Issue #178, opened 2025-01-20
Restore some faucet tests.


sstanfield:
In order for 'cargo test' to succeed when run in an environment without access to faucet infrastructure four tests were disable temparily.

These tests should not run in general with vanilla cargo test to allow external developers to work on the repo but need to run for final checks by maintainers.  The crates may need to be reorganized or other steps taken to properly isolate then.

The tests to re-enable are in crates/execution/faucet/ and can be found by searching for "should not run with a default cargo test" in that directory.
Two are in crates/execution/faucet/tests/it/faucet.rs and two are in crates/execution/faucet/src/lib.rs.

PR #184, opened 2025-01-21
Run some tests that require credentials during attestation.

sstanfield:
This moves some of the test ignored because they needed some credentials, etc to make cargo test work out of box for contributors to make attest.  These can only be run by a core dev with credentials anyway so make sure they get run at some point.

Closes issue 178 https://github.com/Telcoin-Association/telcoin-network/issues/178


grantkee (approved):
LGTM!

PR #188, opened 2025-01-27
Upgrade reth: v1.1.5

grantkee:
- upgrade reth v1.1.5
- refactor tn_node::engine with compatibility traits
- upgrade rust 1.82
- centralized exports for reth_primitives / alloy types from tn-types
- distinguish `reth::Header` as `ExecHeader` (execution layer)
- nuke `Cargo.toml` and reinstall latest minor versions


crates/engine/src/lib.rs
```diff
@@ -145,7 +145,8 @@ where
             tokio::task::spawn_blocking(|| {
                 // this is safe to call on blocking thread without a semaphore bc it's held in
                 // Self::pending_tesk as a single `Option`
-                let result = execute_consensus_output(evm_config, build_args);
+                let evm_config = evm_config;
```
sstanfield, line 148: This looks pointless?

crates/node/src/engine/worker.rs
```diff
@@ -66,93 +59,45 @@ impl<DB> WorkerComponents<DB> {
...
     fn local_addr(&self) -> SocketAddr {
         (IpAddr::from(std::net::Ipv4Addr::UNSPECIFIED), DEFAULT_DISCOVERY_PORT).into()
     }
 
+    #[allow(deprecated, reason = "EthProtocolInfo::difficulty is deprecated")]
     async fn network_status(&self) -> Result<NetworkStatus, NetworkError> {
-        let chain_spec = adiri_chain_spec();
         Ok(NetworkStatus {
-            client_version: "adiri".to_string(),
+            client_version: "v0.0.1".to_string(),
```
sstanfield, line 87: Not for this PR but we may want to set versions like this from the Cargo.toml versions.

crates/types/Cargo.toml
```diff
@@ -34,7 +34,7 @@ serde_with = { workspace = true }
 serde_json = { workspace = true }
 serde_repr = { workspace = true }
 thiserror = { workspace = true }
-tokio = { workspace = true, features = ["sync", "rt", "macros"] }
+tokio = { workspace = true, features = ["sync", "rt", "macros", "signal"] }
```
sstanfield, line 37: I may have missed it but why the addition of signal to this crate? Not a big deal but may not be needed.

sstanfield (approved):
This is great.

crates/node/src/engine/worker.rs
grantkee: Yes, I agree. The worker's entire RPC needs to be revisited, so I just made an issue #189

crates/engine/src/lib.rs
grantkee: Ah, I agree but it doesn't implement copy so there was an ownership issue. I meant to come back and try `move` before opening this PR

Good catch - I added move to take ownership and got rid of that silly line

crates/types/Cargo.toml
grantkee: I was a compile error for `tokio::signal::ctrl_c` so I added signal

I knew I was forgetting something! I totally nuked `Cargo.lock` and reinstalled latest minor versions as well

That being said, tokio is on 1.43 so we should target a general update soon while reth is still "fresh"

sstanfield (approved):

Issue #186, opened 2025-01-24
Start existing validator with empty DBs.


sstanfield:
If a validator is "reset" by removing it's DBs (execution and consensus) it should be able to restart by applying consensus output to an empty chain.

Reth does not appear to be initialized properly in this case causing errors, main problem seems to be:
ts=2025-01-24T23:02:54.916254Z·level=warn·target=engine·message="failed·to·get·state·for·empty·output"·parent_hash=0xad4b7eadb9d46ec4cba805f255b9fe2c3c3bbd94071d7b0a6174dc4e3ccf0a1e·err="block·hash·0xad4b7eadb9    d46ec4cba805f255b9fe2c3c3bbd94071d7b0a6174dc4e3ccf0a1e·does·not·exist·in·Headers·table
See payload_builder.rs
Also see start_engine() in engine/inner.rs- seems to get an incorrect head (?).

PR #192, opened 2025-01-28
Last fixes to allow startup of a validator with an empty DB.

sstanfield:
https://github.com/Telcoin-Association/telcoin-network/issues/186

Closes the gap on being able to restart after deleting all data.  Note all data includes these dirs:
static_files/
db/
consensus-db/

Not removing all will lead to issues.  This should also apply later to add a new validator after he chain is started.


grantkee (approved):
LGTM!

Issue #191, opened 2025-01-28
Need to clean up batch fetching in executor subscriber.


sstanfield:
Start at fetch_batches().  Note that this functions perform network IO but despite that it does not return an error.

Clean this up to avoid panics induced by outside input or nodes not responding and to basically use errors in network facing code like a sane program.

While here examine any other panic inducing code to make sure it should truly panic and is not just lazy error handling.

PR #193, opened 2025-01-29
Refactor subscriber to support future error handling and libp2p migration.

sstanfield:
This is basically a pre-cursor refactor for https://github.com/Telcoin-Association/telcoin-network/issues/191 that should make it easier to move to libp2p as well.


grantkee (approved):
Nice. LGTM! This will definitely help moving to libp2p

PR #194, opened 2025-01-31
Add a state-sync crate and refactor more of subscriber into it.

sstanfield:
This is decoupling a lot of the networking from subscriber functionality to be cleaner and make the libp2p port easier.

This puts the network facing (a lot anyway) of subscriber into a new state-sync crate and simplifies some of subscriber to just the core logic.

The new crate is all functions currently.  Also the spawning is slightly odd because of the need to make sure no messages are missed by some of the dependent tasks in subscriber.


grantkee (approved):
LGTM!

PR #201, opened 2025-02-01
Atomic GC Round

grantkee:
- Use atomic u32 for gc round
- helper method on consensus bus
- first step towards synchronizer refactor

### DIFF
--- a/crates/consensus/primary/src/consensus/state.rs
+++ b/crates/consensus/primary/src/consensus/state.rs
@@ -253,7 +253,8 @@ impl ConsensusRound {
         Self { committed_round, gc_round: atomic_gc_round }
     }
 
-    /// Calculates the latest CommittedRound by providing a new committed round and the gc_depth by comparing the existing committed round against the new.
+    /// Calculates the latest CommittedRound by providing a new committed round and the gc_depth by
+    /// comparing the existing committed round against the new.
     fn update(&mut self, new_committed_round: Round, gc_depth: Round) {
         let last_committed_round = max(self.committed_round, new_committed_round);
         let last_gc_round = gc_round(last_committed_round, gc_depth);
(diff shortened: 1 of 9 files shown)


crates/consensus/primary/src/consensus/state.rs
```diff
@@ -234,31 +234,77 @@ impl ConsensusState {
 /// When a certificate gets committed then
 /// the corresponding certificate's round is considered a "committed" round. It bears both the
 /// committed round and the corresponding garbage collection round.
-#[derive(Debug, Default, Copy, Clone)]
+#[derive(Debug, Default, Clone)]
 pub struct ConsensusRound {
     pub committed_round: Round,
-    pub gc_round: Round,
+    pub gc_round: AtomicRound,
```
sstanfield, line 240: I assume the need to make this atomic is not obvious from this PR?  Loosing Copy on a simple data structure like ConsensusRound seems suboptimal but I guess if there is some driver for it then it is what it is.
Also, it seems odd that one field is atomic and one is not.  Given the cloning the loss of copy forces not sure this is useful (or is that why AtomicRound has an Arc<Atomic>>?  Do you really just want to keep this copy and wrap in an Arc at a higher level?

crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -282,6 +282,11 @@ impl ConsensusBus {
         &self.inner.tx_consensus_round_updates
     }
 
+    /// Load the atomic GC round for consensus.
+    pub fn atomic_gc_round(&self) -> u32 {
```
sstanfield, line 286: Given the convoluted path to get to this, this seems fine.  In general I like to keep helpers down on the bus so it does not become a confused pile of "helpers" by returning more generic things with the helpers (ie the TnSender/Reciever traits).  This is just a general comment though, in this case this may be the best option.

grantkee:
Several tasks within the primary track the committed rounds and gc round, but it's only set by consensus after a commit. This seemed like a good application for the `AtomicRound` but tasks still needed notification when the value changed, thus I stuck it inside a watch channel.

Still, you bring up good points and I want to avoid anti-patterns. I'm closing in favor of #205

PR #202, opened 2025-02-01
separate watch channels for committed round updates and gc round updates

grantkee:
- Separate gc and committed rounds into different watch channels
  - helper method for consensus_bus to `update_consensus_rounds` that sends on both gc/committed round watch channels
- progress towards separating synchronizer
- also supports `borrow_and_update()` now since single consumer for each watch channel
  - previously it was only safe to call `borrow` because some watchers needed the commit round while others needed the gc round. this separates the two

### DIFF
--- a/crates/consensus/primary/src/consensus_bus.rs
+++ b/crates/consensus/primary/src/consensus_bus.rs
@@ -3,8 +3,10 @@
 //! arguments.
 
 use crate::{
-    certificate_fetcher::CertificateFetcherCommand, consensus::ConsensusRound,
-    proposer::OurDigestMessage, RecentBlocks,
+    certificate_fetcher::CertificateFetcherCommand,
+    consensus::{AtomicRound, ConsensusRound},
+    proposer::OurDigestMessage,
+    RecentBlocks,
 };
 use consensus_metrics::metered_channel::{self, channel_with_total_sender, MeteredMpscChannel};
 use std::sync::{atomic::AtomicBool, Arc};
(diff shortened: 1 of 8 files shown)


crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -55,10 +60,16 @@ struct ConsensusBusInner {
...
+
     /// Outputs the highest committed round & corresponding gc_round in the consensus.
-    tx_consensus_round_updates: watch::Sender<ConsensusRound>,
+    tx_committed_round_updates: watch::Sender<Round>,
     /// Hold onto a receiver to keep it "open".
-    _rx_consensus_round_updates: watch::Receiver<ConsensusRound>,
+    _rx_committed_round_updates: watch::Receiver<Round>,
+
+    /// Outputs the highest gc_round from the consensus.
+    tx_gc_round_updates: watch::Sender<AtomicRound>,
```
sstanfield, line 70: I think putting an atomic in a watch is a bit of an anti-pattern.  What is the purpose?  The watch will give you the latest thing, are you trying to change a value underneath code that has already queried the watch?  It seems like using the latest value when you check the watch should always be sufficient?
Also, being able to take a value from a watch and then changing it via an atomic store is violating the watch contract since subscribers will not be notified so I really think a watch should not contain an atomic.

crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -277,14 +291,27 @@ impl ConsensusBus {
         &self.inner.parents
     }
 
+    /// Update consensus round watch channels.
+    pub fn update_consensus_rounds(&self, update: ConsensusRound) -> eyre::Result<()> {
```
sstanfield, line 295: Don;t love helpers on the bus but in case I think it makes sense.

sstanfield (commented):
I am probably missing something about why you are making GC rounds atomic but I think there are some issues with that in this and the previous PR.

PR #207, opened 2025-02-04
Initial integration of new libp2p lib and consensus.

sstanfield:
Does not work yet (needs to take over the primary address).  But useful for feedback on the direction of the integration.
This will
- put gossip on a channel on the consensus bu
- has the hook on primary to implement request
- will need to wrap the NetworkHandle to recreate the "client" for remote calls

These would all be pretty straightforward.  This break out the initial scaffolding code it's own PR.


crates/consensus/primary/src/primary.rs
```diff
@@ -194,6 +225,41 @@ impl<DB: Database> Primary<DB> {
...
+                NetworkEvent::Request { peer, request, channel, cancel } => tokio::select!(
+                    _ = cancel => {}
+                    _ = Self::event(peer, request, channel, config.clone(), consensus_bus.clone()) => {}
+                ),
+                NetworkEvent::Gossip(message) => {
+                    let _ = consensus_bus.consensus_network_gossip().send(message).await;
+                }
+            }
+        }
+    }
```
grantkee, line 246: Please see `PrimaryNetwork::spawn` https://github.com/Telcoin-Association/telcoin-network/blob/main/crates/consensus/primary/src/network/mod.rs#L61-L73

crates/consensus/primary/src/primary.rs
```diff
@@ -43,12 +49,20 @@ pub mod primary_tests;
...
+    network_event_stream: Option<mpsc::Receiver<NetworkEvent<PrimaryRequest, PrimaryResponse>>>,
 }
 
 impl<DB: Database> Primary<DB> {
-    pub fn new(config: ConsensusConfig<DB>, consensus_bus: &ConsensusBus) -> Self {
+    pub fn new(
+        config: ConsensusConfig<DB>,
+        consensus_bus: &ConsensusBus,
+        network_p2p_handle: NetworkHandle<PrimaryRequest, PrimaryResponse>,
+        network_event_stream: mpsc::Receiver<NetworkEvent<PrimaryRequest, PrimaryResponse>>,
```
grantkee, line 64: I think it makes sense to pass the handle and event stream as args here. 

The only other place where it *might* make sense is if the consensus bus owned these values. However, I don't like that idea since that means the bus is spawning the network. I think this makes the most sense

crates/network-libp2p/src/consensus.rs
```diff
@@ -215,21 +212,19 @@ where
...
-                        }
+    pub async fn run(mut self) -> NetworkResult<()> {
+        loop {
+            tokio::select! {
+                event = self.swarm.select_next_some() => self.process_event(event).await?,
+                command = self.commands.recv() => match command {
+                    Some(c) => self.process_command(c),
+                    None => {
+                        info!(target: "network", topics=?self.topics, "subscriber shutting down...");
+                        return Ok(())
```
grantkee, line 223: Good idea

crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -129,6 +130,10 @@ struct ConsensusBusInner {
 
     /// Flag to indicate a node should restart after a shutdown.
     restart: AtomicBool,
+    /// Consensus output with a consensus header.
+    consensus_network_gossip: broadcast::Sender<GossipMessage>,
```
grantkee, line 134: This will come through as a `NetworkEvent` https://github.com/Telcoin-Association/telcoin-network/blob/main/crates/network-libp2p/src/types.rs#L61-L62

crates/consensus/primary/src/primary.rs
```diff
@@ -43,12 +49,20 @@ pub mod primary_tests;
 pub struct Primary<DB> {
     /// The Primary's network.
     network: Network,
+    network_p2p_handle: NetworkHandle<PrimaryRequest, PrimaryResponse>,
     synchronizer: Arc<Synchronizer<DB>>,
     peer_types: Option<HashMap<PeerId, String>>,
+    // Hold onto the network event stream until spawn "takes" it.
+    network_event_stream: Option<mpsc::Receiver<NetworkEvent<PrimaryRequest, PrimaryResponse>>>,
```
grantkee, line 56: Does this need to be an option even though it's passed as an arg in `new`? I think it can be `PrimaryNetwork` and not an option since everything gets passed to `new`

grantkee: Scratch that - `PrimaryNetwork::spawn` consumes itself so `Option` is a good idea

crates/consensus/primary/src/tests/primary_tests.rs
```diff
@@ -33,7 +35,18 @@ use tn_types::{
...
+fn get_bus_and_primary<DB: Database>(config: ConsensusConfig<DB>) -> (ConsensusBus, Primary<DB>) {
+    let (event_stream, rx_event_stream) = mpsc::channel(1000);
+    let consensus_bus = ConsensusBus::new_with_args(config.config().parameters.gc_depth);
+    let consensus_network = ConsensusNetwork::new_for_primary(&config, event_stream)
+        .expect("p2p network create failed!");
+    let consensus_network_handle = consensus_network.network_handle();
+
+    let primary = Primary::new(config, &consensus_bus, consensus_network_handle, rx_event_stream);
+    (consensus_bus, primary)
+}
```
grantkee, line 49: nice!

grantkee (changes requested):
Great to see our ideas are very similar! There's some redundant code you can remove, but everything else looks good

crates/consensus/primary/src/primary.rs
sstanfield: I agree, I explored doing some of this on the consensus bus but this is what made sense.  It work perfectly with your PrimaryNetwork class as well.

crates/network-libp2p/src/consensus.rs
sstanfield: Thanks, wanted to make this task manager friendly since the loss of this task is fatal...

crates/consensus/primary/src/consensus_bus.rs
sstanfield: The intent here is to just provide gossip, not the NetworkEvents.  It is the consensus_bus hook into libp2p gossip.

crates/consensus/primary/src/tests/primary_tests.rs
sstanfield: Yea, I need to take a little more time on some of these test refactors to do stuff like this and hopefully make future refactors easier...

crates/consensus/primary/src/consensus_bus.rs
sstanfield: Actually this should go away for specific versions for things that should go out via the bus.

sstanfield: Leaving for now but I expect this to become more specific channels as we incorporate gossip.

grantkee (approved):
LGTM!

PR #211, opened 2025-02-08
Move the primary network to libp2p.

sstanfield:
Switch the primary consensus network to libp2p.
- Primary (not worker) are using libp2p now.
- Anemo code for the old primary network is removed (still around for the worker).
- Had to play some games to keep local network happy but should be OK.  Can revisit after fulling moving to libp2p.


crates/consensus/primary/src/certifier.rs
```diff
@@ -174,26 +176,24 @@ impl<DB: Database> Certifier<DB> {
...
+                Ok(PrimaryResponse::MissingParents(parents)) => {
+                    debug!(target: "primary::certifier", ?authority, ?parents, "Ok missing parents response received after request vote");
+                    missing_parents = parents;
                 }
-                Err(status) => {
-                    // TODO: why does this error out so much?
-                    error!(target: "primary::certifier", ?authority, ?status, ?header, "bad request for requested vote");
-                    if status.status() == anemo::types::response::StatusCode::BadRequest {
-                        error!(target: "primary::certifier", ?authority, ?status, ?header, "fatal request for requested vote");
+                Ok(_) => panic!("invalid response, this should have been an error!"),
```
grantkee, line 188: We should handle this gracefully with an error log to prevent malicious peers from causing others to panic

crates/consensus/primary/src/network/handler.rs
```diff
@@ -30,7 +30,7 @@ use tracing::{debug, error, warn};
 
 /// The type that handles requests from peers.
 #[derive(Clone)]
-pub(super) struct RequestHandler<DB> {
+pub struct RequestHandler<DB> {
```
grantkee, line 33: any reason to make this more visible than `pub(crate)`?

crates/consensus/primary/src/network/mod.rs
```diff
@@ -44,7 +47,7 @@ where
     DB: Database,
 {
     /// Create a new instance of Self.
-    pub(crate) fn new(
+    pub fn new(
```
grantkee, line 50: Where is this used outside the `tn-primary`?

crates/consensus/primary/src/network/mod.rs
```diff
@@ -57,7 +60,7 @@ where
     }
 
     /// Run the network.
-    pub(crate) fn spawn(mut self, task_manager: &TaskManager) {
+    pub fn spawn(mut self, task_manager: &TaskManager) {
```
grantkee, line 63: Is this also used outside the `tn-primary`?

crates/consensus/primary/src/primary.rs
grantkee, line 1: wow! this really cleaned up nicely

crates/consensus/worker/src/network.rs
```diff
@@ -20,9 +20,10 @@ use tn_storage::{
 use tn_types::{now, BatchValidation, Committee, SealedBatch, WorkerCache, WorkerId};
 use tracing::{debug, trace};
 
+/* TODO restore these when worker moves to libp2p XXXX (should not make it to main)
 #[cfg(test)]
 #[path = "tests/handlers_tests.rs"]
-pub mod handlers_tests;
+pub mod handlers_tests;*/
```
grantkee, line 26: leftovers

crates/consensus/worker/src/tests/handlers_tests.rs
```diff
@@ -61,6 +61,7 @@ async fn synchronize() {
         batch_fetcher: None,
         validator: Arc::new(NoopBatchValidator),
     };
+    //let handler = RequestHandler::new(target.consensus_config(), cb.clone(), synchronizer.clone());
```
grantkee, line 64: leftover

crates/network-libp2p/src/consensus.rs
```diff
@@ -390,7 +390,7 @@ where
                 send_or_log_error!(reply, peer_id, "LocalPeerId");
             }
             NetworkCommand::Publish { topic, msg, reply } => {
-                let res = self.swarm.behaviour_mut().gossipsub.publish(topic, msg);
+                let res = self.swarm.behaviour_mut().gossipsub.publish(topic.clone(), msg);
```
grantkee, line 393: Does topic still need to be cloned here?

crates/network-libp2p/src/types.rs
```diff
@@ -181,6 +183,12 @@ where
         Self { sender }
     }
 
+    /// Create a handle to no where for test setup.
+    pub fn new_for_test() -> Self {
+        let (sender, _) = mpsc::channel(10_000);
```
grantkee, line 188: nit: can safely reduce channel capacity from 10k to 100 for tests

crates/consensus/worker/src/network.rs
sstanfield: No, fixed in next PR.

crates/consensus/worker/src/network/handler.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub(super) async fn process_gossip(&self, msg: &GossipMessage) -> WorkerNetworkResult<()> {
+        // deconstruct message
+        let GossipMessage { data, .. } = msg;
+
+        // gossip is uncompressed
+        let gossip = try_decode(data)?;
+
+        match gossip {
+            WorkerGossip::Batch(_block_hash) => {
+                // Retrieve the block...
```
grantkee, line 52: Do we have an issue for this yet?

crates/consensus/worker/src/network/handler.rs
```diff
@@ -0,0 +1,123 @@
...
+        Ok(())
+    }
+
+    /// Attempt to return requested batches.
+    pub(crate) async fn process_request_batches(
+        &self,
+        batch_digests: Vec<BlockHash>,
+    ) -> WorkerNetworkResult<RequestBatchesResponse> {
+        const MAX_REQUEST_BATCHES_RESPONSE_SIZE: usize = 6_000_000;
+        const BATCH_DIGESTS_READ_CHUNK_SIZE: usize = 200;
```
grantkee, line 92: I'd like to move these into the config, but okay for this PR. We need to figure out how we can separate worker/primary cleanly without creating headaches in the short term and tech debt in the long term. It's complicated because we need the workers to scale as well

Anyway, fine for this PR just thinking out loud

grantkee (approved):
Wow! Very exciting! Great work, Steve!

Issue #196, opened 2025-01-31
Move `tn_storage::traits::Database` into `tn_types::traits`


grantkee:
The TN trait `Database` is only available through import from `tn_storage` which is a large crate.

Some crates don't need `tn_storage` but need the Database trait for generic compatibility. (For example tn-network-libp2p and tn-config)

Moving the trait definition to `tn_types` allows crates to constrain generics with this type without having to import the large `tn_storage` crate.

ArieJones:
It seems that with this change there would be one potential blocker in regards to tn-config because consensus relies on NodeStorage from tn-storage.  

Would that be ok leaving that as-is and just updating the references to the Database traits?

grantkee:
Hi @ArieJones - thanks for looking into this!

I'm not sure I fully understand the issue you're describing? Looking at your PR #218, it looks like you're on the right track for what I had in mind. `tn-config` and `tn-storage` both have `tn-types` as dependencies, so there shouldn't be any circular issues.

Are you pointing out that `tn-config` still needs `tn-storage` as a dependency since the `ConsensusConfig` uses `NodeStorage`?

ArieJones:
@grantkee yeah just that I won't be able to remove the dependency from everything or at least move them down to a dev dependency where they are only in tests.

grantkee:
Okay, good thinking but it's okay to keep `tn-storage` as a dependency in `tn-config` for now. Thanks again for your help!

PR #218, opened 2025-02-17
Move database traits to a new module in `tn_types`.

ArieJones:
Addresses Issue #196 

- [ ] Refactored database trait definitions out of `tn_storage` into `tn_types::database_traits` for better modularity. 
- [ ] Updated all relevant imports across the codebase and ensured compatibility with the new module structure. 
- [ ] Updated certifier.rs to fix a complaint about type matching


sstanfield (approved):
@ArieJones Thanks for this.  I think it looks good (@grantkee will probably also want to review), just FYI there may be a delay merging while we work out some details this exposed.  You're the first external PR and we have an "attestation" process (basically one of use has to run the tests that require authentication and post the hash to test net on success so CI can pass).  Since external contributors use their own fork we will probably have to tweak the process.  We are going to review this later today.

ArieJones:
@sstanfield Thanks Steve! Let me know if you need me to do anything else with this though

grantkee (approved):
Nice work. Thanks for your help, @ArieJones!

PR #225, opened 2025-02-21
Agnostic txs worker batch

grantkee:
- use bytes for worker transactions
- set global threadpool limits for rayon
   - currently used in batch validator


crates/execution/batch-validator/src/validator.rs
```diff
@@ -108,15 +110,66 @@ where
...
+        transactions: &Vec<Vec<u8>>,
+        digest: BlockHash,
+    ) -> BatchValidationResult<Vec<TransactionSigned>> {
+        if transactions.len() < *PARALLEL_SENDER_RECOVERY_THRESHOLD {
+            transactions
+                .iter()
+                .map(|tx| Self::recover_and_validate(tx, digest))
+                .collect::<BatchValidationResult<Vec<_>>>()
+        } else {
+            transactions
```
sstanfield, line 150: I think you are using rayon ok here but just want to point out that mixing rayon (a thread pool) and Tokyo (also a thread pool and async runtime) can lead to nasty interactions and bugs.  Again, I can't see how this simple use would be a problem but it might make it easier to foot gun ourselves later if we start using rayon more.  Just something worth considering.

sstanfield (dismissed):
Looks good although I am wary of introducing rayon to a tokio project (see comment).

sstanfield (approved):

PR #227, opened 2025-02-26
Publish consensus headers headers and start using it.

sstanfield:
Publish the block num and hash for consensus headers.
When catching up and following consensus use this to improve and stream line state sync.


grantkee (approved):
LGTM! Nice work

Issue #232, opened 2025-02-26
Provide a fully validated latest block for bridging.


sstanfield:
A node following consensus could advertise blocks that may have been caused by malicious peers.  This will be detected very quickly but currently there is a window.  Certificates include a latest executed block num hash, this can be used to only commit (or otherwise advertise) blocks that a quorum of the current committee has signed off on as valid.

Observer nodes are currently be worked on but this is an important security feature to support bridging and other use cases that require fully vetted blocks. 

Note, this may be implemented in such a way as to impact committed reth blocks even for validators.

PR #234, opened 2025-02-28
If not a CVV then do not finalize a block until we see it in a certificate.

sstanfield:
https://github.com/Telcoin-Association/telcoin-network/issues/232

This tests that setting the right flag does not finalize but this we should add some integration tests once we can run "observer" nodes to test this in the wild.

The implementation turned out to be pretty simple and basically contained in the engine.


grantkee (approved):
LGTM! Clean solution

PR #235, opened 2025-03-04
Use triats instead of wrappers for DB functions.

sstanfield:
This is a pure refactor, no new functionality.  Instead of using wrapper structs for DB functions put them in traits.  I think this is simpler and will be less hassle to add new functions (can just add a trait and implement for Database) vs adding a new wrapper and putting in NodeStorage, etc.

Ran across this when I was working batch fetching and wanted to add a new DB function.

NOTE: This stacks on my previous PR.


grantkee (dismissed):
That cleaned up nicely! I didn't see any issues. I think this is good to go

grantkee (approved):

PR #236, opened 2025-03-07
Reworked batch fetching to work with NVV and observer nodes.

sstanfield:
This was a response to https://github.com/Telcoin-Association/telcoin-network/issues/191
It does not add errors since we need batches to progress.  It does the following:
- Asks all its connected peers for batches vs just workers.  If not a CVV then it may not even have access to a particular worker (or any) so this makes it more general to handle NVV and observer nodes as well.
- Refactors some of the error checking into lower level code.
- Changed the test harness code to test more of the fetching stack.

This PR basically sends a batch request to every connected peer.  This is similar to how it used to use workers, it this what we want to do?

### DIFF
--- a/crates/consensus/worker/src/network/mod.rs
+++ b/crates/consensus/worker/src/network/mod.rs
@@ -1,6 +1,7 @@
 use std::{collections::HashSet, sync::Arc, time::Duration};
 
 use error::WorkerNetworkError;
+use futures::{stream::FuturesUnordered, StreamExt};
 use handler::RequestHandler;
 use message::{RequestBatchesResponse, WorkerGossip, WorkerRPCError};
 pub use message::{WorkerRequest, WorkerResponse};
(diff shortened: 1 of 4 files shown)


crates/consensus/worker/src/network/mod.rs
```diff
@@ -98,17 +99,96 @@ impl WorkerNetworkHandle {
...
     }
+
+    /// Request a group of batches by hashes.
+    /// Sends request to all our connected peers at once and returns Ok when we
+    /// get a valid response or Err if no one responds with the batches.
+    pub async fn request_batches_from_all(
+        &self,
+        batch_digests: Vec<BlockHash>,
+    ) -> NetworkResult<RequestBatchesResponse> {
+        let peers = self.handle.connected_peers().await?;
```
grantkee, line 155: It might be worth adding another error type here for `NoPeers` in case `peers` is empty to distinguish between the peer doesn't have any peers vs none of its peers returned batches

crates/consensus/worker/src/network/mod.rs
```diff
@@ -98,17 +99,96 @@ impl WorkerNetworkHandle {
...
+        let mut futures = FuturesUnordered::new();
+        for peer in peers {
+            futures.push(self.request_batches(peer, batch_digests.clone()));
```
grantkee, line 158: I'm glad you took this approach initially because it reduced the complexity to evaluate all the other logic in this PR 🙏

I think the next step is to reduce the redundant messaging in the network. This approach is likely to work, but if we have 5 batches at 1MB each, 5 peers (likely more): the request could generate 25MB download and 20MB would be redundant

I haven't thought of a simple solution to address this. I'm leaning towards partitioning the request amongst peers. Using the example above, we could take modulo batches length and number peers so each peer would receive a request for one batch

The complexity then becomes handling errors/failures. Even though that is more complex, I think it's still important to reduce the amount of redundant messages in our sync strategy. Let's discuss this more

grantkee: I think this approach is okay for this PR because we do not support public nodes yet, but we need to address this soon

grantkee (dismissed):
I approved in case you'd rather address some of these in a follow-up PR

I'd prefer having targets for any logs so we can filter them out. I think this is a decent initial approach. Fetching batches is something we should discuss further before officially supporting public nodes. The current approach is likely to have a significant amount of redundant data for nodes with a healthy peer network.

crates/consensus/worker/src/network/mod.rs
sstanfield: Actually an error here means we probably lost the network or something (got not response on the peers query).  I am going to tighten this up a bit anyway.  If there are no peers I think I will just respond with nothing, that should cause the higher level code to automatically try again when hopefully we have some peers...

grantkee (approved):

Issue #221, opened 2025-02-18
In Worker BatchProvider is the Worker now.


sstanfield:
There is a Worker structure and a BatchProvider structure but Worker is empty and just makes a BatchProvider.  Make BatchProvider the Worker.
- Bring over the functions form Worker that create BatchProviders to Worker.
- Rename BatchProvider (file and structure) Worker.

This will remove one source file from Worker.

PR #238, opened 2025-03-11
Batch fetching is peer agnostic and handles errors better now.

sstanfield:
Includes https://github.com/Telcoin-Association/telcoin-network/issues/221 - leftover from a previous refactor.

Batch fetching should work for any kind of node now, not just committee members.
Also improve the error handling, https://github.com/Telcoin-Association/telcoin-network/issues/191 .


grantkee (approved):
LGTM

PR #239, opened 2025-03-14
Better batch fetching.

sstanfield:
Addresses comment on last PR stack.
- When requesting batches from any peer divide the list up instead of request all batches from all peers.
- Remove the batch response type, the extra field could not be trusted and was never used so just return the vector of batches we could get which may or may not be complete.  All the higher level code already assumed this.


grantkee (dismissed):
Nice work!

grantkee (approved):

Issue #149, opened 2025-01-09
State sync: Store gossip while syncing backlog


grantkee:
Store gossip while syncing historic state so nodes don't fall further behind while they're syncing.

PR #241, opened 2025-03-14
Publish and use batch gossip.

sstanfield:
This is useful for non-CVVs to preemptively get batches.

- A worker will publish a batch hash after it has been accepted by the committee.
- Other nodes will confirm the gossip originated with the committee to help avoid any spoofing.
- Snuck a tiny consensus bus refactor in, no need to hold onto broadcast channel receivers, don't worry about not having anyone to send to.

This supports https://github.com/Telcoin-Association/telcoin-network/issues/149

### DIFF
--- a/crates/consensus/worker/src/network/handler.rs
+++ b/crates/consensus/worker/src/network/handler.rs
@@ -11,6 +11,7 @@ use tn_types::{
 use super::{
     error::{WorkerNetworkError, WorkerNetworkResult},
     message::WorkerGossip,
+    WorkerNetworkHandle,
 };
 
 /// The type that handles requests from peers.
@@ -22,6 +23,8 @@ pub struct RequestHandler<DB> {
     validator: Arc<dyn BatchValidation>,
     /// Consensus config with access to database.
     consensus_config: ConsensusConfig<DB>,
+    /// Network handle- so we can respond to gossip.
+    network_handle: WorkerNetworkHandle,
 }
 
 impl<DB> RequestHandler<DB>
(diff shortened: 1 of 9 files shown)


crates/consensus/worker/src/network/handler.rs
```diff
@@ -33,25 +36,54 @@ where
...
         // gossip is uncompressed
         let gossip = try_decode(data)?;
 
         match gossip {
-            WorkerGossip::Batch(_block_hash) => {
-                // Retrieve the block...
+            WorkerGossip::Batch(batch_hash) => {
+                // Only accept batch gossip from the committee
+                if let Some(source) = source {
+                    if self.consensus_config.committee_peer_ids().contains(source) {
```
grantkee, line 59: This check is already performed before `process_gossip` in the application layer is notified of the new gossip

The [verify_message](https://github.com/Telcoin-Association/telcoin-network/blob/main/crates/network-libp2p/src/consensus.rs#L605-L623) method is [called at the libp2p-network](](https://github.com/Telcoin-Association/telcoin-network/blob/main/crates/network-libp2p/src/consensus.rs#L501-L502)) level anytime a gossip message is received to validate the gossip

grantkee (changes requested):
This is the right idea. The only change I request is removing the additional authorized publisher check inside `process_gossip`

crates/consensus/worker/src/network/handler.rs
sstanfield: Roger, missed that lower level check.

grantkee (approved):
Nice! Great work

PR #242, opened 2025-03-21
Misc refactors and moved network keys to new native libp2p.

sstanfield:
- Add a -observer command line arg (not fully functional yet).
- Minor version bump of libp2p and a couple fixes for that.
- Move Network Keys to native libp2p format and away from fast crypto.
- Derive network keys from a nodes BLS primary key now, no need to save them on disk and should be helpful when we properly protect the BLS key.
- Refactors around KeyConfig so it generates and saves the keys now itself (as well as loads them).  Get this functionality in one place.

### DIFF
--- a/crates/config/src/keys.rs
+++ b/crates/config/src/keys.rs
@@ -1,13 +1,16 @@
 //! Cryptographic keys used by the node.
 
-use crate::{
-    read_validator_keypair_from_file, TelcoinDirs, BLS_KEYFILE, PRIMARY_NETWORK_KEYFILE,
-    WORKER_NETWORK_KEYFILE,
-};
+use crate::{TelcoinDirs, BLS_KEYFILE};
+use blake2::Digest;
+use rand::{rngs::StdRng, SeedableRng};
+use rand_chacha::ChaCha20Rng;
+use reth_chainspec::ChainSpec;
 use std::sync::Arc;
 use tn_types::{
-    traits::{AllowedRng, KeyPair, Signer},
-    BlsKeypair, BlsPublicKey, BlsSignature, BlsSigner, NetworkKeypair, NetworkPublicKey,
+    encode,
+    traits::{AllowedRng, EncodeDecodeBase64 as _, KeyPair, Signer, ToFromBytes},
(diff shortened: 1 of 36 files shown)


crates/types/src/crypto/mod.rs
```diff
@@ -60,17 +62,91 @@ pub type ExecutionPublicKey = secp256k1::PublicKey;
...
+        }
+
+        if deserializer.is_human_readable() {
+            deserializer.deserialize_str(NetworkPublicKeyVisitor)
+        } else {
+            deserializer.deserialize_bytes(NetworkPublicKeyVisitor)
+        }
+    }
+}
+
```
grantkee, line 149: nit: move this to the `crypto::network.rs` mod

crates/config/src/keys.rs
```diff
@@ -1,47 +1,86 @@
...
+                primary_network_keypair,
+                worker_network_keypair,
+            }),
+        })
+    }
+
+    /// Generate a new random primary BLS key and save to the config file.
+    /// Note, this is not very secure in that it is writing the private key to a file...
+    pub fn generate_and_save<TND: TelcoinDirs>(tn_datadir: &TND) -> eyre::Result<Self> {
+        // TODO: discuss with @Utku
```
grantkee, line 70: nit: I think we should remove the specific user and just say "security"/"auditor" or something similar. If you think it should still be directed only towards Utku, then please use his username `utkutugrul`

crates/config/src/keys.rs
```diff
@@ -87,25 +125,55 @@ impl KeyConfig {
...
+        let genesis_bytes = encode(&chain_spec.genesis);
+        msg.extend_from_slice(genesis_bytes.as_slice());
+        let sig = BlsSignature::new_secure(
+            &IntentMessage::new(Intent::telcoin(IntentScope::ProofOfPossession), msg),
+            &self.inner.primary_keypair,
+        );
+        Ok(sig)
+    }
+
+    /// Derive a NetworkKeypair from a BLS signatur, seed string and blake2 hash.
```
grantkee, line 170: nit: instead of saying `blake2` let's just use `[DefaultHashFunction]` so this comment is still valid if/when we upgrade to blake3, etc

grantkee (approved):
Nice work! I have a few nits, but nothing major

PR #243, opened 2025-03-25
Remove the fast crypto crate.

sstanfield:
- Replace the fast crypto BLS key support with the underlying BLST crate support.
- Bring in the Hash trait and Digest structure since these were core for us and small.
- Tweak a few package versions to clear dependabot alerts.
- Remove "systemMessages'- they were never used and they were heavily invested in fast crypto.

### DIFF
--- a/crates/types/src/crypto/bls_keypair.rs
+++ b/crates/types/src/crypto/bls_keypair.rs
@@ -20,7 +20,6 @@ impl BlsKeypair {
     pub fn generate<R: CryptoRng + RngCore>(rng: &mut R) -> Self {
         let mut ikm = [0u8; 32];
         rng.fill_bytes(&mut ikm);
-        // TODO: Consider moving to key gen version 5.
         let private = BlsPrivateKey::key_gen(&ikm, &[]).expect("ikm length should be higher");
         let pubkey = private.sk_to_pk();
         let mut bytes = [0_u8; 96];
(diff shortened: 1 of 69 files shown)


crates/types/src/crypto/bls_keypair.rs
```diff
@@ -0,0 +1,57 @@
...
+const _DST_G2: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_"; // min pk
+impl BlsKeypair {
+    pub fn public(&self) -> &BlsPublicKey {
+        &self.public
+    }
+
+    pub fn generate<R: CryptoRng + RngCore>(rng: &mut R) -> Self {
+        let mut ikm = [0u8; 32];
+        rng.fill_bytes(&mut ikm);
+        // TODO: Consider moving to key gen version 5.
```
grantkee, line 23: What is key gen version 5?

grantkee (dismissed):
Great work! I have one question about the bls key gen version 5, but it's a non-blocker. I just wasn't sure what it was

Also wanted to mention that I plan to still use `SystemMessages` in the Header, but it will be significantly different so better to start with a clean slate

crates/types/src/crypto/bls_keypair.rs
sstanfield: Not sure, that was old comment that I brought to a new place...  I will see about figuring it out but otherwise will remove it since it may be non-sense at this point.

sstanfield: Obvious difference is v5 takes a salt.  Since this is a random key I am not sure what that would help.  I will remove the comment, this key gen code is something that should have a spotlight on it during an audit as well.

grantkee: I added it to our audit notes and agree a salt is not a priority for this random key gen

grantkee (approved):

PR #244, opened 2025-03-28
Refactor the committee for interior mutability in prep for epoch support.

sstanfield:
Basically what the name says, breaking this down into a PR stack for easier review.

This should be fine standalone but will be a base PR.

### DIFF
--- a/crates/config/src/consensus.rs
+++ b/crates/config/src/consensus.rs
@@ -44,7 +44,7 @@ where
         key_config: KeyConfig,
     ) -> eyre::Result<Self> {
         // load committee from file
-        let mut committee: Committee =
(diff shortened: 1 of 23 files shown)


crates/config/src/consensus.rs
```diff
@@ -175,6 +183,7 @@ where
 
     /// Committee network peer ids.
     pub fn committee_peer_ids(&self) -> HashSet<PeerId> {
+        // XXXX- authority map should live on committee or better yet go away.
```
sstanfield, line 186: This comment will be dealt with in an upcoming PR in the stack.

grantkee (approved):

Issue #200, opened 2025-01-31
Use `PeerId` for AuthorityIdentifier


grantkee:
PeerId and AuthorityIdentifier should be consolidated to just PeerId.

Remove AuthorityIdentifier and just use PeerId to identify Authorities.

ArieJones:
@grantkee are you looking to have AuthorityIdentifier replace in the Id with PeerId and then everything rewired or are you just wanting to dump the Id field for the Authorities altogether?

ArieJones:
@sstanfield lol . Ok gotcha. I thought that is what all of these end up being "interesting" issues in the end. Thanks for the clarification.

sstanfield:
Just found this comment while working on something else (test-utils/src/builder.rs- around 130 I am editing it):
// Reset the authority ids so they are in sort order.  Some tests require this.

So yes, some tests will probably get flakey when the authorities are not in the expected order which I think this issue will cause.  I am sure they can be fixed but you may not want to take that on.  Or maybe you do, nothing like pain as a teacher...

I think maintaining sub-optimal stuff just for tests is itself sub-optimal so this should be addressed but it may not make this as "beginner friendly" as intended.

ArieJones:
Yeah I noticed that and the check in committee.rs to check.. so we'll just see where this goes. Just starting to look into it. 
Thanks!

sstanfield:
Sounds good and good luck.  We aren't trying to be sadistic with these issues...  It may not be a big deal at all, something will probably have to be done about the test(s) that expect authorities in a certain order but there should be more than one way to handle that.

ArieJones:
@grantkee i am just checking to ensure that there is not a need to have the authorities ordered necessarily by their order of insertion. Just getting back to this and fixing things around the the tests.I have a few changes around existing code so thought it better at ask ahead of time.

Thanks!

sstanfield:
@ArieJones Grant is at a conference this week so may not be as responsive as usual.  I believe that is true, the order of the authorities should not matter.  There is at least one test that is order dependent but I don't think we should maintain that just for a test(s).  I would say to not worry about the order and see what shakes out.  I think you will need to fix one or two tests but if everything else works it should be good (maybe you can fix it with a deterministic auth id and possibly order?).  When we move to epochs and committee rotation I also think having a fixed order will not be ideal.

ArieJones:
With these types of tests on sizing. How are we calculating what the max_chunk_size should be? 

[...]

PR #245, opened 2025-03-28
Make authority ids peer ids.

sstanfield:
Part of committee refactor for epoch support.  Addresses https://github.com/Telcoin-Association/telcoin-network/issues/200 .

Make the AuthorityId a PeerId. 
- Wrapped the PeerId in the AuthorityIdentifier in an Arc.  This reduces the size but removes Copy from AuthorityIdentifier.
- Use BTrees to sort the authorities by their ids, some code and many tests need this.
- Removes the old mapping from peer to auth ids- they can just "into" each other now.

### DIFF
--- a/crates/types/src/primary/certificate.rs
+++ b/crates/types/src/primary/certificate.rs
@@ -18,7 +18,10 @@ use crate::{
 use base64::{engine::general_purpose, Engine};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
-use std::{collections::VecDeque, fmt};
+use std::{
+    collections::{BTreeMap, VecDeque},
+    fmt,
+};
 
 /// Certificates are the output of consensus.
 /// The certificate issued after a successful round of consensus.
@@ -61,7 +64,7 @@ impl Certificate {
         header: Header,
         votes: Vec<(AuthorityIdentifier, BlsSignature)>,
     ) -> DagResult<Certificate> {
-        Self::new_unsafe(committee, header, votes, true)
+        Self::new_unsafe(committee, header, votes.into_iter().collect(), true)
     }
 
     /// Create a new, unsafe certificate that does not check stake.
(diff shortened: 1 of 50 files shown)


crates/types/src/primary/certificate.rs
```diff
@@ -70,17 +73,16 @@ impl Certificate {
...
-        Self::new_unsafe(committee, header, votes, false)
+        Self::new_unsafe(committee, header, votes.into_iter().collect(), false)
     }
 
     /// Create a new certificate without verifying authority signatures.
     fn new_unsafe(
         committee: &Committee,
         header: Header,
-        mut votes: Vec<(AuthorityIdentifier, BlsSignature)>,
+        votes: BTreeMap<AuthorityIdentifier, BlsSignature>,
```
grantkee, line 83: Instead of `BTreeMap`, let's make this a `VecDeque` and remove L86

`new_unsafe` is only called by `Certificate::new_unverified` and `Certificate::new_unsigned`

`new_unverified` is used in production and passes a `Vec` without having to iter/collect, but `new_unsigned` is only used in tests (usually with `Vec::new()` for votes)

grantkee (changes requested):
One small change from `BTreeMap` to `VecDeque`. Everything else LGTM!

crates/types/src/primary/certificate.rs
grantkee: This would save us an extra iter/collect

sstanfield: The point here was to force this to be sorted in the correct order.  I will double check but I think we need the BTree so that we get a consistent ordering of authorities otherwise stuff will break.

grantkee (approved):

crates/types/src/primary/certificate.rs
grantkee: Yes, good catch. BTreeMap is needed to sort, but we can remove the extra VecDeque. Well done

PR #246, opened 2025-03-31
V0.3.2 adiri patch

grantkee:
- adjust network config
- remove application penalty when processing gossip. this will be handled in a follow-up PR


grantkee:
Skipping attestation due to significant clippy updates on main that can't be merged into this branch without breaking testnet

PR #248, opened 2025-04-02
Final small cleanups to committee refactor.

sstanfield:
- A couple of renames I forgot in my last PR.
- Use an Arc wrapped "inner" for Authority- it is cloned a lot so this should be better.


grantkee (approved):
LGTM!

PR #251, opened 2025-04-12
Subcribe to Topics

grantkee:
- use `String` for topics to support Eq/Ord traits
- refactor authorized publishers as a map to support different publishers per topic

### DIFF
--- a/crates/config/src/network.rs
+++ b/crates/config/src/network.rs
@@ -104,12 +104,12 @@ pub struct LibP2pConfig {
 
 impl LibP2pConfig {
     /// Return topics for primary.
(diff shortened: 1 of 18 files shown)


crates/config/src/network.rs
```diff
@@ -102,6 +102,18 @@ pub struct LibP2pConfig {
     pub px_disconnect_timeout: Duration,
 }
 
+impl LibP2pConfig {
```
sstanfield, line 105: Name is plural, should these return Vec's for future use?  Obviously ok for now and can be dealt this later.

sstanfield (approved):

Issue #220, opened 2025-02-18
Include `lib` data


grantkee:
Each library in TN should include the following attributes to check for consistency.

Once added, remove deps and add missing fields.

This is best accomplished as a single PR per library to keep reviews as small as possible.

Ensure each `lib.rs` contains these attributes (example: `tn-engine`):

```rust
#![doc(
    html_logo_url = "https://www.telco.in/logos/TEL.svg",
    html_favicon_url = "https://www.telco.in/logos/TEL.svg",
    issue_tracker_base_url = "https://github.com/telcoin-association/telcoin-network/issues/"
)]
#![warn(
    missing_debug_implementations,
    missing_docs,
    unreachable_pub,
    rustdoc::all,
    unused_crate_dependencies
)]
#![deny(unused_must_use, rust_2018_idioms)]
#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
```

grantkee:
@ArieJones I would like the missing doc warning taken care of as well. I'm realize the doc comments can be quite challenging for someone less familiar with the workspace, so I'm happy to help with those if needed.

I'd like to review these as 1 PR per crate and can help edit any doc comments that are complicated/incomplete/missing

PR #255, opened 2025-04-14
Refactor around Reth

sstanfield:
This is the first step in getting Reth isolated into it's own crate that will hopefully allow for:
- easier upgrades
- cleaner code elsewhere

More work can be done, the interface is currently just what was needed and some Reth data still leaks.  Can use this as an opportunity to make a better interface in the future.


crates/execution/batch-builder/src/lib.rs
```diff
@@ -205,7 +125,8 @@ where
         let to_worker = self.to_worker.clone();
 
         // configure params for next block to build
-        let config = PendingBlockConfig::new(self.address, self.latest_canon_state.clone());
+        let config =
+            PendingBlockConfig::new(self.address, self.latest_canon_state().unwrap_or_default());
```
grantkee, line 129: Using the default here would set bad values for the next batch. It's unlikely to happen once the node starts, but if we do something in genesis where we have to wait for a bridge transaction then this could become an issue

Instead, maybe you can store the last canonical update as a field on `Self` and store it on L221 when the state_changed is received? That would also save you the extra db calls on L198-209 for `latest_canon_state`. That data can be sent on L221, so I think it makes sense to use it

crates/execution/faucet/src/lib.rs
```diff
@@ -148,34 +147,24 @@ impl Faucet {
...
-        Pool: TransactionPool<Transaction = EthPooledTransaction> + Unpin + Clone + 'static,
-    {
-        Self::spawn_with(provider, pool, config, TokioTaskExecutor::default())
+    pub(crate) fn spawn(reth_env: RethEnv, pool: WorkerTxPool, config: FaucetConfig) -> Self {
+        Self::spawn_with(reth_env, pool, config, TokioTaskExecutor::default())
     }
 
     /// Creates a new async LRU backed cache service task and spawns it to a new task via
-    /// [tokio::spawn].
+    /// [].
```
grantkee, line 155: delete "via []" and "See also [Self::spawn_with]"

crates/tn-reth/src/txn_pool.rs
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    pub async fn add_transaction(
+        &self,
+        recovered: EthPooledTransaction,
+    ) -> Result<TxHash, crate::PoolError> {
+        self.0.add_transaction(TransactionOrigin::Local, recovered).await
+    }
+
+    pub async fn add_transaction_and_subscribe(
```
grantkee, line 224: Only faucet calls this method right now, but please make a comment that explains the `TransactionOrigin` is set to local so we don't accidentally use this for forwarding txs between workers or at epoch boundaries, etc. Those tx origins should be external so they are scrutinized

crates/tn-reth/src/lib.rs
grantkee, line 1: Please add the lib data referenced in #220 for this new crate

```rust

#![doc(
    html_logo_url = "https://www.telco.in/logos/TEL.svg",
    html_favicon_url = "https://www.telco.in/logos/TEL.svg",
    issue_tracker_base_url = "https://github.com/telcoin-association/telcoin-network/issues/"
)]
#![warn(
    missing_debug_implementations,
    missing_docs,
    unreachable_pub,
    rustdoc::all,
    unused_crate_dependencies
)]
#![deny(unused_must_use, rust_2018_idioms)]
#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
```

grantkee (changes requested):
Good work refactoring reth into a single crate. This was a great idea! I have some comments and requested some changes. I'm happy to discuss further. Overall, great job

crates/tn-reth/src/lib.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// All dev related arguments with --dev prefix
+    #[clap(flatten)]
+    pub dev: DevArgs,
+
+    /// All pruning related arguments
+    #[clap(flatten)]
+    pub pruning: PruningArgs,
+    // All engine related arguments
+    //#[clap(flatten)]
+    //pub engine: EngineArgs,
```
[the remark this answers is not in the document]
grantkee: Okay, makes sense. You can leave them if you prefer

crates/execution/batch-builder/src/lib.rs
sstanfield: This is all re-worked (once branch is updated).

crates/tn-reth/src/txn_pool.rs
sstanfield: Added a comment and _local to the name to reinforce this.  Good catch.

crates/tn-reth/src/lib.rs
```diff
@@ -0,0 +1,983 @@
...
+        tokio::spawn(async move {
+            while let Some(latest) = stream.next().await {
+                let header = latest.tip().block.header.clone();
+                if let Err(_e) = tx.send(header).await {
+                    break;
+                }
+            }
+        });
+        rx
+    }
```
grantkee, line 439: This should send out the `SealedBlock`:

```rust
                let block = latest.tip().block.clone();
                if let Err(_e) = tx.send(block).await {
                    break;
                }
```

crates/execution/batch-builder/src/lib.rs
```diff
@@ -99,93 +81,38 @@ pub struct BatchBuilder<BT, Pool> {
...
-        // Canonical update
-        let update = CanonicalStateUpdate {
-            new_tip: &tip.block,          // finalized block
-            pending_block_base_fee,       // current base fee for worker (network-wide)
-            pending_block_blob_fee: None, // current blob fee for worker (network-wide)
-            changed_accounts,             // entire round of consensus
-            mined_transactions,           // entire round of consensus
-            update_kind: PoolUpdateKind::Commit,
+            state_changed,
+            last_canonical_update: None,
```
grantkee, line 111: Instead of an option, let's make this the `SealedBlock` that the worker builds the next batch off of. The `pending_base_fee` is set for the epoch, so no need to manage the extra data included with `LastCanonicalUpdate`.

Also, this cannot be an `Option`. This has to be a `SealedBlock` - even if it's Genesis when the node first starts

PR #259, opened 2025-04-18
Conclude Epoch System Call

grantkee:
- add closing epoch field to consensus output
- support for calling `concludeEpoch` on consensus registry contract
- simple fischer-yates shuffle algo based on leader cert's aggregate BLS signature and chacharand8
- update genesis tests to use main sol interface
- rename `block` -> `batch` leftovers

The `close_epoch` field is always set to `false` on consensus output so the closing epoch is never called. This will be addressed in a follow-up PR.

### DIFF
--- a/crates/tn-reth/Cargo.toml
+++ b/crates/tn-reth/Cargo.toml
@@ -62,4 +62,3 @@ alloy = { workspace = true, features = ["genesis"] }
 [dev-dependencies]
 tempfile = { workspace = true }
 serde_json = { workspace = true }
-tracing-subscriber = { workspace = true }
--- a/crates/tn-reth/Cargo.toml
+++ b/crates/tn-reth/Cargo.toml
@@ -20,7 +20,6 @@ enr = { workspace = true, default-features = false, features = [
     "rust-secp256k1",
 ] }
 
-
 reth = { workspace = true }
 reth-db = { workspace = true }
 reth-db-common = { workspace = true }
@@ -51,7 +50,7 @@ reth-discv4 = { workspace = true }
 reth-eth-wire = { workspace = true }
 reth-tracing = { workspace = true }
 
-
+rand_chacha = { workspace = true }
 futures = { workspace = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
@@ -61,3 +60,6 @@ eyre = { workspace = true }
 alloy = { workspace = true, features = ["genesis"] }
 
 [dev-dependencies]
+tempfile = { workspace = true }
+serde_json = { workspace = true }
+tracing-subscriber = { workspace = true }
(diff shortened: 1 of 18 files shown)


crates/tn-reth/src/lib.rs
```diff
@@ -115,6 +120,13 @@ pub mod worker;
 /// Rpc Server type, used for getting the node started.
 pub type RpcServer = TransportRpcModules<()>;
 
+/// The system address.
+pub const SYSTEM_ADDRESS: Address = address!("fffffffffffffffffffffffffffffffffffffffe");
+
+/// The address for consensus registry.
+pub const CONSENSUS_REGISTRY_ADDRESS: Address =
+    address!("7777777777777777777777777777777777777777");
```
grantkee, line 128: update this address

crates/tn-reth/src/traits.rs
```diff
@@ -205,6 +204,7 @@ impl TNPayload {
         let block_env = BlockEnv {
             // the block's number should come from the canonical tip, NOT the batch's number
             number: U256::from(self.attributes.parent_header.number + 1),
+            // special fee address
```
grantkee, line 207: update address with basefee contract

crates/tn-reth/src/lib.rs
robriks: it can be read from `tn-contracts/deployments/deployments.json` ConsensusRegistry key once I update the submodule's pinned commit. 

feel free to hard code as `0x07e17e17e17e17e17e17e17e17e17e17e17e17e1` for now and I will bring it in during genesis PR

robriks: also, `concludeEpoch(address[])` function selector is `0x7b55a300`

crates/tn-reth/Cargo.toml
```diff
@@ -61,3 +60,6 @@ eyre = { workspace = true }
 alloy = { workspace = true, features = ["genesis"] }
 
 [dev-dependencies]
+tempfile = { workspace = true }
+serde_json = { workspace = true }
+tracing-subscriber = { workspace = true }
```
grantkee, line 65: remove this

crates/tn-reth/src/lib.rs
```diff
@@ -734,6 +760,188 @@ impl RethEnv {
         Ok(sealed_block_with_senders)
     }
 
+    #[inline]
```
sstanfield, line 763: This is harmless so can ignore for the PR but why all these inlines?  These function seem large to be inlined and also seem to be doing a lot so probably would be lost in the noise anyway.  It is unlikely you will outguess the optimizer or that it will even respect these.  Just seems like clutter to me but I may be missing something.

crates/tn-reth/src/lib.rs
```diff
@@ -967,4 +1176,343 @@ impl RethEnv {
             self.evm_executor.executor(&mut db).execute(block)?;
         Ok((state, receipts))
     }
+
+    /// Create an EVM
```
sstanfield, line 1180: Is this comment right?  Or maybe incomplete?

sstanfield (approved):
This PR is awesome.  I need to spend more time with the system call interface for general knowledge but it looks great.

sstanfield (approved):

PR #261, opened 2025-04-22
Allow a node to not be validator (no authority).

sstanfield:
This is necessary for "observer"nodes that are not validators.

Make the authority optional.


crates/config/src/consensus.rs
```diff
@@ -201,10 +215,19 @@ where
     /// Retrieve the worker's network address by id.
     /// Note, will panic if id is not valid (not found in our worker cache).
```
grantkee, line 216: This comment can be updated to something like:
```rust
/// Note, will panic if id is not valid (not found in our worker cache) and the node is an authority. 
/// Otherwise, retrieve the next available udp port assigned by the system.
```

crates/consensus/worker/src/quorum_waiter.rs
```diff
@@ -69,7 +69,7 @@ pub struct QuorumWaiter {
 impl QuorumWaiter {
     /// Create a new QuorumWaiter.
     pub fn new(
-        authority: Authority,
+        authority: Option<Authority>,
```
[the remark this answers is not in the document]
grantkee: Maybe make the `QuorumWaiter` optional on the worker?

crates/consensus/worker/src/worker.rs
```diff
@@ -65,16 +65,23 @@ pub fn new_worker<DB: Database>(
...
+    if let Some(authority) = consensus_config.authority() {
+        // NOTE: This log entry is used to compute performance.
+        info!(target: "worker::worker",
+            "Worker {} successfully booted on {}",
+            id,
+            consensus_config
+                .worker_cache()
+                .worker(authority.protocol_key(), &id)
+                .expect("Our public key or worker id is not in the worker cache")
+                .transactions
```
grantkee, line 77: Is the transactions multiaddr actually connected to the RPC? I think `transactions` is busted and should at the very least log `worker_address` instead

grantkee (approved):
LGTM!

PR #262, opened 2025-04-23
Flag Epoch Boundary in Consensus

grantkee:
- logic to trigger epoch boundary in subscriber


crates/config/src/consensus.rs
```diff
@@ -24,6 +24,7 @@ struct ConsensusConfigInner<DB> {
     local_network: LocalNetwork,
     network_config: NetworkConfig,
     genesis: HashMap<CertificateDigest, Certificate>,
+    epoch_boundary: TimestampSec,
```
sstanfield, line 27: I assume this is a timestamp (hence the name) not a duration.  You should consider putting this in Committee, I added interior mutability to it recently in anticipation of epochs.  Otherwise when you want to update this you may be forced to put a lock in the Arc that wraps this inner.  Also, it is arguably part of the committee since it defines its expiration.

sstanfield (dismissed):
Looks good, with one comment.  You can decide on that but I think it would be a good idea and save you some extra refactors (and be faster) when it is time to update the epoch boundary.

crates/config/src/consensus.rs
grantkee: You bring up really good points. I refactored the epoch boundary to be with the committee instead

sstanfield (approved):

PR #264, opened 2025-04-25
Submodule update

grantkee:
- Update `tn-contracts` submodule
- set consensus registry in genesis on startup through CLI


bin/telcoin-network/tests/it/genesis_tests.rs
```diff
@@ -1,218 +1,241 @@
-#[cfg(test)]
-mod tests {
-    use crate::util::spawn_local_testnet;
-    use alloy::{
-        network::EthereumWallet, primitives::Uint, providers::ProviderBuilder, sol_types::SolCall,
+//! Tests with RPC calls for ConsensusRegistry in genesis (through CLI).
+//!
+//! NOTE: this test contains code for executing a proxy/impl pre-genesis
+//! however, the RPC calls don't work. The beginning of the test is left
+//! because the proxy version may be re-prioritized later.
```
robriks, line 5: is it possible we move to keeping the consensus registry as a proxy indefinitely? or maybe I am misunderestanding this comment

robriks (approved):
Awesome, this is perfect for getting ITS bridge contracts into genesis next. Thanks!!

crates/tn-reth/Cargo.toml
```diff
@@ -58,7 +58,8 @@ tokio = { workspace = true, features = ["sync", "rt", "macros", "signal"] }
 tracing = { workspace = true }
 eyre = { workspace = true }
 alloy = { workspace = true, features = ["genesis"] }
-
-[dev-dependencies]
 tempfile = { workspace = true }
```
sstanfield, line 61: Are we really using temple outside of tests?  Maybe should still be a dev-dependency.

sstanfield (approved):
Looks good.  At some point next week we should run a local testnet in a fresh directory, just to be sure we don't have any tn-contracts paths in use that we need to import the file into the binary.  We probably don't but that is a subtle thing we should check.

crates/tn-reth/Cargo.toml
robriks: Good point. The files that we are reading from tn-contracts are used for genesis configuration only- to fetch precompile addresses, bytecode and storage. I think those would not be included in the binary, is that right? So we should be good, but I agree we should do tests.

To answer your question, yes this PR now uses tempfile for genesis in `tn-reth/src/lib::RethEnv::create_consensus_registry_accounts_for_genesis`. 

It's because the registry's storage values for validator addresses & bls keys aren't known until genesis runtime. To fetch the correct storage configuration for the registry precompile we are spinning up a reth env to simulate the contract initialization. Maybe there is a way to do this without tempfile?

Before, my precompile config yaml generator utility in tn-contracts was handling this by using marker flags to mark the storage slots so they could later be populated by the protocol when the values are known but it was unnecessarily complex and I definitely like this better.

crates/tn-reth/src/lib.rs
```diff
@@ -1258,20 +1260,256 @@ impl RethEnv {
...
+        debug!(target: "engine", "reverts:\n{:#?}", reverts);
+        debug!(target: "engine", "state_size:{:#?}", state_size);
+        debug!(target: "engine", "reverts_size:{:#?}", reverts_size);
+
+        // place initialized bytecode in genesis
+        let storage = state.get(&CONSENSUS_REGISTRY_ADDRESS).map(|account| {
+            account.storage.iter().map(|(k, v)| ((*k).into(), v.present_value.into())).collect()
+        });
+
+        let genesis = genesis.extend_accounts([(
```
robriks, line 1343: we also need to set the `.with_balance(total_stake_balance)` in the real genesis.

grantkee: Good catch

crates/tn-reth/Cargo.toml
grantkee: We need a temporary execution environment for pre-genesis. Another solution could be to use `reth-db` with `test-utils` feature and import `create_test_rw_db` for the `RethEnv`. I'm not crazy about importing tempfile either. We could impl the necessary traits on the db in-memory but this was the fastest approach forward. I'm open to other suggestions

bin/telcoin-network/tests/it/genesis_tests.rs
grantkee: We want to keep consensus registry as a singleton. I think it would be nice to have some tests that demonstrate how to setup a proxy in rust (ie - faucet tests). I believe this code is close to demonstrating a proxy setup, and I didn't want to throw it away just yet. The comment referred to proxies in general

crates/tn-reth/src/system_calls.rs
```diff
@@ -9,51 +9,63 @@ use tn_types::Address;
...
 pub(super) const SYSTEM_ADDRESS: Address = address!("fffffffffffffffffffffffffffffffffffffffe");
 
-/// The address for consensus registry.
-pub(super) const CONSENSUS_REGISTRY_ADDRESS: Address =
+/// The address for consensus registry impl.
+pub const CONSENSUS_REGISTRY_ADDRESS: Address =
     address!("07E17e17E17e17E17e17E17E17E17e17e17E17e1");
 
+/// The address RWTEL.
+pub const RWTEL_ADDRESS: Address = address!("65a38cd7f35b57d7224709bdad5f6717cb5fe58d");
```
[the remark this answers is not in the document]
grantkee: I like this idea a lot. I'm not sure what the best approach is off the top of my head, but using `tn-contracts` as the source-of-truth is ideal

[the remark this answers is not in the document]
robriks: I can take it on in my follow up PR, will want to make sure you guys have a look at how I do it. as Steve pointed out we don't want to dirty up the rust binary with tn-contract files

bin/telcoin-network/tests/it/genesis_tests.rs
robriks: okay gotcha! sorry for the misunderstanding

crates/tn-reth/Cargo.toml
robriks: I have no qualms about tempfile, this looks good to me esp given timelines. As you've mentioned, it only runs once

sstanfield: Tempfile is fine.  Just not used outside tests much so wanted to make sure it was not an accident.

robriks (approved):
LGTM!

PR #265, opened 2025-04-25
Retry test fixes.

sstanfield:
This is primarily to improve the restart tests.  Includes two other "fixes" that seemed to improve the test reliability.
- When starting the node will progress once it can connect with one primary peer.  This is a lot looser then than the old code that expected a node to be in the committee and was also buggy which I believe occasionally derailed restart tests.
- The Reth DB (not used except to pass to Reth) is created early and not recreated on a node "restart" (currently used when a validator has caught up to rejoin consensus).  MDBX was occasionally freaking out when the DB was dropped and quickly recreated.  The consensus DB is also managed this way, always has been.  Forget specific reasons but probably similar.  There may be MDBX issues in general, have been seeing testnet issues now and then as well.  Nothing can be done about this at the moment and this seems to make it happier...

This won't fix all the issues, for instance I have recently seen port conflict issues and once had the nodes on restart advance one block and then stop.  Will get these in issues as data can be collected, it seems to be a lot more reliable now though.

### DIFF
--- a/bin/telcoin-network/tests/it/restarts.rs
+++ b/bin/telcoin-network/tests/it/restarts.rs
@@ -22,15 +22,16 @@ use tracing::{error, info};
 /// One unit of TEL (10^18) measured in wei.
 const WEI_PER_TEL: u128 = 1_000_000_000_000_000_000;
 
-/// Helper function to shutdown child processes and log errors.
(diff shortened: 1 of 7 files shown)


grantkee (approved):
LGTM!

PR #266, opened 2025-04-26
Include interchain token service as precompiles

robriks:
note: built off of the update-submodule branch

Adds ITS contracts as precompiles alongside consensus registry

This logic is not strictly related to creating a committee but it is currently under that abstraction since that's where the registry precompile is added. LMK if this should go somewhere else


bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -98,7 +103,7 @@ pub struct CreateCommitteeArgs {
     #[arg(
         long = "epoch-block-rewards",
         alias = "block_rewards_per_epoch",
-        help_heading = "The amount of TEL (incl 18 decimals) for the committee starting at genesis.",
+        help_heading = "The minimal amount a validator can withdraw. The default is 20mil TEL per 28 days.",
```
grantkee, line 106: This doc comment is for `min_withdrawal`. The previous version was correct

bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -55,9 +55,14 @@ pub struct CreateCommitteeArgs {
...
     pub chain: Arc<RethChainSpec>,
 
-    /// The path to the consensus registry storage yaml file.
-    #[arg(long, value_name = "CONSENSUS_REGISTRY_PATH", verbatim_doc_comment)]
-    pub consensus_registry: Option<PathBuf>,
+    /// The precompile config yaml dir.
+    #[arg(
+        long,
+        value_name = "PRECOMPILES_CONFIG_PATH", 
+        default_value = "../../tn-contracts/deployments/genesis", //todo will does manifest dir be avilable
```
grantkee, line 62: Can we leverage the `Datadir` for these contracts?

There's a leftover `//todo`, but we should address it in this PR. I'm not sure this is the correct approach. At the very least, I think the doc comment above `#[arg(...` should specify this is a relative path. @sstanfield what do you think we should do here?

grantkee: It might be best to remove this from the CLI and just error out if the contracts aren't there

grantkee (commented):
Great job getting the ITS contracts in genesis! I'd like Steve to review as well so we can all discuss the relative path for precompiles. Ultimately, this is genesis so I'm in favor of doing whatever is going to work since it's a one-time lift

bin/telcoin-network/src/genesis/create_committee.rs
robriks: I was unsure what approach we'd want to take here, erroring out if the contracts aren't found sounds good since we definitely need them in genesis.

will remove once we get Steve's opinion or discuss this in standup today

sstanfield: I think my preference would be to use the include macro stuff to put the files in the binary so there is no need to worry about paths at all.  Given this is only used once at genesis though that is probably not required (although this might be a more than one use thing).  However if we don't do that then we need to have some sort of "package" to send to deployers with these files included, I don't think we can expect that the MNOs will just checkout the repo with submodules (but maybe I am wrong there).  Also the only reason to not include them (in my mind) is if they are very large to the point of bloating the executable.  Also if these are part of genesis they may be needed for more than just initial validators, for instance if I start a new observer/validator it has to know the genesis for the chain it is joining and this is part of that right?

I think we used to include some of these with the macros but that was removed recently during all this work?  Ideally someone can just get a compiled binary and run it with a flag for say "main net" or "test net" etc and have it work?  Having a "package" is not a big deal but increases friction a bit (both for us and the deployer).

robriks: resolved by embedding relevant JSON and YAML files into binary as strings. This removes need for node operators to get involved with submodules and can be further abstracted with a YAML/JSON agnostic trait and update to `fetch_from_json_str`

grantkee (approved):
LGTM!

PR #267, opened 2025-04-27
Initial BLS keywrapping (encrypting).

sstanfield:
Opening for initial review.  This PR adds the ability to encrypt (wrap) the validator BLS keys.  It is very simple it can get a passphrase from:
- env via variable
- stdin
- Can ask if not in the background and on a TTY

It will hash the passphrase with Blake2 and use this as an AES key.  It uses the aes-gcm-siv algorithm with a fixed nonce.

It still supports the "old" un-wrapped (plain text) keys for testing but maybe should only wrap them (?).  It does not use any complicated configuration although we may add that later for more customization.  The ability to get a passphrase via env or stdin should allow easy integration with services that manage secrets.


crates/config/src/keys.rs
```diff
@@ -80,8 +108,22 @@ impl KeyConfig {
         let primary_network_keypair =
             Self::generate_network_keypair(&primary_keypair, primary_seed);
         let worker_network_keypair = Self::generate_network_keypair(&primary_keypair, worker_seed);
-        let contents = bs58::encode(primary_keypair.to_bytes()).into_string();
-        std::fs::write(tn_datadir.validator_keys_path().join(BLS_KEYFILE), contents)?;
+        if let Some(passphrase) = passphrase {
+            let mut hasher = DefaultHashFunction::new();
+            hasher.update(passphrase.into_bytes());
```
grantkee, line 113: Instead of blake2, we should use [pbkdf2](https://crates.io/crates/pbkdf2) to hash the passwords. It's much more difficult to crack than blake2. [EIP-2335](https://eips.ethereum.org/EIPS/eip-2335#decryption-key) also suggests this hashing algo. Here's a good explanation, but I've read this discussed several places:

> PBKDF2 is the go-to choice for password hashing, emphasizing security through iteration, while BLAKE2X excels in performance and flexibility for general-purpose hashing. The choice between the two ultimately depends on the specific requirements of the application, with PBKDF2 being preferable for scenarios where password security is paramount. [source](https://ssojet.com/compare-hashing-algorithms/pbkdf2-vs-blake2x/)

crates/config/src/keys.rs
```diff
@@ -80,8 +108,22 @@ impl KeyConfig {
...
+            let passphrase_bytes: [u8; 32] = hasher.finalize().into();
+            let key = Key::<Aes256GcmSiv>::from_slice(&passphrase_bytes);
+            let cipher = Aes256GcmSiv::new(key);
+            let nonce = Nonce::from_slice(b"telcoin net "); // 96-bits
```
grantkee, line 117: Using a static nonce undermines the security of the algo. Here's [a StackExchange discussion](https://crypto.stackexchange.com/questions/26790/how-bad-it-is-using-the-same-iv-twice-with-aes-gcm) I found insightful

> A single nonce reuse leaks the xor of plaintexts, so if one plaintext is known the adversary can completely decrypt the other.

Instead, the best practice is to generate the random 96-bit nonce and store it alongside the cipher text. [EIP-2335](https://eips.ethereum.org/EIPS/eip-2335#procedure) outlines a similar procedure.

This is another [random post](https://www.linkedin.com/pulse/understanding-nonces-use-aes-gcm-derek-hinch/), but I thought it was well-written and affirms a lot of the same information.

grantkee:
```rust
let mut nonce_bytes = [0u8; 12]; // 96-bits
rand::thread_rng().fill(&mut nonce_bytes);
let nonce = Nonce::from_slice(&nonce_bytes);
........
// Combine salt + nonce + ciphertext for storage
// (random salt for PBKDF2)
let encrypted_data = [&salt[..], &nonce_bytes[..], &ciphertext[..]].concat();
let contents = bs58::encode(encrypted_data).into_string();
```

grantkee (changes requested):
This is a great start! I looked into some of the cryptography decisions you've made and I have a few concerns. I've suggested some ways we can enhance the passwords. This is really close. Great work!

grantkee (approved):
User must explicitly declare no passphrase and the salt/nonce use separate rngs. LGTM!

PR #272, opened 2025-04-29
Add genesis CLI arg to optionally pass genesis config file

robriks:
Add genesis CLI arg to optionally pass genesis config file so that local testnet includes genesis precompiles


sstanfield (approved):
LGTM

PR #274, opened 2025-04-30
Fix a mistake that got checked into Makefile.

sstanfield:
Fix test-restarts, this undoes some a testing patch...


robriks (approved):
Good catch, this will re-include all the restart tests. LGTM!

PR #275, opened 2025-04-30
Retrieve TaskManager::submanagers by Name

grantkee:
- Store submanagers as a map to retrieve by name


sstanfield (approved):
I don't have an issue with this assuming you need this functionality.  Otherwise it seems pointless (?).

PR #276, opened 2025-04-30
Task Manager for Worker

grantkee:
- use const names for task managers
- worker task manager for batch builder


crates/consensus/worker/src/worker.rs
```diff
@@ -172,7 +178,7 @@ impl<DB: Database, QW: QuorumWaiterTrait> Worker<DB, QW> {
         let this_clone = this.clone();
         // Spawn a little task to accept batches from a channel and seal them that way.
         // Allows the engine to remain removed from the worker.
-        tokio::spawn(async move {
+        task_manager.spawn_task("batch-builder", async move {
```
sstanfield, line 181: Good catch putting this in the task manager.

sstanfield (approved):
Looks good.

PR #277, opened 2025-05-01
Separate Task Managers for Epoch/Node Related Tasks

grantkee:
- spawn tasks for epoch manager based on epoch responsibility
    - node tasks run for the lifetime of the node
    - epoch tasks run for the lifetime of the epoch

### DIFF
--- a/crates/node/src/manager.rs
+++ b/crates/node/src/manager.rs
@@ -150,7 +150,7 @@ where
 
         // start consensus metrics for the epoch
         if let Some(metrics_socket) = self.builder.consensus_metrics {
-            start_prometheus_server(metrics_socket);
+            start_prometheus_server(metrics_socket, &self.epoch_task_manager);
         }
 
         // create submanager for engine tasks
@@ -396,6 +396,7 @@ where
             consensus_config.clone(),
             consensus_bus.clone(),
             state_sync,
+            &self.epoch_task_manager, // tasks should abort with epoch
         )
         .spawn(&self.node_task_manager);
 
--- a/crates/node/src/manager.rs
+++ b/crates/node/src/manager.rs
@@ -482,9 +482,6 @@ where
         )
         .await?;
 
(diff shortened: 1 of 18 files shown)


sstanfield (commented):
Unless I am somehow missing something this has two task managers that exist outside on a Tokio runtime.  The runtime can cycle underneath them leaving tasks orphaned, I have no idea how the test_restarts_delayed test is able to handle this.  Looking again I guess that the contents of the task managers are being replaced wholesale with just the outer shell surviving.  I think this is a very dangerous path from a clarity perspective, the task managers should not exist outside their associated runtimes.

Maybe this is a stepping stone PR though since we talked about where this going offline?  If so I think I would like to see the final version more fleshed out, I think this iteration has some foot guns.

sstanfield (dismissed):
Good to go with it's stack.

crates/node/src/manager.rs
```diff
@@ -101,27 +113,157 @@ where
...
+        let mut node_task_manager = TaskManager::new(NODE_TASK_MANAGER);
+        let node_task_spawner = node_task_manager.get_spawner();
+        // create oneshot channel to await the first epoch to start
+        let (start, ready) = oneshot::channel();
+        node_task_spawner.spawn_task("epoch-started", ready);
+
+        // track the status for when the node is ready
+        self.node_ready = Some(start);
+        // ensure task is tracked
+        node_task_manager.update_tasks();
```
sstanfield, line 139: This should maybe go after spawn_node_networks() but since you join right after that anyway I don't think it matters and could maybe go away...  This mostly makes sure the proper tasks print when logging before the join, so not a functional issue and we are not logging before the join anyway.

sstanfield (approved):
Looks good.  One comment I missed earlier but it is not a blocker for getting the PR in and can be ignored for now (or maybe forever).

PR #282, opened 2025-05-02
Fix floating point precision issues, update tn-contracts commit

robriks:
replace use of rust scientific notation `e18` (which uses u64 floating point arithmetic prone to precision loss) with `alloy::primitives::utils::parse_ether`

update tn-contracts commit (that PR must be merged first) with improved deployments.json structure for better upstream rust logic


sstanfield (dismissed):
LGTM but I would give Grant a chance to review as well (I am not as up on the genesis contract stuff).

grantkee (dismissed):

sstanfield (approved):
LGTM

PR #286, opened 2025-05-03
Initial code to publish tnxs if not a CVV.

sstanfield:
Attempt to disperse transactions to the network if not a committee member.  A CVV should pick up the ten and include it in a batch.


grantkee (approved):
LGTM. Approving this PR because we discussed tests in a follow-up PR.

bin/telcoin-network/src/node.rs
robriks, line 1: Nice fix for the local testnet dev account funds

robriks (approved):
LGTM!

PR #288, opened 2025-05-04
Spawn Consensus Network for Epoch Manager

grantkee:
- refactor network and epoch manager so swarm survives but primary/worker networks reset

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -27,8 +27,8 @@ use std::{
     collections::{HashMap, HashSet, VecDeque},
     time::Duration,
 };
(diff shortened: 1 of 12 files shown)


crates/network-libp2p/src/consensus.rs
```diff
@@ -431,6 +399,9 @@ where
                 // for now, this only supports the current committee for the epoch
 
                 self.swarm.behaviour_mut().peer_manager.new_epoch(committee);
+
+                // update the stream to forward events
+                self.event_stream = new_event_stream;
```
sstanfield, line 404: Nice, this is the way!

sstanfield (approved):
This looks awesome, good solution to decouple this stuff.

PR #289, opened 2025-05-05
Observer tests and config.

sstanfield:
Makes configuring an observer more straightforward.  Adds a restart test to make sure an observer is participating in a network and adds an observer to the local testate script.

Also updated tn-contracts, I needed to pull master for something and this happened...

### DIFF
--- a/bin/telcoin-network/src/keytool/generate.rs
+++ b/bin/telcoin-network/src/keytool/generate.rs
@@ -1,11 +1,14 @@
 //! Generate subcommand
 
-use crate::args::{clap_address_parser, clap_genesis_parser};
+use crate::{
+    args::{clap_address_parser, clap_genesis_parser},
+    genesis::account_from_word,
+};
 use clap::{value_parser, Args, Subcommand};
 use std::sync::Arc;
 use tn_config::{Config, KeyConfig, TelcoinDirs};
 use tn_reth::{dirs::DataDirPath, MaybePlatformPath, RethChainSpec};
-use tn_types::Address;
+use tn_types::{Address, GenesisAccount, U256};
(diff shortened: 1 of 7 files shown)


bin/telcoin-network/src/keytool/mod.rs
grantkee, line 1: Do you anticipate this code changing? 

I think it would be best to create another method (maybe `Self::generate_keys`?) and consolidate the duplicate code into that method. You can also remove the todos in `init_path`. The `else if` should probably return an error if `!force` to prevent accidentally overwriting validator keys. I dropped that ball

bin/telcoin-network/src/keytool/generate.rs
```diff
@@ -113,3 +119,73 @@ impl ValidatorArgs {
...
+    /// Used to add a funded account (by simple text string).  Use this on a dev cluster
+    /// (must provide on all validator genesis inits) to have an account with a deterministically
+    /// derived key. This is ONLY for dev testing, never use this for other chains.
+    #[arg(long)]
+    pub dev_funded_account: Option<String>,
+}
+
+impl ObserverArgs {
+    /// Create all necessary information needed for validator and save to file.
+    pub fn execute<TND: TelcoinDirs>(
```
grantkee, line 160: Most of this code is identical to `ValidatorArgs::execute`. Is this code anticipated to change? I think a helper function would be nice but not a huge deal

grantkee (approved):
LGTM. I think there are some small improvements to consider, but not a deal breaker for this PR

grantkee (dismissed):

grantkee (approved):

PR #298, opened 2025-05-08
Feat/genesis updates

robriks:
Update to tn-contracts v1.0

### DIFF
--- a/bin/telcoin-network/src/args/utils.rs
+++ b/bin/telcoin-network/src/args/utils.rs
@@ -40,12 +40,9 @@ pub fn clap_address_parser(value: &str) -> eyre::Result<Address> {
 ///
 /// Pass "0" to return zero as u232
 pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
-    let parsed_val = match value {
-        "0" => U232::ZERO,
-        _ => U232::from_str_radix(value, 10)?
-            .checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation"))
-            .expect("U232 parsing"),
-    };
+    let parsed_val = U232::from_str_radix(value, 10)?
+        .checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation"))
+        .expect("U232 parsing");
 
     Ok(parsed_val)
 }
--- a/bin/telcoin-network/src/args/utils.rs
+++ b/bin/telcoin-network/src/args/utils.rs
@@ -1,6 +1,7 @@
 //! Utilities for parsing args
 
 use std::{str::FromStr, sync::Arc};
+use alloy::primitives::aliases::U232;
 use tn_reth::{chain_value_parser, dirs::DataDirPath, MaybePlatformPath, RethChainSpec};
 use tn_types::{adiri_chain_spec_arc, Address};
 
@@ -33,4 +34,16 @@ pub fn clap_address_parser(value: &str) -> eyre::Result<Address> {
     };
 
     Ok(address)
+}
+
+/// Parse 18 decimal U232 from string for ConsensusRegistry.
+///
+/// Pass "0" to return zero as u232
+pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
+    let parsed_val = match value {
+        "0" => U232::ZERO,
+        _ => U232::from_str(value)?.checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation")).expect("U232 parsing"),
+    };
+
+    Ok(parsed_val)
 }
--- a/bin/telcoin-network/src/args/utils.rs
+++ b/bin/telcoin-network/src/args/utils.rs
@@ -1,6 +1,7 @@
 //! Utilities for parsing args
 
 use alloy::primitives::aliases::U232;
+use eyre::OptionExt;
 use std::{str::FromStr, sync::Arc};
 use tn_reth::{chain_value_parser, dirs::DataDirPath, MaybePlatformPath, RethChainSpec};
 use tn_types::{adiri_chain_spec_arc, Address};
@@ -37,12 +38,19 @@ pub fn clap_address_parser(value: &str) -> eyre::Result<Address> {
 }
 
 /// Parse 18 decimal U232 from string for ConsensusRegistry.
-///
-/// Pass "0" to return zero as u232
 pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
     let parsed_val = U232::from_str_radix(value, 10)?
         .checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation"))
-        .expect("U232 parsing");
+        .ok_or_eyre("U232 parsing")?;
 
     Ok(parsed_val)
+}
+
+/// Parse U232 from string for ConsensusRegistry.
+pub fn clap_u232_parser_with_divisor(value: &str, divisor: &str) -> eyre::Result<U232> {
+    let parsed_val = clap_u232_parser(value)?;
+    let divisor_value = U232::from_str_radix(divisor, 10)?;
+    let divided_val = parsed_val.checked_div(divisor_value).ok_or_eyre("U232 divisor")?;
+
+    Ok(divided_val)
 }
--- a/bin/telcoin-network/src/args/utils.rs
+++ b/bin/telcoin-network/src/args/utils.rs
@@ -1,7 +1,7 @@
 //! Utilities for parsing args
 
-use std::{str::FromStr, sync::Arc};
 use alloy::primitives::aliases::U232;
+use std::{str::FromStr, sync::Arc};
 use tn_reth::{chain_value_parser, dirs::DataDirPath, MaybePlatformPath, RethChainSpec};
 use tn_types::{adiri_chain_spec_arc, Address};
 
@@ -42,7 +42,9 @@ pub fn clap_address_parser(value: &str) -> eyre::Result<Address> {
 pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
     let parsed_val = match value {
         "0" => U232::ZERO,
-        _ => U232::from_str_radix(value, 10)?.checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation")).expect("U232 parsing"),
+        _ => U232::from_str_radix(value, 10)?
+            .checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation"))
+            .expect("U232 parsing"),
     };
 
     Ok(parsed_val)
--- a/bin/telcoin-network/src/args/utils.rs
+++ b/bin/telcoin-network/src/args/utils.rs
@@ -45,12 +45,3 @@ pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
 
     Ok(parsed_val)
 }
-
-/// Parse U232 from string for ConsensusRegistry.
-pub fn clap_u232_parser_with_divisor(value: &str, divisor: &str) -> eyre::Result<U232> {
-    let parsed_val = clap_u232_parser(value)?;
-    let divisor_value = U232::from_str_radix(divisor, 10)?;
-    let divided_val = parsed_val.checked_div(divisor_value).ok_or_eyre("U232 divisor")?;
-
-    Ok(divided_val)
-}
--- a/bin/telcoin-network/src/args/utils.rs
+++ b/bin/telcoin-network/src/args/utils.rs
@@ -42,7 +42,7 @@ pub fn clap_address_parser(value: &str) -> eyre::Result<Address> {
 pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
     let parsed_val = match value {
         "0" => U232::ZERO,
-        _ => U232::from_str(value)?.checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation")).expect("U232 parsing"),
+        _ => U232::from_str_radix(value, 10)?.checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation")).expect("U232 parsing"),
     };
 
     Ok(parsed_val)
--- a/bin/telcoin-network/src/genesis/create_committee.rs
+++ b/bin/telcoin-network/src/genesis/create_committee.rs
@@ -1,9 +1,8 @@
 //! Create a committee from the validators in genesis.
 
 use crate::args::{clap_address_parser, clap_genesis_parser, clap_u232_parser};
-use alloy::primitives::{aliases::U232, ruint::aliases::U256, utils::parse_ether};
+use alloy::primitives::{aliases::U232, ruint::aliases::U256};
 use clap::Args;
-use core::panic;
 use std::{path::PathBuf, str::FromStr, sync::Arc};
 use tn_config::{
     Config, ConfigFmt, ConfigTrait, NetworkGenesis, TelcoinDirs as _, DEPLOYMENTS_JSON,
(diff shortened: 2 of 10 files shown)


bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -79,30 +79,30 @@ pub struct CreateCommitteeArgs {
         long = "initial-stake-per-validator",
         alias = "stake",
         help_heading = "The initial stake credited to each validator in genesis. The default is 1mil TEL.",
-        default_value_t = U256::try_from(parse_ether("1_000_000").expect("parse_ether")).expect("initial stake"),
+        default_value_t = U232::from(U256::try_from(parse_ether("1_000_000").expect("parse_ether")).expect("initial stake")),
```
grantkee, line 82: For these conversions, I'd prefer if we refactor this to just use the `U232` api:

```rust
U232::from(U232::from(1_000_000).checked_mul(U232::from(10).checked_pow(18).expect("U232 doesn't overflow 10^18")))
```

The cleanest solution is to add this as a `fn clap_u232_parser` in `args/utils` like [this](https://github.com/Telcoin-Association/telcoin-network/blob/958c9518f102499de269cfe8d4a13976ce629656/bin/telcoin-network/src/args/utils.rs#L29) and remove the `expect` to use `?` with `eyre::Result<U232>`

bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -181,19 +170,31 @@ impl CreateCommitteeArgs {
...
+        let itel_address =
+            match RethEnv::fetch_from_json_str(DEPLOYMENTS_JSON, Some("its.InterchainTEL")) {
+                Ok(res) => match res {
+                    serde_json::Value::String(s) => {
+                        Address::from_str(&s).expect("ITEL addr incorrect")
+                    }
+                    _ => panic!("ITEL address not a string"),
+                },
+                _ => panic!("ITEL address not found"),
+            };
```
grantkee, line 197: nit: instead of panics, use the `?` operator since this method returns a result

```rust
        let itel_address_str: String =
            RethEnv::fetch_from_json_str(DEPLOYMENTS_JSON, Some("its.InterchainTEL"))?;
        let itel_address = Address::from_str(&itel_address_str)?;
```

crates/tn-reth/src/lib.rs
```diff
@@ -1735,11 +1725,10 @@ mod tests {
...
         let initial_stake_config = ConsensusRegistry::StakeConfig {
-            stakeAmount: U256::try_from(parse_ether("1_000_000").unwrap()).unwrap(),
-            minWithdrawAmount: U256::try_from(parse_ether("1_000").unwrap()).unwrap(),
-            epochIssuance: U256::try_from(parse_ether("20_000_000").unwrap())
-                .unwrap()
-                .checked_div(U256::from(28))
+            stakeAmount: U232::from(U256::try_from(parse_ether("1_000_000").unwrap()).unwrap()),
+            minWithdrawAmount: U232::from(U256::try_from(parse_ether("1_000").unwrap()).unwrap()),
+            epochIssuance: U232::from(U256::try_from(parse_ether("20_000_000").unwrap()).unwrap())
+                .checked_div(U232::from(28))
```
grantkee, line 1731: flagging this for the `U232` approach as well

grantkee (changes requested):
Great job with these updates! I requested some technical "rusty" syntax changes to keep the codebase consistent, but the logic is sound

crates/tn-reth/src/lib.rs
```diff
@@ -1344,50 +1336,48 @@ impl RethEnv {
...
-        // extract the specific field we want
-        let abi = json["deployedBytecode"]["object"]
-            .as_str()
-            .ok_or_eyre("Invalid json abi format for bytecode")?;
+    /// Fetches json info from the given string
+    ///
+    /// If a key is specified, return the corresponding nested object.
+    /// Otherwise return the entire JSON
+    /// With a generic this could be adjused to handle YAML also
+    pub fn fetch_from_json_str(json_content: &str, key: Option<&str>) -> eyre::Result<Value> {
```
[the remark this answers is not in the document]
robriks: Good call, this specifies the return type! resolved in 9fa9f3b7

bin/telcoin-network/src/genesis/create_committee.rs
robriks: Definitely, thanks for the guidance! Resolved in f5951dff

bin/telcoin-network/src/genesis/create_committee.rs
robriks: Oh that's so much prettier! Done in 8a1a20c4

bin/telcoin-network/src/args/utils.rs
```diff
@@ -34,3 +35,17 @@ pub fn clap_address_parser(value: &str) -> eyre::Result<Address> {
...
+/// Pass "0" to return zero as u232
+pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
+    let parsed_val = match value {
+        "0" => U232::ZERO,
+        _ => U232::from_str_radix(value, 10)?
+            .checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation"))
+            .expect("U232 parsing"),
+    };
+
+    Ok(parsed_val)
```
grantkee, line 50: This is the right approach, but please remove the `match` statement so it's always `U32::from_str_radix(.....`. The zero should parse correctly

bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -79,30 +78,33 @@ pub struct CreateCommitteeArgs {
         long = "initial-stake-per-validator",
         alias = "stake",
         help_heading = "The initial stake credited to each validator in genesis. The default is 1mil TEL.",
-        default_value_t = U256::try_from(parse_ether("1_000_000").expect("parse_ether")).expect("initial stake"),
+        value_parser = clap_u232_parser,
+        default_value_t = clap_u232_parser("1_000_000").expect("U232 parsing"),
```
grantkee, line 82: these can just be `default_value = "1_000_000"`

The `default_value_t` expects a type, but since you have the value parser, it's okay to pass a string with `default_type` which will then use the `value_parser`

bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -79,30 +78,33 @@ pub struct CreateCommitteeArgs {
...
+    pub initial_stake: U232,
 
     /// The minimum amount a validator can withdraw.
     #[arg(
         long = "min-withdraw-amount",
         alias = "min_withdraw",
         help_heading = "The minimal amount a validator can withdraw. The default is 1_000 TEL.",
-        default_value_t = U256::try_from(parse_ether("1_000").expect("parse_ether")).expect("min withdraw"),
+        value_parser = clap_u232_parser,
+        default_value_t = clap_u232_parser("1_000").expect("U232 parsing"),
```
grantkee, line 93: `default_value` here as well

bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -79,30 +78,33 @@ pub struct CreateCommitteeArgs {
...
+    pub min_withdrawal: U232,
 
     /// The amount of block rewards per epoch starting in genesis.
     #[arg(
         long = "epoch-block-rewards",
         alias = "block_rewards_per_epoch",
         help_heading = "The amount of TEL (incl 18 decimals) for the committee starting at genesis.",
-        default_value_t = U256::try_from(parse_ether("20_000_000").expect("parse_ether")).expect("block rewards").checked_div(U256::from(28)).expect("U256 div works"),
+        value_parser = clap_u232_parser,
+        default_value_t = clap_u232_parser("20_000_000").expect("U232 parsing").checked_div(U232::from(28)).expect("U256 div works"),
```
grantkee, line 104: `default_value`

bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -181,19 +172,20 @@ impl CreateCommitteeArgs {
...
                     self.consensus_registry_owner,
-                    itel_address,
                 )
             })?
         };
         // use embedded ITS config from submodule, passing in decremented ITEL balance
         let genesis_stake = self
             .initial_stake
-            .checked_mul(U256::from(validators.clone().len()))
+            .checked_mul(U232::from(validators.clone().len()))
```
grantkee, line 181: no need for `clone` here

crates/tn-reth/src/lib.rs
```diff
@@ -1735,11 +1743,11 @@ mod tests {
 
         let epoch_duration = 60 * 60 * 24; // 24hrs
         let initial_stake_config = ConsensusRegistry::StakeConfig {
-            stakeAmount: U256::try_from(parse_ether("1_000_000").unwrap()).unwrap(),
-            minWithdrawAmount: U256::try_from(parse_ether("1_000").unwrap()).unwrap(),
-            epochIssuance: U256::try_from(parse_ether("20_000_000").unwrap())
+            stakeAmount: clap_u232_parser("1_000_000").unwrap(),
+            minWithdrawAmount: clap_u232_parser("1_000").unwrap(),
+            epochIssuance: clap_u232_parser("20_000_000")
```
grantkee, line 1748: `U232::from(parse_ether(...))` instead of of the clap parser

grantkee (changes requested):

bin/telcoin-network/src/genesis/create_committee.rs
robriks: oh sweet I was wondering why the tests were double parsing these values. resolved in 839c5961

bin/telcoin-network/src/genesis/create_committee.rs
robriks: resolved in 839c5961

bin/telcoin-network/src/genesis/create_committee.rs
robriks: Gotcha, removed in 839c5961

bin/telcoin-network/src/genesis/create_committee.rs
robriks: This one results in a precision loss as it stands. It's not a critical calculation but it would be nice for it to be correct. 

I can't figure out how to get rust to parse a U232 string with a decimal in it, if there's a way to do that it would resolve the precision loss.

Another option is to do the division in place after scaling so the precision is kept.

crates/tn-reth/src/lib.rs
robriks: resolved in a9c85100

bin/telcoin-network/src/genesis/create_committee.rs
sstanfield: U232 means unsigned int 232 bits long.  It's an integer so no decimal.  You would need a float (non-starter here I assume) or track the remainder in another int if you need it.  Even if it parsed a decimal it would be lost once a U232 so it is being more precise by refusing to parse it.

bin/telcoin-network/src/args/utils.rs
```diff
@@ -34,3 +35,14 @@ pub fn clap_address_parser(value: &str) -> eyre::Result<Address> {
 
     Ok(address)
 }
+
+/// Parse 18 decimal U232 from string for ConsensusRegistry.
+///
+/// Pass "0" to return zero as u232
+pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
+    let parsed_val = U232::from_str_radix(value, 10)?
+        .checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation"))
```
sstanfield, line 44: Since this returns an error and takes user input for value you might want to consider returning errors vs the expects unless you are really sure that those conditions will never happen.

sstanfield (approved):
I think this is good.  One comment you can decide if needs addressing, I think it does since a clap parser will work on user input.  Then again it is from the command line so maybe crashing early is OK.,

bin/telcoin-network/src/args/utils.rs
robriks: I must be missing something because the compiler won't let me use ?, ie:

```
the `?` operator can only be used on `Result`s, not `Option`s, in a function that returns `Result`
the trait `FromResidual<std::option::Option<Infallible>>` is not implemented for `Result<alloy::alloy_primitives::Uint<232, 4>, ErrReport>`
the trait `FromResidual<Result<Infallible, E>>` is implemented for `Result<T, F>`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0
utils.rs(49, 1): this function returns a `Result`
```

so I just replaced them with expects and then the compiler is happy. maybe there is something I need to import or implement to access the ? operator?

robriks: ah the `ok_or_eyre()` thing seems to work in place of the expects, that seems like it's probably what I should be using to propagate the errors correctly, yes?

bin/telcoin-network/src/genesis/create_committee.rs
robriks: to preserve the precision of 20_000_000 / 28 I added a second clap parser fn which accepts a divisor in commit [fb83e91](https://github.com/Telcoin-Association/telcoin-network/pull/298/commits/fb83e915e34c1f65a7fc1175b1e4d19888cdebfb)

it first performs the 18 decimal scaling using `clap_u232_parser` and then applies the divisor so there is no precision loss

bin/telcoin-network/src/args/utils.rs
sstanfield: Seems like you resolved this but the root cause is ? has too match the return type.  You can't ? an Option unless the function returns an Option for instance.  The ok_or_eyre() should be fine given this is an eyre error.  There will usually be some sort of mapping function for this (like the ok_or_eyre() you found of map_err() if the error you have is not the same as the return type).  Think of ? as just sugar around "match X { Err(e) => return Err(e),..." so your types still have to work (it can use From/Into for Errors but that won't help for going from Option to Result).

robriks: that makes sense! could not for the life of me understand why it sometimes worked and sometimes not, but makes sense that would be up to type matching

bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -79,30 +80,33 @@ pub struct CreateCommitteeArgs {
...
+    pub min_withdrawal: U232,
 
     /// The amount of block rewards per epoch starting in genesis.
     #[arg(
         long = "epoch-block-rewards",
         alias = "block_rewards_per_epoch",
         help_heading = "The amount of TEL (incl 18 decimals) for the committee starting at genesis.",
-        default_value_t = U256::try_from(parse_ether("20_000_000").expect("parse_ether")).expect("block rewards").checked_div(U256::from(28)).expect("U256 div works"),
+        value_parser = |s: &str| clap_u232_parser_with_divisor(s, "28"),
+        default_value = "20_000_000",
```
grantkee, line 106: This works, but I think it's confusing. I'm in favor of using the `clap_u232_parser`, default value, and updating the help heading to explain that this should be:
```rust 
help_heading = "The per block reward (int) for each epoch. Ex) 20mil rewards per month / 31 days / 25 hour epoch interval. It's best to use conservative values."
```

Then set the default to "25806". This is only used during genesis so I think it's okay to be a little rough on the edges and ask devs to do their own math.

grantkee (changes requested):

bin/telcoin-network/src/genesis/create_committee.rs
robriks: Ok sweet, so block issuance will be 19999650 TEL per 31 days for 25 hour epochs. Resolved in commit [ab3e9fc](https://github.com/Telcoin-Association/telcoin-network/pull/298/commits/ab3e9fc7bdc0f479a4a92da6549ed9a53fb2f4df)

grantkee (approved):

PR #300, opened 2025-05-08
Remove reth args that are confusing or broken.

sstanfield:
Some of these args we inherited from Reth will at best be confusing and at worst will lead to a broken node.  Remove them and use the defaults we want.


crates/tn-reth/src/lib.rs
```diff
@@ -237,11 +224,11 @@ impl RethConfig {
...
             txpool,
-            builder,
-            debug,
+            builder: PayloadBuilderArgs::default(),
+            debug: DebugArgs::default(),
             db,
-            dev,
-            pruning,
+            dev: DevArgs::default(),
+            pruning: PruningArgs::default(),
```
grantkee, line 231: explicitly specify "default" values to force attention on future reth upgrades

grantkee (changes requested):

grantkee (approved):

PR #308, opened 2025-05-16
Disable Reth metrics, just our own metrics now.

sstanfield:
We aren't using them, there may be a Reth DB issue related to them on testate, we need to review metrics in general and we should be to provide the metrics we care about ourselves after review.


grantkee (dismissed):

grantkee (approved):

grantkee (approved):
LGTM! Nice work

PR #310, opened 2025-05-19
Keep engine alive across epochs

grantkee:
- spawn engine tasks on node startup
- only restart the engine's `BatchBuilder` per epoch
- add tx pool to `BatchValidator` to submit gossip transactions to worker's tx pool
- add task spawner to `RethEnv` for spawning other engine tasks


crates/node/src/manager.rs
```diff
@@ -83,6 +84,8 @@ pub struct EpochManager<P> {
     /// and returns the ack that the first epoch has started. At this point,
     /// the node manager has other tasks to await and will run normally.
     node_ready: Option<oneshot::Sender<()>>,
+    /// Output channel for consensus blocks to be executed.
+    consensus_output: broadcast::Sender<ConsensusOutput>,
```
sstanfield, line 88: I see why you did this but don't love channels living outside the consensus bus.  One of its uses is to make it simple to find code that send and receives a channel and this violates that.  Probably no way around it, but might be worth noting it should only be accesses via the consensus bus.

crates/execution/batch-builder/src/lib.rs
```diff
@@ -377,7 +369,7 @@ mod tests {
         let tmp_dir = TempDir::new().unwrap();
         let reth_env =
             RethEnv::new_for_temp_chain(chain.clone(), tmp_dir.path(), &task_manager).unwrap();
-        let txpool = reth_env.worker_txn_pool().clone();
+        let txpool = reth_env.init_txn_pool().unwrap().clone();
```
sstanfield, line 372: Doesn't this return a new instance?  Why the clone?

crates/execution/batch-builder/src/lib.rs
```diff
@@ -529,7 +521,7 @@ mod tests {
         // task manger
         let task_manager = TaskManager::new("Test Task Manager");
         let reth_env = RethEnv::new_for_temp_chain(chain.clone(), path, &task_manager).unwrap();
-        let txpool = reth_env.worker_txn_pool().clone();
+        let txpool = reth_env.init_txn_pool().unwrap().clone();
```
sstanfield, line 524: Clone?

sstanfield (approved):
Looks good, couple minor comments.  This will be great to have working.

sstanfield (approved):
Looks good.

Issue #189, opened 2025-01-27
Worker RPC


grantkee:
`tn_node::engine::worker` implements eth RPC endpoints that should work, but are hardcoded and need attention

For example, use the Cargo.toml version as client version

PR #314, opened 2025-05-22
Fill out some of the info for the RPCs.

sstanfield:
https://github.com/Telcoin-Association/telcoin-network/issues/189

This should cover or bases.  A lot of this stuff will not be used since we are not supporting the Reth admin and txpool namespaces.

### DIFF
--- a/crates/network-libp2p/src/types.rs
+++ b/crates/network-libp2p/src/types.rs
@@ -14,10 +14,7 @@ use libp2p::{
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use tn_types::{encode, BlsPublicKey, BlsSignature, NetworkPublicKey};
-use tokio::sync::{
-    mpsc,
-    oneshot::{self},
-};
+use tokio::sync::{mpsc, oneshot};
 
 #[cfg(test)]
 #[path = "tests/types.rs"]
--- a/crates/network-libp2p/src/types.rs
+++ b/crates/network-libp2p/src/types.rs
@@ -14,7 +14,10 @@ use libp2p::{
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use tn_types::{encode, BlsPublicKey, BlsSignature, NetworkPublicKey};
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::{
+    mpsc,
+    oneshot::{self},
+};
 
 #[cfg(test)]
 #[path = "tests/types.rs"]
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -21,6 +21,7 @@ use tn_types::{
     Address, BatchSender, BatchValidation, ConsensusOutput, ExecHeader, Noticer, SealedHeader,
     TaskSpawner, WorkerId, B256, MIN_PROTOCOL_BASE_FEE,
 };
+use tn_worker::WorkerNetworkHandle;
 use tokio::sync::broadcast;
 use tokio_stream::wrappers::BroadcastStream;
 use tracing::{error, info};
@@ -113,11 +114,12 @@ impl ExecutionNodeInner {
     pub(super) async fn initialize_worker_components(
         &mut self,
         worker_id: WorkerId,
+        network_handle: WorkerNetworkHandle,
     ) -> eyre::Result<()> {
         let transaction_pool = self.reth_env.init_txn_pool()?;
 
-        // TODO: update WorkerNetwork - issue #189
-        let network = WorkerNetwork::new(self.reth_env.chainspec());
+        let network =
+            WorkerNetwork::new(self.reth_env.chainspec(), network_handle, self.tn_config.version);
         let mut tx_pool_latest = transaction_pool.block_info();
         tx_pool_latest.pending_basefee = MIN_PROTOCOL_BASE_FEE;
         let last_seen = self.reth_env.finalized_block_hash_number()?;
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -42,8 +42,6 @@ pub(super) struct ExecutionNodeInner {
     pub(super) opt_faucet_args: Option<FaucetArgs>,
     /// Collection of execution components by worker.
     pub(super) workers: HashMap<WorkerId, WorkerComponents>,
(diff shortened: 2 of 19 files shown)


crates/network-libp2p/src/types.rs
```diff
@@ -14,7 +14,10 @@ use libp2p::{
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use tn_types::{encode, BlsPublicKey, BlsSignature, NetworkPublicKey};
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::{
+    mpsc,
+    oneshot::{self},
```
grantkee, line 19: is this a clippy delta?

crates/tn-reth/src/lib.rs
```diff
@@ -1871,4 +1915,35 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_rpc_validator() {
```
grantkee, line 1920: niiice

crates/tn-reth/src/worker.rs
```diff
@@ -46,43 +52,82 @@ impl WorkerComponents {
...
     async fn network_status(&self) -> Result<NetworkStatus, NetworkError> {
         Ok(NetworkStatus {
-            client_version: "v0.0.1".to_string(),
-            protocol_version: 1,
+            client_version: self.version.to_string(), // web3_clientVersion
+            protocol_version: 1,                      // eth_protocolVersion
             eth_protocol_info: EthProtocolInfo {
                 difficulty: None,
-                network: self.chain_id(),
+                network: 1, // ??
```
grantkee, line 122: This should be the network id, which is also the chain id for TN

crates/tn-reth/src/worker.rs
```diff
@@ -97,21 +142,24 @@ impl NetworkInfo for WorkerNetwork {
 }
 
 impl PeersInfo for WorkerNetwork {
+    // net_peerCount
     fn num_connected_peers(&self) -> usize {
-        0
+        *self.peer_count.read()
```
grantkee, line 147: Good workaround for this sync trait

crates/tn-reth/src/worker.rs
```diff
@@ -97,21 +142,24 @@ impl NetworkInfo for WorkerNetwork {
...
     }
 
+    // TN Unused
     fn local_node_record(&self) -> NodeRecord {
         NodeRecord::new(self.local_addr(), PeerId::random())
     }
 
-    // TODO: this is not supported
+    // TN Unused
     fn local_enr(&self) -> Enr<SecretKey> {
```
grantkee, line 156: This could be the kad record, but I think the ENR only supports ECDSA signatures so probably not worth it rn

crates/tn-reth/src/worker.rs
```diff
@@ -46,43 +52,82 @@ impl WorkerComponents {
...
-        Self { chain_spec: chain_spec.reth_chain_spec() }
+    pub fn new(
+        chain_spec: ChainSpec,
+        worker_network: WorkerNetworkHandle,
+        version: &'static str,
+    ) -> Self {
+        let peer_count = Arc::new(RwLock::new(0));
+        let peer_count_clone = peer_count.clone();
+        let spawner = worker_network.get_task_spawner().clone();
+        spawner.spawn_task("Worker Network Peers", async move {
```
grantkee, line 80: good solution

crates/node/src/manager.rs
```diff
@@ -662,18 +662,23 @@ where
...
                 .as_mut()
                 .ok_or_eyre("worker network handle missing from epoch manager")?;
+
             network_handle.update_task_spawner(epoch_task_spawner.clone());
+            // initialize worker components on startup
+            // This will use the new epoch_task_spawner on network_handle.
+            if *initial_epoch {
+                engine.initialize_worker_components(*worker_id, network_handle.clone()).await?;
+            } else {
+                // We updated our epock task spawner so make sure worker network tasks are
```
grantkee, line 678: nit: epock -> epoch

crates/node/src/engine/mod.rs
```diff
@@ -71,9 +72,21 @@ impl ExecutionNode {
...
+        &self,
+        worker_id: WorkerId,
+        network_handle: WorkerNetworkHandle,
+    ) -> eyre::Result<()> {
         let mut guard = self.internal.write().await;
-        guard.initialize_worker_components(worker_id).await
+        guard.initialize_worker_components(worker_id, network_handle).await
+    }
+
+    /// Respqn any tasks on the worker network when we get a new epoch task manager.
```
grantkee, line 84: nit: respqn -> respawn

crates/node/src/engine/inner.rs
```diff
@@ -41,6 +42,8 @@ pub(super) struct ExecutionNodeInner {
     pub(super) opt_faucet_args: Option<FaucetArgs>,
     /// Collection of execution components by worker.
     pub(super) workers: HashMap<WorkerId, WorkerComponents>,
+    /// Keep the WorkerNetwork around so we can update it's task(s).
+    pub(super) network: Option<WorkerNetwork>,
```
grantkee, line 46: Instead of making this a separate field, please add to `WorkerComponents`

grantkee (changes requested):
I caught a few typos. The biggest thing is adding the engine's `WorkerNetwork` to the `WorkerComponents`. The logic LGTM

crates/network-libp2p/src/types.rs
sstanfield: Don't know, I suspect this was a rust analyzer thing.  I did add it manually and not sure what it did not trip a warning somewhere.

crates/tn-reth/src/worker.rs
sstanfield: Oops, that was from trying o find out where it was exposed.  I don't think I found it (hence the ???) on the RPC so I don't think it actually matters but agree it should be the chain id anyway.

crates/tn-reth/src/worker.rs
sstanfield: Sure.  I did not find the RPC to expose this also so I put it in the don't care pile.  I know ENRs are a thing but given our net arch they may or may not fit as well.

grantkee (approved):
LGTM

PR #315, opened 2025-05-23
Epoch Transition

grantkee:
- update epoch manager to fully close epoch at boundary
- move logic to close epoch from subscriber to epoch manager
- use mpsc instead of broadcast for consensus output
- export canonical state stream directly
- follow up PR to address test failures and cleanup

### DIFF
--- a/crates/consensus/primary/src/consensus_bus.rs
+++ b/crates/consensus/primary/src/consensus_bus.rs
@@ -7,10 +7,7 @@ use crate::{
     proposer::OurDigestMessage, state_sync::CertificateManagerCommand, RecentBlocks,
 };
 use consensus_metrics::metered_channel::{self, channel_with_total_sender, MeteredMpscChannel};
-use std::{
-    error::Error,
-    sync::{atomic::AtomicBool, Arc},
-};
+use std::{error::Error, fmt, sync::Arc};
 use tn_config::Parameters;
 use tn_primary_metrics::{ChannelMetrics, ConsensusMetrics, ExecutorMetrics, Metrics};
 use tn_types::{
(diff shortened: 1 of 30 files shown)


crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -57,6 +54,38 @@ impl NodeMode {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match self {
+            Self::Unknown => "unknown",
+            Self::Sync => "sync",
+            Self::Epoch => "epoch",
+        };
+        write!(f, "{s}")
+    }
+}
+
```
grantkee, line 88: I removed this in the follow up pr. The logic handles errors and loops regardless of the "reason"

crates/tn-reth/src/lib.rs
```diff
@@ -116,7 +113,7 @@ pub use reth_cli_util::{parse_duration_from_secs, parse_socket_address};
 pub use reth_errors::{CanonicalError, ProviderError, RethError};
 pub use reth_node_core::args::LogArgs;
 pub use reth_primitives_traits::crypto::secp256k1::sign_message;
-pub use reth_provider::ExecutionOutcome;
+pub use reth_provider::{CanonStateNotificationStream, ExecutionOutcome};
```
grantkee, line 116: re-export to keep reth isolated still

sstanfield: This is good.

sstanfield (approved):
LGTM, assume you still have some followup engine/reth stuff in works.

sstanfield (approved):

PR #318, opened 2025-05-27
update tests to read from genesis.json

grantkee:
- refactor cli to ensure chainspec passed through CLI is used
  - prev behavior was inconsistent with default `tn-config` loading
- validate genesis instead of chainspec
  - correct the validation process to sign the correct genesis
- rename `Config::store_path` to `Config::write_to_path`
- add timestamp to genesis for e2e tests to prevent epoch boundary closing on first block
- uses genesis.json for e2e tests to ensure consistent 
- add `TaskSpawner` to engine for spawning blocking engine task

### DIFF
--- a/bin/telcoin-network/src/genesis/mod.rs
+++ b/bin/telcoin-network/src/genesis/mod.rs
@@ -132,7 +132,7 @@ impl GenesisArgs {
 
                 // initialize genesis
                 tn_config.genesis = self.chain.genesis().clone();
-                debug!(target: "init", ?config_path, "load config from path:\n{:#?}", tn_config);
+                debug!(target: "cli", ?config_path, "load config from path:\n{:#?}", tn_config);
 
                 if !init.is_empty() {
                     // Changed a default config setting so update and save.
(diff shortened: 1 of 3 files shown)


bin/telcoin-network/src/genesis/create_committee.rs
```diff
@@ -207,15 +200,16 @@ impl CreateCommitteeArgs {
         // write genesis and config to file
         //
         // NOTE: CLI parser only supports JSON format for genesis
-        Config::store_path(data_dir.genesis_file_path(), tn_config.genesis(), ConfigFmt::JSON)?;
-        Config::store_path(config_path, tn_config, ConfigFmt::YAML)?;
+        Config::write_to_path(data_dir.genesis_file_path(), tn_config.genesis(), ConfigFmt::JSON)?;
```
sstanfield, line 203: There is a reason genesis is JSON not YAML like the other config files?  I don;t think this really matters but a small consistency thing.

crates/types/src/task_manager.rs
```diff
@@ -123,6 +123,20 @@ impl TaskSpawner {
         }
     }
 
+    /// Spawns a non-critical, blocking task on tokio and records the JoinHandle and name.
+    ///
+    /// Other tasks are unaffected when this task resolves.
+    pub fn spawn_blocking_task<F, S: ToString>(&self, name: S, task: F)
```
sstanfield, line 129: Nice.  I don't think these tasks will respond to aborts, etc but also we have timeouts so I don't think that matters.

sstanfield (approved):
Looks good, nice startup cleanups.

PR #320, opened 2025-05-28
Add work on the tn RPC namespace.

sstanfield:
This adds a couple of new functions to the tn RPC namespace.  It also makes some changes to serde for more human readable JSON and some logs.

Can call new functions added with:
```curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"tn_latestHeader","params":[],"id":74}' 127.0.0.1:8545```
and
```curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"tn_genesis","params":[],"id":74}' 127.0.0.1:8545```

### CODE REVIEW

File: crates/network-libp2p/src/consensus.rs [hunk 1]
```diff
@@ -371,10 +371,13 @@ where
 
     /// Publish our network addresses and peer id under our BLS public key for discovery.
     fn publish_our_data(&mut self) {
-        if let Some(record) = self.get_peer_record() {
+        if let Some(kad_record) = self.get_peer_record() {
+            // Convert to KadRecord for a "cleaner" debug print.  Do we really need this record in
+            // log?
+            let record: KadRecord = kad_record.clone().into();
             info!(target: "network-kad", ?record, "Publishing our record to kademlia");
             if let Err(err) =
```
grantkee, line 379: We can remove this in a follow up PR. I mostly added it as a debug tool, but it isn't very helpful. I think a better log would be the confirmed `observed_addr` from the identify event

crates/node/src/lib.rs
```diff
@@ -48,3 +52,31 @@ where
     // return result after shutdown
     res
 }
+
+pub struct EngineToPrimaryRpc<DB> {
+    /// Container for consensus channels.
+    consensus_bus: ConsensusBus,
+    /// Consensus DB
+    db: DB,
+}
```
grantkee, line 61: I'm weary of adding more consensus domain to the execution layer. However, I think this is an elegant solution for the monolithic node approach we're taking to launch mainnet. In the future, this should be an api call between processes/servers. Good job finding a simple approach to meet our deadline

grantkee (approved):
I had two comments, but they are not show stoppers. LGTM

crates/node/src/lib.rs
sstanfield: Agree totally and the goal was to keep this behind a trait we could replace with IPC or whatever later.

PR #323, opened 2025-06-02
Rename ValidatorInfo and validator.yaml to reflect they are for a node not just a validator.

sstanfield:
This just renames stuff, both structure and files, to better reflect that the info is for any node not just a validator.  By labelling this stuff "validator" it could create confusion when trying to understand the code (or config).


grantkee (approved):
LGTM

PR #324, opened 2025-06-05
Implement infrastructure for setting base fees per worker per epoch.

sstanfield:
This should provide what we need to manage base fees per epoch.

Need to do a calculation.  Also may need to load the accumulator on a mid-epoch restart.

### DIFF
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -5,7 +5,7 @@
 use crate::error::ExecutionError;
 use eyre::OptionExt;
 use jsonrpsee::http_client::HttpClient;
-use std::{collections::HashMap, net::SocketAddr, sync::Arc};
+use std::{net::SocketAddr, sync::Arc};
 use tn_batch_builder::BatchBuilder;
 use tn_batch_validator::BatchValidator;
 use tn_config::Config;
@@ -18,6 +18,7 @@ use tn_reth::{
 };
 use tn_rpc::{EngineToPrimary, TelcoinNetworkRpcExt, TelcoinNetworkRpcExtApiServer};
 use tn_types::{
+    gas_accumulator::{BaseFeeContainer, GasAccumulator},
     Address, BatchSender, BatchValidation, ConsensusOutput, ExecHeader, Noticer, SealedHeader,
     TaskSpawner, WorkerId, B256, MIN_PROTOCOL_BASE_FEE,
 };
@@ -40,7 +41,8 @@ pub(super) struct ExecutionNodeInner {
     /// Optional args to turn on faucet (for testnet only).
     pub(super) opt_faucet_args: Option<FaucetArgs>,
     /// Collection of execution components by worker.
-    pub(super) workers: HashMap<WorkerId, WorkerComponents>,
+    /// Index of vec is worker id.
+    pub(super) workers: Vec<WorkerComponents>,
 }
 
 impl ExecutionNodeInner {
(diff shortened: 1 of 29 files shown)


crates/node/src/manager.rs
```diff
@@ -155,7 +150,12 @@ where
 
         // create the engine
         let engine = self.create_engine(&engine_task_manager, reth_db)?;
-        engine.start_engine(for_engine, self.node_shutdown.subscribe()).await?;
+        // Create our epoch gas accumulator, we currently have one worker.
+        // All nodes have to agree on the worker count, do not change this for an existing chain.
+        let gas_accumulator = GasAccumulator::new(1);
```
grantkee, line 155: this will not survive restarts. is there an issue to add this feature?

crates/types/src/gas_accumulator.rs
```diff
@@ -0,0 +1,109 @@
...
+use alloy::eips::eip1559::MIN_PROTOCOL_BASE_FEE;
+use parking_lot::{Mutex, RwLock};
+
+use crate::WorkerId;
+
+/// An interiour mutable container for a workers base fee.
+#[derive(Clone, Debug)]
+pub struct BaseFeeContainer {
+    base_fee: Arc<RwLock<u64>>,
+}
```
grantkee, line 12: nit: this doesn't change during the epoch. since the container is passed to batch builder and batch validator, `AtomicU64` is probably fine here. not a blocker though

crates/types/src/gas_accumulator.rs
```diff
@@ -0,0 +1,109 @@
+use std::sync::Arc;
```
grantkee, line 1: module doc comment here?

crates/types/src/gas_accumulator.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn default() -> Self {
+        Self::new(MIN_PROTOCOL_BASE_FEE)
+    }
+}
+
+#[derive(Debug, Default)]
+struct Inner {
+    blocks: u64,
+    gas_used: u64,
+    gas_limit: u64,
```
grantkee, line 41: docs?

grantkee (approved):
LGTM! I found some nits, but nothing worth blocking this PR

crates/node/src/manager.rs
sstanfield: The next PR addresses this.

crates/node/src/engine/inner.rs
```diff
@@ -165,7 +174,11 @@ impl ExecutionNodeInner {
 
         // take ownership of worker components
         let components = WorkerComponents::new(rpc_handle, transaction_pool, network);
-        self.workers.insert(worker_id, components);
+        // Must call this function in accending worker_id order or will panic.
+        if worker_id as usize != self.workers.len() {
+            panic!("initialize_worker_components not called with sequencial worker ids!")
```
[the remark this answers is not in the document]
sstanfield: The panic was deliberate.  This function should be called as part or initialization and should never be called "wrong" unless the calling code is just doing it wrong.  I put the panic in since I think blowing up quickly and noisily beats an error someone might ignore, etc.

crates/types/src/gas_accumulator.rs
sstanfield: I think these are private and kinda obvious but it does not hurt to doc them as being on a per epoch since that may not be obvious.

crates/node/src/engine/inner.rs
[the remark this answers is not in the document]
grantkee: Good point

PR #327, opened 2025-06-07
Update tn contracts

grantkee:
Update tn-contracts

### DIFF
--- a/bin/telcoin-network/src/args/utils.rs
+++ b/bin/telcoin-network/src/args/utils.rs
@@ -3,8 +3,7 @@
 use eyre::OptionExt;
 use std::str::FromStr;
 use tn_reth::{dirs::DataDirPath, MaybePlatformPath};
-use tn_types::Address;
-use tn_types::U256;
+use tn_types::{Address, U256};
 
 pub use tn_reth::clap_genesis_parser;
 
--- a/bin/telcoin-network/src/args/utils.rs
+++ b/bin/telcoin-network/src/args/utils.rs
@@ -1,10 +1,10 @@
 //! Utilities for parsing args
 
-use alloy::primitives::aliases::U232;
 use eyre::OptionExt;
 use std::str::FromStr;
 use tn_reth::{dirs::DataDirPath, MaybePlatformPath};
 use tn_types::Address;
+use tn_types::U256;
 
 pub use tn_reth::clap_genesis_parser;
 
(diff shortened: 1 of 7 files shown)


bin/telcoin-network/src/args/utils.rs
```diff
@@ -27,11 +26,11 @@ pub fn clap_address_parser(value: &str) -> eyre::Result<Address> {
     Ok(address)
 }
 
-/// Parse 18 decimal U232 from string for ConsensusRegistry.
-pub fn clap_u232_parser(value: &str) -> eyre::Result<U232> {
-    let parsed_val = U232::from_str_radix(value, 10)?
-        .checked_mul(U232::from(10).checked_pow(U232::from(18)).expect("1e18 exponentiation"))
-        .ok_or_eyre("U232 parsing")?;
+/// Parse 18 decimal U256 from string for ConsensusRegistry.
+pub fn clap_u256_parser_to_18_decimals(value: &str) -> eyre::Result<U256> {
```
robriks, line 30: In theory the `alloy::parse_ether` utility would suffice here but it does exactly the same thing and this is lightweight without adding a dependency to the binary

crates/tn-reth/src/lib.rs
```diff
@@ -1079,27 +1081,16 @@ impl RethEnv {
...
-        let validators = token_ids
-            .into_iter()
-            .map(|token_id| {
-                // read validator info
-                self.get_validator_by_token_id(token_id, &mut tn_evm)
-            })
-            .collect::<eyre::Result<Vec<_>, _>>()?;
 
+        // retrieve the committee
+        let validators = self.get_committee_validators_by_epoch(epoch, &mut tn_evm)?;
```
robriks, line 1093: Oh this is so much nicer, glad you mentioned adding in this fn

crates/tn-reth/src/system_calls.rs
```diff
@@ -86,18 +94,40 @@ sol!(
             /// The governance-issued consensus NFT token id.
             uint24 tokenId;
             /// The validator balance.
-            uint232 balance;
+            uint256 balance;
```
robriks, line 97: now that `tokenId == uint256(uint160(validatorAddress))` this struct can be deleted! it no longer exists in the `StakeManager` implementation

robriks (changes requested):
Awesome stuff, thanks for this! Only requested change is to remove the `StakeInfo` struct

robriks (approved):
LGTM!

sstanfield (approved):

PR #330, opened 2025-06-10
Update Epoch Value Retrieved from State

grantkee:
- sort validator array before close epoch - shuffle test passing
- fix unit test


sstanfield (approved):
Looks good.

PR #335, opened 2025-06-10
Config readme

grantkee:
- add readme for config
- cleanup some old todos
- add some doc comments


crates/config/README.md
```diff
@@ -0,0 +1,97 @@
...
+### KeyConfig
+
+The most critical config for a node's security.
+This contains the node's keys used to authenticate protocol messages.
+
+The protocol requires access to three keys in the current version:
+
+- Primary BLS12-381 key
+- Primary Ed25519 network key
+- Worker Ed25519 network key (only support 1 worker per node for now)
```
sstanfield, line 36: It is probably worth mentioning in this section that the BLS private key is not exposed outside KeyConfig while the network private keys are so that they can be used with libp2p.  The BLS key is the more important key to impersonate a validator and network keys can theoretically be rotated (although we don't do that yet).

sstanfield (approved):

robriks (approved):
LGTM! Great info for auditors & new devs

PR #336, opened 2025-06-11
close epochs unit test

grantkee:
- unit test to close multiple epochs and shuffle in new validator
- remove closing epoch logs appended to receipts
    - tn-reth static file writer assumes tx per receipt


sstanfield (approved):
LGTM

PR #338, opened 2025-06-11
Refactor CLI chain/genesis and remove old adiri for a new one.

sstanfield:
The added configs are just placeholders for now.


chain-configs/mainnet/committee.yaml
```diff
@@ -0,0 +1,32 @@
+---
+authorities:
+  yBWc3pW4bXPmYyADsoPhspx6UnFEhx5ByRpU8mEvGYftVcHMC4cXL1ZQAgkyjqE7XHowKtf8Z7Cf1yUghYFBtS72WfDjzg51ZaQ85ohvQwAWDhgTBffrtPhmJqMd83J51Yb:
+    protocol_key: yBWc3pW4bXPmYyADsoPhspx6UnFEhx5ByRpU8mEvGYftVcHMC4cXL1ZQAgkyjqE7XHowKtf8Z7Cf1yUghYFBtS72WfDjzg51ZaQ85ohvQwAWDhgTBffrtPhmJqMd83J51Yb
+    voting_power: 1
+    primary_network_address: /ip4/127.0.0.1/udp/49306/quic-v1
+    execution_address: "0x3333333333333333333333333333333333333333"
```
grantkee, line 7: nit: in preparation for the security audits, we should probably use random addresses here. Otherwise, it could be flagged as an invalid execution address without a known private key

chain-configs/testnet/committee.yaml
```diff
@@ -0,0 +1,32 @@
+---
+authorities:
+  yBWc3pW4bXPmYyADsoPhspx6UnFEhx5ByRpU8mEvGYftVcHMC4cXL1ZQAgkyjqE7XHowKtf8Z7Cf1yUghYFBtS72WfDjzg51ZaQ85ohvQwAWDhgTBffrtPhmJqMd83J51Yb:
+    protocol_key: yBWc3pW4bXPmYyADsoPhspx6UnFEhx5ByRpU8mEvGYftVcHMC4cXL1ZQAgkyjqE7XHowKtf8Z7Cf1yUghYFBtS72WfDjzg51ZaQ85ohvQwAWDhgTBffrtPhmJqMd83J51Yb
+    voting_power: 1
+    primary_network_address: /ip4/127.0.0.1/udp/49306/quic-v1
+    execution_address: "0x3333333333333333333333333333333333333333"
```
grantkee, line 7: For testnet, we probably want to start using real addresses as well so we can test validator stake/withdraw

grantkee (approved):
I left two thoughts regarding the configs, but your description mentions they're just placeholders for now. We just need to be sure we communicate that for the competition. LGTM - nice work

chain-configs/testnet/committee.yaml
sstanfield: Agree on both points about addresses.  My plan is to update both of these files (hopefully we have the "real" testnet files pre comp) and probably should add a small readme to the directory.  If someone selects main net pre launch we can either panic or use testnet.  Probably panic.

PR #341, opened 2025-06-13
Rework initial p2p network config.

sstanfield:


bin/telcoin-network/src/keytool/generate.rs
```diff
@@ -71,33 +91,54 @@ impl KeygenArgs {
...
+            .unwrap_or_else(|| get_available_udp_port(&worker_host).unwrap_or(49584));
+        for worker in node_info.p2p_info.worker_index.0.iter_mut() {
             worker.name = network_publickey.clone();
+            worker.worker_address =
+                format!("/ip4/{worker_host}/udp/{worker_udp_port}/quic-v1").parse()?;
+            worker_udp_port = if self.worker_base_port.is_none() {
+                get_available_udp_port(&worker_host).unwrap_or(49584)
+            } else {
+                worker_udp_port + 1
+            };
```
grantkee, line 123: this works for workers with the same IP, but will be a problem if they're on different servers

bin/telcoin-network/src/keytool/generate.rs
```diff
@@ -71,33 +91,54 @@ impl KeygenArgs {
...
         // network keypair for authority
         let network_publickey = key_config.primary_network_public_key();
-        node_info.primary_info.network_key = network_publickey;
+        node_info.p2p_info.network_key = network_publickey;
+        let primary_host = self.primary_host.clone().unwrap_or("127.0.0.1".to_string());
+        let primary_udp_port = self
+            .primary_port
+            .unwrap_or_else(|| get_available_udp_port(&primary_host).unwrap_or(49584));
+        node_info.p2p_info.network_address =
+            format!("/ip4/{primary_host}/udp/{primary_udp_port}/quic-v1").parse()?;
```
grantkee, line 107: this only supports IPv4 which is fine for now, but I think it's likely node operators want to use IPv6

bin/telcoin-network/src/keytool/generate.rs
sstanfield: Uses multiaddr now and supports a list for worker addresses.

bin/telcoin-network/src/keytool/generate.rs
sstanfield: Uses multi addrs now.

bin/telcoin-network/src/keytool/generate.rs
```diff
@@ -71,33 +86,65 @@ impl KeygenArgs {
...
-        node_info: &mut NodeInfo,
         tn_datadir: &TND,
         passphrase: Option<String>,
     ) -> eyre::Result<()> {
         info!(target: "tn::generate_keys", "generating keys for full validator node");
+        let mut node_info = NodeInfo::default();
+        if self.workers != 1 {
+            return Err(eyre::eyre!("Only supports a single worker at this time!"));
+        }
+        /* Uncomment when multi-worker support is enabled
```
sstanfield, line 134: Comment deliberately left in.

etc/local-testnet.sh
```diff
@@ -102,6 +102,7 @@ else
         target/${RELEASE}/telcoin-network genesis \
             --datadir "${ROOTDIR}" \
             --chain-id 0x1e7 \
+            --epoch-duration-in-secs 60 \
```
sstanfield, line 105: 60 second epochs for testnet seem OK?  I think we should shorten these for testing but maybe this is too far?

grantkee (approved):
LGTM

PR #344, opened 2025-06-14
Epoch Boundary E2E Test

grantkee:
Epoch transition e2e test


crates/consensus/primary/src/certifier.rs
grantkee, line 1: silence some noisy logs

crates/consensus/primary/src/proposer.rs
```diff
@@ -243,7 +243,8 @@ impl<DB: Database> Proposer<DB> {
         let mut total_inclusion_secs = 0.0;
         for digest in &digests {
             let batch_inclusion_secs =
-                Duration::from_secs(*header.created_at() - digest.timestamp).as_secs_f64();
+                Duration::from_secs(header.created_at().saturating_sub(digest.timestamp))
```
grantkee, line 246: this overflowed when the epoch tests were set to 3s epochs?

sstanfield (approved):

PR #345, opened 2025-06-14
Add base-fee recipient

robriks:
Includes the base fee recipient as the TAO governance safe genesis precompile with TEL balance for gas


sstanfield (dismissed):
Looks good.  I assume you did a sanity check with a local testnet to make sure the safe got the funds when using the default?  I think there is a setting in the local-test net script that should probably just be this default?  Or maybe not since someone running a local instance may just want baseless to go to an address of there choosing.

grantkee (dismissed):

PR #347, opened 2025-06-16
update dockerfile for deployments

grantkee:
dockerfile updates to bring in chain configs once they are up to date


sstanfield (approved):
Is the worker_cache right?

PR #348, opened 2025-06-17
Tweak epoch code to hopefully avoid some races.

sstanfield:
- Move the metrics service to the node task manager.
- Attempt to wait until the epoch task manager has shutdown before restarting.  This can timeout pretty quickly.
- Fixed a race with the network event_stream during epoch rotation.


grantkee (approved):
LGTM! Great work

PR #354, opened 2025-07-30
Cantina-74: Support bls precompiles

grantkee:
This PR is a precursor to verifying bls proof-of-possession on-chain when validators stake. See cantina #845

- upgrade to cancun 
  - reject blob transactions
- upgrade to prague
  - access to bls precompiles


chain-configs/testnet/genesis.yaml
```diff
@@ -242,4 +244,12 @@ alloc:
       "0x457f3fc26bf430b020fe76358b1bfaba57e1657ace718da6437cda9934eabfe9": "0x0000000000000000000000000000000000000000000000000000000000000001"
       "0x6c0afcdb9e0038f7b4b65f772a283b1f0460275f9e07faa8cf0c9a492a238f13": "0x0000000000000000000000000000000000000000000000000000000000000001"
       "0xca458dc12368669a3b8c292bc21c1b887ab1aa386fa3fcc1ed972afd74a330ca": "0x000000000000000000000000c1612c97537c2cc62a11fc4516367ab6f62d4b23"
+  "0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02":
```
robriks, line 247: I didn't realize we were putting the genesis account allocs here also, is the plan to manually copy them over from the programatically generated values in tn-contracts genesis yaml? just so I know to do that in the future in case of any changes in the solidity-generated precompile outputs

crates/tn-reth/src/evm/block.rs
```diff
@@ -320,6 +354,98 @@ where
...
+            return Ok(());
+        }
+
+        let mut res = match self.evm.transact_system_call(
+            SYSTEM_ADDRESS,
+            BEACON_ROOTS_ADDRESS,
+            parent_beacon_block_root.0.into(),
+        ) {
+            Ok(res) => res,
+            Err(e) => {
```
robriks, line 390: at some point it could be nice to bubble up tx execution that fails just to make debugging a bit more straightforward when using `transact_system_call` or `transact_pre_genesis_create`. currently we only check for non-EVM `Err`s so executions that fail/revert/halt/return false slip through without bubbling up and require some investigation. def a dev X nit and not worth adding in everywhere right now but will keep it in mind if we get more bandwidth later

robriks (approved):
great stuff! LGTM, put a couple questions/musings in but they're superfluous in nature

sstanfield (approved):

crates/tn-reth/src/evm/block.rs
grantkee: Good idea. I added error logging so these errors will show up in the logs

sstanfield (approved):

PR #361, opened 2025-08-04
Cantina-30: sanitize missing certificates requests

grantkee:
- sanitize requested max items before allocating memory for vec
- use same limit for all requested certs from peers


sstanfield (approved):
LGTM

PR #362, opened 2025-08-05
Cantina-364: batch validation overflow

grantkee:
- use checked addition when validating batches to handle possible u64 gas overflow


sstanfield (approved):
LGTM

PR #363, opened 2025-08-05
Cantina-592: catchup accumulator `blockHeight`

grantkee:
- use `blockHeight` instead of epoch start timestamp for block range
- make basefee address non-optional
    - makes testing easier
    - this should not be an option in production
    - default through CLI was always `Some`

The first commit [59fd548](https://github.com/Telcoin-Association/telcoin-network/pull/363/commits/59fd54877169a53f8d715ba94d5fe3eacfdd8f52) implements the patch. The rest of the changes were test related.


sstanfield (approved):
Looks good.  Like the removed clones, feels like a clip thing so not sure how they survived to this point...

PR #365, opened 2025-08-11
Cantina-594: Catchup Accumulator Rewards

grantkee:
- recover rewards for current epoch once per round (not per executed block)


crates/consensus/primary/src/network/handler.rs
grantkee, line 1: clippy

PR #366, opened 2025-08-13
Cantina-615: Closing Epoch Batch Index

grantkee:
- update output logic for closing epoch to account for batches `pop`ing


sstanfield (approved):
Looks good.

PR #374, opened 2025-09-05
use message size constraints for MissingCertificatesRequest

grantkee:
WIP


crates/consensus/primary/src/state_sync/cert_collector.rs
```diff
@@ -130,17 +163,24 @@ where
...
+    /// - the time limit has expired
     fn max_limits_reached(&self) -> bool {
-        self.items_returned >= self.max_items
-            || self.start_time.elapsed()
-                >= self
-                    .config
-                    .network_config()
-                    .sync_config()
-                    .max_db_read_time_for_fetching_certificates
+        let size = self.max_message_size - self.accumulated_size <= *LOCAL_MIN_REQUEST_SIZE;
```
DrasticWatermelon, line 170: `<=` here imo should be `<` but I'm assuming that a message with size exactly at `LOCAL_MIN_REQUEST_SIZE` can exist and be provided here

DrasticWatermelon:
I'm not sure I understand how the PR addresses the ban of peers requesting >1MB worth of batches/certs. I understand that we're now limiting fetch responses from honest nodes to 1MB but I dont see how this changes the ban behaviour

PR #377, opened 2025-09-12
Cantina 3.1.17 - Block Beneficiary

grantkee:
- use correct beneficiary address for execution block priority fees
  - remove leader's address from `ConsensusOutput` since this is managed by `RewardsCounter` now
  - include batch producer's execution address with consensus output batches
- update engine tests


crates/consensus/executor/src/subscriber.rs
```diff
@@ -309,29 +326,13 @@ impl<DB: Database> Subscriber<DB> {
...
-
-        let early_finalize = if self.consensus_bus.node_mode().borrow().is_active_cvv() {
-            // We are a CVV so we can finalize early.
-            true
-        } else {
-            // Not a CVV so be more conservative about finalizing blocks.
-            false
-        };
+        // active cvvs always finalize early bc they are the authorities
+        let early_finalize = self.consensus_bus.node_mode().borrow().is_active_cvv();
```
sstanfield, line 330: I wonder who wrote this "if" originally (think it was me)....  I can only hope there was more to that "if" at one point....  Still no excuse for this, good catch.

crates/state-sync/src/lib.rs
```diff
@@ -140,10 +140,12 @@ pub fn save_consensus<DB: Database>(
 ) -> eyre::Result<()> {
     match db.write_txn() {
         Ok(mut txn) => {
-            for batch in consensus_output.batches.iter().flatten() {
-                if let Err(e) = txn.insert::<Batches>(&batch.digest(), batch) {
-                    error!(target: "state-sync", ?e, "error saving a batch to persistant storage!");
-                    return Err(e);
+            for certified_batches in consensus_output.batches.iter() {
```
sstanfield, line 143: Just curious but what was wrong with flatten() here?  It seems more efficient (or at least less code) but assume you had a reason to double iterate?

crates/engine/tests/it/main.rs
```diff
@@ -343,34 +379,88 @@ async fn test_empty_output_executes_late_finalize() -> eyre::Result<()> {
...
 #[tokio::test]
-async fn test_queued_output_executes_after_sending_channel_closed() -> eyre::Result<()> {
+async fn test_happy_path_full_execution_even_after_sending_channel_closed() -> eyre::Result<()> {
     let tmp_dir = TempDir::new().expect("temp dir");
     // create batches for consensus output
     let mut batches_1 = tn_reth::test_utils::batches(4); // create 4 batches
     let mut batches_2 = tn_reth::test_utils::batches(4); // create 4 batches
 
+    // add eip1559 transactions to set max priority fee per gas so batch producer earns fees
+    let genesis = adiri_genesis();
```
sstanfield, line 393: Probably want to use test_genesis() instead of adiri_genesis() since adiri_genesis() will produce a genesis with increasingly stale start time.  This could throw some non-determinism into the first epoch for instance.  If you do this you may want to update this everywhere in this file.  Unless you want the stale start times but not sure why you would.

sstanfield (approved):
Looks good.  Approving since you can make the call about adiri_genesis() but I think that should be replaced per the comment.  The other comments are not critical.

crates/state-sync/src/lib.rs
grantkee: `ConsensusOutput::batches` returns a struct with another field called `batches` that returns a `Vec<Batch>`

The subscriber is responsible for converting batch digests into batches with transactions for the engine. I was torn on where the business logic for converting `AuthorityIdentifier` -> ecdsa `Address`. The engine has access to this through the `GasAccumulatr::RewardsCounter` but I didn't want to have to obtain a lock for every batch. That being said, we could retrieve the map at the start of execution in the payload builder

crates/engine/tests/it/main.rs
grantkee: Good catch

crates/engine/tests/it/main.rs
DrasticWatermelon, line 1: changes to this file seem (mostly) unrelated to fixing 3.1.17 a part from some tests verifying the basefee payout is handled correctly, am I right or am I missing something?

grantkee: Yes, I updated the tests for more complete coverage. The focus for 3.1.17 was mostly [lines 666-704](https://github.com/Telcoin-Association/telcoin-network/blob/3d68f43b4029b29343a6ec438039e64169cfc556/crates/engine/tests/it/main.rs#L666-L704)

The basefees go to the governance safe, priority fees to the batch producer (3.1.17), and block rewards at the end of the epoch match the `RewardCounter` (updated by `EpochManager` normally)

PR #379, opened 2025-09-18
Fix a regression in test speed, don't call test_genesis() in loops.

sstanfield:
test_genesis() is more complete now and should not be called in loops in tests.  This will slow them down a lot.


grantkee (dismissed):
Good work on this!

PR #381, opened 2025-09-19
Use procs instead of multiple tokio runtimes for a test that sometimes failed.

sstanfield:
This test sometimes just started failing.  This seems to improve things (made a laptop that was in a failing state work).


grantkee (approved):
LGTM!

PR #383, opened 2025-09-25
Cantina-244 Precursor: Kad uses Peer Manager

grantkee:
- update target number of peers to better support network behaviours
- manually add peers to kademlia routing table so PeerManager has control over connections with peers
- additional testing
- remove `inbound_requests` on success (prev growing infinitely large)
- use bls key for `PeerExchange` so receiving node can also disconnect (happy path to support honest nodes)


sstanfield (approved):
LGTM, like the test expansion.

Issue #402, opened 2025-09-29
`QuorumWaiter::verify_batch` doesn't decrease `available_stake` when matches `Ok`


leopashov:
**Cantina Finding:** # 804  
**Severity:** Low  

---

#### Finding Description

`total_stake` is increased but `available_stake` isn't decreased when a peer successfully acknowledges a batch (`Ok(stake)`), and this causes inflation of `total_stake + available_stake`.

This causes the anti-quorum condition `total_stake + available_stake < threshold` to never be met, even when quorum is no longer possible. As a result, the function waits until timeout rather than exiting early, wasting time and reducing throughput.

This harms Telcoin Network’s performance — reducing TPS and breaking core functionality of quorum responsiveness.

---

#### Proof of Concept

Create a committee with four authorities, each with equal stake.  
Set quorum threshold = 3.

Run `verify_batch` where:
- One peer replies `Ok`
- Two peers reply with `Network` errors
- One peer never responds

Current behavior:
- `total_stake = 2`
- `available_stake = 1` (only reduced for the two failures)
- `total_stake + available_stake = 3`, so the anti-quorum condition `(< 3)` does **not** trigger
- Function waits until the final peer times out

Expected behavior:
- After the two failures and one success, `available_stake` should be `0`
- Then: `total_stake + available_stake = 2 < 3`, so the function exits early with `AntiQuorum`

Example fix in Rust:

```rust
match res {
    Ok(stake) => {
        total_stake += stake;
        available_stake -= stake; // new line
        if total_stake >= threshold {
            return Ok(());
        }
    }
    ...
}
```

---

#### Recommendation

Update `QuorumWaiter::verify_batch` to decrease `available_stake` when receiving a successful `Ok(stake)` response:

```rust
Ok(stake) => {
    total_stake += stake;
    available_stake -= stake;
}
```

Add a regression test simulating the described failure scenario. Ensure that quorum failure is detected early and the function exits without waiting for timeout.

PR #452, opened 2025-10-03
Make sure we count available stake correctly.

sstanfield:
https://github.com/Telcoin-Association/telcoin-network/issues/402


grantkee (approved):
Great work!

PR #455, opened 2025-10-06
Break CommitteeFixture and related structs out of tn-test-utils.

sstanfield:
This removes a circular dependency in tests that messed up rust analyzer.

This just moves some of the code in tn-test-utils to tn-test-utils-committee which does not need to depend on primary.


grantkee (approved):
Perfect! Thank you for fixing this

Issue #441, opened 2025-09-30
primary: requested_parents DoS issue


leopashov:
**Cantina finding**: #1095
**Severity**: Medium

---

**Summary**  
Primary is vulnerable to Denial of Service issue.

Malicious committee members can cause unbounded memory growth of `requested_parents` map by sending vote requests for arbitrarily future rounds.

These entries stay forever in memory, because garbage collection mechanism only removes entries from past rounds.

---

**Finding Description**  
Let's look at the code from `handler.rs`, to see how incoming vote requests are processed:

```rust
pub(crate) async fn vote(
    &self,
    peer: PeerId,
    header: Header,
    parents: Vec<Certificate>,
) -> PrimaryNetworkResult<PrimaryResponse> {
    // current committee
    let committee = self.consensus_config.committee();

    // validate header
    header.validate(committee, self.consensus_config.worker_cache())?;

    // validate parents
    let num_parents = parents.len();
    ensure!(
        num_parents <= committee.size(),
        HeaderError::TooManyParents(num_parents, committee.size()).into()
    );
    self.consensus_bus
        .primary_metrics()
        .node_metrics
        .certificates_in_votes
        .inc_by(num_parents as u64);

    let committee_peer: AuthorityIdentifier = peer.into();
    ensure!(
        self.consensus_config.in_committee(&committee_peer),
        HeaderError::UnknownNetworkKey(peer).into()
    );
    ensure!(header.author() == &committee_peer, HeaderError::PeerNotAuthor.into());

    ...
    ...
    if parents.is_empty() {
        // check if any parents missing
        let missing_parents = self.check_for_missing_parents(&header).await?;
        if !missing_parents.is_empty() {
            // return request for missing parents
            debug!(
                "Received vote request for {:?} with unknown parents {:?}",
                header, missing_parents
            );
            return Ok(PrimaryResponse::MissingParents(missing_parents));
        }
    } else {
        ...
    }
}
```

Let's look at the `check_for_missing_parents` function:

```rust
async fn check_for_missing_parents(
    &self,
    header: &Header,
) -> HeaderResult<Vec<CertificateDigest>> {
    // identify parents that are neither in storage nor pending
    let mut unknown_certs = self.state_sync.identify_unkown_parents(header).await?;

    // ensure header is not too old
    let limit = self.consensus_bus.primary_round_updates().borrow().saturating_sub(
        self.consensus_config.network_config().sync_config().max_proposed_header_age_limit,
    );
[...]

Issue #399, opened 2025-09-29
Vote Resending Logic is Unbounded and Allows Request Amplification


leopashov:
**Cantina Finding:** # 720  
**Severity:** Low    

---

#### Summary

Vote resending is done without rate limiting or caching. A peer can continuously request votes for the same header. If the digest matches, the node regenerates and resends the vote each time.

---

#### Root cause

No memoization or throttle logic in vote resend path.

memoization - (results of expensive function calls are stored (cached) so that when the same inputs occur again, the cached result is returned instead of recomputing it)

---

#### Finding Description

When a previous vote is found for a round and the digest matches, this code runs:

```rust
return Ok(PrimaryResponse::Vote(vote));
```

This regenerates a vote object using `Vote::new(...)` which includes signing and can be expensive.

---

#### Proof of Concept

```rust
use std::time::{Duration, Instant};

#[derive(Clone, Debug, PartialEq)]
struct Header {
    author: u64,
    round: u64,
    digest: u64,
}

#[derive(Clone, Debug, PartialEq)]
struct Vote {
    header_digest: u64,
    round: u64,
    signature: u64,
}

#[derive(Clone, Debug)]
struct VoteInfo {
    round: u64,
    vote_digest: u64,
}

struct ConsensusConfig {
    authority_id: u64,
}

impl ConsensusConfig {
    fn new() -> Self {
        ConsensusConfig { authority_id: 1 }
    }

    fn authority_id(&self) -> Option<u64> {
        Some(self.authority_id)
    }

    fn read_vote_info(&self, author: u64) -> Result<Option<VoteInfo>, String> {
        if author == 1 {
            Ok(Some(VoteInfo {
                round: 100,
                vote_digest: 12345,
            }))
        } else {
            Ok(None)
        }
    }
}

// Simulate expensive vote creation (e.g., cryptographic Hull cryptographic signing
fn create_vote(header: &Header, authority_id: u64) -> Vote {
    let start = Instant::now();
    let mut sum = 0;
    for _ in 0..1_000_000 {
        sum += header.digest;
    }
    println!("Vote creation took {:?}", start.elapsed());
    Vote {
        header_digest: header.digest,
        round: header.round,
        signature: header.digest ^ authority_id,
    }
}

// Vulnerable vote processing function
fn process_vote(header: &Header, config: &ConsensusConfig) -> Result<Vote, String> {
    let previous_vote = config
        .read_vote_info(header.author)
[...]

Issue #451, opened 2025-10-03
Voting Weight


grantkee:
# Problem

Committee tracks stake as voting weight.

After careful consideration, weights should remain equal. We can remove this unused code on `Committee` and reduce complexity throughout the codebase.

## Solution

Remove weight calculations to prevent inconsistent security models for consensus and epoch boundary certificates.

NOTE: stake still has an effect on the amount of rewards a validator earns when `applyIncentives` is called on consensus registry.

PR #471, opened 2025-10-17
Reject votes for rounds too far in future.

sstanfield:
This should keep a node from sending vote requests with crazy future rounds to waste another nodes resources.
closes https://github.com/Telcoin-Association/telcoin-network/issues/441
Leverages the caching for vote tracking to cache vote results.
closes https://github.com/Telcoin-Association/telcoin-network/issues/399

### DIFF
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -30,7 +30,10 @@ use tn_types::{
 use tokio::sync::oneshot;
 use tracing::{debug, error, warn};
 
-type AuthEquivocationMap = BTreeMap<AuthorityIdentifier, (Epoch, Round, HeaderDigest)>;
+/// Map to hold vote info to detect invalid votes, equivocation and cache responses in case of
+/// rerequests.
+type AuthEquivocationMap =
+    BTreeMap<AuthorityIdentifier, (Epoch, Round, HeaderDigest, Option<PrimaryResponse>)>;
 
 /// The type that handles requests from peers.
 #[derive(Clone)]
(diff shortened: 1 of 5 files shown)


crates/consensus/primary/src/network/handler.rs
```diff
@@ -221,6 +221,18 @@ where
 
         // validate header
         header.validate(committee)?;
+        let max_round = *self.consensus_bus.committed_round_updates().borrow()
```
grantkee: Can a header outside of GC window be received anymore? I was thinking this would be caught on startup when consensus gossip is received

I suppose there is a race condition, but I'm not sure this is necessary since a future header would also have missing parents and hang the vote up

sstanfield: I am not sure what would catch this?  Another validator can request a vote on any header it wants, I don't see anything to filter that.  This all depends on a malicious validator going rogue but if it does it could craft any header it wants, this should just narrow the window on the kinds of crap it can sent and close the issue.

crates/types/src/primary/header.rs
```diff
@@ -76,6 +76,11 @@ impl Header {
             return Err(HeaderError::InvalidEpoch { theirs: self.epoch, ours: committee.epoch() });
         }
 
+        // Ensure we don't have too many parents.
+        if self.parents.len() > committee.size() {
+            return Err(HeaderError::TooManyParents(self.parents.len(), committee.size()));
+        }
```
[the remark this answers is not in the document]
grantkee: Ah, yes these are different "parents". I still think it would be worth checking that the umber of parents is reasonable since we are planning to change voting weight in #451 to be 1:1, so if `num_parents < quorum => reject`

crates/consensus/primary/src/network/handler.rs
grantkee: > I am not sure what would catch this?

I think the current flow on main is:
- node restarts outside of GC window
- node receives header that is `TooNew` and hangs on `consensus_bus.wait_for_execution`
- eventually receives consensus gossip
- restarts in `InactiveCVV` mode

The new flow would return an error to the header proposer, which would be logged in the `Certifier`

> ...it could craft any header it wants, this should just narrow the window on the kinds of crap it can sent and close the issue

The issue as I understand it is that I a malicious validator could send a header for this round but include bogus parents which would be permanently appended to `RequestHandler::requested_parents`, but this tracks by round so I believe your patch does work.

grantkee (dismissed):
Thanks for clarifying. I understand the issue now and agree this patch works. Well done

crates/consensus/primary/src/network/handler.rs
```diff
@@ -216,11 +223,42 @@ where
...
         parents: Vec<Certificate>,
     ) -> PrimaryNetworkResult<PrimaryResponse> {
+        {
+            // Check for validator equivocation early and reject if so
+            let mut auth_last_vote = self.auth_last_vote.lock();
+            let (last_epoch, last_round, last_digest) =
+                *auth_last_vote.get(&header.author).unwrap_or(&(0, 0, HeaderDigest::default()));
+            if last_digest != header.digest()  // can ask for the same vote again
+                && (header.epoch() < last_epoch
+                    || (last_epoch == header.epoch() && last_round >= header.round()))
```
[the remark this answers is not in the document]
sstanfield: Agree, that is what last_digest != header.digest() is for or you think that won't work?  At least one test agrees with it.  Also, agree on 399, I can just roll that into this PR since it is almost done with this one.

Issue #443, opened 2025-09-30
Non-persistent header tracking leads to transaction loss on node restart


leopashov:
**Cantina finding**: #1177
**Severity**: Medium

---

**Summary**  
The Proposer fails to persist proposed, but failed to be committed headers and digests across node restarts, potentially resulting in transaction loss when a node restarts during specific scenarios.

---

**Finding Description**  
The proposer maintains two key data structures to track proposed headers:

1. An in-memory `proposed_headers` `BTreeMap` that keeps track of all headers that have been proposed but not yet committed to the DAG. When a header is proposed in `propose_next_header()`, it's sent to other peers for voting. After that in `handle_proposal_result()` it is added to this in-memory map:

```rust
fn handle_proposal_result(&mut self, result: ProposerResult<Header>) -> ProposerResult<()> {
    // receive result from oneshot channel
    let header = result?;

    // track latest header
    self.opt_latest_header = Some(header.clone());
    // Reset advance flag.
    self.advance_round = false;
    // reschedule intervals
    self.min_delay_interval.reset_after(self.calc_min_delay());
    self.max_delay_interval.reset_after(self.calc_max_delay());
    // track header so proposer can repropose the digests and system messages
    // if this header fails to be committed for some reason
    self.proposed_headers.insert(header.round(), header);

    Ok(())
}
```

As stated in the comment above the insert, this is performed in case the header fails to be committed (rejected by the committee, the header for the next round is proposed, etc.). So a rescheduling of the digests of this header should be performed.

2. A database table `last_proposed` that stores the last proposed header:

```rust
async fn store_and_send_header(
    header: &Header,
    proposer_store: DB,
    consensus_bus: &ConsensusBus,
    reason: &str,
) -> ProposerResult<()> {
    // Store the last header.
    proposer_store
        .write_last_proposed(header)
        .map_err(|e| ProposerError::StoreError(e.to_string()))?;

    // Send the new header to the `Certifier` that will broadcast and certify it.
    let result =
        consensus_bus.headers().send(header.clone()).await.map_err(|e| Box::new(e).into());
    let num_digests = header.payload().len();
    consensus_bus
        .primary_metrics()
        .node_metrics
        .num_of_batch_digests_in_header
        .with_label_values(&[reason])
        .observe(num_digests as f64);

    result
}
```

In `crates/storage/src/stores/proposer_store.rs`:

```rust
pub const LAST_PROPOSAL_KEY: ProposerKey = 0;

[...]

Issue #440, opened 2025-09-30
Transactions Are Removed From The MemPool Prematurely Which Results In Transactions Getting Lost After Epoch Transitions


leopashov:
**Cantina finding:** #967
**Severity**: Medium

---

**Summary**  
Transactions are prematurely removed from validator transaction pools when batches reach quorum but before they are actually included in a canonical block, leading to potential transaction loss when a new epoch is started.

---

**Finding Description**  
The Telcoin network uses a unique transaction pool architecture where each validator maintains its own isolated transaction pool (`WorkerTxPool`). Unlike traditional blockchains where transactions are gossiped across the network, Telcoin validators only share transactions when they are sealed in batches that have reached consensus quorum.

The critical issue occurs in the `BatchBuilder::poll()` method. When a batch successfully reaches quorum (2f+1 workers acknowledge it), the batch info itself is saved in the Batches Table:

```rust
self.store.insert::<Batches>(&digest, &batch)
```

and the digest is sent to the Primary/proposer where it is saved in the digest vectors of the proposer:

```rust
self.digests.push_back(digest); //i: add the digest to the queue of the proposer
```

The system immediately removes the transactions from the validator's transaction pool via `update_canonical_state()`:

```rust
// update pool to remove mined transactions
this.pool.update_canonical_state(
    &this.last_canonical_update,
    base_fee_per_gas,
    None,
    mined_transactions,
    vec![],
);
```

However, this removal happens before the transactions are actually included in a canonical block. The transactions are only considered "mined" when they appear in the canonical chain through the engine's execution process, which happens asynchronously after the batch reaches quorum.

> **Note**  
> The `update_canonical_state()` method in `WorkerTxPool` which is called directly after a batch was validated by the other workers removes transactions from the pool immediately, not when they are actually committed to the canonical chain.

This creates a critical vulnerability during epoch transitions. On epoch transition every validator launches a new proposer which starts with an empty digest queue and an empty certificates table (which is cleaned at the end of the epoch). This means that if a transaction exists only in a single validator's pool, the epoch ends and the transaction in the batch is not yet included in blockchain — it is lost and will never be executed no matter if it was still "a digest" in the proposer or already included in a header which did not get added to the blockchain.

---

**Issue Scenario**  
- **Transaction Isolation**: A transaction T1 is submitted to Validator A's pool but not propagated to other validators  
- **Batch Creation**: Validator A includes T1 in a batch that reaches quorum  
[...]

Issue #398, opened 2025-09-29
Transaction can be dropped in the gossip network


leopashov:
**Cantina Finding:** # 718  
**Severity:** Low  

---

#### Finding Description

A transaction can be gossiped via worker p2p network's Txn message. It leverage a simple mechanism to include transaction to local mempool, if it's a validator node:

```rust
fn submit_txn_if_mine(&self, tx_bytes: &[u8], committee_size: u64, committee_slot: u64) {
    if let Ok(tx) = bytes_to_txn(tx_bytes) {
        if let Some(tx_pool) = &self.tx_pool {
            let mut bytes = [0_u8; 8];
            let hash = tx.hash();
            bytes.copy_from_slice(&hash[0..8]);
            if (u64::from_ne_bytes(bytes) % committee_size) == committee_slot {
                // add transaction to mempool
            }
        }
    }
}
```

The first 8 bytes of the transaction hash are used to compute a modulo with the current committee size. If the result matches the local validator's slot in the committee, the transaction is added to the mempool. Here it used u64::from_ne_bytes() to convert bytes to the number for calculation. This is not portable across systems. A transaction could be lost if validator nodes run on systems with different native endianness.

For example, a committee has size of 10, the first 8 bytes of transaction hash is \x01\x00\x00\x00\x00\x00\x00\x00. Some nodes interpret the value as little-endian, producing 1, thus index 1, while others interpret it as big-endian, resulting in 72057594037927936 and selecting index 6 after modulo. If the validator at index 1 uses a big-endian system and the validator at index 6 uses little-endian, the transaction won't be accepted by either.

---

#### Recommendation

Use explicit byte-order conversions to ensure consistent behavior across all environments. For example:

```rust
if (u64::from_le_bytes(bytes) % committee_size) == committee_slot {
    // consistent behavior across all platforms
}
```

PR #474, opened 2025-10-21
Cache local batches until they reach consensus.

sstanfield:
This will allow transactions that would be orphaned to be re-injected into the mempool.
closes https://github.com/Telcoin-Association/telcoin-network/issues/440
closes https://github.com/Telcoin-Association/telcoin-network/issues/443

Also
closes https://github.com/Telcoin-Association/telcoin-network/issues/398
this is a one character fix (use little vs native endian on a byte encode) so just putting in with this.

Also fixes an issue with incorrectly saving an epoch record we had to fetch (caused rare epoch boundary test failures).

### DIFF
--- a/crates/node/src/engine/mod.rs
+++ b/crates/node/src/engine/mod.rs
@@ -175,7 +175,7 @@ impl ExecutionNode {
     }
 
     /// Return an owned instance of all the worker's transaction pools.
-    pub async fn get_worker_transaction_pools(&self) -> Vec<WorkerTxPool> {
+    pub async fn get_all_worker_transaction_pools(&self) -> Vec<WorkerTxPool> {
         let guard = self.internal.read().await;
         guard.get_worker_transaction_pools()
     }
--- a/crates/node/src/manager.rs
+++ b/crates/node/src/manager.rs
@@ -432,7 +432,7 @@ where
     }
 
     /// Collect any batches that never got into consensus (at epoch change or node restart) and
-    /// Re-introduce them into the mempool for incusion in future batches.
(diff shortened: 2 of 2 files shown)


crates/node/src/engine/mod.rs
```diff
@@ -174,6 +174,12 @@ impl ExecutionNode {
         guard.get_worker_transaction_pool(worker_id)
     }
 
+    /// Return an owned instance of all the worker's transaction pools.
+    pub async fn get_worker_transaction_pools(&self) -> Vec<WorkerTxPool> {
```
grantkee: nit: the method names comes across like a worker can have multiple pools, but this is all worker pools

Maybe add "all" to the name? `get_all_...`

crates/node/src/manager.rs
```diff
@@ -426,6 +431,48 @@ where
...
+        engine: ExecutionNode,
+        worker: Worker<DB, QuorumWaiter>,
+    ) -> eyre::Result<()> {
+        let mut orphan_batches: Vec<(BlockHash, Batch)> =
+            self.consensus_db.iter::<NodeBatchesCache>().collect();
+        if !orphan_batches.is_empty() {
+            self.consensus_db.clear_table::<NodeBatchesCache>()?;
+            let consensus_bus = self.consensus_bus.clone();
+            epoch_task_manager.spawn_task("Orphaned Batches", async move {
+                info!(target: "epoch-manager", "Re-introducing orpahned batchs {} transactions", orphan_batches.len());
```
grantkee: sp: "orpahned" -> `orphaned`

PR #479, opened 2025-10-28
More stable restarts.

sstanfield:
Result of focusing on occasional restart test failures.
- Added a new restart test to lag the restarted node- this exercises different paths that had bugs.
- Removed the DB (store) from Bullshark.  Unrelated but it noted it was dead code so took out.
- Expanded on comments as reviewing code.
- Change the code on startup (CVV) for executing missed consensus.  Tighten up this process and not do some unnecessary steps.
- Remove the concept of "early finalize", now that all consensus output is certified before it can go anywhere should not need to delay finalize anymore.  This has a side effect of closing some potential races.
- Use certificates to also detect we are behind on consensus and move to inactive if an active CVV.
- Certificates are "ordered" by BlsPublicKey not AuthorityId now.  This allows us to use committees that are only a sequence of bls keys (i.e. future committees from an epoch record).
- Clear the Votes DB table on startup.  This is mostly a testing concession but also closes some races that could hang a full network restart.
- Had to take extra care when run_epoch joins- it was possible for a shutdown to be triggered before we join so need to make sure to handle that.

Together these changes seem to improve restart reliability.

### DIFF
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -79,7 +79,7 @@ where
     }
 
     /// Detect if we are too far behind the given epoch, round and switch to CVV inactive if we are
-    /// active. This will put us in a "catch up" mode until we have caught up enogh to rejoin
+    /// active. This will put us in a "catch up" mode until we have caught up enough to rejoin
     /// consensus.
     async fn behind_consensus(&self, epoch: Epoch, round: Round, number: Option<u64>) -> bool {
         // Last consensus block we have executed, us this to determine if we are
@@ -102,7 +102,7 @@ where
         // epoch_behind will work in that case.
         let outside_gc_window = epoch == exec_epoch && (exec_round + gc_depth) < round;
         // are we on older epoch?
-        // note, we need to make wure we are not at the epoch boundary otherwise
+        // note, we need to make sure we are not at the epoch boundary otherwise
         // we can get false positives
         let epoch_behind = if let Some(number) = number {
             epoch > exec_epoch && exec_number + 1 < number
@@ -166,7 +166,7 @@ where
                     match unverified_cert.verify_cert(&committee) {
                         Ok(cert) => {
                             if self.behind_consensus(epoch, cert.header().round, None).await {
-                                warn!(target: "primary", "certificate indicates we are behind, go to catchup mode!", );
+                                warn!(target: "primary", "certificate indicates we are behind, go to catchup mode!");
                                 return Ok(());
                             }
                             self.state_sync.process_peer_certificate(cert).await?;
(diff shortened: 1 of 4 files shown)


crates/consensus/primary/src/network/handler.rs
```diff
@@ -78,6 +78,66 @@ where
         }
     }
 
+    /// Detect if we are too far behind the given epoch, round and switch to CVV inactive if we are
+    /// active. This will put us in a "catch up" mode until we have caught up enogh to rejoin
```
grantkee: sp: `enogh` -> `enough`

crates/consensus/primary/src/network/handler.rs
```diff
@@ -78,6 +78,66 @@ where
...
+        // the current DAG. Trying to ride the GC window exactly can lead to subtle races
+        // (allow some time to get going).
+        let gc_depth = self.consensus_config.parameters().gc_depth.saturating_sub(10);
+        let active_cvv = self.consensus_bus.node_mode().borrow().is_active_cvv();
+        // is our round outside the GC window
+        // Will be false when not the same epoch (can't compare rounds) but
+        // epoch_behind will work in that case.
+        let outside_gc_window = epoch == exec_epoch && (exec_round + gc_depth) < round;
+        // are we on older epoch?
+        // note, we need to make wure we are not at the epoch boundary otherwise
```
grantkee: sp: `wure` -> `sure`

crates/consensus/primary/src/network/handler.rs
```diff
@@ -98,7 +158,28 @@ where
                 );
                 // process certificate
                 let unverified_cert = cert.validate_received().map_err(CertManagerError::from)?;
-                self.state_sync.process_peer_certificate(unverified_cert).await?;
+
+                let epoch = unverified_cert.header().epoch;
```
grantkee: Good catch. The only way we can receive gossip consensus if we are behind is if a validator is in a future epoch (network layer filters on this node's current committee). Tricky situation, might be worth thinking about a cleaner architecture, but this is a good solution for now

crates/consensus/primary/src/network/handler.rs
```diff
@@ -78,6 +78,66 @@ where
         }
     }
 
+    /// Detect if we are too far behind the given epoch, round and switch to CVV inactive if we are
+    /// active. This will put us in a "catch up" mode until we have caught up enogh to rejoin
+    /// consensus.
+    async fn behind_consensus(&self, epoch: Epoch, round: Round, number: Option<u64>) -> bool {
```
grantkee: Good abstraction for consensus header and cert gossip

crates/types/src/primary/certificate.rs
```diff
@@ -236,11 +250,13 @@ impl Certificate {
...
     /// Will clear the state first and revalidate even if it appears to be valid.
-    pub fn verify_cert(mut self, committee: &Committee) -> CertificateResult<Certificate> {
+    pub fn verify_cert(mut self, committee: &[BlsPublicKey]) -> CertificateResult<Certificate> {
+        self = self.validate_received()?;
         self = self.validate_received()?;
         let (weight, pks) = self.signed_by(committee);
 
-        let threshold = committee.quorum_threshold();
+        // All validator have a vote weight of 1.
+        let threshold = 2 * committee.len() as u64 / 3 + 1;
```
grantkee: This should be a method on committee, not rewritten everywhere (issue #451)

PR #485, opened 2025-11-07
Healthcheck

grantkee:
- basic, lightweight healtcheck endpoint
- CLI support


sstanfield (dismissed):
This looks useful.  It might be possible to use prometheus for this or an RPC endpoint?  But this is simple enough and focused so probably better.  Had a couple of comments, you can disregard but I think should be addressed.

PR #492, opened 2025-11-18
Update faucet address

robriks:


grantkee (changes requested):
The `genesis.yaml` is created at deployment time and stored here. We only want to update the account after a new deployment.

Instead, please update the address in `testnet_accounts.yaml` since that is used to create genesis on deployments

PR #498, opened 2025-11-21
cantina-174: add quadratic scaling for gas limit penalty

### LINKED ISSUE #424: DOS through high gas limit declarations
**Cantina Finding:** # 174   
**Severity:** High  

---

#### Summary

Malicious actors can exploit gas limit declarations to perform a denial-of-service attack on the Telcoin network by submitting transactions with high gas limits but minimal actual usage, reducing network throughput by 99%+.

---

#### Finding Description

The batch builder acknowledges a gas limit exploitation vulnerability:

```rust
/// NOTE: it's possible to under utilize resources if users submit transactions
/// with very high gas limits. It's impossible to know the amount of gas a transaction
/// will use without executing it, and the worker does not execute transactions.
```

The `BatchBuilder::add_pool_transactions` function includes transactions based solely on declared gas limits:

```rust
// Check if the transaction will exceed the gas limit
if total_possible_gas + pool_tx.gas_limit() > gas_limit {
    //@99Crits-Note: only checks declared limit, not actual usage
    warn!(
        target = "batch_builder::add_pool_transactions",
        "Stopping early as we've reached the gas limit: {} > {}",
        total_possible_gas + pool_tx.gas_limit(),
        gas_limit
    );
    break;
}
...
total_possible_gas += pool_tx.gas_limit();
```

This enables attackers to:

- Submit transactions with `gas_limit = 30,000,000` (near batch maximum)  
- Execute simple operations using only 21,000 gas (0.07% utilization)  
- Block entire batches from including legitimate transactions  
- Pay standard fees based on actual usage, not reservation  

Without penalties for gas underutilization or dynamic fee adjustments, attackers can sustain this attack indefinitely at minimal cost.

---

#### Recommendation

Simulate the gas consumption first before putting them into batches, instead of just taking cumulative `gas_limit`. In case a batch fails, due to too much gas consumed, move the transactions back into the pool. Maybe add a safety margin so there is some room for increase.

grantkee:
- add quadratically scaling penalty for txs that don't use at least 10% of their gas limit
     - economically discourage excessive gas limits which DOS batch space
- update tests and utils

closes #424

### DIFF
--- a/crates/tn-reth/src/evm/utils.rs
+++ b/crates/tn-reth/src/evm/utils.rs
@@ -8,7 +8,7 @@ const MIN_GAS_LIMIT_THRESHOLD: u64 = 210_000;
 /// Precision for calculating usage ratio with 10^18.
 // 10^9 precision (1 billion) - sufficient for 0.001% granularity
 static PRECISION: LazyLock<u128> = LazyLock::new(|| 10_u128.pow(9));
-/// Usage ratio threshold below which penalties apply (10%)
+/// Usage ratio threshold below which penalties apply (used to calc 10%)
 // 10% threshold = 10^8 with 10^9 precision
 static THRESHOLD: LazyLock<u128> = LazyLock::new(|| 10_u128.pow(8));
 /// THRESHOLD squared
@@ -80,8 +80,8 @@ pub fn calculate_gas_penalty(gas_limit: u64, gas_used: u64) -> u64 {
         "assessing penalty"
     );
 
-    // clamp to unused gas (safety fallback) and convert to u64
-    penalty.min(unused_gas) as u64
+    // unused_gas is the fallback if penalty overflows (u128 cast from u64)
+    penalty.try_into().unwrap_or(unused_gas as u64)
 }
 
 #[cfg(test)]


crates/tn-reth/src/evm/utils.rs
```diff
@@ -0,0 +1,195 @@
+//! Utility functions used by EVM.
+
+use std::sync::LazyLock;
+use tracing::debug;
+
+/// Minimum gas limit threshold (10x minimum transaction cost)
+const MIN_GAS_LIMIT_THRESHOLD: u64 = 210_000;
+/// Precision for calculating usage ratio with 10^18.
```
sstanfield: Out of date comment?

crates/tn-reth/src/evm/utils.rs
```diff
@@ -0,0 +1,195 @@
...
+        ?gas_limit,
+        ?gas_used,
+        ?penalty,
+        ?unused_gas,
+        final=penalty.min(unused_gas) as u64,
+        "assessing penalty"
+    );
+
+    // clamp to unused gas (safety fallback) and convert to u64
+    penalty.min(unused_gas) as u64
```
sstanfield: Maybe use try_into() for the cast with an unwrap_or_else to unused_gas.  I don't think this number should wind up bigger than a u64 but would be slightly safer in case something were to change.

sstanfield (approved):
Looks good.  I did not see any problems with the math and your tests look like good coverage.  Couple of minor comments you can decide on.

crates/tn-reth/src/evm/utils.rs
grantkee: Good idea

PR #503, opened 2025-12-11
Fix some races when restarting epochs to improve restart tests.

sstanfield:
Fix some issues exposed by restart tests.
Closes https://github.com/Telcoin-Association/telcoin-network/issues/505

### DIFF
--- a/crates/node/src/manager.rs
+++ b/crates/node/src/manager.rs
@@ -69,10 +69,15 @@ const ENGINE_TASK_MANAGER: &str = "Engine Task Manager";
 /// The worker's base task manager name. This is used by `fn worker_task_manager_name(id)`.
 pub(super) const WORKER_TASK_BASE: &str = "Worker Task";
 
+/// Modes for an epoch.
 #[derive(Debug, Copy, Clone)]
 enum RunEpochMode {
+    /// This is the initial epoch when the system starts, will need to get established.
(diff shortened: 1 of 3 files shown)


crates/node/src/manager.rs
```diff
@@ -480,6 +477,122 @@ where
         Ok(())
     }
 
+    /// Turn a CommittedSubDag with consensus header info into ConsensusOutput.
+    /// It will retrieve any missing Batches so the ConsensusOutput will be ready
+    /// to execute.
+    /// Note, an error here is BAD and will most likely cause node shutdown (clean).  Do
+    /// not provide a bogus sub dag...
+    async fn fetch_local_batches(
```
grantkee: This seems like duplicate logic that exists when `Subscriber::run` receives a committed subdag on `consensus_bus.sequence()` channel

The subscriber downloads these batches if they are missing from local storage, but I agree they should be in local storage by this point. I think it's fine to have this in this PR because we are planning to refactor the startup/state-sync flow but wanted to point this out as something we want to think through from an architectural perspective moving forward. Ideally, everything goes through one flow with logic gates. We'll sync on this more during the refactor

grantkee (approved):
Few nits,  no show stoppers. Thanks for digging into restart tests improvements

Issue #506, opened 2025-12-22
Database Configuration: Static File Storage


grantkee:
# Implement Static File Storage for Finalized Data

## Summary

This proposal introduces a static file storage layer for finalized, immutable data (consensus headers, finalized batches) to complement our existing libmdbx-based storage. This approach is used by reth and addresses fundamental limitations of B+tree databases when storing append-only historical data that grows into the terabytes.

### Architecture Overview

libmdbx is a memory-mapped B+tree key-value store that provides ACID transactions without using a Write-Ahead Log (WAL). It achieves this through **copy-on-write (CoW)** semantics:

```
                    ┌─────────────────────────────────────┐
                    │         libmdbx Architecture         │
                    ├─────────────────────────────────────┤
                    │                                      │
  Write Transaction │   1. Copy modified pages            │
                    │   2. Build new tree structure       │
                    │   3. Atomic root pointer swap       │
                    │   4. Old pages → freelist           │
                    │                                      │
                    │  ┌─────┐    ┌─────┐                 │
                    │  │Root │───▶│Root'│ (atomic swap)   │
                    │  └──┬──┘    └──┬──┘                 │
                    │     │          │                     │
                    │  ┌──┴──┐    ┌──┴──┐                 │
                    │  │Page │    │Page'│ (CoW copy)      │
                    │  └─────┘    └─────┘                 │
                    │     │                                │
                    │     ▼                                │
                    │  Freelist (for reuse)               │
                    │                                      │
                    └─────────────────────────────────────┘
```

### Key Characteristics

| Property | libmdbx Behavior |
|----------|------------------|
| **Durability** | Instant crash recovery (no WAL replay needed) |
| **Concurrency** | MVCC - readers never block writers |
| **Write Amplification** | Low (~1-2x) compared to LSM trees |
| **Memory Model** | Memory-mapped files, OS manages page cache |
| **Transactions** | Full ACID with serializable isolation |

### Why We Chose libmdbx

1. **MVCC for Consistent State Queries**: Allows "stitching together" complex state from multiple tables without locks
2. **Instant Crash Recovery**: No WAL replay on restart—critical for validator uptime
3. **Predictable Read Latency**: Single B+tree traversal vs. LSM's multi-level lookups
4. **Battle-Tested in Ethereum**: Used by reth and Erigon at mainnet scale

## The Problem: libmdbx Limitations at Scale

[...]

PR #510, opened 2026-01-07
Add support for bulk storage of consensus data by epoch.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/506


crates/storage/src/archive/pack.rs
```diff
@@ -0,0 +1,594 @@
...
+    version: u16,     // Holds the version number
+    uid: u64,         // Unique ID generated on creation
+    appnum: u64,      // Application defined constant
+}
+
+impl DataHeader {
+    pub(crate) fn new() -> Self {
+        let mut hasher = FxHasher::default();
+        let now = time::SystemTime::now();
+        let now_millis = now
```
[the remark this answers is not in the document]
sstanfield: Good call on using the epoch for the UID, adds a little extra sanity check and is still deterministic.

crates/storage/src/archive/pack_iter.rs
```diff
@@ -0,0 +1,106 @@
...
+            }
+            return Err(FetchError::IO(err));
+        }
+        crc32_hasher.update(&val_size_buf);
+        let val_size = u32::from_le_bytes(val_size_buf);
+        buffer.resize(val_size as usize, 0);
+        file.read_exact(buffer)?;
+        crc32_hasher.update(buffer);
+        let calc_crc32 = crc32_hasher.finalize();
+        let val = decode::<V>(&buffer[..]);
```
grantkee: Should this happen after the crc check?

crates/storage/src/archive/digest_index/index.rs
```diff
@@ -0,0 +1,823 @@
...
+        if !self.read_only && !self.synced {
+            if !std::thread::panicking() {
+                tracing::warn!("HdxIndex dropped with unsynced data - caller should call sync()");
+            }
+            if let Err(e) = self.write_header() {
+                if !std::thread::panicking() {
+                    tracing::error!("HdxIndex: failed to write header on drop: {e}");
+                }
+            }
+            if let Err(e) = self.save_bucket_cache() {
```
grantkee: nit: this is an effort of last resort, but I think it's _slightly_ more correct to write buckets before header so the header acts as the commit point

Issue #508, opened 2025-12-29
rm: `monitored_future`


grantkee:
# Problem

- not being used for metrics
- spawns with tokio, not task manager
- `monitored_future` and `monitored_scope` and `monitored_channel` should be removed

## Solution

Remove consensus metrics. These metrics are not used and inconsistent with our new approach to manage tasks. We can make a better version once metrics are prioritized.

PR #518, opened 2026-01-20
refactor: remove monitored_future, monitored_scope, and related code

TanguyDeTaxis:
## Summary                                                                
  - Remove `monitored_scope` function and `MonitoredScopeGuard`                                                                                                                                                        
  - Remove `MonitoredFutureExt` trait and `MonitoredScopeFuture`                                                                                                                                                       
  - Remove `Metrics` struct and related static initialization                                                                                                                                                            
  - Remove unused dependencies (scopeguard, once_cell) 
  
  These monitoring utilities were not being used for actual metrics collection and spawned tasks via tokio directly instead of the task manager.                                                                                                            
  Refs #508

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2401,10 +2401,8 @@ version = "0.1.0"
 dependencies = [
  "axum 0.8.6",
  "futures",
- "once_cell",
  "parking_lot",
  "prometheus",
- "scopeguard",
  "tn-types",
  "tokio",
  "tracing",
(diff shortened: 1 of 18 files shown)


sstanfield (approved):
This looks good to me, thanks.  One of the goals is to remove all the "old" metrics code to be replaced with something new leveraging open tracing.  I don't think the ticket exists yet but if you are interested in continuing ripping out old metrics I can add an issue.

I was going to suggest getting rid of the "async move" blocks so that errors could make it to the task manager but alas it seems like the task manager will need some tuning (it uses it's own async moves and erases the errors as well).  The will be a larger refactor though and this PR does what it should.  It still works from a high level but sometimes it looks like a critical task ended with Ok and not the error that killed it which is not ideal for debugging.

Note we will need also do an "attest" (Grant or I have to do that) before final merge but I don't think that will be an issue (it does check format and nightly clippy- see etc/test-and-attest.sh for the steps).  Basically runs all tests including the slow ones and especially the faucet tests which need credentials.  We are going to work on fixing that in the future though (the need for the creds) to get rid of the permissioned testing.

PR #519, opened 2026-01-20
fix JSON serialization for tn_latestHeader RPC (#375)

### LINKED ISSUE #375: The tn_latestHeader RPC fails.
The ConsensusHeader type will not serialize to JSON currently causing an internal error on this call.

Should disable this RPC until fixed.  Note the tn namespace RPCs are not fleshed out and fully usable yet but can be used for testing.

We will probably need a "shadow" structure to generate the JSON from.  The reputation structure seems to be the culprit since it contains a Map with non-String keys.  We should probably also make this JSON data complete (include certificates) while doing this so that multiple calls will not be needed to collect consensus output when using RPC.

TanguyDeTaxis:
## Summary      
  - Implement custom Serialize/Deserialize for AuthorityIdentifier using bs58 for JSON and raw bytes for bincode                                                                                                           
  - Add FromStr impl with proper error handling                                                                                                                                                                            
  - Remove the TODO comment in rpc_ext.rs                                                                                                                                                                                  
                                                                                                                                                                                                                           
  This fixes the root cause: any HashMap<AuthorityIdentifier, T> now serializes correctly to JSON automatically.                                                                                                           
                                                                                                                                                                                                                           
  Closes #375

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12149,6 +12149,7 @@ dependencies = [
  "roaring",
  "secp256k1 0.31.0",
  "serde",
+ "serde_json",
  "serde_repr",
  "serde_with 2.3.3",
  "serde_yaml",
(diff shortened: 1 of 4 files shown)


sstanfield (approved):
This looks really good, thanks.  I like you included the FromStr implementation.

Want to give @grantkee a chance to review your PRs and we will attest them soon for merge.

PR #520, opened 2026-01-22
fix: use sequential sub_dag_index instead of nonce for LeaderSchedule

### LINKED ISSUE #472: bug: LeaderSchedule based on Certificate `nonce` instead of subdag index
# Problem

The current leader schedule is based on the leader's `nonce`, but Bullshark logic assumes this is the sequential subdag index (monotonically increasing per commit).

## Solution

Track this value in state

The subdag index was changed in [this PR](https://github.com/Telcoin-Association/telcoin-network/pull/146/files#diff-e73b6663a668d107ef2473d46af3ee7444e26ccd3fa1a841850654e88346fd99R249)

TanguyDeTaxis:
This commit fixes issue #472 where the LeaderSchedule was using the Certificate's nonce (epoch << 32 | round) instead of a sequential subdag index for reputation score calculations.

Changes:
- Add `sub_dag_index` field to CommittedSubDag for sequential tracking
- Replace `leader.nonce()` with `state.next_sub_dag_index()` in Bullshark
- Remove the `/2` division hack that was compensating for round-based indices
- Persist sub_dag_index across epoch boundaries by recovering it from the global latest subdag before filtering by current epoch
- Add test `sub_dag_index_persists_across_epochs` to verify the fix

The sequential index ensures that:
- ReputationScores are recalculated at consistent intervals (every 60 commits)
- final_of_schedule is triggered correctly regardless of epoch changes
- LeaderSwapTable updates happen predictably

### DIFF
--- a/crates/consensus/primary/src/consensus/bullshark.rs
+++ b/crates/consensus/primary/src/consensus/bullshark.rs
@@ -70,10 +70,8 @@ impl Bullshark {
     ) -> ReputationScores {
         // we reset the scores for every schedule change window, or initialise when it's the first
         // sub dag we are going to create.
-        // sub_dag_index is based on epoch and round so / 2 so this check works on commit rounds
-        // (every other one).
         let mut reputation_score =
-            if (sub_dag_index / 2).is_multiple_of(self.num_sub_dags_per_schedule as u64) {
+            if sub_dag_index.is_multiple_of(self.num_sub_dags_per_schedule as u64) {
                 ReputationScores::new(&self.committee)
             } else if let Some(last) = state.last_committed_sub_dag.as_ref() {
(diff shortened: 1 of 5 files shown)


sstanfield (commented):
I want to go over this one with Grant.  That ticket was actually something we were not in total agreement on but had shelved (sorry, should have had a note in the issue).  I removed the original sub dag index and just used the nonce since it was one less data field in the consensus "chain" and I think the nonce is good enough.  Anyway now that you have implemented this that may resolve it...  I also may be wrong about the nonce being good enough but it does grow per round it just will have large gaps per epoch.

grantkee:
Closing this in favor of a different approach. Thank you @TanguyDeTaxis for looking into this.

PR #533, opened 2026-02-04
feat(tn-reth): parallelize transaction recovery with rayon

TanguyDeTaxis:
## Summary
- Parallelize transaction recovery using rayon for improved block processing performance
- Uses rayon's parallel iterators to recover transaction senders concurrently instead of sequentially

### DIFF
--- a/crates/tn-reth/src/lib.rs
+++ b/crates/tn-reth/src/lib.rs
@@ -229,10 +229,6 @@ pub struct RethConfig(NodeConfig<RethChainSpec>);
 
 const DEFAULT_UNUSED_ADDR: IpAddr = IpAddr::V4(Ipv4Addr::UNSPECIFIED);
 
-/// Minimum number of transactions to justify parallel ECDSA recovery via rayon.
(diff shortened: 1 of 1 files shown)


sstanfield (commented):
I think this is good but if we could get rid of the collects it would be better (less allocations).  Is it possible to have recovered_txs be the actual iterator vs a vector and then use the iterator directly avoiding the allocation?  The borrow checker will likely give you grief with the if (can probably work around it but not 100% sure).  It might be worth taking the hit if less that 5 txns and just use rayon for all cases if so.  You loose performance on the par_iter but avoid an allocation to pick some up.  Plus with that few transactions a slight performance hit should be noise vs a lot of transactions where this will shine.

crates/tn-reth/src/lib.rs
```diff
@@ -1802,6 +1813,57 @@ mod tests {
         Ok(())
     }
 
+    #[test]
```
grantkee: I like this test

PR #534, opened 2026-02-04
feat(consensus): make channel send no-op when no receiver subscribed

### LINKED ISSUE #457: Double check consensus bus to make sure we don't have channels not being read from.
If we have any MPSC channels that are not read (for instance because a task is not started in the current mode) then they could "fill up" and block calling code. 

See consensus/primary/src/primary.rs around line 103 for a task that is started just to deal with this.

We want something more robust then this, we use a wrapper around MPSC channels already so maybe we can adapt it to turn send() into a no-op when there is no receiver for instance.  The goal would be to keep sending code "clean" and not have to start otherwise useless tasks.

TanguyDeTaxis:
## Summary

- Adds an `AtomicBool` (`subscribed`) flag to `MeteredMpscChannel` and `QueChannel` so that `send()` / `try_send()` become no-ops when no receiver is subscribed, preventing MPSC channels from filling up and blocking calling code.
- Removes the workaround task in `primary.rs` that existed solely to drain the `parents` channel for non-CVV nodes, replacing it with `drop(consensus_bus.parents().subscribe())`.
- Extends test coverage with property-based tests on BFT invariants, DAG state tests, recovery persistence tests, equivocation detection, and serialization round-trips.

Closes #457

### DIFF
--- a/crates/node/src/manager.rs
+++ b/crates/node/src/manager.rs
@@ -417,6 +417,96 @@ fn spawn_epoch_vote_collector<DB: TNDatabase>(
     });
 }
 
+#[cfg(test)]
+mod epoch_vote_collector_tests {
+    use super::*;
+    use rand::{rngs::StdRng, SeedableRng as _};
+    use tn_network_libp2p::types::{MessageId, NetworkCommand};
+    use tn_primary::network::{PrimaryRequest, PrimaryResponse};
+    use tn_storage::mem_db::MemDatabase;
+    use tn_types::{BlsKeypair, TnSender as _};
+
+    /// Happy path: committee of 4, node signs + receives 3 peer votes → cert stored.
+    #[tokio::test]
+    async fn test_collector_reaches_quorum_and_stores_cert() {
+        let mut rng = StdRng::from_os_rng();
+        let kp1 = BlsKeypair::generate(&mut rng);
+        let kp2 = BlsKeypair::generate(&mut rng);
+        let kp3 = BlsKeypair::generate(&mut rng);
+        let kp4 = BlsKeypair::generate(&mut rng);
+        let pk1 = *kp1.public();
+        let pk2 = *kp2.public();
+        let pk3 = *kp3.public();
+        let pk4 = *kp4.public();
+
+        // Node is kp1
+        let key_config = KeyConfig::new_with_testing_key(kp1);
+
+        // Committee of 4: super_quorum = (4*2)/3 + 1 = 3
+        let epoch_rec = EpochRecord {
+            epoch: 0,
+            committee: vec![pk1, pk2, pk3, pk4],
+            next_committee: vec![pk1, pk2, pk3, pk4],
+            ..Default::default()
+        };
+        let epoch_hash = epoch_rec.digest();
+
+        let consensus_bus = ConsensusBus::new();
+        let db = MemDatabase::default();
+
+        // Mock network: drain commands and reply to Publish
+        let (net_tx, mut net_rx) =
+            tokio::sync::mpsc::channel::<NetworkCommand<PrimaryRequest, PrimaryResponse>>(100);
+        let primary_network = PrimaryNetworkHandle::new_for_test(net_tx);
+        tokio::spawn(async move {
+            while let Some(cmd) = net_rx.recv().await {
+                if let NetworkCommand::Publish { reply, .. } = cmd {
+                    let _ = reply.send(Ok(MessageId::new(b"test")));
+                }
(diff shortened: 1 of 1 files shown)


TanguyDeTaxis:
Thanks for the review @sstanfield ! I fully agree with both points — here's what I've done:                                                                                                                                             
                                                                                                                                                                                                                                    
  **Channels now start unsubscribed by default** (`subscribed=false`). This eliminates the subscribe/drop dance entirely — the `else` branch in `primary.rs` is simply removed since unsubscribed channels are already no-ops. No   
  more workaround needed.                                                                                                                                                                                                           
                                                                                                                                                                                                                                    
  As you suspected, switching to `false` by default did break a lot of tests — they were implicitly relying on channels being subscribed from the start. Tracking those down was a bit tedious but it flushed out every place that
  was sending without an explicit subscribe, which is exactly what we want.

  To make this safe going forward, I also went further on the API side:

  - **`subscribe()` removed from `TnSender` trait** — it's now an inherent method on `MeteredMpscChannel` and `QueChannel`, so you can't accidentally subscribe through a generic `&impl TnSender<T>`. This gives us compile-time
  enforcement.
  - **Explicit `subscribe_*()` methods on `ConsensusBus`** — all 13 channels have a dedicated subscribe method (e.g. `subscribe_parents()`, `subscribe_new_certificates()`). Every call site now goes through these, making it easy
  to audit which components subscribe to what.
  - All subscriptions happen in the synchronous `spawn()` context, before `spawn_critical_task`, so there's no race between subscribing and sending.

  This should catch exactly the kind of cases you mentioned — channels filling up in observer mode are now fully handled by the default no-op behavior, without relying on anyone remembering the subscribe/drop pattern.

  Hope I understood correctly what you had in mind — let me know if you'd like any adjustments.

sstanfield:
> I like this approach a lot! The epoch tests are failing at the epoch boundary because the `QueChannel` is dropping then no-oping
> 
> This is an edge case because `EpochVote` might come in as gossip after the epoch boundary hits https://github.com/TanguyDeTaxis/telcoin-network/blob/457-consensus-bus-channels/crates/consensus/primary/src/network/handler.rs#L263
> 
> I'm still thinking through this, but wanted to share immediately. My initial thoughts are:
> 
> * dedicated epoch votes channel?
> * arg to ignore the drop?
> * better idea? @sstanfield

@grantkee  I see what you are getting at.  If we get gossip while we are between epochs the flag is false so it gets lost when it used to just get parked in the receiver until next subscribe...  I think there are two solutions, either we subscribe once in run_epochs (note the 's') and then pass the receiver around so it is only subscribed once.  It will have a two deep call stack so that is probably Ok but is drifting a bit from the quick find purpose of the bus.  Or we have to allow the flag to be true even when not subscribed in some cases.  I don't love the special case but it would be contained to the channel and consensus bus so could live with it.  @TanguyDeTaxis If you are implementing your call I think if Grant is cool with that.  Maybe there is another option I missed?

TanguyDeTaxis:
@sstanfield @grantkee Went ahead and implemented both option 2 and option 3 together since they complement each other nicely.

**Signature verification moved to handler** — BLS signature and committee membership checks now happen in the network handler before sending on the channel. This removes the `oneshot::Sender` entirely, so `new_epoch_votes` is now `QueChannel<EpochVote>` instead of `QueChannel<(EpochVote, oneshot::Sender<...>)>`. This also plugs the potential DOS vector Steven identified — a malicious or catching-up node sending epoch votes while the collector isn't running can no longer orphan handler threads waiting on a oneshot that never gets read.

**Epoch vote collector is now a node-lifetime actor** — `spawn_epoch_vote_collector` subscribes once at startup and runs on `node_task_manager`. It waits for `EpochRecord` updates via a watch channel, then drains and aggregates buffered votes. The receiver never drops across epoch boundaries, so no votes are lost during the gap. The node's own vote is signed and added automatically when the epoch record arrives.

Added an integration test (`test_collector_reaches_quorum_and_stores_cert`) that spins up a 4-validator committee, buffers 3 peer votes, pushes an EpochRecord, and verifies the resulting certificate is stored in DB with valid aggregated signatures.

Ready for another look when you get a chance.

grantkee:
Thanks for the updates @TanguyDeTaxis ! They LGTM so far. I'm still evaluating the `spawn_epoch_vote_collector` actor logic and will follow up soon

sstanfield (approved):
Looks really good.  Approving but you should probably address getting epoch recs by number not hash in the handler.  My note on the test does not really matter.

PR #535, opened 2026-02-04
feat(engine): skip EVM execution for empty non-epoch-closing output

TanguyDeTaxis:
## Summary

When consensus output contains no batches and the epoch is not closing, EVM execution is now skipped entirely. The current canonical header is returned unchanged, avoiding unnecessary block allocation, state trie writes, and database persistence for empty rounds.

The leader count is still incremented before the early return so validator rewards tracking remains correct.

When `close_epoch` is true, an empty block is still produced as before to anchor the epoch transition.

## Changes

- **`crates/engine/src/payload_builder.rs`**: Add early return when `batches.is_empty() && !output.close_epoch`, after leader count increment
- **`crates/engine/tests/it/main.rs`**: Add three tests covering the new behavior:
  - `test_empty_output_skips_execution` — no blocks produced for empty non-epoch-closing output
  - `test_empty_output_with_close_epoch_still_executes` — empty block still created when `close_epoch = true`
  - `test_empty_output_increments_leader_count` — leader count incremented even when execution is skipped

### DIFF
--- a/crates/consensus/primary/src/consensus/tests/bullshark_tests.rs
+++ b/crates/consensus/primary/src/consensus/tests/bullshark_tests.rs
@@ -488,7 +488,7 @@ async fn dead_node() {
     let dummy_parent = SealedHeader::new(ExecHeader::default(), B256::default());
     cb.recent_blocks()
         .send_modify(|blocks| blocks.push_latest(0, B256::default(), Some(dummy_parent)));
-    let mut rx_output = cb.sequence().subscribe();
+    let mut rx_output = cb.subscribe_sequence();
     let task_manager = TaskManager::default();
     Consensus::spawn(config, &cb, bullshark, &task_manager);
     let cb_clone = cb.clone();
@@ -621,7 +621,7 @@ async fn not_enough_support() {
     let dummy_parent = SealedHeader::new(ExecHeader::default(), B256::default());
     cb.recent_blocks()
         .send_modify(|blocks| blocks.push_latest(0, B256::default(), Some(dummy_parent)));
-    let mut rx_output = cb.sequence().subscribe();
+    let mut rx_output = cb.subscribe_sequence();
     let task_manager = TaskManager::default();
     Consensus::spawn(config, &cb, bullshark, &task_manager);
     let cb_clone = cb.clone();
@@ -720,7 +720,7 @@ async fn missing_leader() {
     let dummy_parent = SealedHeader::new(ExecHeader::default(), B256::default());
     cb.recent_blocks()
         .send_modify(|blocks| blocks.push_latest(0, B256::default(), Some(dummy_parent)));
-    let mut rx_output = cb.sequence().subscribe();
+    let mut rx_output = cb.subscribe_sequence();
     let task_manager = TaskManager::default();
     Consensus::spawn(config, &cb, bullshark, &task_manager);
     let cb_clone = cb.clone();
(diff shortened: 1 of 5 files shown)


sstanfield (dismissed):
This seems right to me.  I think this means that a leader will get no rewards (makes since since nothing happened) but might encourage leaders to always include some txn for the rewards (?) @grantkee .  Anyway, not filling execution up with empty blocks is worth while I think.  We may need to do something for consensus when the network is idle as well but this another issue to discuss.

crates/engine/src/payload_builder.rs
```diff
@@ -35,14 +38,22 @@ pub fn execute_consensus_output(
     let _guard = span.enter();
     debug!(target: "engine", ?output, "executing output");
 
+    // Skip execution entirely when output contains no batches AND the epoch is not closing.
+    // The leader count has already been incremented above for rewards tracking.
+    if batches.is_empty() && !output.close_epoch {
```
grantkee: nit: `batches.is_empty()` is also checked on L62

I think it would be cleaner to move the `if !output.close_epoch {...}` inside L62 block as part of that conditional branch

```rust
// L62 of current branch
if batches.is_empty() {
     if !output.close_epoch {
         // log + channel update + return
     }
    // execute single block with not transactions to close the epoch
}
```

grantkee: Please update L63 comment to mention closing epoch

grantkee:
> This seems right to me. I think this means that a leader will get no rewards (makes since since nothing happened) but might encourage leaders to always include some txn for the rewards (?) @grantkee . Anyway, not filling execution up with empty blocks is worth while I think. We may need to do something for consensus when the network is idle as well but this another issue to discuss.

Good point. The `gas_accumulator.rewards_counter()` is still updated in the correct spot [here](https://github.com/TanguyDeTaxis/telcoin-network/blob/531-skip-empty-execution/crates/engine/src/payload_builder.rs#L25). Leaders still receive rewards to incentivize advancing consensus rounds. However, the [pub fn catchup_accumulator](https://github.com/TanguyDeTaxis/telcoin-network/blob/531-skip-empty-execution/crates/node/src/manager.rs#L186) will need to be updated as well to ensure rewards are incremented correctly.

There is a test for this, but every output contains batches [here](https://github.com/TanguyDeTaxis/telcoin-network/blob/531-skip-empty-execution/crates/node/tests/it/main.rs#L53-L54). @TanguyDeTaxis please update this test as well to ensure test coverage for commits without batches of transactions still increments leader rewards

TanguyDeTaxis:
Thanks for the detailed review @grantkee!                               
                                                                                                       
  The second commit (cac59c06) already implements the mpsc channel approach you described — replacing  
  CanonStateNotificationStream with mpsc::Sender<(Round, B256, Option<SealedHeader>)> passed to
  execute_consensus_output. The channel always sends the leader round and consensus hash (even on
  skip), so RecentBlocks is updated for every output and wait_for_consensus_execution no longer hangs.

  The third commit (b2d49a19) addresses the remaining items:

  1. e2e restart tests — network_advancing() was waiting for block_number to increase, which never
  happens during idle periods with skip-empty-execution. Changed it to check RPC responsiveness
  instead; actual block production is already verified by send_and_confirm().
  2. Engine update delivery — replaced let _ = try_send() with try_send() + error propagation so
  failures aren't silently ignored.
  3. Node IT test for rewards with empty outputs — added test_catchup_accumulator_with_empty_outputs
  that mixes real and empty consensus outputs, then verifies:
    - gas_accumulator counts ALL leaders (including skipped rounds)
    - catchup_accumulator only counts leaders for rounds that produced blocks
    - Gas values match between both accumulators
    - The difference in leader counts equals exactly the number of empty outputs injected

  Re: catchup_accumulator — it currently only replays from blocks in the DB, so it won't see leaders
  from skipped rounds. This means after a restart, leader counts will be lower for the current epoch.
  This seems acceptable since gas values (which determine actual rewards) are consistent. Let me know
  if you think this needs further handling.

crates/test-utils/src/consensus.rs
grantkee: These seem useful for generating test data for `catchup_accumulator`, but isn't called anywhere yet. I really like that the caller can pass `empty_rounds`. 

I'm in favor of replacing the existing `create_signed_certificates_for_rounds` with this approach. There are only 5 tests that use this function for now, so shouldn't be too much to refactor. The only test that needs to pass `empty_rounds` is the new catchup accumulator test. This approach will also better support fuzz testing

TanguyDeTaxis:
@grantkee Rebase done! Thanks

PR #536, opened 2026-02-04
refactor(consensus): replace watch send() with send_replace() and remove stored receivers

### LINKED ISSUE #529: Get rid of the stored watch receivers on the consensus bus.
There are several instances of watch::Receivers on the consensus bus.  They are stored so that send() will work but if all instances of send() are replaced with send_replace() on the Sender half then the call becomes infallible and updates the watch even with no receivers.  This will be cleaner that hanging onto Receivers that are never used.

TanguyDeTaxis:
## Summary

- Replace all `watch::Sender::send()` calls with `send_replace()` across the consensus layer, making channel updates infallible
- Remove 7 stored `watch::Receiver` fields from `ConsensusBusAppInner` that were only kept to prevent `send()` from failing
- Clean up now-unnecessary error handling paths (e.g. `SubscriberError::ClosedChannel` for watch channels)

Closes #529

### DIFF
--- a/crates/consensus/executor/src/subscriber.rs
+++ b/crates/consensus/executor/src/subscriber.rs
@@ -141,7 +141,7 @@ impl<DB: Database> Subscriber<DB> {
         // We aren't doing consensus now but still need to update these watches before
         // we send the consensus output.
         self.consensus_bus.committed_round_updates().send_replace(last_round);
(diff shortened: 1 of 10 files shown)


TanguyDeTaxis:
@grantkee Resolved. conflicts about line 161-162 on consensus_bus.rs. Thanks for review

grantkee (approved):

PR #539, opened 2026-02-04
feat(rpc): add tn_syncing, tn_epochInfo, tn_currentCommittee methods

### LINKED ISSUE #530: feat: `tn_` RPC Namespace Methods
# Add `tn_*` RPC Methods for Network Sync Status and Epoch Info

## Summary

Add new RPC methods to the `tn` namespace to support validator operations and network monitoring. These methods will help validators determine when their node is synced and ready for activation, as well as provide epoch and committee information.

## Motivation

New validator nodes need a way to determine if their node is fully synced before calling `activate()` on the `ConsensusRegistry` contract. Currently, there's no RPC method to check if a node's execution state is caught up with the current consensus epoch. Validators activating before their node is synced risk being slashed or missing consensus participation.

Additionally, operators and tooling need easy access to current epoch information and committee membership details through RPC.

NOTE: `eth_syncing` always returns `false` because peers do not exchange block execution info (requires overriding default `reth` RPC behavior).

### 1. `tn_syncing`

Returns sync status for the consensus chain. Follows the same pattern as `eth_syncing`: returns `false` when synced, or an object with sync progress when syncing.

**Request:**

```json
{
  "jsonrpc": "2.0",
  "method": "tn_syncing",
  "params": [],
  "id": 1
}
```

**Response (when synced):**

```json
{
  "jsonrpc": "2.0",
  "result": false,
  "id": 1
}
```

**Response (when syncing):**

```json
{
  "jsonrpc": "2.0",
  "result": {
    "startingBlock": 0,
    "currentBlock": 5000,
    "highestBlock": 10000
  },
  "id": 1
}
```

- `false` - Node is synced (execution has caught up to the current consensus epoch)
- `{ startingBlock, currentBlock, highestBlock }` - Node is syncing; values refer to consensus chain block numbers (not execution chain)

**Sync Definition:**

- A node is considered "synced" when its execution state is within the current consensus epoch
- Example: If the committee is committing subdags for epoch 10, a node executed into epoch 10 returns `false` (not syncing), while a node still executing epoch 9 output returns the sync status object
- Nodes within the current epoch but outside the garbage collection window should still return `false` (considered synced)
- This applies to all node types (validators, observers, etc.)

**Implementation Notes:**

- The `ConsensusBus` provides access to `NodeMode` via `node_mode()` which tracks `CvvActive`, `CvvInactive`, and `Observer` states
- However, `NodeMode` alone is insufficient - we need to compare the local execution epoch against the network's current consensus epoch
- The RPC layer will need to:
  1. Get the highest known consensus block from the network (via gossip)
  2. Get the local node's current executed consensus block
  3. Track the starting block from when sync began
  4. Return `false` if synced, or the sync status object otherwise
- The `EngineToPrimary` trait may need extension to expose this information to the RPC layer
- `startingBlock`: The consensus block number when the node started syncing
- `currentBlock`: The consensus block number the node has executed up to
- `highestBlock`: The highest consensus block number known from the network

### 2. `tn_epochInfo`

Returns the current `EpochInfo` for the committee's epoch (not the local node's view if it differs during sync).

**Request:**

```json
{
  "jsonrpc": "2.0",
  "method": "tn_epochInfo",
  "params": [],
  "id": 1
}
```

**Response:**

```json
{
  "jsonrpc": "2.0",
  "result": {
    "epoch": 10,
    "committee": ["0x...", "0x...", "0x..."],
    "nextCommittee": ["0x...", "0x...", "0x..."],
    "parentHash": "0x...",
    "parentState": {
      "number": 12345,
      "hash": "0x..."
    },
    "parentConsensus": "0x..."
  },
  "id": 1
}
```

**Implementation Notes:**

- Should return the `EpochRecord` for the current committee's epoch
- The existing `tn_epochRecord` method requires an epoch number parameter; this method returns the _current_ epoch without requiring the caller to know the epoch number
- Useful for tooling that needs to discover the current epoch state

### 3. `tn_currentCommittee`

Returns validator information for the current committee members.

**Request:**

```json
{
  "jsonrpc": "2.0",
  "method": "tn_currentCommittee",
  "params": [],
  "id": 1
}
```

**Response:**

```json
{
  "jsonrpc": "2.0",
  "result": {
    "epoch": 10,
    "validators": [
      {
        "blsPublicKey": "0x...",
        "address": "0x..."
      },
      ...
    ]
  },
  "id": 1
}
```

**Implementation Notes:**

- Returns the current committee's validator information
- Should include both BLS public keys (from consensus layer `EpochRecord.committee`) and Ethereum addresses (if available)
- The `EpochRecord` stores `Vec<BlsPublicKey>` for the committee

### RPC Layer Changes

1. **Extend `TelcoinNetworkRpcExtApi` trait** in `crates/execution/tn-rpc/src/rpc_ext.rs`:

   ```rust
   #[method(name = "syncing")]
   async fn syncing(&self) -> TelcoinNetworkRpcResult<SyncStatus>;

   #[method(name = "epochInfo")]
   async fn epoch_info(&self) -> TelcoinNetworkRpcResult<EpochRecord>;

   #[method(name = "currentCommittee")]
   async fn current_committee(&self) -> TelcoinNetworkRpcResult<CommitteeInfo>;
   ```

   Where `SyncStatus` is an enum that serializes to either `false` or the sync progress object:

   ```rust
   pub enum SyncStatus {
       Synced,  // serializes to `false`
       Syncing { starting_block: u64, current_block: u64, highest_block: u64 },
   }
   ```

2. **Extend `EngineToPrimary` trait** in `crates/execution/tn-rpc/src/lib.rs` to expose:
   - Current consensus epoch (from network/committee perspective)
   - Local execution epoch
   - Current committee information

3. **Update `EngineToPrimaryRpc` implementation** in `crates/node/src/lib.rs` to provide the required data from `ConsensusBus`

### Data Flow

```
ConsensusBus (consensus layer)
    │
    ├── last_consensus_header() -> current consensus epoch
    ├── recent_blocks() -> local execution state
    └── node_mode() -> sync status hint
    │
    ▼
EngineToPrimaryRpc (implements EngineToPrimary trait)
    │
    ▼
TelcoinNetworkRpcExt (RPC implementation)
    │
    ▼
JSON-RPC Response
```

## Use Cases

1. **Validator Activation**: Before calling `ConsensusRegistry.activate()`, validators query `tn_syncing` to ensure their node is caught up
2. **Monitoring**: Operators use `tn_syncing` to monitor node health and sync status
3. **Tooling**: Block explorers and dashboards use `tn_epochInfo` and `tn_currentCommittee` to display network state

## Open Questions

1. How should we determine the "network's current epoch" for a syncing node that may not be connected to the full committee?
   - The node receives consensus headers via gossip; the latest verified header indicates the network epoch
2. Should `tn_currentCommittee` include additional validator metadata (stake amount, status, etc.)?
   - Start with minimal info (BLS key, address); can extend later if needed
3. Where should `startingBlock` be tracked? This value needs to persist across the sync process.
   - Reference the most-recent `EpochRecord`
   - ex)
     - Committe is on epoch 10
     - syncing node is half-way through epoch 8
     - `startingBlock` = EpochRecord for epoch 7

## Related

- Existing `tn_*` methods: `tn_latestHeader`, `tn_genesis`, `tn_epochRecord`, `tn_epochRecordByHash`
- `ConsensusRegistry.activate()` - the contract method validators call after syncing
- `NodeMode` enum: `CvvActive`, `CvvInactive`, `Observer`
- any other RPC method that is considered useful for node operators, TN users, etc.

TanguyDeTaxis:
## Summary

- Add `tn_syncing`, `tn_epochInfo`, and `tn_currentCommittee` RPC methods to the `tn` namespace
- Add `current_committee` watch channel to `ConsensusBus` so RPC can access committee info
- Fix `tn_syncing` to check **both** epoch and execution progress before declaring a node synced (addresses [review feedback](https://github.com/Telcoin-Association/telcoin-network/issues/530#issuecomment-3848398575))
- Add `currentEpoch` / `highestEpoch` fields to the sync progress response for better validator visibility
- Add comprehensive tests: 7 serialization tests + 11 sync logic tests
- Fix JSON-RPC error codes to use the standard server error range (`-32000` to `-32099`)

### `tn_syncing` behavior

Returns `false` when synced, or a progress object when syncing:

```json
{
  "startingBlock": 100,
  "currentBlock": 5000,
  "highestBlock": 10000,
  "currentEpoch": 3,
  "highestEpoch": 5
}
```

A node is considered synced when:
1. It is an active CVV (`CvvActive`), **or**
2. Its local epoch >= network epoch **and** execution has caught up to the latest known consensus block

Previously, condition 2 only checked the epoch — a node could report as synced while execution was still lagging behind consensus. This is now fixed.

### Files changed

| File | Change |
|------|--------|
| `crates/consensus/primary/src/consensus_bus.rs` | Add `current_committee` watch channel |
| `crates/node/src/manager.rs` | Populate committee channel on epoch creation |
| `crates/execution/tn-rpc/src/rpc_ext.rs` | New RPC types, methods, and serialization tests |
| `crates/execution/tn-rpc/src/lib.rs` | Extend `EngineToPrimary` trait with new methods |
| `crates/execution/tn-rpc/src/error.rs` | Fix JSON-RPC error codes |
| `crates/node/src/lib.rs` | Implement sync logic fix + add 11 logic tests |
| `crates/execution/faucet/tests/it/faucet.rs` | Update mock to implement new trait methods |


crates/node/src/lib.rs
```diff
@@ -100,6 +104,86 @@ impl<DB: Database> EngineToPrimary for EngineToPrimaryRpc<DB> {
             (None, None) => None,
         }
     }
+
+    fn sync_status(&self) -> SyncStatus {
+        let mode = *self.consensus_bus.node_mode().borrow();
+        if matches!(mode, NodeMode::CvvActive) {
+            return SyncStatus::Synced;
+        }
```
grantkee: Nice and clean

crates/node/src/lib.rs
```diff
@@ -100,6 +104,86 @@ impl<DB: Database> EngineToPrimary for EngineToPrimaryRpc<DB> {
...
+            highest_block,
+            current_epoch,
+            highest_epoch,
+        })
+    }
+
+    fn current_epoch_info(&self) -> Option<EpochInfo> {
+        let (_, record) = self.db.last_record::<EpochRecords>()?;
+        Some(record.into())
+    }
```
grantkee: Epoch records are only generated _after_ the epoch concludes, so this logic needs to be adjusted

crates/node/src/lib.rs
```diff
@@ -50,11 +51,14 @@ pub struct EngineToPrimaryRpc<DB> {
     consensus_bus: ConsensusBus,
     /// Consensus DB
     db: DB,
+    /// The block number when sync started, captured on first syncing status check.
+    /// Derived from the last EpochRecord's parent_state as specified in issue #530.
+    sync_starting_block: OnceLock<u64>,
```
grantkee: I made `RethEnv` thread-safe and cheap to clone in #549 so we can use it in the RPC struct

```rust
pub struct EngineToPrimaryRpc<DB> {
    /// Access to the execution database.
    reth_env: RethEnv,
    //...snip
}
```

crates/execution/tn-rpc/src/rpc_ext.rs
```diff
@@ -6,8 +6,126 @@ use crate::{
...
+    pub starting_block: u64,
+    /// The current block number the node has reached.
+    pub current_block: u64,
+    /// The highest known block number.
+    pub highest_block: u64,
+    /// The node's current local epoch.
+    pub current_epoch: u64,
+    /// The network's current epoch.
+    pub highest_epoch: u64,
+}
```
grantkee: The `starting_block`, `current_block`, `highest_block` should all mention and correlate to consensus blocks (aka - `ConsensusHeader`). Please add another field to this struct that includes `execution_block_height` or something along those lines that uses `RethEnv::canonical_tip()` number

grantkee (changes requested):
@TanguyDeTaxis Thank you for taking this on. It's going in the right direction, but I think there is more to consider. I'm leaving feedback for you, but I am still reviewing/thinking about some of the implementation details. Your work reflects the issue #530 however, I'm realizing there are still outstanding design decisions I need to figure out. I will follow up soon with more comments

crates/node/src/lib.rs
TanguyDeTaxis: Updated in e98b524d. network_epoch is now derived from the highest downloaded epoch record + 1, with explicit handling for the epoch-0 dummy record.

crates/node/src/lib.rs
```diff
@@ -100,6 +104,86 @@ impl<DB: Database> EngineToPrimary for EngineToPrimaryRpc<DB> {
...
+
+        // Epoch-based sync decision: compare network epoch (from committee) vs local epoch (from
+        // DB). This follows the issue #530 spec: "A node is considered synced when its
+        // execution state is within the current consensus epoch."
+        let network_epoch =
+            self.consensus_bus.current_committee().borrow().as_ref().map(|c| c.epoch());
+
+        let local_epoch =
+            self.db.last_record::<EpochRecords>().map(|(epoch, _)| epoch).unwrap_or(0);
+
```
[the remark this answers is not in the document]
TanguyDeTaxis: Implemented as suggested: local_epoch now comes from reth_env.epoch_state_from_canonical_tip(), with committee epoch as fallback only if that lookup fails.

crates/node/src/manager.rs
```diff
@@ -1454,6 +1454,9 @@ where
         // load committee
         committee.load();
 
+        // update the consensus bus with the current committee for RPC access
+        self.consensus_bus.current_committee().send_replace(Some(committee.clone()));
+
```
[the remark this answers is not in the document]
TanguyDeTaxis: Good call. I now publish current_committee after persisting a new epoch record (manager.rs around save_epoch_record path), and also publish at committee load so RPC reflects the live epoch promptly.

PR #541, opened 2026-02-07
fix(types): handle broadcast channel lag instead of silently dropping observers

TanguyDeTaxis:
## Summary

Fix a critical bug where broadcast channel lag silently killed observers. The `TnReceiver::recv()` impl for `broadcast::Receiver` converted `RecvError::Lagged` to `None` via `.ok()`, causing observer recv loops to exit entirely — observers would stop following consensus with no log or error.

The fix wraps `recv()` in a loop that catches `Lagged`, logs the number of lost messages, and continues receiving from the oldest buffered message.

This is likely related to the observer stalling issues reported on devnet where council members and personal observer nodes experience problems after running for a while (ref: #457, PR #534 discussion).

## Test plan

- [x] `test_broadcast_lag_does_not_return_none` — verifies recv() returns Some after lag
- [x] `test_broadcast_closed_returns_none` — verifies recv() returns None only when channel is closed
- [x] `test_broadcast_try_recv_lagged` — verifies try_recv() correctly reports Lagged
- [x] All `tn-types` tests pass
- [x] `cargo clippy --workspace --all-features` — zero warnings
- [x] `cargo +nightly fmt --check` — clean

### DIFF
--- a/crates/types/src/sync.rs
+++ b/crates/types/src/sync.rs
@@ -215,7 +215,26 @@ impl<T: Send + 'static> TnSender<T> for mpsc::Sender<T> {
 
 impl<T: Send + Clone> TnReceiver<T> for broadcast::Receiver<T> {
     async fn recv(&mut self) -> Option<T> {
-        broadcast::Receiver::recv(self).await.ok()
+        loop {
+            match broadcast::Receiver::recv(self).await {
+                Ok(value) => return Some(value),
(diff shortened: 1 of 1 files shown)


sstanfield (approved):
@TanguyDeTaxis Did you see this in the wild?  Either way this is a good catch and is strictly better.

Issue #254, opened 2025-04-14
Network Metrics


grantkee:
Add support for peer manager and network metrics in `crates/tn-network-libp2p`

Issue #233, opened 2025-02-26
Add a metric to track how far behind fully validated blocks are from the chain tip.


sstanfield:
This delta should typically be one block.  Add a metric to track this for visibility.

This is the delta between the last block from consensus output and the latest execution block that is included in a certificate.

PR #542, opened 2026-02-07
feat: observer diagnostics, sync-aware healthcheck and e2e tests

TanguyDeTaxis:
## Summary

Adds observability tooling for observer nodes — addressing the lack of diagnostics that makes it hard to understand why observers stall on devnet (ref: #233, #254, PR #499 discussion).

### Sync-aware healthcheck

Extends the existing healthcheck endpoint to report sync status in the response body:
- `"OK"` when fully synced
- `"behind N"` when N consensus blocks behind
- `"syncing"` when no consensus data available yet

HTTP 200 always for load balancer compatibility. Inspired by Solana's \`GET /health\` pattern.

### Observer diagnostics logging

Structured tracing with \`target: "tn::observer"\` for OpenTelemetry export:
- Sync distance on each gossip update (Issue #233)
- Catchup distance when catching up consensus blocks
- Node mode at subscriber and state-sync startup
- Progress logging every 100 blocks during follow_consensus
- Warning when consensus header channel closes
- Fork detection logging (digest mismatch, execution fork)

### Peer metrics (Issue #254)

Heartbeat logging with connected count, connected-or-dialing count, and banned count.

### E2E tests

- `test_observer_late_join_catchup` — observer starts after validators produce blocks, verifies catchup
- `test_observer_reconnect_after_pause` — SIGSTOP/SIGCONT simulates partition, verifies recovery

## Test plan

- [x] 3 healthcheck unit tests pass
- [x] All `tn-executor`, `tn-network-libp2p`, `state-sync` tests pass (107 total)
- [x] `cargo clippy --workspace --all-features` — zero warnings
- [x] `cargo +nightly fmt --check` — clean

Depends on #541

### DIFF
--- a/crates/consensus/executor/src/subscriber.rs
+++ b/crates/consensus/executor/src/subscriber.rs
@@ -61,6 +61,7 @@ pub fn spawn_subscriber<DB: Database>(
     let committee = config.committee().clone();
     let client = config.local_network().clone();
     let mode = consensus_bus.current_node_mode();
+    info!(target: "tn::observer", node_mode = ?mode, "subscriber starting in mode");
     let subscriber = Subscriber {
         consensus_bus,
         config,
@@ -198,22 +199,23 @@ impl<DB: Database> Subscriber<DB> {
 
             // Periodically log observer progress (every 100 blocks)
             if processed_count.is_multiple_of(100) {
-                let latest_executed = self.consensus_bus.latest_block_num_hash().number;
-                let latest_known = self
+                let latest_processed_consensus = self
                     .consensus_bus
-                    .last_consensus_header()
+                    .recent_blocks()
                     .borrow()
-                    .as_ref()
-                    .map(|h| h.number)
-                    .unwrap_or(0);
-                let sync_distance = latest_known.saturating_sub(latest_executed);
+                    .latest_consensus_block_num_hash()
+                    .number;
+                let (latest_network_consensus, _) =
+                    self.consensus_bus.published_consensus_num_hash();
+                let consensus_sync_distance =
+                    latest_network_consensus.saturating_sub(latest_processed_consensus);
                 info!(
                     target: "tn::observer",
                     processed_count,
                     header_number,
-                    latest_executed,
-                    latest_known,
-                    sync_distance,
+                    latest_processed_consensus,
+                    latest_network_consensus,
+                    consensus_sync_distance,
                     "observer follow progress"
                 );
             }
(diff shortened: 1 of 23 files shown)


sstanfield (commented):
@TanguyDeTaxis Great work debugging the observer issues.  I assume you have been running your own for the required time to see the error?  I have not finished a review but this looks good so far.  I notice you often keep an old trace event and add a new.  You can probably just replace some of the old events if you want.  To date they have mostly been used as logs so as long as the information if still logged nothing should be lost.  If you think it will be valuable to have similar event with different targets that is probably fine.  Open Tracing is still early days, we have not setup jaeger for devnet yet for instance.

TanguyDeTaxis:
Hey @sstanfield ,
Thanks for the review! I answer also your comment on #541 in same comment here : To answer your question - no, I haven't personally run observers long enough to hit this in the wild. I found the bug through code inspection (with AI assistance, not gonna lie) while investigating the observer stalling issues reported on devnet (refs #457, #534).
When analyzing the TnReceiver impl, I spotted the .ok() on RecvError::Lagged - it was clear this would silently kill any observer recv loop since they'd just stop following consensus with zero visibility. Given the devnet reports of observers stalling after running for a while, this seemed like a very likely culprit.
The fix should make observers significantly more resilient to temporary broadcast channel pressure, and the diagnostics in #542 will help us confirm whether lag events were actually occurring in production.
Re: the trace events - good point about potentially consolidating some of them. I kept both in a few places to maintain existing log coverage while adding the new target: "tn::observer" events for structured OpenTelemetry export, but I can clean that up if you think it's redundant. Let me know your preference!

sstanfield (dismissed):
@TanguyDeTaxis Sorry for the delay- I have been down a rabbit hole on the new consensus DB.  I think at this point we sould consolidate traces.  As we approach main net we may adjust that but at this point I don't know that we need both.  I don't have a strong opinion on this though.

Good job on the new restart tests.  The STOP/CONT is interesting and I like it, not sure how often that will come up but the more robust we can make it the better so fine with me.

On a side note, I am on vacation next week so will probably not be responsive until the week after.

crates/consensus/executor/src/subscriber.rs
```diff
@@ -190,9 +200,35 @@ impl<DB: Database> Subscriber<DB> {
...
+                let sync_distance = latest_known.saturating_sub(latest_executed);
+                info!(
+                    target: "tn::observer",
+                    processed_count,
+                    header_number,
+                    latest_executed,
+                    latest_known,
+                    sync_distance,
+                    "observer follow progress"
+                );
```
grantkee: Great idea!

PR #560, opened 2026-02-14
refactor(committee): enforce equal voting power across authorities

### LINKED ISSUE #451: Voting Weight

TanguyDeTaxis:
## Summary
- enforce equal authority voting power in committee logic
- make committee thresholds count-based via authority count (uniform weight model)
- remove weighted voting distribution support from committee test fixture builder
- update committee/property tests and certificate order test to match the equal-weight model

Closes #451

### DIFF
--- a/crates/types/src/committee.rs
+++ b/crates/types/src/committee.rs
@@ -80,21 +80,13 @@ impl Authority {
     /// it via Committee (more specifically can use [CommitteeBuilder]). As some internal properties
     /// of Authority are initialised via the Committee, to ensure that the user will not
     /// accidentally use stale Authority data, should always derive them via the Commitee.
-    fn new(
-        protocol_key: BlsPublicKey,
-        _voting_power: VotingPower,
-        execution_address: Address,
-    ) -> Self {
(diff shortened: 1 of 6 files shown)


crates/types/src/committee.rs
```diff
@@ -82,20 +84,32 @@ impl Authority {
     /// accidentally use stale Authority data, should always derive them via the Commitee.
     fn new(
         protocol_key: BlsPublicKey,
-        voting_power: VotingPower,
+        _voting_power: VotingPower,
```
sstanfield: Any reason to keep this?

sstanfield (approved):
Looks good, don't think we need to keep some of the voting power params/fields though.

PR #562, opened 2026-02-14
Use new pack files as consensus DB.

sstanfield:
This replaces the long lived consensus tables (except the Epoch Records table) with the new pack file implementaion.

### DIFF
--- a/crates/storage/src/consensus.rs
+++ b/crates/storage/src/consensus.rs
@@ -8,6 +8,7 @@ use std::{
     io::{Read, Seek, SeekFrom, Write},
     path::{Path, PathBuf},
     sync::Arc,
+    thread::JoinHandle,
 };
 
 use parking_lot::Mutex;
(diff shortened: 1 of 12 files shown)


crates/storage/src/consensus.rs
```diff
@@ -0,0 +1,775 @@
...
+        let _ = rx.await;
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct ConsensusChain {
+    base_path: PathBuf,
+    current_pack: Option<ConsensusPack>,
+    latest_consensus: LatestConsensus,
+    recent_packs: Arc<Mutex<VecDeque<ConsensusPack>>>,
```
grantkee: Missing doc comments

crates/storage/src/consensus.rs
```diff
@@ -0,0 +1,775 @@
...
+                next_committee: committee.bls_keys().iter().copied().collect(),
+                ..Default::default()
+            },
+            committee,
+        )
+        .await?;
+        Ok(me)
+    }
+
+    pub async fn new_epoch(
```
grantkee: Missing doc comment

crates/storage/src/consensus.rs
```diff
@@ -0,0 +1,775 @@
...
+            let _ = File::create_new(&slot2_path);
+        }
+        let mut slot1 = OpenOptions::new().read(true).write(true).open(&slot1_path)?;
+        let mut slot2 = OpenOptions::new().read(true).write(true).open(&slot2_path)?;
+        let mut buffer32_epoch = [0_u8; 4];
+        let mut buffer32_crc = [0_u8; 4];
+        let mut buffer64 = [0_u8; 8];
+        let _ = slot1.read_exact(&mut buffer32_epoch);
+        let _ = slot1.read_exact(&mut buffer64);
+        let _ = slot1.read_exact(&mut buffer32_crc);
```
grantkee: Why are these errors ignored?

crates/storage/src/consensus.rs
```diff
@@ -0,0 +1,775 @@
...
+    pub async fn save_consensus_output(
+        &mut self,
+        consensus: ConsensusOutput,
+    ) -> Result<(), ConsensusChainError> {
+        if let Some(pack) = &self.current_pack {
+            if consensus.number() > self.latest_consensus.number {
+                self.latest_consensus
+                    .update(consensus.sub_dag().leader_epoch(), consensus.number())
+                    .await;
+                pack.save_consensus_output(consensus).await?;
```
grantkee: nit: swap this order so the memory/disk slot is updated after the file is written?

crates/storage/src/consensus.rs
```diff
@@ -0,0 +1,775 @@
...
+            while let Some(try_epoch) = epoch {
+                if let Ok(pack) = self.get_static(try_epoch).await {
+                    if let Ok(header) = pack.consensus_header_by_number(number).await {
+                        return Ok(Some(header));
+                    }
+                }
+                epoch = if try_epoch > 0 { Some(try_epoch - 1) } else { None };
+            }
+            Ok(None)
+        }
```
grantkee: Is there a DoS concern with this approach? If a malicious request requested non-existent headers, this would open O(epochs) pack files?

crates/storage/src/consensus.rs
```diff
@@ -0,0 +1,775 @@
...
+                }
+            }
+        });
+        Ok(me)
+    }
+
+    async fn update(&mut self, epoch: Epoch, number: u64) {
+        self.epoch = epoch;
+        self.number = number;
+        match self.current_slot {
```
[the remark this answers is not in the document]
sstanfield: That is probably better but would mean grabbing the state lock twice so I don't think it is worth it for the very tiny amount of safety it might provide.  These records are rewritten quickly and we mostly care about the epoch which rarely changes more than the number.

PR #566, opened 2026-02-20
feat: Multi-Worker Support - PR 4: EpochManager Multi-Worker Loop

### LINKED ISSUE #557: feat: Multi-Worker Support — PR 4: `EpochManager` Multi-Worker Loop
## Problem

## Goal

- RPC server (unique port)
- Transaction pool

## Why

## Design Constraints

## Current State

---

## This PR: `EpochManager` Multi-Worker Loop (core change)

This is the core refactor. The `EpochManager` stops hardcoding `worker_id = 0` and instead loops over all workers when spawning components, creating consensus nodes, and managing epochs. With `num_workers = 1` the loop runs once — identical behavior to today.

### Scope

**`crates/node/src/manager.rs` — `GasAccumulator` initialization:**
- Change `GasAccumulator::new(1)` to `GasAccumulator::new(num_workers as usize)`

**`crates/node/src/manager.rs` — `spawn_worker_node_components()`:**
  - Get the worker's network handle from `self.worker_network_handles[worker_id]`
  - Call `engine.initialize_worker_components(worker_id, ...)` (initial epoch only)
  - Call `engine.new_batch_validator(&worker_id, ...)`
  - Call `self.spawn_worker_network_for_epoch(...)` for this worker
  - Create `WorkerNode::new(worker_id, ...)`
- Return `Vec<WorkerNode>` instead of a single `WorkerNode`

**`crates/node/src/manager.rs` — `create_consensus()`:**
- Change return type from `(PrimaryNode, WorkerNode)` to `(PrimaryNode, Vec<WorkerNode>)`
- Worker creation is done inside the loop above

**`crates/node/src/manager.rs` — `run_epoch()`:**
- Receive `Vec<WorkerNode>` from `create_consensus()`
- Loop over each `WorkerNode`:
  - Call `worker_node.new_worker().await?` to get `Worker` instance
  - Call `worker.spawn_batch_builder(...)`
  - Call `engine.start_batch_builder(worker.id(), worker.batches_tx(), ...)`
- Handle orphan batches per-worker

**`crates/node/src/manager.rs` — `catchup_accumulator()`:**
- The existing code already extracts `worker_id` from the difficulty field per-block and calls `gas_accumulator.inc_block(worker_id, ...)` — this already works for N workers
- **Fix:** The hardcoded `gas_accumulator.base_fee(0)` for initial base fee restoration needs to scan backwards through blocks, extract `worker_id` from difficulty, and collect the latest `base_fee` for each worker
- Set `gas_accumulator.base_fee(worker_id).set_base_fee(...)` for each worker

### Backward Compatibility

- With `num_workers = 1`, all loops run once and produce the same single-worker result as today
- `GasAccumulator::new(1)` behavior is preserved when `num_workers = 1`
- `catchup_accumulator` with a single worker only finds `worker_id = 0` blocks — same as current behavior

blocked by #555 (needs `Vec<WorkerNetworkHandle>`) and #556 (needs `Vec<LocalNetwork>`)

TanguyDeTaxis:
## Summary
- Loop EpochManager across configured workers instead of hardcoding worker 0.
- Branch: `stack/557-epoch-manager-multi-worker-loop`

Closes #557

### DIFF
--- a/MULTI_WORKER_STACK_PLAN.md
+++ b/MULTI_WORKER_STACK_PLAN.md
@@ -1,0 +1,72 @@
+# Multi-Worker Stacked Branch Plan
+
+This file tracks the branch/PR stacking strategy for the multi-worker implementation.
+
+## Objective
+
+Implement multi-worker support through stacked PRs with one branch per PR, each branch based on the previous one.
+
+## Branch Naming and Stack Order
+
+1. `stack/554-config-foundation` (base: `main`)
+2. `stack/555-per-worker-swarms` (base: `stack/554-config-foundation`)
+3. `stack/556-per-worker-local-network` (base: `stack/555-per-worker-swarms`)
+4. `stack/557-epoch-manager-multi-worker-loop` (base: `stack/556-per-worker-local-network`)
+5. `stack/558-engine-initialization-multi-worker` (base: `stack/557-epoch-manager-multi-worker-loop`)
+6. `stack/559-multi-worker-tests` (base: `stack/558-engine-initialization-multi-worker`)
+
+## PR Mapping
+
+- Branch `stack/554-config-foundation` -> Issue #554
+- Branch `stack/555-per-worker-swarms` -> Issue #555
+- Branch `stack/556-per-worker-local-network` -> Issue #556
+- Branch `stack/557-epoch-manager-multi-worker-loop` -> Issue #557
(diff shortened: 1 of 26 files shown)


grantkee:
Closing this due to significant upstream changes. Still relevant, but too much drift

PR #567, opened 2026-02-20
feat: Multi-Worker Support - PR 5: Execution Engine Initialization

### LINKED ISSUE #558: feat: Multi-Worker Support — PR 5: Execution Engine Initialization
## Problem

## Goal

- RPC server (unique port)
- Transaction pool

## Why

## Design Constraints

## Current State

---

## This PR: Execution Engine Initialization

Initialize `WorkerComponents` (TX pool, RPC server, worker network) for each worker. The `ExecutionNodeInner` already has a `workers: Vec<WorkerComponents>` — this PR populates it with N entries instead of 1. With `num_workers = 1`, behavior is identical to today.

### Scope

**`crates/node/src/engine/inner.rs` — `initialize_worker_components()`:**
- Called in a loop for each `worker_id` in order (0, 1, ..., N-1)
- Each call creates:
  - Its own `WorkerTxPool` via `reth_env.init_txn_pool()`
  - Its own `WorkerNetwork` with the worker-specific `WorkerNetworkHandle`
  - Its own RPC server (each on a different port, allocated by the reth RPC infrastructure)
- **Faucet:** `self.opt_faucet_args.take()` is called once — only worker 0 gets the faucet extension. For `worker_id > 0`, the faucet branch is skipped (the `Option` is already `None` after the first take)

**`crates/node/src/engine/inner.rs` — `start_batch_builder()`:**
- Already parameterized by `worker_id` — no change needed
- Will be called once per worker from `run_epoch()`

**`crates/node/src/engine/inner.rs` — `new_batch_validator()`:**
- Already parameterized by `worker_id` — no change needed
- Returns a `BatchValidator` bound to that worker's TX pool

**`crates/node/src/engine/inner.rs` — `respawn_worker_network_tasks()`:**
- Currently takes a single `WorkerNetworkHandle` but loops over `self.workers`
- Change signature to accept `&[WorkerNetworkHandle]` and match handles to workers by index
- Each worker's network gets the correct handle on epoch rollover

**RPC port allocation:**
- Each worker's RPC server needs a unique port
- The existing `reth_env.get_rpc_server()` allocates ports — ensure it assigns distinct ports for each call

### Backward Compatibility

- With `num_workers = 1`, the loop runs once, producing one `WorkerComponents` entry — same as today
- Faucet attaches to worker 0 only, which is the existing behavior
- `respawn_worker_network_tasks(&[single_handle])` with one element behaves identically to the current single-handle version

blocked by #557  (`EpochManager` Multi-Worker Loop) — the loop that calls `initialize_worker_components` lives in `spawn_worker_node_components()`

TanguyDeTaxis:
## Summary
- Respawn worker networks and initialize engine components with per-worker handles.
- Branch: `stack/558-engine-initialization-multi-worker`

Closes #558

### DIFF
--- a/MULTI_WORKER_STACK_PLAN.md
+++ b/MULTI_WORKER_STACK_PLAN.md
@@ -1,0 +1,72 @@
+# Multi-Worker Stacked Branch Plan
+
+This file tracks the branch/PR stacking strategy for the multi-worker implementation.
+
+## Objective
+
+Implement multi-worker support through stacked PRs with one branch per PR, each branch based on the previous one.
+
+## Branch Naming and Stack Order
+
+1. `stack/554-config-foundation` (base: `main`)
+2. `stack/555-per-worker-swarms` (base: `stack/554-config-foundation`)
+3. `stack/556-per-worker-local-network` (base: `stack/555-per-worker-swarms`)
+4. `stack/557-epoch-manager-multi-worker-loop` (base: `stack/556-per-worker-local-network`)
+5. `stack/558-engine-initialization-multi-worker` (base: `stack/557-epoch-manager-multi-worker-loop`)
+6. `stack/559-multi-worker-tests` (base: `stack/558-engine-initialization-multi-worker`)
+
+## PR Mapping
+
+- Branch `stack/554-config-foundation` -> Issue #554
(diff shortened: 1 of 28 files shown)


grantkee:
Closing this due to significant upstream changes. Still relevant, but too much drift

Issue #513, opened 2026-01-14
Streaming Protocol for Network Layer


grantkee:
# Problem

The current network layer uses libp2p's `request_response` behavior for all inter-node messaging. This has two significant limitations:

1. **Message size bottleneck**: The `MissingCertificates` response is capped at 1 MiB, requiring multiple round-trips to sync large certificate ranges
2. **Per-message latency overhead**: Every request-response incurs connection setup and framing overhead, even over QUIC

These limitations become problematic during state sync and certificate fetching, where nodes may need to transfer thousands of certificates/batches.

## Solution

Replace the request-response pattern with `libp2p-stream` for all RPC messaging. This provides bidirectional streaming over QUIC with the following benefits:

- **No message size limit**: Bulk data (certificates, batches) can be streamed in chunks without artificial caps
- **Reduced latency**: Single stream opening amortizes connection overhead across multiple items
- **Backpressure support**: Bounded channels provide natural flow control for large transfers

### Design Decisions

- **One request per stream**: Each request opens a new stream and closes when complete. QUIC handles stream creation efficiently, avoiding multiplexing complexity for now.
- **Single frame for small responses**: Responses like `Vote` or `ConsensusHeader` write one serialized frame directly with no chunking overhead.
- **Chunked streaming for bulk responses**: Responses like `RequestedCertificates` stream items in configurable chunks, with a zero-length end marker.
- **Eager stream opening**: Open an outbound stream when a peer connects to validate protocol support immediately.

### Wire Protocol

Requests and single-item responses use a simple length-prefixed frame:
- 4-byte little-endian length prefix
- Snappy-compressed, BCS-encoded payload

Bulk responses use the same frame format repeated, with a 4-byte zero marker indicating end-of-stream.

PR #569, opened 2026-02-24
refactor: network codec

grantkee:
- update wire format for codec to include compressed len to support large data transfers
- make encode/decode functions pub for application to use

precursor for #513


sstanfield (approved):
LGTM

PR #571, opened 2026-02-28
feat: stream worker batches

### LINKED ISSUE #611: Replace RPC-based batch sync with stream-based transfer between workers
## Problem

Workers synchronize batches (collections of transactions) with peers by requesting missing batches over the network. The current implementation uses individual RPC request-response calls for batch transfer, which creates several problems:

First, the RPC approach is constrained by the `max_rpc_message_size` limit. Large batch sets must be artificially capped to fit within the message size, and the handler must juggle both the requestor's reported max size and the node's own limit. This size-juggling logic is brittle and produces partial responses when batches don't fit, forcing the caller to make multiple round trips with rotating peer assignments to collect all missing batches.

Second, the current fan-out strategy (`request_batches`) sends requests to all connected peers simultaneously, distributing digests across them in a round-robin pattern. While this parallelizes fetching, it amplifies network traffic and creates contention. A single large fetch can generate N simultaneous RPC calls (one per peer), with no concurrency control on the responder side. There is no mechanism to prevent a malicious or misbehaving peer from flooding a node with batch requests.

Third, the `RequestBatches` RPC returns `Vec<Batch>` in a single response, meaning the entire batch payload must be serialized, compressed, and held in memory at once. For committed subdags near protocol limits (300 batches at up to 1MB each), this creates a significant memory spike on both sender and receiver.

## Solution

Replace the RPC-based `RequestBatches` / `RequestBatches(Vec<Batch>)` exchange with a two-phase stream protocol. The first phase uses an RPC to negotiate the transfer: the requestor sends a `RequestBatchesStream` message containing the set of missing digests and the epoch, and the responder replies with an `ack` boolean indicating whether it can fulfill the request. If accepted, the second phase opens a raw libp2p stream where batches flow as length-prefixed, snappy-compressed chunks.

The stream protocol should use a chunked wire format -- a 4-byte batch count followed by individually framed batches (uncompressed length, compressed length, compressed data) -- flushed per chunk rather than per batch to optimize throughput. Chunks of up to 200 batches bound per-chunk memory on the sender, and the receiver validates each batch in real-time (digest verification, duplicate detection, count checks) so protocol violations are caught immediately.

Concurrency should be controlled with a global semaphore bounding the total number of concurrent batch stream operations, combined with per-peer rate limiting to prevent a single peer from monopolizing all slots. A periodic cleanup task should prune stale pending requests (peers that negotiate via RPC but never open the stream) to recover leaked semaphore permits. Oversized requests should be truncated rather than rejected, processing as many digests as possible within the cap.

The requestor should try peers sequentially with bounded retries and backoff rather than fan-out to all peers simultaneously. This reduces network amplification while still providing resilience -- if one peer rejects (semaphore full), the next is tried, and the outer `BatchFetcher` loop re-checks local storage between attempts.

grantkee:
## Overview

Replaces the RPC-based worker batch synchronization with a stream-based protocol, adding concurrency control, chunked transfer, and comprehensive security validation.

### Stream-based batch transfer protocol
- Introduced `RequestBatchesStream` request/response replacing `RequestBatches` with a two-phase negotiation: RPC for acceptance, then raw libp2p stream for data transfer
- Added `stream_codec` module implementing the chunked wire format: `[4-byte count][batch...]` per chunk, with each batch framed as `[uncompressed_len][compressed_len][snappy_data]`
- Chunks capped at 200 batches (`BATCH_DIGESTS_READ_CHUNK_SIZE`) with per-chunk flush for backpressure
- Added `SEND_STREAM_TIMEOUT` (200s) on the sender side to prevent slow-reader attacks

### WorkerNetworkHandle refactor
- Extracted `WorkerNetworkHandle` from `network/mod.rs` into its own `network/handle.rs` module
- Replaced `max_rpc_message_size` parameter with `epoch` for protocol-aware batch size validation
- New `request_batches()` method tries peers sequentially with bounded retries (`MAX_BATCH_REQUEST_RETRIES=3`, `BATCH_REQUEST_RETRY_DELAY=500ms`) instead of parallel fan-out
- `read_and_validate_batches_with_timeout()` validates each batch in real-time: digest verification, duplicate detection, count checks, per-batch timeout

### Concurrency control
- Global semaphore (`MAX_CONCURRENT_BATCH_STREAMS=5`) bounds total concurrent batch stream operations
- Per-peer rate limit (`MAX_PENDING_REQUESTS_PER_PEER=2`) prevents a single peer from monopolizing slots
- `PendingBatchStream` struct holds semaphore permit via RAII -- dropped when request completes or times out
- Periodic prune task (every 15s) cleans up stale pending requests older than `PENDING_REQUEST_TIMEOUT` (30s)
- Oversized requests truncated to `MAX_BATCH_DIGESTS_PER_REQUEST=500` rather than rejected

### Inbound stream handling
- `WorkerNetwork` now handles `NetworkEvent::InboundStream` (previously `unimplemented!()`)
- Reads 32-byte correlation digest from stream to match against pending requests
- Spawns task to look up batches from DB and stream them back via `send_batches_over_stream()`

### BatchFetcher simplification
- Renamed `fetch()` to `fetch_for_primary()` with clearer ownership semantics
- `fetch_local()` now takes mutable references to avoid unnecessary cloning
- Removed `RequestBatchesError` enum -- errors unified under `WorkerNetworkError`
- Removed `RequestBatchesNetwork` trait indirection -- uses `WorkerNetworkHandle` directly

### Primary-worker communication refactor
- Extracted `PrimaryReceiverHandler` into new `network/primary.rs` module
- Removed `request_batches_timeout` field (timeout now handled inside stream protocol)
- `fetch_batches()` return type changed from `FetchBatchResponse` wrapper to `HashMap<BlockHash, Batch>`
- Removed `FetchBatchResponse` type and `response.rs` module from `network-types`

### Error handling improvements
- Added new error variants: `TooManyBatches`, `UnexpectedBatch`, `DuplicateBatch`, `StreamClosed`, `UnknownStreamRequest`, `RequestHashMismatch`, `StdIo`, `DBInsert`, `DBCommit`, `DBRead`
- Renamed `Decode` to `Bcs` for clarity
- IO errors mapped to penalties based on error kind (connection resets = Mild, other = Medium)
- Protocol violations (unexpected batches, duplicates, hash mismatches) mapped to Fatal penalty

### Subscriber cleanup
- Updated `fetch_batches_from_peers` to work with new `HashMap` return type (removed `FetchBatchResponse` wrapper)
- Added safety comment documenting max batch digest bounds
- Removed unnecessary `.clone()` on `batch_digests`

### Primary network cleanup
- Removed redundant `consensus_chain` field from `PrimaryNetwork` -- accessed through `RequestHandler` instead
- Added `consensus_chain()` accessor to primary `RequestHandler`

### Network type exports
- Re-exported `Stream` from `tn-network-libp2p`
- Exported new public constants and types from `tn-worker`: `PendingBatchStream`, `MAX_CONCURRENT_BATCH_STREAMS`, `MAX_BATCH_DIGESTS_PER_REQUEST`, `MAX_PENDING_REQUESTS_PER_PEER`, `PENDING_REQUEST_TIMEOUT`

### Test infrastructure
- Replaced `TestRequestBatchesNetwork` mock with simpler `create_test_batches()`, `setup_batch_db()`, `encode_batches_to_stream_bytes()` helpers
- Removed `rand` dependency from worker (was only used in test mock)
- Added `bcs` dev-dependency for wire format verification tests
- Added `snap` and `parking_lot` dependencies to worker

### Test coverage
- Stream codec: roundtrip, empty batch, multiple batches, stream closed, oversized length prefix, wire format structure, multi-chunk roundtrip, partial DB
- Batch validation: correct stream, too many batches, unexpected digest, duplicate batch, stream closed early, oversized digest set truncation, partial stream fulfillment
- BatchFetcher: all-local fetch, many-local fetch, partial local + stream composition
- Concurrency: semaphore exhaustion, permit release on drop, per-peer limits, stale cleanup, exact capacity boundary
- Negotiation: no peers, peer rejection fallback, retry after rejection, digest preservation across fallback, request digest determinism, pending request correlation
- Error penalties: fatal for protocol violations, none for timeouts/stream closed

### Removed code
- Fixed duplicate `process_gossip` error handling (was calling `process_gossip` twice on failure)
- Removed `LOCAL_MIN_REQUEST_SIZE` and `MESSAGE_OVERHEAD` lazy statics (no longer needed without RPC size negotiation)
- Removed old `request_batches_from_peer`, `request_batches_from_all` fan-out logic

### Documentation
- Added `network/README.md` covering architecture, wire format, chunking strategy, retry logic, security measures, and compression

closes #611


crates/consensus/worker/src/network/error.rs
```diff
@@ -63,14 +99,32 @@ impl From<WorkerNetworkError> for Option<Penalty> {
...
-            WorkerNetworkError::InvalidTopic | WorkerNetworkError::Decode(_) => {
-                Some(Penalty::Fatal)
-            }
+            // protocol violations - fatal penalty
+            WorkerNetworkError::InvalidTopic
+            | WorkerNetworkError::Bcs(_)
+            | WorkerNetworkError::TooManyBatches { .. }
+            | WorkerNetworkError::UnexpectedBatch(_)
+            | WorkerNetworkError::DuplicateBatch(_)
+            | WorkerNetworkError::UnknownStreamRequest(_)
```
grantkee, line 120: this might be too aggressive?

crates/consensus/worker/src/network/primary.rs
```diff
@@ -0,0 +1,112 @@
...
+            let (mut batch, digest) = sealed_batch.split();
+            if missing.remove(&digest) {
+                // Set received_at timestamp for remote batch.
+                batch.set_received_at(now());
+                let mut tx = self.store.write_txn().map_err(|e| {
+                    WorkerNetworkError::Internal(format!(
+                        "failed to create batch transaction to commit: {e:?}"
+                    ))
+                })?;
+                tx.insert::<NodeBatchesCache>(&digest, &batch).map_err(|e| {
```
sstanfield, line 83: Since this tx only does a single insert you could just use self.store.insert() directly and avoid getting and committing the TXN manually.  Functionally equivalent but less code.  Alternatively you could move getting the TX and committing the TX outside the for loop to batch all the inserts together which should be more performant.

crates/consensus/worker/src/network/stream_codec.rs
```diff
@@ -0,0 +1,395 @@
...
+    // allocate reusable buffers
+    let mut encode_buffer = Vec::with_capacity(max_size);
+    let mut compressed_buffer = Vec::with_capacity(snap::raw::max_compress_len(max_size));
+
+    // loop through all batch digests (chunked)
+    let digests: Vec<_> = batch_digests.iter().copied().collect();
+    for chunk in digests.chunks(BATCH_DIGESTS_READ_CHUNK_SIZE) {
+        // look up batches from db
+        let batches: Vec<_> = store
+            .multi_get::<NodeBatchesCache>(chunk.iter())
```
sstanfield: I think this comes from the old code but I think only checking the NodeBatchesCache may be an issue (at least until we are streaming pack files).  This implies it can only serve batches from the current epoch not any from previous epoch (which will usually work).  For instance historic syncing (until pack file streaming then this will go away) and possibly at epoch boundaries (it may introduce a race when a slower node is asking for batches at an epoch boundary and other nodes have moved on).  It seems like we want to also check the consensus packs as well.  This may have been an existing problem that somehow slid through or I may be missing something.

sstanfield (dismissed):
Looks good.  Couple of comments, one is optional and one I think might need addressing.  In the short term I think we need to address it or create and merge pack streaming first.

crates/consensus/worker/src/network/primary.rs
grantkee: Good catch. Addressed

sstanfield (approved):

PR #572, opened 2026-03-11
Fix all the tests for the new consensus pack data store.

sstanfield:
This should close out the remaining issues for restart tests.  The primary issue was cloning an ConsensusChain did not work correctly (copying critical data vs sharing it).

Also fixing the remaining unit tests.


grantkee (approved):
LGTM! Great work on this

PR #579, opened 2026-03-13
Use pack files for epoch records.

### LINKED ISSUE #573: Add functionality to wait for an epoch record.
See the loop with sleeps in manager.rs open_epoch_pach.  Would be cleaner to provide something to wait for an epoch record to arrive.
See storage/src/stores/certificate_store.rs:L250 - fn notify_read as an example or maybe the waits on consensus_bus.  We probably need a structure to manage epochs (possibly part of ConsensusChain) that could provide this feature.

### LINKED ISSUE #577: Use new pack file DB for epoch records.
Need to store the epoch records in static pack (log) files.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/573
Closes https://github.com/Telcoin-Association/telcoin-network/issues/577


crates/storage/src/epoch_records.rs
```diff
@@ -0,0 +1,950 @@
...
+    error: watch::Receiver<Option<EpochDbError>>,
+    /// Vector to map epochs to the last consensus header number.
+    /// Used for quickly deducing an epoch for a given consensus header number.
+    final_numbers: Arc<Mutex<Vec<u64>>>,
+}
+
+fn run_db_loop(
+    mut inner: Inner,
+    mut rx: Receiver<EpochDbMessage>,
+    tx_error: watch::Sender<Option<EpochDbError>>,
```
grantkee: ### 2. Error watch channel never cleared — permanent DB failure
- **Severity**: High
- **Location**: `crates/storage/src/epoch_records.rs:82-99, 171-176`
- **Description**: When the background thread encounters an error, it records it via `tx_error.send_replace(Some(e))`. All public write methods call `get_error()?` first, which returns `Err` if any error is set. However, the error is **never cleared** — there is no `send_replace(None)` anywhere. The background thread continues processing after errors, but the client side sees the stale error forever.
- **Impact**: A single transient IO error (e.g., disk full momentarily, interrupted write) permanently disables the epoch record database for all future writes until process restart. Scenario: message 1 fails with IO error → error is set; message 2 succeeds → error still set; all future `save_record()` / `save()` / `persist()` calls fail immediately from `get_error()`.
- **Proposed Fix**: Clear the error after successful operations in the background loop, or clear it on read:
```rust
// Option A: Clear on successful operation in run_db_loop
EpochDbMessage::SaveRecord(record) => {
    if let Err(e) = inner.save_record(record) {
        tx_error.send_replace(Some(e));
    } else {
        tx_error.send_replace(None);
    }
}

// Option B: Clear on read in get_error (consume-once semantics)
pub fn get_error(&self) -> Result<(), EpochDbError> {
    match &*self.error.borrow() {
        Some(e) => Err(e.clone()),
        None => Ok(()),
    }
    // Note: needs redesign since watch::Receiver doesn't allow clearing sender
}
```
Option A is simpler. Apply the same pattern to all message handlers that currently only set errors.

grantkee (dismissed):
I found a few nits and claude's review had two catches I thought seemed worth sharing

I'm not sure if the fallback for `self.epoch_idx.is_empty()` should return `None` or the dummy record since `None` falls back to dummy record anyway at the call sites

crates/storage/src/epoch_records.rs
sstanfield: get_error clears the error now.  I don't think Option A would work reliably since many commands that would put something on the write queue would fail with the error set.  Also, these errors will typically bring a node down, not being able to open or write to a DB is not something we can ignore.  So I don't think this is a high priority but clearing it should not hurt anything, if the node needs to die then one returned error should do it and if not for some reason at least it can try again.

grantkee (approved):
LGTM - great work!

PR #582, opened 2026-03-19
Refactor manager.

sstanfield:
Attempt to make manager.rs more readable by splitting it into node and epoch scoped tasks.
Also refactor the consensus_bus into node and epoch scoped channels.  This was driven by a potential footgun in the reset for epoch logic that could leave different clones with different epoch channels.  It was not triggered in current code but would be easy to happen in future so don't allow it.


grantkee (approved):
LGTM!

PR #594, opened 2026-03-25
Fix race in vote equivocation code.

### LINKED ISSUE #593: Vote Equivocation Detection Race Condition
**File:** `crates/consensus/primary/src/network/handler.rs:328-387`

The equivocation map (`auth_last_vote`) is protected by a mutex, but the lock is released between checking the last vote and inserting the new one. Concurrent vote messages from the same validator could pass detection.

**Recommendation:** Hold the lock for the entire check-and-insert operation.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/593


grantkee (approved):
LGTM! Good idea adding the timeout sanity check

PR #604, opened 2026-03-28
Forward transactions based on address

### LINKED ISSUE #605: Optimize transaction recovery and routing for gossipped and orphaned transactions
## problems 

1. **Redundant ECDSA recovery**: When gossipped transactions are received or orphaned batch transactions are recovered at epoch boundaries, the system performs ECDSA signature recovery multiple times for the same transaction. This is an expensive cryptographic operation that adds unnecessary latency.

2. **Nonce gap delays from hash-based routing**: Gossipped transactions are routed to committee validators based on the transaction hash. This means transactions from the same sender can land on different validators, causing nonce gaps that delay inclusion.

grantkee:
- accept transaction gossip based on signer, not tx hash, to limit nonce gap delays in validator mempools

closes #605

### DIFF
--- a/crates/batch-validator/src/validator.rs
+++ b/crates/batch-validator/src/validator.rs
@@ -315,7 +315,8 @@ mod tests {
     async fn test_valid_batch() {
         let tmp_dir = TempDir::new().unwrap();
         let task_manager = TaskManager::default();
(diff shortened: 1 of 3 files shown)


sstanfield (approved):
LGTM

PR #607, opened 2026-03-28
Fix some restart issues.

### LINKED ISSUE #606: TOCTOU port collision in tests
Occasionally we get port collisions when running test across node.  This is probably a TOCTOU error.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/606


grantkee (approved):
Good catch

PR #615, opened 2026-04-03
feat: basefee adjustments per worker per epoch

### LINKED ISSUE #614: Implement per-epoch base fee adjustment via WorkerConfigs contract
## Problem

The telcoin-network protocol accumulates per-worker gas usage across each epoch but never acts on it. The `adjust_base_fees` method in the epoch manager is a no-op stub that reads accumulated gas values and discards them. This means the network runs with a fixed base fee (MIN_PROTOCOL_BASE_FEE, 7 wei) regardless of demand, offering no congestion pricing or fee market signal to users and validators.

Without dynamic fee adjustment, the network cannot respond to sustained high utilization. During congestion, transactions compete purely on inclusion order rather than willingness to pay, and there is no mechanism to dampen runaway gas consumption. Governance also lacks a way to set per-worker fee strategies (e.g., fixed fees for specialized workers vs. market-driven fees for general-purpose ones).

A related gap is that per-worker fee configuration has no on-chain representation. Any fee policy must be baked into node software rather than governed through contract state, making changes require coordinated software upgrades instead of governance transactions.

## Solution

Introduce a `WorkerConfigs` contract that stores per-worker fee strategies on-chain and wire the epoch manager to read these configs and apply them at each epoch boundary.

The contract should support two fee strategies: EIP-1559 (strategy 0), which adjusts the base fee up or down by at most 12.5% based on gas utilization relative to a governance-set target, and Static (strategy 1), which sets the fee to a fixed governance-configured value. The contract should be deployed at genesis with initial configs passed via CLI flags.

At each epoch close, the epoch manager should read the canonical chain state to retrieve each worker's fee config from the contract, then compute the new base fee. For EIP-1559 workers, the computation should compare total gas consumed during the epoch against the target, applying the standard 1/8 adjustment factor. The result should be clamped to a floor of `MIN_PROTOCOL_BASE_FEE` (7 wei) with a minimum increase of 1 wei when utilization exceeds the target which ensures fees can always escape the floor under sustained demand.

The genesis CLI should accept `--worker-fee-config WORKER_ID:STRATEGY:VALUE` flags to configure initial per-worker strategies. When omitted, a single worker with EIP-1559 at a 100 billion gas target should be the default.

related to [tn-contracts #87](https://github.com/Telcoin-Association/tn-contracts/pull/87) and [#88](https://github.com/Telcoin-Association/tn-contracts/pull/88)

grantkee:
## Overview

Implements per-epoch base fee adjustment by introducing a `WorkerConfigs` contract, an EIP-1559-style fee computation, and wiring it all through genesis, the epoch manager, and a new E2E test.

### EIP-1559 base fee computation (`crates/types/src/gas_accumulator.rs`)
- Added `WorkerFeeConfig` enum with `Eip1559 { target_gas }` and `Static { fee }` variants
- Added `compute_next_base_fee_eip1559()` that adjusts fee by up to 12.5% based on gas utilization vs target, floored at `MIN_PROTOCOL_BASE_FEE` (7 wei), with a minimum 1 wei increase when over target
- 12 unit tests covering edge cases: zero target, overflow safety, min fee floor, partial utilization, cap at 12.5%

### WorkerConfigs contract integration (`crates/tn-reth/src/lib.rs`, `crates/tn-reth/src/system_calls.rs`)
- Added `WorkerConfigs` sol interface with `getWorkerConfig`, `setWorkerConfig`, `setNumWorkers`, and `numWorkers` functions
- Deploy `WorkerConfigs` contract during genesis ceremony alongside existing contracts (BLSG1, ConsensusRegistry, Issuance)
- Added `get_worker_fee_configs()` method on `RethEnv` that reads per-worker configs from on-chain state via EVM calls
- Unit test verifying round-trip deploy and read of multi-worker configs

### Epoch manager (`crates/node/src/manager/node/epoch.rs`)
- Replaced no-op `adjust_base_fees` with real implementation that reads fee configs from `WorkerConfigs` contract and applies EIP-1559 or Static strategy per worker
- `close_epoch` now accepts an `engine` reference to access the EVM for contract reads
- All three `close_epoch` call sites updated

### Genesis CLI (`crates/telcoin-network-cli/src/genesis/mod.rs`)
- Added `--worker-fee-config` CLI flag with format `WORKER_ID:STRATEGY:VALUE`
- Default when omitted: single worker 0 with EIP-1559 at 100B gas target
- Parsing validates contiguous worker IDs starting from 0
- 6 unit tests for parsing logic

### Genesis config (`crates/config/src/genesis.rs`)
- Added `WORKER_CONFIGS_ADDRESS` and `WORKER_CONFIGS_JSON` constants for the new contract

### Epoch duration default change
- Default epoch duration changed from 24 hours (86400s) to 8 hours (28800s)
- Genesis tests and staking tests updated to match

### E2E test (`crates/e2e-tests/tests/it/basefees.rs`)
- New `test_basefee_adjustment_and_sync` test that spins up 5 validators, sends value transfers across epochs, verifies fees increase above 7 wei, kills and restarts a node, and confirms the restarted node syncs to the correct base fee

### tn-contracts submodule
- Updated to commit `9edfc49e` which includes the `WorkerConfigs` contract

closes #614


crates/tn-reth/src/lib.rs
```diff
@@ -1412,6 +1460,49 @@ impl RethEnv {
         self.call_consensus_registry::<_, Vec<ConsensusRegistry::ValidatorInfo>>(evm, calldata)
     }
 
+    /// Read fee configs for all workers from the [`WorkerConfigs`] contract on-chain.
+    ///
+    /// Creates a single EVM against the canonical tip and makes one `getWorkerConfig` call
+    /// per worker, returning a [`WorkerFeeConfig`] for each.
+    pub fn get_worker_fee_configs(&self, num_workers: usize) -> eyre::Result<Vec<WorkerFeeConfig>> {
+        let canonical_tip = self.canonical_tip();
```
sstanfield: Per our discussion we want to use the previous epoch's closing block not the canonical tip here?

sstanfield (dismissed):

grantkee:
Closing in favor of #656

PR #621, opened 2026-04-08
Pull batches from both the cache and the consensus chain.

sstanfield:
Expose the consensus chain DB to the worker network and use both the batch cache and consensus chain to find requested batches.


grantkee (approved):
Good catch

PR #625, opened 2026-04-09
Cleaned up issues with batch caching.

### LINKED ISSUE #623: Review batch cache code.
- Have an extra write at worker/worker.rs 267.
- Check epoch of batches in worker/network/handler.rs process_gossip.  Only subscribe when seem to be in the current epoch.
- In orphan_batches should only process our batches.
- Clear batch cache on epoch change- handle orphans in case any batches were made post epoch close.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/623

Reviewed and cleaned up some issues with batch caching and orphan batches.

### DIFF
--- a/crates/consensus/worker/src/worker.rs
+++ b/crates/consensus/worker/src/worker.rs
@@ -12,10 +12,10 @@ use crate::{
 use std::{sync::Arc, time::Duration};
 use tn_config::ConsensusConfig;
 use tn_network_types::{local::LocalNetwork, WorkerOwnBatchMessage, WorkerToPrimaryClient};
-use tn_storage::tables::NodeBatchesCache;
+use tn_storage::tables::{NodeBatchesCache, OurNodeBatchesCache};
 use tn_types::{
-    error::BlockSealError, BatchReceiver, BatchSender, BatchValidation, Database, SealedBatch,
-    TaskManager, WorkerId,
+    error::BlockSealError, BatchReceiver, BatchSender, BatchValidation, Database, DbTxMut as _,
+    SealedBatch, TaskManager, WorkerId,
 };
 use tracing::{error, info, instrument};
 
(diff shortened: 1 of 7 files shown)


grantkee (dismissed):
LGTM! Good work

crates/consensus/worker/src/worker.rs
```diff
@@ -240,6 +240,8 @@ impl<DB: Database, QW: QuorumWaiterTrait> Worker<DB, QW> {
                     Err(e) => {
                         return Err(match e {
                             crate::quorum_waiter::QuorumWaiterError::QuorumRejected => {
+                                // Explicitly rejected — don't re-inject at epoch boundary
+                                let _ = self.store.remove::<OurNodeBatchesCache>(&digest);
                                 BlockSealError::QuorumRejected
                             }
                             crate::quorum_waiter::QuorumWaiterError::AntiQuorum => {
```
grantkee, line 249: Should `AntiQuorum` remove the batch too?

crates/node/src/manager/node/epoch.rs
```diff
@@ -308,28 +308,19 @@ where
...
+        // We are reintroducing the transactions so these batches are now defunct.
+        self.consensus_db.clear_table::<OurNodeBatchesCache>()?;
         if !orphan_batches.is_empty() {
             let consensus_bus = self.consensus_bus.clone();
             let span =
                 info_span!(target: "telcoin", "orphan-batches", epoch = tracing::field::Empty);
             span.record("epoch", epoch.to_string());
             epoch_task_manager.spawn_task("Orphaned Batches", async move {
-                info!(target: "epoch-manager", "Re-introducing orphaned batchs {} transactions", orphan_batches.len());
+                info!(target: "epoch-manager", "Re-introducing orphaned batches {} transactions", orphan_batches.len());
```
grantkee, line 323: nit: log says "transactions" but logs num batches

sstanfield: Fixed in my next PR.

crates/consensus/worker/src/worker.rs
sstanfield: Good catch, should remove for any error.  On error the txns should be available for a future batch so need to get rid of this cached batch in all cases.

grantkee (approved):

PR #626, opened 2026-04-10
patch: fix proposer min delay logic for odd rounds

### LINKED ISSUE #627: Proposer skips min delay on odd rounds, starving the certifier
## Problem

The primary proposer's `calc_min_delay` method controls how long a node waits before proposing its next header. The delay is supposed to drop to zero only when the node expects to be the leader of the upcoming even round — giving it the best chance of getting its leader header committed. On all other rounds (odd rounds, or even rounds where someone else leads), the node should wait `min_header_delay` before proposing.

The current implementation uses a negated compound condition that inverts this intent. The logic resolves to: return zero delay whenever the next round is odd *or* whenever this node is the upcoming even-round leader. In practice, every node proposes with zero delay on every odd round regardless of leadership.

This matters because even-round headers can carry batches. If the certifier hasn't finished certifying the previous even-round header before the next proposal fires, that proposal cancels the in-flight certificate. Under zero delay on odd rounds, the window for certification shrinks to near-zero, increasing the chance that batch-bearing headers are dropped and must be re-proposed. The effect is worse on slower or resyncing nodes, which already have tighter timing margins.

## Solution

Replace the negated condition with a direct positive check: zero delay only when the next round is even and this node is the anticipated leader. All other cases — odd rounds and even rounds where another node leads — should use `min_header_delay`.

This aligns `calc_min_delay` with the structure already used by `calc_max_delay`, which checks the same positive condition (next round is even, this node is leader) and applies a reduced delay only in that case. The two methods should mirror each other in structure since they govern the same timing window for the same reason.

The doc comment should also be updated to reflect the corrected intent: leaders are elected on even rounds only, odd rounds always use the standard delay to give the certifier breathing room.

grantkee:
- Fix inverted boolean in `calc_min_delay`: the negated compound condition was granting zero delay on all odd rounds instead of only when this node is the next even-round leader. Now mirrors the positive check used by `calc_max_delay`.
- Odd rounds now wait `min_header_delay`, giving the certifier time to certify the previous even-round header before the next proposal cancels it.
- Updated doc comment to match corrected behavior.

closes #627

### DIFF
--- a/crates/consensus/primary/src/proposer.rs
+++ b/crates/consensus/primary/src/proposer.rs
@@ -295,21 +295,15 @@ impl<DB: Database> Proposer<DB> {
     /// The min delay is reduced when this authority expects to become the leader of the next round.
     /// Reducing the min delay increases the chances of successfully committing a leader.
     ///
-    /// NOTE: If the next round is even, the leader schedule is used to identify the next leader. If
-    /// the next round is odd, the whole committee is used in order to keep the proposal rate as
-    /// high as possible (which leads to a higher round rates). Using the entire committee here also
-    /// helps boost scores for weaker nodes that may be trying to resync.
+    /// Leaders are only elected on even rounds. For even rounds, zero delay when this node is the
+    /// anticipated leader increases commit chance. For odd rounds, use min_header_delay so the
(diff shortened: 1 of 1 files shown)


sstanfield (approved):
Looks good and great catch.

PR #630, opened 2026-04-15
Stream pack files for historic epoch data sync.

### LINKED ISSUE #629: Stream pack files for historic epoch data when syncing.
Use the DB pack files and the new network stream support to stream entire DB pack files for historic epoch data.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/629

### DIFF
--- a/crates/consensus/primary/src/consensus_bus.rs
+++ b/crates/consensus/primary/src/consensus_bus.rs
@@ -215,7 +215,7 @@ pub struct ConsensusBusAppInner {
     /// Watch tracking most recently seen consensus header.
     tx_last_consensus_header: watch::Sender<Option<ConsensusHeader>>,
     /// Watch tracking the last gossipped consensus block number and hash.
-    tx_last_published_consensus_num_hash: watch::Sender<(u64, BlockHash)>,
+    tx_last_published_consensus_num_hash: watch::Sender<(Epoch, u64, BlockHash)>,
 
     /// Consensus header.  Note this can be used to create consensus output to execute for non
     /// validators.
@@ -232,6 +232,10 @@ pub struct ConsensusBusAppInner {
     tx_epoch_record: watch::Sender<Option<EpochRecord>>,
     /// The que channel for primary network events.
     primary_network_events: QueChannel<NetworkEvent<crate::network::Req, crate::network::Res>>,
+    /// Sender for epoch records that need to have a pack file downloaded.
+    epoch_request_queue_tx: tokio::sync::mpsc::Sender<EpochRecord>,
+    /// Reciever for epoch records to download pack files for.
+    epoch_request_queue_rx: Arc<tokio::sync::Mutex<tokio::sync::mpsc::Receiver<EpochRecord>>>,
 }
 
 /// The thread-safe inner type that holds all the channels for inner-consensus
@@ -261,7 +265,8 @@ impl ConsensusBusApp {
 
         let (tx_primary_round_updates, _) = watch::channel(0u32);
         let (tx_last_consensus_header, _) = watch::channel(None);
-        let (tx_last_published_consensus_num_hash, _) = watch::channel((0, BlockHash::default()));
+        let (tx_last_published_consensus_num_hash, _) =
+            watch::channel((0, 0, BlockHash::default()));
 
         let (tx_recent_blocks, _) = watch::channel(RecentBlocks::new(recent_blocks as usize));
         let (tx_sync_status, _) = watch::channel(NodeMode::default());
@@ -271,6 +276,8 @@ impl ConsensusBusApp {
 
         let (tx_epoch_record, _) = watch::channel(None);
 
+        let (epoch_request_queue_tx, epochs_rx) = tokio::sync::mpsc::channel(10_000);
+        let epoch_request_queue_rx = Arc::new(tokio::sync::Mutex::new(epochs_rx));
         Self {
             inner: Arc::new(ConsensusBusAppInner {
                 tx_committed_round_updates,
@@ -286,6 +293,8 @@ impl ConsensusBusApp {
                 new_epoch_votes: QueChannel::new(),
                 tx_epoch_record,
                 primary_network_events: QueChannel::new_always_subscribed(),
+                epoch_request_queue_tx,
+                epoch_request_queue_rx,
             }),
         }
     }
@@ -351,12 +360,12 @@ impl ConsensusBusApp {
     /// Track the latest published consensus header block number and hash seen on the gossip
     /// network. This value will have been verified and can be trusted to be the correct hash
     /// for block number.  DO NOT send unverified values to this watch.
-    pub fn last_published_consensus_num_hash(&self) -> &watch::Sender<(u64, BlockHash)> {
+    pub fn last_published_consensus_num_hash(&self) -> &watch::Sender<(Epoch, u64, BlockHash)> {
         &self.inner.tx_last_published_consensus_num_hash
     }
 
     /// Returns the latest verified consensus block number and hash from gossip.
-    pub fn published_consensus_num_hash(&self) -> (u64, BlockHash) {
+    pub fn published_consensus_num_hash(&self) -> (Epoch, u64, BlockHash) {
         *self.inner.tx_last_published_consensus_num_hash.borrow()
     }
 
@@ -536,6 +545,18 @@ impl ConsensusBusApp {
             .consensus_header_by_digest(epoch, latest_consensus.hash)
             .await
             .unwrap_or_default()
+    }
+
+    /// Send a request to down load the epoch pack file for the provided EpochRecord.
+    pub async fn request_epoch_pack_file(&self, epoch_record: EpochRecord) {
+        let _ = self.inner.epoch_request_queue_tx.send(epoch_record).await;
+    }
+
+    /// Retreive the next request to down load an epoch pack file.
+    /// Will not resolve until a request is ready and will only ever
+    /// provide each request once.  Returns None when the underlying channel closes.
+    pub async fn get_next_epoch_pack_file_request(&self) -> Option<EpochRecord> {
+        self.inner.epoch_request_queue_rx.lock().await.recv().await
     }
 }
 
--- a/crates/consensus/primary/src/consensus_bus.rs
+++ b/crates/consensus/primary/src/consensus_bus.rs
@@ -547,12 +547,12 @@ impl ConsensusBusApp {
             .unwrap_or_default()
     }
 
-    /// Send a request to down load the epoch pack file for the provided EpochRecord.
+    /// Send a request to download the epoch pack file for the provided EpochRecord.
     pub async fn request_epoch_pack_file(&self, epoch_record: EpochRecord) {
         let _ = self.inner.epoch_request_queue_tx.send(epoch_record).await;
     }
 
-    /// Retreive the next request to down load an epoch pack file.
+    /// Retrieve the next request to down load an epoch pack file.
     /// Will not resolve until a request is ready and will only ever
     /// provide each request once.  Returns None when the underlying channel closes.
     pub async fn get_next_epoch_pack_file_request(&self) -> Option<EpochRecord> {
--- a/crates/consensus/primary/src/consensus_bus.rs
+++ b/crates/consensus/primary/src/consensus_bus.rs
@@ -233,9 +233,10 @@ pub struct ConsensusBusAppInner {
     /// The que channel for primary network events.
     primary_network_events: QueChannel<NetworkEvent<crate::network::Req, crate::network::Res>>,
     /// Sender for epoch records that need to have a pack file downloaded.
-    epoch_request_queue_tx: tokio::sync::mpsc::Sender<EpochRecord>,
+    epoch_request_queue_tx: tokio::sync::mpsc::Sender<(EpochRecord, EpochRecord)>,
     /// Reciever for epoch records to download pack files for.
-    epoch_request_queue_rx: Arc<tokio::sync::Mutex<tokio::sync::mpsc::Receiver<EpochRecord>>>,
+    epoch_request_queue_rx:
+        Arc<tokio::sync::Mutex<tokio::sync::mpsc::Receiver<(EpochRecord, EpochRecord)>>>,
 }
 
 /// The thread-safe inner type that holds all the channels for inner-consensus
@@ -548,14 +549,18 @@ impl ConsensusBusApp {
     }
 
     /// Send a request to download the epoch pack file for the provided EpochRecord.
-    pub async fn request_epoch_pack_file(&self, epoch_record: EpochRecord) {
-        let _ = self.inner.epoch_request_queue_tx.send(epoch_record).await;
+    pub async fn request_epoch_pack_file(
+        &self,
+        previous_epoch_record: EpochRecord,
+        epoch_record: EpochRecord,
+    ) {
+        let _ = self.inner.epoch_request_queue_tx.send((previous_epoch_record, epoch_record)).await;
     }
 
     /// Retrieve the next request to down load an epoch pack file.
     /// Will not resolve until a request is ready and will only ever
     /// provide each request once.  Returns None when the underlying channel closes.
-    pub async fn get_next_epoch_pack_file_request(&self) -> Option<EpochRecord> {
+    pub async fn get_next_epoch_pack_file_request(&self) -> Option<(EpochRecord, EpochRecord)> {
         self.inner.epoch_request_queue_rx.lock().await.recv().await
     }
 }
--- a/crates/consensus/primary/src/consensus_bus.rs
+++ b/crates/consensus/primary/src/consensus_bus.rs
@@ -276,7 +276,7 @@ impl ConsensusBusApp {
 
         let (tx_epoch_record, _) = watch::channel(None);
 
-        let (epoch_request_queue_tx, epochs_rx) = tokio::sync::mpsc::channel(10_000);
+        let (epoch_request_queue_tx, epochs_rx) = tokio::sync::mpsc::channel(1024);
         let epoch_request_queue_rx = Arc::new(tokio::sync::Mutex::new(epochs_rx));
         Self {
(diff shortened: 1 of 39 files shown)


crates/state-sync/src/consensus.rs
```diff
@@ -150,65 +147,57 @@ pub(crate) async fn spawn_track_recent_consensus<DB: TNDatabase>(
...
+                let epoch = epoch_record.epoch;
+                let prev_epoch_num = epoch.saturating_sub(1);
+                if let Some((previous_epoch, _)) = consensus_chain.epochs().get_epoch_by_number(prev_epoch_num).await {
+                    info!(target: "state-sync", "epoch consensus fetcher {task_index} retreiving epoch {epoch}");
+                    loop {
+                        tokio::select! {
+                            result = network.request_epoch_pack(&epoch_record, &previous_epoch, &consensus_chain) => {
+                                match result {
+                                    Ok(_) => break,
+                                    Err(e) => {
```
grantkee: Should we add a backoff delay here? The underlying method `request_epoch_pack` requests from 3 random peers, but might be good to sleep here for a bit before looping again like `spawn_epoch_record_collector`

crates/consensus/primary/src/network/handler.rs
```diff
@@ -903,4 +911,78 @@ where
...
+    async fn send_epoch_over_stream<S>(
+        stream: &mut S,
+        consensus_chain: &ConsensusChain,
+        epoch: Epoch,
+    ) -> PrimaryNetworkResult<()>
+    where
+        S: AsyncWrite + Unpin + Send,
+    {
+        let mut bytes = vec![0_u8; 16 * 1024]; // Use a 16kb read buffer.
+        if let Ok(mut epoch_stream) = consensus_chain.get_epoch_stream(epoch).await {
```
grantkee: What is the advantage of silently failing when `consensus_chain.get_epoch_stream(epoch)` returns an `Err` here? Would it be better to return that an error to the requestor?

crates/consensus/primary/src/network/mod.rs
```diff
@@ -295,12 +456,39 @@ where
...
+            loop {
+                tokio::select! {
+                    // process network events
+                    next = self.network_events.recv() => {
+                        match next {
+                            Some(event) => {
+                                self.process_network_event(event);
+                            }
+                            None => {
+                                warn!(target: "worker::network", "critical worker network events channel dropped");
```
grantkee: nit: `target: "primary::network"`

crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -537,6 +546,18 @@ impl ConsensusBusApp {
             .await
             .unwrap_or_default()
     }
+
+    /// Send a request to down load the epoch pack file for the provided EpochRecord.
```
grantkee: sp: "down load" -> "download"

crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -537,6 +546,18 @@ impl ConsensusBusApp {
...
+    pub async fn request_epoch_pack_file(&self, epoch_record: EpochRecord) {
+        let _ = self.inner.epoch_request_queue_tx.send(epoch_record).await;
+    }
+
+    /// Retreive the next request to down load an epoch pack file.
```
grantkee: sp: "Retreive" -> "Retrieve"

crates/state-sync/src/consensus.rs
```diff
@@ -150,65 +147,57 @@ pub(crate) async fn spawn_track_recent_consensus<DB: TNDatabase>(
...
-                        let _ = tx.send((epoch, num, hash)).await;
-                    } else {
-                        info!(target: "state-sync", ?new_epoch, ?epoch, "tracking recent consensus for an epoch - changed epochs, finished");
-                        // We are done so can accept a new epoch.
-                        next_permit.take();
+            Some((_permit, epoch_record)) = next_epoch(&consensus_bus, &next_sem) => {
+                let epoch = epoch_record.epoch;
+                let prev_epoch_num = epoch.saturating_sub(1);
+                if let Some((previous_epoch, _)) = consensus_chain.epochs().get_epoch_by_number(prev_epoch_num).await {
+                    info!(target: "state-sync", "epoch consensus fetcher {task_index} retreiving epoch {epoch}");
```
grantkee: sp: "retreiving" -> "retrieving"

crates/consensus/primary/src/network/mod.rs
```diff
@@ -463,6 +656,86 @@ where
...
+                    // permit drops here, freeing the slot
+                    false
+                } else {
+                    let mut hasher = tn_types::DefaultHashFunction::new();
+                    hasher.update(&epoch.to_le_bytes());
+                    let request_digest = B256::from_slice(hasher.finalize().as_bytes());
+                    let pending = PendingEpochStream::new(epoch, permit);
+                    // If the same peer requests the same epoch then replace request.
+                    // If the peer tries to stream twice the second attempt will be
+                    // punished as a protocol violation.
```
grantkee: I do not see where the peer is punished? We need to ensure round-robin requests are limited so syncing nodes don't overwhelm bandwidth with requests

crates/consensus/primary/src/network/mod.rs
```diff
@@ -463,6 +656,86 @@ where
...
+                    if pending_map.insert((peer, request_digest), pending).is_some() {
```
grantkee, line 788: The comment mentions "identical epoch request" but this would update the timestamp and delay pruning the request

Attacker could resend the request every ~20s to keep the entry alive without opening a stream

I didn't consider this for the worker, so this comment applies to both

crates/consensus/primary/src/error/network.rs
```diff
@@ -108,6 +111,7 @@ impl From<&PrimaryNetworkError> for Option<Penalty> {
             },
             | PrimaryNetworkError::InvalidRequest(_)
             | PrimaryNetworkError::UnknownConsensusHeaderDigest(_)
+            | PrimaryNetworkError::UnknownStreamRequest(_)
```
grantkee, line 117: This is more appropriate penalty. I adjusted this in the stacked PR

crates/consensus/primary/src/network/mod.rs
```diff
@@ -480,6 +753,52 @@ where
...
+        self.task_spawner.spawn_task(task_name, async move {
+            // read the request digest (32-bytes) from the stream with timeout
+            let mut digest_buf = [0u8; tn_types::DIGEST_LENGTH];
+            match tokio::time::timeout(
+                std::time::Duration::from_secs(5),
+                stream.read_exact(&mut digest_buf),
+            ).await {
+                Ok(Ok(())) => {}
+                Ok(Err(e)) => {
+                    warn!(target: "worker::network", %peer, ?e, "failed to read request digest from stream");
```
grantkee: nit: `target: worker`

crates/consensus/primary/src/network/mod.rs
```diff
@@ -480,6 +753,52 @@ where
...
+                    return;
+                }
+                Err(_) => {
+                    warn!(target: "worker::network", %peer, "timeout reading request digest from stream");
```
grantkee: nit: `target: worker`

crates/consensus/primary/src/network/mod.rs
```diff
@@ -480,6 +753,52 @@ where
...
+            let opt_pending_req = pending_map
+                .lock()
+                .remove(&(peer, request_digest));
+
+            // process stream
+            if let Err(err) = request_handler
+                .process_request_epoch_stream(peer, opt_pending_req, stream, request_digest, &consensus_chain)
+                .await {
+                    // apply applicable penalty for error
+                    warn!(target: "worker::network", ?err, "error processing request batches stream");
```
grantkee: nit: `target: worker`

crates/consensus/primary/src/network/handler.rs
```diff
@@ -903,4 +911,78 @@ where
     pub(super) fn consensus_chain(&self) -> &ConsensusChain {
         &self.consensus_chain
     }
+
+    /// Send epoch pack file over stream.
+    async fn send_epoch_over_stream<S>(
```
grantkee, line 916: How can we make this timeout more granular? Right now, each 16kb chunk has 300s before the timeout triggers

crates/consensus/primary/src/network/mod.rs
```diff
@@ -249,6 +312,81 @@ impl PrimaryNetworkHandle {
...
+                })?;
+
+                info!(
+                    target: "primary::network",
+                    %peer,
+                    epoch,
+                    "stream opened - reading and validating epoch pack file..."
+                );
+
+                consensus_chain
```
grantkee: I think this should have some timeout to prevent peers from ack'ing the stream and then slow streams. Ideally, we can use a per-record timeout inside `ConsensusPack::stream_import` since it's easier to estimate per record timeouts than entire pack files?

crates/consensus/primary/src/network/handler.rs
```diff
@@ -903,4 +911,78 @@ where
...
+        debug!(
+            target: "primary::network",
+            %peer,
+            ?request_digest,
+            epoch = request.epoch,
+            "processing inbound epoch stream"
+        );
+
+        // set timeout to prevent slow-read attack
+        match tokio::time::timeout(
```
grantkee: These errors should result in penalties too. I think elapsed and `io_err` can both be mild depending on how individual entries timeout

crates/consensus/primary/src/network/mod.rs
```diff
@@ -249,6 +312,81 @@ impl PrimaryNetworkHandle {
...
+        let request = PrimaryRequest::StreamEpoch { epoch };
+        let mut hasher = tn_types::DefaultHashFunction::new();
+        hasher.update(&epoch.to_le_bytes());
+        let request_digest = B256::from_slice(hasher.finalize().as_bytes());
+
+        for _ in 0..3 {
+            // send request and await response from peer
+            //
+            // SAFETY: network layer handles request timeout
+            if let PrimaryResponse::RequestEpochStream { ack, peer } =
```
grantkee: We should capture the peer's BLS key if `ack` and report a mild penalty if they ack but fail to stream

grantkee (changes requested):
I found some issues, made some requests, and stacked a PR for something I missed in the worker path too. This is a tricky path to secure, but overall it's looking good!

crates/storage/src/consensus.rs
```diff
@@ -353,25 +353,114 @@ impl ConsensusChain {
...
+                        }
+                    }
+                    None => {
+                        // Missing a final consensus header...
+                        let _ = std::fs::remove_dir_all(&path);
+                        return Err(ConsensusChainError::EmptyImport);
+                    }
+                }
+                let mut current_pack = self.current_pack.lock();
+                let replace_current = if let Some(current_pack) = &*current_pack {
```
grantkee, line 412: In `stream_import`, the sequence:
1. `current_pack.lock()` → `*current_pack = None` if current_pack.epoch() == epoch (consensus.rs:400-408).
2. `drop(pack); drop(current_pack)` (consensus.rs:409-410).
3. `if fs::exists(&base_dir)` → `recent_packs.lock().retain(|p| p.epoch() != epoch)` (consensus.rs:417).
4. `fs::remove_dir_all(&base_dir)` (consensus.rs:418).
5. `fs::rename(&path_base_dir, &base_dir)` (consensus.rs:420).

Between steps 3 and 5, a concurrent `get_static(epoch)` call (called from `get_epoch_stream` which is in turn called from `send_epoch_over_stream` when any peer requests this epoch):
- Sees `current_pack = None` (step 1).
- Sees `recent_packs` no entry (step 3).
- Falls through to `ConsensusPack::open_static(&base_path, epoch)`.

If this fallback executes between steps 3 and 4, it succeeds on the pre-remove directory, then gets pushed into `recent_packs` (consensus.rs:694). After step 5 completes, the `recent_packs` entry holds an open FD to the OLD inode (on Unix, kept alive until the FD closes). Future readers through `recent_packs` receive stale bytes until LRU eviction.

Alternatively, if the fallback executes between steps 4 and 5, it fails ENOENT → `StreamUnavailable` → peer receives empty stream (and `get_epoch_stream` silently swallows the error).

grantkee: Proposed fix:

```rust
// Make sure we don't have any cruft in the final dir.
if std::fs::exists(&base_dir).unwrap_or_default() {
    // If this exists it is incomplete (see check at start of function).
    // Cached readers of the old pack keep working via their open FDs
    // (inode is pinned until they drop). Fall-through readers during
    // this narrow window may see ENOENT and get StreamUnavailable.
    let _ = std::fs::remove_dir_all(&base_dir);
}
std::fs::rename(&path_base_dir, &base_dir)?;
// Invalidate the cache AFTER the rename so a concurrent get_static that
// missed the cache and opened FDs on the old (now-unlinked) inode cannot
// leave a stale entry behind for other callers — any entry cached during
// the race is purged here. Readers after this point fall through and
// see the new on-disk pack.
self.recent_packs.lock().retain(|p| p.epoch() != epoch);
```

crates/consensus/primary/src/network/handler.rs
sstanfield: Not sure I follow, we have 300 seconds for the entire file.  Or are you suggesting we timeout per 16kb chunk vs the entire file?

crates/consensus/primary/src/network/mod.rs
sstanfield: Fixed in your PR.

crates/consensus/primary/src/network/mod.rs
sstanfield: In process_inbound_stream- I think your PR added this though.

crates/storage/src/consensus.rs
sstanfield: This makes sense.  I had iterated on all this already but missed that edge case.

crates/consensus/primary/src/network/handler.rs
```diff
@@ -6,30 +6,36 @@ use super::{
...
-    Vote,
+    Vote, B256,
 };
-use tokio::sync::Mutex as TokioMutex;
+use tokio::{io::AsyncReadExt, sync::Mutex as TokioMutex, time::timeout};
 use tracing::{debug, error, info, warn};
 
+/// Total timeout for sending all batches over a stream.
+/// Prevents slow-reader attacks where a peer accepts a stream but never reads.
+/// Set to 5 minutes as a an arbitrary upper bound on downloading a pack file.
```
grantkee: Outdated comment

crates/consensus/primary/src/network/handler.rs
```diff
@@ -965,24 +968,24 @@ where
...
-                warn!(target: "primary::network", %peer, ?request_digest, "sending epoch stream timed out");
-            }
-        }
+        };
 
         // attempt to close the stream gracefully
         let _ = stream.close().await;
 
-        Ok(())
+        result
```
grantkee: nit: simplify this by moving the `stream.close().await` into `Self::send_epoch_over_stream` and using `inspect_err` after the await:
```rust
      Self::send_epoch_over_stream(
            &mut stream,
            consensus_chain,
            request.epoch,
            SEND_STREAM_BUFFER_TIMEOUT,
        )
        .await
        .inspect_err(
            |e| warn!(target: "primary::network", %peer, ?e, "failed to send epoch over stream"),
        )
```
If you'd rather close the stream in the outer `process_request_epoch_stream` that's okay but it seems a little more straightforward IMO to close the stream inside `send_epoch_over_stream` and returning the call directly

crates/storage/src/consensus_pack.rs
```diff
@@ -714,8 +728,8 @@ impl Inner {
         .map_err(OpenError::IndexFileOpen)?;
         let mut parent_digest = previous_epoch.final_consensus.hash;
         let mut batches = HashSet::new();
-        while let Some(record) = stream_iter.next().await {
-            let record = record.map_err(|e| PackError::ReadError(e.to_string()))?;
+        while let Some(record) = next(&mut stream_iter, timeout).await? {
+            //XXXXlet record = record.map_err(|e| PackError::ReadError(e.to_string()))?;
```
grantkee: leftover

crates/storage/src/consensus_pack.rs
```diff
@@ -668,19 +670,31 @@ impl Inner {
...
     /// Create a new set of epoch static files to write consensus output into.
-    async fn stream_import<P: AsRef<Path>, R: AsyncRead + AsyncSeek + Unpin>(
+    async fn stream_import<P: AsRef<Path>, R: AsyncRead + Unpin>(
         path: P,
         stream: R,
         epoch: Epoch,
         previous_epoch: &EpochRecord,
+        timeout: Duration,
     ) -> Result<Self, PackError> {
+        async fn next<R: AsyncRead + Unpin>(
```
grantkee, line 720: Missing doc comment

crates/storage/src/archive/pack_iter.rs
grantkee, line 160: There is no upper bound on this value. The vec allocates based on peer's reported size and precedes CRC verification. What would be a reasonable max record size to enforce?

crates/consensus/primary/src/network/mod.rs
grantkee: Ready to penalize peer here?

crates/consensus/primary/src/network/mod.rs
```diff
@@ -249,6 +312,87 @@ impl PrimaryNetworkHandle {
...
+                    self.report_penalty(peer, Penalty::Mild).await;
+                }
+                res?;
+                info!(
+                    target: "primary::network",
+                    %peer,
+                    epoch,
+                    "streamed epoch pack file"
+                );
+
```
grantkee, line 429: Should this `continue` here to use the full 3 peer attempts? Otherwise, this could exit early

grantkee (approved):
Good work

PR #633, opened 2026-04-23
idempotent request replacement

grantkee:
- Primary and worker pending-request replacement now preserves `created_at`, so a peer re-requesting the same epoch / batch set cannot rearm the 30s cleanup timer and squat a slot indefinitely
- Worker's `UnknownStreamRequest` drops from `Penalty::Fatal` to `Penalty::Mild`, matching the primary. The error fires when a stream open races the sweeper — a transient network condition, not a malicious act
- Worker gets a `debug!` log on replacement, mirroring the primary's existing hook for squat-detection telemetry
- Unit tests on both paths assert: replacement preserves the original `created_at`, the stale entry is evicted on schedule, and the semaphore permit is returned


sstanfield (approved):
LGTM!

PR #639, opened 2026-04-29
Tighten some of consensus chain data.

### LINKED ISSUE #638: Include reputation scores in sub dag digest.
Need to do this so they are verified as part of the digest for syncing pack files.
Also remove the unused (except for one unnecessary metric) created_at field from Certificates.  This will save a lot of storage.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/638

Make sure we include rep scores in digests and remove created_at from certs (not really needed so save the 8 bytes).

### DIFF
--- a/crates/types/src/primary/block.rs
+++ b/crates/types/src/primary/block.rs
@@ -49,6 +49,9 @@ impl ConsensusHeader {
         hasher.update(parent_hash.as_slice());
         hasher.update(sub_dag.digest().as_ref());
         hasher.update(number.to_le_bytes().as_ref());
+        // Include the extra field.
+        // Not using this yet but include the default in the hash in prep when we do.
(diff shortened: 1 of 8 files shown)


crates/types/src/primary/block.rs
```diff
@@ -49,6 +49,9 @@ impl ConsensusHeader {
         hasher.update(parent_hash.as_slice());
         hasher.update(sub_dag.digest().as_ref());
         hasher.update(number.to_le_bytes().as_ref());
+        // Include the extra field.
+        // Not using this yet but include the default in the hash in prep when we do.
+        hasher.update(B256::default().as_slice());
```
grantkee, line 54: nit: future proof this now by passing  `extra_field` as an arg

grantkee (approved):
LGTM!

PR #641, opened 2026-04-30
Only store headers not entire certs in subdags.

sstanfield:
This reduces storage and makes it easier to validate pack files (no need to verify all certs from a peer).

### DIFF
--- a/crates/types/src/primary/output.rs
+++ b/crates/types/src/primary/output.rs
@@ -264,7 +264,7 @@ impl Display for ConsensusOutput {
 /// Note it stores Headers without certificates, all validation
 /// should be complete.  Future validation can be done by verifying
 /// the consensus chain against signed checkpoints (like epoch records).
-#[derive(PartialEq, Serialize, Deserialize, Clone, Debug, Default)]
+#[derive(PartialEq, Serialize, Deserialize, Clone, Debug)]
 pub struct CommittedSubDag {
     /// The sequence of committed certificates.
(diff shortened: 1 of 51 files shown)


crates/types/src/primary/output.rs
```diff
@@ -278,9 +277,26 @@ pub struct CommittedSubDag {
...
+            reputation_score: Default::default(),
+            commit_timestamp: Default::default(),
+            randomness: Default::default(),
+        }
+    }
 }
 
 impl CommittedSubDag {
+    /// Create a new CommittedSubDag.
+    /// Note that leader MUST be the first element or certificates or this will panic.
```
grantkee, line 299: nit: leader is expected to be the last (not first)

crates/types/src/primary/output.rs
```diff
@@ -295,36 +311,48 @@ impl CommittedSubDag {
...
-            leader.header().created_at(), previous_sub_dag_ts, commit_timestamp);
+                leader.header().created_at(), previous_sub_dag_ts, commit_timestamp);
         }
-
-        Self { certificates, leader, reputation_score, commit_timestamp }
-    }
-
+        // Make sure the leader is the LAST certificate.
+        //
+        assert_eq!(leader.digest(), certificates.last().map(|c| c.digest()).unwrap_or_default());
```
grantkee, line 318: I'm not crazy about this but I suppose it is a `panic`-able offense

grantkee (approved):
LGTM!

PR #643, opened 2026-05-02
Delay remaval from our batches cache until we are sure.

### LINKED ISSUE #642: Cache for our batches may be cleared to soon.
Occasional failures of the epoch_boundary test seem to be caused by transaction loss due to being too aggressive with removal for our batches cache.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/642


grantkee (approved):
Great catch! This is a clean and elegant solutions. Nice work

PR #645, opened 2026-05-02
Fix race uncovered by restart failure.

### LINKED ISSUE #644: Failure in test_restarts_delayed.
Got a rare failure in test_resrts_delayed.  Claude produced this for the issue which seems reasonable:
Root cause identified and fixed.

  The bug: In handler.rs, when a proposer (node1) re-proposes a header with 0 parents (fresh certifier restart), but the handler's auth_last_vote cache had a previous MissingParents([X]) response, the handler
  returned WrongNumberOfParents(1, 0) — a fatal/irrecoverable error in the certifier. This permanently broke node1's ability to get node0's vote for round 174, leaving consensus stuck (only 2 votes out of 3
  needed).

  Why it triggered: Node2 restarted after a 70s delay, joining consensus at round 62 (too far behind to vote). Node1 needed votes from nodes 0, 1(self), and 3 for quorum. Node0 had requested a missing parent cert
  from node1 on the first vote attempt. Node1's certifier couldn't find it in the store at that moment (race condition) and exited. On re-proposal, node1 sent 0 parents → WrongNumberOfParents(1, 0) → irrecoverable
  → deadlocked.

  The fix (crates/consensus/primary/src/network/handler.rs): When last_response = MissingParents([X]) and the proposer sends empty parents (fresh request), re-issue MissingParents([X]) instead of
  WrongNumberOfParents. This lets the certifier receive the hint again on each re-proposal and eventually provide the cert once it's available in the store.

  Test updated (src/tests/primary_tests.rs): Phase 2 now asserts MissingParents is returned (correct new behavior), Phase 3 updated to match reality.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/644

Root cause identified and fixed.

  The bug: In handler.rs, when a proposer (node1) re-proposes a header with 0 parents (fresh certifier restart), but the handler's auth_last_vote cache had a previous MissingParents([X]) response, the handler
  returned WrongNumberOfParents(1, 0) — a fatal/irrecoverable error in the certifier. This permanently broke node1's ability to get node0's vote for round 174, leaving consensus stuck (only 2 votes out of 3
  needed).

  Why it triggered: Node2 restarted after a 70s delay, joining consensus at round 62 (too far behind to vote). Node1 needed votes from nodes 0, 1(self), and 3 for quorum. Node0 had requested a missing parent cert
  from node1 on the first vote attempt. Node1's certifier couldn't find it in the store at that moment (race condition) and exited. On re-proposal, node1 sent 0 parents → WrongNumberOfParents(1, 0) → irrecoverable
  → deadlocked.

  The fix (crates/consensus/primary/src/network/handler.rs): When last_response = MissingParents([X]) and the proposer sends empty parents (fresh request), re-issue MissingParents([X]) instead of
  WrongNumberOfParents. This lets the certifier receive the hint again on each re-proposal and eventually provide the cert once it's available in the store.

  Test updated (src/tests/primary_tests.rs): Phase 2 now asserts MissingParents is returned (correct new behavior), Phase 3 updated to match reality.

### DIFF
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -372,6 +372,21 @@ where
                     match last_response {
                         None | Some(PrimaryResponse::RecoverableError(_)) => {}
                         Some(PrimaryResponse::MissingParents(missing)) => {
+                            if parents.is_empty() {
+                                // Proposer retried with a fresh request (no parents provided).
+                                // Re-issue the missing parents request so the proposer knows
+                                // what to provide on the next attempt. This avoids a deadlock
+                                // where the proposer's certifier restarts (losing the missing
+                                // parent hint) and the cached state here causes a fatal
+                                // WrongNumberOfParents error on every subsequent attempt.
+                                *auth_last_vote = Some((
+                                    last_epoch,
+                                    last_round,
+                                    last_digest,
+                                    Some(PrimaryResponse::MissingParents(missing.clone())),
+                                ));
+                                return Ok(PrimaryResponse::MissingParents(missing));
+                            }
                             // A proper response to missing parents will include exactly the missing
                             // parents.
                             if parents.len() == missing.len() {
--- a/crates/consensus/primary/src/tests/primary_tests.rs
+++ b/crates/consensus/primary/src/tests/primary_tests.rs
@@ -342,28 +342,34 @@ async fn test_request_vote_has_missing_parents() {
     let received_missing: HashSet<_> = missing.into_iter().collect();
     assert_eq!(expected_missing, received_missing);
 
-    // TEST PHASE 2: Handler should not return additional unknown digests.
-    // No additional missing parents will be requested.
+    // TEST PHASE 2: Handler should re-issue the same MissingParents when proposer retries with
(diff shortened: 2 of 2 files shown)


crates/consensus/primary/src/network/handler.rs
```diff
@@ -372,6 +372,21 @@ where
...
+                                // where the proposer's certifier restarts (losing the missing
+                                // parent hint) and the cached state here causes a fatal
+                                // WrongNumberOfParents error on every subsequent attempt.
+                                *auth_last_vote = Some((
+                                    last_epoch,
+                                    last_round,
+                                    last_digest,
+                                    Some(PrimaryResponse::MissingParents(missing.clone())),
+                                ));
+                                return Ok(PrimaryResponse::MissingParents(missing));
```
grantkee, line 388: I think this is fine. The potential downsides are:
- requesting missing parents that a different peer already provided
- header's round may already be `TooOld`

The cost is bandwidth. However, I think it's likely best to err on the side of "caution" and return a missing parent that was requested to ensure it's present. Worse case scenario is slight inefficiency. Downstream tasks will GC old information

grantkee (approved):
LGTM

PR #648, opened 2026-05-04
AI suggested race fix.

### LINKED ISSUE #647: Restart test race.
Summary from Claude, this seems to be a legit fix and works.

All three fixes compile cleanly. Here's a summary of what was changed:
                                                                                                                                                                                                                        Fix 3 (crates/consensus/primary/src/network/mod.rs — request_consensus):
  Added a 10-second tokio::time::timeout per attempt around res.await. Previously this awaited a peer oneshot with no timeout, permanently freezing spawn_track_recent_consensus if a peer was unresponsive.
  Fix 1 (crates/state-sync/src/consensus.rs — spawn_fetch_consensus):                                                                                                                                                   After a successful epoch pack download (Ok(_)), reads the pack's final header via consensus_chain.consensus_header_by_number and updates consensus_bus.last_consensus_header(). This triggers
  spawn_stream_consensus_headers via the locally-available pack path, bypassing the network-dependent gossip path entirely.
  Fix 2 (crates/state-sync/src/lib.rs — spawn_stream_consensus_headers):                                                                                                                                                Restructured to call borrow_and_update() at startup before entering the loop, consuming the current watch value immediately. The processing logic runs once on the initial value before waiting for changed(). This
  handles the race where an epoch pack is downloaded and last_consensus_header() updated before a new epoch's spawn_stream_consensus_headers subscribes — previously, that update would be silently missed.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/647

### DIFF
--- a/crates/consensus/primary/src/network/mod.rs
+++ b/crates/consensus/primary/src/network/mod.rs
@@ -271,12 +271,20 @@ impl PrimaryNetworkHandle {
         number: u64,
         hash: BlockHash,
     ) -> NetworkResult<ConsensusHeader> {
+        const TIMEOUT: Duration = Duration::from_secs(10);
         let request = PrimaryRequest::ConsensusHeader { number, hash };
         // Try up to three times (from three peers) to get consensus.
         // This could be a lot more complicated but this KISS method should work fine.
         for _ in 0..3 {
             let res = self.handle.send_request_any(request.clone()).await?;
-            let res = res.await?;
+            let res = match tokio::time::timeout(TIMEOUT, res).await {
+                Ok(r) => r,
+                Err(_) => {
+                    tracing::warn!(target: "state-sync", ?number, "request_consensus timed out waiting for peer response");
+                    continue;
+                }
+            };
+            let res = res?;
             if let Ok(NetworkResponseMessage {
                 peer: _,
                 result: PrimaryResponse::ConsensusHeader(header),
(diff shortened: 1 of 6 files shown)


crates/consensus/primary/src/network/mod.rs
```diff
@@ -271,12 +271,20 @@ impl PrimaryNetworkHandle {
...
         let request = PrimaryRequest::ConsensusHeader { number, hash };
         // Try up to three times (from three peers) to get consensus.
         // This could be a lot more complicated but this KISS method should work fine.
         for _ in 0..3 {
             let res = self.handle.send_request_any(request.clone()).await?;
-            let res = res.await?;
+            let res = match tokio::time::timeout(TIMEOUT, res).await {
+                Ok(r) => r,
+                Err(_) => {
+                    tracing::warn!(target: "state-sync", ?number, "request_consensus timed out waiting for peer response");
```
grantkee: nit: `target: "primary::network"`

crates/state-sync/src/consensus.rs
```diff
@@ -208,7 +208,30 @@ pub async fn spawn_fetch_consensus(
                     tokio::select! {
                         result = network.request_epoch_pack(&epoch_record, &previous_epoch_record, &consensus_chain, Duration::from_secs(PACK_RECORD_TIMEOUT_SECS)) => {
                             match result {
-                                Ok(_) => break,
+                                Ok(_) => {
+                                    // After a successful pack download, signal spawn_stream_consensus_headers
+                                    // that locally-available blocks are ready. This unblocks streaming even
+                                    // when the gossip/network path (request_consensus) is slow or unresponsive.
+                                    if let Ok(Some(final_header)) = consensus_chain
```
grantkee: I think it would be worth logging the `Ok(None)` and `Err(e)` variants

grantkee (dismissed):
LGTM! Small logging suggestions that might help future debugging

grantkee (approved):

PR #656, opened 2026-05-05
Worker configs

grantkee:
- `RethEnv::get_worker_fee_configs(num_workers)` reads from canonical tip and rejects worker-count drift; `get_worker_fee_configs_at_block(hash)` reads at a specific header for replay paths.
- CLI: `--worker-fee-config WORKER_ID:STRATEGY:VALUE` (repeatable). Parser sorts by worker id, validates contiguous ids starting from 0, errors on bad format. Default when omitted: one EIP-1559 worker, 100M target gas.
- `compute_next_base_fee_eip1559` in `tn-types::gas_accumulator` — ±12.5% bound, `MIN_PROTOCOL_BASE_FEE` floor, saturating arithmetic. 11 unit tests cover at-target, over/under, overflow, min-fee floor, and partial utilization.
- Default epoch duration: 24h → 8h. `genesis_tests` and `staking` updated to match.
- `tn-contracts` submodule bumped to pull in the `WorkerConfigs` contract
- `GasAccumulator::new_with_rewards` constructor so callers can hand in a pre-built `RewardsCounter` instead of always starting from default.

precursor for #614 
unblocks #657


sstanfield (approved):
LGTM

PR #662, opened 2026-05-06
engineering: Updated test harness for certifier tests

ArieJones:
## Summary

Adds a structured test harness for the certifier and expands coverage from 5 ad-hoc tests to 21 fully independent, hermetic tests.

### `crates/consensus/primary/src/tests/certifier_tests.rs`

Refactored the 5 existing tests to use the new `CertifierContext` harness, and added 16 new tests covering previously untested paths:

**New tests (T1–T16):**

| # | Test | What it covers |
|---|------|----------------|
| T1 | `missing_parents_happy_path` | Certifier waits for missing parents then proceeds |
| T2 | `missing_parents_fake_digests` | Unknown digests that never arrive don't stall indefinitely |
| T3 | `missing_parents_store_miss` | Storage miss on parent lookup is handled gracefully |
| T4 | `transient_network_error_retries` | Vote request retries on transient network failures |
| T5 | `new_header_cancels_inflight` | Arriving new header cancels the in-flight proposal |
| T6 | `wrong_epoch_header_rejected` | Headers from wrong epoch are rejected before broadcasting |
| T7 | `non_cvv_node_skips_certifier` | Non-validator nodes don't spawn a certifier at all |
| T8 | `vote_wrong_author` | Vote signed by wrong authority is discarded |
| T9 | `vote_wrong_header_digest` | Vote referencing wrong digest is discarded |
| T10 | `vote_wrong_origin` | Vote with wrong origin field is discarded |
| T11 | `vote_epoch_mismatch` | Vote from wrong epoch is discarded |
| T12 | `vote_round_mismatch` | Vote from wrong round is discarded |
| T13 | `vote_unknown_authority` | Vote from unknown authority is discarded |
| T14 | `duplicate_vote_same_peer` | Second vote from same peer does not count toward quorum |
| T15 | `startup_republish_highest_cert` | On restart, the highest known certificate is re-published |
| T16 | `minimum_quorum_exactly_threshold` | Certificate forms at exactly `2f+1` votes, not before |

**Refactored existing tests** now use `CertifierContext` for consistent setup/teardown:
- `propose_header_to_form_certificate`
- `propose_header_failure`
- *(and the remaining 3 original tests)*

### `crates/test-utils-committee/src/committee.rs`

- Added `make_committee_fixture_with_n(n, new_db)` — a one-liner helper for building a `CommitteeFixture` with an arbitrary committee size. Useful for any future test that needs a non-default-sized committee.

### `.gitignore`

- Added `.github/.plans/` to exclude local planning files from version control.

## Testing

All 21 tests pass:

```
cargo test -p tn-primary --lib -- certifier
```

### DIFF
--- a/crates/consensus/primary/src/tests/certifier_tests.rs
+++ b/crates/consensus/primary/src/tests/certifier_tests.rs
@@ -8,138 +8,1181 @@ use crate::{
 use rand::{rngs::StdRng, SeedableRng};
 use std::{
     collections::HashMap,
-    num::{NonZero, NonZeroUsize},
+    num::NonZeroUsize,
 };
 use tn_network_libp2p::types::{NetworkCommand, NetworkHandle, NetworkResponseMessage};
 use tn_storage::mem_db::MemDatabase;
-use tn_test_utils_committee::CommitteeFixture;
-use tn_types::{BlsKeypair, BlsSigner, SignatureVerificationState, TnSender};
+use tn_test_utils_committee::{AuthorityFixture, CommitteeFixture};
+use tn_types::{BlsKeypair, BlsSigner, Epoch, Round, SignatureVerificationState, TnSender};
 use tokio::sync::mpsc;
 
-#[tokio::test(flavor = "current_thread")]
-async fn propose_header_to_form_certificate() {
-    let fixture = CommitteeFixture::builder(MemDatabase::default).randomize_ports(true).build();
-    let committee = fixture.committee();
-    let primary = fixture.authorities().last().unwrap();
-    let id = primary.id();
+// ===== Certifier test harness =====
+//
+// Mirrors the `TestTypes` / `create_test_types` pattern from `network_tests.rs`.
+// Each certifier test previously repeated ~30 lines of identical setup. These
+// types encapsulate that boilerplate so individual tests stay focused on the
+// scenario under test.
 
-    // Create a fake header.
-    let proposed_header = primary.header(&committee);
+/// Assembled context for certifier unit tests.
+///
+/// Holds the committee fixture, a live `Certifier` task (spawned on the last
+/// authority), the consensus bus for sending headers and receiving certs, and
+/// the raw network receiver for simulating peer vote responses.
+#[allow(dead_code)]
+struct CertifierContext {
+    /// All authorities + their configs.
+    fixture: CommitteeFixture<MemDatabase>,
+    /// Per-epoch consensus bus: send headers, read certificates.
+    consensus_bus: ConsensusBus,
+    /// Keeps the spawned `Certifier` task alive for the test duration.
+    task_manager: TaskManager,
+    /// Raw receiver of `NetworkCommand`s sent by the certifier to peers.
+    ///
+    /// Drain this in tests to intercept outgoing `Vote` requests and reply
+    /// with the desired [`VoteResponseConfig`].
+    network_rx: mpsc::Receiver<NetworkCommand<PrimaryRequest, PrimaryResponse>>,
+}
 
-    // Set up network handle- this is all we need to simulate then network for the certifier.
-    let (sender, mut network_rx) = mpsc::channel(100);
-    let network: NetworkHandle<PrimaryRequest, PrimaryResponse> = NetworkHandle::new(sender);
-
-    // Set up remote primaries responding with votes.
-    let mut peer_votes = HashMap::new();
-    for peer in fixture.authorities().filter(|a| a.id() != id) {
-        let name = peer.authority().protocol_key();
-        let id = peer.authority().id();
-        let vote = Vote::new(&proposed_header, id, peer.consensus_config().key_config());
-        peer_votes.insert(name, vote);
+#[allow(dead_code)]
+impl CertifierContext {
+    /// 4-authority context; `Certifier` spawned on the last authority.
+    fn new() -> Self {
+        let fixture =
+            CommitteeFixture::builder(MemDatabase::default).randomize_ports(true).build();
+        Self::from_fixture(fixture)
     }
 
-    let cb = ConsensusBus::new();
-    let mut rx_new_certificates = cb.subscribe_new_certificates();
-    // Spawn the core.
-    let task_manager = TaskManager::default();
-    let synchronizer =
-        StateSynchronizer::new(primary.consensus_config(), cb.clone(), task_manager.get_spawner());
+    /// `n`-authority context; `Certifier` spawned on the last authority.
+    fn with_size(n: usize) -> Self {
+        let fixture = CommitteeFixture::builder(MemDatabase::default)
+            .committee_size(NonZeroUsize::new(n).expect("committee size must be non-zero"))
+            .randomize_ports(true)
+            .build();
+        Self::from_fixture(fixture)
+    }
 
-    synchronizer.spawn(&task_manager);
-    Certifier::spawn(
-        primary.consensus_config(),
-        cb.clone(),
-        synchronizer,
-        network.clone().into(),
-        &task_manager,
-    );
+    /// Build from an already-constructed fixture (useful for custom committee params).
+    fn from_fixture(fixture: CommitteeFixture<MemDatabase>) -> Self {
+        let (sender, network_rx) = mpsc::channel(100);
+        let network: NetworkHandle<PrimaryRequest, PrimaryResponse> = NetworkHandle::new(sender);
+        let cb = ConsensusBus::new();
+        let task_manager = TaskManager::default();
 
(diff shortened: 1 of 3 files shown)


ArieJones:
Need to wait for PR #665 since it will solidify the toolchain and make this easier

grantkee:
Thank you for taking this on @ArieJones ! Please update with the latest `main` when you get a chance #665 is merged now

ArieJones:
@grantkee  This has been rebased to current main. No issues with the rebase.

crates/consensus/primary/src/tests/certifier_tests.rs
```diff
@@ -6,140 +6,1176 @@ use crate::{
...
+        .round(base.round())
+        .epoch(wrong_epoch)
+        .parents(base.parents().clone())
+        .created_at(*base.created_at())
+        .build();
+
+    cx.consensus_bus.headers().send(wrong_epoch_header).await.unwrap();
+
+    // No vote requests should arrive (certifier rejects before sending requests).
+    assert!(
```
grantkee, line 562: I think it's safe to remove the `|| matches!....` and just use the first half of the assertion

 Both halves consume from `network_rx`. If a vote request arrives within 2s, the LHS is `false` (timeout returned `Ok`). The RHS then waits an additional 100ms for a **second** message and passes if none arrives, so the test passes despite a vote request having leaked through. Under the intended-success case (no request), the LHS already succeeds and the RHS is redundant. The OR masks the failure case

crates/test-utils-committee/src/committee.rs
```diff
@@ -180,3 +183,22 @@ impl<DB: Database> CommitteeFixture<DB> {
...
+/// One-liner to create a [`CommitteeFixture`] with `n` authorities.
+///
+/// Randomizes ports so tests don't collide. Panics if `n == 0`.
+/// Useful when a test needs a non-default committee size without repeating the full builder chain.
+///
+/// # Example
+/// ```ignore
+/// let fixture = make_committee_fixture_with_n(7, MemDatabase::default);
+/// ```
+pub fn make_committee_fixture_with_n<DB: Database, F: Fn() -> DB>(
```
grantkee: Please remove this from the PR. It isn't used anywhere. The `CommitteeFixture::builder().committee_size()` method is the recommended solution for tests

crates/consensus/primary/src/tests/certifier_tests.rs
```diff
@@ -6,140 +6,1176 @@ use crate::{
...
+// Each certifier test previously repeated ~30 lines of identical setup. These
+// types encapsulate that boilerplate so individual tests stay focused on the
+// scenario under test.
+
+/// Assembled context for certifier unit tests.
+///
+/// Holds the committee fixture, a live `Certifier` task (spawned on the last
+/// authority), the consensus bus for sending headers and receiving certs, and
+/// the raw network receiver for simulating peer vote responses.
+#[allow(dead_code)]
```
grantkee: Please remove all dead code

crates/consensus/primary/src/tests/certifier_tests.rs
```diff
@@ -6,140 +6,1176 @@ use crate::{
...
+}
+
+// ---------------------------------------------------------------------------
+// T14: duplicate_vote_same_peer
+// The same peer sends two identical votes for the same header.  VotesAggregator
+// deduplicates, so the cert is assembled from the proposer's self-vote plus the
+// unique votes of the remaining peers — a full quorum still forms.
+// ---------------------------------------------------------------------------
+#[tokio::test(flavor = "current_thread")]
+async fn duplicate_vote_same_peer() {
```
grantkee: I see `duplicate_peer_id` is assigned at L1010 and never read. Only one `PrimaryResponse::Vote` is sent per peer (L999-1001), so the certifier's `VotesAggregator` only ever sees unique authors ( `authorities_seen.insert(...)` at `aggregators/votes.rs:43` never returns `false`). The test logic follows happy-path certification which contradicts the doc-comment at L967-971 

The comment at L1004-1008 is correct, you can't easily re-inject into the certifier's `tx_votes` channel because it's created inside `propose_header` at `certifier.rs:278`, cloned into each peer task, then dropped. 

To work around this, I'd test `VotesAggregator` directly. It's `pub(crate)` and reachable from for tests. The assertion can also be sharper (you can match `DagError::AuthorityReuse`), and it removes the indirection through the network mock.

Suggested replacement (sketch — adjust types/imports to match the file):

```rust
#[tokio::test(flavor = "current_thread")]
async fn duplicate_vote_rejected_quorum_still_forms() {
    // Build committee + a header authored by the proposer, exactly as other tests do.
    let cx = CertifierContext::new();
    let header = cx.proposer_header();
    let committee = cx.committee();

    let mut agg = VotesAggregator::new();

    // Pick one non-proposer peer to duplicate.
    let mut peers = cx.fixture.authorities().filter(|a| a.id() != cx.proposer().id());
    let dup = peers.next().expect("at least one non-proposer");

    // First vote from `dup` is accepted (no quorum yet).
    let v1 = Vote::new(&header, dup.id(), dup.consensus_config().key_config());
    assert!(matches!(agg.append(v1, &committee, &header), Ok(None)));

    // Second vote from the SAME author is rejected with AuthorityReuse.
    let v2 = Vote::new(&header, dup.id(), dup.consensus_config().key_config());
    assert!(matches!(
        agg.append(v2, &committee, &header),
        Err(DagError::AuthorityReuse(_))
    ));

    // Remaining peers vote — quorum still forms despite the rejected duplicate.
    let mut cert = None;
    for peer in peers {
        let v = Vote::new(&header, peer.id(), peer.consensus_config().key_config());
        cert = agg.append(v, &committee, &header).expect("valid vote");
        if cert.is_some() {
            break;
        }
    }
    let cert = cert.expect("quorum reached from unique voters");
    assert_eq!(cert.header().digest(), header.digest());
}
```

This verifies the two properties the original doc-comment names: (1) a second vote from a seen author yields `DagError::AuthorityReuse` at `aggregators/votes.rs:42-45`, and (2) a quorum still forms from the remaining unique voters. No `network_rx` plumbing, no reliance on the certifier silently swallowing errors. Drop the `// T14` test as-is in favor of this; the network-loop coverage is already provided by `propose_header_to_form_certificate`.

crates/consensus/primary/src/tests/certifier_tests.rs
```diff
@@ -6,140 +6,1176 @@ use crate::{
...
+        .filter(|a| a.id() != proposer_id)
+        .map(|a| (a.id(), VoteResponseConfig::FatalNetworkError))
+        .collect();
+
+    // Propose header and verify we get no certificate back.
+    cx.consensus_bus.headers().send(header.clone()).await.unwrap();
+    drive_vote_requests(&mut cx.network_rx, &cx.fixture, &proposer_id, &header, &peer_configs)
+        .await;
+
+    if let Ok(result) = tokio::time::timeout(Duration::from_secs(5), cert_rx.recv()).await {
```
grantkee: In general, I'd like to avoid testing "nothing happened" with arbitrary time limits. `propose_header` returns `Err(DagError::CouldNotFormCertificate)` almost immediately, so the 5s is dead time. It also wouldn't pick up if there was a legitimate hang from a race condition. Maybe this would be better to test directly on `VotesAggregator`?

grantkee (changes requested):
Thanks for looking into improved test harness @ArieJones . These tests are critical, and there's definitely room to improve them. I have a few requests for these changes. LMK if you have any questions

crates/consensus/primary/src/tests/certifier_tests.rs
ArieJones: updated this to remove the RHS

crates/test-utils-committee/src/committee.rs
ArieJones: gone

crates/consensus/primary/src/tests/certifier_tests.rs
ArieJones: gone

crates/consensus/primary/src/tests/certifier_tests.rs
ArieJones: I updated this slightly to have a more rigid deterministic check that I thought fit. Removed the time limit. I had moved the duplicate vote logic to go against VoteAggregator but I felt this was more of an orchestration related behavior.

crates/consensus/primary/src/tests/certifier_tests.rs
ArieJones: moved to test against VoteAggregator directly

PR #668, opened 2026-05-07
Store validator certs in a pack file for diagnostics.

### LINKED ISSUE #664: Store certificates for validators for diagnostics.
Need to store validator certs in case we have to do forensics on the network.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/664


crates/consensus/primary/src/consensus/state.rs
```diff
@@ -376,14 +387,23 @@ impl<DB: Database> Consensus<DB> {
                 return Ok(());
             }
         }
+        if let Some(certificate_pack) = &self.certificate_pack {
+            if let Err(e) = certificate_pack.save(certificate.clone()).await {
+                tracing::error!(target: "telcoin::consensus_state", ?e, "Failed to save certificate to cert pack file");
```
grantkee, line 396: I fully support saving certificates, but I think this should happen after bullshark or as a separate background task to prevent db from blocking a critical path on `await`

grantkee (dismissed):
LGTM!

grantkee (approved):

PR #665, opened 2026-05-07
centralize rust version - update gh action to fix ci for public PRs

grantkee:
- fix CI
- centralized source for rust versions to prevent this in the future

PR #670, opened 2026-05-07
fix: correct transposed UDP port values in .env.example

blu3chipsx:
PRIMARY_UDP_PORT and WORKER_UDP_PORT values were swapped.

PRIMARY_UDP_PORT corrected to 49590
WORKER_UDP_PORT corrected to 49594


sstanfield (approved):
Good find.  Not even sure where that file came from or why it is "hidden".  I assume because it is supposed to be .env.

grantkee:
@blu3chipsx please merge upstream `main` into your branch. We made a change to fix CI https://github.com/Telcoin-Association/telcoin-network/pull/665

PR #674, opened 2026-05-08
Fix observers getting stuck on epoch 0.

### LINKED ISSUE #673: Observers that start behind can't execute epoch 0.
The "dummy" initial record was not being written soon enough for all cases and the initial consensus header hash for epoch 0 was not calculated correctly when verifying a pack file for epoch 0 (correct file, incorrect validation).

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/673

Note, this accidentally picked up a small fix when checking the last vote and a spot that missed clearing cache tables on a new epoch.  Both are small and should be valid fixes (non-breaking).


grantkee (approved):
LGTM! Thank you for the quick turnaround on this

PR #676, opened 2026-05-11
Regression tests for observer epoch-0 sync + stale prior-epoch vote

### LINKED ISSUE #675: Add regression tests for observer epoch-0 sync and stale prior-epoch vote handling
## Problem

PR #674 ("Fix observers getting stuck on epoch 0") landed two correctness fixes that currently have no test coverage:

1. **Pack-import parent-hash convention.** When an observer joins after epoch 0 has closed and the validators have certified an epoch-0 record, the only path to catch up runs through `consensus_pack::stream_import`. Before #674, the consumer initialised the chain's first `parent_digest` from `previous_epoch.final_consensus.hash`, which for the synthesised epoch-0 predecessor is `B256::ZERO`. The producer side, in contrast, writes the genesis `parent_hash` as `ConsensusHeader::default().digest()` — a non-zero value derived from a default `CommittedSubDag`. The mismatch surfaced as `PackError::InvalidConsensusChain` and observers never advanced past genesis. The fix matches the convention on both sides, but no end-to-end test forces an observer down the post-epoch-close pack-import path, so a future regression to either side of the convention would be caught only in production.

2. **Stale prior-epoch `VoteInfo` rows.** PR #674 added a defensive branch in the vote handler that allows a vote on a current-epoch header even when the `Votes` table still holds an entry from a prior epoch. The `Votes` table is supposed to be cleared on every epoch close (PR #674 also closed the last gap in that coverage on the replay-driven close path), so the leaked-row scenario is theoretically unreachable — but if any future close path ever skips the clear, the defensive branch is what keeps consensus making progress. With no test pinning the branch in place, it is at risk of being removed as "dead code" in a later cleanup.

Both gaps were called out as recommended follow-ups by the security review of #674 (`reports/security-eval-issue-673.md`, finding F6).

## Solution

Introduce two regression tests, one at the e2e layer and one at the primary-handler unit-test layer, each pinning down exactly the invariant that #674 restored.

The e2e test should bring up a four-validator local testnet with a short epoch duration, wait for epoch 0 to close on the validator side, wait for the validators to certify and persist the epoch-0 record, and only then start a fresh observer from genesis. Because the epoch-0 pack is already on disk by the time the observer comes up, any catch-up path must go through `consensus_pack::stream_import` — exactly the path the parent-hash convention bug broke. The test asserts that the observer reaches the validator's consensus header height, that the observer's reconstructed epoch-0 record self-verifies and matches the validator's record byte-for-byte, and that a transaction submitted to the observer is confirmed by a validator (a liveness check that catches "synced but serving stale state" failure modes). To keep the test inside a normal CI budget, the e2e harness should expose an opt-in epoch-duration override so the test does not have to wait the genesis-default epoch length.

The unit test should construct a `RequestHandler` whose committee reports a non-zero epoch, seed the `Votes` table with a stale entry from the prior epoch keyed under the authority that will submit the current-epoch header, and submit a current-epoch header from that authority. The handler should accept the vote, return `PrimaryResponse::Vote`, and overwrite the stale row with a current-epoch entry. To make this construction possible, the `CommitteeFixture` builder should thread its `epoch` field through to `Committee::new_for_test` so the fixture's reported epoch matches the one the handler validates against.

To keep the new e2e test colocated with related sync tests but separate from the restart suite, the shared process-management, key-derivation, balance-polling, and transaction-submission helpers currently inlined in `restarts.rs` should move into the existing `common.rs` test module so both `restarts.rs` and a new `sync.rs` module can consume them without duplication.

grantkee:
- Adds `test_observer_pack_imports_after_epoch_close` (e2e): starts 4 validators with a 10s epoch duration, waits for epoch 0 to close and certify on the validator side, then starts a fresh observer so the only catch-up path is `consensus_pack::stream_import`. Asserts the observer reaches the validator's consensus height, returns a self-verifying epoch-0 record that matches the validator's byte-for-byte, and confirms a transaction submitted to the observer — pinning down the parent-hash convention fix from #674 (Finding F2).
- Adds `test_vote_succeeds_with_stale_prior_epoch_vote_info` (primary unit): seeds the `Votes` table with an epoch-0 entry, submits an epoch-1 header from the same author, and asserts the handler returns `PrimaryResponse::Vote` and overwrites the stale row — pinning down the `header.epoch() > vote_info.epoch()` defensive branch from #674 (Findings F1 / F6).
- Threads `CommitteeFixture::epoch` through to `Committee::new_for_test` so fixtures actually report the requested epoch (was hardcoded to `0`); needed by the new vote test.
- Adds `config_local_testnet_with_epoch_duration` so e2e tests can override the genesis 8-hour epoch default; original `config_local_testnet` is preserved as a thin wrapper.
- Moves shared process / key / balance / tx helpers from `restarts.rs` into `common.rs` so both `restarts.rs` and the new `sync.rs` module can use them. No behaviour change in `restarts.rs`.

closes #675

### DIFF
--- a/crates/consensus/primary/src/tests/network_tests.rs
+++ b/crates/consensus/primary/src/tests/network_tests.rs
@@ -3,7 +3,7 @@
 use crate::{
     error::PrimaryNetworkError,
     network::{
-        message::{ConsensusResult, PrimaryGossip},
+        message::{ConsensusResult, PrimaryGossip, PrimaryResponse},
         MissingCertificatesRequest, PendingEpochStream, RequestHandler,
         MAX_CONCURRENT_EPOCH_STREAMS, PENDING_REQUEST_TIMEOUT,
     },
@@ -20,12 +20,12 @@ use std::{
 use tempfile::TempDir;
 use tn_config::Parameters;
 use tn_network_libp2p::{GossipMessage, TopicHash};
-use tn_storage::{consensus::ConsensusChain, mem_db::MemDatabase};
+use tn_storage::{consensus::ConsensusChain, mem_db::MemDatabase, tables::Votes};
 use tn_test_utils_committee::CommitteeFixture;
 use tn_types::{
     error::HeaderError, now, AuthorityIdentifier, BlockHash, BlockHeader, BlockNumHash,
-    BlsPublicKey, Certificate, EpochVote, ExecHeader, Hash as _, HeaderDigest, SealedHeader,
-    TaskManager, B256,
+    BlsPublicKey, Certificate, Database, Epoch, EpochVote, ExecHeader, Hash as _, HeaderDigest,
+    SealedHeader, TaskManager, VoteDigest, VoteInfo, B256,
 };
 use tracing::debug;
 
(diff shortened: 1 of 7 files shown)


sstanfield (dismissed):
This all looks great.

sstanfield (approved):

PR #681, opened 2026-05-13
Always have bootstraps and dont crash on a network event error.

### LINKED ISSUE #680: Missing bootstrap servers and network event errors stopped node.
Need to always include the bootstrap servers for any committee even if not epoch 0.

Also, do not shut down the node for a random network error.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/680


crates/network-libp2p/src/consensus.rs
```diff
@@ -429,13 +429,13 @@ where
 
         loop {
             tokio::select! {
-                event = self.swarm.select_next_some() => self.process_event(event).await.inspect_err(|e| {
-                    error!(target: "network", ?e, "network event error")
-                })?,
+                event = self.swarm.select_next_some() => if let Err(e) = self.process_event(event).await {
```
grantkee, line 447: If `Err(AllListenersClosed)`, the node has no network connectivity and the manager doesn't check/retry. I think this error should probably be treated as fatal and shut the node down

It will eventually be obvious the node has no connectivity but this might delay offchain alerts if the node can't listen to any network traffic

grantkee (dismissed):
Good work. I think there's a small improvement for node operators, but not a blocker

grantkee (approved):
Thanks @sstanfield ! Great work

PR #683, opened 2026-05-14
Kad improvements

### LINKED ISSUE #682: kad records accumulate without local eviction, filling the store and rejecting new puts
## Problem

The validator-discovery DHT relies on each node publishing a `NodeRecord` keyed by its BLS public key so peers can resolve consensus identities to libp2p addresses. Today these records are written to the local kad store with `expires: None`, meaning the store treats them as permanent. libp2p's `set_record_ttl` drives republication cadence but does not evict rows from the underlying `RecordStore` — that responsibility belongs to the store implementation, and ours has none.

Two amplifiers make this worse in practice. First, the node pushes its full record to every newly connected peer via `publish_our_data_to_peer`, with no deduplication — peers that flap (and banned peers that re-reach the `PeerConnected` event, as observed in adiri testnet logs) cause repeated direct pushes. Second, every `ExternalAddrConfirmed` event re-publishes our record to the DHT, even though the node's external address is static and the value being published is identical. Each redundant publish lands as a new row in some peer's kad store.

The combination is that local kad stores fill up over time with rows for peers that have long since departed, rotated keys, or are otherwise stale, and once `num_records` reaches `max_records` every new put is rejected with `MaxRecords`. Validator discovery quietly degrades: fresh records from new or recovering peers cannot be stored, while ancient rows from dead peers remain. The same pattern affects the provider-record table.

There is also a related defense-in-depth gap: banned peers occasionally surface through the `PeerConnected` swarm event despite the peer-manager's connection guards. When that happens today, the banned peer is added to the kad routing table and gossipsub, which triggers a redial loop and additional record traffic before the ban is re-enforced.

## Solution

The local kad store should drive its own eviction off a real TTL carried on every record, decoupled from libp2p's republication timer. Published `NodeRecord`s should be written with `expires` set to `now + kad_record_ttl`, and both the TTL and the republication interval should move from hardcoded constants in `ConsensusNetwork::new` into `LibP2pConfig` so operators can tune them. The TTL must remain strictly greater than the publication interval, otherwise records expire between refreshes.

The `KadStore` `RecordStore` impl should filter expired rows on every read path (`get`, `records`, `providers`) so consumers never observe a stale value, and should sweep expired rows opportunistically on `put` / `add_provider` when the store is at capacity. This keeps the store self-healing without adding a background task: a saturated store of expired rows will free space the moment a new put arrives, rather than rejecting it. Provider records, which are stored as `Vec<ProviderRecord>` per key, should drop a key only when every provider in the vector has expired.

The publish path should be tightened to remove redundant traffic. The node should track which peers it has already pushed its record to for the lifetime of the process and skip the direct push on reconnects — the remote peer already has the record in its persistent store. The `ExternalAddrConfirmed` handler should stop triggering a fresh DHT publish and warn operators if their external ip differs from the one they publish. The protocol assumes a static external address, so a mismatch is a configuration warning, not a reason to republish. Finally, the `PeerConnected` handler should re-check the peer-manager's ban state and refuse to register the connection with kad and gossipsub if the peer is banned, as a defense-in-depth backstop against the upstream guard.

Failed `PutRecord` queries log as errors today, but the failure is routine when peers are temporarily unreachable and floods logs without conveying actionable information. It should drop to `debug` level.

grantkee:
- Set `expires = now + kad_record_ttl` on our own `NodeRecord` so the local store can actually evict it; libp2p's record TTL only drives republish, not eviction.
- Move `kad_record_ttl` (48h) and `kad_publication_interval` (12h) from hardcoded `ConsensusNetwork::new` constants into `LibP2pConfig` so operators can tune them.
- `KadStore` now filters expired records on `get` / `records` / `providers` and opportunistically sweeps expired rows on `put` / `add_provider` when at capacity, so a saturated store of stale rows self-heals on the next put instead of returning `MaxRecords`. Provider keys drop only when all providers under the key have expired.
- Track `published_to_peers` per process and skip the direct record push on reconnects — the remote already has it in its persistent kad store. Cuts amplification from flapping/banned peers seen in adiri.
- `ExternalAddrConfirmed` no longer re-publishes; the external address is static, so a mismatch just warns.
- `PeerConnected` now re-checks `peer_manager.peer_banned` and refuses to register banned peers with kad/gossipsub, backstopping the upstream peer-manager guard.
- `PutRecord` failures drop from `error!` to `debug!` (routine when peers are unreachable, was flooding logs).
- Two new unit tests cover (1) expired records being filtered from reads and (2) `put` evicting expired rows when the store is full. One existing test loosened to compare stable fields only, because `Instant`-encoded `expires` doesn't survive the `SystemTime` ↔ `Instant` round-trip exactly.

closes #682

### DIFF
--- a/crates/config/src/network.rs
+++ b/crates/config/src/network.rs
@@ -113,6 +113,15 @@ pub struct LibP2pConfig {
     pub px_disconnect_timeout: Duration,
     /// The k-bucket size for kademlia.
     pub k_bucket_size: NonZeroUsize,
+    /// The TTL applied to kademlia records — both the libp2p record TTL and the
+    /// local store's `expires` timestamp. Drives eviction of records that are
+    /// never refreshed. Also used for provider record TTL.
+    pub kad_record_ttl: Duration,
+    /// How often this node republishes its own kademlia records.
+    ///
+    /// Must be < `kad_record_ttl`, otherwise records expire before they are
+    /// refreshed.
+    pub kad_publication_interval: Duration,
 }
 
 impl LibP2pConfig {
@@ -165,6 +174,8 @@ impl Default for LibP2pConfig {
             max_px_disconnects: 10,
             px_disconnect_timeout: Duration::from_secs(3),
             k_bucket_size: K_VALUE,
+            kad_record_ttl: Duration::from_secs(48 * 60 * 60),
+            kad_publication_interval: Duration::from_secs(12 * 60 * 60),
         }
     }
 }
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -31,7 +31,7 @@ use libp2p::{
 };
 use std::{
     collections::{HashMap, HashSet, VecDeque},
-    time::Duration,
+    time::{Duration, Instant},
 };
 use tn_config::{KeyConfig, LibP2pConfig, NetworkConfig, PeerConfig};
 use tn_types::{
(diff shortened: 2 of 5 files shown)


sstanfield (dismissed):
These look like great improvements.

sstanfield (approved):

PR #684, opened 2026-05-14
Improve sync.

### LINKED ISSUE #679: Fix observers.
Observers are having issue currently.  Get them fixed.
This will include:
- Auditing the tasks that run as an observer.
- Fixing the current epoch backfill.

Note that pack sync seems to be working but things go south once caught up to the current epoch.

sstanfield:
This is primarily cleans up the sync code and tasks.

Closes https://github.com/Telcoin-Association/telcoin-network/issues/679


crates/state-sync/src/consensus.rs
```diff
@@ -75,112 +83,79 @@ async fn get_consensus_header<DB: TNDatabase>(
...
-                        consensus_bus
-                            .request_epoch_pack_file(previous_epoch_record, epoch_record.clone())
-                            .await;
-                    }
-                    previous_epoch_record = epoch_record;
-                } else {
-                    break;
+    // If we still have epochs to fetch then add to the queue until we are out of epoch records.
+    // For epoch 0: `saturating_sub(1)` yields 0, so record_by_epoch(0) would return the
+    // *real* epoch-0 record- use a synthetic epoch record instead.
```
grantkee, line 105: nit: restore the comment explaining the rationale behind this decision:

```rust
        // If we still have epochs to fetch then add to the queue until we are out of epoch records.
        // For epoch 0: `saturating_sub(1)` yields 0, so record_by_epoch(0) would return the
        // *real* epoch-0 record whose final_state is the end-of-epoch execution state.
        // But epoch 0's pack was written with genesis defaults as its genesis_exec_state
        // (BlockNumHash::default()), so passing the real epoch-0 record as `previous_epoch`
        // to stream_import would cause verify_epoch_meta to reject the pack.
        // Fix: synthesise a sentinel record that has the genesis defaults for final_state /
        // final_consensus but copies the real epoch-0 committee into next_committee so the
        // committee check in verify_epoch_meta still passes.
```

crates/state-sync/src/consensus.rs
```diff
@@ -294,3 +286,198 @@ pub async fn spawn_fetch_consensus(
...
+                    &consensus_chain_clone,
+                    epoch,
+                    0,
+                )
+                .await;
+                tasks_clone.fetch_sub(1, Ordering::Relaxed);
+                Ok(())
+            },
+        );
+        tasks.store(1, Ordering::Relaxed);
```
grantkee: This should be called before the `spawn_task` spawns to eliminate any possible race condition

crates/state-sync/src/consensus.rs
```diff
@@ -294,3 +286,198 @@ pub async fn spawn_fetch_consensus(
...
+                        &consensus_chain_clone,
+                        min_epoch,
+                        end_number,
+                    )
+                    .await;
+                    tasks_clone.fetch_sub(1, Ordering::Relaxed);
+                    Ok(())
+                },
+            );
+            tasks.fetch_add(1, Ordering::Relaxed);
```
grantkee: Same as above - `fetch_add` before spawn to eliminate any possible race condition

crates/state-sync/src/consensus.rs
```diff
@@ -294,3 +286,198 @@ pub async fn spawn_fetch_consensus(
...
+    db: DB,
+    consensus_bus: ConsensusBusApp,
+    network: PrimaryNetworkHandle,
+    consensus_chain: ConsensusChain,
+    rx_shutdown: Noticer,
+    task_spawner: TaskSpawner,
+) {
+    // Get the epoch of our last executed consensus.
+    let mut current_fetch_epoch = consensus_chain.latest_consensus_epoch();
+    let mut rx_consensus_request = consensus_bus.subscribe_consensus_request_queue();
```
grantkee: nit: this should be called outside the async context and passed as an arg to eliminate any race condition. It would possibly miss some gossip, but best practice based (as mentioned in  `consensus_bus:L91` comment)

grantkee (commented):
Overall, this looks great. I think I found a few potential race conditions that we should discuss/address before merging

crates/state-sync/src/consensus.rs
sstanfield: The old comment was not totally accurate anymore.  That is why I removed it.  Part of the fix got pushed into the pack verifier, this was not enough and epoch 0 was complex.  So I don't think adding this extra verbiage back is a good idea- it is misleading.  If someone is in the code that matters now it should have the fuller comment there.

crates/state-sync/src/consensus.rs
sstanfield: I'll move this, I guess that is more technically correct, but there is not a meaningful race here.  The int is signed so going negative for a beat is no big deal.  Also this code is just an arbitrary governor to keep the consensus requests in check if the fetches pile up.  Having one more (or less) fetch task won't matter.  I am not convinced it will even change anything, it grabs the existing value upfront so if this codepath fires in parallel the "race" is there anyway.  But I do not think it is worth spending time on this (well introducing a full lock that will be slower in the general case) since it is crude and arbitrary and will serve its purpose either way.  Also, we should not be able to get validated consensus in parallel anyway but worst case our number of tasks is a little off but the throttle will still work.  I am adding a comment to document that even if racey this works as intended.

grantkee (dismissed):
LGTM!

grantkee (approved):

PR #686, opened 2026-05-15
use serde default to deserialize missing fields on network config

### LINKED ISSUE #685: Network config deserialization breaks on field additions
## Problem

`NetworkConfig` and its sub-structs (`LibP2pConfig`, `SyncConfig`, `QuicConfig`, `PeerConfig`, `ScoreConfig`) in `crates/config/src/network.rs` are loaded from a YAML file on disk via `serde_yaml::from_str` inside `ConfigTrait::load_from_path_or_default`. None of these structs opt into serde's default-on-missing behavior, so every field listed in the Rust definition must be present in the YAML — if one is missing, parsing fails and the node refuses to start.

This turns any field addition into a breaking upgrade. The most recent example is #683 ("Kad improvements"), which added `kad_record_ttl` and `kad_publication_interval` to `LibP2pConfig`. Nodes that had previously written a `network-config` file now crash on startup against that file with a serde error, even though the new fields have sensible defaults defined on the struct. The fix today is for an operator to delete or hand-edit the on-disk config — neither is acceptable for a validator network where downtime is costly and config files may be operator-customized.

The fragility is structural: every future field addition has the same failure mode unless the contributor remembers to do something about it. There is no test that catches the regression.

## Solution

The network config should treat missing fields as a request for the default value, matching the behavior callers already get from `T::default()` when the file is absent entirely. Serde's container-level `#[serde(default)]` does exactly this when the struct already derives or implements `Default` — any field omitted from the input is filled in from `Default::default()`. All six config structs in `network.rs` already have a `Default` impl, so the change is one attribute per struct and covers future fields automatically.

Applying it to `NetworkConfig` itself also protects against entire sub-sections being absent (an older config with no `peer_config:` key at all), which matters as new sub-structs get added under the top-level container.

A regression test should pin this contract in place: deserializing an empty YAML mapping must produce `NetworkConfig::default()`, and stripping the kad fields from a serialized default must still parse cleanly. Any future contributor who adds a field to one of these structs without preserving the default-on-missing behavior will see the test fail before the change reaches a release.

Out of scope: rewriting the on-disk file with the new defaults on read, and applying the same treatment to other config files (`parameters.yaml`, `committee.yaml`, key configs). Those are separate concerns and broadening the change would mix unrelated risk into the patch.

grantkee:
- use serde default for network config
- add regression test for network config

closes #685

### DIFF
--- a/crates/config/src/network.rs
+++ b/crates/config/src/network.rs
@@ -10,6 +10,7 @@ impl ConfigTrait for NetworkConfig {}
 
 /// The container for all network configurations.
 #[derive(Serialize, Deserialize, Debug, Default, Clone)]
+#[serde(default)]
 pub struct NetworkConfig {
     /// The configurations for libp2p library.
     ///
@@ -78,6 +79,7 @@ impl NetworkConfig {
 
 /// Configurations for libp2p library.
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(default)]
 pub struct LibP2pConfig {
     /// The supported inbound/outbound protocols for request/response behavior.
     /// - ex) "/telcoin-network/mainnet/0.0.1"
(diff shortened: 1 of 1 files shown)


sstanfield (approved):
Looks good- I assume we are fine with all these structs having defaults for serde now?  I think so, seems worth it for these issues but will be a subtle config change.

PR #688, opened 2026-05-15
Close banned-peer loopholes in peer manager

### LINKED ISSUE #687: Banned peers can still be dialed through gaps in the peer-manager ban predicate
## Problem

The libp2p peer manager treats banned peers as un-dialable, but the predicate that gates outbound dials does not consistently reflect the union of all ban states. Several code paths can leave a peer in a "banned" state by one measure (reputation, IP, or the `temporarily_banned` LRU) while another measure reports the peer as eligible to dial. Because Kademlia drives its own dial attempts through the peer manager, any disagreement between these measures is enough for Kad to keep re-dialing a peer the rest of the system considers banned.

Three concrete gaps allow this:

- The unified `can_dial` check consulted only the connection-status table (`peers.can_dial`) plus the `temporarily_banned` LRU. When a peer's reputation was `Banned` but its `ConnectionStatus` had been flipped back to `Disconnected` via the `Unbanned` status transition, neither check fired and `can_dial` returned true.
- `PeerAction::Ban`, the action emitted when reputation transitions to `Banned`, did not insert the peer into `temporarily_banned`. The sibling actions `Disconnect` and `DisconnectWithPX` did. This left a window in which the dial-time veto did not match the reputation-level decision.
- `handle_pending_outbound_connection` short-circuited when an in-flight dial was already registered, returning success without re-evaluating ban state. A peer that became banned after the dial was queued would still complete the connection, and on the subsequent disconnect Kad would re-dial, producing tight retry loops and log flapping.

The observable impact is log spam from repeated dial/disconnect cycles, wasted connection slots and bandwidth, and a weakening of the ban contract — banned peers can still consume resources and influence routing tables even after the manager has decided to exclude them.

## Solution

The peer manager should treat "is this peer banned" as a single predicate evaluated everywhere a dial decision is made, rather than open-coding subsets of the ban state at each call site.

grantkee:
- `can_dial` now consults the full `peer_banned` predicate (reputation + IP + temp-ban LRU) before the status-based check, so reputation-banned peers in `Disconnected` status are no longer dialable.
- `PeerAction::Ban` arms `temporarily_banned`, matching `Disconnect` / `DisconnectWithPX` and closing the window where the dial-time veto disagreed with the reputation decision.
- `handle_pending_outbound_connection` re-checks the ban predicate on the in-flight short-circuit path; peers banned mid-dial are denied at the pending stage instead of producing a retry/log-flap loop after the connection establishes.
- Three regression tests cover each gap (rep-banned + Disconnected, `PeerAction::Ban` arming the LRU, pending outbound denial while in `Dialing` state).

closes #687

### DIFF
--- a/crates/network-libp2p/src/peers/behavior.rs
+++ b/crates/network-libp2p/src/peers/behavior.rs
@@ -34,6 +34,12 @@ impl NetworkBehaviour for PeerManager {
             // PeerManager and Kad may initiate dials
             // intercept kad dial attempts, sanitize, and register
             if self.dial_attempt_already_registered(&peer_id) {
+                if self.peer_banned(&peer_id) {
+                    debug!(target: "peer-manager", ?peer_id, "rejecting in-flight dial for banned peer");
+                    return Err(ConnectionDenied::new(
+                        "Outbound connection to banned peer".to_string(),
+                    ));
+                }
                 // peer manager has already approved this dial attempt
                 return Ok(vec![]);
             }
--- a/crates/network-libp2p/src/peers/manager.rs
+++ b/crates/network-libp2p/src/peers/manager.rs
@@ -288,6 +288,7 @@ impl PeerManager {
         match action {
             PeerAction::Ban(ip_addrs) => {
                 debug!(target: "peer-manager", ?peer_id, ?ip_addrs, "reputation update results in ban");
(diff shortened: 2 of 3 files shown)


sstanfield (approved):
LGTM

PR #695, opened 2026-05-20
Add rpc tn_info.

sstanfield:
Add tn_info to expose a lot of the nodes network and identity info.

### DIFF
--- a/crates/execution/tn-rpc/src/lib.rs
+++ b/crates/execution/tn-rpc/src/lib.rs
@@ -11,7 +11,7 @@ use tn_types::{
     EpochCertificate, EpochRecord, Multiaddr, NetworkPublicKey,
 };
 
-/// Contain the nodes identifying info to provide over RPC.
+/// Contain the node's identifying info to provide over RPC.
 #[derive(PartialEq, Serialize, Clone, Debug)]
 pub struct RpcNodeInfo {
     /// Chain id this node is part of.
@@ -23,9 +23,9 @@ pub struct RpcNodeInfo {
     /// prepended with 'node-'. The operator can overwrite
     /// this value since it is not used when writing to file.
     pub name: String,
-    /// The nodes BLS public key.
+    /// The node's BLS public key.
     pub bls_public_key: BlsPublicKey,
-    /// The nodes authority id (hash of BLS key).
+    /// The node's authority id (hash of BLS key).
     /// Used for some tables (like reputation).
     pub authority_id: AuthorityIdentifier,
     /// Address that will receive rewards if this node participates in consensus.
@@ -50,6 +50,6 @@ pub trait EngineToPrimary {
         epoch: Option<Epoch>,
         hash: Option<BlockHash>,
     ) -> impl std::future::Future<Output = Option<(EpochRecord, EpochCertificate)>> + Send;
-    /// Return the nodes static information.
+    /// Return the node's static information.
     fn node_info(&self) -> &RpcNodeInfo;
 }
--- a/crates/execution/tn-rpc/src/lib.rs
+++ b/crates/execution/tn-rpc/src/lib.rs
@@ -5,7 +5,38 @@ mod error;
 mod rpc_ext;
 
 pub use rpc_ext::{TelcoinNetworkRpcExt, TelcoinNetworkRpcExtApiServer};
-use tn_types::{BlockHash, ConsensusHeader, Epoch, EpochCertificate, EpochRecord};
+use serde::{Deserialize, Serialize};
+use tn_types::{
+    Address, AuthorityIdentifier, BlockHash, BlsPublicKey, ConsensusHeader, Epoch,
+    EpochCertificate, EpochRecord, Multiaddr, NetworkPublicKey,
+};
+
+/// Contain the nodes identifying info to provide over RPC.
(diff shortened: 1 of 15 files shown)


crates/execution/tn-rpc/src/lib.rs
```diff
@@ -5,7 +5,38 @@ mod error;
...
+
+/// Contain the nodes identifying info to provide over RPC.
+#[derive(PartialEq, Serialize, Deserialize, Clone, Debug)]
+pub struct RpcNodeInfo {
+    /// The name for the validator. The default value
+    /// is the hashed value of the validator's
+    /// execution address. The operator can overwrite
+    /// this value since it is not used when writing to file.
+    ///
+    /// Keccak256(Address)
```
grantkee: The default is actually `bs58(bls_public_key.to_bytes()[0..8])`. Another comment in `config/src/genesis.rs:189-194` also incorrectly mentions this

crates/execution/tn-rpc/src/lib.rs
```diff
@@ -5,7 +5,38 @@ mod error;
...
-use tn_types::{BlockHash, ConsensusHeader, Epoch, EpochCertificate, EpochRecord};
+use serde::{Deserialize, Serialize};
+use tn_types::{
+    Address, AuthorityIdentifier, BlockHash, BlsPublicKey, ConsensusHeader, Epoch,
+    EpochCertificate, EpochRecord, Multiaddr, NetworkPublicKey,
+};
+
+/// Contain the nodes identifying info to provide over RPC.
+#[derive(PartialEq, Serialize, Deserialize, Clone, Debug)]
+pub struct RpcNodeInfo {
```
grantkee, line 16: This duplicates a lot of `tn_config::NodeInfo` and adds `authority_id`. Maybe use `#[serde(flatten)]`? Is there an advantage of using similar but different structs between config and rpc?

It's probably not a bad idea to include the PoP (although this is easily copyable for immitation attacks). The only downside is more data

crates/node/src/manager/node/epoch.rs
```diff
@@ -756,8 +757,28 @@ where
             )
             .await?;
 
+        let node_info = RpcNodeInfo {
```
grantkee, line 761: This is a small heap allocation, but is it necessary to reconstruct it every epoch? Maybe add this as a helper method and cache it on startup on the manager?

crates/execution/tn-rpc/src/lib.rs
```diff
@@ -5,7 +5,38 @@ mod error;
...
+    /// The nodes BLS public key.
+    pub bls_public_key: BlsPublicKey,
+    /// The nodes authority id (hash of BLS key).
+    /// Used for some tables (like reputation).
+    pub authority_id: AuthorityIdentifier,
+    /// Address that will receive rewards if this node participates in consensus.
+    pub execution_address: Address,
+    /// Network public key for the primary network.
+    pub primary_network_key: NetworkPublicKey,
+    /// Network public key for the workes network.
```
grantkee: sp: "workers network"

crates/node/src/manager/node/epoch.rs
```diff
@@ -756,8 +757,28 @@ where
...
+            name: self.builder.tn_config.node_info.name.clone(),
+            bls_public_key: self.key_config.public_key(),
+            authority_id: self.key_config.public_key().into(),
```
grantkee: nit: `public_key()` is called twice

crates/execution/tn-rpc/src/rpc_ext.rs
```diff
@@ -14,6 +14,9 @@ use tn_types::{BlockHash, ConsensusHeader, Epoch, EpochCertificate, EpochRecord,
 /// TN-specific RPC endpoints.
 #[rpc(server, namespace = "tn")]
 pub trait TelcoinNetworkRpcExtApi {
+    /// Return the nodes information.
```
grantkee: nit: This is public facing and could benefit from more details, like fields, security posture, etc.

grantkee (dismissed):
LGTM - I found a few small tweaks since this is a public facing RPC endpoint that are worth considering

crates/execution/tn-rpc/src/lib.rs
sstanfield: It is more than just a flatten.  The intention of the RPC structure is to only provide external network address and not leak the possibly different internal network addresses (this will also invalidate the POP).  Basically I wanted this to include anything that could identify the node publicly so also adds some stuff.  If you feel strongly about this we can iterated but at the very least I don't think we should expose internal network details (I think we do elsewhere but should probably clean that up as minor security issue).  I am going to also see about adding the version string to this as well.

crates/node/src/manager/node/epoch.rs
sstanfield: Technically it's on the stack, should not be an allocation.  I like it here because it is close to where it is needed plus we can easily add epoch scoped data is we want (maybe the nodes current epoch?).  I don't think this will make a difference in epoch startup time but could be pushed up and we never add epoch data.

grantkee (dismissed):
LGTM! Thank you @sstanfield

grantkee (approved):

PR #696, opened 2026-05-21
Fsync parent directory after file create and rename in archive

### LINKED ISSUE #551: Add fsync on parent directory after file creation and rename
## Problem

After creating new files, directories, or renaming files, the parent directory is never `fsync`'d. On Linux/Unix, `sync_all()` on a file only guarantees the **file data and metadata** are durable — it does not guarantee the **directory entry** (the name-to-inode mapping) is durable. A power loss after file creation or rename can leave the filesystem in a state where the file does not exist, even though `sync_all()` was called on the file itself.

No directory fsync exists anywhere in the archive module today.

### 1. `DataFile::rename()` — `data_file.rs:127-140` (Critical)

- `fs::rename(&self.data_file_path, path)` at line 135
- No fsync on the parent directory of either old or new path
- **Risk:** A committed and renamed pack file could revert to its old name (or disappear entirely) after a crash

### 2. `DataFile::open()` — `data_file.rs:34-76` (Important)

- `File::create_new(path)` at line 40 creates a new file when opening for write
- No fsync on the parent directory after creation
- **Risk:** A newly created data file may not exist after a crash

### 3. `HdxIndex::open_hdx_file()` — `digest_index/index.rs:264-332` (Important)

- `fs::create_dir(dir)` at line 270 creates the index directory
- `OpenOptions::new().create(true).open(dir.join("index.hdx"))` at lines 271-276 creates the index file
- No fsync on parent directory after `create_dir`, no fsync on `dir` after file creation
- **Risk:** Index directory or file may not exist after a crash

### 4. `OdxHeader::open_odx_file()` — `digest_index/odx_header.rs:32-61` (Important)

- `OpenOptions::new().create(true).open(path)` at line 39 creates the overflow index file
- No fsync on the parent directory
- **Risk:** Overflow index file may not exist after a crash

### 5. `PositionIndex::open_pdx_file()` — `position_index/index.rs:121-152` (Important)

- `fs::create_dir(dir)` at line 127 creates the index directory
- `DataFile::open(dir.join("index.pdx"), read_only)` at line 128 creates the index file (via `File::create_new` internally)
- No fsync on parent directory after `create_dir`, no fsync on `dir` after file creation
- **Risk:** Position index directory or file may not exist after a crash

### Step 1: Add a `fsync_directory` helper function

Add a utility function in `data_file.rs` (or a new small utility area in the archive module) that opens a directory as a read-only file and calls `sync_all()` on it:

```rust
/// Fsync a directory to ensure durable directory entries (file creates, renames).
pub(crate) fn fsync_directory(path: &Path) -> Result<(), io::Error> {
    let dir = File::open(path)?;
    dir.sync_all()
}
```

On Unix, `File::open` on a directory path returns a valid file descriptor, and `sync_all()` issues an `fsync()` syscall on it. This is the POSIX-standard way to make directory entries durable.

### Step 2: Fsync after `DataFile::rename()` — `data_file.rs`

After the successful `fs::rename()` at line 135, fsync the parent directory of the **new** path (and the old path's parent too, if it differs). Since `rename` within the same directory is the common case, one fsync usually suffices.

```rust
pub fn rename<P: AsRef<Path>>(&mut self, path: P) -> Result<(), RenameError> {
    let path = path.as_ref();
    if self.data_file_path == path {
        return Ok(());
    }
    if path.exists() {
        return Err(RenameError::FilesExist);
    }
    let old_parent = self.data_file_path.parent().map(Path::to_owned);
    let res = fs::rename(&self.data_file_path, path);
    if res.is_ok() {
        self.data_file_path = path.to_owned();
        // Fsync the new path's parent directory to make the rename durable.
        if let Some(new_parent) = path.parent() {
            fsync_directory(new_parent).map_err(RenameError::RenameIO)?;
        }
        // If the old parent differs, fsync it too (removal of old entry).
        if let Some(old_parent) = &old_parent {
            if path.parent() != Some(old_parent.as_path()) {
                fsync_directory(old_parent).map_err(RenameError::RenameIO)?;
            }
        }
    }
    res.map_err(RenameError::RenameIO)
}
```

### Step 3: Fsync after `DataFile::open()` file creation — `data_file.rs`

When `File::create_new(path)` succeeds (file was actually created), fsync the parent directory. Change line 40 from ignoring the result to checking it:

```rust
if !ro {
    if File::create_new(path).is_ok() {
        // New file was created — fsync parent dir so the entry is durable.
        if let Some(parent) = path.parent() {
            let _ = fsync_directory(parent);
        }
    }
}
```

Note: Errors are still tolerated here (the `let _ =`) since this is a best-effort creation path — the file may already exist.

### Step 4: Fsync after `open_hdx_file()` — `digest_index/index.rs`

After `fs::create_dir(dir)` at line 270 and after the file open/create at lines 271-276, add directory fsyncs:

```rust
let dir = dir.as_ref();
let dir_created = fs::create_dir(dir).is_ok();
// Fsync parent so the new directory entry is durable.
if dir_created {
    if let Some(parent) = dir.parent() {
        let _ = fsync_directory(parent);
    }
}
let mut hdx_file = OpenOptions::new()
    .read(true)
    .write(true)
    .create(true)
    .truncate(false)
    .open(dir.join("index.hdx"))?;
```

Then after the file has been initialized (header written + initial buckets written for the new-file branch), fsync the directory:

```rust
// After the if/else block that writes the header for new files:
// Fsync the index directory so the file entry is durable.
let _ = fsync_directory(dir);
```

### Step 5: Fsync after `open_odx_file()` — `digest_index/odx_header.rs`

After creating and writing the header for a new odx file, fsync the parent directory:

```rust
// After the if/else block for header:
if let Some(parent) = path.parent() {
    let _ = fsync_directory(parent);
}
```

### Step 6: Fsync after `open_pdx_file()` — `position_index/index.rs`

Same pattern — after `fs::create_dir(dir)` at line 127, fsync the parent. After `DataFile::open()` at line 128, fsync `dir`:

```rust
let dir = dir.as_ref();
let dir_created = fs::create_dir(dir).is_ok();
if dir_created {
    if let Some(parent) = dir.parent() {
        let _ = fsync_directory(parent);
    }
}
let mut pdx_file = DataFile::open(dir.join("index.pdx"), read_only)?;
// ... after header init ...
// Fsync the index directory so the pdx file entry is durable.
let _ = fsync_directory(dir);
```

## Files to Modify

| File                                                    | Change                                                                               |
| ------------------------------------------------------- | ------------------------------------------------------------------------------------ |
| `crates/storage/src/archive/data_file.rs`               | Add `fsync_directory()` helper, call after `File::create_new` and `fs::rename`       |
| `crates/storage/src/archive/digest_index/index.rs`      | Call `fsync_directory` after `create_dir` and after file creation in `open_hdx_file` |
| `crates/storage/src/archive/digest_index/odx_header.rs` | Call `fsync_directory` after file creation in `open_odx_file`                        |
| `crates/storage/src/archive/position_index/index.rs`    | Call `fsync_directory` after `create_dir` and after file creation in `open_pdx_file` |

## Error Handling Strategy

- **`DataFile::rename()`**: Directory fsync errors are propagated via `RenameError::RenameIO` — rename durability is critical.
- **`DataFile::open()` and index file opens**: Directory fsync errors are best-effort (`let _ =`) since these paths already tolerate the file pre-existing (the create calls are conditional). The file will exist in memory and be fsync'd on the next `commit()`. If the process crashes before commit, the data wasn't durable anyway.

## Testing

- Existing tests in `pack.rs`, `digest_index/index.rs`, and `position_index/index.rs` should continue to pass — the fsync calls are additive.
- No new tests needed: directory fsync correctness is a kernel guarantee; we only need to verify we don't introduce regressions (e.g., trying to fsync a nonexistent directory). The existing tests exercise the create/open/rename paths and will surface any `io::Error` panics.

Related #510

MavenRain:
## Summary

  Adds `fsync` on parent directories after file creates and renames in the
  archive module, closing a durability gap on Unix where `File::sync_all` alone
  does not flush the directory entry that names a file.

  ## Changes

  - [x] New helper `fsync_directory(path)` in `archive::data_file`.
  - [x] `DataFile::open`: fsync parent after successful `File::create_new`.
  - [x] `DataFile::rename`: fsync new (and old, if different) parent after
  `fs::rename`; errors propagate via `RenameError::RenameIO`.
  - [x] `HdxIndex::open_hdx_file`: fsync parent when the index directory is
  freshly created; fsync the index directory after initial header + bucket
  writes.
  - [x] `OdxHeader::open_odx_file`: fsync parent after the new header is written.
  - [x] `PositionIndex::open_pdx_file`: fsync parent when the index directory is
  freshly created; fsync the index directory after the new header is written.

  Closes #551.  Related: #510.

  ## Testing

  Existing archive tests (`test_archive_hdx_index`, `test_archive_pdx_index`,
  `test_consensus_pack`) exercise the create/open/rename paths and surface any
  `io::Error` from the new fsync calls.  No new tests added; per the issue,
  directory fsync correctness is a kernel guarantee.

  Local: `cargo +nightly-2026-03-20 fmt` clean, `cargo +nightly-2026-03-20 clippy
   -p tn-storage --all-features` clean, `cargo test -p tn-storage --lib`

### DIFF
--- a/crates/storage/src/archive/data_file.rs
+++ b/crates/storage/src/archive/data_file.rs
@@ -12,6 +12,18 @@ use crate::archive::error::rename::RenameError;
 const READ_BUFFER_SIZE: usize = 16 * 1024; // 16kb
 const WRITE_BUFFER_SIZE: usize = 16 * 1024; // 16kb
 
+/// Fsync a directory so recent directory-entry changes (file creates, renames)
+/// are durable on Unix.
+///
+/// `File::sync_all` on a regular file does not flush the parent directory entry
+/// that names it.  A crash between a create/rename and a subsequent directory
+/// fsync can leave the filesystem with the entry missing even though the file's
+/// own data and metadata are durable.  Calling this on the parent directory
+/// closes that gap.
+pub(crate) fn fsync_directory(path: &Path) -> Result<(), io::Error> {
+    File::open(path)?.sync_all()
+}
+
 /// Wrapper around a file that implements read and write buffers and manages
 /// them seamlessly via standard IO traits.
 #[derive(Debug)]
@@ -38,7 +50,12 @@ impl DataFile {
             // If we are opening for write then make sure the file exists.
             // This function will create it if it does not exist or produce
             // an error if it does so ignore the errors.
-            let _ = File::create_new(path);
+            if File::create_new(path).is_ok() {
+                // New file was created; fsync the parent so the directory entry is durable.
+                if let Some(parent) = path.parent() {
+                    let _ = fsync_directory(parent);
+                }
+            }
         }
         let mut data_file = OpenOptions::new().read(true).append(!read_only).open(path)?;
         let data_file_end = data_file.seek(SeekFrom::End(0))?;
@@ -164,9 +181,21 @@ impl DataFile {
         if path.exists() {
             return Err(RenameError::FilesExist);
         }
+        let old_parent = self.data_file_path.parent().map(Path::to_owned);
         let res = fs::rename(&self.data_file_path, path);
         if res.is_ok() {
             self.data_file_path = path.to_owned();
+            // Fsync the new path's parent so the rename is durable on the entry side.
+            if let Some(new_parent) = path.parent() {
+                fsync_directory(new_parent).map_err(RenameError::RenameIO)?;
+            }
+            // If the source was in a different directory, fsync that too so the old entry's
+            // removal is durable.
+            if let Some(old_parent) = &old_parent {
+                if path.parent() != Some(old_parent.as_path()) {
+                    fsync_directory(old_parent).map_err(RenameError::RenameIO)?;
+                }
+            }
         }
         res.map_err(RenameError::RenameIO)
     }
--- a/crates/storage/src/archive/data_file.rs
+++ b/crates/storage/src/archive/data_file.rs
@@ -20,6 +20,12 @@ const WRITE_BUFFER_SIZE: usize = 16 * 1024; // 16kb
 /// fsync can leave the filesystem with the entry missing even though the file's
 /// own data and metadata are durable.  Calling this on the parent directory
 /// closes that gap.
+///
+/// # Precondition
+///
+/// `path` must refer to a directory.  Passing a regular file silently fsyncs
+/// that file's contents instead of providing the directory-entry durability
+/// guarantee callers expect, so all in-crate call sites pass a directory.
 pub(crate) fn fsync_directory(path: &Path) -> Result<(), io::Error> {
     File::open(path)?.sync_all()
 }
--- a/crates/storage/src/archive/digest_index/index.rs
+++ b/crates/storage/src/archive/digest_index/index.rs
@@ -335,6 +335,9 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
                 hdx_file.write_all(&chunk[..n * bucket_size])?;
                 remaining -= n;
             }
+            // Header and initial buckets were just written; fsync the directory so the index.hdx
+            // entry is durable.
+            let _ = fsync_directory(dir);
             (header, bloom)
         } else {
             let header = HdxHeader::load_header(&mut hdx_file)?;
@@ -359,11 +362,6 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
             let bloom: Bloom = bloom_bits.try_into()?;
             (header, bloom)
         };
-        if file_end == 0 {
-            // Header and initial buckets were just written; fsync the directory so the index.hdx
-            // entry is durable.
-            let _ = fsync_directory(dir);
-        }
         let (odx_file, _odx_header) = OdxHeader::open_odx_file(
             header.version(),
             header.uid(),
--- a/crates/storage/src/archive/digest_index/index.rs
+++ b/crates/storage/src/archive/digest_index/index.rs
@@ -4,6 +4,7 @@ use tn_types::B256;
 
 use crate::archive::{
     crc::{add_crc32, check_crc},
+    data_file::fsync_directory,
     digest_index::{
         bloom::{Bloom, BLOOM_SIZE_BYTES},
         bucket_iter::BucketIter,
@@ -292,7 +293,13 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
         read_only: bool,
     ) -> Result<HdxIndex<KSIZE, S>, LoadHeaderError> {
         let dir = dir.as_ref();
-        let _ = fs::create_dir(dir);
+        let dir_created = fs::create_dir(dir).is_ok();
+        if dir_created {
+            // The index directory is brand new; fsync the parent so the entry survives a crash.
+            if let Some(parent) = dir.parent() {
+                let _ = fsync_directory(parent);
+            }
+        }
         let mut hdx_file = if read_only {
             OpenOptions::new().read(true).write(false).open(dir.join("index.hdx"))?
         } else {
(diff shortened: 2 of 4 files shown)


crates/storage/src/archive/digest_index/index.rs
```diff
@@ -352,6 +359,11 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
             let bloom: Bloom = bloom_bits.try_into()?;
             (header, bloom)
         };
+        if file_end == 0 {
```
sstanfield: Any reason to not add this to the existing file_end == 0 branch above?

crates/storage/src/archive/digest_index/odx_header.rs
```diff
@@ -65,6 +66,13 @@ impl OdxHeader {
             }
             header
         };
+        if file_end == 0 {
```
sstanfield: Why not use the existing file_end == 0 branch?

crates/storage/src/archive/position_index/index.rs
```diff
@@ -148,6 +155,11 @@ impl PositionIndex {
             }
             header
         };
+        if was_empty {
```
sstanfield: Same comment about the if, now I wonder if I am just missing something or you find this more clear?

crates/storage/src/archive/data_file.rs
```diff
@@ -12,6 +12,18 @@ use crate::archive::error::rename::RenameError;
...
+/// Fsync a directory so recent directory-entry changes (file creates, renames)
+/// are durable on Unix.
+///
+/// `File::sync_all` on a regular file does not flush the parent directory entry
+/// that names it.  A crash between a create/rename and a subsequent directory
+/// fsync can leave the filesystem with the entry missing even though the file's
+/// own data and metadata are durable.  Calling this on the parent directory
+/// closes that gap.
+pub(crate) fn fsync_directory(path: &Path) -> Result<(), io::Error> {
+    File::open(path)?.sync_all()
```
sstanfield, line 30: Should we confirm this is a directory?  It does not really matter, we only call it with dirs just wondering if that is clearer.  Not a big deal to me either way.

sstanfield (dismissed):
This looks good.  I had a couple of comments but they are trivial and you may prefer your way.  The only minor concern I have is all the syncs throwing of timing (ie being slow)- of course if they are we have to solve that since this is correct.  If you wanted to run the restart tests (sound familiar... and just as fun) you can use ```cargo test -p e2e-tests -- --ignored``` .  Any issues should show up then but I don't think they will.  Those are optional and slow.  One of us will have to do a full "attest" (basically a slow CI run that gets noted on testnet for GitHub to confirm) anyway that will do this.  I do NOT expect any issues.  Restarts have been working well lately so should be safe.  Thanks for the PR.

crates/storage/src/archive/digest_index/index.rs
MavenRain: No good reason, as the duplicate guard was just a holdover from how I added the fsync incrementally.  I folded the call into the new-file arm above.

crates/storage/src/archive/digest_index/odx_header.rs
MavenRain: Done.  I moved the parent fsync into the new-file arm so there is one branch on file_end == 0 now.

crates/storage/src/archive/position_index/index.rs
MavenRain: You weren't missing anything.  The second if was_empty was redundant.  I moved into the new-file arm.

crates/storage/src/archive/data_file.rs
MavenRain: I left the runtime alone since every in-crate call site passes a directory and adding a metadata() syscall before each fsync would charge every caller for a precondition the call boundary already enforces.  To address the clarity side I extended the doc with an explicit #Precondition section, which should make the contract obvious without the extra syscall.  I'd be happy to swap in a runtime is_dir() check if you'd rather have the belt-and-suspenders.

sstanfield (approved):

grantkee (approved):
Great work @MavenRain ! Thank you for taking a look and opening this PR

Issue #622, opened 2026-04-08
Refactor general consensus side storage.


sstanfield:
We keep ephemeral data in one data store and the consensus chain in another.  There is a trait (Database) over the ephemeral data that is currently used as a generic but the consensus chain is a concrete type.  We should consider making the Database trait more generic and expand it to cover all types of data access.  This will allow us to access any data with one common interface and not require use of two different data stores.

MavenRain:
@sstanfield, I'd like to grab this, but I want to draw your attention to something first and get your opinion on the direction before proceeding.

Here is a complication that I see: `Database` is sync, mutable, KV-table.  `ConsensusChain` / `ConsensusPack` / `EpochRecordDb` / `CertificatePack` are async, append-only, domain-typed (`ConsensusHeader`, `Batch`, `EpochMeta`, plus methods like `count_leaders` and `read_last_committed`).  Forcing the pack stores into the current `Database` shape shoehorns async into sync and loses log/immutability semantics.

There are a few directions to take:

  1. **Extend the `Database` trait directly** to cover consensus chain data: add domain-typed read/write methods (or associated types for non-table data sources) so a single `Database` impl can carry both.  This is closest to the issue text.  The cost is that trait grows, sync/async impedance leaks into the interface, and callers that only want one side still see the whole surface.

  2. **Interface segregation**: split `Database` into facet traits (`TableStore` = current `Database`, `ConsensusReader`/`ConsensusWriter` for the consensus-chain side), with a `UnifiedStore: TableStore + ConsensusReader +
  ...` blanket for places that need both.  Each crate depends on the facet it actually uses, and existing concrete types stay.

  3. **Single unified handle**: a `NodeStore` that owns both `<DB: Database>` and `ConsensusChain` and exposes one trait spanning their union.  There is a thin wrapper with untouched impls;  The payoff is "one handle to thread through" rather than abstraction-level unification.

  4. **Wrap each consensus store as a `Database` impl with synthetic tables**:  This is the smallest call-site change with the biggest mismatch on the impl side (async-on-sync, table-on-log).

I, personally, lean toward (2) for clean interfaces, but if you'd prefer to literally grow `Database` per the issue text I'd take (1) instead.  (3) is the right pick if the goal is reducing constructor plumbing rather than abstraction-level unification.  (4) feels wrong for the pack stores.

Before drafting, can you confirm:
[...]

PR #698, opened 2026-05-21
Add ConsensusChainReader/Writer facet traits

MavenRain:
Refs #622.  First half of the storage-unification work confirmed with
  @sstanfield: introduce facet traits over the consensus-chain handle
  without changing any call sites.

  - Define `ConsensusChainReader` and `ConsensusChainWriter` in a new `tn_storage::traits` module, exposing the read and read+write surface of the consensus chain over `impl Future + Send`.
  - Add a `UnifiedStore` blanket trait combining `tn_types::Database` with `ConsensusChainWriter` for call sites that want a single bound for the table-KV layer plus the consensus chain.
  - Implement both facet traits for the existing `ConsensusChain` struct by delegating to the inherent methods; no behaviour change.

  Call-site migration to `<C: ConsensusChainWriter>` lives in a follow-up
  PR per maintainer guidance to keep test churn isolated.  `EpochRecordDb`
  and `CertificatePack` remain behind the `ConsensusChain` abstraction
  for now.

  The `write_subdag_for_test` helper deliberately stays as inherent-only.  Tests
  that use it work with the concrete handle, and the trait stays clean for
  non-test callers.

### DIFF
--- a/crates/storage/src/consensus.rs
+++ b/crates/storage/src/consensus.rs
@@ -15,11 +15,12 @@ use std::{
 use parking_lot::Mutex;
 use tn_types::{
     gas_accumulator::RewardsCounter, AuthorityIdentifier, Batch, BlockHash, BlockNumHash,
-    CommittedSubDag, Committee, ConsensusHeader, ConsensusOutput, Epoch, EpochRecord, Round, B256,
+    CommittedSubDag, Committee, ConsensusChainReader, ConsensusChainWriter, ConsensusHeader,
+    ConsensusOutput, Epoch, EpochRecord, ReadStream, Round, B256,
 };
 use tokio::{
     fs::File as AsyncFile,
-    io::{AsyncRead, AsyncSeek},
+    io::AsyncRead,
     sync::{
         mpsc::{self, Sender},
         oneshot,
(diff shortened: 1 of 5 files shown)


sstanfield (commented):
This looks great.  I was originally thinking of one trait but prefer the read and write split after seeing it, it should clarify code a bit.  Also I prefer using the Future syntax for the traits for async vs async_trait- we have both ways in the codebase but now that you don't have to have async_trait I tend to use this method.

One thing I think has to be fixed though.  The traits need to live in the tn_types crate otherwise we will run into dep issues once it is deployed.  See telcoin-network/crates/types/src/database_traits.rs for an example for the existing Database trait.  You can use that file if want since these are "DB" traits or use a new file as you see fit.

Otherwise I think this is the perfect start for this.  Thanks.

MavenRain:
Thanks for the review.  I moved the traits to a new `tn_types::consensus_chain_traits` module alongside `database_traits`, kept the `-> impl Future + Send` declarations on the trait, and switched the trait method return types to `eyre::Result` to match `Database`.  `ReadStream` moved along with them and now has a blanket impl over the underlying tokio I/O supertraits so concrete stream types don't need an explicit `impl ReadStream` each.

sstanfield (approved):

grantkee (approved):
This is great! Agreed that the read/write split is a good approach. Thanks @MavenRain

PR #699, opened 2026-05-28
Fixes some observer issues.

sstanfield:
- Improves some of the DB failure logs for debugging.
- Clears the consensus header cache on a restart.
- Closes a hole that could leak consensus headers in the cache (never remove them).
- Allow a pack file with the same batch in multiple subdags to stream.
- Increase the max size of some of the mdmx DBs.
- Focus some of the catch up logs to be cleaner.


grantkee (dismissed):
LGTM!

grantkee (approved):

PR #701, opened 2026-06-01
Handle duplicate batch digests in a committed sub-DAG

### LINKED ISSUE #700: Node crashes when the same batch digest appears in two certificates of one committed sub-DAG
## Problem

When Bullshark commits a sub-DAG, the subscriber rebuilds `ConsensusOutput` by fetching every referenced batch and mapping it back to the certificate that included it.
Batches are fetched into a digest-keyed map and then drained per certificate with `remove(digest)`.
This assumes each batch digest is referenced by exactly one certificate in the committed sub-DAG.

That assumption does not hold.
A proposer re-queues the digests of any of its headers that fall at or below `max_committed_round` but were never actually committed, so those digests get re-proposed in a later header.
When such a header is later certified and committed in the same sub-DAG as the original, the same batch digest now appears in two certificates.
The first certificate's `remove` consumes the entry; the second `remove` returns `None`.

The old code treated that `None` as a missing batch and returned `MissingFetchedBatch`, which is a fatal error that triggers a clean node shutdown.
On adiri this surfaced as a repeating crash: a node would commit a sub-DAG containing a duplicated batch digest and immediately go down, even though nothing was actually wrong with the data - the batch was present, it had just already been attached to the earlier certificate.

## Solution

The subscriber should distinguish a genuinely missing batch from one that was already consumed by an earlier certificate in the same sub-DAG.

Keep the full, non-deduplicated list of referenced digests for the sub-DAG alongside the deduplicated fetch set.
When `remove(digest)` returns `None`, check that list: if the digest is part of the sub-DAG's payload, it is a known duplicate that the earlier certificate already attached, and the execution engine will ignore the repeat - so skip it and continue.
Only when the digest is absent from the sub-DAG entirely should the subscriber treat it as a real protocol violation and return `MissingFetchedBatch`.

The error variant should also carry the offending digest so the fatal case is diagnosable from logs, and the duplicate path should log a warning rather than crash.

grantkee:
* `fetch_batches` no longer crashes when a batch digest appears in more than one certificate of a committed sub-DAG. A `None` from `fetched_batches.remove` is now checked against the sub-DAG's full digest list: a known duplicate is skipped (the engine ignores the repeat), and only a digest absent from the sub-DAG returns `MissingFetchedBatch`.
* `MissingFetchedBatch` now includes the offending `BlockHash` so the fatal case is diagnosable from logs.
* Added `test_duplicate_batch_digest`: builds a DAG where one authority reuses the same batch digest across two rounds, commits both in one sub-DAG, and asserts the subscriber produces output instead of shutting down.
* Added `test_process_committed_headers` for the proposer's re-queue logic — the source of the duplicate: headers at or below `max_committed_round` that weren't committed get their digests re-queued, while headers above it do not. Covers the re-queue, no-re-queue, and empty-committed cases.

The duplicate originates in the proposer re-queue path and only became fatal in the subscriber, so both sides are covered.

closes #700

### DIFF
--- a/crates/consensus/executor/src/errors.rs
+++ b/crates/consensus/executor/src/errors.rs
@@ -135,6 +135,6 @@ pub enum SubscriberError {
     /// This is a protocol violation that occurs when consensus references a batch
     /// but it's not present in the fetched results. This should not happen if workers
     /// are behaving correctly and may indicate either a worker bug or data corruption.
-    #[error("A fetched batch is missing from the collection.")]
+    #[error("A fetched batch is missing from the collection: {0}")]
     MissingFetchedBatch(BlockHash),
 }
--- a/crates/consensus/executor/src/subscriber.rs
+++ b/crates/consensus/executor/src/subscriber.rs
@@ -4,7 +4,7 @@ use crate::{errors::SubscriberResult, SubscriberError};
 use futures::{stream::FuturesOrdered, StreamExt};
 use state_sync::{last_consensus_parent, save_consensus, spawn_state_sync};
 use std::{
-    collections::{BTreeSet, HashMap, VecDeque},
+    collections::{BTreeSet, HashMap, HashSet, VecDeque},
     sync::Arc,
     time::Duration,
 };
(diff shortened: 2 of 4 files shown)


sstanfield (approved):
LGTM

PR #704, opened 2026-06-01
Remove unused UpdateAuthorizedPublishers network command

### LINKED ISSUE #703: Remove unused `UpdateAuthorizedPublishers` network command
## Problem

The `network-libp2p` crate exposes a `NetworkCommand::UpdateAuthorizedPublishers` variant and a corresponding `update_authorized_publishers` helper on the network handle. Both were originally intended to mutate the per-topic authorized publisher set in-place at epoch boundaries, so the gossipsub layer could keep filtering messages from the previous committee while it learned about the next one.

Authorization is currently established at subscribe time: each epoch, the node calls `subscribe_with_publishers(topic, committee_pub_keys)` for every gossipsub topic, and the subscription itself carries the authorized publisher set. There is no remaining call site that uses `update_authorized_publishers` outside the one test that was simulating a whitelist reset — production code never reaches the command. The handler still mutates `self.authorized_publishers` on the network task and ships a `oneshot::Sender<NetworkResult<()>>` reply channel that nothing awaits.

Carrying a dead command path on the network task is a small but real liability. It widens the public API of the subscriber handle, gives the impression that runtime mutation of the publisher whitelist is a supported operation, and leaves a second mechanism for changing authorization that could drift out of sync with the subscribe-time path if either changes. For a security-sensitive surface (who is allowed to publish on a consensus topic), having one authoritative path is preferable to two.

## Solution

The `UpdateAuthorizedPublishers` variant should be removed from `NetworkCommand`, along with the `update_authorized_publishers` helper on the subscriber handle and the matching arm in the command dispatcher. The remaining `subscribe_with_publishers` path should be treated as the single authoritative way to bind a topic to an authorized publisher set.

The one in-tree caller — a test that previously reset the whitelist mid-run by sending `update_authorized_publishers(HashMap::new())` — should switch to re-subscribing the topic with an empty publisher set, which produces the same observable effect (subsequent messages from the previously-authorized publisher are rejected) through the supported API.

No on-the-wire behavior or epoch-transition semantics change: production code already routes all authorization updates through subscribe-time setup, so removing the unused command is a pure API and dead-code cleanup that narrows the gossipsub authorization surface to a single path.

grantkee:
- Drop `NetworkCommand::UpdateAuthorizedPublishers`, the matching dispatcher arm in `consensus.rs`, and the `update_authorized_publishers` helper on the subscriber handle — no production caller remains.
- Authorization is already set per topic via `subscribe_with_publishers` at each epoch boundary, so this removes a second, now-redundant path for mutating the publisher whitelist.
- Update the one affected test (`test_msg_verification_ignores_unauthorized_publisher`) to clear the whitelist by re-subscribing with an empty publisher set instead of calling the removed helper.

closes #703

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -518,11 +518,6 @@ where
     /// Process commands for the network.
     fn process_command(&mut self, command: NetworkCommand<Req, Res>) -> NetworkResult<()> {
         match command {
-            NetworkCommand::UpdateAuthorizedPublishers { authorities, reply } => {
-                // this value should be updated at the start of each epoch
-                self.authorized_publishers = authorities;
-                send_or_log_error!(reply, Ok(()), "UpdateAuthorizedPublishers");
-            }
             NetworkCommand::StartListening { multiaddr, reply } => {
                 let res = self.swarm.listen_on(multiaddr);
                 send_or_log_error!(reply, res, "StartListening");
--- a/crates/network-libp2p/src/tests/network_tests.rs
+++ b/crates/network-libp2p/src/tests/network_tests.rs
@@ -737,7 +737,7 @@ async fn test_msg_verification_ignores_unauthorized_publisher() -> eyre::Result<
     }
 
     // remove cvv from whitelist and try to publish again
-    nvv.update_authorized_publishers(HashMap::new()).await?;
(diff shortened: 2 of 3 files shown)


sstanfield (approved):
Nice catch, I guess we stopped needing this a while ago with newer epoch start...

PR #705, opened 2026-06-01
Make consensus headers cheap to clone.

sstanfield:
Moved to an internal Arc pattern and made fields private.  The cached digest can not be inconsistent anymore outside of a bug in header.rs.  Also removed the now unnecessary Arc from Certificate and also made its fields private to discourage invalidating certs.  Removed an error and test for invalid digests on headers- this is not expressible anymore.  Added a test to make sure the Header changes do not change the serialize format.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2237,7 +2237,7 @@ dependencies = [
  "anstream",
  "anstyle",
  "clap_lex",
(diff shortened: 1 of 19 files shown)


grantkee (approved):
LGTM! Well done @sstanfield

PR #707, opened 2026-06-03
Track previous/current/next committees in the network layer

### LINKED ISSUE #708: Network peer layer tracks only the current committee, mishandling epoch boundaries
## Problem

The libp2p peer layer (`AllPeers` / `PeerManager`) tracks a single committee per epoch. When an epoch starts, `new_epoch` clears the previous committee set and rebuilds it from the incoming committee. A peer counts as a validator only if it is in that one current set, and validator status is what protects a peer from banning, pruning, and reputation penalties.

This single-committee view does not match how committee membership actually behaves across an epoch boundary. The on-chain `ConsensusRegistry` already distinguishes previous, current, and next committees, but the network mirrors only one of them. Two problems follow:

- **The just-completed committee is dropped instantly.** The moment a new epoch begins, the outgoing committee's peers stop counting as validators and become eligible for pruning/banning - even though late gossip from the prior epoch (final certificates, votes, batches) may still be in flight and worth accepting. There is no window during which the previous committee remains trusted.
- **The next committee is unprotected until it is already voting.** Peers joining in the upcoming epoch are not treated as validators until the epoch they vote in has begun. They can be banned or deprioritized in the run-up to their first round, exactly when they need to be dialable and synced.

The net effect is unnecessary disconnects and reputation churn around every epoch transition, and no foundation for accepting late gossip from a committee that just rotated out.

## Solution

The peer layer should mirror the same three committees the registry tracks: previous, current, and next. A peer should count as a validator if it appears in **any** of the three slots, so the just-completed committee is retained for one rotation while late gossip may still arrive, and the next committee is protected before it begins voting.

Introduce an explicit rotation model in place of the single `new_epoch` call:

- On the initial epoch, **seed** the current and next slots (the previous slot is empty because no epoch has closed yet).
- On every subsequent epoch boundary, **rotate** the slots forward — `previous <- current`, `current <- next`, `next <- incoming` — and record the timestamp at which the just-closed epoch ended.

Because trusting a committee peer is a one-way ratchet, only the newly-arriving committee needs to be run through the unban/trust loop on each rotation; peers already promoted in a prior rotation keep their status. The epoch manager should drive this by seeding on the first epoch and rotating thereafter, sourcing the next committee from the consensus config and passing the previous epoch's boundary timestamp.

The existing deadlock-breaker pre-dial path (used when a node comes up with no peers) should forgive bans for a committee so it can be dialed **without** prematurely mutating the slot state, since the real slot update follows shortly after through the seed/rotate path.

The recorded epoch-end timestamp should be stored ahead of its consumer: a follow-up will read it to accept late gossipsub messages from the previous committee within a grace window. Non-validator-voter (NVV) support and the late-gossip acceptance window itself remain future work; this change establishes the three-committee tracking they build on.

grantkee:
- The peer layer now tracks three committees (previous/current/next) instead of one. `is_peer_validator` matches any of the three, so the outgoing committee survives one rotation (late gossip still counts) and the next committee is trusted before it votes.
- `NewEpoch` splits into three commands: `InitializeCommittees` (seed current + next on epoch 0), `NextCommittee` (rotate slots forward + record the closed epoch's end timestamp), and `PrepareCommitteeDial` (deadlock-breaker pre-dial — unban without mutating slots).
- The epoch manager seeds on the initial epoch and rotates thereafter, sourcing `next` from `consensus_config.next_committee_keys()` and the boundary from a new `previous_epoch_boundary` field captured before `epoch_boundary` is overwritten.

Foundational change; NVV support and the late-gossip window land in follow-ups.

closes #708

### DIFF
--- a/crates/network-libp2p/src/peers/all_peers.rs
+++ b/crates/network-libp2p/src/peers/all_peers.rs
@@ -734,20 +734,6 @@ impl AllPeers {
         }
     }
 
-    /// Boolean indicating if this peer is in the current committee of voting validators.
-    ///
-    /// This is the narrow current-only check; [`Self::is_peer_validator`] widens membership to the
-    /// previous and next committees as well. Kept as the building block for future NVV support.
-    #[allow(dead_code)]
-    fn is_peer_cvv(&self, peer_id: &PeerId) -> bool {
-        match self.identity_for(peer_id) {
-            PeerIdentity::Confirmed(bls_public_key) => {
-                self.current_committee.contains(&bls_public_key)
-            }
-            PeerIdentity::Unidentified(_) => false,
-        }
-    }
-
     /// Boolean indicating if the ip address is associated with a banned peer.
     pub(super) fn ip_banned(&self, ip: &IpAddr) -> bool {
         self.banned_peers.ip_banned(ip)
--- a/crates/network-libp2p/src/peers/all_peers.rs
+++ b/crates/network-libp2p/src/peers/all_peers.rs
@@ -37,16 +37,17 @@ pub(super) struct AllPeers {
     peers: HashMap<PeerId, Peer>,
     /// Peers in the previous epoch's committee.
     ///
-    /// Retained for one rotation so late gossip from the just-completed committee still counts as
-    /// validator traffic and those peers are not pruned mid-rotation.
+    /// Set every epoch from authoritative state so late gossip from the just-completed committee
+    /// still counts as validator traffic and those peers are not pruned mid-rotation.
     previous_committee: HashSet<PeerId>,
     /// Peers in the current epoch's committee.
     current_committee: HashSet<PeerId>,
     /// Peers in the next epoch's committee, pre-emptively trusted so they are not banned before
     /// they begin voting.
     next_committee: HashSet<PeerId>,
-    /// Timestamp (sec) at which the previous epoch closed. Recorded on every rotation; the `0`
-    /// sentinel means no previous epoch boundary has been observed yet (initial epoch).
+    /// Timestamp (sec) at which the previous epoch closed. Set every epoch from authoritative
+    /// state; the `0` sentinel means no previous epoch boundary has been observed yet (initial
+    /// epoch).
     ///
     /// Stored ahead of its consumer: a future PR reads this to accept late gossipsub messages from
     /// the previous committee within a grace window.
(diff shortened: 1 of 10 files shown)


crates/network-libp2p/src/peers/all_peers.rs
```diff
@@ -817,26 +823,88 @@ impl AllPeers {
...
+            .collect();
+        for peer_id in previously_tracked.difference(&still_tracked) {
+            if let Some(peer) = self.peers.get_mut(peer_id) {
+                peer.make_untrusted();
+            }
+        }
+
+        // unban + trust every member once (peers in multiple committees are handled a single time)
+        let mut processed: HashSet<PeerId> = HashSet::new();
+        let members: Vec<(BlsPublicKey, NetworkInfo)> = previous
```
sstanfield: Would it make more sense to just make this a set vs a Vec with a spare Set to dedupe?  Could pass a set or iter to apply_committee_membership().

crates/node/src/manager/node/epoch.rs
```diff
@@ -164,17 +164,29 @@ where
...
 
+        // The networks need their one-time, per-process setup (start listening, register bootstrap
+        // peers) on the first iteration that actually reaches `create_consensus`. This is usually
+        // the `Initial` epoch, but the replay above can return early before
+        // `create_consensus` on a restart that replays-and-closes an epoch boundary, so the first
+        // real setup then happens on a following `NewEpoch` iteration. Drive the decision off
+        // whether the network has actually been set up yet (not off `RunEpochMode::Initial`) so the
+        // setup is never skipped on that restart path. (Committee slots are set every epoch
+        // regardless via `update_committees`.)
+        let network_first_init = epoch_mode.initial_epoch() || !self.network_initialized;
```
sstanfield, line 175: Good catch.

sstanfield (dismissed):
Looks good with one minor comment.

crates/network-libp2p/src/peers/all_peers.rs
```diff
@@ -833,13 +833,13 @@ impl AllPeers {
...
-    /// more than one committee are processed once. The demote and re-trust sets are disjoint by
-    /// construction, so no peer is demoted then re-trusted in a single call.
+    /// more than one committee are processed once by map construction. The demote and re-trust
+    /// sets are disjoint by construction, so no peer is demoted then re-trusted in a single call.
     pub(super) fn update_committees(
         &mut self,
-        previous: Vec<(BlsPublicKey, NetworkInfo)>,
-        current: Vec<(BlsPublicKey, NetworkInfo)>,
-        next: Vec<(BlsPublicKey, NetworkInfo)>,
+        previous: HashMap<PeerId, (BlsPublicKey, NetworkInfo)>,
```
sstanfield: Did you really want to introduce the PeerId to the committees?  I assume so and yea this will will probably need to update soon.  I also assume we do not care about the order of these?  I don't think this layer verifies any sigs so should not.

sstanfield (dismissed):

sstanfield (dismissed):

sstanfield (approved):

PR #711, opened 2026-06-03
This should address the missing pack files (never retrieved) found in testnet observer testing.

sstanfield:
The main fix was removing an unnecessary guard on pack file retrieval.  Also does a request for a pack file when starting an epoch when consensus is on a future epoch (sanity check).  Small cleanups around the general issue.


grantkee (approved):
LGTM!

Issue #710, opened 2026-06-03
Validate and map PeerIds from kademlia records


grantkee:
## Problem

Validators are identified by BLS keys throughout consensus and the protocol. The network layer publishes signed kademlia records keyed on the validator's BLS public key: each record contains a `NodeRecord` whose `NetworkInfo` (network public key and multiaddrs) is signed by the BLS key. The libp2p `PeerId` is derived from the `NetworkPublicKey` in the record and is only used by libp2p - consensus and the protocol never reference it.

Record validation (`peer_record_valid` in `crates/network-libp2p/src/consensus.rs`) verifies the BLS signature and that the record's publisher matches the PeerId derived from the record's network key. It does not check the multiaddrs inside the record: a record can carry a multiaddr with a `/p2p/<PeerId>` component that disagrees with the record's network key. Downstream, the BLS-to-PeerId mapping used by `AllPeers` is built from the record's network key without cross-checking the addresses peers actually dial.

## Solution

When validating a kad record, derive the expected PeerId from the record's `NetworkPublicKey` and compare it against the PeerId embedded in each multiaddr's `/p2p` component when present; records that disagree should be rejected. When a multiaddr carries no `/p2p` component, the record's network public key should be used as the source of the PeerId.

The validated network-key-to-PeerId derivation should be the single source for mapping BLS keys to PeerIds in `AllPeers`. This keeps the separation explicit: the BLS key identifies the validator to consensus and the protocol, while the PeerId is an implementation detail used only by libp2p.

## Open question

`AllPeers` currently keys its peers map by `PeerId`. Since validator identity is the BLS key everywhere outside libp2p, should `AllPeers` key peers by `BlsPublicKey` instead and resolve PeerIds only at the libp2p boundary?

PR #716, opened 2026-06-04
Key AllPeers by domain identity instead of libp2p PeerId.

MavenRain:
`AllPeers` previously keyed every peer by its libp2p `PeerId`, an implementation-level id, even though committee/trusted/known peers are identified in the telcoin domain by their `BlsPublicKey`.   A blanket `PeerId` -> `BlsPublicKey` swap is unsound: `AllPeers` also tracks anonymous inbound and kad-dialed peers that have no bls key (ensure_peer_exists inserts Peer::default()), and libp2p connection events only ever carry a `PeerId`.

Introduce a `PeerIdentity` sum type that captures the real invariant:

```
      enum PeerIdentity {
          Confirmed(BlsPublicKey),   // bls known (committee/trusted/known)
          Unidentified(PeerId),      // no bls learned yet
      }
```
`AllPeers.peers` is now keyed by `PeerIdentity` and `current_committee` by `BlsPublicKey`.  A private `PeerId -> BlsPublicKey` index (kept in lockstep with the Confirmed entries) resolves incoming libp2p events to the right key; the three bls-ingesting methods (`add_trusted_peer`, `upsert_peer`, `new_epoch`) promote a peer from `Unidentified` onto its `Confirmed` identity, preserving accumulated state.  Methods that return ids for libp2p calls recover the `PeerId` from the peer's network key.

The change is contained to the peers module: the `PeerManager` API, `PeerEvent`, and `PeerExchangeMap` (already bls-keyed) are unchanged, so no `PeerId` leaks across the crate boundary.

This relates to https://github.com/Telcoin-Association/telcoin-network/issues/710

### DIFF
--- a/crates/network-libp2p/src/peers/all_peers.rs
+++ b/crates/network-libp2p/src/peers/all_peers.rs
@@ -94,10 +94,11 @@ impl AllPeers {
     /// A peer is `Confirmed` once its bls key is known (committee/trusted/known peers); otherwise
     /// it is `Unidentified` and keyed by its libp2p id.
     fn identity_for(&self, peer_id: &PeerId) -> PeerIdentity {
-        match self.confirmed.get(peer_id) {
-            Some(bls_public_key) => PeerIdentity::Confirmed(*bls_public_key),
-            None => PeerIdentity::Unidentified(*peer_id),
-        }
+        self.confirmed
+            .get(peer_id)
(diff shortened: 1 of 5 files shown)


sstanfield (approved):
LGTM!

grantkee (approved):
Thanks, @MavenRain ! Good work. Follow up PR #722

PR #718, opened 2026-06-05
Security and performance fixes for sync.

sstanfield:
This is a batch of small sync fixes to address security concerns, performance and general correctness.  They are precursors to a larger sync refactor (addressing inefficient following of the current epoch).

Automated security finding addressed:

- **"Send watch value if newer" compare-and-update is hand-rolled five times** — Two distinct watches each repeat a borrow-current / compare / `send_replace` dance. For `last_consensus_header`: consensus.rs:55-64 (`get_consensus_header`) and consensus.rs:203-215 (`spawn_fetch_consensus`) both read `last_consensus_header().borrow()...map(|h| h.number).unwrap_or_default()` then `send_replace` only if `header.number >` it. For `last_published_consensus_num_hash`: epoch.rs:64-75 and epoch.rs:164-172 both read via `published_consensus_num_hash()`, compare `number > old_number`, then `send_replace`; handler.rs:253-255/289-291 does the same compare-and-send. A `ConsensusBusApp::send_consensus_header_if_newer(header)` and `send_published_num_hash_if_newer((epoch, number, hash))` would collapse all five into one guarded call each and put the monotonic invariant in one place. The `requested_missing_epoch` write at handler.rs:300-305 is the same shape (`epoch > latest_missing` then `send_replace`) and could take a third helper. (Note: subscriber.rs:312 is an *unconditional* `send_replace` and is intentionally outside this pattern — the active-CVV path knows it is the source of truth.)

- **`manage_new_consensus` first-gossip and subsequent branches differ only in `end_number` and the counter op** — The `if first_gossipped_epoch.is_none()` arm (consensus.rs:332-354) and the `else` arm (consensus.rs:355-383) each clone the same four handles, set `*last_number = Some(number + 1)`, adjust `tasks`, and spawn an identically-bodied task that calls `get_consensus_header_range(number, hash, ..., end_number)` then `fetch_sub`. The only real differences: first gossip uses `end_number = 0` with `tasks.store(1)`, subsequent uses `end_number = last_number.unwrap_or_default()` with `tasks.fetch_add(1)` guarded by `task_num < 6`. Compute `end_number` and the throttle decision up front (first gossip = unconditional, `end_number` 0; otherwise gated on `< 6`), then spawn the backfill task once — removing the duplicated 20-line spawn block.

- **Per-worker `Semaphore::new(1)` in `spawn_fetch_consensus` is vestigial** — Each worker creates its own `Arc<Semaphore::new(1)>` (consensus.rs:175) and `next_epoch` acquires a permit before fetching the next request (consensus.rs:171-172). Since each worker is a single sequential task that holds the permit across the whole inner retry loop and only loops back to `next_epoch` after the previous request finishes, the semaphore can never have contention — there is exactly one acquirer. The permit, the `Arc`, and the `SemaphorePermit<'s>` in the `next_epoch` signature can all be dropped; `next_epoch` becomes a thin wrapper over `get_next_epoch_pack_file_request()`. (Lower confidence only insofar as it may be a guard against a future refactor, but as written it does nothing.)

- **Two separate "find incomplete epochs" scans that could share a predicate** — `request_missing_packs` walks *backwards* from `latest_record()` by parent hash collecting the lowest incomplete epoch (consensus.rs:457-467), then hands `first_missing` to `request_epochs`, which walks *forwards* by epoch number re-checking `is_epoch_complete` on every epoch and enqueueing the incomplete ones (consensus.rs:106-121). The backward scan exists only to find a starting point that the forward scan then re-derives. `request_missing_packs` could simply call `request_epochs(&mut 0, ...)` (or seed from the lowest known record) and let the forward walk's existing `is_epoch_complete` filter handle the gaps, deleting the entire backward-by-hash loop. (Lower confidence: the backward walk also tolerates a broken parent chain via `get_epoch_by_hash`, so confirm the forward `record_by_epoch` walk covers the same gap cases before removing it.)

- **`get_consensus_header`'s `Option<(u64, B256)>` return overloads three meanings** — `None` means both "stop, reached settled chain data" (consensus.rs:41) and "stop at genesis" (consensus.rs:44, 68); `Some(parent)` means "continue to parent" (consensus.rs:44, 66); and `Some((number, hash))` returning the *same* input means "retry, I couldn't fetch it" (consensus.rs:82, after a 5s sleep). The caller loop at consensus.rs:295-307 can't distinguish retry from progress. A small enum (`Stop` / `Continue(u64, B256)` / `Retry`) would make the protocol explicit and let the caller decide retry policy rather than burying a `sleep(5s)` inside the fetch helper.

### DIFF
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -299,7 +299,7 @@ where
                 } else {
                     // Not sure we can sanity check this epoch.  However if it is bogus the code
                     // to handle it should be fine, it stops when out of epochs.
-                    self.consensus_bus.request_missing_epoch_if_newer(epoch);
+                    self.consensus_bus.set_request_missing_epoch_if_newer(epoch);
                 }
             }
             PrimaryGossip::EpochVote(vote) => {
@@ -850,10 +850,11 @@ where
         let mut my_number = self.consensus_chain.latest_consensus_number();
         // If we are behind then wait up to two seconds to catch up.
         let mut count = 0;
-        while my_number < number {
+        let delta = number.saturating_sub(my_number);
+        while my_number < number && delta < 4 {
             tokio::time::sleep(Duration::from_millis(100)).await;
             my_number = self.consensus_chain.latest_consensus_number();
-            if count > 20 {
+            if count >= 20 {
                 // Don't wait more than 2 seconds for this to show up.
                 return Err(PrimaryNetworkError::UnknownConsensusHeaderDigest(hash));
             }
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -297,12 +297,9 @@ where
                         self.consensus_certs.lock().insert(consensus_result_hash, 1);
                     }
                 } else {
-                    let latest_missing = *self.consensus_bus.requested_missing_epoch().borrow();
-                    if epoch > latest_missing {
-                        // Not sure we can sanity check this epoch.  However if it is bogus the code
-                        // to handle it should be fine, it stops when out of epochs.
-                        self.consensus_bus.requested_missing_epoch().send_replace(epoch);
-                    }
+                    // Not sure we can sanity check this epoch.  However if it is bogus the code
+                    // to handle it should be fine, it stops when out of epochs.
+                    self.consensus_bus.request_missing_epoch_if_newer(epoch);
                 }
             }
             PrimaryGossip::EpochVote(vote) => {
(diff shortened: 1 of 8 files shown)


crates/consensus/primary/src/network/handler.rs
```diff
@@ -851,10 +848,17 @@ where
         hash: BlockHash,
     ) -> PrimaryNetworkResult<PrimaryResponse> {
         let mut my_number = self.consensus_chain.latest_consensus_number();
-        // If we are one behind then wait to catch up.
-        while my_number + 1 == number {
+        // If we are behind then wait up to two seconds to catch up.
+        let mut count = 0;
+        let delta = number.saturating_sub(my_number);
+        while my_number < number && delta < 4 {
```
grantkee, line 854: nit: `while ... delta < 4` never changes in the loop. The behavior is correct though, it just looks like it might be a bug. I think it's more readable to add this condition as an if gate, but again this is nit
```rust
        let mut my_number = self.consensus_chain.latest_consensus_number();
        // Only wait for a small gap to close; if we are far behind, don't stall the request.
        // When within a few blocks, poll for up to ~2 seconds (20 * 100ms) for the header to show
        // up locally. If it still hasn't arrived, `get_header_by_hash` returns the same
        // `UnknownConsensusHeaderDigest` error the wait would have.
        if number.saturating_sub(my_number) < 4 {
            let mut count = 0;
            while my_number < number && count < 20 {
                tokio::time::sleep(Duration::from_millis(100)).await;
                my_number = self.consensus_chain.latest_consensus_number();
                count += 1;
            }
        }
}

grantkee (approved):
One small nit, but LGTM!

crates/consensus/primary/src/network/handler.rs
sstanfield: Not a bug and was intentional.  I'll fix it in my next PR to avoid another attest/review loop on this.

PR #720, opened 2026-06-05
Separate operator trust from committee (validator) trust on Peer (715)

### LINKED ISSUE #715: Add `is_validator` field to `Peer` to separate committee trust from operator trust
Problem:

  - The Peer struct in `crates/network-libp2p/src/peers/peer.rs` tracks a single is_trusted flag that conflates two distinct trust provenances:
    - peers explicitly allowlisted by node operators
    - validators promoted to trusted status because they sit in the previous, current, or next committee
  - When a validator rotates out of all committee slots, the peer manager clears is_trusted - an operator-allowlisted peer that also served on a committee silently loses its
  operator trust
  - There is no way to distinguish "protected because committee member" from "protected because the operator said so" when making scoring, pruning, or banning decisions

  Proposed solution:

  - Add an is_validator: bool field to Peer alongside is_trusted
  - is_trusted becomes operator-allowlist only and is never touched by committee rotation
  - is_validator is set and cleared by the epoch rotation logic for peers in the prev/current/next committees
  - Validators receive the same protections trusted peers have today (bypass the score model, exempt from severe/fatal penalties) - the effective check becomes is_trusted ||
  is_validator
  - Node operators can maintain trusted peers independently while the protocol protects validators during epoch rotations

follow up to #707

MavenRain:
## Summary

  `Peer` tracked a single `is_trusted` flag that conflated two distinct trust
  provenances:

  - peers explicitly **allowlisted by the node operator**, and
  - peers trusted because they sit in the **committee** (set via
  `Peer::make_trusted` on epoch entry).

  The flag was also *sticky*: `make_trusted` only ever set it to `true` and
  nothing ever cleared it, so once a peer served on a committee it stayed
  exempt from scoring, penalties, pruning, and remained an "important"
  (explicit gossip) peer **forever**, even after rotating out. There was no
  way to tell "protected because committee member" from "protected because the
  operator said so" when making scoring/pruning/banning decisions.

  Closes #715. Follow-up to #707.

  ## Approach

  The linked issue proposed adding a second `is_validator: bool` next to
  `is_trusted`. This PR takes a slightly different route to avoid two failure
  modes of that shape;  boolean blindness at the call sites (`is_trusted ||
  is_validator` repeated and easy to get wrong) and a second denormalized flag
  that has to be kept in lockstep with committee rotation:

  - **`Peer` stores only operator trust.** `is_trusted: bool` becomes
  `operator_allowlisted: bool`, set at construction and **never touched by
  epoch rotation**.
  - **Validator trust is derived, not stored.** A new `TrustBasis { Operator,
  Validator }` names *why* a peer is exempt this epoch.
  `AllPeers::trust_basis(peer_id)` / `exemption(...)` compute it from
  `current_committee` membership (validator) and the peer's
  `operator_allowlisted` flag: `current_committee` is the single source of
  truth, so validator status can never drift out of sync with rotation.
  - **The exemption is threaded into the peer methods that need it.**
  `Peer::apply_penalty`, `heartbeat`, and `ensure_banned` take
  `Option<TrustBasis>`; the four scattered consult sites now go through one
  typed check. The severe/fatal-penalty `warn!` now records *which* provenance
  suppressed the penalty.
  - **`make_trusted` is removed.** Committee entry in `new_epoch` now calls
  `Peer::reset_score_to_max()` (primes the score, stores no trust flag).

  Net: one provenance lives on the peer (operator), the other is derived
  (validator); they can never be conflated, and an operator allowlist can
  never be stripped by committee rotation.

  ## ⚠️ Behavior change

  Because validator exemption is now derived from `current_committee`, **a
  validator that rotates out of the committee (and is not
  operator-allowlisted) loses its exemption immediately**;  it becomes subject
  to scoring/decay, prunable, and no longer "important".  Previously the
  sticky `is_trusted` flag kept it exempt indefinitely.

  This is the intended #715 fix, but note it currently revokes *immediately*:
  the one-epoch grace window for an outgoing committee (so late gossip from
  the previous committee still counts) arrives with #707's
  previous/current/next committee tracking, at which point this same
  derivation yields the grace window automatically.  Operator-allowlisted peers
  are unaffected in all cases.

  ## Scope / sequencing

  This is scoped to the `peers` module; the `PeerManager` public API is
  unchanged (`peer_is_important` / `is_peer_validator` keep their signatures).
  It's based directly on `main` and is independent of #716. When #707 lands,
  the validator derivation picks up prev/next committees with no further
  `Peer` changes.

  ## Testing

  - New regression test `test_committee_rotation_revokes_validator_exemption`:
  a committee validator is exempt from a `Fatal` penalty while in the
  committee, then is banned by the same penalty after rotating out.
  - `cargo clippy -p tn-network-libp2p --all-features --all-targets` clean;
  `cargo test -p tn-network-libp2p` peers suite 73/73.


crates/network-libp2p/src/peers/penalty.md
```diff
@@ -24,7 +24,7 @@ Score deltas are applied in `crates/network-libp2p/src/peers/score.rs:84-100`.
 - Disconnect threshold (`min_score_before_disconnect`): `-20.0`. `crates/config/src/network.rs:442`.
 - Score halflife: `300.0` seconds; decay factor is `e^(-ln(2)/halflife * dt)`. `crates/config/src/network.rs:440`, `crates/network-libp2p/src/peers/score.rs:130-132`.
 - `banned_before_decay_secs`: `30 * 60` (30 min). When a peer crosses the ban threshold, `last_updated` is pushed forward by this duration so the score does not decay during the lockout. `crates/config/src/network.rs:441`, `crates/network-libp2p/src/peers/score.rs:149-154`.
-- Trusted peers skip penalty application entirely — `Peer::apply_penalty` short-circuits when `self.is_trusted`. `crates/network-libp2p/src/peers/peer.rs:158-161`.
+- Exempt peers skip penalty application entirely — `Peer::apply_penalty` short-circuits when the caller passes a `TrustBasis` (operator allowlist or committee validator); validator status is derived from `current_committee` in `AllPeers::trust_basis`, not stored on the peer. `crates/network-libp2p/src/peers/peer.rs`, `crates/network-libp2p/src/peers/all_peers.rs`.
```
grantkee: Thank you for updating this file. Small change here after merging into `main`. Instead of `current_committee`:

> validator status is derived from the three tracked committee slots (previous/current/next) in `AllPeers::trust_basis`, not stored on the peer.

grantkee (dismissed):
Minor correction in the penalty.md after an upstream merge. Everything else LGTM!

sstanfield (dismissed):
LGTM!

sstanfield (approved):

PR #723, opened 2026-06-08
Preemptive refactors for future sync.

sstanfield:
This PR:
- Add serde ConsensusOutput so it can be cached in a DB.
- Make the OnceCell hash on consensus output just a hash that must be set (all fields private now as well).
- Move CommittedSubDag to an inner Arc pattern to simplify it's use (with an outer Arc).


grantkee (approved):
LGTM!

PR #724, opened 2026-06-08
Use specific types for specific digests.

sstanfield:
This has no functional changes.  It introduces a macro to make it easy to add a new specific digest type.  Use it to make specific types for ConsensusHeaders and EpochRecords.  This should make code easier to read and understand and avoid digest based foot guns.


grantkee (approved):
LGTM!

PR #726, opened 2026-06-08
separate protocols for primary and worker (by id)

### LINKED ISSUE #725: Primary and worker networks share protocol names, letting worker records contaminate the primary's peer store
## Problem

A validator runs both a primary and a worker `ConsensusNetwork` as two libp2p swarms inside the same process. Both swarms speak the same wire protocols: `/telcoin-network/0.0.0` for request/response and `/tn-kad/1.0.0` for Kademlia. Nothing at the protocol layer keeps the two roles apart.

Kademlia records are keyed on the validator's BLS public key, and that key is identical for a validator's primary and worker. So when the primary and worker of the same (or a peer's) validator connect across roles, the primary can ingest a *worker's* `NodeRecord` under a BLS key it expects to belong to a primary. Once that record lands in the primary's store, the primary resolves the key to a worker address and dials/RPCs it using primary protocols.

The cross-role session never negotiates cleanly — the peer is reachable at the transport level but speaks the wrong application protocols — so the request fails. The peer manager attributes the failure to the remote peer, penalizes it, and can eventually ban an otherwise-healthy validator. The result is degraded connectivity and peer churn that gets worse as more nodes co-locate primary and worker roles, with no misbehavior by any party.

## Solution

Make the network role the single source of truth for everything that must differ between primary and worker, including the wire-protocol names. Introduce a `NetworkType` enum with `Primary` and `Worker(WorkerId)` variants (replacing the storage-only `KadStoreType`) and have each role derive its own request/response and Kademlia protocol strings:

- primary: `/tn-primary/*` and `/tn-primary-kad/*`
- worker `id`: `/tn-worker-{id}/*` and `/tn-worker-{id}-kad/*`

Because the protocol names now differ by role (and by worker id), cross-role substreams can never negotiate: a worker's pushed kad record is never ingested by a primary, and a misdirected req/res surfaces a clean outbound failure instead of contaminating peer scoring. Worker protocols carry the worker id so that future multi-worker nodes stay isolated from one another as well.

The per-role protocol names should be treated as a peer-compatibility contract — locked by a test — since a silent change would prevent honest peers from connecting. The static `supported_req_res_protocols` config field and the shared `protocol()` helpers are no longer needed, since protocols are now derived from role rather than configured.

grantkee:
- Replace `KadStoreType` with `NetworkType` (`Primary` / `Worker(WorkerId)`); each role derives its own req/res and kad protocol names (`/tn-primary/*`, `/tn-worker-{id}/*`), so cross-role substreams can't negotiate and a worker's BLS-keyed kad record can no longer contaminate a primary's peer store.
- Drop the `supported_req_res_protocols` config field, `protocol_vec` serde, and the shared `protocol()`/`identify_protocol()` helpers — protocols come from role now, not config.
- `new_for_worker` takes a `worker_id`; add `DEFAULT_WORKER_ID = 0` and thread it through the node/epoch wiring.
- Add a cross-role isolation regression test and a test locking the per-role protocol name strings.

closes #725

### DIFF
--- a/crates/config/src/network.rs
+++ b/crates/config/src/network.rs
@@ -1,7 +1,7 @@
 //! Configuration for network variables.
 
 use crate::{ConfigFmt, ConfigTrait, TelcoinDirs};
-use libp2p::{kad::K_VALUE, request_response::ProtocolSupport, StreamProtocol};
+use libp2p::kad::K_VALUE;
 use serde::{Deserialize, Serialize};
 use std::{num::NonZeroUsize, time::Duration};
 use tn_types::Round;
@@ -81,10 +81,6 @@ impl NetworkConfig {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(default)]
 pub struct LibP2pConfig {
-    /// The supported inbound/outbound protocols for request/response behavior.
-    /// - ex) "/telcoin-network/mainnet/0.0.1"
-    #[serde(with = "protocol_vec")]
-    pub supported_req_res_protocols: Vec<(StreamProtocol, ProtocolSupport)>,
     /// Maximum message size between request/response network messages in bytes.
     pub max_rpc_message_size: usize,
     /// Maximum message size for gossipped network messages in bytes.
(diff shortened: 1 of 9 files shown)


sstanfield (approved):
Looks good.

MavenRain (approved):

MavenRain:
Straightforward . . . LGTM

PR #730, opened 2026-06-10
Peer rpc info

### LINKED ISSUE #729: # Advertise validator JSON-RPC endpoints through peer discovery
## Problem

Validators run worker nodes that expose a JSON-RPC endpoint for submitting transactions, but there is no way for a client to find those endpoints. The kademlia node record a validator publishes carries only network-layer information — the libp2p multiaddrs and network key used for the consensus swarm. The application-layer JSON-RPC URL that wallets, dapps, and front-ends actually need is nowhere in the record.

Today an operator who wants to route client traffic across the validator set has to maintain that mapping out of band: a hand-curated list of endpoints, kept in sync by hand as the committee rotates. There is no on-network source of truth, so a gateway or load balancer cannot discover where to send transactions, and a validator that wants to make its endpoint reachable has no way to announce it.

The node record is also signed and encoded with BCS, which is not self-describing. Any change to the record layout has to stay backward compatible: a node running new software must still be able to decode and verify records published by peers running the old software, or peer discovery breaks across the upgrade boundary.

## Solution

Add an optional RPC descriptor to the worker node record so a validator can advertise its JSON-RPC endpoint through the same kademlia record it already publishes for discovery. The descriptor should carry a required HTTP(S) URL and an optional WebSocket URL, and it should be strictly opt-in — primary records and workers that do not expose RPC publicly leave it unset. These URLs are consumed by external clients only; the libp2p swarm should never dial them.

The network handle should expose two lookups — fetch one authority's advertised endpoint, and snapshot every known authority's endpoint — so a worker gateway can discover validators' RPC endpoints and route client traffic across them without any out-of-band configuration.

Advertised URLs should be validated rather than trusted. A signature proves a record is authentic, not that its endpoint is well formed, so validation should happen at more than one layer: fail fast at node startup if the operator's own configured endpoint is invalid, and on the receiving side strip a malformed endpoint from an otherwise-authentic peer record while keeping the rest of the record usable. Validation should reject endpoints with the wrong URL scheme and cap URL length so an oversized value cannot push the record past kademlia's size limit. When a record is too large to store, the node should surface that the endpoint was not advertised rather than failing silently.

Backward compatibility should be handled by decoding against the historical layout when the current decode fails. Because BCS encodes the new trailing `Option` field as a single tag byte and the legacy layout ends with a length-prefixed signature, the two encodings are mutually exclusive, so the fallback is unambiguous. Legacy records must be verified against the bytes they were actually signed over, not re-encoded under the new schema. Records already persisted in the local store should load through the same fallback, and any record that decodes under neither layout should be purged so the store's accounting stays accurate.

grantkee:
# Advertise validator JSON-RPC endpoints over kademlia

- Add an optional `RpcInfo` (required `http`, optional `ws`) to the worker `NodeRecord` so validators can advertise their JSON-RPC endpoint through peer discovery. Opt-in: `None` on primary records and on workers that don't expose RPC.
- New worker network-handle lookups `get_validator_rpc` / `get_all_validator_rpcs` back a worker gateway that discovers advertised endpoints and routes client traffic across the validator set. These URLs are for external clients only and are never dialed by the swarm.
- Validate endpoints in two places: fail-fast at node startup against the operator's own config, and again in `add_known_peer`, which drops a malformed endpoint from an authentic signed record while keeping the rest. Checks cover URL scheme (`http`/`https`, `ws`/`wss`) and a 2048-byte length cap.
- Backward-compatible decoding for pre-upgrade peers: `LegacyNodeRecord` mirrors the old BCS field order, and `try_decode_compat` / `decode_and_verify` fall back to it. The new trailing `Option` tag byte and the legacy signature length prefix can't collide, so the fallback is deterministic; legacy records verify against the bytes they were originally signed over.
- Persisted store records load through the same fallback; undecodable entries are purged on load so kad's record accounting stays accurate.
- Surface kad `ValueTooLarge` explicitly — log that the RPC endpoint was *not* advertised instead of a generic "failed to store record" error.
- Drop the duplicate BLS verification on the kad query path: `peer_record_valid` now verifies once and `process_kad_query_result` trusts the pre-validated record.

closes #729


crates/network-libp2p/src/types.rs
```diff
@@ -639,13 +688,53 @@ pub struct NodeRecord {
...
+    signature: BlsSignature,
+}
+
+impl From<LegacyNodeRecord> for NodeRecord {
+    fn from(legacy: LegacyNodeRecord) -> Self {
+        let LegacyNodeRecord {
+            info: LegacyNetworkInfo { pubkey, multiaddrs, timestamp },
+            signature,
+        } = legacy;
+        Self { info: NetworkInfo { pubkey, multiaddrs, timestamp, rpc: None }, signature }
```
sstanfield, line 722: Probably won't matter with when this is used (sig already verified) but won't this create a record with an invalid signature?  Probably nothing to be done about this even if true as long as we are on the other side of having verified the sig and won't try to do so again.  If this is infrequent should it verify the sig one last time to be sure the legacy record was legit?  This is just commentary unless you think there is anything here.

sstanfield: Nevermind, I see you addressed this decode_and_verify() below.  The saved sig will no longer be valid but once in the store it should be past the trust boundary and OK.  Although if those records get propagated they will be invalid to other nodes now.  Maybe that is also OK since new records should be forthcoming as long as it does not trigger a penalty or ban.

sstanfield (approved):
Looks good.  Not sure about the signatures with a legacy conversion but they may be fine.

crates/network-libp2p/src/consensus.rs
```diff
@@ -411,7 +434,13 @@ where
         info!(target: "network-kad", ?record, "Providing our record to kademlia for peer {:?}", self.swarm.local_peer_id());
         let key = record.key.clone();
         if let Err(err) = self.swarm.behaviour_mut().kademlia.put_record(record, kad::Quorum::One) {
-            error!(target: "network-kad", "Failed to store record locally: {err}");
+            match &err {
+                kad::store::Error::ValueTooLarge => error!(
+                    target: "network-kad",
+                    "node record exceeds kad value-size limit; RPC endpoint NOT advertised to peers ({err})"
+                ),
+                _ => error!(target: "network-kad", "Failed to store record locally: {err}"),
```
MavenRain, line 442: I tend to want to exhaustively match all enums, but don't yet have a handle on the extent to which this is a pattern around here.

MavenRain (approved):

crates/network-libp2p/src/consensus.rs
grantkee: I tend to agree. In this case, the two errors are `MaxRecords` and `MaxProvidedKeys` - both of which have unique error messages. I didn't have anything else to add in the log that would differentiate them so the catch all keeps the code dry. I'm in favor of leaving as is until we have additional/differentiating data to include in the log. Open to exhaustive matching in the future as this code changes

PR #731, opened 2026-06-10
Native BLS proof-of-possession precompile at 0x...b151

Huwonk:
## Summary

Adds a native precompile at the canonical `BLS_G1_LIBRARY` address (`0x000000000000000000000000000000000000b151`) that verifies BLS proofs of possession by delegating to tn-types blst verification, replacing the Solidity `BlsG1` library and its on-chain EIP-2537 reimplementation.

- `crates/tn-reth/src/evm/bls_precompile/`: new `DynPrecompile` implementing the two selectors `ConsensusRegistry` calls, `verifyProofOfPossession` and `proofOfPossessionMessage`. Gas is priced at 150k, matching the equivalent EIP-2537 pairing work. Structured the same way as the existing TEL precompile.
- `crates/tn-reth/src/evm/factory.rs`: registers the precompile at `0x...b151` alongside the TEL precompile, so it is live at that address on all chains and shadows any genesis-deployed `BlsG1` library code.
- `crates/types`: adds uncompressed-bytes constructors for `BlsPublicKey` and `BlsSignature`, used by the precompile to decode calldata without round-tripping through compressed encodings.

The tn-contracts submodule is unchanged in this PR. The follow-up that routes genesis PoP validation through this precompile and adopts the updated `ConsensusRegistry` ABI is stacked on top in a separate PR, since it depends on tn-contracts changes that have not merged yet.

## Testing

- Unit tests in `bls_precompile/mod.rs` cover a valid proof of possession, rejection when the call targets the wrong address, and malformed input.

### DIFF
--- a/crates/tn-reth/src/evm/bls_precompile/mod.rs
+++ b/crates/tn-reth/src/evm/bls_precompile/mod.rs
@@ -1,0 +1,390 @@
+//! Native BLS12-381 proof-of-possession precompile.
+//!
+//! Replaces the Solidity `BlsG1` library (linked at [`BLS_G1_PRECOMPILE_ADDRESS`], `0x…b151`) with
+//! a native implementation that delegates to the **same** `blst` (`min_sig`) verification the
+//! consensus layer uses to *produce* proofs of possession
+//! ([`tn_types::verify_proof_of_possession_bls`]). Having one implementation behind both signing
+//! and verification removes the byte-for-byte drift risk between the Rust signer and an independent
+//! on-chain reimplementation.
+//!
+//! # Structure
+//!
+//! Mirrors [`tel_precompile`](super::tel_precompile): a [`DynPrecompile`] registered via
+//! [`add_bls_precompile`] and dispatched by 4-byte selector. The ABI matches `BlsG1.sol`, so
+//! `ConsensusRegistry`'s existing library `delegatecall`s resolve to this precompile unchanged.
+//!
+//! | Selector | Behavior |
+//! |----------|----------|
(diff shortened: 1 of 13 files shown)


sstanfield (commented):
This looks great, good work.  Leaving this a comment just because I think Grant needs to review this as well since he is the pre compile expert.  But consider this an approval once he reviews.

crates/tn-reth/src/evm/bls_precompile/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+    let pubkey = BlsPublicKey::from_uncompressed_bytes(&decoded.uncompressedPubkey)
+        .map_err(|e| PrecompileError::Other(format!("invalid uncompressed pubkey: {e:?}").into()))?;
+    let message = proof_of_possession_message_bytes(&pubkey, &decoded.validatorAddress)
+        .map_err(|e| PrecompileError::Other(format!("message build failed: {e}").into()))?;
+
+    Ok(PrecompileOutput::new(POP_MESSAGE_GAS_COST, Bytes::from(Bytes::from(message).abi_encode())))
+}
+
+#[cfg(test)]
+mod tests {
```
MavenRain, line 175: How feasible would a `bls_precompile` test be?

Huwonk: I think I could write similar tests to what is written against the current smart contract.  Let me check it out.

MavenRain (approved):

grantkee (approved):
LGTM! This is a solid precursor for overhauling the BLS PoP verification for validator nodes. Good work @Huwonk

Issue #114, opened 2024-11-01
Base Fee


grantkee:
Protocol is currently set to always use the minimum protocol basefee (7 wei).

The complexity comes from latency in the feedback process. Batches aren't executed until after consensus, so adjustments aren't made until a few rounds later.

Once the output is executed, the protocol can adjust base fees for the entire network by calculating:
- target gas used in each batch
- total gas used in each batch
- total possible number of batches for each round (based on worker, time, and max number of batches)
- total possible number of authorities proposing blocks for the round

However, it's unclear how these values will change during periods of high and low network activity.

Ethereum adjusts base-fees for the entire protocol, but it also uses a shared mempool.

TN has individual mempools, so capturing network activity is a little more complex.

If workers managed basefees locally, there could be unintended consequences. It's reasonable to assume users would chase the lowest base fees available. This could lead to huge spikes in transactions for specific workers between batches. It would also incentivize users to submit duplicate transactions and limits the effectiveness of simple load balancers for the network.

Instead, a network-wide base fee adjustment seems more desirable albeit more complicated to figure out.

### Adjusting after consensus

After a successful commit to the subdag, consensus chain produces output for the next wave. The protocol can't reliably determine if batches were maxed out because there are min/max time intervals at play.

For example, worker tries to produce a batch every 1s. Primary seals a block between 5-10s. Accounting for computation and voting latency, a worker could potentially reach quorum on 4 batches before the primary's min interval (5s) is reached. If network activity is very high, these batches should be full. 

What if another worker isn't receiving any transactions and produces no blocks for the round?

Should the protocol take the average and leave basefees alone? The node with full batches should increase basefees, but the node without any transactions might want to lower basefees.

### Voting for basefee adjustments

Workers could include a simple vote with each batch to increase, decrease, or stay the same. Then, votes are tallied after execution and applied to the basefee.

The only problem with this approach is that the DAG commits a leader, then executes previous rounds.

For example, commit `12` would have a leader from round 12, but all other children would be from previous rounds (8, 9, 10, etc). This introduces significant latency that could result in basefee adjustments that undermine the protocols success. If rounds 8 and 9 were extremely high volume, but 10, 11, and 12 were normal the protocol might decide to still increase basefees. This is expected to reduced transactions, but rounds 10, 11, and 12 already indicate that network activity was slowing down to begin with.

[...]

PR #732, opened 2026-06-10
Route genesis PoP through the BLS precompile and adopt the integration ConsensusRegistry ABI

Huwonk:
## Blocked: do not merge until tn-contracts lands

This PR bumps the tn-contracts submodule to `933703c` on `integration/e2e-0835-sets-transient`, which combines tn-contracts #114 (per-status EnumerableSet validator indexing), #118 (transient committee-membership set), and the solc 0.8.35 bump. Before this can merge into main we need to:

1. Merge those changes into tn-contracts master.
2. Re-point the submodule here to the resulting master commit.

It is stacked on #731 (the native BLS precompile), so the base is `feat/bls-precompile`; GitHub will retarget to main automatically once #731 merges.

## Summary

- `crates/tn-reth/src/lib.rs`: routes genesis proof-of-possession validation through the new BLS precompile at `0x...b151` instead of validating natively, so genesis and runtime validation share one code path, and adopts the updated `ConsensusRegistry` ABI from the integration branch.
- `crates/tn-reth/src/evm/block.rs` and `system_calls.rs`: updated for the new ABI and PoP flow.
- `crates/e2e-tests/tests/it/genesis_tests.rs`: updated for the new genesis flow.
- `tn-contracts`: submodule bumped to the integration branch (see blocker above).

## Testing

- e2e genesis tests updated and passing against the integration ConsensusRegistry.

### DIFF
--- a/crates/execution/tn-rpc/src/rpc_ext.rs
+++ b/crates/execution/tn-rpc/src/rpc_ext.rs
@@ -242,9 +242,39 @@ where
         &self,
         status: ConsensusRegistry::ValidatorStatus,
     ) -> TelcoinNetworkRpcResult<Vec<ConsensusRegistry::ValidatorInfo>> {
-        let calldata =
-            ConsensusRegistry::getValidatorsCall { status: status as u8 }.abi_encode().into();
-        self.registry_read(calldata).await
+        use ConsensusRegistry::ValidatorStatus;
+
+        // `getValidatorsInfo` returns exactly one status set and reverts on the `Undefined`/`Any`
+        // sentinels (the registry no longer folds statuses on-chain). Emulate the documented
+        // `"Any"` => "all validators" behavior by unioning every concrete status set off-chain.
+        // Each validator lives in exactly one set, so the union is a plain concatenation (no
+        // dedup); retired validators intentionally appear in no set, matching the old scan. Any
+        // other status (including `Undefined`) is passed straight through, so `Undefined` still
+        // surfaces the on-chain revert that callers rely on.
+        if status == ValidatorStatus::Any {
+            let mut all = Vec::new();
+            for set_status in [
+                ValidatorStatus::Staked,
+                ValidatorStatus::PendingActivation,
+                ValidatorStatus::Active,
+                ValidatorStatus::PendingExit,
+                ValidatorStatus::Exited,
+            ] {
+                let calldata =
+                    ConsensusRegistry::getValidatorsInfoCall { status: set_status as u8 }
+                        .abi_encode()
+                        .into();
+                let mut infos: Vec<ConsensusRegistry::ValidatorInfo> =
+                    self.registry_read(calldata).await?;
+                all.append(&mut infos);
+            }
+            Ok(all)
+        } else {
+            let calldata = ConsensusRegistry::getValidatorsInfoCall { status: status as u8 }
+                .abi_encode()
+                .into();
+            self.registry_read(calldata).await
+        }
     }
 
     async fn get_committee_validators(
--- a/crates/execution/tn-rpc/src/rpc_ext.rs
+++ b/crates/execution/tn-rpc/src/rpc_ext.rs
@@ -8,7 +8,11 @@ use async_trait::async_trait;
 use jsonrpsee::proc_macros::rpc;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
-use tn_reth::{error::RegistryReadError, system_calls::ConsensusRegistry, RethEnv};
+use tn_reth::{
+    error::{RegistryReadError, RegistryReadResult},
+    system_calls::ConsensusRegistry,
+    RethEnv,
+};
 use tn_types::{
     Address, Bytes, ConsensusHeader, Epoch, EpochCertificate, EpochDigest, EpochRecord, Genesis,
     SolCall, SolType, SolValue, B256, U256,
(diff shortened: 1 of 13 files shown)


grantkee (dismissed):

crates/execution/tn-rpc/src/rpc_ext.rs
```diff
@@ -242,9 +242,39 @@ where
...
     ) -> TelcoinNetworkRpcResult<Vec<ConsensusRegistry::ValidatorInfo>> {
-        let calldata =
-            ConsensusRegistry::getValidatorsCall { status: status as u8 }.abi_encode().into();
-        self.registry_read(calldata).await
+        use ConsensusRegistry::ValidatorStatus;
+
+        // `getValidatorsInfo` returns exactly one status set and reverts on the `Undefined`/`Any`
+        // sentinels (the registry no longer folds statuses on-chain). Emulate the documented
+        // `"Any"` => "all validators" behavior by unioning every concrete status set off-chain.
+        // Each validator lives in exactly one set, so the union is a plain concatenation (no
```
MavenRain, line 256: This invariant is only enforced on-chain.  If a validator ever appears under two statuses (e.g. a registry bug or state observed in-transition), then that validator will be double-counted.

Consider deduplicating before shuffling.

crates/execution/tn-rpc/src/rpc_ext.rs
```diff
@@ -242,9 +242,39 @@ where
...
+                ValidatorStatus::Active,
+                ValidatorStatus::PendingExit,
+                ValidatorStatus::Exited,
+            ] {
+                let calldata =
+                    ConsensusRegistry::getValidatorsInfoCall { status: set_status as u8 }
+                        .abi_encode()
+                        .into();
+                let mut infos: Vec<ConsensusRegistry::ValidatorInfo> =
+                    self.registry_read(calldata).await?;
```
MavenRain: These five calls are sequential against the canonical tip, yeah?  So, a validator that transitions status between call N and call N+1 seems like it would get double-counted.

Consider pinning all five reads to a single block hash/number snapshot taken before the loop.

crates/types/src/crypto/bls_signature.rs
```diff
@@ -35,10 +35,8 @@ impl BlsSignature {
     /// uncompressed bytes the protocol passes to the on-chain `BlsG1` library as a proof of
     /// possession. Used by the native BLS precompile so it can verify the exact bytes the
     /// consensus layer produced, without re-implementing point (de)compression.
-    pub fn from_uncompressed_bytes(bytes: &[u8]) -> eyre::Result<Self> {
-        let sig = CoreBlsSignature::deserialize(bytes)
-            .map_err(|_| eyre::eyre!("Invalid uncompressed signature bytes!"))?;
-        Ok(Self(sig))
+    pub fn from_uncompressed_bytes(bytes: &[u8]) -> Result<Self, blst::BLST_ERROR> {
```
MavenRain: `blst::BLST_ERROR` is a C-FFI error with no `std::error::Error` impl.  This will break downstream consumers writing idiomatic error handling (i.e. `?` operator, combinators, etc.).

This contract change may actually be intended, but it is hard to tell from the context, and the old implementation could have been simplified as follows:

```
CoreBlsSignature::deserialize(bytes).map_err(|_| eyre::eyre!("Invalid uncompressed signature bytes!")).map(Self)
```

crates/tn-reth/src/lib.rs
```diff
@@ -1273,17 +1259,16 @@ impl RethEnv {
...
             Self::fetch_value_from_json_str(ISSUANCE_JSON, Some("deployedBytecode.object"))?;
         let issuance_runtimecode =
             hex::decode(issuance_json_binding.as_str().ok_or_eyre("invalid issuance json")?)?;
         let genesis = genesis.extend_accounts([
-            (blsg1_address, GenesisAccount::default().with_code(Some(blsg1_runtimecode.into()))),
+            // The BLS proof-of-possession library is a precompile at `blsg1_address`
+            // (`BLS_G1_PRECOMPILE_ADDRESS`). Mirror the TEL precompile and give it a single `0xfe`
+            // (INVALID) byte of code so the account is non-empty (never state-pruned) and any call
+            // that bypasses precompile dispatch reverts instead of succeeding against an EOA.
+            (blsg1_address, GenesisAccount::default().with_code(Some(vec![0xfe_u8].into()))),
```
MavenRain: Instead of a magic sentinal, consider a more formal `const PRECOMPILE_SENTINEL_CODE: &[u8] = &[0xfe];` that can also be reused in the tests.

crates/execution/tn-rpc/src/rpc_ext.rs
grantkee: Agreed - we have pretty significant testing around this. The status is only held in `ValidatorInfo` which is only used in a map (`address` is the key). As a result, it's impossible to have an address associated with multiple status

crates/tn-reth/src/lib.rs
grantkee: Great idea!

sstanfield (dismissed):

crates/types/src/crypto/bls_signature.rs
grantkee: There was only one call site, but I agree with your concern. I reverted back to the `eyre::Result` approach pending follow up PR to only used the compressed BLS bytes. Thanks for flagging

crates/execution/tn-rpc/src/rpc_ext.rs
grantkee: I was initially concerned about gas limits, but this does present a race condition that could return inconsistent results across different canonical tips. I will address it

MavenRain (approved):
All good

PR #733, opened 2026-06-10
Metrics

grantkee:
- New `tn-metrics` crate: a global recorder that passes `tn_*` names through and prefixes
  everything else with `reth_` (stock-reth dashboards keep working), plus a minimal raw-TCP
  `/metrics` endpoint modeled on the healthcheck server — pre-scrape hooks sample process
  stats and reth db table sizes, with 5s registry upkeep.
- `--metrics <addr>` now actually serves the endpoint. It was previously parsed and threaded
  into `TnBuilder` but never consumed. The recorder is installed before the reth database is
  constructed (handles bind to the live recorder at construction); a bind failure fails
  startup rather than degrading silently.
- Per-crate `tn_*` instrumentation across worker, batch-builder, primary, executor, engine,
  network-libp2p (`network={primary,worker}` label), state-sync, the epoch manager, and the
  node (`tn_node_mode`, `tn_node_consensus_sync_distance`).
- Consensus watch channels (round, committed round, heights, node mode) are mirrored into
  gauges by one node-lifetime `spawn_bus_metrics_mirror` task instead of instrumenting every
  send site — rot-proof and only needs scrape-interval freshness.
- `metrics` bumped 0.23 → 0.24 to match reth (a version skew silently splits the global
  registry); added `metrics-exporter-prometheus`/`metrics-process`/`metrics-util`/
  `reth-metrics`; removed the now-dead `axum` "consensus metrics" dep and the unused
  `PrometheusMetricsParameters` config type.
- Tests: e2e asserts the endpoint serves both `tn_*` and `reth_*` on a live validator;
  per-crate unit tests use a local `DebuggingRecorder` (only `tn-metrics`'s own test installs
  the global recorder).
- Grafana dashboard at `etc/grafana/telcoin-node-metrics.json` plus operator docs (tn-metrics
  README + CLI README); docs note `--instance` does not offset `--metrics`/`--healthcheck`.


crates/tn-metrics/src/server.rs
```diff
@@ -0,0 +1,196 @@
...
+                            // unread bytes in the kernel buffer sends RST instead of FIN,
+                            // resetting the response mid-flight on the scraper's side. the
+                            // timeout bounds idle probes that connect but never send.
+                            let mut buf = [0u8; 1024];
+                            let _ = time::timeout(
+                                Duration::from_secs(2),
+                                socket.read(&mut buf),
+                            )
+                            .await;
+                            hooks.run();
```
MavenRain: Does `hooks.run` run synchronously?  If so, it could pin a worker for its duration.

crates/tn-metrics/src/server.rs
```diff
@@ -0,0 +1,196 @@
...
+    // IMPORTANT: use firewall to protect this endpoint
+    let listener = TcpListener::bind(addr).await?;
+    let listen_on = listener.local_addr()?;
+
+    // static info series so dashboards can surface the running version
+    metrics::gauge!("tn_info", "version" => version).set(1.0);
+
+    info!(target: "tn::metrics", ?listen_on, "prometheus metrics endpoint listening");
+
+    task_spawner.spawn_critical_task("metrics", async move {
```
MavenRain: Any reason why this isn't just `spawn_task` instead of `spawn_critical_task`?  Do we want a monitoring choke to take down the node?

MavenRain (dismissed):

crates/tn-metrics/src/server.rs
grantkee: Good point. I'd rather metrics crash than the entire node

crates/tn-metrics/src/server.rs
grantkee: Yes, good point. Looking into this more I think I have a better solution

MavenRain (approved):
LGTM

sstanfield (approved):

PR #734, opened 2026-06-10
db debug tool command

grantkee:
- `db` command for CLI to validate epoch-pack files


crates/storage/src/consensus_pack.rs
```diff
@@ -1185,6 +1133,61 @@ impl Inner {
     }
 }
 
+/// Verify a streamed [`EpochMeta`] record links correctly to the previous epoch's record.
+///
+/// Extracted from [`Inner::stream_import`] as a free function (it is stateless) so the offline pack
```
sstanfield, line 1170: You can probably drop this part of the comment.  After this PR I don't think this will have any meaning.

crates/storage/src/pack_validate.rs
```diff
@@ -0,0 +1,568 @@
...
+) -> Result<PackValidationReport, PackError> {
+    // Read-only open of just the data file — `Pack::open` loads/cross-checks the header (the wrong
+    // epoch fails here with an open error) and needs no sidecar index files.
+    let pack = Pack::<PackRecord>::open(path, epoch as u64, true, PackCompression::ZStd)?;
+
+    // ---- Pass 1: global batch presence. ----
+    // Collect every batch digest present anywhere in the file. This is what lets us tell an
+    // *absent* batch (a real data gap) apart from a *misordered* one (present, wrong group).
+    let mut all_batch_digests: HashSet<BlockHash> = HashSet::new();
+    for record in pack.raw_iter().map_err(|e| PackError::ReadError(e.to_string()))? {
```
sstanfield: No big deal for a diag tool but I don't love the double iteration. but this is just a debug tool so we can optimize later if/when it gets too slow (which may never happen).

sstanfield (dismissed):
Nice, this will come in handy.

crates/storage/src/consensus_pack.rs
grantkee: Agreed and removed

crates/storage/src/pack_validate.rs
grantkee: Good point. I went ahead and fixed it

crates/storage/src/pack_validate.rs
```diff
@@ -0,0 +1,582 @@
...
+                batch_count += 1;
+                // Compute the (re-encode + hash) digest once and record it in both the per-group
+                // set and the persistent global set.
+                let digest = batch.digest();
+                batches.insert(digest);
+                all_batch_digests.insert(digest);
+            }
+            PackRecord::Consensus(consensus_header) => {
+                consensus_count += 1;
+                let number = consensus_header.number;
```
MavenRain, line 251: Out of curiosity, what happens if there is a mis-numbered pack?  Like, what if there is a skip?

crates/storage/src/pack_validate.rs
```diff
@@ -0,0 +1,582 @@
...
+        /// Consensus number that closed the group containing the orphan batch.
+        number: u64,
+        /// The unreferenced batch digest.
+        digest: BlockHash,
+    },
+    /// A consensus number exceeds the epoch's known final consensus number.
+    ///
+    /// Only emitted when a final number is known (not for a bare-file validate, which has no
+    /// authoritative final number to compare against).
+    ConsensusNumberTooHigh {
```
MavenRain: Is this variant ever constructed?

MavenRain (dismissed):
A couple of questions, but nothing blocking

crates/storage/src/pack_validate.rs
grantkee: The number is part of parent's hash so it would be caught indirectly. I went ahead and added an explicit check

crates/storage/src/pack_validate.rs
grantkee: It was never constructed. I was thinking it might be useful in the future, but I'm in favor of removing dead code. We can add this back if/when we need it. Thanks for flagging!

sstanfield (approved):

PR #736, opened 2026-06-10
Epoch Info RPC 

### LINKED ISSUE #735: Expose on-chain ConsensusRegistry state through the `tn` RPC namespace
## Problem

The `tn` RPC namespace only serves consensus-layer epoch records read from the node's database (`tn_epochRecord` and friends). Nothing in the namespace reads the on-chain `ConsensusRegistry`: committee membership, validator status and info, accrued rewards, stake config versions, BLS public keys, delegation digests, or proof-of-possession messages. The node already reads this exact state internally when it builds committees, so the capability exists inside the process — it just isn't reachable from outside.

Clients that need registry state — block explorers, validator dashboards, delegation flows, monitoring — have no choice but to fall back to raw `eth_call` against the registry address with hand-encoded ABI calldata. That ties every client to the contract's ABI and deployed address, and offers no typed or documented surface to build against. A delegator that wants the digest to sign, or a dashboard that wants the current committee, has to reimplement the encoding the node already knows.

Error behavior compounds the problem. The existing `tn` errors map onto HTTP-style `401` codes that don't follow `eth_call` conventions, so a client can't distinguish an on-chain revert (a bad argument) from a missing record from an internal node fault. There's no revert data to decode and no stable code to branch on.

## Solution

Expose each `ConsensusRegistry` view function as its own `tn_*` method — epoch info, validators by status, committee membership, per-validator info, rewards, balance breakdown, BLS pubkeys, delegation digest, proof-of-possession message, stake version, and undistributed issuance. Each method ABI-encodes its call and resolves against EVM state at the canonical tip, returning typed results rather than opaque bytes.

The read path should funnel through a single `read_consensus_registry<T>` helper on `RethEnv` that builds an EVM at the tip, executes the call, and decodes the result. The existing per-purpose readers (committee validators, committee BLS pubkeys) should fold onto this helper so there is one code path for every registry read.

Errors should follow `eth_call` conventions so clients can branch on them. An on-chain revert should surface as code `3` with the raw revert bytes in the `data` field; a missing record should map to EIP-1474 resource-not-found (`-32001`); an internal fault should be logged server-side and return a generic internal error (`-32603`) without leaking node detail to the caller.

Because each read is synchronous database and CPU work executed on a blocking thread, RPC traffic could otherwise exhaust the runtime's blocking pool or contend with consensus-critical reads. A semaphore at the RPC layer should bound the number of concurrent registry reads — held for the lifetime of the blocking work, queueing rather than rejecting at capacity — mirroring reth's own `eth_call` guard while staying well below it so the shared pool keeps headroom.

grantkee:
# Expose ConsensusRegistry view functions through the `tn` namespace

- Add `tn_*` methods for every `ConsensusRegistry` view function: epoch info, validators by status, committee, per-validator info/rewards/balance breakdown, BLS pubkeys, delegation digest, proof-of-possession message, stake version, and undistributed issuance. Each reads on-chain state at the canonical tip and returns typed results.
- Fold `validators_for_epoch` / `bls_pubkeys_for_epoch` onto a shared `read_consensus_registry<T>` helper on `RethEnv`, so there's one EVM-build-and-decode path for all registry reads.
- Standardize `tn` errors to match `eth_call`: on-chain reverts → code `3` with revert bytes in `data`; missing records → `-32001`; internal faults → `-32603`, logged server-side only. Drops the old `401`-style `InvalidProofOfPossession`.
- Bound concurrent registry reads with a 64-permit semaphore at the RPC layer, held across the blocking read and queueing (not rejecting) at capacity. Keeps RPC load from exhausting tokio's blocking pool or throttling consensus-critical reads; sits below reth's `eth_call` guard since they share the pool.
- Derive `serde` + `PartialEq` on `ValidatorStatus` / `ValidatorInfo` / `EpochInfo`, and add the missing registry bindings (`isValidator`, `isDelegated`, `isRetired`, `getBalanceBreakdown`, `delegationDigest`, `getCurrentStakeVersion`, `undistributedIssuance`).
- e2e: cross-check the `tn` endpoints against direct `eth_call` reads at the tip, covering a revert case, a not-found case, and a 192-request burst against the 64-permit guard.

closes #735


sstanfield (approved):
LGTM!

MavenRain (approved):

PR #737, opened 2026-06-11
Breakdown epoch manager

grantkee:
- no functional changes
- organize epoch manager into start/middle/close epoch to reduce the LOC per mod


crates/node/src/manager/node.rs
```diff
@@ -28,36 +34,44 @@ use tn_worker::{WorkerNetworkHandle, WorkerRequest, WorkerResponse};
...
+/// small amount of cross-epoch carry-over state that the next epoch needs to start correctly -
+/// notably [`last_consensus_header`](Self::last_consensus_header),
+/// [`last_forwarded_consensus_number`](Self::last_forwarded_consensus_number), and
+/// [`network_initialized`](Self::network_initialized). Per-epoch consensus components are built and
+/// dropped inside the epoch loop rather than stored here.
 #[derive(Debug)]
 pub(crate) struct EpochManager<P, DB> {
     /// The builder for node configuration
-    builder: TnBuilder,
+    pub(super) builder: TnBuilder,
```
sstanfield: This was organized in the last epoch refactor so we could avoid all this pub(super) noise on everything (why epoch lived in the node dir/crate).  I would prefer not having all this junk but not going to die on that hill if you want the file org that requires it.  Also provided a little tighter encapsulation but not a lot of other code can touch it anyway.

sstanfield (dismissed):
Good refactor.  I would prefer if the files lived in the deleted node directory so we did not have to litter the code with pub(super) everywhere.  Maybe change node to a better name if you find the dir name offensive?  This was something I figured out for the last refactor just to avoid this.

crates/node/src/manager/node.rs
grantkee: I removed as many `pub(super)` as I could. My goal was to break the single `node/epoch.rs` from ~1500 LOC into smaller chunks to reason about. I noticed redundant code during the refactor, but I'm hesitant to refactor more heavily at this stage since so many race conditions were already addressed with the current code

crates/node/src/manager/node/close_epoch.rs
```diff
@@ -0,0 +1,275 @@
+//! Epoch teardown.
+//!
+//! The methods here form the closing half of [`EpochManager`]'s per-epoch
+//! lifecycle. `run_epoch` (in the sibling `epochs` module) calls them in
```
MavenRain: Sibling module appears to no longer exist

crates/node/src/manager/node.rs
```diff
@@ -1,8 +1,14 @@
...
-//! This oversees the tasks that run for each epoch. Some consensus-related
-//! tasks run for one epoch. Other resources are shared across epochs.
-//! This file defines the struct and the node/application scoped code.
+//! This file owns the [`EpochManager`] struct and the resources that live for the entire process:
+//! the long-running primary/worker p2p networks, the execution engine, the consensus DB, the
+//! consensus chain (epoch pack files), and the app-scoped fetch/collector tasks. It also drives
+//! the epoch loop: `run` builds the process-lifetime components, then `run_epochs` repeatedly
+//! invokes `run_epoch` until shutdown.
+//!
+//! Per-epoch orchestration lives in the sibling `epochs` module. Code here is concerned with
```
MavenRain: Sibling module appears to no longer exist

MavenRain (dismissed):

crates/node/src/manager/node/close_epoch.rs
grantkee: Good catch thanks @MavenRain ! Updated

crates/node/src/manager/node.rs
grantkee: Good catch thanks @MavenRain !

sstanfield (approved):
Looks good, like the removed pub(super) noise.

PR #740, opened 2026-06-11
Re-key bls_by_peer_id on network-key rotation

### LINKED ISSUE #722: Network-key rotation breaks the bls_by_peer_id iff-invariant (stale entries, counter loss, zombie records)
Flagged by @grantkee during review of #716

## Summary

PR #716 ("Key `AllPeers` by domain identity instead of libp2p `PeerId`") re-keys the peer collection by domain identity and adds a `bls_by_peer_id` resolution index. The re-key holds on the common paths (anonymous-inbound to `Confirmed` migration and epoch transitions), but `bls_by_peer_id` is **not** maintained when a peer rotates its network key while keeping the same BLS key.  This is a supported flow (`add_known_peer` accepts any newer signature-valid kad record and calls `upsert_peer`), and a single rotation produces three distinct bugs: a stale index entry, skipped status-counter bookkeeping, and a self-perpetuating "zombie" record under a `Confirmed` key.

This code is on the `peerid-to-peeridentity` branch and is **not yet on `main`**; this is a follow-up cleanup before #716 lands.  Network-key rotation is not exercised by any existing test, which is why the gap was not caught.

## The invariant

`bls_by_peer_id` is the single inbound boundary resolver mapping a libp2p `PeerId` to the `BlsPublicKey` of a `Confirmed` peer. Its documented contract (`crates/network-libp2p/src/peers/all_peers.rs:46-52`) is an iff:

> It is kept in lockstep with the `Confirmed` entries in `peers`: an entry exists here **iff** the peer is stored under a `[PeerIdentity::Confirmed]` key.

This is bidirectional: an entry `peer_id -> bls` exists **iff** `peers` holds a `Confirmed(bls)` record whose derived `PeerId` is `peer_id`. `identity_for` (`all_peers.rs:94-100`) trusts this map to decide whether a connection event resolves to `Confirmed` or `Unidentified`, and `evict` (`all_peers.rs:114-122`) is the **only** site that removes from it.

## Reproduction scenario

A known peer publishes a newer, signature-valid kad record carrying the **same** BLS key `B` but a **new** network key `N2` (a realistic operator action: network-keypair seed rotation mid-epoch).  The peer is already stored as `Confirmed(B)` from network key `N1`, with `bls_by_peer_id[peer_id(N1)] = B`.

`add_known_peer` accepts the newer record and calls `upsert_peer(B, N2, …)`. Because `N2` derives a brand-new `PeerId` (`peer_id(N2)`) that has never been seen, `identity_for(peer_id(N2))` returns `Unidentified(peer_id(N2))` . . . it does not resolve to the existing `Confirmed(B)` record.

### 1. Stale entry on rotation

**Location:** `all_peers.rs:143-162` (`upsert_peer`); same missing step in `all_peers.rs:127-140` (`add_trusted_peer`).

**Mechanism:** `upsert_peer` inserts `bls_by_peer_id[peer_id(N2)] = B` (line 160) but never removes the prior `bls_by_peer_id[peer_id(N1)] = B`. The only removal site is `evict` (line 118), which keys on the record's **current** `peer_id` (`peer.peer_id()`), so it can only ever reach the latest entry . . . the old `peer_id(N1)` mapping is orphaned. `add_trusted_peer` has the same defect independently (the fix must address both methods, not via shared code): it removes only the `Unidentified(peer_id)` record (line 137) and inserts the new mapping (line 138), never clearing a prior entry.

**Impact:** Both `peer_id(N1)` and `peer_id(N2)` map to `B` while only one record exists, violating the iff-invariant. Stale entries accumulate across rotations. Because `is_peer_cvv` (`all_peers.rs:708-715`) resolves via `identity_for`, `is_peer_cvv(peer_id(N1))` keeps reporting the rotated-away id as a current committee member as long as `B` is in `current_committee`.

### 2. Counter loss

**Location:** `all_peers.rs:152-161` (`upsert_peer`).

**Mechanism:** Line 152 calls `identity_for(peer_id(N2))`, which returns `Unidentified(peer_id(N2))` because `N2` is unseen. Line 153 (`self.peers.remove(&current)`) therefore looks under the wrong key and **misses** the record already stored under `Confirmed(B)`, returning `None`. A fresh `Peer::new(...)` is built (line 158) and inserted under `Confirmed(B)` (line 161) via a raw `HashMap` insert, silently overwriting the prior record. That insert bypasses the proper state-transition path (`identity_for` + `get_mut`, e.g. `handle_banned_transition` at `all_peers.rs:598-603`), so if the displaced record was `Disconnected` or `Banned`, `disconnected_peers` / `banned_peers` are never decremented.

**Impact:** The `disconnected_peers` / banned counters are permanently skewed. Pruning math depends on them (`prune_disconnected_peers` at `all_peers.rs:880-898` reads `disconnected_peers` to compute excess), so the skew permanently distorts pruning thresholds.

### 3. Zombie record

**Location:** `all_peers.rs:217-252` (`ensure_peer_exists`); `all_peers.rs:114-122` (`evict`); `peers/peer.rs:135-142` (`peer_id`).

**Mechanism:** Once the stale entry from Problem 1 exists and the live `Confirmed(B)` record is later pruned, a libp2p event carrying `peer_id(N1)` resolves (via `identity_for`) to `Confirmed(B)`, finds no record (`!self.peers.contains_key(&id)`), and inserts `Peer::default()` under the `Confirmed(B)` key (`all_peers.rs:235`). `Peer::default()` carries neither a network key nor a bls key, so the record both structurally mismatches the `Confirmed(B)` key it is stored under and returns `None` from `peer.peer_id()` (`peer.rs:140-141`), directly violating the guarantee documented on `peer_id` ("any peer with a recorded bls key … also has a recoverable `PeerId`").  The state is self-perpetuating: `evict` can only clear the index entry when `peer.peer_id()` is `Some` (`all_peers.rs:117-119`), so it can never remove the stale `peer_id(N1)` mapping, and the next event from `peer_id(N1)` recreates the keyless record.

**Impact:** A `Confirmed` record exists whose `peer_id_for` (`all_peers.rs:106-111`) returns `None`. Several `peer_id_for` `None` paths that callers assume are unreachable for `Confirmed` peers become live . . . `heartbeat_maintenance`, `update_peer_scores`, `connected_peer_ids`, `connected_or_dialing_peers`, `connected_peers_by_score_and_routability`, and `prune_banned_peers` (call sites around `all_peers.rs:269, 298, 737, 749, 790, 868`).

## Root cause

`bls_by_peer_id` has no removal or re-key path triggered by network-key rotation: `upsert_peer` / `add_trusted_peer` only ever insert, and the sole removal site (`evict`) keys on the record's **current** `peer_id`, so it can never reach an entry left behind by a prior network key. All three bugs are downstream of this single missing rotation-cleanup step.

## Suggested test design

No test in `crates/network-libp2p` exercises network-key rotation; the peer tests in `crates/network-libp2p/src/tests/peers.rs` cover anonymous-to-confirmed promotion and the full status-transition matrix, but never call `upsert_peer` twice with the same BLS key and differing network keys.

Add a rotation test that:

1. Creates a peer with BLS key `B` and network key `N1` (`upsert_peer(B, N1, …)`), then drives it into a non-default status (e.g. `Disconnected` or `Banned`) so counter bookkeeping is observable.
2. Rotates by calling `upsert_peer(B, N2, …)` with `N2 != N1`.
3. Asserts the iff-invariant holds: `peer_id(N1)` no longer resolves to the peer (no stale `bls_by_peer_id` entry), `peer_id(N2)` does, and exactly one `Confirmed(B)` record exists in `peers`. **Also asserts** `disconnected_peers` / `banned_peers` are unchanged by the rotation (Problem 2: would otherwise pass on the structural checks alone).
4. Repeats the structural assertion **after the peer is evicted/pruned**, confirming no orphaned index entry remains and no keyless `Peer::default()` can be resurrected under `Confirmed(B)` by a subsequent event carrying `peer_id(N1)`.

Helper setup can reuse the keypair-generation pattern from `test_peer_exchange` (`tests/peers.rs:108`): generate two `NetworkKeypair`s and reuse a single `BlsKeypair`. A test of this shape catches all three problems.

## Acceptance criteria

- [ ] `upsert_peer` and `add_trusted_peer` re-key `bls_by_peer_id` on network-key rotation: the prior `PeerId -> bls` entry is removed (e.g. by evicting the existing `Confirmed(bls)` record before inserting the rotated peer) so no stale entry survives (Problem 1 restored).
- [ ] When `upsert_peer` displaces an existing `Confirmed(bls)` record due to network-key rotation, the old record is removed through the proper path — `evict(&Confirmed(bls))` before the new peer is inserted — so `disconnected_peers` / `banned_peers` are decremented rather than dropped by the raw overwrite (Problem 2).
- [ ] `ensure_peer_exists` never inserts a keyless `Peer::default()` under a `Confirmed` key (no zombie record reachable from a stale index entry) (Problem 3).
- [ ] A rotation test (same BLS key, new network key) is added to `crates/network-libp2p/src/tests/peers.rs` asserting the `bls_by_peer_id` <-> `Confirmed` iff-invariant and the status counters before and after eviction, and the test passes.

MavenRain:
Fixes #722.

`bls_by_peer_id` is documented to stay in lockstep with the `Confirmed` entries in `peers`: an entry exists iff the peer is stored under a `Confirmed` key whose record derives that `PeerId`.  Network-key rotation (`upsert_peer` with the same bls key and a new network key, a supported `add_known_peer` flow) broke the invariant three ways: the index entry derived from the previous network key was never removed (the only removal site, `evict`, keys on the record's current peer id); the existing `Confirmed` record was raw-overwritten, skipping `disconnected_peers`/`banned_peers` bookkeeping; and once the live record was later pruned, the stale entry resurrected a keyless `Peer::default()` under the `Confirmed` key, which `evict` can never clear because the record has no recoverable peer id.

`upsert_peer` now treats rotation explicitly.  A record already stored under the confirmed identity but unreachable from the new peer id is displaced through `evict`, clearing the stale index entry.  In the common rotation case the displaced record carries forward via `update_net` (same domain peer, new transport key), so accumulated score, trust, and Banned/Disconnected status survive; in particular a banned peer cannot launder its reputation by rotating its network key.  Transport-liveness statuses (`Connected`/`Dialing`/`Disconnecting`) describe the rotated-away key's connection, so the carry normalizes them to `Disconnected` (or `Banned` when a ban was pending) with matching counter bookkeeping; otherwise the new identity would be wedged behind a phantom connection (`can_dial` false, dial attempts short-circuiting with `AlreadyConnected`) that no libp2p event for the new id ever corrects.  When an anonymous record for the new peer id also exists (the peer reconnected before publishing its kad record), that record is promoted onto the confirmed identity as before, and the displaced record releases its counter bookkeeping through the new `release_displaced_record` helper.

`add_trusted_peer` gets the same eviction treatment, including one case `upsert_peer` already handled by construction: a network key re-presented under a different bls key.  Resolving the current identity through the index and evicting it prevents an orphaned `Confirmed` record whose later pruning would delete the legitimate index entry and silently make the trusted peer unresolvable.  Its old `remove_banned_peer(trusted_peer.known_ip_addresses())` call is dropped: ever since `Peer::new_trusted` lost its addr parameter that iterator is always empty, so the call could never unban an IP and only decremented the banned total spuriously (`BannedPeers::remove_banned_peer` decrements `total` unconditionally).  Displacement-release now does that job with the displaced record's actual addresses.

`ensure_peer_exists` self-heals as a backstop for the third bug: a confirmed identity with no backing record means a stale index entry, so it repairs the index and stores the default record under `Unidentified` instead of creating a keyless record under a `Confirmed` key.

Twelve new tests cover the acceptance criteria and the edge cases above: pure rotation (iff-invariant, status counters before and after eviction, no orphaned index entry, no zombie resurrection from the rotated-away peer id), rotation of a banned peer, rotation while connected (transport-status normalization), rotation displacing a disconnected record behind an anonymous reconnect, the same merge displacing a banned record, trusted-peer rotation with and without counters to release, anonymous-record displacement by `add_trusted_peer`, a network key re-bound to a different bls key through both `upsert_peer` and `add_trusted_peer`, banned-total preservation in `add_trusted_peer`, and the `ensure_peer_exists` index repair.

Two known limitations are left as possible follow-ups (both pre-existing semantics, now documented in code comments): a displaced banned record releases its `banned_peers` bookkeeping without emitting an `Unbanned` event (these methods have no `PeerAction` channel), and in the anonymous-reconnect merge the surviving record keeps its fresh score, so a banned peer that reconnects anonymously before its kad record arrives sheds its AllPeers-level ban there, exactly as it did before this change.

### DIFF
--- a/crates/network-libp2p/src/peers/all_peers.rs
+++ b/crates/network-libp2p/src/peers/all_peers.rs
@@ -134,6 +134,57 @@ impl AllPeers {
         removed
     }
 
+    /// Release the status-counter bookkeeping held by a record dropped outside the normal
+    /// status-transition path (displaced by [`Self::add_trusted_peer`] or [`Self::upsert_peer`]).
+    ///
+    /// `disconnected_peers` counts records in `Disconnected` status and `banned_peers` tracks
+    /// records in `Banned` status; a record removed without passing through
+    /// [`Self::update_connection_status`] must release them here so pruning math stays accurate.
+    fn release_displaced_record(&mut self, displaced: &Peer) {
+        match displaced.connection_status() {
+            ConnectionStatus::Disconnected { .. } => {
+                self.disconnected_peers = self.disconnected_peers.saturating_sub(1);
+            }
+            ConnectionStatus::Banned { .. } => {
+                let _ = self.banned_peers.remove_banned_peer(displaced.known_ip_addresses());
+            }
+            ConnectionStatus::Connected { .. }
+            | ConnectionStatus::Dialing { .. }
+            | ConnectionStatus::Disconnecting { .. }
+            | ConnectionStatus::Unknown => {}
+        }
+    }
+
+    /// Normalize the transport-liveness status of a record carried across a network-key
+    /// rotation.
+    ///
+    /// Reputation (score, trust, ban) is a property of the domain identity and survives the
+    /// rotation, but `Connected`/`Dialing`/`Disconnecting` describe the rotated-away key's
+    /// connection, which no longer belongs to this record. Left in place they wedge the new
+    /// identity: `can_dial` reports false and dial attempts short-circuit with
+    /// `AlreadyConnected` even though no connection exists for the new peer id, and no
+    /// libp2p event for the new id ever corrects the record. Live transport states map onto
+    /// `Disconnected` (or `Banned` when a ban was pending) with matching counter bookkeeping.
+    fn normalize_carried_status(&mut self, peer: &mut Peer) {
+        match *peer.connection_status() {
+            ConnectionStatus::Connected { .. }
+            | ConnectionStatus::Dialing { .. }
+            | ConnectionStatus::Disconnecting { banned: false } => {
+                peer.set_connection_status(ConnectionStatus::Disconnected {
+                    instant: Instant::now(),
+                });
+                self.disconnected_peers = self.disconnected_peers.saturating_add(1);
+            }
+            ConnectionStatus::Disconnecting { banned: true } => {
+                peer.set_connection_status(ConnectionStatus::Banned { instant: Instant::now() });
+                self.banned_peers.add_banned_peer(peer);
+            }
+            ConnectionStatus::Disconnected { .. }
+            | ConnectionStatus::Banned { .. }
+            | ConnectionStatus::Unknown => {}
+        }
+    }
+
     /// Create a peer that is "trusted".
     ///
     /// This overwrites peer records and unbans ips.
@@ -143,12 +194,20 @@ impl AllPeers {
         network_key: NetworkPublicKey,
     ) {
         let peer_id: PeerId = network_key.clone().into();
-        let trusted_peer = Peer::new_trusted(bls_public_key, network_key);
-        let _ = self.banned_peers.remove_banned_peer(trusted_peer.known_ip_addresses());
-        // overwrite any prior record and key the peer by its confirmed (bls) identity
-        self.peers.remove(&PeerIdentity::Unidentified(peer_id));
+        let confirmed = PeerIdentity::Confirmed(bls_public_key);
+        let current = self.identity_for(&peer_id);
+        // overwrite any prior record: anonymous under this peer id, confirmed under a different
+        // bls key that previously presented this network key, or confirmed under a previous
+        // network key; each displaced record is removed through `evict` / released so no
+        // `bls_by_peer_id` entry or status counter goes stale
+        if let Some(displaced) = (current != confirmed).then(|| self.evict(&current)).flatten() {
+            self.release_displaced_record(&displaced);
+        }
+        if let Some(displaced) = self.evict(&confirmed) {
+            self.release_displaced_record(&displaced);
+        }
         self.bls_by_peer_id.insert(peer_id, bls_public_key);
-        self.peers.insert(PeerIdentity::Confirmed(bls_public_key), trusted_peer);
+        self.peers.insert(confirmed, Peer::new_trusted(bls_public_key, network_key));
     }
 
     /// Create a peer.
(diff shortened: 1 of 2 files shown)


sstanfield (dismissed):
LGTM!

crates/network-libp2p/src/tests/peers.rs
```diff
@@ -162,6 +162,348 @@ fn test_upsert_peer_seeds_multiaddrs_for_new_peer() {
...
+    // an event for the stale peer id repairs the index instead of storing a keyless default
+    // record under the confirmed key
+    all_peers.update_connection_status(&peer_id_1, NewConnectionStatus::Dialing);
+    assert!(all_peers.bls_by_peer_id.is_empty());
+    assert!(all_peers.peers.contains_key(&PeerIdentity::Unidentified(peer_id_1)));
+    assert!(!all_peers.peers.contains_key(&PeerIdentity::Confirmed(bls)));
+}
+
+#[test]
+fn test_upsert_peer_rotation_normalizes_carried_connected_status() {
```
grantkee, line 354: nit: add `test_upsert_peer_rotation_normalizes_carried_pending_ban_status` for testing the `Disconnecting { banned: true } → Banned + add_banned_peer` arm of
  `normalize_carried_status`

crates/network-libp2p/src/peers/all_peers.rs
```diff
@@ -159,18 +218,35 @@ impl AllPeers {
...
+        // connection, so it is normalized onto the new identity; a peer never seen before
+        // starts fresh
+        let carried = migrated.is_none();
+        let mut peer = migrated
+            .or(rotated)
+            .unwrap_or_else(|| Peer::new(bls_public_key, network_key.clone(), Vec::new()));
+        if carried {
+            self.normalize_carried_status(&mut peer);
+        }
+        peer.update_net(bls_public_key, network_key, addrs);
```
grantkee: Should the oder of operations be switched here?

When `normalize_carried_status` matches on `Disconnecting { banned: true }`, it calls `add_banned_peer` but the peer's IPs are updated until after the call is made on L247

grantkee (dismissed):
LGTM! I flagged a potential bug and a gap in test coverage, but there is so much work to be done in the networking layer that I'm okay if prefer to address these in a follow up PR (assuming you agree with the comments)

crates/network-libp2p/src/peers/all_peers.rs
MavenRain: Good catch on this . . . addressed

crates/network-libp2p/src/tests/peers.rs
MavenRain: I went ahead and added it

sstanfield (approved):

grantkee (approved):
Great work @MavenRain !

PR #741, opened 2026-06-12
Index so we can access all the bytes (batches plus consensus header) for consensus output in a pack file.

sstanfield:
This is a foundational refactor to allow coherent access to relevant bytes from pack files of improving sync.

It also greatly reduces the IO ops need to load a ConsensusOutput.  The core refactor is replacing the position index with a more complete version that in addition to the consensus header it also stores the bytes that make up the complete consensus output.


grantkee (approved):
LGTM!

PR #742, opened 2026-06-12
network-libp2p: bound codec decode buffers to delivered bytes

### LINKED ISSUE #728: network-libp2p: req-res codec commits attacker-declared buffer sizes before the body arrives
## Summary

`TNCodec::decode_message` sizes and zero-fills its decode buffers from two 4-byte length prefixes that arrive on the wire before any message body, validating each prefix only against a maximum.  An unauthenticated peer can therefore send an 8-byte header that declares a ~1 MiB uncompressed length and a ~1.17 MiB compressed length, then withhold the body.  The node commits ~2.17 MiB of zero-filled resident memory per inbound substream and parks at the body read until the 10s request timeout.  The timed-out request is not penalized, so the attack is cost-free and indefinitely repeatable, and the same codec path serves both the primary and worker swarms.  With the default `request_response::Config` (100 concurrent streams per connection) this is roughly 217 MiB of attacker-controlled resident memory per accepted connection from a few hundred bytes of input.

The codec is the only request-response message path for the consensus networks, and there is no per-RPC memory budget, no connection-limits behaviour, and no per-IP admission cap anywhere in the crate, so nothing bounds the aggregate.  Severity is bounded by config (a finite per-connection ceiling rather than unbounded growth), but the per-stream amplification is large, the cost to the attacker is negligible, and the pattern can be sustained indefinitely, so it is an availability concern worth closing before mainnet.

## Wire format and the validated maxima

`decode_message` (`crates/network-libp2p/src/codec.rs:21-72`) reads a fixed header before the body:

```
[4-byte uncompressed_len][4-byte compressed_len][compressed_body]
```

The two size checks are the only constraints on the declared lengths:

- `codec.rs:41` rejects `uncompressed_len > max_message_size` (`max_rpc_message_size`, 1 MiB by default, `crates/config/src/network.rs:173`).
- `codec.rs:52` rejects `compressed_len > snap::raw::max_compress_len(uncompressed_len)` (~1.17 MiB for a 1 MiB `uncompressed_len`).

Neither check relates the declared sizes to the number of bytes that have actually arrived on the substream.

## The defect

**Location:** `crates/network-libp2p/src/codec.rs:59-63`.

**Mechanism:** Both buffers are resized (and therefore zero-filled, committing resident pages) from the attacker-declared prefixes before the compressed body is read:

```rust
// resize buffers to reported sizes
decode_buffer.resize(uncompressed_len, 0);     // ~1 MiB committed       (codec.rs:59)
compressed_buffer.resize(compressed_len, 0);   // ~1.17 MiB committed    (codec.rs:60)

// read compressed data
io.read_exact(compressed_buffer).await?;       // body read happens AFTER (codec.rs:63)
```

`Vec::resize(n, 0)` writes zeros across every page, so this is committed RSS, not a lazy reservation.  A peer that sends the 8-byte header and then sends no body leaves the future parked at the `read_exact` await (`codec.rs:63`) holding ~2.17 MiB.  This contrasts with libp2p's own reference codec (`libp2p-request-response-0.29.0/src/cbor.rs`), which uses `io.take(limit).read_to_end(...)` and grows only as bytes physically arrive.  The eager resize-from-prefix is a TN-specific deviation.

**Impact:** ~2.17 MiB of resident memory committed per stalled inbound substream from an 8-byte header, an amplification of roughly 280,000x, held until the request timeout.

## Why it is cost-free and repeatable

The codec read is wired into `request_response::Behaviour` with `request_response::Config::default()` (`crates/network-libp2p/src/consensus.rs:263`), which yields `request_timeout = 10s` and `max_concurrent_streams = 100` per connection (`libp2p-request-response-0.29.0/src/lib.rs:318-319`).  So:

- Each connection can park up to 100 stalled reads, ~217 MiB, for up to 10s.
- At the timeout the behaviour emits `InboundFailure::Timeout`, which the network loop handles with no penalty (`crates/network-libp2p/src/consensus.rs:1029-1031`: the `Timeout | ConnectionClosed` arm is an explicit no-op).  The peer's score is untouched, so it never bans itself and can re-open the streams immediately.

There is no admission control to bound the number of such connections.  `handle_established_inbound_connection` (`crates/network-libp2p/src/peers/behavior.rs:79-93`) rejects only banned peers; the peer cap is applied only after a connection is fully established and then only as a graceful peer-exchange disconnect (`peers/behavior.rs:247-252`); and `TNBehavior` wires in no libp2p `ConnectionLimits` and no per-IP cap.  An attacker can rotate ephemeral `PeerId`s at will, and the identical codec path serves both the primary and worker swarms.

## Reproduction scenario

1. Dial the target over QUIC from an unauthenticated peer.  The connection is accepted (only banned peers are refused at establishment).
2. Negotiate the request-response protocol (`/telcoin-network/0.0.0`, `ProtocolSupport::Full`, `crates/config/src/network.rs:162,169-171`) and open up to 100 inbound substreams on the connection.
3. On each substream write exactly 8 bytes: `uncompressed_len = 0x0010_0000` (1 MiB, passes `codec.rs:41`) and `compressed_len` just under `snap::raw::max_compress_len(1 MiB)` (~1.17 MiB, passes `codec.rs:52`).  Send no further bytes.
4. The node commits ~2.17 MiB per stream (~217 MiB for the connection) and stalls at `codec.rs:63` for 10s.  On timeout the streams are reaped with no penalty; re-open immediately and repeat.  Open additional connections, and target both swarms, to multiply the held memory.

## Reproduction (runtime-verified)

The eager allocation is directly observable at the unit level by feeding `decode_message` only the 8-byte header and inspecting the buffers afterward.  Adding the following to `crates/network-libp2p/src/tests/codec_tests.rs` and running `cargo test -p tn-network-libp2p --lib poc_decode_message` passes on current `main`:

```rust
use std::{
    pin::Pin,
    task::{Context, Poll},
};

/// Reader that yields a fixed header then stalls forever on the body read,
/// modeling a peer that sends the length prefixes and withholds the body.
struct HeaderThenStall {
    header: Vec<u8>,
    pos: usize,
}

impl futures::AsyncRead for HeaderThenStall {
    fn poll_read(
        self: Pin<&mut Self>,
        _cx: &mut Context<'_>,
        buf: &mut [u8],
    ) -> Poll<std::io::Result<usize>> {
        let this = self.get_mut();
        if this.pos < this.header.len() {
            let n = std::cmp::min(buf.len(), this.header.len() - this.pos);
            buf[..n].copy_from_slice(&this.header[this.pos..this.pos + n]);
            this.pos += n;
            Poll::Ready(Ok(n))
        } else {
            // body never arrives
            Poll::Pending
        }
    }
}

/// 8 bytes total: declares a 1 MiB uncompressed length (passes the `<= max_message_size`
/// check at codec.rs:41) and a ~1.2 MB compressed length (`<= snap::raw::max_compress_len(1 MiB)`
/// which is ~1_223_370, passes codec.rs:52).
fn attacker_header() -> Vec<u8> {
    let uncompressed_len: u32 = 1024 * 1024;
    let compressed_len: u32 = 1_200_000;
    let mut header = Vec::with_capacity(8);
    header.extend_from_slice(&uncompressed_len.to_le_bytes());
    header.extend_from_slice(&compressed_len.to_le_bytes());
    header
}

#[tokio::test]
async fn poc_decode_message_commits_buffers_from_prefix_before_body() {
    let header = attacker_header();
    assert_eq!(header.len(), 8, "attacker input is only 8 bytes");

    let mut decode_buffer = Vec::new();
    let mut compressed_buffer = Vec::new();
    let max_message_size = 1024 * 1024;

    // header-only reader: 8 bytes then EOF (no body)
    let mut reader = header.as_slice();
    let res = decode_message::<_, u8>(
        &mut reader,
        &mut decode_buffer,
        &mut compressed_buffer,
        max_message_size,
    )
    .await;

    // the read fails (no body) but the allocation already happened
    assert!(res.is_err(), "decode must fail with no body");

    // 8 bytes of input forced these committed (zero-filled, len-bearing) buffers:
    assert_eq!(decode_buffer.len(), 1024 * 1024, "decode_buffer sized to declared uncompressed_len");
    assert_eq!(compressed_buffer.len(), 1_200_000, "compressed_buffer sized to declared compressed_len");
    let committed = decode_buffer.len() + compressed_buffer.len();
    assert!(committed >= 2_200_000, "~2.15 MiB committed from 8 input bytes (got {committed})");
}

#[tokio::test]
async fn poc_decode_message_holds_memory_while_body_withheld() {
    let mut reader = HeaderThenStall { header: attacker_header(), pos: 0 };
    let mut decode_buffer = Vec::new();
    let mut compressed_buffer = Vec::new();

    let fut =
        decode_message::<_, u8>(&mut reader, &mut decode_buffer, &mut compressed_buffer, 1024 * 1024);

    // the body never arrives; the read stays parked until our short test timeout fires
    // (on the wire the libp2p request_timeout, 10s by default, governs the hold)
    let res = tokio::time::timeout(std::time::Duration::from_millis(200), fut).await;
    assert!(res.is_err(), "future stays parked at the body read while the body is withheld");

    // while parked, the ~2.15 MiB stayed committed
    assert_eq!(decode_buffer.len(), 1024 * 1024);
    assert_eq!(compressed_buffer.len(), 1_200_000);
}
```

Result:

```
running 2 tests
test codec::codec_tests::poc_decode_message_commits_buffers_from_prefix_before_body ... ok
test codec::codec_tests::poc_decode_message_holds_memory_while_body_withheld ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 102 filtered out; finished in 0.21s
```

The first test confirms that 8 bytes of input leave `decode_buffer.len() == 1_048_576` and `compressed_buffer.len() == 1_200_000`, that is ~2.15 MiB committed before any body byte is read.  The second confirms the buffers stay committed while the read is parked awaiting the withheld body.  The per-connection figure (100 concurrent streams, ~217 MiB) follows from `request_response::Config::default()` (`libp2p-request-response-0.29.0/src/lib.rs:318-319`) and is derived from source rather than executed; the unit PoC above establishes the per-stream amplification primitive that the connection-level numbers scale.

## Root cause

`decode_message` trusts the unauthenticated length prefixes to pre-size zero-filled allocations before the corresponding bytes have arrived.  The bounded request timeout limits the hold time but not the per-stream allocation, and nothing in the crate budgets aggregate RPC memory or caps inbound connections, so a single peer converts a trivial amount of input into a large, sustained, resident-memory commitment.

## Suggested fix

Primary (`codec.rs`, `decode_message`):

- Do not pre-size from the prefixes.  Read the compressed body incrementally with a bounded reservation (for example reserve `min(compressed_len, CHUNK)` and grow as bytes arrive), so a withheld or short body never forces the full ~1.17 MiB up-front commit; the existing 10s request timeout reaps a stall.
- Grow `decode_buffer` only during decompression, bounding the `FrameDecoder` with `.take(u64::try_from(max_message_size).unwrap_or_default())` and erroring if the decompressed length disagrees with the declared `uncompressed_len`.  Keep the prefix validations at `codec.rs:41` and `codec.rs:52`.

Defense in depth:

- Set `request_response::Config::with_max_concurrent_streams(...)` at `consensus.rs:263` to a value matched to expected peer behaviour, to tighten the per-connection blast radius.
- Wire a libp2p `ConnectionLimits` behaviour, and/or a per-IP admission cap, into `TNBehavior` so inbound connections are bounded before establishment.
- Apply a Mild penalty for repeated `InboundFailure::Timeout` from the same peer within a short window so the abuse is no longer cost-free, and reconsider the no-op arm at `consensus.rs:1029-1031`.

## Suggested test design

No test in `crates/network-libp2p/src/tests/codec_tests.rs` exercises a header-only / withheld-body read.  Add a test that drives `decode_message` against a reader that yields only the 8-byte header (a valid 1 MiB / ~1.17 MiB prefix pair) and then returns `Pending` or blocks.  Assert that `decode_message` does not commit buffers proportional to the declared prefixes before the body is available.  A straightforward way to make the regression observable is to inject a reader that records the largest single allocation requested, or to restructure `decode_message` so the buffer length is a function of bytes actually delivered and assert it tracks the delivered bytes rather than the declared prefix.

## Acceptance criteria

- `decode_message` no longer commits memory proportional to the declared prefixes before the body bytes have arrived; a peer that sends only the header and stalls holds no more than a small bounded buffer until the request timeout.
- Repeated inbound timeouts from a peer are no longer entirely cost-free (penalty applied, or connections bounded), so the pattern cannot be sustained indefinitely at zero reputational cost.
- A regression test in `codec_tests.rs` covers the header-only / withheld-body case.

MavenRain:
network-libp2p: bound codec decode buffers to delivered bytes (#728)

## What this fixes

`TNCodec::decode_message` sized and zero-filled its decode buffers from the two attacker-declared 4-byte length prefixes before reading the body. An unauthenticated peer could send an 8-byte header declaring a ~1 MiB uncompressed length and a ~1.17 MiB compressed length, withhold the body, and commit ~2.17 MiB of resident memory per inbound substream until the 10s request timeout reaped the stream. With the default `request_response::Config` (100 concurrent streams per connection) that is roughly 217 MiB of attacker-controlled resident memory from a few hundred bytes of input, repeatable at no cost.

Fixes #728.

## The change

`crates/network-libp2p/src/codec.rs`, `decode_message`:

- Removed the two eager `resize(reported_len, 0)` calls that committed memory from the prefixes before any body byte arrived.
- The compressed body is now streamed in bounded 8 KiB chunks, appending only what actually arrives, so committed memory tracks delivered bytes rather than the declared `compressed_len`. A peer that declares a large body and withholds it holds only what it has sent until the timeout. A body shorter than the declared length is rejected with `UnexpectedEof`.
- Decompression now bounds its output to the validated `uncompressed_len` (`FrameDecoder::take(uncompressed_len)`), grows `decode_buffer` only as bytes are produced, and errors if the decompressed size disagrees with the declared prefix. This preserves the previous decompression-bomb ceiling (`max_message_size`) without the up-front commit.

The prefix validations (`uncompressed_len <= max_message_size`, `compressed_len <= snap::raw::max_compress_len(uncompressed_len)`) and the wire format are unchanged, so this is wire-compatible.

Note on implementation: the body read uses an explicit chunked loop rather than `io.take(compressed_len).read_to_end(...)` on purpose. `futures` `read_to_end` zero-fills the entire spare capacity of the target `Vec` on its first growth, which would re-introduce an eager commit for the worker's reused (pre-sized) buffers. The chunked `extend_from_slice` commits only the bytes that arrive regardless of the buffer's pre-existing capacity.

## Coverage

The same codec serves both the consensus request-response swarms (`consensus.rs`) and the worker batch-sync streams (`crates/consensus/worker/src/network/stream_codec.rs` calls the same `decode_message`/`encode_message`), so both paths are fixed by this change.

## Tests

Three regression tests added to `crates/network-libp2p/src/tests/codec_tests.rs`:

- `test_header_only_does_not_commit_declared_buffers`: a header-only read (body withheld, EOF) fails and leaves both buffers at length 0 rather than the ~2.15 MiB the prefixes declared.
- `test_partial_body_commits_only_delivered_bytes`: a peer that declares a ~1 MB body but delivers only 4096 bytes before closing commits exactly the delivered bytes, not the declared length.
- `test_withheld_body_holds_no_declared_memory_while_parked`: while the decode future is parked awaiting a withheld body, neither buffer holds memory proportional to the declared prefixes.

All existing codec tests and the 12 worker `stream_codec` tests (including the back-to-back multi-batch and multi-chunk roundtrips, which confirm the chunked read leaves the stream correctly positioned for the next message) pass unchanged.

Local gate (pinned toolchains):
- `cargo +nightly-2026-03-20 fmt --all -- --check`: clean
- `cargo +nightly-2026-03-20 clippy -p tn-network-libp2p --all-features --tests -- -D warnings`: clean
- `cargo test -p tn-network-libp2p --lib codec`: 7 passed
- `cargo test -p tn-worker --lib stream_codec`: 12 passed

## Scope and recommended follow-ups (not in this PR)

This PR removes the amplification at its root: a stalled or short body now costs the node only the bytes the peer actually sent, so the attack is bounded by attacker bandwidth rather than by an 8-byte header. The issue also lists outer-layer defenses that involve peer-scoring and connection policy decisions better owned by the maintainers, and which are deliberately left out of this focused fix:

- `request_response::Config::with_max_concurrent_streams(...)` at `consensus.rs` to tighten the per-connection blast radius.
- A libp2p `ConnectionLimits` behaviour and/or per-IP admission cap in `TNBehavior`.
- Revisiting the no-op `InboundFailure::Timeout` arm so sustained inbound timeouts from a peer are not entirely cost-free.

I'd be happy to follow up with any or all of these in a separate PR if you want them.

### DIFF
--- a/crates/network-libp2p/src/codec.rs
+++ b/crates/network-libp2p/src/codec.rs
@@ -88,8 +88,9 @@ where
     // Decompress, bounding the output to the validated `uncompressed_len` so a malformed snappy
     // stream cannot expand past `max_message_size`, and growing `decode_buffer` only as bytes are
     // produced rather than committing it up front.
+    let decode_limit = u64::try_from(uncompressed_len).map_err(std::io::Error::other)?;
     let reader = std::io::Cursor::new(&*compressed_buffer);
-    let mut decoder = FrameDecoder::new(reader).take(uncompressed_len as u64);
+    let mut decoder = FrameDecoder::new(reader).take(decode_limit);
     decoder.read_to_end(decode_buffer)?;
 
     // the decompressed output must match the reported uncompressed length
--- a/crates/network-libp2p/src/codec.rs
+++ b/crates/network-libp2p/src/codec.rs
@@ -13,11 +13,22 @@ use std::{
 };
 use tn_types::encode_into_buffer;
 
+/// Maximum number of bytes pulled from the wire in a single read while streaming in the compressed
+/// body.
+///
+/// Reading the body in bounded increments keeps committed memory proportional to the bytes that
+/// have actually arrived rather than to the attacker-declared length prefix.
+const BODY_READ_CHUNK_SIZE: usize = 8 * 1024;
+
 /// Decode a single length-prefixed, snappy-compressed BCS message from an async reader.
 ///
 /// Wire format: `[4-byte uncompressed_len][4-byte compressed_len][compressed_data]`
 ///
 /// The caller provides reusable buffers to avoid repeated allocation.
+///
+/// The length prefixes are only used to bound the read; the buffers grow with the bytes that
+/// actually arrive, so a peer that declares a large body and then withholds it cannot force a large
+/// up-front allocation.
 pub async fn decode_message<T, M>(
     io: &mut T,
     decode_buffer: &mut Vec<u8>,
@@ -55,17 +66,38 @@ where
         ));
     }
 
-    // resize buffers to reported sizes
-    decode_buffer.resize(uncompressed_len, 0);
-    compressed_buffer.resize(compressed_len, 0);
+    // Stream the compressed body in bounded chunks so committed memory tracks the bytes that
+    // actually arrive rather than the attacker-declared `compressed_len`. A peer that declares a
+    // large body but withholds it stalls here holding only what it has sent, until the request
+    // timeout reaps the stream, instead of forcing a large zero-filled allocation up front.
+    let mut remaining = compressed_len;
+    let mut chunk = [0u8; BODY_READ_CHUNK_SIZE];
+    while remaining > 0 {
+        let want = remaining.min(chunk.len());
+        let read = io.read(&mut chunk[..want]).await?;
+        if read == 0 {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::UnexpectedEof,
+                "compressed body shorter than reported compressed size",
+            ));
+        }
+        compressed_buffer.extend_from_slice(&chunk[..read]);
+        remaining -= read;
+    }
 
-    // read compressed data
-    io.read_exact(compressed_buffer).await?;
-
-    // decompress
+    // Decompress, bounding the output to the validated `uncompressed_len` so a malformed snappy
+    // stream cannot expand past `max_message_size`, and growing `decode_buffer` only as bytes are
+    // produced rather than committing it up front.
     let reader = std::io::Cursor::new(&*compressed_buffer);
-    let mut decoder = FrameDecoder::new(reader);
-    decoder.read_exact(decode_buffer)?;
+    let mut decoder = FrameDecoder::new(reader).take(uncompressed_len as u64);
+    decoder.read_to_end(decode_buffer)?;
+
+    // the decompressed output must match the reported uncompressed length
+    if decode_buffer.len() != uncompressed_len {
+        return Err(std::io::Error::other(
+            "decompressed size does not match reported uncompressed size",
+        ));
+    }
 
     // deserialize
     bcs::from_bytes(decode_buffer).map_err(std::io::Error::other)
--- a/crates/network-libp2p/src/tests/codec_tests.rs
+++ b/crates/network-libp2p/src/tests/codec_tests.rs
@@ -225,3 +225,127 @@ async fn test_malicious_prefix_deceives_peer_to_read_message_and_fails() {
     let res = honest_peer.read_response(&protocol, &mut encoded.as_ref()).await;
     assert!(res.is_err());
 }
+
+/// Build the 8-byte length header a malicious peer would send: a 1 MiB uncompressed length (passes
+/// the `<= max_message_size` check) paired with a ~1.2 MB compressed length (passes the
+/// `<= snap::raw::max_compress_len(1 MiB)` check, which is ~1_223_370).
+fn malicious_header(uncompressed_len: u32, compressed_len: u32) -> Vec<u8> {
+    let mut header = Vec::with_capacity(8);
+    header.extend_from_slice(&uncompressed_len.to_le_bytes());
+    header.extend_from_slice(&compressed_len.to_le_bytes());
+    header
+}
+
+/// Regression test for issue #728: a peer that sends only the 8-byte length header and withholds
+/// the body must not cause the decoder to commit memory proportional to the declared prefixes.
+#[tokio::test]
+async fn test_header_only_does_not_commit_declared_buffers() {
+    let max_chunk_size = 1024 * 1024; // 1mb
+    let header = malicious_header(1024 * 1024, 1_200_000);
+    assert_eq!(header.len(), 8, "attacker input is only the 8-byte header");
+
+    let mut decode_buffer = Vec::new();
+    let mut compressed_buffer = Vec::new();
+    // reader yields the 8 header bytes then EOF (no body)
+    let mut reader = header.as_slice();
+    let res = decode_message::<_, TestPrimaryRequest>(
+        &mut reader,
+        &mut decode_buffer,
+        &mut compressed_buffer,
+        max_chunk_size,
+    )
+    .await;
+
+    // the decode fails because the body never arrives
+    assert!(res.is_err(), "decode must fail when the body is withheld");
+
+    // crucially, neither buffer was sized to the attacker-declared prefixes: with no body
+    // delivered, both hold zero bytes rather than the ~2.15 MiB the prefixes declared
+    assert_eq!(compressed_buffer.len(), 0, "no body arrived, so no compressed bytes are committed");
+    assert_eq!(decode_buffer.len(), 0, "decompression never ran, so no bytes are committed");
+}
+
+/// Regression test for issue #728: a peer that declares a large compressed body but delivers only
+/// part of it before closing commits memory proportional to what it actually sent, not the declared
+/// prefix.
+#[tokio::test]
+async fn test_partial_body_commits_only_delivered_bytes() {
+    let max_chunk_size = 1024 * 1024; // 1mb
+    let compressed_len: u32 = 1_000_000;
+    let delivered = 4096; // only a small slice of the declared body is sent
+
+    let mut wire = malicious_header(1024 * 1024, compressed_len);
+    wire.resize(wire.len() + delivered, 0);
+
+    let mut decode_buffer = Vec::new();
+    let mut compressed_buffer = Vec::new();
+    let mut reader = wire.as_slice();
+    let res = decode_message::<_, TestPrimaryRequest>(
+        &mut reader,
+        &mut decode_buffer,
+        &mut compressed_buffer,
+        max_chunk_size,
+    )
+    .await;
+
+    // fails because fewer bytes arrived than the declared compressed length
+    assert!(res.is_err(), "decode must fail on a short body");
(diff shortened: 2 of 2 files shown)


sstanfield (approved):
Good catch, looks good to me.

grantkee (approved):
LGTM!

PR #748, opened 2026-06-12
Fix two critical/high issues found with a scan.

sstanfield:
We need to sync testnet past epoch 74 as a sanity check.  Should work based on how packs provide ConsesusOutput.

Issues this addresses:

### 1. Epoch-closing system call silently skipped when final output contains a duplicate batch digest
- **Severity (initial)**: Critical
- **Category**: Consensus Safety
- **Location**: `crates/types/src/primary/output.rs:240-244` (interacting with `crates/consensus/executor/src/subscriber.rs:450-456`)
- **Claim**: `close_epoch_for_last_batch(index)` returns `Some(true)` only when `(index + 1) >= batch_digests.len()`. The subscriber keeps duplicate occurrences in `batch_digests` (pushed at subscriber.rs:450-456) but drops them from `batches` (subscriber.rs:477-493), so the maximum flattened `batch_index` reached by the engine loop is `batch_digests.len() - dup_count - 1` and the condition is never satisfied. If the epoch-closing `ConsensusOutput` contains a duplicated batch digest, no `TNPayload` receives `close_epoch: Some(randomness)`, so `apply_consensus_block_rewards` and `apply_closing_epoch_contract_call` are never invoked. All validators skip identically; the on-chain epoch is never concluded. A single Byzantine validator can trigger this by including an already-referenced batch digest in its header during the epoch-boundary commit window.
- **Key Question**: With `close_epoch == true` and `batch_digests` of length N containing one duplicate while flattened `batches` has N-1 entries, does any iteration of the payload loop satisfy `(index + 1) >= N`, and is there any other code path that applies the closing system calls for a non-empty final output?
- **Relevant Files**: crates/types/src/primary/output.rs, crates/consensus/executor/src/subscriber.rs, crates/engine/src/payload_builder.rs, crates/tn-reth/src/payload.rs, crates/tn-reth/src/evm/block.rs, crates/node/src/manager/node/epoch.rs
- **Source**: tn-review

### 2. Duplicate batch digests desynchronize `batch_digests` from `batches`, misaligning digest→block mapping
- **Category**: Consensus Safety
- **Location**: `crates/consensus/executor/src/subscriber.rs:477-493` (root cause), manifesting at `crates/engine/src/payload_builder.rs:117-120`
- **Claim**: In `Subscriber::fetch_batches`, `batch_digests` is built by pushing every payload digest of every header including duplicates, but duplicate occurrences are dropped from `cert_batches`/`batches`. The resulting `ConsensusOutput` has `batch_digests.len() > flatten_batches().len()`. The engine's `execute_consensus_output` loop indexes `output.get_batch_digest(batch_index)` by the flattened-batches enumeration index, so every batch positioned after a duplicate's second occurrence executes with the wrong `batch_digest` (used as the block's ommers-hash field) and wrong `mix_hash`/prev_randao. The guard at payload_builder.rs:51-55 is only a `debug_assert_eq!`. Behavior is identical on all honest validators (no divergence), but canonical blocks permanently record incorrect batch-digest bindings and duplicated randomness.
- **Key Question**: In a sub-dag where one batch digest appears in two headers, does `output.batch_digests().len()` exceed `output.flatten_batches().len()`, and does the block built for the post-duplicate batch receive the duplicate's digest as its `batch_digest`/`mix_hash` instead of its own?
- **Relevant Files**: crates/consensus/executor/src/subscriber.rs, crates/engine/src/payload_builder.rs, crates/types/src/primary/output.rs, crates/tn-reth/src/payload.rs
- **Source**: tn-review

### DIFF
--- a/crates/consensus/executor/Cargo.toml
+++ b/crates/consensus/executor/Cargo.toml
@@ -32,3 +32,7 @@ tempfile = { workspace = true }
 
 [lints]
 workspace = true
+
+[features]
+default = []
+adiri = []
--- a/crates/consensus/executor/src/subscriber.rs
+++ b/crates/consensus/executor/src/subscriber.rs
@@ -491,12 +491,14 @@ impl<DB: Database> Subscriber<DB> {
                         .find(|b| b.digest() == *digest)
                     {
                         warn!(target: "subscriber", ?digest, ?batch_digests, "failed to remove fetched batch - duplicate");
-                        if sub_dag.leader_epoch() != 74
+                        if number != 832748
                             || self.config.config().genesis().config.chain_id != 2017
                         {
-                            // Epoch 74 of adiri testnet had a bug with duplicate batches.
-                            // We have to recreate it in order to sync testnet so we skip this push
-                            // on adiri in epoch 74.
+                            // ADIRI BUG
+                            // Epoch 74 consensus number 832748 of adiri testnet had a bug with
+                            // duplicate batches. We have to recreate it
+                            // in order to sync testnet so we skip this push
+                            // on adiri for 832748.
                             cert_batches.push(batch.clone());
                         }
                     } else {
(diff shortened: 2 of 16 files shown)


grantkee (approved):
LGTM! Great work

Issue #739, opened 2026-06-11
refactor(network): consolidate the bulk request-response paths onto hardened bidirectional streams, keep single-shot RPCs on request-response


MavenRain:
## Summary

The consensus networks run two parallel point-to-point mechanisms: `request_response::Behaviour` for small single-shot RPCs and a hand-rolled `/tn-stream/0.0.1` behaviour for bulk transfers.  The bulk transfers reach their stream through the RPC behaviour: the requester sends an RPC asking permission, waits for an `ack` response, then opens a raw stream and writes a 32-byte digest so the responder can correlate the stream back to the original RPC.  That two-phase dance, the digest-correlation surface it creates, and the deliberately minimal stream behaviour underneath it are the cost we pay for keeping the two mechanisms separate.

This issue proposes consolidating the genuinely stream-shaped exchanges (worker batch sync, primary epoch packs, and primary `MissingCertificates`) onto a single hardened bidirectional-streaming layer that carries the request as its first frame, and keeping the single-shot consensus RPCs (`Vote`, `ReportBatch`, `ConsensusHeader`, `EpochRecord`) on `request_response`.  It is explicitly a hybrid: the goal is to delete the awkward seams, not to replace `request_response` wholesale.

## Why this is a hybrid and not a full replacement

It is tempting to frame this as "replace request-response with streaming," but there is something worth stating up front so it does not get re-litigated mid-review.

The consensus transport is QUIC-only (`crates/network-libp2p/src/consensus.rs:303-322`).  `libp2p-request-response` already opens a fresh bidirectional QUIC substream for every single request (`libp2p-request-response-0.29.0/src/handler.rs:450-451`, one `OutboundSubstreamRequest` per request).  Every RPC in this crate is therefore already bidirectional streaming at the wire level.  Replacing `request_response` for the single-shot RPCs buys no transport improvement; it only moves the RPC machinery that libp2p maintains for us (request-id correlation, the per-request 10s timeout enforced by `futures_bounded`, the 100-concurrent-stream cap, dial-on-demand via `send_request_with_addresses`, the cancel lifecycle, and the `OutboundFailure`/`InboundFailure` taxonomy that the entire peer-scoring system is written against, `consensus.rs:943-1037`) into code we own and maintain forever, under voting liveness, for little wire-level gain.

So the line we draw is by message shape, not by mechanism:

[...]

PR #749, opened 2026-06-13
refactor(network): own ResponseChannel and outbound-failure types (739)

MavenRain:
Step 2 of the #739 plan: **type decoupling, zero behaviour change.**

Today the crate leaks two libp2p request-response types through its public API:
`request_response::ResponseChannel` (re-exported from `lib.rs`, used in
`NetworkEvent::Request`/`Error`, `NetworkCommand::SendResponse`, and
`NetworkHandle::send_response`) and `OutboundFailure` (wrapped by
`NetworkError::Outbound` via `#[from]`). A libp2p upgrade that touches either
ripples into all four consumer crates.

This PR owns both:

- **`ResponseChannel<Res>`** is now a crate type (`types.rs`) wrapping the
  libp2p sink plus the `PeerId` of the peer that opened the exchange. It keeps
  the same name and is re-exported from `lib.rs`, so every consumer site
  compiles unchanged. The swarm task wraps the libp2p channel at the two event
  producers and unwraps it via `pub(crate) into_inner()` when answering. The
  new `peer_id()` accessor is the identity-first surface the later
  verdict-handle work builds on.
- **`RpcFailure`** (`error.rs`) mirrors `OutboundFailure`'s five variants and
  their Display strings verbatim, with `From<OutboundFailure>`.
  `NetworkError::Outbound` now wraps `RpcFailure`. The peer-scoring match in
  `consensus.rs` still classifies the raw libp2p failure from the swarm event;
  only the single public conversion that forwards the error to the caller was
  retargeted.

### Why no consumer changes

`ResponseChannel` is opaque to consumers: they receive it from a
`NetworkEvent` and forward it to `send_response`, never constructing it or
calling methods on it. Keeping the name makes the substitution invisible. No
consumer references `OutboundFailure`/`InboundFailure` or matches on
`NetworkError::Outbound`'s payload, so the failure change is fully internal.

### Scope / non-goals

Behaviour is unchanged: same wire protocol, same penalties, same Display text.
The per-exchange verdict handle, dial-on-demand, and the stream-path work are
later steps in #739.

### Tests

- New `rpc_failure_mirrors_outbound_failure` locks the variant mapping and
  Display parity across all five `OutboundFailure` variants (proves the
  conversion is loss-free).
- Existing `assert_matches!(NetworkError::Outbound(_))` network tests continue
  to pass (the wildcard matches regardless of the inner type).


sstanfield (dismissed):
LGTM!

crates/network-libp2p/src/types.rs
```diff
@@ -95,6 +95,42 @@ impl NetworkType {
     }
 }
 
+/// A channel for sending the response to an inbound RPC, bound to the peer that
+/// opened the exchange.
+///
+/// This is the crate-owned equivalent of libp2p's
+/// [`ResponseChannel`](libp2p::request_response::ResponseChannel): it wraps the
+/// same response sink but also carries the [`PeerId`] captured when the inbound
+/// request was accepted, so the answerer's identity travels with the channel
```
grantkee: nit: the "requestor's identity" is the one that's stored in the channel?

crates/network-libp2p/src/error.rs
grantkee: nit: can drop the `Libp2p` reference in the doc comment

grantkee (dismissed):
Great PR. Thank you for keeping these changes narrow. Two nits regarding docs but not blockers if you'd rather address in a follow up PR

sstanfield (approved):

PR #750, opened 2026-06-14
refactor(network): harden the /tn-stream behaviour (739, step 3)

MavenRain:
Step 3 of the #739 plan: **stream-module hardening** (prerequisite for any
later cutover). The `/tn-stream` behaviour was intentionally minimal; this makes
it safe to carry more traffic. Scope is the stream module plus its consensus
wiring; the request-response path is untouched.

### What changed

- **Dial-on-demand + connection tracking.** `StreamBehavior` now tracks
  established connections via `on_swarm_event`. `open_stream(peer, addrs, reply)`
  dispatches immediately when the peer is connected, otherwise emits
  `ToSwarm::Dial` (using the addresses `auth_to_peer` returns, which the
  `OpenStream` handler previously discarded as `_addrs`) and dispatches once the
  connection establishes. The peer manager still gates the dial in
  `handle_pending_outbound_connection`, so banned/undialable peers are rejected
  and surface as an explicit error.
- **Explicit `NotConnected` + timeouts.** An open to an unreachable peer (no
  addresses, or a failed dial) resolves to `StreamError::NotConnected` instead
  of hanging on a dropped oneshot. Negotiation is bounded by
  `SubstreamProtocol::with_timeout`; the overall open is bounded by a deadline
  swept by a periodic timer (which also guarantees `poll` runs to expire stale
  opens).
- **Bounded queues** in the behaviour (`events`, `pending`) and the handler
  (`pending_outbound`, events), shedding load with an explicit error rather than
  growing without limit.
- **Per-peer inbound rate limiting** with a sliding window; over-limit inbound
  streams are dropped and reported.
- **Classified failures, metrics-only.** A `StreamFailure` taxonomy mirrors the
  request-response one (`StreamFailure::penalty()`), routed to
  `process_stream_event`. Per #739's cross-cutting rule, new penalty
  classifications are **logged for telemetry but not enforced** yet; flipping to
  enforcement is a one-line change once telemetry confirms they do not fire on
  healthy peers.
- **Doc fix:** the module doc advertised `/tn-stream/1.0.0` while the code
  registers `/tn-stream/0.0.1`; the doc now matches.
- **Tests:** first direct unit tests of the stream behaviour, handler
  classification, and the penalty mapping (none existed before).

### Correctness note (disconnect race)

A `Connected` open whose peer drops before dispatch must not be sent to a
handler that no longer exists (libp2p silently drops the command and strands the
caller's reply). On `ConnectionClosed` the behaviour re-queues that peer's
dispatch-ready opens for dialing, and because the swarm processes
`ConnectionClosed` before `poll`, the open re-dials (recovering it or failing it
explicitly) instead of being lost. Multiple opens to the same disconnected peer
coalesce onto a single dial. A narrow residual remains (a connection dropping
*during* the handler's bounded negotiation after handoff), which is inherent to
the raw-stream handoff and yields a prompt error, not a hang.

### Verification

`cargo build`, `cargo +nightly fmt --check`, `cargo +nightly clippy --workspace
--all-features` (consumers compile, zero edits), `cargo test -p tn-network-libp2p
--lib` (**147 pass**, incl. 14 new stream tests), and `RUSTDOCFLAGS=-D warnings
cargo doc` all green. A 3-lens adversarial review drove the disconnect-race fix,
the dial coalescing, and the extra regression tests.

### Non-goals

The raw-stream keep-alive question (a handed-off stream's lifetime depends on the
swarm idle timeout) is unchanged and out of scope for this step. No
request-response, gossipsub, kad, or scoring-weight changes.


sstanfield (approved):
LGTM

grantkee (commented):
LGTM!

PR #754, opened 2026-06-15
fix(network): handle unresolved peer identity in gossip and response paths (#743)

### LINKED ISSUE #743: network-libp2p: valid consensus gossip is dropped silently (and valid RPC responses become spurious `PeerMissing` errors) when the sender's BLS identity is not yet resolved, even though the request path handles that exact race
## Summary

The consensus network forwards an inbound gossip message to the local application layer only when it can resolve the relaying peer's `BlsPublicKey`.  When `peer_to_bls(&propagation_source)` returns `None`, the message has already been reported to gossipsub as `Accept` (so it is propagated to the rest of the mesh) but it is never delivered to this node's consensus layer, and there is no log line for the drop.  The symmetric request-response *response* path has the same dependency and turns an otherwise-successful response into an `Err(NetworkError::PeerMissing)` for the same reason.

The reason this is a bug and not an intended invariant is internal inconsistency: the request-response *request* path already treats "we don't know this peer yet" as a real, handleable race and sends a `NetworkEvent::Error` back, while the gossip and response paths assume it cannot happen.  The inline comment on the gossip path ("We should not be able to recieve a message from an unknown peer so this should always work") states that assumption explicitly, but the assumption does not hold: a peer is `Connected` before its `NodeRecord` is processed, so there is a real window in which a live mesh neighbor is not yet BLS-resolvable.

### Gossip: silent drop, no log (`crates/network-libp2p/src/consensus.rs:785-823`)

The message is reported to gossipsub first (Accept propagates it to the mesh), then handed to the app only inside an `if let Some(bls)` with no `else`:

```rust
// report message validation results to propagate valid messages
if !self.swarm.behaviour_mut().gossipsub.report_message_validation_result(
    &message_id,
    &propagation_source,
    msg_acceptance.into(),
) { /* ... */ }

// process gossip in application layer
if valid {
    // We should not be able to recieve a message from an unknown peer so this
    // should always work.
    if let Some(bls) =
        self.swarm.behaviour().peer_manager.peer_to_bls(&propagation_source)
    {
        // forward gossip to handler
        if let Err(e) =
            self.event_stream.try_send(NetworkEvent::Gossip(message, bls))
        { /* logged + return */ }
    }
    // <-- no else: when peer_to_bls is None the accepted message is dropped with no log
} else {
    /* invalid: fatal penalty */
}
```

### Response: valid response becomes `PeerMissing` (`crates/network-libp2p/src/consensus.rs:906-924`)

```rust
request_response::Message::Response { request_id, response } => {
    // ...
    let _ = self.outbound_requests.remove(&(peer, request_id)).map(|ack| {
        if let Some(key) = self.swarm.behaviour().peer_manager.peer_to_bls(&peer) {
            let _ = ack.send(Ok(NetworkResponseMessage { peer: key, result: response }));
        } else {
            let _ = ack.send(Err(NetworkError::PeerMissing)); // a good response, reported as a failure
        }
    });
}
```

### Request: the same race handled correctly (`crates/network-libp2p/src/consensus.rs:865-904`)

```rust
// We should not be able to recieve a message from an unknown peer so this
// should always work. It is possible (mostly in
// testing) to have a race where we don't know the requester YET.
// If so send an error back ...
if let Some(bls) = self.swarm.behaviour().peer_manager.peer_to_bls(&peer) {
    // forward request ...
} else if let Err(e) = self.event_stream.try_send(NetworkEvent::Error(
    format!("requesting peer unknown: {peer:?}"),
    channel,
)) { /* ... */ }
```

## Why the unresolved-peer window is reachable

`peer_to_bls` is backed by the `Confirmed`-identity index (`bls_by_peer_id`).  `register_peer_connection` (`crates/network-libp2p/src/peers/manager.rs:431`) sets a peer to `Connected` but does not populate that index; the mapping is filled later, when the peer's `NodeRecord` is processed (`add_known_peer` / kad record handling).  A peer can therefore be an active gossipsub mesh neighbor, relaying a certificate, before this node has resolved its BLS key.  The gossip topics affected are the consensus-critical ones (`tn_certificates`, `tn_consensus_headers`, `tn_batches`, epoch votes).

## Impact

Reliability and observability.  On the gossip side, a certificate or consensus-output message that this node accepts and re-propagates to the network is never delivered to this node's own consensus layer, with no log entry, so the loss is invisible in production and the node falls back to slower recovery or sync paths without any signal as to why.  On the response side, a successfully received RPC response is converted into a `PeerMissing` error purely because of a transient resolution gap, producing spurious request failures.

This is not a high-frequency event on a warm node, so it should be filed as a correctness and observability gap rather than a constant outage.  The strongest argument that it is a real defect is the asymmetry: the request path in the same file already handles the identical race deliberately.

## Suggested fix

Give the gossip and response paths the same explicit `None` handling the request path has.

- Gossip: add an `else` to the `if let Some(bls)`.  At minimum emit a `warn!` so the drop is observable; better, surface it to the application (a `NetworkEvent` carrying the raw `propagation_source`) so the consensus layer can decide whether to buffer until the identity resolves or to penalize.
- Response: when `peer_to_bls(&peer)` is `None`, the response payload is still valid; consider delivering it with the resolved identity deferred, or at least distinguish "responder identity not yet resolved" from a genuine `PeerMissing` so callers do not retry a request that actually succeeded.

## Scope and novelty

Confirmed end to end on `main` (`0bd36414`).  Distinct from #713 (gossip multiple-counting a validator), #709 (late gossip grace window), and the closed #437 (gossip authorization bypass): none of those address local non-delivery of an accepted message when the relay's identity is unresolved, nor the missing log.  Not covered by the `network-libp2p` security review (PR #620, findings NL-001..NL-010).

MavenRain:
## Summary

Issue #743: when a relaying peer is `Connected` but its BLS identity has not yet
resolved (it connected before its `NodeRecord` populated the confirmed-identity
index `bls_by_peer_id`), `peer_to_bls` returns `None`. Two delivery paths in
`consensus.rs` mishandled that window:

- **Gossip:** a valid, accepted (and already re-propagated) message was dropped
  locally with no log when `peer_to_bls(&propagation_source)` returned `None`, so
  the loss was invisible in production.
- **Response:** a genuine RPC response was converted into
  `NetworkError::PeerMissing` (a spurious failure) when `peer_to_bls(&peer)`
  returned `None`.

The request path in the same file already handles this exact race deliberately
(it sends a `NetworkEvent::Error` back). This PR brings the gossip and response
paths in line.

## Changes

- **Gossip:** add the missing `else` branch and emit a `warn!` so an undelivered
  accepted message is observable instead of silently dropped. Corrected the
  inline comment that asserted the unknown-peer case "should always work"; the
  issue documents why the window is reachable.
- **Response:** introduce `NetworkError::PeerUnresolved` and a small pure helper
  `resolve_response`, so an unresolved responder identity is reported as a
  transient, retryable race distinct from `PeerMissing` ("not in our local
  store"). The success path is unchanged.
- `NetworkError::PeerUnresolved` is additive. No code matches `NetworkError`
  variants exhaustively, so existing callers are unaffected; callers that want to
  can now distinguish the transient race from a genuine miss.

## Scope

This addresses the reliability and observability gap (the "at minimum, emit a
`warn!`" option in the issue). It deliberately does not yet surface the
unresolved-gossip event to the application layer for buffer-or-penalize handling
(the issue's "better" option). That is a larger cross-crate design change that
overlaps the #739 plan to turn peer admission into an explicit decision, so it is
better as a focused follow-up than bundled here.

## Testing

- `cargo +nightly-2026-03-20 fmt --all -- --check`: clean
- `cargo +nightly-2026-03-20 clippy -p tn-network-libp2p --all-features --all-targets -- -D warnings`: clean
- `cargo +1.94 test -p tn-network-libp2p --all-features`: 135 passed, 0 failed

Adds two deterministic regression tests covering the resolved and unresolved
response paths.

## Dependencies

Independent of the other open network PRs. If #749 (owning outbound-failure
types) lands first, the only interaction is a trivial adjacency in `error.rs`
(both add a `NetworkError` variant). #740 (#722 rotation) is complementary and
does not close this window, so the fix is still needed regardless.

Closes #743.

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -832,8 +832,11 @@ where
 
                 // process gossip in application layer
                 if valid {
-                    // We should not be able to recieve a message from an unknown peer so this
-                    // should always work.
+                    // A peer is `Connected` before its `NodeRecord` resolves its BLS
+                    // identity, so a live mesh neighbor can relay a message before
+                    // `peer_to_bls` can resolve it. This is rare on a warm node but
+                    // reachable, so handle the unresolved case explicitly instead of
+                    // dropping an accepted message with no trace.
                     if let Some(bls) =
                         self.swarm.behaviour().peer_manager.peer_to_bls(&propagation_source)
                     {
@@ -846,6 +849,19 @@ where
                             // During epoch change the event_stream reciever can be closed.
                             return Ok(());
                         }
+                    } else {
+                        // The message was accepted into the mesh (and re-propagated),
+                        // but this node cannot yet resolve the relaying peer's BLS
+                        // identity, so it is not delivered to the local consensus
+                        // layer. Warn so the drop is observable rather than silent;
+                        // the node recovers the payload via sync once the identity
+                        // resolves.
+                        warn!(
+                            target: "network",
+                            ?propagation_source,
+                            ?message_id,
+                            "accepted gossip not delivered locally: relaying peer's BLS identity is not yet resolved"
+                        );
                     }
                 } else {
                     let GossipMessage { source, topic, .. } = message;
@@ -951,16 +967,13 @@ where
 
                         // try to forward response to original caller
                         let _ = self.outbound_requests.remove(&(peer, request_id)).map(|ack| {
-                            if let Some(key) =
-                                self.swarm.behaviour().peer_manager.peer_to_bls(&peer)
-                            {
-                                let _ = ack.send(Ok(NetworkResponseMessage {
-                                    peer: key,
-                                    result: response,
-                                }));
-                            } else {
-                                let _ = ack.send(Err(NetworkError::PeerMissing));
-                            }
+                            // The response payload is genuine (we still hold the
+                            // matching outbound request). If the responder's BLS
+                            // identity has not resolved yet, report a transient
+                            // `PeerUnresolved` rather than a misleading `PeerMissing`
+                            // so the caller does not retry a request that succeeded.
+                            let resolved = self.swarm.behaviour().peer_manager.peer_to_bls(&peer);
+                            let _ = ack.send(resolve_response(resolved, response));
                         });
                     }
                 }
@@ -1642,4 +1655,20 @@ where
             .field("swarm", &"<swarm>") // Skip detailed debug for swarm
             .finish()
     }
+}
+
+/// Pair a response payload with the responding peer's resolved BLS identity.
+///
+/// The payload is genuine whenever this node still holds the matching outbound
+/// request, so a peer whose identity has not resolved yet (it connected before
+/// its `NodeRecord` populated the confirmed-identity index) is reported as a
+/// transient [`NetworkError::PeerUnresolved`] rather than the misleading
+/// [`NetworkError::PeerMissing`].
+fn resolve_response<Res: TNMessage>(
+    resolved: Option<BlsPublicKey>,
+    response: Res,
+) -> NetworkResult<NetworkResponseMessage<Res>> {
+    resolved
+        .map(|peer| NetworkResponseMessage { peer, result: response })
+        .ok_or(NetworkError::PeerUnresolved)
 }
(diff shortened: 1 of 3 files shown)


sstanfield (approved):
LGTM!

grantkee (approved):
LGTM!

PR #756, opened 2026-06-15
Make sure we can not save consensus to the wrong epoch pack file.

sstanfield:
This was added to address this finding.  Note this should not happen and has not on testnet or else packs would be corrupt but this is a reasonable defense.

### 4. Epoch-mismatched ConsensusOutput can be appended into the wrong pack file
- **Category**: State corruption
- **Location**: `crates/storage/src/consensus.rs:494-498`, `crates/storage/src/consensus_pack.rs:882-896`
- **Claim**: `ConsensusChain::save_consensus_output` branch 1 writes to `current_pack` without checking `consensus.sub_dag().leader_epoch() == pack.epoch()`. In `Inner::save_consensus_output`, `consensus_idx = consensus_number.saturating_sub(self.epoch_meta.start_consensus_number)` maps any output with `number < start_consensus_number` to index 0; on a freshly-created empty epoch pack, the guards `0 < 0` and `0 != 0` both pass, so a prior-epoch header and its batches are appended at index 0 of the new epoch's pack, and the legitimate first output of the epoch is then silently dropped as "already saved". Symmetrically, the first output of epoch E+1 arriving while `current_pack` is still the sealed epoch-E pack hits `consensus_idx == len` and is appended into the sealed pack. The corrupted pack persists across restarts and `trunc_and_heal` cannot repair it (the record is CRC-valid and indexed). Reachability: after a restart with stale `LatestConsensus` slot files (fsynced only on epoch change), and `handle_consensus_header` (subscriber.rs:139-145) only rejects headers with epoch *greater* than the committee epoch, not lesser.
- **Key Question**: Can `save_consensus_output` be invoked with an output whose leader_epoch differs from `current_pack.epoch()` while `consensus.number() > latest_consensus.number()` — specifically after a restart with stale slot files and a freshly-opened empty epoch pack, or at the sealed/next-pack boundary before `new_epoch` runs?
- **Relevant Files**: crates/storage/src/consensus.rs, crates/storage/src/consensus_pack.rs, crates/consensus/executor/src/subscriber.rs, crates/node/src/manager/node/epoch.rs
- **Source**: tn-review

### DIFF
--- a/crates/consensus/primary/src/network/mod.rs
+++ b/crates/consensus/primary/src/network/mod.rs
@@ -485,6 +485,7 @@ impl PrimaryNetworkHandle {
             ConsensusChainError::StreamUnavailable
             | ConsensusChainError::NoCurrentEpoch
             | ConsensusChainError::EpochDbError(_)
+            | ConsensusChainError::InvalidPackEpoch(_, _)
             | ConsensusChainError::IO(_) => None,
         }
     }
--- a/crates/storage/src/consensus.rs
+++ b/crates/storage/src/consensus.rs
@@ -494,8 +494,19 @@ impl ConsensusChain {
         if number > self.latest_consensus.number() {
             let epoch = consensus.sub_dag().leader_epoch();
             if let Some(pack) = &self.current_pack() {
-                pack.save_consensus_output(consensus).await?;
-                self.latest_consensus.update(epoch, number).await;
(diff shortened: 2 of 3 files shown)


MavenRain (approved):
This is solid.  Good stuff.

grantkee (approved):
Agreed - good work @sstanfield !

PR #757, opened 2026-06-15
Reduce the default hash index size (buckets).

sstanfield:
The default size will goes from 126M to a few M.  Rough testing on a testnet sync and they seem to perform fine.

We can do other benchmarking later but this will reduce wasted disk space by a LOT.

Using less default buckets uncovered a bug in a test (not in prod code but still a bug).  Also performed a Claude security, performance and correctness scan on the hash index while here.

### DIFF
--- a/crates/storage/src/archive/digest_index/index.rs
+++ b/crates/storage/src/archive/digest_index/index.rs
@@ -264,14 +264,7 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
     /// Each bucket element is a (KSIZE bytes, u64)- (digest, record_pos).
     pub const BUCKET_ELEMENT_SIZE: usize = KSIZE + 8;
     /// Number of buckets to allocate in a fresh index.
-    /// Test builds use fewer buckets to avoid ~130MB of zero-initialized file I/O per HDX file
-    /// during epoch init (open_epoch_pack → ConsensusPack::open_append). The index auto-grows
-    /// via split_one_bucket() if capacity is exceeded.
-    #[cfg(all(feature = "test-utils", not(test)))]
     pub const INITIAL_BUCKETS: usize = 1_000;
-    /// Number of buckets to allocate in a fresh index.
-    #[cfg(any(not(feature = "test-utils"), test))]
-    pub const INITIAL_BUCKETS: usize = 100_000;
     /// Number of elements each bucket can hold before overflowing.
     pub const BUCKET_ELEMENTS: usize = 32;
     /// How large (in bytes) is each bucket.
--- a/crates/storage/src/archive/digest_index/index.rs
+++ b/crates/storage/src/archive/digest_index/index.rs
@@ -550,26 +550,24 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
         // Make sure we only keep the first (most recent) key if we find duplicates.
         let mut rec_hashes = BTreeSet::new();
         while let Some((rec_hash, rec_pos)) = self.next_bucket_element(&mut iter) {
-            if rec_pos > 0 {
-                let hash = B256::from_slice(rec_hash);
-                if !rec_hashes.contains(&hash) {
-                    let bucket = self.hash_to_bucket(rec_hash);
-                    if bucket != split_bucket && bucket != new_bucket {
-                        panic!(
-                            "got bucket {}, expected {} or {}, mod {}",
-                            bucket,
-                            split_bucket,
-                            self.buckets() - 1,
-                            self.modulus
-                        );
-                    }
-                    if bucket == split_bucket {
-                        self.save_to_bucket_buffer(rec_hash, rec_pos, &mut buffer, false)?;
-                    } else {
-                        self.save_to_bucket_buffer(rec_hash, rec_pos, &mut buffer2, false)?;
-                    }
-                    rec_hashes.insert(hash);
+            let hash = B256::from_slice(rec_hash);
+            if !rec_hashes.contains(&hash) {
+                let bucket = self.hash_to_bucket(rec_hash);
+                if bucket != split_bucket && bucket != new_bucket {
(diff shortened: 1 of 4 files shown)


crates/storage/src/archive/digest_index/index.rs
```diff
@@ -264,14 +264,7 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
...
     pub const BUCKET_ELEMENT_SIZE: usize = KSIZE + 8;
     /// Number of buckets to allocate in a fresh index.
-    /// Test builds use fewer buckets to avoid ~130MB of zero-initialized file I/O per HDX file
-    /// during epoch init (open_epoch_pack → ConsensusPack::open_append). The index auto-grows
-    /// via split_one_bucket() if capacity is exceeded.
-    #[cfg(all(feature = "test-utils", not(test)))]
     pub const INITIAL_BUCKETS: usize = 1_000;
-    /// Number of buckets to allocate in a fresh index.
-    #[cfg(any(not(feature = "test-utils"), test))]
-    pub const INITIAL_BUCKETS: usize = 100_000;
```
MavenRain, line 274: So, we'll go from rarely running `split_one_bucket` to always running it?  That'll be OK, yeah?

crates/storage/src/archive/digest_index/index.rs
```diff
@@ -557,26 +565,24 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
...
-                        self.save_to_bucket_buffer(rec_hash, rec_pos, &mut buffer, false)?;
-                    } else {
-                        self.save_to_bucket_buffer(rec_hash, rec_pos, &mut buffer2, false)?;
-                    }
-                    rec_hashes.insert(hash);
+            let hash = B256::from_slice(rec_hash);
+            if !rec_hashes.contains(&hash) {
+                let bucket = self.hash_to_bucket(rec_hash);
+                if bucket != split_bucket && bucket != new_bucket {
+                    panic!(
```
MavenRain, line 572: I know this isn't a part of this PR, but . . . either in this PR or a follow-up, could we return `Err(AppendError::CrcError)` here instead of panicking?

crates/storage/src/archive/digest_index/index.rs
```diff
@@ -351,6 +344,16 @@ impl<const KSIZE: usize, S: BuildHasher + Default> HdxIndex<KSIZE, S> {
             if header.uid() != data_header.uid() {
                 return Err(LoadHeaderError::InvalidIndexUID);
             }
+            // The on-disk bucket geometry must match this binary's compile-time layout.
+            // Reads size buffers from the header (bucket_size) while the bucket/overflow
+            // iterators index using the compile-time BUCKET_SIZE / BUCKET_ELEMENT_SIZE
+            // (derived from KSIZE and BUCKET_ELEMENTS).  A mismatch would silently corrupt
+            // reads, so reject it up front like the other header fields.
+            if header.bucket_size() != Self::BUCKET_SIZE as u16
```
MavenRain, line 352: Very nitty: I am of the (perhaps unpopular) opinion that the fact that panic-indicating as-casting is idiomatic for intended truncation is unfortunate.  I don't know if this team would consider it, but surfacing a helper trait, like the one I've illustrated below, would have the net effect of giving me a panic attack (no pun intended) in one place, instead of many.

```
trait Truncate { fn u16(self) -> u16; }
impl Truncate for usize { fn u16(self) -> u16 { self as u16 } }
```

MavenRain (approved):

MavenRain:
No blockers for merge

crates/storage/src/archive/digest_index/index.rs
```diff
@@ -799,14 +813,14 @@ mod tests {
             let mut hasher = DefaultHashFunction::new();
             hasher.update(&format!("idx-{i}").into_bytes());
             let hash = B256::from_slice(hasher.finalize().as_bytes());
-            idx.save(hash, i).expect("add to index");
+            idx.save(hash, i).expect(&format!("add to index {i}"));
```
MavenRain, line 816: Nit:  I think this eagerly allocates the string, and this may run a bunch of times.  Consider `unwrap_or_else`ing the expectation.

crates/storage/src/archive/digest_index/index.rs
sstanfield: Yep, avoiding that was the original reason for the large bucket count.  Functionally it is fine, I am confident splitting buckets works, just adds some more IO ops is all.

crates/storage/src/archive/digest_index/index.rs
sstanfield: I will double check this but I would only have used panic! if tripping this was a coding error not something data driven.  In other words I think this check is just a sanity check to make sure some finicky code is not buggy.  In this case I think panic is appropriate since the code is broken and assumptions are not holding, etc.  I.e. we are into undefined behavior territory.  At least that's the idea.

crates/storage/src/archive/digest_index/index.rs
sstanfield: I often play loose with these casts but will admit it is sloppy.  In this case I am OK with it because we are casting a constant and if it won't fit in u16 a quick panic is probably appropriate.  So I guess I am good with casts that correct at compile time or infallible (u32 -> u64 for instance) but am good with being more rigorous when the cast could fail.

Looking at your trait, its point is to cause a panic at one place?  I think I am missing something, we can discuss tomorrow though.

crates/storage/src/archive/digest_index/index.rs
sstanfield: I think you're right.  It is just a test but no reason to be inefficient.

grantkee (approved):
No blockers on my end either - let's discuss some of the suggestions @MavenRain brought up. I tend to agree with following idiomatic best-practices when possible

PR #758, opened 2026-06-16
fix(network): recover partial peers from GetClosestPeers query timeouts (#744)

### LINKED ISSUE #744: network-libp2p: `GetClosestPeers` query timeouts discard every peer the query already discovered, starving the discovery loop exactly when the node is short on peers
## Summary

The consensus network drives peer discovery by issuing kademlia `get_closest_peers` queries and feeding their results into the discovery candidate pool.  The kad event handler only processes the `Ok` arm of `GetClosestPeers`.  The `Err` arm, which kademlia returns on query timeout and which carries the peers discovered so far, falls into the catch-all `_ => {}` and is thrown away.  Discovery queries are scheduled precisely when the node is below its peer target, and that same low-connectivity condition is what makes queries slow enough to hit the 60 second query timeout, so the partial results are lost in exactly the situation where they are most needed.

## Where

`crates/network-libp2p/src/consensus.rs:1371-1379`:

```rust
kad::QueryResult::GetClosestPeers(Ok(result)) => {
    // process peers for potential discovery attempts
    debug!(target: "network-kad", ?result, "GetClosestPeers for discovery");
    self.swarm
        .behaviour_mut()
        .peer_manager
        .process_peers_for_discovery(result.peers);
}
_ => {}
```

There is no `kad::QueryResult::GetClosestPeers(Err(..))` arm.  On timeout, kademlia yields `GetClosestPeersError::Timeout`, whose payload includes the peers the query had already found; that whole set is discarded by the catch-all.

## Why it is reachable

- `PeerEvent::Discovery` is handled by issuing the query: `kademlia.get_closest_peers(PeerId::random())` (`crates/network-libp2p/src/consensus.rs:1226-1228`).
- `discovery_heartbeat` emits `PeerEvent::Discovery` whenever the discovery pool is below target (`crates/network-libp2p/src/peers/manager.rs:835-838`), which is the under-connected state.
- The kad query timeout is 60 seconds (`set_query_timeout(Duration::from_secs(60))`, `crates/network-libp2p/src/consensus.rs:277`).
- The sink that would have consumed the peers is `PeerManager::process_peers_for_discovery` (`crates/network-libp2p/src/peers/manager.rs:777`), which filters for eligibility and stores candidates for the next dial cycle.

## Impact

Reliability.  A node trying to reach its connection target loses the partial discovery output of every timed-out query.  Because the discovery loop only fires when the node is short on peers, and because a peer-starved node is the most likely to time out, the failure mode compounds: the node stays under-connected longer than necessary even though it had already located usable peers.  This is recoverable (the next heartbeat issues a fresh query), so it is a degraded-connectivity reliability issue rather than a liveness halt.

## Suggested fix

Add an explicit `GetClosestPeers(Err(..))` arm that still consumes the peers the query discovered before timing out, mirroring the `Ok` arm:

```rust
kad::QueryResult::GetClosestPeers(Err(err)) => {
    let peers = err.peers().to_vec(); // peers found before the timeout
    debug!(target: "network-kad", ?err, recovered = peers.len(), "GetClosestPeers timed out; using partial results");
    self.swarm
        .behaviour_mut()
        .peer_manager
        .process_peers_for_discovery(peers);
}
```

(Field accessors should be matched to the exact `GetClosestPeersError` shape in the pinned libp2p version.)

## Scope and novelty

Confirmed end to end on `main` (`0bd36414`).  No existing issue or PR tracks the discarded-partial-results behavior; the closest prior art is the closed #612 (worker startup race), which is unrelated.  Not covered by the `network-libp2p` security review (PR #620, NL-001..NL-010).

MavenRain:
# fix(network): recover partial peers from GetClosestPeers query timeouts (#744)

## Summary

Issue #744: the kademlia `QueryResult` handler in `consensus.rs` matched only
`GetClosestPeers(Ok(..))`. The `Err(..)` arm fell into the catch-all `_ => {}`
and was discarded. Kademlia returns `GetClosestPeersError::Timeout` on a query
timeout, and that payload carries the peers the query had already located, so
every timed-out discovery query threw away usable peers.

This compounds badly: `PeerEvent::Discovery` queries are issued precisely when
the node is below its peer target, and that same low-connectivity state is what
makes queries slow enough to hit the 60-second kad timeout.  A peer-starved node
therefore loses the partial discovery output exactly when it most needs peers.
It is recoverable (the next `discovery_heartbeat` issues a fresh query), so this
is a degraded-connectivity reliability issue, not a liveness halt.

## Changes

- Add an explicit `kad::QueryResult::GetClosestPeers(Err(err))` arm that feeds the
  timed-out query's discovered peers into `process_peers_for_discovery`, the same
  sink the `Ok` arm uses.  This mirrors the `StartProviding(Ok/Err)` two-arm
  pattern directly above it in the same match.
- Extract a pure `pub(crate)` helper `partial_peers_from_get_closest_timeout` that
  destructures the timeout error to its peers, so the recovery logic is unit
  testable in isolation.
- The eligibility filtering (`eligible_for_discovery`: valid unbanned IP and
  `can_dial`), deduplication, and discovery-pool semantics are unchanged; only the
  source of the peers differs, so a recovered peer is treated identically to one
  from a successful query (self/banned/connected peers are still excluded).

### Note on the suggested fix

The issue sketched an `err.peers()` accessor. libp2p-kad `0.48.0` has no such
method: `GetClosestPeersError` is a single, non-`#[non_exhaustive]` variant
`Timeout { key: Vec<u8>, peers: Vec<PeerInfo> }` exposing only `key()`/`into_key()`.
The helper therefore destructures the variant directly:

```rust
pub(crate) fn partial_peers_from_get_closest_timeout(
    err: kad::GetClosestPeersError,
) -> Vec<kad::PeerInfo> {
    let kad::GetClosestPeersError::Timeout { peers, .. } = err;
    peers
}
```

The irrefutable `let` is exhaustive today and fail-loud by design: if a future
libp2p adds a variant or marks the enum `#[non_exhaustive]`, this stops compiling
and forces an explicit decision rather than silently mishandling the new case.
Confirmed against the libp2p source that `Timeout.peers` is built from the same
`query.peers.into_peerinfos_iter().collect()` the success path uses, so the
recovered set is genuine discovered peers.

## Testing

- `cargo +nightly-2026-03-20 fmt --check`: clean
- `cargo +nightly-2026-03-20 clippy -p tn-network-libp2p --all-features --all-targets -- -D warnings`: clean
- `cargo +1.94 test -p tn-network-libp2p`: 134 passed, 0 failed

Adds one deterministic regression test (`test_get_closest_peers_timeout_recovers_partial_peers`)
that builds a `Timeout` error carrying an eligible peer, runs it through the
recovery helper, and asserts the peer lands in `discovery_peers` with its address,
mirroring the existing `test_process_peers_for_discovery_filters_duplicates`.

### Test coverage scope (disclosure)

The regression test exercises the extraction helper and the
`process_peers_for_discovery` sink directly. It does not drive `process_kad_event`,
because `kad::Event` cannot be constructed in a unit test: `kad::QueryId` wraps a
private `usize` with no public constructor in libp2p-kad `0.48.0`. The match-arm
dispatch itself is therefore not directly covered, and a swarm-level integration
test that induces a real 60-second timeout was judged too slow and flaky to be
worth it. The test guards the recovery logic; the arm wiring is a short explicit
arm that mirrors the adjacent `Ok`/`StartProviding` arms.

## Dependencies

Independent of the other open network PRs. The change touches the kad handler
region of `consensus.rs` (around line 1417) plus a new end-of-file free function,
and `tests/peer_manager.rs`. The other open PRs that edit `consensus.rs`
(#754, #750, #749, #733) modify disjoint regions of the file and none touch the
`GetClosestPeers`/`QueryResult` match, so at most a trivial rebase is needed if one
lands first. No `Cargo.toml`/`Cargo.lock` change.

Closes #744.


sstanfield (dismissed):
LGTM!

grantkee (dismissed):
LGTM!

grantkee (approved):

sstanfield (approved):

Issue #107, opened 2024-10-15
Deployment: `tn genesis create-committee`


grantkee:
The `genesis create-committee` command fails during `adiri-genesis::genesis.sh` is called.

The script uses the deployed docker image to create the committee. The current workaround is to copy the genesis directory into telcoin-network and run the binary/command locally. This works. Using the docker container on mac does not.

grantkee:
The cli generated a default config when running `create-committee` in an empty directory which was then causing issues with the deployment. This is fixed now in the script that runs during deployments.

PR #759, opened 2026-06-17
tn-reth: generic BLS verify precompile (reusable beyond proof-of-possession)

Huwonk:
﻿## Summary

Node-side half of the consensus-registry upgrade. Two layered changes:

1. **Compressed + generic BLS verification.** The native precompile is now a generic `blsVerify(signature, pubkey, message)` over compressed inputs (48-byte G1 signature, 96-byte G2 pubkey), reusable beyond proof-of-possession. The `ConsensusRegistry` proof-of-possession is one caller: it builds the message (`intentPrefix || compressedPubkey || address`) on-chain and verifies it here. (An earlier iteration of this branch was a PoP-specific compressed precompile; this generalizes it per review.)
2. **Validator-set migration support.** The tn-contracts submodule adds `migrateValidatorSets`, a one-time system-gated back-fill of the per-status sets for an in-place upgrade.

Stacked on #732. Pairs with Telcoin-Association/tn-contracts#126 (`feat/compressed-pop` @ `f398a3f`), which this branch's submodule points at.

## Changes

- `bls_precompile`: generic `blsVerify(signature, pubkey, message)` - length-gate (48/96) then `bls_verify_secure` over the raw message. Unit + property tests cover the generic semantics, the compressed-only length gate (S1), and the DELEGATECALL relay path.
- tn-types: `construct_proof_of_possession_message` returns the raw, on-chain-constructible message; `generate`/`verify` use the raw sign/verify path; add the generic `bls_verify_secure` primitive and a layout-guard test pinning the intent prefix to `0x000000`. `verify_proof_of_possession_bls` is retained for off-chain use (genesis validation, tests).
- `KeyConfig::generate_proof_of_possession_bls` signs the new message.
- Engine close-epoch tests build from a fresh-registry genesis helper (`test_genesis_with_consensus_registry`) so they exercise `getValidatorsInfo`; the static `TESTNET_GENESIS` fixture is untouched.
- Bump the tn-contracts submodule to `f398a3f`.

## Validation

All green in WSL: tn-types / tn-config / tn-reth lib, tn-reth `it` (45, including the BLS precompile property tests and the DELEGATECALL relay), tn-engine `it` (8), e2e genesis + stake + epoch + sync (6), `fmt`, and `clippy` (lib/bins). tn-contracts consensus + fuzz suite 122/122.

## Audit (self-review)

Differential audit of the Rust and Solidity diffs, 0 Critical / 0 High:

- **Medium (Solidity):** `migrateValidatorSets` is an `O(totalSupply)` loop; paginate before any large validator set (not triggerable at testnet's 5).
- **Low (both, hardened):** the on-chain message builder now asserts `pubkey.length == 96`; the precompile's flat gas is documented as calibrated for the fixed PoP message (add per-word pricing before non-PoP callers).
- The verify primitive is sound: full signature + pubkey subgroup checks, deterministic, no panics; the message binds the compressed pubkey (rogue-key defense). The on-chain/off-chain message byte-match is proven by the e2e suite (real precompile + real crypto), which the contract-level mock cannot show.

## Deployment analysis (observer sync)

This bears on how the testnet adopts the change. TN observers **re-execute** consensus packs from genesis (confirmed in `state-sync` / `consensus_pack`); there is no checkpoint or state-snapshot fast-sync. Therefore:

- A true in-place hardfork (keeping block 0) would require the node binary to carry **both** the pre-fork and post-fork registry/precompile behavior so new nodes can replay across the fork - a large, consensus-critical dual-ABI surface, where any divergence is a chain split.
- A **state-preserving relaunch** (a new genesis baking the live validators/balances + the migrate-computed sets) avoids all of that: new nodes sync the new chain with the new binary only. It preserves validator state (no re-staking); only the block history resets.

Recommendation for the testnet: the state-preserving relaunch. `migrateValidatorSets` is the set-reconstruction primitive it (and any future mainnet in-place upgrade) uses.

## Storage compatibility with the deployed registry (#107)

`forge inspect storageLayout` of #107 (`8d4cb030`) versus this branch is a clean append: slots 0-41 are byte-for-byte identical; only `validatorSets` (slot 42) and `eligibleValidatorCount` (slot 43) are appended. So the deployed validator state maps onto the new code unchanged, which is what makes the migration (and a relaunch snapshot) faithful.

## Status

Draft for team review. Stacked on #732. The testnet deployment approach (a state-preserving relaunch, per the analysis above) is the open decision for the team.

## Update - BlsG1 is now an `IBlsG1` interface (supersedes the DELEGATECALL / library notes above)

Follow-up refactor (node `a7ed03af`, contracts `a44d726`): on the tn-contracts side `BlsG1` became a thin `interface IBlsG1`, and `ConsensusRegistry` now reaches the precompile with a direct low-level `staticcall` instead of a linked-library `delegatecall`. Node-side consequences:

- The registry bytecode no longer carries a link placeholder, so both `link_solidity_library` calls and the helper are removed. The registry initcode/runtimecode deploy as-is, with a fail-fast guard against a stale `__$..$__` placeholder. The unused `BLSG1_JSON` constant is deleted.
- The precompile property test's relay is retargeted to `STATICCALL` to match the registry's new call path; the precompile itself is unchanged.
- The tn-contracts submodule now points at `a44d726` (not `f398a3f`).

Re-validated in WSL after the refactor: tn-reth and tn-engine (94 tests, including the now-`STATICCALL` relay and the 8 genesis / close-epoch tests that verify PoPs through the precompile on the temp chain), plus the e2e genesis tests (2). Nightly `fmt` and `clippy` (default and `--all-features`) are clean; tn-contracts consensus suite 122/122.

### DIFF
--- a/crates/tn-reth/src/lib.rs
+++ b/crates/tn-reth/src/lib.rs
@@ -1191,8 +1191,7 @@ impl RethEnv {
                 };
                 let bls_pubkey: tn_types::Bytes = v.bls_public_key.to_bytes().into();
                 let proof = ConsensusRegistry::ProofOfPossession {
-                    uncompressedPubkey: v.bls_public_key.serialize().into(),
-                    uncompressedSignature: v.proof_of_possession.serialize().into(),
+                    signature: v.proof_of_possession.to_bytes().into(),
                 };
 
                 (validator, (bls_pubkey, proof))
@@ -1938,8 +1937,7 @@ mod tests {
             calldata,
         );
         let proof = ConsensusRegistry::ProofOfPossession {
-            uncompressedPubkey: new_validator.bls_public_key.serialize().into(),
-            uncompressedSignature: new_validator.proof_of_possession.serialize().into(),
+            signature: new_validator.proof_of_possession.to_bytes().into(),
         };
         let calldata = ConsensusRegistry::stakeCall {
             blsPubkey: new_validator.bls_public_key.to_bytes().into(),
--- a/crates/tn-reth/src/lib.rs
+++ b/crates/tn-reth/src/lib.rs
@@ -30,7 +30,7 @@ use crate::{
     traits::{DefaultEthPayloadTypes, TNExecution},
 };
 use alloy::{
-    hex::{self, ToHexExt},
+    hex,
     primitives::{Bytes, ChainId},
     sol_types::{SolCall, SolConstructor},
 };
@@ -1167,13 +1167,12 @@ impl RethEnv {
             .with_bundle_update()
             .build();
 
-        // The BlsG1 proof-of-possession library is a native precompile registered by the EVM
-        // factory at `BLS_G1_PRECOMPILE_ADDRESS`, so there is nothing to deploy at genesis:
-        // the registry is linked directly against the precompile address. The registry
-        // constructor's PoP verification `delegatecall`s this address and is served
-        // natively by the precompile, which shares the exact `blst` verification the
-        // consensus layer uses to produce the proofs (so the on-chain check cannot drift
-        // from the off-chain encoding).
+        // The BLS proof-of-possession precompile is registered by the EVM factory at
+        // `BLS_G1_PRECOMPILE_ADDRESS`, so there is nothing to deploy at genesis. The registry
+        // constructor's PoP verification staticcalls this address and is served natively by the
+        // precompile, which shares the exact `blst` verification the consensus layer uses to
+        // produce the proofs (so the on-chain check cannot drift from the off-chain
+        // encoding).
         let blsg1_address = BLS_G1_PRECOMPILE_ADDRESS;
 
         // prepare registry deployment
@@ -1218,11 +1217,15 @@ impl RethEnv {
             Self::fetch_value_from_json_str(CONSENSUS_REGISTRY_JSON, Some("bytecode.object"))?;
         let registry_initcode_str =
             registry_initcode_binding.as_str().ok_or_eyre("invalid registry json")?;
-        // link the BlsG1 library address into the registry bytecode
-        let linked_registry_initcode =
-            Self::link_solidity_library(registry_initcode_str, &blsg1_address.encode_hex())?;
-
-        let mut create_registry = linked_registry_initcode;
+        // The registry calls the BLS precompile directly at `BLS_G1_ADDRESS` (no linked library),
+        // so its bytecode carries no link placeholder; deploy it as-is. Guard against a
+        // stale artifact that still contains an unresolved `__$..$__` placeholder.
+        if registry_initcode_str.contains("__$") {
+            eyre::bail!(
+                "ConsensusRegistry initcode has an unresolved library link placeholder; regenerate tn-contracts artifacts"
+            );
+        }
+        let mut create_registry = hex::decode(registry_initcode_str)?;
         create_registry.extend(constructor_args);
 
         // after adding bls proof of possession, registry precompile exceeds size limit so disable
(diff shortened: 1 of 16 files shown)


crates/types/src/crypto/bls_signature.rs
```diff
@@ -173,55 +160,57 @@ pub fn generate_proof_of_possession_bls_for_test(
...
-pub fn proof_of_possession_message_bytes(
+/// The proof-of-possession message a validator signs to prove control of its BLS key and bind it to
+/// its execution `address`. Layout: `intentPrefix(3) || compressedBlsPubkey(96) || address(20)` =
+/// 119 raw bytes, where `intentPrefix` is the serialized telcoin proof-of-possession [`Intent`]
+/// (`0x000000`). Using the compressed pubkey and the raw address makes this cheaply reconstructable
+/// on-chain, so the `ConsensusRegistry` builds the identical bytes and verifies them through the
+/// native BLS precompile ([`bls_verify_secure`]). Returned as raw bytes (not wrapped in an
+/// [`IntentMessage`]) since both the signer and the on-chain verifier operate on the bytes
+/// directly.
+pub fn construct_proof_of_possession_message(
```
MavenRain, line 206: What's the plan for the old PoP signatures?

crates/tn-reth/src/lib.rs
```diff
@@ -1219,11 +1217,15 @@ impl RethEnv {
...
             registry_initcode_binding.as_str().ok_or_eyre("invalid registry json")?;
-        // link the BlsG1 library address into the registry bytecode
-        let linked_registry_initcode =
-            Self::link_solidity_library(registry_initcode_str, &blsg1_address.encode_hex())?;
-
-        let mut create_registry = linked_registry_initcode;
+        // The registry calls the BLS precompile directly at `BLS_G1_ADDRESS` (no linked library),
+        // so its bytecode carries no link placeholder; deploy it as-is. Guard against a
+        // stale artifact that still contains an unresolved `__$..$__` placeholder.
+        if registry_initcode_str.contains("__$") {
```
MavenRain, line 1223: Consider adding this same guard in front of the runtimecode call at 1301.  Otherwise, a stale build will either error non-obviously or store incorrect code.

MavenRain (approved):
One question, one suggestion

sstanfield (approved):

grantkee (approved):
LGTM!

crates/tn-reth/src/lib.rs
grantkee: This is a good idea, but not a blocker IMO. We are basically code frozen in `tn-contracts` so this is likely not an issue much longer

crates/types/src/crypto/bls_signature.rs
Huwonk: I think this would only affect the fork for testnet, since any new network would only use the new PoP signature.  We should talk about it with @grantkee tomorrow

Huwonk: Or the claude version:  In a fork the old PoPs are already spent: each one was verified once at its original stake tx and now just lives on as recorded validator state, so existing validators don't re-submit anything and their signatures are never re-checked after the fork (the registry only verifies a PoP at stake/delegateStake/genesis, never again). The real work is making the node fork-scheduled: it keeps the old verification (uncompressed, old ABI) active for pre-fork blocks so a fresh node syncing from genesis re-executes the historical stake txs identically, then switches to the new compressed blsVerify at the fork block, with new stakes after the fork using the new 119-byte message. So there's no PoP migration to do; the cost is carrying both verification paths (old for history, new post-fork) and swapping the ConsensusRegistry code at the fork.

Issue #689, opened 2026-05-18
Peer penalty system bans honest peers during observer joins on WAN


grantkee:
## Problem

The peer-reputation system currently treats transport-level failures (peer disconnects, dial failures, IO errors, connection closures) the same way it treats protocol-level violations. The outbound-request path applies `Penalty::Medium` (-5.0) on every `OutboundFailure` regardless of cause, and the inbound path applies `Penalty::Medium` on any IO error and `Penalty::Fatal` on any unsupported-protocol mismatch. With the default ban threshold at -50.0, roughly ten transport blips against the same peer push it past the threshold.

This produces a ban cascade when a new observer joins a WAN-deployed committee. The joining node opens many concurrent request/response streams to the same small set of authorities. A single brief network interruption — common over real WANs — flushes N in-flight requests, each surfacing as an `OutboundFailure::ConnectionClosed`. The requester applies N × Medium against the responder, instantly banning a perfectly honest peer. Once banned, the 12-hour lockout (`banned_before_decay_secs = 43200`) and 600-second score halflife mean the peer cannot recover within any operator-attention window; the observer is effectively isolated until manual intervention.

Several other call sites compound the problem by penalizing behavior that is benign on a healthy network: a peer asking for a consensus header the local node has not yet produced, a peer republishing a slightly stale (but signature-valid) kademlia record after a restart, and a peer running a slightly older protocol version during a rolling upgrade. None of these are misbehavior signals — they are normal artifacts of distributed systems catching up to each other.

The risk is not theoretical. The existing `test_outbound_failure_malicious_request` test had been disabled with a TODO referencing this exact failure mode, and observer-join failures on WAN deployments are the symptom that prompted this work.

## Solution

The penalty system should distinguish between transport-level failures (which indicate network conditions, not peer fault) and protocol-level violations (which indicate either misbehavior or a real correctness problem). Transport failures should carry no penalty; protocol violations should retain their existing severity. Concretely:

- Outbound failures on dial, connection-closed, and IO errors should not penalize the responder. Outbound timeouts remain a `Mild` signal. Unsupported-protocol responses are a real misconfiguration and should remain `Severe`, but not `Fatal` — rolling upgrades briefly run mixed versions and should not produce permanent bans.
[...]

PR #620, opened 2026-04-08
network-libp2p security review 

fabcrowd:
## Summary

Adds a **review-tn**-formatted security assessment for [\crates/network-libp2p\](https://github.com/Telcoin-Association/telcoin-network/tree/main/crates/network-libp2p) (\	n-network-libp2p\).

- **New files:** \.audit/README.md\, \.audit/reports/network-libp2p-review.md\
- **Findings:** NL-001 through NL-010 (5 High, 5 Medium); each marked **Unconfirmed** pending maintainer triage.
- **Scope:** Peer ban / IP accounting, temporary ban vs dial and discovery, disconnect + peer-exchange flow, inbound Kad \AddProvider\, \SendRequestDirect\ vs \uth_to_peer\, gossip validation window, Kad vs live connection on PX disconnect, \expect\ in hot path, unbounded \discovery_peers\ growth.

## Notes

- This PR is **documentation only**; no runtime code changes.
- Source: AI-assisted review with line-level citations in the report; human confirmation still required before treating items as confirmed vulnerabilities.

## Suggested next steps for reviewers

1. Walk the cited locations in \
etwork-libp2p\ and confirm or dispute each NL item.
2. For accepted items, open follow-up issues or fixes as appropriate.

PR #760, opened 2026-06-17
  fix(network-libp2p): make operator-set peer-scoring thresholds take effect

### LINKED ISSUE #746: network-libp2p: operator-set peer-scoring ban/disconnect thresholds are silently ignored (`Peer` is pinned to `PeerConfig::default()`), and two documented `ScoreConfig` fields are dead
## Summary

The peer-scoring config exposes ban and disconnect thresholds that an operator can set in their node config, but the values that actually drive ban and disconnect decisions are taken from a defaulted config that the operator's settings never reach.  Each `Peer` decides its reputation from its own `self.config` field, and every `Peer` is constructed with `PeerConfig::default()`.  The operator's `PeerConfig` is deserialized and stored in the `PeerManager`, but it is never threaded into the individual `Peer` structs, so `min_score_for_ban` and `min_score_for_disconnect` set by an operator have no effect.  Separately, two documented `ScoreConfig` fields are effectively dead in production.

### The live ban/disconnect decision reads the per-`Peer` config (`crates/network-libp2p/src/peers/peer.rs:135-141`)

```rust
pub(super) fn reputation(&self) -> Reputation {
    match self.score.aggregate_score() {
        score if score <= self.config.min_score_for_ban => Reputation::Banned,
        score if score <= self.config.min_score_for_disconnect => Reputation::Disconnected,
        _ => Reputation::Trusted,
    }
}
```

`reputation()` is the production decision path: it is consumed by `process_penalty` (`crates/network-libp2p/src/peers/all_peers.rs:179-183`), by the heartbeat reputation update (`crates/network-libp2p/src/peers/peer.rs:284-301`), and by `peer_banned` (`crates/network-libp2p/src/peers/all_peers.rs:747`).

### Every `Peer` is built with a defaulted config (`crates/network-libp2p/src/peers/peer.rs`)

The `config: PeerConfig` field (peer.rs:30) is set to `Default::default()` in all constructors (peer.rs:61, 80, 100), and the derived `Peer::default()` is used for unknown peers (`crates/network-libp2p/src/peers/all_peers.rs:247`), which also yields `PeerConfig::default()`.  Nothing anywhere mutates `peer.config` after construction.

### The operator's `PeerConfig` is read, but never reaches the `Peer` (`crates/network-libp2p/src/peers/manager.rs:78-102`)

`PeerManager::new` stores the operator `PeerConfig` (`config: *config`, manager.rs:93) and uses it for heartbeat, dial, and connection-limit math, and it does call `init_peer_score_config(config.score_config)` (manager.rs:90).  It never passes `PeerConfig` (or its `min_score_for_ban` / `min_score_for_disconnect`) into the `Peer` structs, which is where the ban/disconnect thresholds are actually read.

The operator knobs that are ignored are real config fields: `PeerConfig::min_score_for_disconnect` (`crates/config/src/network.rs:292`, default -20.0) and `min_score_for_ban` (`crates/config/src/network.rs:294`, default -50.0).

### Two dead `ScoreConfig` fields

- `ScoreConfig::min_application_score_before_ban` (`crates/config/src/network.rs:400`, default -60.0) is never read in production code.
- `ScoreConfig::min_score_before_disconnect` (`crates/config/src/network.rs:410`, default -20.0) is referenced only in tests; production code does not read it.
- For contrast, `ScoreConfig::min_score_before_ban` (network.rs:412) is read, but only by `Score::update_score`'s decay-lockout (`crates/network-libp2p/src/peers/score.rs:141-160`), not by the ban decision itself.  So even the live `ScoreConfig` ban threshold governs decay timing, not the actual ban trigger.

## Impact

User experience and operability.  The config surface presents ban and disconnect tuning controls that are silent no-ops.  Concretely, an operator who relaxes `min_score_for_ban` to stop banning honest peers during WAN sync lag (the situation in #689) gets zero behavioral change and no warning, because the actual threshold is always the compiled-in default.  This also means the documented `ScoreConfig` floors do not do what their names and docs imply.

## Suggested fix

Thread the operator's scoring thresholds into the decision path.  Either:

1. Pass the operator `PeerConfig` into each `Peer` at construction so `reputation()` reads operator-set `min_score_for_ban` / `min_score_for_disconnect`, or
2. Move the ban/disconnect thresholds to a single source of truth (the global score config, already plumbed via `init_peer_score_config`) and have `reputation()` read from it, removing the per-`Peer` `PeerConfig` copy.

Then either wire up or delete `ScoreConfig::min_application_score_before_ban` and `min_score_before_disconnect` so the config surface matches behavior.

## Scope and novelty

Confirmed end to end on `main` (`0bd36414`).  This is a config-plumbing defect; it spans `crates/config` and `crates/network-libp2p` but the behavior lives in the network crate.  No existing issue or PR tracks it; related but distinct from #689 (penalty harshness on WAN), which assumes the thresholds are tunable.  Not covered by the `network-libp2p` security review (PR #620, NL-001..NL-010).

MavenRain:
Fixes #746.

## Problem

The peer-scoring config exposed ban/disconnect thresholds an operator could set, but the
values that actually drove ban/disconnect decisions came from a defaulted config the
operator's settings never reached.

`Peer::reputation()` decided ban/disconnect from `self.config` (a per-`Peer` `PeerConfig`),
and every `Peer` was constructed with `PeerConfig::default()`. The operator's `PeerConfig`
was deserialized and stored on the `PeerManager`, but never threaded into the individual
`Peer` structs. So `min_score_for_ban` / `min_score_for_disconnect` set by an operator were
silent no-ops: the live thresholds were always the compiled-in defaults.

Separately, two documented `ScoreConfig` fields were effectively dead:
`min_application_score_before_ban` (never read in production) and `min_score_before_disconnect`
(referenced only in tests).

## Fix

Consolidate the ban/disconnect thresholds onto the single source of truth that already
governs scoring: the global `ScoreConfig`. `Score` is already fully driven by
`GLOBAL_SCORE_CONFIG` (initialized from the operator config via `init_peer_score_config`),
and `penalty.md` already documented `ScoreConfig::min_score_before_ban` /
`min_score_before_disconnect` as *the* thresholds. This change makes the code match that.

- `Score::reputation()` (new) reads the ban/disconnect thresholds from the global
  `ScoreConfig`. `Peer::reputation()` now delegates to it. Because the ban threshold is the
  same `min_score_before_ban` used by `Score::is_banned()`, the reputation decision and the
  ban/decay lockout stay consistent.
- Removed the now-redundant, never-threaded `PeerConfig::min_score_for_ban` /
  `min_score_for_disconnect`. Their values duplicated the `ScoreConfig` fields (same defaults,
  -50 / -20).
- Removed the never-read `ScoreConfig::min_application_score_before_ban`. With this change
  `min_score_before_disconnect` is now wired into the live decision, so the config surface
  matches behavior.
- Pure threshold logic is split into a `reputation_for(score, &ScoreConfig)` helper so it can
  be unit-tested against arbitrary operator configs without touching the process-global
  `OnceLock`.
- Updated `penalty.md` line references that shifted in `network.rs`.

## Behavior change for operators

An operator who relaxes the ban/disconnect thresholds (e.g. to stop banning honest peers
during WAN sync lag, the situation in #689) now gets the expected behavior. The knobs live in
the `score_config` section: `min_score_before_ban` and `min_score_before_disconnect`.

Config migration note: the removed keys (`min_score_for_ban`, `min_score_for_disconnect`,
`min_application_score_before_ban`) were already inert. `PeerConfig` and `ScoreConfig` use
`#[serde(default)]` (no `deny_unknown_fields`), so existing TOML that still sets the old keys
continues to deserialize; the keys are simply ignored. Operators should move ban/disconnect
tuning to `score_config.min_score_before_ban` / `score_config.min_score_before_disconnect`.

## Tests

- New `reputation_honors_operator_thresholds`: a peer at score -60 is `Banned` under strict
  (-50 ban) thresholds but `Trusted` under an operator-relaxed config (-95 ban). This is the
  property that was broken before #746.
- New `reputation_threshold_boundaries`: covers the at-or-below boundary and ban-over-disconnect
  precedence.
- Existing `tn-network-libp2p` lib tests already assert reputation transitions at
  `ScoreConfig::min_score_before_disconnect`; they now exercise the genuinely wired path.

`cargo test -p tn-network-libp2p --lib`: 152 passed (150 + 2 new). `cargo fmt --check` and
`cargo clippy --all-targets -- -D warnings` clean for both crates.

## Scope

Confirmed on `main` at `1bbec7c9`. Config-plumbing defect spanning `crates/config` and
`crates/network-libp2p`. Related but distinct from #689 (penalty harshness on WAN), which
assumed these thresholds were tunable. Not covered by the `network-libp2p` security review
(PR #620, NL-001..NL-010).

### DIFF
--- a/crates/config/src/network.rs
+++ b/crates/config/src/network.rs
@@ -288,10 +288,6 @@ pub struct PeerConfig {
     pub target_num_peers: usize,
     /// The timeout for dialing a peer.
     pub dial_timeout: Duration,
-    /// The threshold before a peer is disconnected
-    pub min_score_for_disconnect: f64,
-    /// The threshold before a peer is banned.
-    pub min_score_for_ban: f64,
     /// A fraction of `Self::target_num_peers` that is allowed to connect to this node in excess of
     /// `PeerManager::target_num_peers`.
     ///
@@ -335,8 +331,6 @@ impl Default for PeerConfig {
             heartbeat_interval: 30,
             target_num_peers,
             dial_timeout: Duration::from_secs(15),
-            min_score_for_disconnect: -20.0,
-            min_score_for_ban: -50.0,
             peer_excess_factor: 0.3,
             priority_peer_excess: 0.2,
             target_outbound_only_factor: 0.3,
@@ -395,9 +389,6 @@ impl PeerConfig {
 pub struct ScoreConfig {
     /// The default score for new peers.
     pub default_score: f64,
-    /// The threshold for a peer's score before they are banned, regardless of any other scoring
-    /// parameters.
-    pub min_application_score_before_ban: f64,
     /// The maximum score a peer can obtain.
     pub max_score: f64,
     /// The minimum score a peer can obtain.
@@ -406,9 +397,15 @@ pub struct ScoreConfig {
     pub score_halflife: f64,
     /// The minimum amount of time (seconds) a peer is banned before their score begins to decay.
     pub banned_before_decay_secs: u64,
-    /// Minimum score before a peer is disconnected.
+    /// Aggregate score at or below which a peer is disconnected.
+    ///
+    /// This is the live disconnect threshold read by the reputation decision; lowering it makes
+    /// the node more tolerant of misbehaving peers before disconnecting them.
     pub min_score_before_disconnect: f64,
-    /// Minimum score before a peer is banned.
+    /// Aggregate score at or below which a peer is banned.
+    ///
+    /// This is the live ban threshold: it drives both the reputation decision and the ban/decay
+    /// lockout, so the two always agree.
     pub min_score_before_ban: f64,
 }
 
@@ -416,7 +413,6 @@ impl Default for ScoreConfig {
     fn default() -> Self {
         ScoreConfig {
             default_score: 0.0,
-            min_application_score_before_ban: -60.0,
             max_score: 100.0,
             min_score: -100.0,
             // 5-minute halflife: transient WAN penalties (peer flap, slow request) decay
--- a/crates/network-libp2p/src/peers/peer.rs
+++ b/crates/network-libp2p/src/peers/peer.rs
@@ -11,7 +11,6 @@ use libp2p::{
     PeerId,
 };
 use std::{collections::HashSet, net::IpAddr, time::Instant};
-use tn_config::PeerConfig;
 use tn_types::{BlsPublicKey, NetworkPublicKey};
 use tracing::{error, warn};
 
@@ -26,8 +25,6 @@ pub(super) struct Peer {
     bls_public_key: Option<BlsPublicKey>,
     /// The peers network public key (libp2p public key).
     network_key: Option<NetworkPublicKey>,
-    /// The config
-    config: PeerConfig,
     /// The peer's score - used to derive [Reputation].
     score: Score,
     /// The multiaddrs this node has witnessed the peer using.
@@ -62,7 +59,6 @@ impl Peer {
             network_key: Some(network_key),
             score: Score::new_max(),
             operator_allowlisted: true,
-            config: Default::default(),
             multiaddrs: Default::default(),
             connection_status: Default::default(),
             connection_direction: Default::default(),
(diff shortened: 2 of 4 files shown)


grantkee (approved):
Good find! LGTM

sstanfield (approved):

PR #764, opened 2026-06-17
 Serialize epoch pack install against epoch transition

### LINKED ISSUE #763: bug: Restarting validator can crash during epoch catch-up: ENOENT race on the epoch pack directory
## Problem

A validator that restarts while behind the chain runs two subsystems at once: the epoch-transition loop advances it forward epoch by epoch, and state-sync downloads any epoch pack files it is missing. Both write to the same on-disk `epoch-{N}` directory, and there is no coordination between them.

State-sync installs a freshly downloaded pack by deleting the live `epoch-{N}` directory and renaming the imported copy into place. The delete leaves a brief window where `epoch-{N}` does not exist — a window the code already acknowledges in a comment but does not guard against. If the epoch-transition loop opens that same epoch's pack during the window (its append-open runs `create_dir_all` and then opens the data file), the open lands after the directory has been unlinked and fails with `No such file or directory`.

That error is fatal. It propagates up through the epoch task, the consensus-output channel closes, the execution engine exits, and the node dies on boot — it never rebinds its RPC port. Externally this looks like a node that restarts and silently fails to come back. The pack-file subsystem is recent (#741), and the restart-and-catch-up path is exactly when both subsystems run concurrently against the same epoch, so the window is reachable in normal operation rather than only under contrived timing. An end-to-end epoch-sync test reproduces it as an intermittent crash.

The replace cannot be made safe by reordering alone: there is no portable way to atomically swap a directory that another task may be opening, so the two paths genuinely need to be serialized.

## Solution

Introduce a single install lock on the consensus chain that serializes the two directory-mutating critical sections against each other. The epoch-transition path should hold the lock across its open/append and current-pack swap; the state-sync path should hold the same lock across its delete + rename and the subsequent cache invalidation. With both sides serialized, the append-open always observes a consistent directory — either before the delete or after the rename, never inside the window — so the ENOENT cannot occur.

The lock should be acquired in state-sync only after the network download has finished, not around it. The download writes solely into a temporary import directory and never touches the live directory, so holding the lock across multi-second network I/O would needlessly block unrelated epoch transitions for no safety benefit.

Because both critical sections cross `.await` points, the lock should be an async mutex rather than the synchronous mutexes used for the chain's other state. It should always be acquired before those existing locks so there is one consistent lock order and no deadlock; read-only paths do not take it and are unaffected. The fix is localized to the consensus-chain storage layer — no change to the pack or data-file layers, and no new channels or task restructuring.

Read-only opens can still hit the same transient window, but they return errors that callers already degrade on gracefully and do not crash the node, so they are intentionally left outside the lock to avoid over-serializing reads.

grantkee:
- Adds a `pack_install` async mutex to `ConsensusChain` that serializes the two paths which mutate the same `epoch-{N}` directory: `new_epoch` (open/append) and `stream_import` (remove + rename). Without it, a restarting validator catching up across epochs could open a pack during the brief window where state-sync had unlinked the directory, hit `No such file or directory`, and crash on boot before binding its RPC port.
- `new_epoch` holds the lock across the whole function; `stream_import` takes the same lock only after the download completes — guarding the remove/rename + cache invalidation but not the multi-second network I/O, which only touches the temp import dir.
- Lock order is always `pack_install` → `current_pack`/`recent_packs`; read-only paths don't take it, so no deadlock and no added contention on reads. `tokio::sync::Mutex` because both critical sections cross `.await`.
- Localized to `crates/storage/src/consensus.rs` — no pack/data-file changes. `cargo check`, `clippy`, and `cargo nextest run -p tn-storage` (92/92) all pass.

closes #763

### DIFF
--- a/crates/storage/src/consensus.rs
+++ b/crates/storage/src/consensus.rs
@@ -282,6 +282,13 @@ pub struct ConsensusChain {
     /// Simple cache of recent pack files.
     recent_packs: Arc<Mutex<VecDeque<ConsensusPack>>>,
     epochs: Arc<EpochRecordDb>,
+    /// Serializes epoch-{N} directory mutation between `new_epoch` (open/append)
+    /// and `stream_import` (remove+rename), preventing a transient-ENOENT crash.
+    ///
+    /// Both critical sections cross `.await` points, so this is a `tokio::sync::Mutex`
+    /// (not the `parking_lot::Mutex` used for the other fields). Always acquired *before*
+    /// `current_pack`/`recent_packs` to keep a single lock order and avoid deadlock.
+    pack_install: Arc<tokio::sync::Mutex<()>>,
 }
 
 impl ConsensusChain {
@@ -297,7 +304,8 @@ impl ConsensusChain {
         ));
         let recent_packs = Arc::new(Mutex::new(VecDeque::default()));
         let epochs = Arc::new(EpochRecordDb::open(&base_path)?);
-        Ok(Self { base_path, current_pack, latest_consensus, recent_packs, epochs })
+        let pack_install = Arc::new(tokio::sync::Mutex::new(()));
+        Ok(Self { base_path, current_pack, latest_consensus, recent_packs, epochs, pack_install })
     }
 
     /// Create a new empty consensus chain with a dummy epoch 0 pack ready.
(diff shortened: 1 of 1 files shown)


MavenRain (approved):
LGTM

sstanfield (approved):
Good catch.

PR #768, opened 2026-06-18
feat(network): typed sync-frame layer registered alongside /tn-stream (739, step 4)

MavenRain:
Step 4 of the #739 plan: the **typed sync-frame layer**, additive. It introduces
the framing and typed requests that the worker-batch, epoch-pack, and
missing-certificate cutovers (steps 5 to 7) will ride on, and registers the new
protocols, while leaving every current call site on the legacy `/tn-stream`
path. No bulk traffic moves yet.

### What changed

- **New `sync` module (`crates/network-libp2p/src/sync/`).**
  - `SyncFrame<R>` is one frame per `encode_message`/`decode_message` unit
    (length-prefixed, snappy-compressed, BCS), so the BCS enum discriminant is
    the frame tag: `Req(R)` (REQ, the typed request), `Ack` (ACK), `Deny(DenyReason)`
    (DENY), `Data(Vec<u8>)` (DATA, opaque payload), `End` (END), `Err(SyncFrameError)`
    (ERR). An exchange opens with `Req`, the responder answers `Ack` or `Deny`,
    and an accepted exchange streams `Data` frames terminated by `End` (or aborts
    with `Err`).
  - `DenyReason` (`AtCapacity`, `Unavailable`) gives an over-capacity responder a
    way to decline immediately, so the requester does not burn its open timeout.
  - `read_frame` / `write_frame` are thin, named, documented wrappers over
    `decode_message` / `encode_message` that pin the message type to `SyncFrame`.
- **Typed request enums (shared, `pub` and re-exported from the crate root).**
  - `WorkerSyncRequest::Batches { batch_digests, epoch }` folds the legacy
    `RequestBatchesStream` ack-plus-digest handshake into a single open: the
    digests travel in the request frame, so the responder needs no
    `(peer, digest)` pending map.
  - `PrimarySyncRequest::EpochPack { epoch }` and
    `PrimarySyncRequest::MissingCertificates { exclusive_lower_bound, skip_rounds }`.
    The missing-certs variant drops `max_response_size`: a streamed, responder-ended
    reply is no longer bounded by the request-response codec's fixed cap, so the
    caller need not pre-declare a budget.
  - Defined with `tn-types` primitives only (the crate cannot depend on the
    consumer crates), which is why the missing-certs fields are modeled inline
    rather than reusing `primary`'s `MissingCertificatesRequest`.
- **Protocol registration.** `NetworkType::sync_protocol()` returns
  `/tn-primary-sync/1.0.0` or `/tn-worker-{id}-sync/1.0.0`, mirroring the existing
  `req_res_protocol` / `kad_protocol`. `TNStreamProtocol` now advertises a list,
  `[/tn-stream/0.0.1, <sync>]`, threaded `StreamBehavior::new(network_type)` to
  each `StreamHandler` and into both the inbound listen protocol and outbound
  open. The legacy protocol is first in dialer-preference order, so every existing
  open still negotiates `/tn-stream`; a responder now also accepts the sync
  protocol.

### Why it stays additive

The legacy `/tn-stream/0.0.1` path is byte-for-byte unchanged: the inbound and
outbound upgrades still return the raw stream regardless of which protocol
negotiated, and no call site opens a sync stream. Because `/tn-stream` is
advertised first, existing opens negotiate it against both item-4 and pre-item-4
peers, and a pre-item-4 dialer (which only proposes `/tn-stream`) negotiates it
against an item-4 listener. The sync protocols are registered but inert: nothing
opens one, so none is ever negotiated in an item-4 fleet. Routing an inbound
stream by negotiated protocol arrives with the first cutover (step 5), which is
also where the first opener lands.

### Cross-cutting rule honored

BCS encodes enum variant indices, so `frame_tags_are_stable` pins each
`SyncFrame` tag's exact bytes (`Req`=0, `Ack`=1, `Deny`=2, `Data`=3, `End`=4,
`Err`=5, plus the nested reason indices). No variant may be reordered or removed
before the coordinated `/0.0.2` bump; the test makes a regression loud.

### Verification

All green on stable 1.94 (build/test) and nightly-2026-03-20 (fmt/clippy):

- `cargo build -p tn-network-libp2p`
- `cargo +nightly fmt --check`
- `cargo +nightly clippy -p tn-network-libp2p --all-features --all-targets` (also
  compiles the consumer crates `tn-primary`, `tn-worker`, `tn-executor`,
  `tn-node` as dev-deps: they build zero-edit against the new public surface,
  which only grows)
- `cargo test -p tn-network-libp2p --lib`: **181 pass**, including 10 new (frame
  round-trips for every tag, `frame_tags_are_stable`, oversized-frame rejection,
  the three request round-trips, `request_tags_are_stable`, protocol-info
  ordering, and sync-protocol registration)
- `RUSTDOCFLAGS=-D warnings cargo doc -p tn-network-libp2p --no-deps`

A 4-lens adversarial review (wire-format, libp2p negotiation, conventions,
completeness) drove two additions: `request_tags_are_stable` pins the typed
request enums' BCS discriminants (they ride on the wire inside `Req`, so they get
the same encoding-compat pin as the frame tags), and a note at `upgrade_inbound`
marks the negotiated-protocol seam that the step-5 cutover branches on. A live
multistream-select negotiation test (asserting an outbound open still picks
`/tn-stream` against a dual-protocol peer) is intentionally deferred to step 5,
which the issue assigns the cross-version integration harness; step 4 covers the
advertised order with `protocol_info_preserves_order` and relies on libp2p's
dialer-preference negotiation.

### Non-goals

No call site switches to the sync protocols (steps 5 to 7). No request-response,
gossipsub, kad, or scoring-weight changes. The `Data` payload encoding and the
buffer-ownership model for a live exchange are deliberately left to the cutovers
that define them.
```

### DIFF
--- a/crates/network-libp2p/src/sync/mod.rs
+++ b/crates/network-libp2p/src/sync/mod.rs
@@ -23,7 +23,7 @@
 //! ## Protocols
 //!
 //! The frames ride on streams negotiated with the per-role sync protocols
-//! (`/tn-primary-sync/1.0.0`, `/tn-worker-{id}-sync/1.0.0`), which the stream
+//! (`/tn-primary-sync/0.0.1`, `/tn-worker-{id}-sync/0.0.1`), which the stream
 //! behaviour registers alongside the legacy `/tn-stream/0.0.1` upgrade. This
 //! layer is inert for now: no call site opens a sync stream until the
 //! per-exchange cutovers migrate the worker batch, epoch pack, and missing
--- a/crates/network-libp2p/src/sync/mod.rs
+++ b/crates/network-libp2p/src/sync/mod.rs
@@ -1,0 +1,39 @@
+//! Typed sync-frame layer for bulk request/response exchanges.
+//!
+//! The legacy bulk path opens a raw `/tn-stream` substream and correlates it
+//! back to a prior request-response handshake by writing a 32-byte digest. This
+//! module is the additive groundwork for folding that handshake into the stream
+//! itself: an exchange carries its request as the first frame, and the responder
+//! answers in the same stream.
+//!
+//! ## Framing
+//!
+//! Every exchange is a sequence of [`SyncFrame`]s, each written as a single
+//! [`encode_message`]/[`decode_message`] unit (length-prefixed, snappy
+//! compressed, BCS encoded). The BCS enum discriminant is the frame tag:
+//!
+//! 1. The requester opens with a [`SyncFrame::Req`] carrying the typed request
+//!    ([`WorkerSyncRequest`] or [`PrimarySyncRequest`]).
+//! 2. The responder answers [`SyncFrame::Ack`] to accept or [`SyncFrame::Deny`] to shed load, so an
+//!    over-capacity responder declines immediately instead of making the requester wait out its
+//!    timeout.
+//! 3. An accepted exchange streams zero or more [`SyncFrame::Data`] frames and finishes with
+//!    [`SyncFrame::End`], or aborts with [`SyncFrame::Err`].
+//!
+//! ## Protocols
+//!
+//! The frames ride on streams negotiated with the per-role sync protocols
+//! (`/tn-primary-sync/1.0.0`, `/tn-worker-{id}-sync/1.0.0`), which the stream
+//! behaviour registers alongside the legacy `/tn-stream/0.0.1` upgrade. This
+//! layer is inert for now: no call site opens a sync stream until the
+//! per-exchange cutovers migrate the worker batch, epoch pack, and missing
+//! certificate paths.
+//!
+//! [`encode_message`]: crate::encode_message
+//! [`decode_message`]: crate::decode_message
+
+mod frame;
+mod request;
+
+pub use frame::{read_frame, write_frame, DenyReason, SyncFrame, SyncFrameError};
+pub use request::{PrimarySyncRequest, WorkerSyncRequest};
--- a/crates/network-libp2p/src/sync/request.rs
+++ b/crates/network-libp2p/src/sync/request.rs
@@ -6,7 +6,7 @@ use tn_types::{AuthorityIdentifier, BlockHash, Epoch, Round};
 
 /// A bulk-sync request from a worker, carried in the opening
 /// [`SyncFrame::Req`](crate::sync::SyncFrame::Req) of a
-/// `/tn-worker-{id}-sync/1.0.0` exchange.
+/// `/tn-worker-{id}-sync/0.0.1` exchange.
 ///
 /// The variant order is the wire format: this enum's BCS discriminant rides on
 /// the wire inside the `Req` frame, so variants must not be reordered or removed
@@ -28,7 +28,7 @@ pub enum WorkerSyncRequest {
 }
 
 /// A bulk-sync request from a primary, carried in the opening
-/// [`SyncFrame::Req`](crate::sync::SyncFrame::Req) of a `/tn-primary-sync/1.0.0`
+/// [`SyncFrame::Req`](crate::sync::SyncFrame::Req) of a `/tn-primary-sync/0.0.1`
 /// exchange.
 ///
 /// The variant order is the wire format: this enum's BCS discriminant rides on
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: crates/network-libp2p/src/types.rs [hunk 1]
```diff
@@ -93,6 +93,23 @@ impl NetworkType {
...
+    /// Bulk-sync streaming wire protocol, isolated per role (and per worker).
+    ///
+    /// The stream behaviour registers this alongside the legacy
+    /// `/tn-stream/0.0.1` upgrade so a responder accepts it; the typed
+    /// [`SyncFrame`](crate::sync::SyncFrame) layer rides on streams negotiated
+    /// with this protocol. Additive for now: no call site opens it until the
+    /// per-exchange cutovers migrate the bulk paths.
+    pub(crate) fn sync_protocol(&self) -> StreamProtocol {
+        match self {
+            Self::Primary => StreamProtocol::new("/tn-primary-sync/1.0.0"),
```
grantkee: nit: other protocols are using `0.0.1` - I don't have a great reason to use 0/1 for the major semver but I would like to keep it consistent until mainnet launch

crates/network-libp2p/src/sync/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+//! 2. The responder answers [`SyncFrame::Ack`] to accept or [`SyncFrame::Deny`] to shed load, so an
+//!    over-capacity responder declines immediately instead of making the requester wait out its
+//!    timeout.
+//! 3. An accepted exchange streams zero or more [`SyncFrame::Data`] frames and finishes with
+//!    [`SyncFrame::End`], or aborts with [`SyncFrame::Err`].
+//!
+//! ## Protocols
+//!
+//! The frames ride on streams negotiated with the per-role sync protocols
+//! (`/tn-primary-sync/1.0.0`, `/tn-worker-{id}-sync/1.0.0`), which the stream
```
grantkee: nit: consistent versions across all protocols pre-mainnet

crates/network-libp2p/src/sync/request.rs
```diff
@@ -0,0 +1,132 @@
+//! Typed requests carried in the opening frame of a bulk-sync exchange.
+
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeSet;
+use tn_types::{AuthorityIdentifier, BlockHash, Epoch, Round};
+
+/// A bulk-sync request from a worker, carried in the opening
+/// [`SyncFrame::Req`](crate::sync::SyncFrame::Req) of a
+/// `/tn-worker-{id}-sync/1.0.0` exchange.
```
grantkee: nit: consistent versions across all protocols pre-mainnet

crates/network-libp2p/src/sync/request.rs
```diff
@@ -0,0 +1,132 @@
...
+    Batches {
+        /// The batch digests being requested.
+        batch_digests: BTreeSet<BlockHash>,
+        /// The epoch that produced these batches.
+        epoch: Epoch,
+    },
+}
+
+/// A bulk-sync request from a primary, carried in the opening
+/// [`SyncFrame::Req`](crate::sync::SyncFrame::Req) of a `/tn-primary-sync/1.0.0`
```
grantkee: nit: consistent versions across all protocols pre-mainnet

crates/network-libp2p/src/types.rs
```diff
@@ -93,6 +93,23 @@ impl NetworkType {
...
+            Self::Worker(id) => {
+                StreamProtocol::try_from_owned(format!("/tn-worker-{id}-sync/1.0.0"))
```
grantkee: nit: consistent versions across all protocols pre-mainnet

grantkee (dismissed):
LGTM! I'd like to standardize versions before mainnet, but don't have a strong preference between v1.0.0 or v0.0.1

Not a blocker for this PR. Good work!

grantkee (approved):

sstanfield (approved):
LGTM!

sstanfield:
> LGTM! I'd like to standardize versions before mainnet, but don't have a strong preference between v1.0.0 or v0.0.1
> 
> Not a blocker for this PR. Good work!

Philosophically I feel like main net should be 1.0.0.  It is the first real release, no?  Not a hill to die on or anything, these are for keeping compatible protocols working etc but makes sense to me.  Seems like something to address in the final lead up to main net though.

MavenRain:
> > LGTM! I'd like to standardize versions before mainnet, but don't have a strong preference between v1.0.0 or v0.0.1
> > Not a blocker for this PR. Good work!
> 
> Philosophically I feel like main net should be 1.0.0. It is the first real release, no? Not a hill to die on or anything, these are for keeping compatible protocols working etc but makes sense to me. Seems like something to address in the final lead up to main net though.

Yes, we are in 100% agreement on this . . . I will remember to raise the issue closer to the time.

PR #624, opened 2026-04-09
feat: add Execution Extensions (ExEx) framework for observer nodes

ArieJones:
### Summary

Adds a plugin system for Execution Extensions (ExEx) — long-running tasks that receive real-time notifications about chain state changes and consensus events. This enables external tooling (indexers, bridges, analytics) to hook into the node without modifying core code.

### Motivation

Observer nodes need a way to react to block execution and consensus activity. Rather than forking the node binary, ExExs provide a composable, backpressure-aware notification pipeline that external developers can use to build custom integrations.

### Changes

**New crate: `tn-exex`**
- `TnExExNotification` — event type carrying `ChainCommitted`, `CertificateCreated`, and `ConsensusCommitted` variants
- `TnExExManager` — central dispatcher that fans out notifications to all installed ExExs with backpressure (blocks execution if any ExEx falls behind)
- `TnExExLauncher` — install-and-launch API for registering ExEx factory functions
- `TnExExContext` — per-ExEx handle providing blockchain state access and a notification stream
- `ReplayStream` — wraps the notification receiver with optional historical block replay on startup
- `TnExExEvent` — ExExs report `FinishedHeight` back to the manager to track readiness
- `TnExExManagerHandle` — lightweight cloneable handle used throughout the engine layer; `TnExExManagerHandle::empty()` is a zero-overhead no-op suitable for tests and non-ExEx builds
- Comprehensive docs: `crates/tn-exex/README.md` and `crates/tn-exex/CUSTOM_NODE.md`

**Consensus bus integration** (`tn-primary`)
- Added `exex_certificates` and `exex_committed_sub_dags` broadcast channels to `ConsensusBusApp` (app-lifetime, persists across epoch boundaries)
- `Certifier` sends `Arc<Certificate>` on `exex_certificates` after each own certificate is produced
- `CertManager` sends `Arc<Certificate>` on `exex_certificates` after each verified peer certificate is accepted
- Consensus `State` sends `Arc<CommittedSubDag>` on `exex_committed_sub_dags` after each Bullshark commit

**Engine + tn-reth integration**
- `ExecutorEngine` accepts a `TnExExManagerHandle` and threads it through to `execute_consensus_output`
- `RethEnv::finish_executing_output` gains a third parameter (an `ExExHandle` trait object) — sends `ChainCommitted` notifications after each block execution
- `tn-reth` adds `exex_handle` module with `EmptyExExHandle` for call sites that don't need ExEx notifications (tests, staking helpers)

[...]

PR #770, opened 2026-06-19
Exex2

### LINKED ISSUE #618: Add Execution Extensions (ExEx) plugin system
## Problem

Telcoin Network currently has no way for external modules to react to chain state changes in real time. Bridges, wallets, block explorers, and analytics pipelines that need to track transactions must poll RPC endpoints or parse logs after the fact. This creates latency, missed events, and duplicated infrastructure across every consumer that needs lifecycle visibility.

The gap is wider than what reth's built-in ExEx system covers. Reth's ExEx only emits notifications after EVM execution, but TN's Bullshark BFT consensus creates two earlier observable stages — certificate acceptance and sub-DAG commit — that happen before execution. A bridge monitoring system, for example, needs to know when a transaction's certificate is accepted (earliest confirmation) and when the sub-DAG commits (ordering finalized), not just when the EVM processes it. Without these signals, consumers have no visibility into the consensus pipeline and can't provide accurate confirmation latency to users.

There's also no mechanism for ExEx plugins to catch up after restarts. If a plugin crashes and restarts, it needs to replay historical blocks from where it left off and then seamlessly switch to live notifications. Without built-in replay support, every plugin would need to implement its own catch-up logic against the database.

## Solution

Introduce a `tn-exex` crate that provides a full-lifecycle plugin system covering three stages: certificate accepted, consensus committed, and chain executed. Each stage produces an irrevocable notification thanks to BFT finality — there are no reorgs to handle.

The system should have four main components. A `TnExExNotification` enum with three variants (`CertificateAccepted`, `ConsensusCommitted`, `ChainExecuted`) that map to the three observable lifecycle stages. A `TnExExManager` that subscribes to ConsensusBus broadcast channels and reth's canonical state stream, wrapping each event into the appropriate notification variant and fanning out to all registered plugins via bounded channels. A `TnExExContext` that gives each plugin a notification stream, an event channel for reporting progress, and access to reth's blockchain provider for querying chain state. And a `ReplayStream` that reads finalized blocks from the database in ascending order, so plugins can catch up from any starting block to the chain tip before switching to live notifications.

To feed the manager, three new broadcast channels should be added to ConsensusBus: one for own certificates (sent after the Certifier produces a certificate), one for peer certificates (sent when the cert manager accepts and forwards a peer certificate), and one for committed sub-DAGs (sent when Bullshark commits). The manager polls all four sources (three consensus channels plus the canonical state stream) in a single `Future` implementation and distributes to plugins using non-blocking sends.

Registration should use `TnBuilder::install_exex(name, async_fn)` following reth's `LaunchExEx` pattern. Each plugin is spawned as a critical task on the node-level task manager. A lifecycle example should demonstrate all three notification types to serve as a reference implementation.

https://www.paradigm.xyz/2024/05/reth-exex

grantkee:
# Add full-lifecycle ExEx plugin system

- New `crates/exex` crate: a full-lifecycle Execution Extension system covering certificate-accepted → consensus-committed → chain-executed, plus a `Lagged` reconcile signal. ExExes register via `TnBuilder::install_exex[_with_capacity]` and run as isolated non-critical tasks — a slow/stuck/panicking ExEx cannot stall consensus, delay execution, or crash the node.
- Manager fans out from 4 sources (own certs, peer certs, committed sub-DAGs, reth canonical state) over bounded per-ExEx channels with non-blocking `try_send`. Drops are not silent: the next delivery is preceded by `Lagged { missed }`. Per-ExEx capacity is configurable (default 256).
- Replay is the authoritative catch-up path: `replay_from` / `replay_and_subscribe` read finalized blocks ascending from the DB (no WAL/reorg handling — Bullshark has immediate finality). Documented asymmetry: live `ChainExecuted` carries full `BundleState`; replayed carries receipts + empty `BundleState` (read diffs from `reth_env` by block number).
- Consensus hot path stays cheap: per-stage broadcasts skip the payload clone unless `receiver_count() > 0` (the sub-DAG clone is O(n) and can be large under a Byzantine leader).
- `ConsensusBus` gains three ExEx broadcast channels + notify helpers; node manager spawns the manager as a non-critical task; `RethEnv` gains `replay_block_as_chain` and a small read surface. Returns errors on missing reth info (state-inconsistency signal) rather than swallowing.
- Two example crates: `exex-lifecycle` (logs every notification) and `exex-indexer` (stateful skeleton: replay-on-startup, uniform replay/live processing, `Lagged` reconciliation, progress reporting). Crate-level docs cover the build/replay/fidelity contract.

closes #618 
credit to #624 as a source of inspiration


crates/exex/src/context.rs
```diff
@@ -0,0 +1,198 @@
...
+    ///     // Process identically — no distinction between replay and live
+    /// }
+    /// ```
+    ///
+    /// # Delivery contract
+    ///
+    /// `ChainExecuted` notifications are delivered **at-least-once and in
+    /// monotonic height order**: the overlap window (blocks buffered on the live
+    /// channel while replay runs) is de-duplicated by height so a block is never
+    /// processed twice. If the live channel overflowed during a slow replay, the
```
MavenRain, line 154: It's curious . . . the claim is that a block can never be processed twice, although `ChainExecuted` itself can be delivered twice.

crates/exex/src/notification.rs
```diff
@@ -0,0 +1,84 @@
...
+    ///
+    /// Emitted in two cases, both surfacing the same "gap exists, go reconcile"
+    /// signal:
+    /// 1. **Downstream (manager → ExEx):** this ExEx's bounded channel was full, so the manager
+    ///    dropped notifications rather than block consensus.
+    /// 2. **Upstream (reth → manager):** the manager itself fell behind one of its source streams
+    ///    (reth's canonical-state stream or a consensus broadcast), which drop lagged items
+    ///    silently. The manager detects the resulting discontinuity and re-emits it as `Lagged`.
+    ///
+    /// On receipt, a stateful ExEx should reconcile by replaying from its last
```
MavenRain: I think I can see this for `ChainExecuted`.  Is it true for `Lagged`?

MavenRain (dismissed):
Just a couple of questions, mainly due to my ignorance in this area

crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -530,6 +548,68 @@ impl ConsensusBusApp {
         self.inner.consensus_header.subscribe()
     }
 
+    /// Broadcast sender for own certificates (ExEx).
+    pub fn exex_own_certificates(&self) -> &broadcast::Sender<Certificate> {
```
sstanfield: Exposing the sender like this will allow someone to skip to the checks done in notify.  I would not do this OR wrap the broadcast in a type that does the check.  That has been the pattern in the past and reduces external dependency on tokio (that is part of the purpose of the TnSender/Reciever interfaces).  I am not super worried about that but I think you should remove these and force the use of subscribe and notify variants- I think this will only lead to footguns and pain.

crates/exex/src/context.rs
```diff
@@ -0,0 +1,198 @@
...
+pub struct TnExExContext {
+    /// Async stream of lifecycle notifications (certificates, commits, executions).
+    ///
+    /// Live delivery is best-effort (see [`TnExExNotification`]); replay is the
+    /// authoritative catch-up path.
+    pub notifications: mpsc::Receiver<TnExExNotification>,
+    /// Bounded channel to report events (e.g., `FinishedHeight`) back to the
+    /// manager.
+    ///
+    /// Report with a non-blocking `try_send` — `FinishedHeight` is latest-wins,
```
sstanfield: Suggest you make all these fields private and use methods to enforce the expected contract.  This code is intended to be used by outside programmers, yes?  If so then I don't think documenting "thou shall not" is the best move, someone will ignore it and enforcing via code can't be ignored.

crates/exex/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+//!
+//! TN uses Bullshark BFT consensus with immediate finality. There are no reorgs,
+//! no WAL, and no reorg handling needed.
+//!
+//! # Building an ExEx
+//!
+//! An ExEx is an `async fn(`[`TnExExContext`]`) -> eyre::Result<()>` registered
+//! with `TnBuilder::install_exex` (or `install_exex_with_capacity` to size its
+//! notification buffer for a heavyweight consumer). It runs as an isolated,
+//! non-critical task: a slow, stuck, or panicking ExEx can never stall consensus,
```
sstanfield, line 22: I think this should optionally be a critical task.  If I run a node ONLY for the exex functionality then it is critical.  That's not all use cases though so should be optional.

crates/exex/src/notification.rs
```diff
@@ -0,0 +1,84 @@
...
+        is_own: bool,
+    },
+    /// A sub-DAG has been committed by the Bullshark consensus protocol.
+    ///
+    /// This means the certificates in the sub-DAG are ordered and will
+    /// be executed. The committed sub-DAG includes the leader certificate
+    /// and all certificates it commits.
+    ConsensusCommitted {
+        /// The committed sub-DAG containing ordered certificates.
+        sub_dag: Arc<CommittedSubDag>,
```
sstanfield: Should this be the ConsensusOutput so it provides everything about the committed consensus?

crates/exex/src/manager.rs
```diff
@@ -0,0 +1,472 @@
...
+                let height = h.finished_height?;
+                Some(Some(acc.map_or(height, |a: BlockNumber| a.min(height))))
+            })
+            .flatten();
+
+        self.min_finished_height_tx.send_replace(min_height);
+    }
+}
+
+impl futures::Future for TnExExManager {
```
sstanfield, line 221: Implementing Future directly seems way more complex than a simple async fn that loops over a select?  Maybe I am missing something but I don't think we should be using raw Futures when a basic select loop will be clearer and lot less error prone (no need to make sure the cx is registered everywhere it needs to be for one).  I won't die on this hill but I would rather see current Future implementations refactored out (unless they truly need to be there for some reason) vs extending this pattern into new code.

sstanfield (dismissed):
I think this looks good but had a few comments.  Not fully covered but I don't think the sub dag is sufficient for consensus output, at least the ConsensusHeader would allow the hash and number for the consensus chain.  Could probably use the parent beacon field to inject consensus when playing back historic data as well.  Also, it seems like we should allow access to the RethDB and ConsensusChain data as well.

sstanfield:
@grantkee One more thing, without access to the DBs couldn't this just be a RPC/websocket thing.  Then the exex process could be it's own process and multiple could share the same observer etc (I know you can have multiple exex). This is out of scope but might be something to consider, could dove tail with some of the light client and observers over RPC ideas.

crates/exex/src/lib.rs
grantkee: Good idea

MavenRain (dismissed):

MavenRain (approved):

crates/exex/src/manager.rs
grantkee: I agree it's more complicated, but I think the poll order matters

crates/exex/src/notification.rs
grantkee: Good point. Updated

crates/exex/src/context.rs
grantkee: Done

crates/consensus/primary/src/consensus_bus.rs
grantkee: Addressed

crates/consensus/primary/src/consensus_bus.rs
```diff
@@ -548,66 +549,52 @@ impl ConsensusBusApp {
...
-    /// Broadcast sender for own certificates (ExEx).
-    pub fn exex_own_certificates(&self) -> &broadcast::Sender<Certificate> {
-        &self.inner.exex_own_certificates
-    }
-
-    /// Broadcast sender for peer certificates (ExEx).
-    pub fn exex_peer_certificates(&self) -> &broadcast::Sender<Certificate> {
-        &self.inner.exex_peer_certificates
+    /// Broadcast sender for verified certificates (ExEx).
+    pub fn exex_certificates(&self) -> &broadcast::Sender<Certificate> {
```
sstanfield, line 553: I don't want to force another review cycle over this but why is this and the notify/sub methods both provided?  Either this is clutter or the not/sub are clutter.  When trying to figure out how events flow through the system having one thing to search for vs two or three is very helpful.  I think some other similar noise has crept into the bus but not sure we need to establish this as a pattern.  Would be better to have one way to do something.

crates/exex/src/context.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// [`report_finished_height`](Self::report_finished_height).
+    events: mpsc::Sender<TnExExEvent>,
+    /// Read-only handle to reth for EVM chain state and history.
+    reth_env: RethEnv,
+    /// Handle to the consensus chain (consensus DB); exposed for reads via
+    /// [`consensus_chain`](Self::consensus_chain).
+    consensus_chain: ConsensusChain,
+}
+
+impl TnExExContext {
```
sstanfield, line 41: Out of scope for this PR but we may want to add some common DB lookup functions in the future so a user can do some simple stuff easily.  Thinking looking up exec or consensus block type stuff.  Still expose the reth env and chain though for complex stuff.  Or maybe not, this will leak some of the Reth API insanity out but probably no way around that I guess.

sstanfield (approved):
Looks good, nothing to block merging but would not mind cleaning up the extra cruft on the consensus bus later.

crates/consensus/primary/src/consensus_bus.rs
ArieJones: I think we could do a simple follow-up to this.  This looks like it's only used by the tests now so we could clean that up now with something like this..and then it would just be simply updating the tests and address @sstanfield concern ... if we need counts or something like that in the future then we can add a narrowly named method or just export it as a metric.  

```    /// Number of ExEx certificate subscribers.
    #[cfg(test)]
    fn exex_certificates_receiver_count(&self) -> usize {
        self.inner.exex_certificates.receiver_count()
    }

    /// Number of ExEx consensus-output subscribers.
    #[cfg(test)]
    fn exex_consensus_output_receiver_count(&self) -> usize {
        self.inner.exex_consensus_output.receiver_count()
    }

ArieJones (approved):

crates/exex/src/context.rs
grantkee: Addressed

PR #771, opened 2026-06-20
Results of scan of consensus chain code.

sstanfield:
Fixes an issue where a corrupted consensus chain slot file would invalidate the other good slot file.  Same issue as:

### 9. Torn write to a single LatestConsensus slot file prevents node startup (pre-existing)
- **Location**: `crates/storage/src/consensus.rs:128-131` (propagation), `consensus.rs:93-115` (read_slot)
- **Claim**: `LatestConsensus::new` propagates `read_slot` errors with `?`: a CRC failure or short read in either slot file (e.g., torn 16-byte in-place `write_all` during power failure — the update path rewrites in place with no temp-file/rename) makes `ConsensusChain::new` fail, so the node cannot start even though the other slot is intact and packs carry ground truth. The dual-slot + CRC scheme is designed to tolerate exactly one corrupt slot, but error handling treats one bad slot as fatal. Pre-existing code in an in-scope changed file.
- **Key Question**: Does a crash mid-`write_all` of one slot file produce a CRC-failing file that hard-fails `ConsensusChain::new` (and node startup) rather than falling back to the surviving slot or pack-derived recovery?
- **Relevant Files**: crates/storage/src/consensus.rs
- **Source**: tn-review

### DIFF
--- a/crates/storage/src/consensus.rs
+++ b/crates/storage/src/consensus.rs
@@ -127,8 +127,18 @@ impl LatestConsensus {
         }
         let mut slot1 = OpenOptions::new().read(true).write(true).open(&slot1_path)?;
         let mut slot2 = OpenOptions::new().read(true).write(true).open(&slot2_path)?;
-        let (slot1_epoch, slot1_number) = Self::read_slot(&mut slot1)?;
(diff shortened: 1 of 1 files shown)


MavenRain (approved):
All good

grantkee (approved):
LGTM! Good catch

PR #772, opened 2026-06-22
network: incorporate the chain id into every libp2p protocol string and gossip topic

### LINKED ISSUE #765: Incorporate the chain id into the libp2p protocol string.
This will provide a small bit of safety to keep different chains from ever trying to cross talk.  This should not cause any lasting damage and should not really happen but if kademillia for instance "shared" records across networks it could lead to confusion.

MavenRain:
Fixes #765.

## Problem

Two Telcoin networks are kept apart only by non-overlapping bootnodes and app-layer
validation, not by the protocol layer itself. Four protocol families carry no chain
component, so nodes on different chains can negotiate connections and share gossip meshes
until app-layer validation rejects the cross-talk:

- request-response: `/tn-primary/0.0.1`, `/tn-worker-{id}/0.0.1`
- kademlia: `/tn-primary-kad/0.0.1`, `/tn-worker-{id}-kad/0.0.1`
- stream: `/tn-stream/0.0.1`
- gossipsub topics: `tn-primary`, `tn-consensus-output`, `tn-epoch-vote`, `tn-worker`, `tn-txn`

Gossipsub is the widest bleed vector: topics are not baked into the behaviour, so all gossip
rides one global `/meshsub/1.1.0` mesh and isolation is purely the topic-string hash. Two
chains using these five names would form a shared mesh and gossip each other's messages.

## Fix

A coordinated hard cutover: fold the genesis chain id into all four families so the wire
layer itself keeps chains apart. Names become `/tn-primary-{chain_id}/0.0.1`,
`/tn-primary-kad-{chain_id}/0.0.1`, `/tn-stream-{chain_id}/0.0.1`, `tn-primary-{chain_id}`,
and so on.

Genesis is the single source of truth. The chain id (`genesis.config.chain_id`) is stamped
onto the network config once at node startup and read back everywhere a protocol or topic is
built:

- `LibP2pConfig` gains a `chain_id` field. It is `#[serde(skip)]`: genesis-derived, not an
  operator tunable, never written to (or read from) the config file, so it cannot desync from
  genesis. `NetworkConfig::set_chain_id` stamps it (from `self.builder.tn_config.genesis()`)
  right after `NetworkConfig::read_config` in `EpochManager::run`. That same config object is
  threaded into both the network swarms and the per-epoch `ConsensusConfig`, so one assignment
  propagates to every reader.
- The five `LibP2pConfig::*_topic` fns take a `chain_id: u64` and format `tn-...-{chain_id}`.
- `NetworkType::req_res_protocol` / `kad_protocol` take a `chain_id` and a new
  `types::stream_protocol(chain_id)` builds the (role-agnostic) stream name. All three now use
  `StreamProtocol::try_from_owned` and return `NetworkResult`, replacing the previous
  `StreamProtocol::new` / `.expect` (a const-fn over a `&'static str` cannot embed a runtime
  id, and the result keeps construction panic-free).
- The stream subsystem carried a single `const TN_STREAM_PROTOCOL`. That const is deleted and
  an owned `StreamProtocol` is threaded `ConsensusNetwork::new` → `TNBehavior::new` →
  `StreamBehavior` → `StreamHandler` → `TNStreamProtocol`.
- The two config-less gossip publish handles (`PrimaryNetworkHandle`, `WorkerNetworkHandle`)
  gain a `chain_id` field, stamped at construction; the gossip-validation handlers read it via
  `ConsensusConfig::chain_id()`. Publish and validate therefore read the same value.

`WorkerNetworkHandle::publish_txn` previously published to a hardcoded `"tn-txn"` literal
(bypassing `worker_txn_topic()`); it now uses the namespaced fn, so the published and
subscribed txn topics match.

### Why a field on `LibP2pConfig` rather than a new `ConsensusNetwork::new` parameter

The issue suggested a `chain_id` field so wire protocols and topics read one source; this PR
follows that. Reading the stamped value everywhere means a missed stamp degrades to a
self-consistent `0` namespace (no chain isolation) rather than a publish/validate split that
would break a node talking to itself. It also leaves `ConsensusNetwork::new`'s signature (and
its test call sites) untouched.

## Deployment (breaking wire change)

This changes multistream-select negotiation, so old and new nodes will not peer; every node
must upgrade together (a coordinated cutover, tied to a release or epoch boundary). There is
no version-negotiation fallback in the codebase, and this PR adds none. A dual-advertise
transition release (advertise both old and namespaced protocols, drop the old later) was
considered and is the alternative if a rolling upgrade is required; it is meaningfully more
scope and was not pursued for this hardening.

## Tests

- `tn-config`: new `gossip_topics_are_chain_namespaced` (all five topics + the
  different-id-different-topic property), `set_chain_id_is_read_back`, and
  `chain_id_is_never_persisted`. The last asserts the field is never serialized AND that a
  config file containing an explicit `chain_id: 999` is ignored on read (reads back `0`), so an
  operator cannot override genesis from disk.
- `tn-network-libp2p`: the `test_network_type_protocol_names` lock test now asserts the
  chain-namespaced req-res/kad/stream strings (still a peer-compatibility contract). The
  existing `test_primary_worker_protocol_isolation` still passes (role isolation holds under
  namespacing).
- `tn-worker`: new `test_batch_gossip_topic_is_chain_namespaced` stamps a non-zero chain id
  (2017) onto the fixture and asserts the handler accepts the `2017` topic but rejects a
  `chain-0`-namespaced message with `InvalidTopic`. This makes the gossip-enforcement path
  load-bearing: it fails if the validator ever hardcoded the namespace instead of reading it
  from config. Existing worker/primary gossip-enforcement tests updated to the new signatures.
- The 0-defaulting `From<NetworkHandle> for PrimaryNetworkHandle` (only ever used by a test) is
  now `#[cfg(test)]`-gated, so the only way to build a production handle is `new` with an
  explicit genesis chain id.

Verification on `main` @ `27de5398`:
- `cargo test --workspace --no-run`: compiles clean.
- `cargo test` for the affected crates: `tn-config` network tests (6), `tn-network-libp2p`
  protocol-lock + 14 stream + isolation tests, `tn-primary` and `tn-worker` gossip-enforcement
  tests all pass.
- `cargo +nightly fmt --check`: clean.
- CI clippy (`cargo +nightly clippy --locked --all --all-features -- -D warnings`): clean.

## Scope

Confirmed on `main` at `27de5398`. Spans `crates/config`, `crates/network-libp2p`,
`crates/consensus/{primary,worker}`, and `crates/node`. The three `tn_*` topic consts in
`network-libp2p/src/types.rs` (`WORKER_BATCH_TOPIC`, `PRIMARY_CERT_TOPIC`,
`CONSENSUS_HEADER_TOPIC`) are unused dead code and were left untouched.

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -252,6 +252,12 @@ where
         external_addr: Multiaddr,
         rpc: Option<RpcInfo>,
     ) -> NetworkResult<Self> {
+        // Namespace every wire protocol by the genesis chain id so nodes on
+        // different chains never negotiate a connection. The id is stamped onto
+        // the network config from genesis at node startup; see
+        // `NetworkConfig::set_chain_id`.
+        let chain_id = network_config.libp2p_config().chain_id;
+
         let gossipsub_config = gossipsub::ConfigBuilder::default()
             // explicitly set default
             .heartbeat_interval(Duration::from_secs(1))
@@ -259,6 +265,14 @@ where
             .validation_mode(gossipsub::ValidationMode::Strict)
             // TN specific: filter against authorized_publishers for certain topics
             .validate_messages()
+            // Gossipsub negotiates its own `/meshsub` protocol, independent of the
+            // req-res/kad/stream names below, so without this it is the one wire
+            // protocol two chains still share: namespacing the topics keeps their
+            // messages apart but still lets cross-chain peers negotiate a gossip
+            // substream. Folding the chain id into the protocol id closes that gap.
+            // The builder appends `/1.1.0` and `/1.0.0`, yielding
+            // `/tn-meshsub-{chain_id}/1.1.0` and `/tn-meshsub-{chain_id}/1.0.0`.
+            .protocol_id_prefix(crate::types::gossip_protocol_id_prefix(chain_id))
             .build()?;
         let gossipsub = gossipsub::Behaviour::new(
             gossipsub::MessageAuthenticity::Signed(keypair.clone()),
@@ -268,12 +282,6 @@ where
 
         let tn_codec =
             TNCodec::<Req, Res>::new(network_config.libp2p_config().max_rpc_message_size);
-
-        // Namespace every wire protocol by the genesis chain id so nodes on
-        // different chains never negotiate a connection. The id is stamped onto
-        // the network config from genesis at node startup; see
-        // `NetworkConfig::set_chain_id`.
-        let chain_id = network_config.libp2p_config().chain_id;
 
         let req_res = request_response::Behaviour::with_codec(
             tn_codec,
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -29,7 +29,7 @@ use libp2p::{
         ProtocolSupport,
     },
     swarm::{NetworkBehaviour, SwarmEvent},
-    Multiaddr, PeerId, Swarm, SwarmBuilder,
+    Multiaddr, PeerId, StreamProtocol, Swarm, SwarmBuilder,
 };
 use std::{
     collections::{HashMap, HashSet, VecDeque},
@@ -94,9 +94,10 @@ where
         kademlia: kad::Behaviour<KadStore<DB>>,
         peer_config: &PeerConfig,
         metrics: PeerManagerMetrics,
+        stream_protocol: StreamProtocol,
     ) -> Self {
         let peer_manager = PeerManager::new(peer_config, metrics);
-        let stream = StreamBehavior::new();
+        let stream = StreamBehavior::new(stream_protocol);
         Self { peer_manager, gossipsub, req_res, kademlia, stream }
     }
 }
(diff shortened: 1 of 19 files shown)


sstanfield (dismissed):
Looks good.  One question (probably me missing something) but do we need to stamp the gossip topics with the chain id?  Won't including it in the protocols stop any cross chain gossip?  No big deal and maybe it is just there for completeness or something.

crates/network-libp2p/src/consensus.rs
grantkee, line 262: I think @sstanfield brought up a good point. It seems like gossip can still hit cross-chain.

@MavenRain would adding a namespace to the gossipsub protocol id using [.protocol_id_prefix()](https://github.com/libp2p/rust-libp2p/blob/v0.56.0/protocols/gossipsub/src/config.rs#L574) work?

MavenRain: I dig it . . . done

sstanfield (approved):

grantkee (approved):
LGTM!

PR #773, opened 2026-06-22
Add server side calls to retrieve consensus bytes for an entire ConsensusOutput.

sstanfield:
Expose the server side of pulling the pack formatted bytes for a ConsensusOutput.  This is a step in improving syncing within the current epoch.  Should also reduce server load, can just read and send bytes- no real processing on the server.

### DIFF
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -895,7 +895,8 @@ where
         match self.consensus_chain.consensus_output_bytes_by_number(number).await {
             Ok(Some(bytes)) => Ok(bytes),
             // Missing locally, or the requested number is outside the pack's range.
-            _ => Err(PrimaryNetworkError::UnknownConsensusOutput(number)),
+            Ok(None) => Err(PrimaryNetworkError::UnknownConsensusOutput(number)),
+            Err(e) => Err(PrimaryNetworkError::ConsensusChainError(e)),
         }
     }
 
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -6,7 +6,7 @@ use super::{
 };
 use crate::{
     error::{CertManagerError, PrimaryNetworkError, PrimaryNetworkResult},
-    network::{message::PrimaryGossip, PendingEpochStream},
+    network::{message::PrimaryGossip, PendingStreamRequest, StreamRequestKind},
     state_sync::{CertificateCollector, StateSynchronizer},
     ConsensusBusApp, NodeMode,
 };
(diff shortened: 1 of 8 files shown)


grantkee (changes requested):
The req/res protocol codec caps message sizes at 1MB which is also the max batch size. Consensus output contains many batches, so we have to use the stream protocol for this

crates/consensus/primary/src/network/mod.rs
```diff
@@ -71,12 +70,49 @@ pub const MAX_CONCURRENT_EPOCH_STREAMS: usize = 5;
...
+    /// Stream the full pack file for an epoch.
+    EpochPack(Epoch),
+    /// Stream the raw bytes for a single consensus output (by consensus chain number).
+    ConsensusOutput(u64),
+}
+
+/// Correlation digest for an epoch-pack stream request.
+fn epoch_stream_digest(epoch: Epoch) -> B256 {
+    let mut hasher = tn_types::DefaultHashFunction::new();
+    hasher.update(&epoch.to_le_bytes());
```
grantkee, line 96: Would this benefit from a prefix `b"epoch-stream"` for the sake of consistency?

crates/consensus/primary/src/network/handler.rs
```diff
@@ -978,11 +1010,53 @@ where
...
+        peer: BlsPublicKey,
+    ) -> PrimaryNetworkResult<()>
+    where
+        S: AsyncWrite + Unpin + Send,
+    {
+        match self.consensus_output_bytes(number).await {
+            Ok(bytes) => {
+                // write in chunks so a single timeout bounds a slow reader
+                for chunk in bytes.chunks(16 * 1024) {
+                    timeout(buffer_timeout, stream.write_all(chunk)).await??;
```
MavenRain, line 1032: Each chunk is time-bound, but the aggregation isn't.  I am assuming that these chunks must be written in-order, and so I will not recommend `FuturesUnordered`, but I will recommend an additional overarching timeout.  I feel like there also may be an async version of `for_each` out of `StreamExt` that might make this more ergonomic.

crates/consensus/primary/src/network/handler.rs
```diff
@@ -867,6 +867,38 @@ where
...
+                    // Don't wait more than 2 seconds for this to show up.
+                    return Err(PrimaryNetworkError::UnknownConsensusOutput(number));
+                }
+                count += 1;
+            }
+        }
+        match self.consensus_chain.consensus_output_bytes_by_number(number).await {
+            Ok(Some(bytes)) => Ok(bytes),
+            // Missing locally, or the requested number is outside the pack's range.
+            _ => Err(PrimaryNetworkError::UnknownConsensusOutput(number)),
```
MavenRain: Is it safe to react the same way to `Ok(None)` as `Err(_)`?

MavenRain (commented):
Just a couple of questions . . . will approve once the outstanding changes-requested flag is cleared

MavenRain (dismissed):

grantkee (dismissed):
LGTM!

MavenRain (dismissed):

grantkee (approved):

PR #775, opened 2026-06-22
Add `keytool generate pop` for execution-address rotation

### LINKED ISSUE #774: Allow rotating a node's execution address without regenerating keys
## Problem

A node's proof of possession (PoP) is a BLS signature that commits to the execution address. The on-chain `stake()` call verifies the PoP against the staking address, so once an operator wants to stake or earn rewards under a *different* execution address, the original signature no longer validates and the call reverts with "proof of possession is incorrect."

Today the only way to produce a PoP for a new address is to run `keytool generate validator|observer`, which mints a fresh BLS keypair and overwrites the node's identity. That is destructive: it changes the BLS public key, network keys, p2p peer IDs, and node name. An operator who simply wants to point an existing node at a new payout address has no safe path. They either lose their identity or hand-edit `node-info.yaml`, which can't re-sign the PoP.

The result is that execution-address rotation, a potential operational need before staking, forces operators into a workflow that risks destroying validator identity.

## Solution

Introduce a `keytool generate pop` subcommand (alias `proof-of-possession`) that re-signs the PoP for a new execution address using the node's *existing* keys. It should load the existing BLS keys and `node-info.yaml`, re-sign against `--address`, and rewrite only `proof_of_possession` and `execution_address`. The BLS public key, network keys, p2p peer IDs, and the deterministically derived node name should remain byte-for-byte identical.

The command should never generate or overwrite keys, and should error with a clear hint (pointing at `generate validator|observer`) when keys or `node-info.yaml` are missing — rather than panicking or silently minting new keys. It should mirror the ergonomics of the existing generate subcommands: the `--execution-address` alias and the `EXECUTION_ADDRESS` env var, and a logged warning when the address actually changes.

To keep the two paths consistent, the PoP-signing logic should be factored into a shared helper used by both `generate validator|observer` (fresh keys) and `generate pop` (existing keys), so both produce identical PoP-derived fields.

grantkee:
- New `generate pop` subcommand (alias `proof-of-possession`) re-signs the proof of possession for a new `--address` using the node's existing BLS keys — only `execution_address` and `proof_of_possession` in `node-info.yaml` change; BLS key, network keys, p2p peer IDs, and node name are untouched.
- Errors with a hint to `generate validator|observer` when keys or `node-info.yaml` are missing; never mints or overwrites keys.
- Extracted the PoP-signing logic into a shared `set_proof_of_possession` helper so `generate validator|observer` and `generate pop` produce identical PoP-derived fields.
- Tests cover the happy path (identity unchanged, new PoP verifies for the new address and fails for the old), the missing-keys error, and clap wiring of both the subcommand and its alias.
- README: documents the rotation workflow and re-exporting staking args afterward.
- Makefile: `docker-builder` now reuses an existing `tn-builder` buildx instance instead of failing when it already exists (error while creating a docker image to deploy this update)

closes #774


crates/telcoin-network-cli/src/keytool/generate.rs
```diff
@@ -1,11 +1,28 @@
...
+/// `bls_public_key` / `proof_of_possession` / `execution_address` / `name` values.
+pub(crate) fn set_proof_of_possession(
+    node_info: &mut NodeInfo,
+    key_config: &KeyConfig,
+    address: Address,
+) -> eyre::Result<()> {
+    node_info.bls_public_key = key_config.primary_public_key();
+    node_info.proof_of_possession = key_config.generate_proof_of_possession_bls(&address)?;
+    node_info.execution_address = address;
+    node_info.name =
```
sstanfield: Can't the node name have been set to a non-default and won't this overwrite it with a default if so?  It looks like one call site might have an issue.  Maybe make name Optional and only default if None?

sstanfield (changes requested):
Maybe my comment is off base but I think it needs to be addressed.

crates/telcoin-network-cli/src/keytool/pop.rs
```diff
@@ -0,0 +1,94 @@
...
+        // re-sign against the existing keys (shared with `generate validator|observer`)
+        set_proof_of_possession(&mut node_info, &key_config, self.address)?;
+        Config::write_to_path(dir.node_info_path(), &node_info, ConfigFmt::YAML)?;
+
+        // PoP is printed in the same bs58 encoding `node-info.yaml` uses (BlsSignature Display).
+        println!("OK  node-info.yaml updated: {}", dir.node_info_path().display());
+        println!("    execution_address:   {}", node_info.execution_address);
+        println!("    proof_of_possession: {}", node_info.proof_of_possession);
+        println!(
+            "Next: telcoin keytool export-staking-args --node-info {}",
```
MavenRain: I think this should be `"Next: telcoin-network keytool export-staking-args --node-info {}"`

crates/telcoin-network-cli/src/keytool/pop.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub fn execute<TND: TelcoinDirs>(
+        &self,
+        dir: &TND,
+        passphrase: Option<String>,
+    ) -> eyre::Result<()> {
+        // existing keys - never generated here; error with a hint if absent
+        let key_config = KeyConfig::read_config(dir, passphrase).map_err(|e| {
+            eyre::eyre!(
+                "could not load existing BLS keys from {}: {e}\n\
+                 hint: generate keys first with `keytool generate validator|observer`",
```
MavenRain: As written, I believe that this hint is surfaced if the key exists, but the passphrase is incorrect.

crates/telcoin-network-cli/src/keytool/generate.rs
```diff
@@ -1,11 +1,28 @@
...
+/// Sign the proof of possession for `address` from `key_config` and write the
+/// PoP-derived fields into `node_info`. Shared by `generate validator|observer`
+/// (fresh keys) and `generate pop` (existing keys), so both produce identical
+/// `bls_public_key` / `proof_of_possession` / `execution_address` / `name` values.
+pub(crate) fn set_proof_of_possession(
+    node_info: &mut NodeInfo,
+    key_config: &KeyConfig,
+    address: Address,
+) -> eyre::Result<()> {
+    node_info.bls_public_key = key_config.primary_public_key();
```
MavenRain: Consider checking to see if the key matches what is currently loaded into `node_info`

crates/telcoin-network-cli/src/keytool/pop.rs
grantkee: Good catch. I'm thinking about renaming the binary, but it is currently `telcoin-network`. Thanks for flagging!

crates/telcoin-network-cli/src/keytool/pop.rs
grantkee: Good idea. Since this is public facing, I expanded the hints based on actual error type

crates/telcoin-network-cli/src/keytool/generate.rs
grantkee: This is a reasonable sanity check - added

crates/telcoin-network-cli/src/keytool/generate.rs
grantkee: Added a `--name` flag with a default fallback if empty. It's not overwritten anymore. Good idea

MavenRain (approved):

sstanfield (approved):
LGTM!

PR #778, opened 2026-06-22
fix(worker): retry transient RPC errors instead of treating them as  rejections (#747)

### LINKED ISSUE #747: network-libp2p: `NetworkError::RPCError(String)` conflates an explicit peer rejection with a transient remote-side failure, so the worker `QuorumWaiter` permanently gives up on a peer that hit a one-off internal error
## Summary

`NetworkError::RPCError(String)` is the single application-layer RPC error variant, and the worker `QuorumWaiter` uses it as the discriminator between "peer rejected this, do not retry" and "transient network error, retry with backoff."  The problem is that transient, recoverable conditions on the responder side (a batch-store write failure, an internal channel failure during an epoch transition) are mapped into the same `RPCError` string.  When that reaches the requester, `QuorumWaiter` classifies it as a permanent rejection and stops retrying that peer, even though the failure was a momentary local condition that a retry would clear.

### The conflated variant (`crates/network-libp2p/src/error.rs:76-78`)

```rust
/// Request/response RPC Error
#[error("{0}")]
RPCError(String),
```

`Display` is just the inner string, so no structured cause survives.

### Transient responder-side conditions become `Internal` (`crates/consensus/worker/src/network/handler.rs`)

```rust
// :167  batch-store write failure
WorkerNetworkError::Internal(format!("failed to write to batch store: {e}"))
// :174  local-network channel failure
.map_err(|e| WorkerNetworkError::Internal(e.to_string()))?;
// :103  internal send failure
WorkerNetworkError::Internal(format!( ... ))
```

### `Internal` is shipped as an `RPCError` string to the requester (`crates/consensus/worker/src/network/handle.rs:96`)

```rust
WorkerResponse::Error(WorkerRPCError(s)) => Err(NetworkError::RPCError(s)),
```

(`WorkerNetworkError::Internal` is rendered into the `WorkerRPCError(String)` response payload, then mapped back to `NetworkError::RPCError` on receipt.)

### The requester treats `RPCError` as a permanent rejection (`crates/consensus/worker/src/quorum_waiter.rs:65-97`)

The doc comment states the intended contract:

```rust
/// - `Err(Rejected)` — peer explicitly rejected (RPCError), no retry
/// - `Err(Network)` — all retry attempts exhausted
```

and the code implements it:

```rust
Err(NetworkError::RPCError(msg)) => {
    // ...
    return Err(WaiterError::Rejected(deliver)); // no retry
}
// every other NetworkError variant: retry up to MAX_RETRIES with exponential backoff
```

So a responder that hits a momentary batch-store or channel error is treated identically to a responder that deliberately rejected the batch.

## Impact

Reliability.  The retry behavior is asymmetric and, in the transient case, wrong: a transport flap gets several retries with backoff, but a peer that hit a one-off local condition (a batch-store write error, channel pressure during an epoch transition) is permanently classified as a rejecter and dropped from the quorum attempt with no retry.  Under epoch-transition load, where transient internal errors are most likely, this can shrink the effective set of peers a worker will collect acknowledgements from, for failures that a single retry would have cleared.

## Suggested fix

Distinguish "explicit rejection" from "transient remote error" at the type level rather than by string.  Options:

1. Split `RPCError(String)` into a rejection variant and a retryable remote-error variant, and have the responder map `WorkerNetworkError::Internal` (and similar transient cases) to the retryable one.
2. Keep one variant but carry a structured, non-stringly cause (an enum discriminant) that `QuorumWaiter` can match on to decide retry versus give-up.

Either way, the responder's transient `Internal` conditions should land on the retry path, not the no-retry `Rejected` path.

## Scope and novelty

Confirmed end to end on `main` (`0bd36414`).  This spans `crates/network-libp2p` (the error type) and `crates/consensus/worker` (the producer and the consumer), but the root cause is the under-specified error type in the network crate.  No existing issue or PR tracks the rejection-versus-transient conflation; #250 ("Inbound Failures: Report Penalty from Codec") is about a different error path.  Not covered by the `network-libp2p` security review (PR #620, NL-001..NL-010).

MavenRain:
## Summary

  `NetworkError::RPCError(String)` was the single application-layer RPC error
  variant, and the worker `QuorumWaiter` used it as the discriminator between
  "peer rejected this, do not retry" and "transient remote error, retry with
  backoff." Transient, recoverable conditions on the responder side (a
  batch-store write failure, an internal channel failure during an epoch
  transition) were mapped into that same `RPCError` string, so the requester
  classified them as a permanent rejection and stopped retrying that peer,
  even though a single retry would likely have cleared the failure.

  This PR distinguishes "explicit rejection" from "transient remote error" at
  the type level, so the responder's transient conditions land on the retry
  path instead of the no-retry rejection path.

  Closes #747.

  ## Root cause

  The flow that conflated the two cases:

  1. A transient responder-side failure in `process_report_batch` became
  `WorkerNetworkError::Internal(...)`
  (`crates/consensus/worker/src/network/handler.rs`).
  2. The responder stringified any error into
  `WorkerResponse::Error(WorkerRPCError(String))`
  (`crates/consensus/worker/src/network/mod.rs`).
  3. The requester mapped that back to `NetworkError::RPCError(String)`
  (`crates/consensus/worker/src/network/handle.rs`).
  4. `QuorumWaiter::waiter` treated `RPCError` as a permanent rejection
  (`WaiterError::Rejected`, no retry), while every other error variant was
  retried with backoff (`crates/consensus/worker/src/quorum_waiter.rs`).

  So a responder that hit a momentary batch-store or channel error was treated
  identically to one that deliberately rejected the batch.

  ## Fix

  Carry the rejection-versus-transient distinction through the type system,
  mirroring the pattern the primary network already uses
  (`PrimaryResponse::RecoverableError` + `PrimaryResponse::into_error_ref`):

  - **`network-libp2p`**: add `NetworkError::RPCRetryable(String)` as the
  retryable counterpart of `RPCError`. `RPCError` keeps its meaning (permanent
  rejection, no retry).
  - **worker `message.rs`**: add
  `WorkerResponse::RecoverableError(WorkerRPCError)` and
  `WorkerResponse::into_error_ref(&WorkerNetworkError)`, which classifies each
  `WorkerNetworkError` as either recoverable or a permanent rejection
  (exhaustive match, so new error variants must be classified deliberately).
  - **worker `mod.rs`**: the responder builds its error response through
  `into_error_ref(&err)`, so transient conditions are sent as
  `RecoverableError`.
  - **worker `handle.rs`**: the requester maps `RecoverableError` to
  `NetworkError::RPCRetryable` (and `Error` to `RPCError`, unchanged).
  - **worker `quorum_waiter.rs`**: no behavior code change is needed.
  `RPCRetryable` is not `RPCError`, so it falls into the existing
  retry-with-backoff arm and only resolves to `WaiterError::Network` after
  retries are exhausted, never to `WaiterError::Rejected`. The doc comment is
  updated to make the contract explicit.

  This keeps the retry policy where it belongs (the `QuorumWaiter`, which owns
  the backoff loop and metrics) and only adds the type-level signal it was
  missing.

  ### Transient vs permanent classification

  `into_error_ref` routes these to `RecoverableError` (requester retries):

  `Internal`, `Timeout`, `Network`, `StreamClosed`, `DBInsert`, `DBCommit`,
  `DBRead`, `StdIo`.

  Everything else stays a permanent `Error` (requester does not retry): `Bcs`,
  `BatchValidation`, `NonCommitteeBatch`, `InvalidRequest`, `InvalidTopic`,
  `TooManyBatches`, `UnexpectedBatch`, `DuplicateBatch`,
  `UnknownStreamRequest`, `RequestHashMismatch`, `BatchEpochMismatch`. These
  are decisions the peer would make the same way on an identical retry
  (invalid batch or request, protocol violation, or a committee/epoch view
  mismatch for this request).

  ## Compatibility note

  `WorkerResponse::RecoverableError` is appended to the end of the enum, so
  existing BCS variant indices are preserved. A pre-fix node still cannot
  decode the new variant, so this wants a coordinated worker upgrade. This is
  the same property the primary network's `RecoverableError` already carries
  on `main`, and it is consistent with the project's existing breaking-wire
  changes. No version-negotiation machinery is added, for parity with how the
  primary handles it.

  ## Testing

  - New unit tests for the classifier: a transient `Internal` error is
  reported as recoverable, and a `NonCommitteeBatch` rejection stays
  permanent.
  - New `QuorumWaiter` regression tests: a `RecoverableError` on the first
  attempt is retried and reaches quorum, and a peer that keeps returning
  `RecoverableError` exhausts its retries and is counted as `AntiQuorum`,
  never `QuorumRejected`.
  - Full suites pass: `tn-network-libp2p` (171), the worker classifier unit
  tests, and the `quorum_waiter` integration tests (11). `cargo fmt` and
  `clippy` are clean for the touched crates.

  ## Scope

  Touches `crates/network-libp2p` (the error type) and
  `crates/consensus/worker` (the producer and the consumer). The root cause
  was the under-specified error type in the network crate, now split into a
  permanent and a retryable variant.

### DIFF
--- a/crates/consensus/worker/src/network/handle.rs
+++ b/crates/consensus/worker/src/network/handle.rs
@@ -94,6 +94,9 @@ impl WorkerNetworkHandle {
                 "Got wrong response, not a report batch is peer exchange!".to_string(),
             )),
             WorkerResponse::Error(WorkerRPCError(s)) => Err(NetworkError::RPCError(s)),
+            WorkerResponse::RecoverableError(WorkerRPCError(s)) => {
+                Err(NetworkError::RPCRetryable(s))
+            }
         }
     }
 
@@ -274,6 +277,9 @@ impl WorkerNetworkHandle {
                 "Got wrong response: peer exchange instead of stream ack".to_string(),
             )),
             WorkerResponse::Error(WorkerRPCError(s)) => Err(NetworkError::RPCError(s)),
+            WorkerResponse::RecoverableError(WorkerRPCError(s)) => {
+                Err(NetworkError::RPCRetryable(s))
+            }
         }
     }
 
--- a/crates/consensus/worker/src/network/message.rs
+++ b/crates/consensus/worker/src/network/message.rs
@@ -2,6 +2,7 @@
 
 use std::collections::BTreeSet;
 
+use crate::network::error::WorkerNetworkError;
 use serde::{Deserialize, Serialize};
 use tn_network_libp2p::{PeerExchangeMap, TNMessage};
 use tn_types::{BlockHash, Epoch, SealedBatch};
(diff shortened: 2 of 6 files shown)


sstanfield (approved):
LGTM!

PR #779, opened 2026-06-22
Make sure to only count a consensus sig once per validator.

### LINKED ISSUE #713: Consensus gossip can multiple count the same validator.
In primary/src/network/handler.rs when we handle PrimaryGossip::Consensus it is possible to count the same validators signature more than once.  Need to check so we have one validator on vote.

sstanfield:
Closes https://github.com/Telcoin-Association/telcoin-network/issues/713

Also added a Digest type for ConsensusResult which resulted in ConsensusResult moving to the tn-types crate.

### DIFF
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -269,15 +269,15 @@ where
                         // possible and don't want to do the cert verify if we can
                         // skip it.  So this seems like the place.  And of course
                         // close any windows to double counting a signature.
-                        let mut guard = self.consensus_certs.lock();
+                        let guard = self.consensus_certs.lock();
                         if guard.get(&consensus_result_hash).and_then(|set| set.get(&key)).is_some()
                         {
                             // We have already counted this signature so ignore.
                             return Ok(());
                         }
-                        // An argument could be made to do this outside the lock.  This would mean
-                        // it happens on duplicates but holds the lock for
-                        // less time.
+                        drop(guard);
+                        // Drop the lock for the expensive verify op. If a dup gets through the Set
+                        // will still only count it once.
                         ensure!(
                             signature
                                 .verify_secure(&to_intent_message(consensus_result_hash), &key),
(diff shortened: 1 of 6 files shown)


grantkee (dismissed):
LGTM! Merge conflict, so ping me when it's resolved

crates/consensus/primary/src/network/handler.rs
```diff
@@ -265,37 +262,60 @@ where
...
+                        let mut guard = self.consensus_certs.lock();
+                        if guard.len() > 20 {
+                            // Small sanity check to make it more difficult for a malicious
+                            // committee member to fill up
+                            // consensus_certs.  Attempt to evict any records that don't appear
+                            // relavent when more that 20 entries
+                            // (should not have this many under normal conditions).
+                            // We would expect a malicious validator to flood a lot of singleton
+                            // results.
+                            guard.retain(|k, s| *k == consensus_result_hash || s.len() > 1);
```
MavenRain, line 298: `s.len() > 1` implies that any two members can collude to stuff entries without bound that will never clear for committees larger than five.

Consider, alternatively, a hard cap on entries or ejecting stale entries (entries with `number <= published_number`).

sstanfield: Not sure what published_number is?  But yes this will not handle multiple malicious committee members, that is intentional to keep it simple.  It is just a sanity check in case one goes rouge.  Since we have a closed vetted validator set the odds of one being rogue is extremely low and more that one colluding to DOS validator peers seems even less likely.  Plus in that case (multiple byzantine validators) we have bigger issues.  This is also cleared when quorum is hit so the odds of causing an OOM before that clear is probably low (but not impossible).  I may be missing some of your point though?  Finally increasing the threshold will make it far more likely to throw out in process accounting- they may open another DOS window even easier to trigger than an OOM...

MavenRain: My understanding is that quorum is never hit for exactly the bogus entries that you'd want to clear.  That said, your point about the permissioned validator set is well taken.

MavenRain (approved):
Prior concern addressed

PR #780, opened 2026-06-23
Do not penalize UnsupportedProtocols; guard against self-dial/self-penalty (#777)

### LINKED ISSUE #777: Over-eager peer banning on protocol-negotiation failure: downgrade `UnsupportedProtocols` penalty + guard against self-dial
## Summary

  A live `adiri` observer node (build `telcoin-network 0.1.0 (977e44a5)`) was
  observed banning a cluster of peers **including its own worker identity**
  off the back of libp2p req/res `UnsupportedProtocols` failures. Two distinct
  defects combine here:

  - **A.** `OutboundFailure::UnsupportedProtocols` /
  `InboundFailure::UnsupportedProtocols` are scored `Penalty::Severe`. A failure
  to negotiate a common protocol is *version/role skew*, not peer misbehavior,
  so it should not be penalized at all.
  - **B.** The node dials and then penalizes **its own peer id** (worker swarm).
  There is no self-dial / self-penalty guard, so a self-connection that fails
  protocol negotiation feeds straight into A and bans the node's own worker.

  Net effect: honest, simply-not-matching peers get banned; banned peers then
  have their Kademlia records rejected, which degrades `NodeRecord` propagation
  (PeerId→BLS resolution) and req/res availability for sync. This is an
  amplifier of peering/sync failure, and it is a prerequisite blocker for
  landing the chain-id protocol split (#765, see below).

  ## Evidence (observer log, 2026-06-22, build `977e44a5`)

  - The set of peers ever `peer banned` (12) is **identical** to the set that
  hit `outbound failure: unsupported protocol` (12); no peer was banned without
  one. 42 ban events, 96 unsupported-protocol events.
  - The node's **own worker id** (`12D3KooWMJDGCTrus…`, confirmed via `Worker 0
  successfully booted on …/p2p/12D3KooWMJDGCTrus…` and `local_peer_id=…`)
  accrued **9** unsupported-protocol failures and **4** bans, all on
  `target=network`.
  - Self-dial captured: `new connection established endpoint=Dialer { address:
  /ip4/192.168.8.1/udp/49594/quic-v1/p2p/12D3KooWMJDGCTrus…, role_override:
  Dialer, port_use: Reuse }` — the node dialing its own worker peerid on the
  worker's own port. The node's **primary** id was never banned (asymmetry that
  rules out a primary↔worker protocol mismatch; this is self-connection, not
  cross-role dialing).
  - Knock-on: banned peers' DHT records are rejected (114× `rejecting put
  request for record`, gated by ban status before record validation), and
  `banned_count` spikes in waves (8 at the 10:16 restart, 5 at the 15:29
  restart).

  > Caveat: this capture is `info`/`warn`/`error` only, so the exact penalty
  that tips a given peer past the ban threshold (`min_score_before_ban`) is not
  directly visible — `Penalty::Severe` (−10) over the observed 2-failure windows
  does not on its own reach −50. The case for the fix does **not** rest on the
  precise arithmetic; it rests on the principle that protocol-negotiation
  failure is not a scoreable offense (see the rolling-upgrade contract below).

  ---

  ## Part A: Do not penalize `UnsupportedProtocols`

  ### Root cause

  `crates/network-libp2p/src/consensus.rs`:

  ```rust
  // outbound, ~1065
  ReqResOutboundFailure::UnsupportedProtocols => {
      warn!(target: "network", ?peer, ?request_id, "outbound failure:
  unsupported protocol");
      self.swarm.behaviour_mut().peer_manager.process_penalty(peer,
  Penalty::Severe);
  }
  // inbound, ~1108
  ReqResInboundFailure::UnsupportedProtocols => {
      warn!(target: "network", ?peer, ?request_id, ?error, "inbound failure:
  unsupported protocol");
      self.swarm.behaviour_mut().peer_manager.process_penalty(peer,
  Penalty::Severe);
  }
  ```

  The stream-path taxonomy mirrors this
  (`crates/network-libp2p/src/stream/upgrade.rs:132`: *"… unsupported protocols
  are severe."*).

  ### Why this is wrong

  `UnsupportedProtocols` means multistream-select found no common protocol . . . the
  peer runs a different/older/role-distinct protocol set. That is the textbook
  *honest-skew* condition, and the codebase already commits to not penalizing it
  elsewhere: `crates/network-libp2p/src/tests/network_tests.rs`
  (`test_pre_upgrade_record_accepted_with_default_rpc`) documents the
  rolling-upgrade contract . . . *"The honest not-yet-upgraded sender is not
  penalized."* The transport layer should honor the same contract. Note that
  sibling transport faults `DialFailure` and `ConnectionClosed` are already
  treated as not-the-peer's-fault in the same match.

  This also blocks #765: `765-chain-id-protocol` deliberately makes old/new
  nodes fail to peer ("all nodes must upgrade together; old and new will not
  peer"). If `UnsupportedProtocols => Severe` stays, the #765 rollout will have
  every node *ban* every not-yet-upgraded peer (and DHT-poison them), turning a
  clean non-peering into network partitioning. Fixing the penalty is a
  precondition for landing #765 safely.

  ### Proposed change

  - Map both `OutboundFailure::UnsupportedProtocols` and
  `InboundFailure::UnsupportedProtocols` to **no penalty** (keep the `warn!`,
  drop the `process_penalty`, or route through an explicit `None`), grouping
  them with `DialFailure`/`ConnectionClosed`.
  - Update the taxonomy comment at `stream/upgrade.rs:132` and the table in
  `crates/network-libp2p/src/peers/penalty.md` (rows ~53/56) and the
  `StreamFailure::UnsupportedProtocol` arm to match.
  - Update `penalty_mapping_mirrors_reqres` to assert
  `…UnsupportedProtocol.penalty().is_none()`.

  ### Acceptance criteria

  - A peer whose only failures are `UnsupportedProtocols` (in or outbound) is
  never disconnected or banned.
  - Taxonomy doc + `penalty.md` + the penalty-mapping test reflect "unsupported
  protocol = not penalized".

  ---

  ## Part B: Guard against self-dial / self-penalty

  ### Root cause

  `PeerManager` does not know its own peer id, and `dial_peer`
  (`crates/network-libp2p/src/peers/manager.rs:132`) guards only
  banned/dialing/connected: no self-check. Per the note on
  `dial_attempt_already_registered`, *"Self and kad behaviors can initiate dial
  attempts,"* so the self-dial most likely originates from Kademlia auto-dialing
  the node's own published record (`"Providing our record to kademlia for peer
  <local_peer_id>"`, `consensus.rs:444`), bypassing `dial_peer` and its guards.
  libp2p's built-in self-dial rejection is keyed on peer id, so a learned
  hairpin address (`192.168.8.1`) can slip through. The result is a
  self-connection that fails req/res negotiation and is then penalized as if it
  were a remote peer.

  ### Proposed change

  Defense in depth, both at the source and the sink:

  1. **Never penalize a local id.** In the req/res failure handlers (and ideally
  centrally), skip `process_penalty` when `peer == *self.swarm.local_peer_id()`.
  Cover all local identities (primary and every worker swarm), since each runs its own peer id.
  2. **Never dial/learn ourselves.** Filter the node's own peer id(s) out of the
  discovery/dial inputs (Kademlia-provided addresses, peer-exchange entries,
  committee dial list) so a self-connection is not initiated in the first place;
  drop an inbound/loopback connection whose remote id is one of our own without
  scoring it.

  (Implementation note: threading the local peer id into `PeerManager`, or
  adding the guard in the `consensus.rs` handlers where
  `self.swarm.local_peer_id()` is already in scope, both work; the latter is the
  smaller change.)

  ### Acceptance criteria

  - The node's own peer id(s) are never dialed and never scored/banned.
  - Regression test: routing the node's own peer id through the discovery/dial
  path is a no-op; a self-connection, if it occurs, is dropped without a
  penalty.

  ---

  ## Related

  - #743: added the gossip `relaying peer's BLS identity is not yet resolved`
  warning that surfaced the downstream resolution damage.
  - #747 (`split-rpc-error-rejection-from-transient`): same philosophy
  (transient/non-fault failures should not be permanent rejections) applied to
  the **application** RPC layer; this issue is the **transport** analog
  (`UnsupportedProtocols`), a separate code path that #747 does not touch.
  - #765 (`chain-id-protocol`): will intentionally introduce
  `UnsupportedProtocols` between un-upgraded and upgraded nodes; Part A should
  land first.
  - #760: operator-configurable scoring thresholds; tighter thresholds make the
  over-eager ban fire sooner.

  ## Notes / open questions

  - The exact ban-tipping penalty per peer is not observable in the available
  `info`-level capture. A `debug`/`trace` run on `network` + `peer-manager`
  would confirm the per-ban penalty sequence and the offered-vs-supported
  protocol lists. Not required to fix A/B, but useful to confirm the remote
  peers' skew is genuine version mismatch.

MavenRain:
## Summary

Fixes #777. Stops a live `adiri` observer node from banning honest peers, including its own worker identity, off the back of libp2p req/res `UnsupportedProtocols` failures.  Two independent defects, fixed together:

- **Part A** — `Outbound/InboundFailure::UnsupportedProtocols` was scored `Penalty::Severe`.  Failing to negotiate a common protocol is honest version/role skew, not misbehavior, so it is now **not penalized** (warn-only), grouped with the existing no-penalty `DialFailure`/`ConnectionClosed` arms.
- **Part B** — the node dialed and then penalized **its own peer id** (worker swarm) via a learned hairpin address that libp2p's peer-id self-dial check missed.  There was no self-dial / self-penalty guard anywhere in the crate. Added one, in depth.

This is a prerequisite for landing #765 (`chain-id-protocol`): that change intentionally makes old/new nodes fail to peer with `UnsupportedProtocols`. With the old `Severe` mapping, the #765 rollout would have every node *ban* (and DHT-poison) every not-yet-upgraded peer, turning a clean non-peering into a network partition.

## Part A — do not penalize `UnsupportedProtocols`

`multistream-select` finding no common protocol means the peer runs a different/older/role-distinct protocol set . . . the textbook honest-skew condition, and the same not-the-peer's-fault class as `DialFailure`/`ConnectionClosed`. The codebase already commits to this contract elsewhere (`test_pre_upgrade_record_accepted_with_default_rpc`).

- `consensus.rs` outbound + inbound `UnsupportedProtocols` arms: drop `process_penalty(peer, Penalty::Severe)`, keep the `warn!` (now labelled `(not penalized)`) for operator visibility.
- `stream/upgrade.rs`: `StreamFailure::UnsupportedProtocol.penalty()` → `None`, keeping the stream taxonomy a mirror of the req/res one; `penalty_mapping_mirrors_reqres` updated to assert `.is_none()`.
- `peers/penalty.md`: rows for both `UnsupportedProtocols` entries → `None`, plus a restraint-invariant note.

## Part B — guard against self-dial / self-penalty

The committee dial loop already self-filters by BLS key at the app layer (`start_epoch.rs`), so the self-dial enters purely via the kad/discovery path, which is keyed on `PeerId` and had no self-check.  `PeerManager` did not even hold its own peer id.

Keystone: thread the swarm's own `peer_id` (already computed in `ConsensusNetwork::new`) into `PeerManager` and add an `is_local_peer` predicate.  Each swarm (primary and each worker) builds its own `PeerManager` with its own keypair-derived id, so a per-swarm single-id guard naturally covers every one of the node's identities . . . no multi-id allowlist needed.  Defense in depth, source and sink:

- **Penalty sink (central):** `PeerManager::process_penalty` short-circuits before the score model (and before the metric) for the local id . . . covers every penalty call site (gossip / kad / req-res) in one place.
- **Dial source:** `dial_peer` skips the local id and reports success, so retrying callers (`dial_peer_bls`) treat it as a no-op rather than a failure to back off on.
- **Discovery source:** `eligible_for_discovery` excludes the local id, so a self entry learned via kad closest-peers or peer exchange never enters the discovery feed.
- **Connection sink:** `handle_pending_outbound_connection` denies a dial to the local id (the precise fix for kad auto-dialing our own re-learned record); `handle_established_{inbound,outbound}_connection` deny a self-connection that still reaches the established stage (the inbound peer id is only known there). None of these call `process_penalty`, so the self-connection is dropped without scoring.

## Tests

- `stream/upgrade.rs::penalty_mapping_mirrors_reqres` — asserts `UnsupportedProtocol` is no longer penalized.
- `network_tests.rs::test_unsupported_protocol_does_not_penalize` — cross-role primary↔worker req/res produces real `UnsupportedProtocols` outbound failures; 6 of them (pre-fix 6 × Severe = −60, past the −50 ban threshold) leave the peer connected with an unchanged score.
- `tests/peer_manager.rs` — `test_self_id_filtered_from_discovery`, `test_self_dial_request_is_noop`, `test_self_penalty_is_noop`, `test_self_connection_denied`.

## CI

Run against the repo's pinned nightly (`nightly-2026-03-20`) and stable test toolchain:

- [x] `cargo +nightly-2026-03-20 fmt --all -- --check` — clean
- [x] `cargo +nightly-2026-03-20 clippy -p tn-network-libp2p --all-features --all-targets --locked -- -D warnings` — zero warnings
- [x] `cargo test -p tn-network-libp2p --lib` — 176 passed, 0 failed

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -89,13 +89,14 @@ where
 {
     /// Create a new instance of Self.
     pub(crate) fn new(
+        local_peer_id: PeerId,
         gossipsub: gossipsub::Behaviour,
         req_res: request_response::Behaviour<C>,
         kademlia: kad::Behaviour<KadStore<DB>>,
         peer_config: &PeerConfig,
         metrics: PeerManagerMetrics,
     ) -> Self {
-        let peer_manager = PeerManager::new(peer_config, metrics);
+        let peer_manager = PeerManager::new(local_peer_id, peer_config, metrics);
         let stream = StreamBehavior::new();
         Self { peer_manager, gossipsub, req_res, kademlia, stream }
     }
@@ -317,6 +318,7 @@ where
 
         // create custom behavior
         let mut behavior = TNBehavior::new(
+            peer_id,
             gossipsub,
             req_res,
             kademlia,
@@ -1058,16 +1060,15 @@ where
                             .peer_manager
                             .process_penalty(peer, Penalty::Mild);
                     }
-                    // Severe = 5 strikes covers the typical rolling-upgrade window where
-                    // a peer sees <=5 req-res handshakes per halflife. Demote to Medium
-                    // if telemetry shows version-skewed peers banned during normal
-                    // upgrades.
+                    // Not penalized. Failing to negotiate a common protocol is honest
+                    // version/role skew (the peer runs a different/older/role-distinct
+                    // protocol set), not misbehavior — the same not-the-peer's-fault
+                    // class as `DialFailure`/`ConnectionClosed` above. Penalizing it
+                    // bans not-yet-upgraded peers during rolling upgrades and would turn
+                    // the #765 chain-id protocol split into a network partition. Warn for
+                    // operator visibility only.
                     ReqResOutboundFailure::UnsupportedProtocols => {
-                        warn!(target: "network", ?peer, ?request_id, "outbound failure: unsupported protocol");
-                        self.swarm
-                            .behaviour_mut()
-                            .peer_manager
-                            .process_penalty(peer, Penalty::Severe);
+                        warn!(target: "network", ?peer, ?request_id, "outbound failure: unsupported protocol (not penalized)");
                     }
                 }
 
@@ -1101,20 +1102,14 @@ where
                                 .process_penalty(peer, Penalty::Medium);
                         }
                     },
-                    // Severe = 5 strikes covers the typical rolling-upgrade window where
-                    // a peer sees <=5 req-res handshakes per halflife. Demote to Medium
-                    // if telemetry shows version-skewed peers banned during normal
-                    // upgrades.
+                    // Not penalized. The local peer supports none of the protocols the
+                    // remote requested: honest version/role skew, not misbehavior (the
+                    // inbound mirror of the outbound arm above). Penalizing it bans
+                    // not-yet-upgraded peers during rolling upgrades and is a prerequisite
+                    // blocker for the #765 chain-id protocol split. Warn for operator
+                    // visibility only.
                     ReqResInboundFailure::UnsupportedProtocols => {
-                        warn!(target: "network", ?peer, ?request_id, ?error, "inbound failure: unsupported protocol");
-
-                        // the local peer supports none of the protocols requested by the remote
-                        // Severe (not Fatal) so version skew during rolling upgrades does not
-                        // instantly ban a peer that is otherwise well-behaved.
-                        self.swarm
-                            .behaviour_mut()
-                            .peer_manager
-                            .process_penalty(peer, Penalty::Severe);
+                        warn!(target: "network", ?peer, ?request_id, ?error, "inbound failure: unsupported protocol (not penalized)");
                     }
                     ReqResInboundFailure::Timeout | ReqResInboundFailure::ConnectionClosed => {
                         // peer dropped or stalled mid-request — expected on WAN, no penalty
--- a/crates/network-libp2p/src/peers/behavior.rs
+++ b/crates/network-libp2p/src/peers/behavior.rs
@@ -31,6 +31,14 @@ impl NetworkBehaviour for PeerManager {
         //
         // ensure PeerId isn't banned if known and register dial attempt
         if let Some(peer_id) = maybe_peer {
+            // refuse to dial our own identity. Kademlia can auto-dial an address
+            // it learned for a peer id (e.g. our own record re-learned via a
+            // hairpin address); if that id is ours, deny it here before a
+            // self-connection is established.
+            if self.is_local_peer(&peer_id) {
+                debug!(target: "peer-manager", ?peer_id, "denying outbound connection to local peer id");
+                return Err(ConnectionDenied::new("refusing to dial self"));
+            }
             // PeerManager and Kad may initiate dials
             // intercept kad dial attempts, sanitize, and register
             if self.dial_attempt_already_registered(&peer_id) {
@@ -84,6 +92,13 @@ impl NetworkBehaviour for PeerManager {
         remote_addr: &Multiaddr,
     ) -> Result<THandler<Self>, ConnectionDenied> {
         trace!(target: "peer-manager", ?peer, ?remote_addr, "inbound connection established");
+        // drop a self-connection (loopback/hairpin back to our own id) without
+        // scoring it. The inbound peer id is only known at this stage, so this is
+        // the earliest point an inbound self-connection can be rejected.
+        if self.is_local_peer(&peer) {
+            debug!(target: "peer-manager", ?peer, ?remote_addr, "denying inbound self-connection");
+            return Err(ConnectionDenied::new("self-connection: remote peer id is our own"));
+        }
         // ensure banned peers are not accepted
         if self.peer_banned(&peer) {
             return Err(ConnectionDenied::new("peer is banned"));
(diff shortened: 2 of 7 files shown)


grantkee (approved):
Wow! I assumed libp2p would have this in their core swarm logic, but we've seen this issue in logs. Good work @MavenRain

PR #781, opened 2026-06-24
feat(network): cut the worker batch path over to the typed sync protocol with legacy fallback (739, step 5)

MavenRain:
## Summary

Step 5 of #739. The typed sync-frame layer registered in step 4 (#768) was inert: a worker advertised `/tn-worker-{id}-sync/0.0.1` but never opened it, so every batch fetch still ran the legacy `/tn-stream` ack-plus-digest handshake. This PR flips the worker batch path over to the sync protocol as the **first** real opener, with a **penalty-exempt fallback to legacy** for any peer that does not speak it.

The result folds the legacy two-roundtrip handshake (a `RequestBatchesStream` request-response that returns an `ack`, then a raw stream open onto which the requester writes a 32-byte correlation digest) into a **single stream open** whose opening `SyncFrame::Req` carries the request itself.

The primary path is untouched and stays fully legacy; its cutover is step 6.

### Requester (worker `network/handle.rs`)
- `request_batches_from_peer` now prefers the sync protocol. Unless the peer is cached as legacy-only this epoch, it opens a `/tn-worker-{id}-sync` stream, writes `SyncFrame::Req(WorkerSyncRequest::Batches { .. })` in the opening frame, and reads the response (`Ack`, then `Data` frames terminated by `End`).
- A `SyncAttempt` enum classifies the outcome:
  - `Fetched`: the peer served the batches → cache it sync-capable.
  - `Unsupported`: the peer never answered (negotiation failed = a pre-item-4 peer, or it negotiated sync but sent no `Ack` = an item-4 peer that registered the protocol without serving it) → cache it legacy-only and fall back to the legacy path.
  - `Failed`: the exchange failed after the peer proved sync-capable (`Deny`/`Err`/mid-stream error), or a transient transport error occurred (an I/O error opening or writing the stream, distinct from a protocol mismatch) → keep it sync-capable and try the next peer.
- A per-peer `sync_capability` cache (`Arc<Mutex<HashMap<BlsPublicKey, bool>>>`) avoids re-probing a known legacy-only peer every request. It is **cleared on `update_epoch`** so a peer upgraded across the committee-rotation boundary is re-probed.
- The sync probe is **penalty-exempt**, and a transient transport error never demotes a sync-capable peer: a genuine negotiation failure surfaces as `StreamError::UpgradeFailed` and maps to `Unsupported` (fall back, cache legacy-only), while a transport I/O error during the open surfaces as the distinct `StreamError::UpgradeIo` and maps to `Failed` (try the next peer, peer stays sync-capable, cache not poisoned). Neither scores the peer.

### Responder (worker `network/mod.rs`, `handler.rs`, `stream_codec.rs`)
- Inbound streams now dispatch on the negotiated `StreamKind`: `Legacy` → the existing digest-correlation reader, `Sync` → the new `process_inbound_sync_stream`.
- Because the sync request travels in the stream's opening frame (no prior request-response ack), admission against the concurrency caps moves to stream-open time:
  - `try_admit_sync` acquires the **shared** `batch_stream_semaphore` (so the global `MAX_CONCURRENT_BATCH_STREAMS` cap is the combined legacy+sync limit) and checks the peer's combined in-flight count (legacy pending + sync in-flight) against `MAX_PENDING_REQUESTS_PER_PEER`. The legacy admission path (`process_request_batches_stream`) likewise adds the peer's in-flight sync-stream count to its legacy count, so the per-peer cap is enforced **symmetrically across both paths** (both lock `pending_batch_requests` before `sync_stream_peers`, so they never deadlock). A `SyncStreamPermit` RAII guard releases the global slot and decrements the per-peer count on drop.
  - A shedding responder writes `SyncFrame::Deny(DenyReason::AtCapacity)` without reading, so the requester gives up immediately and tries elsewhere instead of waiting out its ack timeout. Both the shed write and the malformed-frame `Err` write are bounded by `SYNC_REQUEST_READ_TIMEOUT` so a non-reading peer cannot stall the responder task on an unbounded write.
- Once admitted, the opening `Req` frame is read (bounded by `SYNC_REQUEST_READ_TIMEOUT`), the digest set is capped to `MAX_BATCH_DIGESTS_PER_REQUEST`, and `send_sync_batches_over_stream` serves `Ack` + one `Data` frame per found batch (chunked DB reads) + `End`. It mirrors the legacy responder's partial-DB tolerance: only batches present in storage are streamed.
- Responder errors are **metrics-only** during the rollout (logged, best-effort `SyncFrame::Err`, no penalty), matching the step-3 stream-failure policy and the legacy responder.

### Plumbing (`network-libp2p`)
- `open_stream` / `NetworkCommand::OpenStream` / `HandlerCommand::OpenStream` / `NetworkEvent::InboundStream` / `StreamEvent::InboundStream` now carry a `StreamKind` (`Legacy` | `Sync`).
- An **outbound** open advertises the single chosen protocol (so a sync open negotiates sync or fails, instead of silently negotiating legacy at the wire level); an **inbound** listen advertises both (legacy first) and the upgrade reports which negotiated, so the application routes the stream.
- `classify_outbound` distinguishes a genuine negotiation failure (`StreamError::UpgradeFailed`: the peer speaks none of our protocols) from a transport I/O error during negotiation (the new `StreamError::UpgradeIo`), so the requester falls back on the former but retries on the latter rather than caching a flapping-but-capable peer as legacy-only.
- A sync open whose negotiation fails (`UnsupportedProtocol`) is **not** reported as an outbound failure for scoring (the caller falls back); every other failure, and every legacy failure, is still reported.

### Primary (`primary/network/mod.rs`)
- Compile-only follow-through: opens `StreamKind::Legacy` and routes inbound `Legacy` to its existing reader. An unexpected inbound `Sync` stream (the primary registers the protocol but opens none until step 6) is dropped with a warning, no penalty.

## Cross-version behaviour

|---|---|---|
| item-5 (sync-first) | item-5 | sync exchange, single open |
| item-5 | item-4 (registered, not served) | no `Ack` → requester falls back to legacy; no penalty |
| item-5 | pre-item-4 (no protocol) | negotiation fails → requester falls back to legacy; no penalty |
| pre-item-5 (legacy-only) | item-5 | responder still accepts the legacy `/tn-stream` open |

## Wire protocol

- Each `SyncFrame` is one length-prefixed, snappy-compressed `encode_message` unit; the BCS enum discriminant is the frame tag (pinned by `frame_tags_are_stable` / `request_tags_are_stable` from step 4).
- A `Data` payload is a BCS-encoded `Batch`; the frame layer adds its own length prefix and compression.
- Every frame is bounded by `max_sync_frame_size(epoch) = max_batch_size(epoch) + 1 KiB` envelope headroom, so a peer cannot force an oversized allocation.

## Tests

- New `stream_codec` round-trip tests proving the item-5 responder serializer and the item-5 requester reader agree on the wire: `test_send_receive_sync_roundtrip`, `test_send_receive_sync_partial_db` (responder holds a subset), `test_send_receive_sync_multi_chunk` (>200 batches, multiple chunks).
- The six legacy-flow IT tests now answer the sync probe via a small `deny_sync_open` helper that returns `StreamError::UpgradeFailed` (exactly what a pre-item-4 peer's failed negotiation produces), so they keep exercising the legacy fallback path end-to-end.

## Gates (all green)

- `cargo +nightly-2026-03-20 fmt --check`
- `cargo test -p tn-network-libp2p` (181), `-p tn-worker` (28 lib + 32 IT), `-p tn-primary` (111 lib + 40 IT)
- `cargo clippy -p tn-network-libp2p -p tn-worker --all-features --all-targets` and `cargo clippy -p tn-primary --all-features --lib` — the changed crates and their consumers compile clean
- `cargo doc --no-deps -D warnings -p tn-network-libp2p -p tn-worker`

A workspace-wide `cargo clippy --all-targets` and `cargo doc --no-deps -D warnings` each trip one **pre-existing** `tn-primary` issue outside this diff, a `never_loop` in `src/tests/certificate_fetcher_tests.rs` and a broken `[LeaderSwapBoard]` intra-doc link in `src/consensus/bullshark.rs`, both unrelated to the worker batch-sync cutover.

BCS encodes enum variant indices: no `SyncFrame` / `WorkerSyncRequest` variant is reordered or removed here. The coordinated `/0.0.2` bump and legacy freeze remain step 9.

### DIFF
--- a/crates/consensus/worker/src/network/handle.rs
+++ b/crates/consensus/worker/src/network/handle.rs
@@ -5,18 +5,22 @@
 //! context.
 
 use std::{
-    collections::{BTreeSet, HashSet},
+    collections::{BTreeSet, HashMap, HashSet},
+    sync::Arc,
     time::Duration,
 };
 
 use futures::{AsyncRead, AsyncWriteExt as _};
+use parking_lot::Mutex;
 use tn_network_libp2p::{
     error::NetworkError,
+    read_frame,
     types::{NetworkHandle, NetworkResponseMessage, NetworkResult},
-    Penalty,
+    write_frame, Penalty, StreamError, StreamKind, SyncFrame, WorkerSyncRequest,
 };
 use tn_types::{
-    encode, max_batch_size, Batch, BlockHash, BlsPublicKey, Epoch, SealedBatch, TaskSpawner, B256,
+    encode, max_batch_size, try_decode, Batch, BlockHash, BlsPublicKey, Epoch, SealedBatch,
+    TaskSpawner, B256,
 };
 use tracing::{debug, warn};
 
@@ -28,11 +32,37 @@ use crate::{
 /// Timeout for streaming a single batch from peer. Batches capped at 1MB.
 const BATCH_STREAM_TIMEOUT: Duration = Duration::from_secs(5);
 
+/// Timeout for the responder's first sync frame (`Ack`/`Deny`) after the request
+/// frame is written. A peer that negotiated the sync protocol but does not answer
+/// (e.g. an item-4 node that registered the protocol but reads the stream on the
+/// legacy digest path) trips this and the requester falls back to legacy.
+const SYNC_ACK_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Headroom added to `max_batch_size` for the sync-frame envelope (the `SyncFrame`
+/// enum tag and the `Data` length prefix) when bounding a decoded frame.
+const SYNC_FRAME_OVERHEAD: usize = 1024;
+
 /// Maximum number of retries through the full peer list in `request_batches()`.
 const MAX_BATCH_REQUEST_RETRIES: usize = 3;
 
 /// Delay between retry attempts in `request_batches()` to give semaphores time to release.
 const BATCH_REQUEST_RETRY_DELAY: Duration = Duration::from_millis(500);
+
+/// The largest sync frame accepted for `epoch`: a `Data` frame carrying one
+/// encoded batch, plus envelope headroom.
+pub(crate) fn max_sync_frame_size(epoch: Epoch) -> usize {
+    max_batch_size(epoch).saturating_add(SYNC_FRAME_OVERHEAD)
+}
+
+/// The outcome of a sync-protocol batch fetch attempt.
+enum SyncAttempt {
+    /// The peer served the requested batches over the sync protocol.
+    Fetched(Vec<(BlockHash, Batch)>),
+    /// The peer did not answer the sync exchange; fall back to legacy.
+    Unsupported,
+    /// The peer answered but the exchange failed; try the next peer.
+    Failed(NetworkError),
+}
 
 /// The wrapper around worker-specific network calls.
 #[derive(Clone, Debug)]
@@ -43,12 +73,21 @@ pub struct WorkerNetworkHandle {
     task_spawner: TaskSpawner,
     /// The current epoch for this node.
     epoch: Epoch,
+    /// Per-peer sync-protocol capability, learned by probing.
+    ///
+    /// Absent means not yet probed (try the sync protocol); `false` means the
+    /// peer did not answer a sync open (a pre-item-4 peer that fails negotiation,
+    /// or an item-4 peer that registered the protocol but does not serve it), so
+    /// go straight to legacy; `true` means the peer served a sync exchange. The
+    /// cache is reset each epoch ([`Self::update_epoch`]) so a peer upgraded over
+    /// the rotation boundary is re-probed.
+    sync_capability: Arc<Mutex<HashMap<BlsPublicKey, bool>>>,
 }
 
 impl WorkerNetworkHandle {
     /// Create a new instance of [Self].
     pub fn new(handle: NetworkHandle<Req, Res>, task_spawner: TaskSpawner, epoch: Epoch) -> Self {
-        Self { handle, task_spawner, epoch }
+        Self { handle, task_spawner, epoch, sync_capability: Arc::new(Mutex::new(HashMap::new())) }
     }
 
     /// Return a reference to the task spawner.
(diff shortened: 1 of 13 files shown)


crates/consensus/worker/src/network/handler.rs
```diff
@@ -241,6 +244,74 @@ where
...
+        .await
+        .map_err(WorkerNetworkError::from)
+        .and_then(|sent| sent);
+
+        // a send failure or timeout is logged and best-effort signalled so the
+        // requester stops waiting; it is not a peer fault, so no penalty
+        if let Err(e) = served {
+            warn!(target: "worker::network", %peer, ?e, "failed to serve sync batch stream");
+            let (mut encode_buffer, mut compressed_buffer) = (Vec::new(), Vec::new());
+            let _ = write_frame(
```
grantkee: I think this should probably have a timeout as well. It's a small attack surface, but a malicious peer could:

1) pass admission
1) send a valid `Req`
1) read the `Ack`
1) stop reading and wait out the full 200s `SEND_STREAM_TIMEOUT` 
1) never poll this stream causing this to `await` indefinitely

The current per-peer cap is 2, so a single peer could pin at most 2 of the 5 global slots. To fully exhaust a node, it would take multiple distinct peers, each burning ~200s per slot. It seems unlikely but still possible

crates/consensus/worker/src/network/handle.rs
```diff
@@ -204,13 +243,246 @@ impl WorkerNetworkHandle {
...
+            read_frame::<_, WorkerSyncRequest>(
+                &mut stream,
+                &mut decode_buffer,
+                &mut compressed_buffer,
+                max_frame,
+            ),
+        )
+        .await
+        .map_err(|_elapsed| SyncAttempt::Unsupported)?
+        .map_err(|e| {
```
grantkee, line 390: If a peer hasn't upgraded yet, would this return `std::io::ErrorKind::UnexpectedEof` ? I think as it stands now, an unsupported peer would be considered `Failed` instead of the intended `Unsupported`. Consider matching in this map in case the peer can't deserialize the unexpected frame

```rust
      .map_err(|e| match e.kind() {
          // a clean EOF / connection close means the peer did not answer the sync
          // protocol (e.g. a pre-cutover peer that closed the misread stream) ->
          // fall back to legacy and cache legacy-only, exactly like the timeout case
          std::io::ErrorKind::UnexpectedEof
          | std::io::ErrorKind::ConnectionReset
          | std::io::ErrorKind::BrokenPipe => SyncAttempt::Unsupported,
          // any other I/O error after a successful negotiation is transient ->
          // keep the peer sync-capable and try the next peer
          _ => SyncAttempt::Failed(NetworkError::RPCRetryable(format!(
              "failed to read sync ack frame: {e}"
          ))),
      })?;
```

crates/consensus/worker/src/network/stream_codec.rs
```diff
@@ -122,6 +123,73 @@ where
...
+
+    // accept the exchange before streaming any data
+    write_frame(
+        stream,
+        &SyncFrame::<WorkerSyncRequest>::Ack,
+        &mut encode_buffer,
+        &mut compressed_buffer,
+        max_frame,
+    )
+    .await?;
```
grantkee: Worth flushing here to ensure the peer receives the `Ack`?

grantkee (dismissed):
Everything LGTM! I have a few comments for your consideration

crates/consensus/worker/src/network/handle.rs
MavenRain: I think I've addressed this below, modulo style . . . LMK

grantkee (approved):
LGTM!

PR #785, opened 2026-06-25
fix(network): deliver accepted gossip when relaying peer's BLS is unresolved

### LINKED ISSUE #776: Deliver accepted gossip even when the relaying peer's BLS identity is   unresolved (drop is stronger than necessary)
## Summary

  When an authenticated gossip message is accepted but the relaying peer's BLS
  identity is not yet resolved, we currently drop the message locally and emit a
  warning. The drop is unnecessary: `propagation_source` is used only for penalty
  attribution and a log label, never for the payload itself. We should deliver
  the payload anyway and handle the unresolved relayer explicitly, which removes
  both the dropped delivery and the warning burst seen on node restart.

  ## Current behavior

  In process_gossip_event (crates/network-libp2p/src/consensus.rs:866-897), after
  a message passes verify_gossip (which authenticates the message author
  gossip.source, not the relayer, consensus.rs:1162) and is reported Accept to
  gossipsub:
```
  if let Some(bls) =
  self.swarm.behaviour().peer_manager.peer_to_bls(&propagation_source) {
      // forward NetworkEvent::Gossip(message, bls)
  } else {
      warn!(target: "network", ?propagation_source, ?message_id,
          "accepted gossip not delivered locally: relaying peer's BLS identity is not yet resolved");
  }
```
  The else branch and warning were added in #743 / #754 to make the drop
  observable. On a cold restart this fires repeatedly ("a bunch"), because peers
  join the gossipsub mesh and relay before their signed NodeRecord (which carries
  the BLS key) has been resolved over kad. That window is expected and
  self-heals; the symptom is a noisy log plus a discarded, already-authenticated
  copy of each affected message.

  ## Why the local drop is unnecessary

  - The relayer identity is not part of the payload.
  NetworkEvent::Gossip(GossipMessage, BlsPublicKey) (types.rs:165) carries the
  relayer BLS as a second field, but the consumers use it for exactly two things:
    - a task-name log label (primary mod.rs:876, worker mod.rs:265), and
    - `report_penalty(propagation_source, penalty)` when deeper processing fails
  (primary mod.rs:883, worker mod.rs:271).
  - The payload decode ignores propagation_source entirely
  (crates/consensus/primary/src/network/handler.rs:195 takes only &msg).
  - Recovery does happen via pull-based sync (cert fetcher keyed on the committee
  object, certificate_fetcher.rs:375; worker batch fetcher,
  batch_fetcher.rs:124-211), so consensus state is not lost. But the dropped copy
  is a real, avoidable waste: gossipsub records the `message_id` in its 60s
  `duplicate_cache` before the app event, so it will never re-deliver that id
  locally even after the identity resolves. The local drop is therefore a
  definite loss of a deliverable message, recovered only later and more
  expensively by sync.

  ## Proposed change

  Stop dropping. Always deliver the accepted payload, and represent the relayer
  identity as optional so consumers can handle the unresolved case.

  Option A (minimal, recommended): change the event to Gossip(GossipMessage,
  Option<BlsPublicKey>). Always forward; when the relayer is unresolved, deliver
  the payload and skip the consensus-layer penalty (it is unattributable without
  the BLS key). Tradeoff: during the brief resolution window, a relayer that
  forwards a payload which later fails deeper processing escapes the
  consensus-layer penalty. This is low risk because the author is already
  authenticated by `verify_gossip` and the relayer remains subject to gossipsub
  peer scoring.

  Option B (preserves attribution): carry the relayer as an enum that always
  holds the PeerId and the BlsPublicKey when known. Add a PeerId-keyed penalty
  path so a deep-fail still penalizes via the existing
  `peer_manager.process_penalty(peer_id, ...)` (manager.rs:385, already used for
  the invalid-gossip branch at consensus.rs:911). Larger change to the penalty
  API, but no attribution loss.

  Recommendation: ship Option A unless we consider the windowed attribution loss
  unacceptable, in which case Option B.

  ## Scope

  - crates/network-libp2p/src/types.rs:165 (event type)
  - crates/network-libp2p/src/consensus.rs:866-897 (deliver instead of drop;
  remove or downgrade the warning)
  - crates/consensus/primary/src/network/mod.rs:630, 872-890 (handle optional
  relayer; penalty)
  - crates/consensus/worker/src/network/mod.rs:204-205, 261-271 (handle optional
  relayer; penalty)
  - crates/network-libp2p/src/types.rs:644 (report_penalty) if Option B

  ## Acceptance criteria

  - An accepted gossip message from a connected-but-unresolved relayer is
  delivered to the consensus layer (both primary and worker paths).
  - The accepted gossip not delivered locally warning at consensus.rs:891 is
  removed (Option A may keep a debug! note for the penalty-skipped case).
  - Penalty behavior is defined for the unresolved case: skipped (A) or routed by
  PeerId (B).
  - Test coverage: extend the existing mapping test around
  crates/network-libp2p/src/tests/network_tests.rs:212 with a case where a
  message relayed by an unresolved peer is still delivered.

  ## Out of scope / notes

  - A kad `get_record` retry at the warning site is not viable: records are keyed
  by BLS public key (consensus.rs:1316), and at that point we only have the
  relayer's PeerId, with no PeerId to RecordKey mapping. The PeerId-keyed query
  that exists (`get_closest_peers`, consensus.rs:1322) returns routing info, not
  the NodeRecord that carries the BLS key. Resolution is already in flight via
  the on-connect record push (consensus.rs:474, 1291) and committee get_record
  (manager.rs:678); this change just stops discarding messages while that
  completes.
  - Related follow-up (separate issue): the worker txn topic subscribes with no
  publisher allowlist (`subscribe(worker_txn_topic)` vs `subscribe_with_publishers`
  for batches), so a non-committee relayer that never publishes a resolvable
  NodeRecord can warn indefinitely with no sync backstop. Worth deciding whether
  to gate or rate-limit that path. This change removes the warning regardless,
  but the underlying never-resolves case is distinct.

MavenRain:
Closes #776.

## Summary

When an authenticated gossip message is accepted but the relaying peer's BLS
identity has not yet resolved, we dropped the message locally and emitted a
warning. The drop is unnecessary: `propagation_source` is used only for penalty
attribution and a log label, never for the payload itself, and the message
author is already authenticated by `verify_gossip`. This delivers the accepted
payload anyway and carries the relayer identity as `Option`, which removes both
the dropped delivery and the warning burst seen on cold restart.

This implements **Option A** from the issue, which the maintainer confirmed is
sufficient ("If the message is valid and published by an authority, then I think
it's fine to deliver to the application layer during the brief resolution
window").

## Why the local drop was unnecessary

- The relayer identity is not part of the payload. The consumers use the second
  field of `NetworkEvent::Gossip` for exactly two things: a task-name log label
  and `report_penalty(relayer, penalty)` when deeper processing fails.
- The payload decode ignores the relayer entirely.
- Dropping was a definite loss: gossipsub records the `message_id` in its 60s
  duplicate cache before the app event, so it never re-delivers that id locally
  even after the identity resolves. The message was recovered only later and more
  expensively by pull-based sync.

## Changes

- `crates/network-libp2p/src/types.rs`: `NetworkEvent::Gossip` now carries
  `Option<BlsPublicKey>` (the relaying peer's BLS identity, `None` while
  unresolved).
- `crates/network-libp2p/src/consensus.rs`: deliver the accepted payload
  unconditionally instead of dropping it; the "accepted gossip not delivered
  locally" `warn!` is removed and replaced with a `debug!` note for the
  penalty-skipped case. Delivery now goes through a small `accepted_gossip_event`
  helper so the unresolved-relayer case is deterministically testable.
- `crates/consensus/primary/src/network/mod.rs` and
  `crates/consensus/worker/src/network/mod.rs`: `process_gossip` takes
  `Option<BlsPublicKey>`. The consensus-layer penalty is applied only when the
  relayer has resolved (`relayer.zip(penalty)`); it is unattributable otherwise.
  The task-name label renders an unresolved relayer as `unresolved`.

## Penalty behavior (the windowed tradeoff)

During the brief resolution window, a relayer that forwards a payload which
later fails deeper processing escapes the consensus-layer penalty. This is low
risk: the author is already authenticated by `verify_gossip`, and the relayer
remains subject to gossipsub peer scoring (the invalid-gossip path still applies
a fatal penalty to the propagation source unchanged).

## Tests

- `crates/network-libp2p/src/tests/network_tests.rs`:
  - `accepted_gossip_with_unresolved_relayer_is_delivered` — regression guard
    that an accepted message with an unresolved relayer is delivered (carried as
    `None`), never dropped.
  - `accepted_gossip_with_resolved_relayer_carries_identity` — the resolved
    identity is attached.
  - `test_msg_verification_ignores_unauthorized_publisher` extended to assert the
    resolved relayer's BLS identity is carried through the real gossip path.

The unresolved-relayer state is a brief, non-deterministic resolution-window
race that cannot be reproduced reliably through the live-network harness (a
connected mesh neighbor resolves via the on-connect record push), so the
delivery invariant is guarded by the focused unit tests above, mirroring the
`resolve_response` unit tests added for the sibling response-path fix in #743 /
#754.

### Gates

- `cargo check -p tn-network-libp2p -p tn-primary -p tn-worker` — green
- `cargo clippy --all-targets -p tn-network-libp2p -- -D warnings` — green
- `cargo clippy --lib -p tn-primary -p tn-worker -- -D warnings` — green
- `cargo fmt --check` — clean
- tests: tn-network-libp2p 183 passed, tn-worker 57 passed, tn-primary 153
  passed (0 failed)

## Out of scope (follow-up)

The worker txn topic subscribes with no publisher allowlist
(`subscribe(worker_txn_topic)` vs `subscribe_with_publishers` for batches), so a
non-committee relayer that never publishes a resolvable `NodeRecord` could warn
indefinitely with no sync backstop. This change removes the warning regardless,
but the underlying never-resolves case is distinct and worth a separate decision
on whether to gate or rate-limit that path.

### DIFF
--- a/crates/consensus/primary/src/network/mod.rs
+++ b/crates/consensus/primary/src/network/mod.rs
@@ -799,8 +799,8 @@ where
                     self.process_consensus_output_stream(peer, number, channel, cancel)
                 }
             },
-            NetworkEvent::Gossip(msg, propagation_source) => {
-                self.process_gossip(msg, propagation_source);
+            NetworkEvent::Gossip(msg, relayer) => {
+                self.process_gossip(msg, relayer);
             }
             NetworkEvent::Error(msg, channel) => {
                 let err = PrimaryResponse::Error(PrimaryRPCError(msg));
@@ -1065,18 +1065,21 @@ where
     }
 
     /// Process gossip from committee.
-    fn process_gossip(&self, msg: GossipMessage, propagation_source: BlsPublicKey) {
+    fn process_gossip(&self, msg: GossipMessage, relayer: Option<BlsPublicKey>) {
         // clone for spawned tasks
         let request_handler = self.request_handler.clone();
         let network_handle = self.network_handle.clone();
-        let task_name = format!("ProcessGossip-{}-{propagation_source}", msg.topic);
+        let relayer_label =
+            relayer.as_ref().map_or_else(|| "unresolved".to_string(), |bls| bls.to_string());
+        let task_name = format!("ProcessGossip-{}-{relayer_label}", msg.topic);
         // spawn task to process gossip
         self.task_spawner.spawn_task(task_name, async move {
             if let Err(e) = request_handler.process_gossip(&msg).await {
                 warn!(target: "primary::network", ?e, "process_gossip");
-                // convert error into penalty to lower peer score
-                if let Some(penalty) = (&e).into() {
-                    network_handle.report_penalty(propagation_source, penalty).await;
+                // convert error into penalty to lower peer score; only attributable
+                // when the relaying peer's BLS identity has resolved
+                if let Some((relayer, penalty)) = relayer.zip((&e).into()) {
+                    network_handle.report_penalty(relayer, penalty).await;
                 }
                 Err(e.into())
             } else {
--- a/crates/consensus/worker/src/network/mod.rs
+++ b/crates/consensus/worker/src/network/mod.rs
@@ -201,8 +201,8 @@ where
                     warn!(target: "worker::network", "worker application received unexpected peer exchange message");
                 }
             },
-            NetworkEvent::Gossip(msg, propagation_source) => {
-                self.process_gossip(msg, propagation_source);
+            NetworkEvent::Gossip(msg, relayer) => {
+                self.process_gossip(msg, relayer);
             }
             NetworkEvent::Error(msg, channel) => {
                 let err = WorkerResponse::Error(message::WorkerRPCError(msg));
@@ -261,17 +261,20 @@ where
     }
 
     /// Process gossip from a worker.
-    fn process_gossip(&self, msg: GossipMessage, propagation_source: BlsPublicKey) {
+    fn process_gossip(&self, msg: GossipMessage, relayer: Option<BlsPublicKey>) {
         // clone for spawned tasks
         let request_handler = self.request_handler.clone();
         let network_handle = self.network_handle.clone();
-        let task_name = format!("process-gossip-{propagation_source}");
+        let relayer_label =
+            relayer.as_ref().map_or_else(|| "unresolved".to_string(), |bls| bls.to_string());
+        let task_name = format!("process-gossip-{relayer_label}");
         self.network_handle.get_task_spawner().spawn_task(task_name, async move {
             if let Err(e) = request_handler.process_gossip(&msg).await {
                 warn!(target: "worker::network", ?e, "process_gossip");
-                // convert error into penalty to lower peer score
-                if let Some(penalty) = e.penalty() {
-                    network_handle.report_penalty(propagation_source, penalty).await;
+                // convert error into penalty to lower peer score; only attributable
+                // when the relaying peer's BLS identity has resolved
+                if let Some((relayer, penalty)) = relayer.zip(e.penalty()) {
+                    network_handle.report_penalty(relayer, penalty).await;
                 }
                 Err(e.into())
             } else {
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -859,34 +859,32 @@ where
                 if valid {
                     // A peer is `Connected` before its `NodeRecord` resolves its BLS
                     // identity, so a live mesh neighbor can relay a message before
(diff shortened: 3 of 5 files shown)


sstanfield (approved):
LGTM!

PR #786, opened 2026-06-26
fix(network): surface the real DialError to the dialer instead of a hardcoded timeout (#745)

### LINKED ISSUE #745: network-libp2p: every dial failure is reported to the caller as the hardcoded string "dial attempt timedout", erasing the real `DialError` (wrong key, refused, firewall) during validator onboarding
## Summary

When an outbound dial fails, the peer manager computes the real `DialError` and also has a hardcoded fallback string.  Because of call ordering, the hardcoded string always wins: `on_dial_failure` calls `register_disconnected` first, and that path consumes the caller's reply channel while sending `Err(NetworkError::Dial("dial attempt timedout"))`.  The immediately following line then tries to deliver the genuine `DialError`, but the reply channel is already gone, so that send is a silent no-op.  The caller therefore sees "dial attempt timedout" for every dial failure, whether the true cause was a wrong network key (`WrongPeerId`), a refused connection, a firewall drop, or an actual timeout.

### The hardcoded error, sent first (`crates/network-libp2p/src/peers/all_peers.rs:512-528`)

The `Unknown | Connected | Dialing` arm of the disconnect transition unconditionally reports a timeout, regardless of the real cause:

```rust
ConnectionStatus::Unknown
| ConnectionStatus::Connected { .. }
| ConnectionStatus::Dialing { .. } => {
    self.disconnected_peers += 1;
    let id = self.identity_for(peer_id);
    if let Some(peer) = self.peers.get_mut(&id) {
        peer.set_connection_status(ConnectionStatus::Disconnected { instant: Instant::now() });
    }

    // notify caller of dial error if present
    self.notify_dial_result(
        peer_id,
        Err(NetworkError::Dial("dial attempt timedout".to_string())),
    );
}
```

### The reply channel is consumed on first send (`crates/network-libp2p/src/peers/all_peers.rs:370-383`)

```rust
pub(super) fn reply_for_dial_attempt(&mut self, peer_id: &PeerId) -> Option<oneshot::Sender<NetworkResult<()>>> {
    self.pending_dials.remove(peer_id) // takes the channel out
}

pub(super) fn notify_dial_result(&mut self, peer_id: &PeerId, result: NetworkResult<()>) {
    if let Some(reply) = self.reply_for_dial_attempt(peer_id) {
        send_or_log_error!(reply, result, "DialResult", peer = peer_id);
    }
}
```

### The real error is computed, then dropped (`crates/network-libp2p/src/peers/behavior.rs:294-303`)

```rust
fn on_dial_failure(&mut self, peer_id: Option<PeerId>, error: &DialError) {
    if let Some(peer_id) = peer_id {
        if !self.is_connected(&peer_id) {
            self.register_disconnected(&peer_id); // <-- consumes the reply with the hardcoded string
        }
        // return error to dialer
        self.notify_dial_result(&peer_id, Err(error.into())); // <-- real DialError, but channel already gone: no-op
    }
}
```

`error.into()` uses `From<&DialError> for NetworkError` (`crates/network-libp2p/src/error.rs:138-142`), which preserves the real cause as `NetworkError::Dial(e.to_string())`.  That conversion runs, but its result never reaches the caller.

## Impact

User experience and operability.  Dial failures are the most common validator-onboarding problem, and this collapses every distinct cause into one misleading message.  An operator who has the wrong network key configured, whose port is firewalled, or whose target is genuinely unreachable all see the identical "failed to dial <bls>: dial attempt timedout" and cannot tell the difference.  This actively slows diagnosis of misconfiguration.

## Suggested fix

Make the real `DialError` reach the caller.  Two options:

1. Reorder `on_dial_failure` so the genuine error is delivered before `register_disconnected` consumes the channel, or
2. Pass the real error into `register_disconnected` / the disconnect-transition path so the `Dialing` arm reports the actual cause instead of the hardcoded `"dial attempt timedout"` string.

Either way, drop the hardcoded literal; it is only ever correct by coincidence.

## Scope and novelty

Confirmed end to end on `main` (`0bd36414`); reachability traced from `DialBls`/`Dial` commands through `dial_peer` to the `DialFailure` swarm event.  No existing issue or PR addresses the error-erasure ordering.  Not covered by the `network-libp2p` security review (PR #620, NL-001..NL-010), which flags dial *policy* gaps (NL-002 temp-ban) but not the lost error cause.

MavenRain:
## Summary

Fixes #745. Makes the genuine `DialError` reach the dialer instead of the hardcoded `"dial attempt timedout"` string, so an operator onboarding a validator can tell a wrong network key (`WrongPeerId`) from a refused connection, a firewall drop, or an actual timeout.

Root cause was call ordering. `on_dial_failure` calls `register_disconnected` first, and the disconnect transition's `Unknown | Connected | Dialing` arm unconditionally notified the dialer with `Err(NetworkError::Dial("dial attempt timedout"))`. That send consumed the caller's reply channel, so the immediately following line that delivers the real `error.into()` found the channel already gone and was a silent no-op. Every dial failure therefore collapsed to the same misleading "timeout".

## Fix

The hardcoded literal lived in a shared transition path but was only ever correct by coincidence. It is removed from there, and the two callers that actually own a live dial now report the real cause:

- **`handle_disconnected_transition`** (`peers/all_peers.rs`): the `Unknown | Connected | Dialing` arm no longer notifies a dial result. It only updates connection status, so it can no longer consume the reply channel ahead of the real error.
- **`on_dial_failure`** (`peers/behavior.rs`): unchanged in behavior, but now the sole notifier on the failure path. Because `register_disconnected` no longer eats the channel, its existing `notify_dial_result(&peer_id, Err(error.into()))` delivers the genuine `DialError` to the caller.
- **`heartbeat_maintenance`** (`peers/all_peers.rs`): this is the one path where a timeout really is the cause (a peer still `Dialing` past `dial_timeout`). It now notifies the dialer with the timeout error explicitly, so the genuine timeout message is preserved exactly where it is true.

The two notifiers never both fire for one dial: whichever reaps the pending dial first (the swarm's `DialFailure` via `on_dial_failure`, or the heartbeat backstop) consumes the reply channel, and the later path finds it already taken and is a no-op.

`on_dial_failure` is bumped from private to `pub(super)` as a test seam (contained to the `peers` module, consistent with the crate's existing test-visibility pattern).

## Tests

- `tests/peer_manager.rs::test_dial_failure_surfaces_real_error` — registers an in-flight dial, drives `on_dial_failure` with a concrete non-timeout cause (`DialError::Aborted`), and asserts the caller receives that real cause and specifically not the old `"dial attempt timedout"` literal.
- `tests/peers.rs::test_heartbeat_dial_timeout_notifies_caller` — ages a `Dialing` peer past `dial_timeout`, runs `heartbeat_maintenance`, and asserts the caller is still notified of the genuine timeout (proving the relocated message).

## CI

Run against the repo's pinned toolchain (`1.94`):

- [x] `cargo fmt --all -- --check` — clean (also clean on `nightly-2026-03-20`)
- [x] `cargo clippy -p tn-network-libp2p --all-targets --locked -- -D warnings` — zero warnings
- [x] `cargo test -p tn-network-libp2p --lib` — 188 passed, 0 failed

### DIFF
--- a/crates/network-libp2p/src/peers/all_peers.rs
+++ b/crates/network-libp2p/src/peers/all_peers.rs
@@ -384,6 +384,14 @@ impl AllPeers {
         // disconnect peers
         for peer_id in peers_to_disconnect {
             self.update_connection_status(&peer_id, NewConnectionStatus::Disconnected);
+
+            // these peers exceeded `dial_timeout` while still `Dialing`, so the genuine cause
+            // is a timeout. Report it to the caller (the disconnect transition no longer
+            // notifies; the dialer's reply channel is only consumed here or in `on_dial_failure`).
+            self.notify_dial_result(
+                &peer_id,
+                Err(NetworkError::Dial("dial attempt timedout".to_string())),
+            );
         }
 
         // update scores for all other peers
@@ -621,11 +629,11 @@ impl AllPeers {
                     });
                 }
 
-                // notify caller of dial error if present
-                self.notify_dial_result(
-                    peer_id,
-                    Err(NetworkError::Dial("dial attempt timedout".to_string())),
-                );
+                // NOTE: the dial result (if any) is notified by the caller that knows the real
+                // cause: `on_dial_failure` delivers the genuine `DialError`, and
+                // `heartbeat_maintenance` reports a timeout. This transition must not report a
+                // hardcoded cause, which previously consumed the reply channel before the real
+                // error could be sent.
             }
         }
 
--- a/crates/network-libp2p/src/peers/behavior.rs
+++ b/crates/network-libp2p/src/peers/behavior.rs
@@ -311,14 +311,16 @@ impl PeerManager {
     ///
     /// NOTE: `AllPeers` is only updated if the peer is _not_ already connected. It's possible that
     /// an outgoing dial attempt fails because the peer connected during the dial.
-    fn on_dial_failure(&mut self, peer_id: Option<PeerId>, error: &DialError) {
+    pub(super) fn on_dial_failure(&mut self, peer_id: Option<PeerId>, error: &DialError) {
         self.metrics.record_dial_failure();
         if let Some(peer_id) = peer_id {
             if !self.is_connected(&peer_id) {
                 self.register_disconnected(&peer_id);
             }
 
-            // return error to dialer
+            // return the genuine dial error to the dialer. `register_disconnected` no longer
+            // consumes the reply channel with a hardcoded cause, so the real `DialError`
+            // (wrong key, refused, firewall, timeout) reaches the caller.
             self.notify_dial_result(&peer_id, Err(error.into()));
         }
     }
--- a/crates/network-libp2p/src/tests/peer_manager.rs
+++ b/crates/network-libp2p/src/tests/peer_manager.rs
@@ -9,7 +9,7 @@ use assert_matches::assert_matches;
 use libp2p::{
     core::Endpoint,
     kad::GetClosestPeersError,
-    swarm::{ConnectionId, NetworkBehaviour as _},
+    swarm::{ConnectionId, DialError, NetworkBehaviour as _},
 };
 use rand::{rngs::StdRng, SeedableRng as _};
 use std::{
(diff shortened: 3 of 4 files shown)


sstanfield (approved):
LGTM!

PR #787, opened 2026-06-26
Hardening patches for consensus chain.

sstanfield:
This addresses a save output gap and some directory syncs missing from pack file management.  It removes optional from current pack to close some gaps.

Addresses these two findings from an earlier scan:

### 6. `save_consensus_output` fallback silently drops un-persisted outputs and returns Ok
- **Location**: `crates/storage/src/consensus.rs:499-515` (also `consensus.rs:296`, `consensus.rs:599-605`)
- **Claim**: When `current_pack` is `None` and the output is not found in the static pack (or `get_static` fails, or `contains_consensus_header_number` errors — swallowed by `unwrap_or_default()` at line 502), the function emits only a `warn!` and returns `Ok(())`. Callers (`state_sync::save_consensus` → executor handlers) treat `Ok` as durably persisted, then broadcast/sign the output. `current_pack` can be `None` outside the intended replay scenario because `ConsensusChain::new` (line 296) swallows `open_append_exists` errors with `.ok()` — a damaged pack at startup yields `None`, and live committed outputs would be acknowledged but never persisted. `persist_current()` (lines 599-605) is also a silent no-op when `current_pack` is `None`.
- **Key Question**: Is there a production sequence where `current_pack` is `None` while live (non-replay) outputs flow through `save_consensus_output`, and does caller behavior depend on `Ok` implying durability?
- **Relevant Files**: crates/storage/src/consensus.rs, crates/state-sync/src/lib.rs, crates/consensus/executor/src/subscriber.rs
- **Source**: tn-review

### 7. Epoch-level directory entries never fsynced in `base_path`
- **Category**: Durability
- **Location**: `crates/storage/src/consensus.rs:431-446` (rename), `crates/storage/src/consensus_pack.rs:548` and `:769` (create_dir_all), `crates/storage/src/epoch_records.rs:482`
- **Claim**: The branch added `fsync_directory` for file creates/renames inside the archive layer, but nothing ever fsyncs `base_path` itself. (a) `ConsensusChain::stream_import` does `remove_dir_all(&base_dir)` then `rename(&path_base_dir, &base_dir)` with no `fsync_directory(&self.base_path)` — after power failure the delete can be durable while the rename dirent is not, leaving the epoch absent despite import success. (b) `open_append`/`stream_import`/`epoch_records::open_append` create the `epoch-N` directory via `create_dir_all` with no parent fsync, so a whole epoch's pack can vanish after power loss. Both recoverable by re-streaming from peers, but they defeat the durability guarantee `persist()`/`stream_import` claim.
- **Key Question**: After power failure between `stream_import`'s rename (or `open_append`'s `create_dir_all`) and journal commit, does node recovery detect the missing epoch dir and re-stream, and is loss of the node's own current-epoch pack (its `read_last_committed` crash-recovery source) tolerable?
- **Relevant Files**: crates/storage/src/consensus.rs, crates/storage/src/consensus_pack.rs, crates/storage/src/epoch_records.rs, crates/storage/src/archive/data_file.rs
- **Source**: tn-review

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -47,7 +47,13 @@ where
     tokio::spawn(async move {
         // create the epoch manager
         let mut epoch_manager =
-            EpochManager::new(builder, tn_datadir, consensus_db, key_config, version).await?;
+            match EpochManager::new(builder, tn_datadir, consensus_db, key_config, version).await {
+                Ok(epoch_manager) => epoch_manager,
+                Err(err) => {
+                    tracing::error!("Error running node (creating EpochManager): {err}");
+                    return Err(err.into());
+                }
+            };
         let result = epoch_manager.run().await;
         if let Err(err) = &result {
             tracing::error!("Error running node: {err}");
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -47,7 +47,7 @@ where
     tokio::spawn(async move {
         // create the epoch manager
         let mut epoch_manager =
-            EpochManager::new(builder, tn_datadir, consensus_db, key_config, version).await;
+            EpochManager::new(builder, tn_datadir, consensus_db, key_config, version).await?;
         let result = epoch_manager.run().await;
         if let Err(err) = &result {
             tracing::error!("Error running node: {err}");
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -51,7 +51,7 @@ where
                 Ok(epoch_manager) => epoch_manager,
                 Err(err) => {
                     tracing::error!("Error running node (creating EpochManager): {err}");
-                    return Err(err.into());
+                    return Err(err);
                 }
             };
         let result = epoch_manager.run().await;
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -230,15 +230,14 @@ where
         {
             committee_zero
         } else {
-            error!(target: "epoch-manager", "Unable to load commitee zero from the genesis committee!");
+            error!(target: "epoch-manager", "Unable to load committee zero from the genesis committee!");
             return Err(eyre::eyre!(
                 "unable to load committee zero (genesis committee), this is fatal"
             ));
         };
         let epochs_db_path = tn_datadir.epochs_db_path();
         let _ = std::fs::create_dir_all(&epochs_db_path);
-        let consensus_chain =
-            ConsensusChain::new(epochs_db_path, committee_zero).expect("open consensus DB");
+        let consensus_chain = ConsensusChain::new(epochs_db_path, committee_zero)?;
         // shutdown long-running node components
         let node_shutdown = Notifier::new();
 
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -226,10 +226,8 @@ where
         // Note this can only fail if the consensus DB is very broken (bad path for instance).
         // So we will panic for now, this will kill the node on startup for a critical error.
         let committee_zero = if let Ok(committee_zero) =
-            Config::load_from_path_or_default::<Committee>(
-                tn_datadir.committee_path(),
-                ConfigFmt::YAML,
-            ) {
+            Config::load_from_path::<Committee>(tn_datadir.committee_path(), ConfigFmt::YAML)
+        {
             committee_zero
         } else {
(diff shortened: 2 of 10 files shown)


crates/node/src/manager/node.rs
```diff
@@ -222,12 +222,23 @@ where
...
+            error!(target: "epoch-manager", "Unable to load commitee zero from the genesis committee!");
+            return Err(eyre::eyre!(
+                "unable to load committee zero (genesis committee), this is fatal"
+            ));
+        };
         let epochs_db_path = tn_datadir.epochs_db_path();
         let _ = std::fs::create_dir_all(&epochs_db_path);
-        let consensus_chain = ConsensusChain::new(epochs_db_path).expect("open consensus DB");
+        let consensus_chain =
+            ConsensusChain::new(epochs_db_path, committee_zero).expect("open consensus DB");
```
MavenRain: We should use `?` here instead of `expect`, I think.

crates/storage/src/consensus.rs
```diff
@@ -351,21 +378,14 @@ impl ConsensusChain {
...
-            if old_pack.epoch() == committee.epoch() {
-                return Ok(());
-            }
-            old_pack.persist().await?;
-            let mut recents = self.recent_packs.lock();
-            // Evict before pushing so the cache stays capped at PACK_CACHE_SIZE.
-            if recents.len() >= Self::PACK_CACHE_SIZE {
-                let _ = recents.pop_front();
-            }
-            recents.push_back(old_pack);
```
MavenRain, line 364: Did we mean to remove the caching of the `old_pack` before eviction?

crates/storage/src/consensus.rs
```diff
@@ -446,15 +466,8 @@ impl ConsensusChain {
...
                 let _install = self.pack_install.lock().await;
-                let mut current_pack = self.current_pack.lock();
-                let replace_current = if let Some(current_pack) = &*current_pack {
-                    current_pack.epoch() == epoch
-                } else {
-                    false
-                };
-                if replace_current {
-                    *current_pack = None;
-                }
```
MavenRain, line 457: If we're going to drop this clearing, should we hold the locks until after the rename and fsync?

crates/node/src/lib.rs
```diff
@@ -47,7 +47,7 @@ where
     tokio::spawn(async move {
         // create the epoch manager
         let mut epoch_manager =
-            EpochManager::new(builder, tn_datadir, consensus_db, key_config, version).await;
+            EpochManager::new(builder, tn_datadir, consensus_db, key_config, version).await?;
```
MavenRain: This error is swallowed unless the enclosing `JoinHandle` is awaited-upon.

crates/node/src/manager/node.rs
```diff
@@ -222,12 +222,23 @@ where
...
-    ) -> Self {
+    ) -> eyre::Result<Self> {
         // Note this can only fail if the consensus DB is very broken (bad path for instance).
         // So we will panic for now, this will kill the node on startup for a critical error.
+        let committee_zero = if let Ok(committee_zero) =
+            Config::load_from_path::<Committee>(tn_datadir.committee_path(), ConfigFmt::YAML)
+        {
+            committee_zero
+        } else {
+            error!(target: "epoch-manager", "Unable to load commitee zero from the genesis committee!");
```
MavenRain: Typo: "commitee" -> "committee"

crates/node/src/manager/node.rs
sstanfield: Yes.  This used to not return an error and a panic just shut the node down early.  But now has returns and error so you are correct.

crates/storage/src/consensus.rs
sstanfield: I think so.  If we have to re-open it for a request we will open it as read-only vs just caching the rw version.  This is not a big deal, we should never try to write to it but I think it is slightly more correct to open it read only.

crates/storage/src/consensus.rs
sstanfield: This should be fine.  We can use the existing current_pack until this one is ready to go.  We need everything saved and synced etc to open it static.  Unless you think I am missing something I think this is the correct way to go.  No need to hold the lock while syncing etc, that would potentially stall a lot of access to the chain.

crates/node/src/lib.rs
sstanfield: Great catch.

grantkee (approved):
LGTM!

MavenRain (approved):
Let's do it!

PR #790, opened 2026-06-26
This uses partial pack downloads to speed up intra-epoch syncing.

sstanfield:
Uses code from previous PR to allow downloading a partial pack file.  This will be MUCH faster when many consensus headers behind to sync the current epoch.

### DIFF
--- a/crates/state-sync/src/consensus.rs
+++ b/crates/state-sync/src/consensus.rs
@@ -12,14 +12,18 @@ use tn_config::ConsensusConfig;
 use tn_primary::{network::PrimaryNetworkHandle, ConsensusBusApp};
 use tn_storage::{consensus::ConsensusChain, tables::ConsensusHeaderCache};
 use tn_types::{
-    ConsensusHeaderDigest, Database as TNDatabase, Epoch, EpochRecord, Noticer, TaskSpawner,
-    TnReceiver, TnSender as _,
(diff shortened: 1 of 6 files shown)


crates/state-sync/src/consensus.rs
```diff
@@ -326,20 +420,29 @@ async fn manage_new_consensus<DB: TNDatabase>(
...
-    }
-    // Note that the way tasks is used is open to "races" but this is a simple throttle for not
-    // firing too many fetch tasks so not worth the overhead of using a full lock here.  I.e.
-    // one more or less task won't matter.
-    let task_num = tasks.load(Ordering::Relaxed);
-    // Skip for now, this number will be subsumed by gossip once enough tasks end.
-    if task_num < 6 {
+        // On the first epoch we will try to do partial pack download to build retrieve consensus.
         let end_number = last_number.unwrap_or_default();
-        *last_number = Some(number + 1);
```
MavenRain, line 337: By removing this, we can have up to 7 concurrent tasks.  That's cool . . . right?

MavenRain (dismissed):
Just a question

crates/state-sync/src/consensus.rs
sstanfield: Yes, the 6 tasks is just a dumb governor to keep things from running amuck.

grantkee (dismissed):
Merge conflict, but the code LGTM!

grantkee (dismissed):

MavenRain (dismissed):

grantkee (approved):

PR #792, opened 2026-06-26
feat(node): source worker pool base fee from the gas accumulator

### LINKED ISSUE #791: Worker transaction pool ignores the gas accumulator's base fee (hardcoded)
## Problem

When a worker's components are initialized at the start of an epoch, the transaction pool's pending base fee is hardcoded to `MIN_PROTOCOL_BASE_FEE`. The gas accumulator already tracks a per-worker base fee, but the pool never reads it, so the value that gates transaction admission and pricing in the pool can diverge from the base fee the rest of the system considers authoritative for that worker.

This matters because the pool uses `pending_basefee` to decide which transactions are eligible and how they are priced. If the accumulator holds a base fee above the protocol minimum, the pool still admits and prices transactions against the minimum, accepting transactions that should be rejected and mispricing the ones it keeps.

The gap also persists across epochs. The base fee is constant within an epoch but can change at epoch boundaries. On the epoch rollover respawn path, worker components are not re-initialized, so even the hardcoded assignment never runs again which leaves the pool with a stale base fee for the life of the node.

## Solution

The worker pool's base fee should come from the gas accumulator rather than a constant. At epoch setup, the per-worker base fee should be snapshotted from the accumulator and passed into worker component initialization, where it sets the pool's `pending_basefee` directly. Since the base fee is fixed for the duration of an epoch, a snapshot taken at epoch start is valid for the whole epoch.

To cover the respawn path (component initialization is skipped) the engine should expose a setter that updates an already-initialized worker pool's base fee. This setter should run on every epoch start so the pool always reflects the accumulator's current value, and no-op when the worker has not yet been initialized. The init path sets the value at creation; the setter additionally handles rollover.

grantkee:
* Worker transaction pool now takes its `pending_basefee` from the gas accumulator's per-worker value at epoch setup instead of the hardcoded `MIN_PROTOCOL_BASE_FEE`. The `u64` is snapshotted at epoch start, which is valid since the base fee is constant within an epoch.
* Added `ExecutionNode::set_worker_base_fee` (and the inner equivalent), called on every epoch start so the respawn path (where `initialize_worker_components` is skipped) still picks up the new epoch's base fee. No-ops when the worker isn't initialized yet.
* `new_batch_validator` now receives the `BaseFeeContainer` directly rather than the unwrapped `u64`.
* Added `test_worker_pool_base_fee_sourced_from_accumulator` covering both the init path and the every-epoch setter, plus a `NoopEngineToPrimary` test helper. Enabled the `tn-worker/test-utils` feature for IT tests.

closes #791

### DIFF
--- a/crates/node/tests/it/main.rs
+++ b/crates/node/tests/it/main.rs
@@ -22,15 +22,18 @@ use tn_primary::{
     ConsensusBus,
 };
 use tn_reth::{test_utils::seeded_genesis_from_random_batches, RethChainSpec};
+use tn_rpc::{EngineToPrimary, RpcNodeInfo};
 use tn_storage::{consensus::ConsensusChain, mem_db::MemDatabase};
 use tn_test_utils::{
     create_signed_certificates_for_rounds, default_test_execution_node, CommitteeFixture,
 };
 use tn_types::{
-    adiri_genesis, gas_accumulator::GasAccumulator, Batch, ConsensusHeaderDigest, ConsensusNumHash,
-    ExecHeader, Notifier, SealedHeader, TaskManager, TnReceiver as _, TnSender as _, B256,
-    DEFAULT_BAD_NODES_STAKE_THRESHOLD,
+    adiri_genesis, gas_accumulator::GasAccumulator, Batch, ConsensusHeader, ConsensusHeaderDigest,
+    ConsensusNumHash, Epoch, EpochCertificate, EpochDigest, EpochRecord, ExecHeader, Notifier,
+    SealedHeader, TaskManager, TnReceiver as _, TnSender as _, WorkerId, B256,
+    DEFAULT_BAD_NODES_STAKE_THRESHOLD, MIN_PROTOCOL_BASE_FEE,
 };
+use tn_worker::WorkerNetworkHandle;
 use tokio::{
     sync::{mpsc, oneshot},
     time::timeout,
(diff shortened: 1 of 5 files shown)


crates/node/tests/it/main.rs
```diff
@@ -172,6 +175,78 @@ async fn test_catchup_accumulator() -> eyre::Result<()> {
...
+        &temp_dir.path().join("reth"),
+        None,
+    )?;
+
+    // keep the task manager alive for the test so the worker RPC + network tasks keep running.
+    let task_manager = TaskManager::default();
+    let network_handle = WorkerNetworkHandle::new_for_test(task_manager.get_spawner());
+
+    let worker_id: WorkerId = 0;
+    // a deliberately non-MIN value: proves the pool does't hardcodes MIN_PROTOCOL_BASE_FEE.
```
MavenRain: Typo: does't -> doesn't

crates/node/tests/it/main.rs
```diff
@@ -172,6 +175,78 @@ async fn test_catchup_accumulator() -> eyre::Result<()> {
...
+        unreachable!("EngineToPrimary RPC is not exercised in this test")
+    }
+}
+
+/// A worker's transaction pool charges the base fee supplied at epoch setup (the
+/// accumulator's per-worker value) instead of a hardcoded `MIN_PROTOCOL_BASE_FEE`, and
+/// `set_worker_base_fee` updates it for the every-epoch (respawn) path.
+#[tokio::test]
+async fn test_worker_pool_base_fee_sourced_from_accumulator() -> eyre::Result<()> {
+    let temp_dir = TempDir::with_prefix("test_worker_pool_base_fee").unwrap();
```
MavenRain: The expression of this unit test is `eyre::Result<()>`, so `unwrap` is unnecessary.  Use `?`.

MavenRain (dismissed):
Just two nits

crates/node/tests/it/main.rs
grantkee: I appreciate the attention to detail! Updated

crates/node/tests/it/main.rs
grantkee: Agreed - good catch

sstanfield (approved):
LGTM!

MavenRain (approved):

PR #794, opened 2026-06-26
feat(batch-validator): take a u64 base-fee snapshot

### LINKED ISSUE #793: Batch validator holds a shared container for basefees
## Problem

The batch validator holds a `BaseFeeContainer` — a shared, interior-mutable handle into the gas accumulator (`crates/types/src/gas_accumulator.rs:154`, an `Arc<AtomicU64>` whose clones all alias the same atomic) — and reads it live at validation time (`crates/batch-validator/src/validator.rs:188`, pre-change) to check each incoming batch's base fee. This is a data race waiting on cross-epoch validation.

The accumulator is created once for the node's whole lifetime (`crates/node/src/manager/node.rs:308`) and its per-worker base-fee containers persist across epochs (`clear()` zeroes gas totals, not the base fee). The container is mutated at runtime through `set_base_fee` — today at startup catch-up (`crates/node/src/manager/node.rs:159`) and, by design, at every epoch boundary through the EIP-1559 seam in `adjust_base_fees` (`crates/node/src/manager/node/run_epoch.rs:542-548`, currently a no-op carrying the comment "Change this base fee to update base fee in batches"). The validator is recreated each epoch, but the container it points at outlives the epoch.

Nothing synchronizes the epoch-transition path writing a worker's base fee for the *next* epoch against a *current* epoch's validator reading the same atomic to validate an in-flight or gossiped batch. Because the read is a live `Acquire` load of state another task can `Release`-store into concurrently, a batch produced and gossiped under epoch N's base fee can be compared against epoch N+1's value in `validate_basefee` — flipping the `base_fee != expected_base_fee` check from accept to reject (or back). The outcome then depends on cross-task scheduling rather than on whether the batch is actually correct, so validators straddling an epoch boundary can disagree on a batch's validity purely by timing.

This looks cosmetic today only because `adjust_base_fees` is a no-op, so the value rarely changes after startup. The moment base-fee adjustment is wired into that seam, it becomes a live, consensus-affecting bug.

Secondary to the race, carrying a shared container is also heavier than the validator needs and threads a `tn_types::gas_accumulator` type through the engine's validator-construction path (`new_batch_validator` on both `ExecutionNode` and `ExecutionNodeInner`) where a primitive would do, coupling those signatures to the accumulator's internals. The epoch-start path even snapshots the container to a `u64` for the transaction pool and then hands the container separately to the validator — sourcing the same value two different ways for two consumers that both want the epoch's fixed base fee.

grantkee:
* `BatchValidator` now stores a plain `u64` base fee instead of a `BaseFeeContainer`. Base fee is fixed per epoch and the validator is rebuilt each epoch, so a snapshot taken at epoch start is valid for its whole lifetime.
* `new_batch_validator` (on `ExecutionNode` and `ExecutionNodeInner`) takes `u64`; the `BaseFeeContainer` import is gone from the engine modules.
* `start_epoch` snapshots the worker base fee to `u64` once and passes it to both the tx pool and the validator, dropping the redundant container read.
* Added `test_validator_uses_u64_base_fee`: accepts a batch whose base fee matches the snapshot, rejects one that doesn't.

closes #793

### DIFF
--- a/crates/batch-builder/tests/it/build_batches.rs
+++ b/crates/batch-builder/tests/it/build_batches.rs
@@ -16,10 +16,10 @@ use tn_reth::{
 };
 use tn_storage::{open_db, tables::NodeBatchesCache};
 use tn_types::{
-    gas_accumulator::{BaseFeeContainer, GasAccumulator},
-    test_genesis, Address, Batch, BatchValidation, Bytes, Certificate, CertifiedBatch,
-    CommittedSubDag, ConsensusHeaderDigest, ConsensusOutput, Database, Encodable2718,
-    GenesisAccount, ReputationScores, SealedBatch, TaskManager, MIN_PROTOCOL_BASE_FEE, U160, U256,
+    gas_accumulator::GasAccumulator, test_genesis, Address, Batch, BatchValidation, Bytes,
+    Certificate, CertifiedBatch, CommittedSubDag, ConsensusHeaderDigest, ConsensusOutput, Database,
+    Encodable2718, GenesisAccount, ReputationScores, SealedBatch, TaskManager,
+    MIN_PROTOCOL_BASE_FEE, U160, U256,
 };
 use tn_worker::{test_utils::TestMakeBlockQuorumWaiter, Worker, WorkerNetworkHandle};
 use tokio::time::timeout;
@@ -151,13 +151,8 @@ async fn test_make_batch_el_to_cl() {
     let sealed_batch = sealed_batch.unwrap();
 
     // ensure batch validator succeeds
-    let batch_validator = BatchValidator::new(
-        reth_env.clone(),
-        Some(txpool.clone()),
-        0,
-        BaseFeeContainer::default(),
-        0,
-    );
+    let batch_validator =
+        BatchValidator::new(reth_env.clone(), Some(txpool.clone()), 0, MIN_PROTOCOL_BASE_FEE, 0);
 
     let valid_batch_result = batch_validator.validate_batch(sealed_batch.clone());
     assert!(valid_batch_result.is_ok());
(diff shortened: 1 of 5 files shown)


crates/batch-validator/src/validator.rs
```diff
@@ -630,6 +632,36 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn test_validator_uses_u64_base_fee() {
+        // A validator built from a plain u64 base fee accepts a batch whose base fee matches and
+        // rejects one that does not -- exercising the BaseFeeContainer -> u64 conversion.
```
MavenRain: Wasn't this conversion removed?

crates/batch-builder/tests/it/build_batches.rs
```diff
@@ -534,13 +524,8 @@ async fn test_canonical_notification_updates_pool() {
...
     // validate batch
-    let batch_validator = BatchValidator::new(
-        reth_env.clone(),
-        Some(txpool.clone()),
-        0,
-        BaseFeeContainer::default(),
-        0,
-    );
+    let batch_validator =
+        BatchValidator::new(reth_env.clone(), Some(txpool.clone()), 0, MIN_PROTOCOL_BASE_FEE, 0);
```
MavenRain, line 528: Nit:  There are a couple of tests here that repeat this expression.  It might be worth extracting into a helper fixture method.

MavenRain (dismissed):
Just nits

sstanfield (dismissed):
This is good.  Shouldn't you also remove BaseFeeContainer with this?

crates/batch-validator/src/validator.rs
grantkee: The conversion is this PR, but the test was only a slight variation of the preceding `test_invalid_batch_base_fee_for_gas`. I just removed it since it didn't add anything really

grantkee:
> This is good. Shouldn't you also remove BaseFeeContainer with this?

Yes, I have some stacked PRs - but removing `BaseFeeContainer` in this PR creates more change in epoch manager. I will address in a follow up. Thanks for flagging

MavenRain (approved):

PR #796, opened 2026-06-26
feat(node): wire the EIP-1559 base-fee compute seam (inert by default)

### LINKED ISSUE #795: Base-fee adjustment seam is a no-op
# Base fees never adjust — `adjust_base_fees` is a no-op

## Problem

Batch base fees are currently static. `adjust_base_fees` (`crates/node/src/manager/node/run_epoch.rs`) reads the per-worker gas accumulator totals at epoch finalization but never feeds a new base fee back into batch production. Every worker holds whatever base fee it starts with indefinitely, regardless of how full or empty its batches actually are.

Without an adjustment mechanism, the network has no seam for reacting to sustained congestion or underuse — batch producers have no fee signal, and nothing computes anything at the epoch boundary today. Introducing that computation later, once it needs to matter, is riskier than wiring an inert version in now: it gives every validator a chance to run the identical deterministic calculation before its output is allowed to actually move fees.

## Solution

`adjust_base_fees` should call into an EIP-1559-style computation (`compute_next_base_fee_eip1559`, already present in `tn_types::gas_accumulator`) at epoch finalization, deriving each worker's next base fee from its gas accumulator totals for the epoch that just ended.

The per-worker target gas used in this computation should be hardcoded to `u64::MAX` for now. Real, governance-driven targets are out of scope for this change and will land in a follow-up. Against an unreachable target, the adjustment is mathematically inert: the computed fee can only ratchet downward, and it floors at `MIN_PROTOCOL_BASE_FEE`. Since every worker already starts at `MIN`, this seam introduces no observable change in behavior yet — it only wires the computation in so every validator runs the identical, deterministic calculation at each epoch boundary.

This keeps the fee-adjustment plumbing separate from the policy question of what the target gas should be, so the two can be reviewed and merged independently.

grantkee:
- `adjust_base_fees` now calls `compute_next_base_fee_eip1559` per worker using that worker's gas accumulator totals from the epoch just ended, instead of reading the totals and discarding them.
- Target gas is hardcoded to `u64::MAX` for now, which keeps the adjustment inert (fee can only ratchet down and floors at `MIN_PROTOCOL_BASE_FEE`, matching current behavior). Real governance-driven targets are a follow-up.
- Moved `adjust_base_fees` from a method to a free function so it's directly unit-testable, and added tests proving it's a no-op at `MIN` and that it ratchets a non-`MIN` fee down to the floor.

closes #795

### DIFF
--- a/crates/node/src/manager/node/run_epoch.rs
+++ b/crates/node/src/manager/node/run_epoch.rs
@@ -23,9 +23,9 @@ use tn_executor::subscriber::spawn_subscriber;
 use tn_primary::ConsensusBus;
 use tn_storage::{certificate_pack::CertificatePack, tables::OurNodeBatchesCache};
 use tn_types::{
-    gas_accumulator::GasAccumulator, BlsPublicKey, Committee, ConsensusHeaderDigest,
-    ConsensusOutput, Database as TNDatabase, EpochRecord, Notifier, TaskJoinError, TaskManager,
-    TaskSpawner, TnReceiver,
+    gas_accumulator::{compute_next_base_fee_eip1559, GasAccumulator},
+    BlsPublicKey, Committee, ConsensusHeaderDigest, ConsensusOutput, Database as TNDatabase,
+    EpochRecord, Notifier, TaskJoinError, TaskManager, TaskSpawner, TnReceiver,
 };
 use tokio::sync::mpsc;
 use tracing::{error, info, warn};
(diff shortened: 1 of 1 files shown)


grantkee:
This is opened as a stack PR but will be rebased onto main once upstream PR #794 lands

MavenRain (dismissed):
This looks solid.

grantkee:
Merged upstream PR and rebased main - ready for rereview

MavenRain (approved):

PR #797, opened 2026-06-27
feat(network): cut the primary epoch-pack path over to the typed sync protocol with legacy fallback (739, step 6)

MavenRain:
# feat(network): cut the primary epoch-pack path over to the typed sync protocol with legacy fallback (739, step 6)

## Summary

Step 6 of #739. Step 5 (#781) cut the worker batch path over to the typed sync-frame layer; the primary stayed fully legacy and dropped any inbound sync stream with a warning. This PR cuts the primary's **full epoch-pack** transfer over to the sync protocol, with a **penalty-exempt fallback to legacy** for any peer that does not speak it.

The result folds the legacy two-roundtrip handshake (a `StreamEpoch` request-response that returns an `ack`, then a raw `/tn-stream` open onto which the requester writes a 32-byte correlation digest) into a **single stream open** whose opening `SyncFrame::Req(PrimarySyncRequest::EpochPack { epoch })` carries the request itself.

Scope is deliberately the full epoch pack only: `PrimarySyncRequest` has `EpochPack` and `MissingCertificates`, and the latter is step 7. The partial-prefix transfer (`StreamEpochPartial`) and the single consensus-output stream (`StreamConsensusOutput`) have no sync request variant, so they stay on the legacy path here. No `SyncFrame` / `PrimarySyncRequest` variant is added, reordered, or removed, so `frame_tags_are_stable` / `request_tags_are_stable` stay green and the wire format is unchanged.

`request_epoch_pack`'s public signature is preserved, so the state-sync retry loop (`crates/state-sync/src/consensus.rs`) that drives it is untouched.

### Requester (primary `network/mod.rs`)

- `request_epoch_pack_inner` gains a sync-first prelude for the full-epoch case: before the existing legacy loop, it calls `request_epoch_pack_sync`, which enumerates connected peers and, for each peer not cached legacy-only this epoch, opens a `/tn-primary-sync` stream, writes `SyncFrame::Req(PrimarySyncRequest::EpochPack { epoch })`, reads the response (`Ack`, then `Data` frames terminated by `End`), and imports it. It returns as soon as one peer's pack imports.
- **The legacy `send_request_any` ack-plus-digest loop below is byte-for-byte unchanged.** The sync attempt is a pure prelude: on any failure it falls through to the proven legacy path, so a mixed or all-legacy fleet keeps working exactly as before.
- An `EpochPackAttempt` enum classifies each exchange:
  - `Imported` -> the peer served a pack that verified and imported; cache it sync-capable, done.
  - `Unsupported` -> the peer never answered (negotiation failed = a pre-cutover peer, or it negotiated sync but sent no `Ack`); cache it legacy-only and skip its probe for the rest of the epoch.
  - `Failed` -> the exchange failed after the peer proved sync-capable (`Deny` / `Err` / mid-stream error, or a transient I/O error opening or writing the stream); keep it sync-capable and try the next peer.
- A per-peer `sync_capability` cache (`Arc<Mutex<HashMap<BlsPublicKey, bool>>>`) on `PrimaryNetworkHandle` avoids re-probing a known legacy-only peer every request. It is **cleared each epoch** (`clear_sync_capability`, called at the per-epoch primary network setup) so a peer upgraded across the committee-rotation boundary is re-probed.
- The probe is bounded to `MAX_EPOCH_SYNC_PROBES` (3) sync exchanges per call (cached legacy-only peers are skipped for free and do not count), so on an all-legacy fleet the prelude never delays the legacy fetch for long while still discovering newly upgraded peers across successive calls.
- The sync probe is **penalty-exempt**, and a transient transport error never demotes a sync-capable peer: a genuine negotiation failure surfaces as `StreamError::UpgradeFailed` and maps to `Unsupported` (fall back, cache legacy-only), while a transport I/O error during the open surfaces as the distinct `StreamError::UpgradeIo` and maps to `Failed` (try the next peer, peer stays sync-capable). Neither scores the peer.

### Pack data path (primary `network/sync_codec.rs`, new)

Unlike the worker, whose every `Data` frame is one whole decodable `Batch`, an epoch pack is a single opaque byte stream (`EpochMeta` then interleaved batch and consensus records). So:

- The responder chunks the pack file (`get_epoch_stream`) into fixed-size `SyncFrame::Data` frames (256 KiB) terminated by `SyncFrame::End`.
- The requester reassembles those `Data` frames into a contiguous `tokio::io::AsyncRead` (`sync_pack_reader`, a `StreamReader` over the frame stream) and feeds it directly to the existing `ConsensusChain::stream_import`, which verifies and imports the pack incrementally exactly as the legacy path does. Nothing in storage changes.

Because the pack is chunked at a fixed size, the per-frame bound (`MAX_SYNC_PACK_FRAME_SIZE`) is a constant, so unlike the worker the primary handle needs no epoch field.

### Responder (primary `network/mod.rs`, `handler.rs`, `sync_codec.rs`)

- Inbound streams now dispatch on the negotiated `StreamKind`: `Legacy` -> the existing digest-correlation reader; `Sync` -> the new `process_inbound_sync_stream` (it previously dropped the stream with a warning).
- Because the sync request travels in the stream's opening frame (no prior request-response ack), admission against the concurrency caps moves to stream-open time:
  - `try_admit_sync` acquires the **shared** `epoch_stream_semaphore` (so the global `MAX_CONCURRENT_EPOCH_STREAMS` cap is the combined legacy+sync limit) and checks the peer's combined in-flight count (legacy pending entries plus sync streams) against `MAX_PENDING_REQUESTS_PER_PEER`. The legacy admission path (`accept_stream_request`) likewise adds the peer's in-flight sync-stream count to its legacy count, so the per-peer cap is enforced **symmetrically across both paths** (both lock `pending_epoch_requests` before `sync_stream_peers`, so they never deadlock). A `SyncStreamPermit` RAII guard releases the global slot and decrements the per-peer count on drop.
  - A shedding responder writes `SyncFrame::Deny(DenyReason::AtCapacity)` without reading, so the requester gives up immediately and tries elsewhere instead of waiting out its ack timeout.
- Once admitted, the opening `Req` frame is read (bounded by `SYNC_REQUEST_READ_TIMEOUT`). A full epoch pack is all-or-nothing, so if the complete pack is not held the responder sheds with `SyncFrame::Deny(DenyReason::Unavailable)`; otherwise it serves `Ack` + one `Data` frame per pack chunk + `End` via `send_sync_epoch_pack_over_stream`.
- A `MissingCertificates` sync request is declined with `Deny(DenyReason::Unavailable)`: the primary registers the request type but does not serve it until step 7.
- Responder errors are **metrics-only** during the rollout (logged, best-effort `SyncFrame::Err`, no penalty), matching the step-3 stream-failure policy and the legacy responder. Every best-effort trailing write (shed `Deny`, malformed `Err`, internal `Err`, final close) is bounded by a timeout so a non-reading peer cannot stall the responder task on an unbounded write.

### Plumbing

- `PrimaryNetworkHandle::clear_sync_capability` is called at the per-epoch primary network setup (`crates/node/src/manager/node/start_epoch.rs`), the same epoch-boundary refresh the worker uses for its cache.
- `tn-primary` gains the `tokio-util` `io` feature for `StreamReader` (used to present the reassembled pack as an `AsyncRead`).

## Cross-version behaviour

|---|---|---|
| item-6 (sync-first) | item-6 with the pack | sync exchange, single open |
| item-6 | item-6 without the pack | `Deny(Unavailable)` -> requester tries the next peer / falls back; no penalty |
| item-6 | pre-item-6 (no sync open served) | no `Ack` (or negotiation fails) -> requester falls back to legacy; no penalty |
| pre-item-6 (legacy-only) | item-6 | responder still accepts the legacy `/tn-stream` open |

## Wire protocol

- Each `SyncFrame` is one length-prefixed, snappy-compressed `encode_message` unit; the BCS enum discriminant is the frame tag (pinned by `frame_tags_are_stable` / `request_tags_are_stable` from step 4).
- The epoch pack is streamed as `Ack` + a sequence of `Data` frames (each a 256 KiB pack chunk) + `End`. Every frame is bounded by `MAX_SYNC_PACK_FRAME_SIZE` (chunk size + 1 KiB envelope headroom), so a peer cannot force an oversized allocation.

## Tests

- New `sync_codec` round-trip tests proving the responder framing and the requester reassembly agree on the wire across chunk boundaries: `sync_pack_round_trip_multi_chunk` (two full chunks plus a partial), `sync_pack_round_trip_empty`, and `sync_pack_reader_rejects_err_frame` / `sync_pack_reader_rejects_unexpected_frame` (an aborted or out-of-place frame surfaces as a read error so `stream_import` aborts).
- `test_sync_epoch_pack_unavailable_denies`: a responder that does not hold the complete pack sheds with `Deny(Unavailable)` rather than erroring. The `Ack`+`Data`+`End` happy path is exercised end-to-end by the (ignored) observer-pack-import e2e test.

## Gates (all green)

- `cargo +nightly-2026-03-20 fmt --check`
- `cargo test -p tn-primary` (162 passed, 1 ignored)
- `cargo clippy -p tn-primary --all-features --all-targets` (0 warnings; the changed crate and its consumers compile clean)
- `cargo doc --no-deps -p tn-primary` with `RUSTDOCFLAGS=-D warnings`
- `cargo check -p tn-node` (the per-epoch `clear_sync_capability` call site)

BCS encodes enum variant indices: no `SyncFrame` / `PrimarySyncRequest` variant is reordered or removed here, and no new variant is added. The coordinated `/0.0.2` bump and legacy freeze remain step 9.

### DIFF
--- a/crates/consensus/primary/Cargo.toml
+++ b/crates/consensus/primary/Cargo.toml
@@ -24,7 +24,7 @@ tokio = { workspace = true, features = [
     "time",
     "test-util",
 ] }
-tokio-util = { version = "0.7", features = ["compat"] }
+tokio-util = { version = "0.7", features = ["compat", "io"] }
 tracing = { workspace = true }
 tn-network-types = { workspace = true }
 tn-types = { workspace = true }
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -15,7 +15,9 @@ use std::{
     time::Duration,
 };
 use tn_config::ConsensusConfig;
-use tn_network_libp2p::{GossipMessage, Stream};
+use tn_network_libp2p::{
+    write_frame, GossipMessage, PrimarySyncRequest, Stream, SyncFrame, SyncFrameError,
+};
 use tn_storage::{consensus::ConsensusChain, CertificateStore, VoteDigestStore};
 use tn_types::{
     ensure,
@@ -32,6 +34,13 @@ use tracing::{debug, error, info, warn};
 /// Prevents slow-reader attacks where a peer accepts a stream but never reads.
 /// Set to an arbitrary 10 seconds to read 16kb buffer.
 const SEND_STREAM_BUFFER_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Total timeout for serving an entire epoch pack over the sync protocol.
+///
+/// A backstop above the per-frame [`SEND_STREAM_BUFFER_TIMEOUT`]: a peer that
+/// drip-reads just fast enough to keep resetting the per-frame timer cannot pin
+/// the responder task (and the admission slot it holds) indefinitely.
+const SEND_SYNC_PACK_TIMEOUT: Duration = Duration::from_secs(200);
 
 /// Map to hold vote info to detect invalid votes, equivocation and cache responses in case of
 /// rerequests.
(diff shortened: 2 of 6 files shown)


grantkee (approved):
LGTM!

sstanfield (approved):
I think this is good.  It is possible that 200s might be tight for full download of a really busy pack file.  Currently we are not anywhere near that being an issue so I am OK with leaving for now and we monitor things at higher traffic levels.

Issue #712, opened 2026-06-03
Worker Gateway: stateless public-facing RPC proxy for worker endpoints


grantkee:
## Summary

This is a large undertaking and covers multiple PRs. Please confirm plans with a maintainer before taking on this issue.

Build a standalone, stateless reverse proxy — the **worker gateway** — that sits between the public internet and a node's worker JSON-RPC endpoints. The gateway absorbs hostile and bursty traffic at the edge, enforces rate limits and request hygiene before anything reaches a worker, and scales horizontally with demand (e.g. via a Kubernetes HPA). Node operators run N gateway replicas behind a load balancer; workers are never exposed directly.

The design must accommodate multiple workers per node, but v1 only needs to support a single worker (worker id `0`), matching what nodes run today.

## Motivation

Today, a worker's RPC endpoint is the standard Ethereum JSON-RPC server (jsonrpsee) built per-worker via `RethEnv::get_rpc_server()` / `start_rpc()` (`crates/tn-reth/src/lib.rs`). Its only protections are transport-level defaults configured in `crates/tn-reth/src/rpc_server_args.rs`:

- a 500-connection cap (`RPC_DEFAULT_MAX_CONNECTIONS`)
- a 15 MB max request size (`RPC_DEFAULT_MAX_REQUEST_SIZE_MB`)

There is **no per-IP rate limiting, no global throttling, and no application-level request filtering**. Any operator who exposes a worker RPC port publicly is one `while true; do curl ...` away from saturating it. Operators currently have to bring their own edge protection (nginx, cloud WAFs, etc.), each hand-rolled and inconsistent.

A first-party gateway gives every operator the same hardened, reproducible edge layer:

- **DoS absorption** — abusive traffic terminates at cheap, disposable, autoscaling gateway pods instead of at the worker, which has consensus responsibilities.
- **Operator security** — workers bind to private interfaces only; the gateway is the sole public surface.
- **Elastic capacity** — gateways are stateless, so replicas scale up and down with traffic without coordination.

## Current architecture (orientation for contributors)

- **Worker RPC**: each worker gets a full Ethereum JSON-RPC server (`eth_*`, `net_*`, `web3_*`, plus `tn_*` extensions). Server construction: `crates/tn-reth/src/lib.rs` (`get_rpc_server`, `start_rpc`); CLI/config surface: `crates/tn-reth/src/rpc_server_args.rs`. Defaults: HTTP on `127.0.0.1:8545`, WS on `8546`.
[...]

PR #798, opened 2026-06-28
feat(node): per-worker readiness endpoint at GET /health/workers (#712)

MavenRain:
## Summary

First PR of the worker-gateway effort (#712). It adds the node-side readiness
contract that the gateway will poll. This PR is deliberately small and
self-contained: it defines the wire contract the remaining three PRs build
against, so it is worth landing (and reviewing the contract) first.

No new runtime behavior is enabled by default: the healthcheck service is still
off unless `--healthcheck <PORT>` is set, and the existing liveness response is
unchanged.

## What it does

- Adds a `GET /health/workers` route to the existing healthcheck server
  (`crates/node/src/health.rs`), served on the **same port** as the liveness
  check. It returns a versioned JSON envelope reporting, per worker, whether the
  worker is accepting transactions.
- Liveness is preserved: any path other than `/health/workers` still returns the
  fixed `200 OK` / `OK`, exactly as before.

### Wire contract (v1)

```json
{ "version": 1, "workers": [ { "worker_id": 0, "accepting_transactions": true } ] }
```

The envelope is versioned and the `workers` field is a list from day one, so the
method-aware routing follow-up can add per-worker fields (epoch, sync, queue
depth) and additional workers without breaking the gateway's parser. v1 reports
exactly one entry, worker `0`, matching what nodes run today.

## Semantics of `accepting_transactions` (please confirm)

Worth a careful look, because the ground truth shifted under the original
proposal. The proposal said this flag should "flip false after `close_epoch()`
clears the accumulator." That framing was based on the pre-`#737` epoch-manager
layout. After the "Breakdown epoch manager (#737)" refactor, the code says
something different:

- The worker's **RPC server and transaction pool accept transactions
  continuously**. They are initialized once and are **not** torn down across
  epoch transitions.
- The only epoch-scoped machinery is the **batch builder**, whose lifetime lives
  in a local `epoch_task_manager` (`run_epoch.rs`) and is aborted at each
  boundary. Nothing observable (no field, atomic, or watch) mirrors "epoch
  active," and the pool keeps accepting across the transition (txs queue for the
  next epoch).
- Worker-side signals are not usable: `eth_syncing` is hardcoded `false` on
  `WorkerNetwork`, and the pool has no "paused/accepting" flag.

So v1 defines `accepting_transactions` as **"this worker's components are
initialized"**:

- `false` during startup, before the node has brought up the worker's RPC + pool;
- `true` once they are up (and it stays true, since they are not torn down);
- a node that is down answers no request at all, which the polling gateway treats
  as not-ready (fail-closed).

This is the cheapest, most honest signal given the current code, and it carries
real startup information. If you would rather have readiness also drop during
epoch transitions (the issue's "readiness lost on epoch transition" wording),
that is a clean follow-up: it means adding a per-worker flag at the `run_epoch`
open/close seams. I left it out of PR 1 because it is new instrumentation, a
larger blast radius, and arguably misreports (the pool is still queuing txs
during a transition). Happy to add it if you want the stricter semantics.

## Implementation notes

- New `ExecutionNode::is_worker_initialized(worker_id)` accessor, mirroring the
  existing `are_workers_initialized()`.
- The healthcheck accept loop now reads the request to route by path. It
  previously read nothing and replied `200` to any bytes, so routing required
  adding a read. Both the read and the readiness probe are bounded by timeouts
  (2s and 1s); on timeout the worker is reported not-ready (fail-closed), so a
  slow client or a lock held during an epoch transition cannot stall the
  single-threaded accept loop. The route ignores any query string.
- `HealthcheckServer::spawn` takes a readiness probe closure rather than the
  engine handle, keeping the server decoupled from the engine and unit testable;
  the call site in `manager/node.rs` captures `engine.clone()` and reads
  `is_worker_initialized` per request.
- Adds `serde` + `serde_json` to the `tn-node` crate (it had no serde before).
- Drive-by doc fix: the flag is `--healthcheck`, not `--enable-healthcheck` (the
  old health.rs doc comment was stale).

## Out of scope (later PRs)

- PR 2: the `bin/worker-gateway/` crate skeleton + proxy core.
- PR 3: edge protections (rate limits, request-size limits, tx sanity checks).
- PR 4: observability + deployment (Prometheus, Dockerfile, k8s/HPA examples).

## Testing

- `test_health_workers_reports_ready` / `test_health_workers_reports_not_ready`:
  end-to-end over a TCP socket with a trivial readiness probe; assert the `200`,
  the `application/json` content type, and the exact JSON body for both states.
- `test_readiness_payload_contract`: locks the exact wire bytes (snake_case keys,
  `version: 1`, single worker) for both the true and false cases.
- `test_request_path_parsing`: unit test for the request-line path parser,
  including a query-string case.
- `test_tcp_healthcheck`: the existing liveness test, updated for the new spawn
  signature; still asserts the unchanged liveness `200 OK` / `OK`.

## Open questions for the maintainer

1. Endpoint path: I went with `GET /health/workers` on the liveness port, as
   proposed. Flag if you would prefer a different path or a dedicated port.
2. `accepting_transactions` semantics: confirm the init-based definition above,
   or ask for the epoch-gated variant (follow-up).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11704,6 +11704,8 @@ dependencies = [
  "rand 0.9.2",
  "reth-metrics",
  "roaring 0.10.12",
+ "serde",
+ "serde_json",
  "state-sync",
  "tempfile",
  "thiserror 1.0.69",
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -35,6 +35,8 @@ jsonrpsee = { workspace = true }
 roaring = { workspace = true }
 metrics = { workspace = true }
 reth-metrics = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
 
 [dev-dependencies]
 metrics-util = { workspace = true }
--- a/crates/node/src/engine/mod.rs
+++ b/crates/node/src/engine/mod.rs
@@ -111,6 +111,16 @@ impl ExecutionNode {
         !self.internal.read().await.workers.is_empty()
     }
 
+    /// Returns true if the worker identified by `worker_id` has been initialized.
+    ///
+    /// A worker's components (RPC server + transaction pool) are created once on
+    /// the node's first epoch and are not torn down across epoch transitions, so
+    /// this reflects "this worker is up and accepting transactions" rather than
+    /// any per-epoch state. Backs the `/health/workers` readiness endpoint.
+    pub async fn is_worker_initialized(&self, worker_id: WorkerId) -> bool {
+        self.internal.read().await.workers.get(worker_id as usize).is_some()
+    }
+
     /// Batch maker
     pub async fn start_batch_builder(
         &self,
--- a/crates/node/src/health.rs
+++ b/crates/node/src/health.rs
@@ -1,30 +1,120 @@
-//! Simple TCP healthcheck endpoint for monitoring service availability.
+//! Simple TCP health/readiness endpoints for monitoring service availability.
 //!
-//! Implements a minimal HTTP/1.1 server that responds with status 200 to all requests.
-//! This is designed for integration with GCP load balancers and similar health monitoring systems.
+//! Implements a minimal HTTP/1.1 server with two routes on a single port:
+//! - any path except `/health/workers` -> liveness: a fixed `200 OK` (the process is up), matching
+//!   the original unconditional behavior.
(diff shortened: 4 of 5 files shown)


sstanfield (approved):
LGTM!

grantkee (approved):
LGTM!

PR #806, opened 2026-06-29
fix(worker): snapshot requested digests so synchronize() accepts fetched batches

### LINKED ISSUE #800: `synchronize()` always errors once any batch must be fetched
- File: `crates/consensus/worker/src/network/primary.rs:55-103`
- Category: logic-edge-case

**Root cause.** `network.request_batches(&mut missing)` drains `missing` as it fetches. The response loop then does `if missing.remove(&digest) { insert } else { return Err("received a batch we did not request") }` (line 78) but `missing` is already empty, so `remove` returns false for every fetched batch and the function returns the spurious "did not request" error.

**Impact.** Worker batch synchronization is broken on every path that actually fetches a batch (the common case). Sync silently fails.

**Patch [precise].** Validate responses against a snapshot taken before the drain; the existing final `missing.is_empty()` check still confirms completeness:
```rust
// snapshot before request_batches drains `missing`
let originally_missing = missing.clone();
let response = network.request_batches(&mut missing).await?;
// ... build sealed_batches_from_response, open tx ...
for sealed_batch in sealed_batches_from_response.into_iter() {
    if !message.is_certified {
        if let Err(err) = self.validator.validate_batch(sealed_batch.clone()) {
            return Err(eyre::eyre!("Invalid batch: {err}"));
        }
    }
    let (mut batch, digest) = sealed_batch.split();
    if originally_missing.contains(&digest) {     // was: missing.remove(&digest)
        batch.set_received_at(now());
        tx.insert::<NodeBatchesCache>(&digest, &batch).map_err(/* ... */)?;
    } else {
        return Err(eyre::eyre!(format!(
            "received a batch {digest} we did not request!"
        )));
    }
}
tx.commit().map_err(/* ... */)?;
// `missing` now holds only digests request_batches could not fetch
if missing.is_empty() { Ok(()) } else { Err(eyre::eyre!("failed to synchronize batches!")) }
```

MavenRain:
## Summary

`PrimaryReceiverHandler::synchronize()` returned an error on every call that had
to fetch a batch from a peer, so worker batch synchronization failed on its
typical path (the failure was also silent to the caller beyond the returned
error).

Root cause: `WorkerNetworkHandle::request_batches(&mut missing)` removes each
successfully fetched digest from the `missing` set as it goes and returns the
fetched batches. `synchronize()` then iterated the returned batches and called
`missing.remove(&digest)` to confirm each one had been requested. Those digests
were already drained from `missing`, so the check returned `false` for every
batch and the function bailed out with "received a batch we did not request!".

## Fix

- Snapshot the requested set (`requested = missing.clone()`) before
  `request_batches` drains it, and check membership against that snapshot.
- Extract the post-fetch reconciliation (validate, verify-requested, persist,
  verify-completeness) into `store_synced_batches`. The final completeness check
  still runs against the drained `missing` set, which now correctly holds only
  the digests that could not be fetched.

No public API changes; the membership check is now defensive (after the fix,
`request_batches` already guarantees it only returns requested digests).

## Tests

The store path had no coverage, which is how this slipped through. Added unit
tests for `store_synced_batches`:

- `store_synced_batches_happy_path_stores_all`: all requested-and-fetched batches
  are validated and persisted, and the sync succeeds. This is the direct
  regression test for #800.
- `store_synced_batches_incomplete_fetch_errors`: a non-empty unfetched set fails
  the sync while still committing the batches that did arrive.
- `store_synced_batches_unrequested_batch_errors`: a returned batch that was never
  requested is rejected.

## Verification

- `cargo +nightly-2026-03-20 fmt`: clean
- `cargo +nightly-2026-03-20 clippy -p tn-worker --all-targets --all-features`: 0 errors, 0 warnings
- `cargo test -p tn-worker --lib`: 3 new tests pass

Closes #800.

### DIFF
--- a/crates/consensus/worker/src/network/primary.rs
+++ b/crates/consensus/worker/src/network/primary.rs
@@ -52,8 +52,44 @@ impl<DB: Database> PrimaryToWorkerClient for PrimaryReceiverHandler<DB> {
             return Ok(());
         }
 
+        // `request_batches` removes each successfully fetched digest from `missing` as it goes,
+        // so snapshot the originally requested set before the call. Afterwards `missing` holds
+        // only the digests that could not be fetched.
+        let requested = missing.clone();
         let response = network.request_batches(&mut missing).await?;
 
+        self.store_synced_batches(response, &requested, &missing, message.is_certified)
+    }
+
+    async fn fetch_batches(
+        &self,
+        digests: BTreeSet<BlockHash>,
+    ) -> eyre::Result<HashMap<BlockHash, Batch>> {
+        Ok(self.batch_fetcher.fetch_for_primary(digests).await?)
+    }
+}
+
+impl<DB: Database> PrimaryReceiverHandler<DB> {
+    /// Validate and persist the batches returned by
+    /// [`WorkerNetworkHandle::request_batches`].
+    ///
+    /// `requested` is the set of digests originally requested, snapshotted *before* the fetch
+    /// (which drains it), and `unfetched` is whatever digests remained afterwards. Each returned
+    /// batch is validated (unless it belongs to a certificate), checked against `requested`, and
+    /// written in a single transaction.
+    ///
+    /// Membership is checked against `requested` rather than the live set because
+    /// `request_batches` already removes every fetched digest from the set it is given;
+    /// re-checking the drained set here would spuriously reject every batch. Returns an error if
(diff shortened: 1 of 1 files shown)


sstanfield (approved):
Good catch.  Not sure how this ever worked right, it probably forced unnecessary retries and eventually everything would be local...

PR #807, opened 2026-06-30
fix: evict oldest bans/disconnects on overflow, not newest (#799)

### LINKED ISSUE #799: Ban/disconnect overflow pruning evicts the NEWEST entries (ban evasion)
- File: `crates/network-libp2p/src/peers/all_peers.rs:999-1029` (`collect_excess_peers`), called by `prune_banned_peers`/`prune_disconnected_peers`
- Category: peer-scoring

**Root cause.** The heap stores `(Reverse(instant), ...)`. `BinaryHeap` is a max-heap, so with `Reverse` the top is the *oldest* instant, and the replace condition `entry.0 < current_max.0` swaps in *newer* candidates. Net result: the heap retains the `excess` NEWEST peers, and those are exactly what the prune functions evict. The doc comments say "oldest first" but the `Reverse` + `<` combination does the opposite.

**Impact.** When `banned_peers.total()` exceeds `max_banned_peers`, the freshest bans are dropped first while stale bans are kept, so a sybil attacker who overflows the ban set gets their just-issued ban evicted and can reconnect. Ban evasion / weakened DoS protection.

**Patch [precise].** Drop the `Reverse` wrapper so it is a max-heap by `instant` (top = newest), and the existing `<` comparison then retains the OLDEST `excess`:
```rust
fn collect_excess_peers<F>(
    &self,
    excess: usize,
    filter: F,
) -> BinaryHeap<(Instant, PeerIdentity, Vec<IpAddr>)>   // was Reverse<Instant>
where
    F: Fn(&ConnectionStatus) -> Option<Instant>,
{
    let mut excess_peers = BinaryHeap::with_capacity(excess);
    for (id, peer) in &self.peers {
        if let Some(instant) = filter(peer.connection_status()) {
            // max-heap by instant: top is the NEWEST collected peer
            let entry = (instant, *id, peer.known_ip_addresses().collect::<Vec<_>>());
            if excess_peers.len() < excess {
                excess_peers.push(entry);
            } else if let Some(current_newest) = excess_peers.peek() {
                // retain the OLDEST `excess`: replace the newest-in-heap when an
                // older candidate appears, so recent bans survive and stale ones
                // are evicted
                if entry.0 < current_newest.0 {
                    excess_peers.pop();
                    excess_peers.push(entry);
                }
            }
        }
    }
    excess_peers
}
```
Callers iterate `for (_, id, ips) in excess_peers`, so they are unaffected. Remove the now-unused `std::cmp::Reverse` import.

MavenRain:
## Summary

`collect_excess_peers` (`crates/network-libp2p/src/peers/all_peers.rs`) is used by
`prune_banned_peers` and `prune_disconnected_peers` to choose which peers to evict when the
banned/disconnected set grows past its cap.  It built a `BinaryHeap` keyed by
`Reverse(instant)`. Because `BinaryHeap` is a max-heap, the `Reverse` wrapper made the heap top
the *oldest* instant, and the replace condition `entry.0 < current_max.0` then swapped in
*newer* candidates.  Net effect: the heap retained the `excess` **newest** peers, and the callers
evict exactly the heap's contents, so the freshest bans were dropped first while stale bans were
kept.

Impact: when `banned_peers.total()` exceeds `max_banned_peers`, a sybil/overflow attacker who
floods the ban set gets their just-issued ban evicted and can reconnect (ban evasion / weakened
DoS protection).  The doc comments said "oldest first", but the `Reverse` + `<` combination did the
opposite.

## Fix

- Drop the `Reverse` wrapper so the heap is an ordinary max-heap keyed by `Instant` (top = the
  newest collected peer).  The existing `entry.0 < current_newest.0` replace condition then retains
  the `excess` **oldest** peers, which the callers evict, keeping the freshest bans/disconnects.
- Remove the now-unused `std::cmp::Reverse` import.
- Rewrite the misleading doc/inline comments to describe the actual (max-heap, retain-oldest)
  behavior.

The two callers iterate the returned heap order-independently (`for (_, id, ips) in ...`), so the
element-type change from `Reverse<Instant>` to `Instant` does not affect them.

## Tests

- `test_prune_banned_evicts_oldest_not_newest` (regression guard for #799): overflows the banned
  set with peers of known, distinct ages and asserts that the `excess` **oldest** bans are pruned
  while every newer ban survives.
- `test_prune_disconnected_evicts_oldest_not_newest`: the same property for the disconnected
  caller, which shares `collect_excess_peers`.

Both new tests **fail on the pre-fix code** with "recent ban (rank 1) must survive pruning" and
pass after the fix.  The pre-existing `test_pruning_logic` only checked the pruned *count*, never
*which* peers were pruned, which is why the inverted direction shipped undetected.

## Checklist

- [x] `collect_excess_peers` retains the oldest `excess` peers; callers evict the oldest, keep the newest
- [x] unused `cmp::Reverse` import removed
- [x] doc/inline comments corrected
- [x] regression tests added (fail before, pass after)
- [x] `cargo +nightly-2026-03-20 fmt` clean
- [x] `cargo +nightly-2026-03-20 clippy -p tn-network-libp2p --all-features --all-targets` clean (0 warnings)
- [x] `tn-network-libp2p` prune tests pass

Closes #799.

### DIFF
--- a/crates/network-libp2p/src/peers/all_peers.rs
+++ b/crates/network-libp2p/src/peers/all_peers.rs
@@ -20,7 +20,6 @@ use crate::{
 use libp2p::{Multiaddr, PeerId};
 use rand::seq::SliceRandom as _;
 use std::{
-    cmp::Reverse,
     collections::{BinaryHeap, HashMap, HashSet},
     net::IpAddr,
     time::{Duration, Instant},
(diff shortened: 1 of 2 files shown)


grantkee (approved):
LGTM!

PR #808, opened 2026-06-30
refactor: replace hand-rolled Future impls with async fns

MavenRain:
## What

Replaces the three hand-rolled `impl Future` event loops with `async fn run`, per the standup decision to prefer async fns over hand-written Futures.

| Loop | Before | After |
|------|--------|-------|
| `ExecutorEngine` (`crates/engine`) | `impl Future` + manual `poll` | `async fn run` + `tokio::select!` |
| `BatchBuilder` (`crates/batch-builder`) | `impl Future` + manual `poll` | `async fn run` + `tokio::select!` |
| `TnExExManager` (`crates/exex`) | `impl Future` + manual `poll` | `async fn run` folding `select_with_strategy(..).try_fold(..)` |

## Details

- **ExecutorEngine / BatchBuilder** become `loop { tokio::select! { biased; .. } }`. `biased` preserves the original poll order (poll consensus / state-changes first to keep the broadcast stream from lagging); the outer loop preserves the greedy non-parking drain.  `BatchBuilder` keeps its post-quorum-failure backoff via a `defer_build` flag and gates the delay interval on `pending_task.is_none()` to match the original (which only polled the interval while idle).
- **TnExExManager** becomes a stream router: the three notification sources and the per-ExEx event receivers are merged via `select_with_strategy([certs, consensus, canon, select_all(event_rxs)], |_| PollNext::Left)` and folded over a `Router`.  The constant left-bias reproduces the original certs > consensus > canon > events priority and the drain-higher-fully-before-next ordering.  `ExExHandle` is split so the event receivers feed `select_all` while the notification senders stay in the fold accumulator for fan-out;  `poll_events`' `while let` drain becomes a per-event fold (the min-height computation was already a `try_fold`).
- **Cargo**: `tokio`'s `macros` feature added to `tn-engine` and `tn-batch-builder` (they use `tokio::select!`). `tn-exex` needs no new feature . . . it uses the `futures` stream router.

## Behavior

Faithful to the originals (each loop was adversarially reviewed against its `poll` body):  poll ordering, shutdown-drain, single-build gating, receiver restore-on-pending, broadcast `Lagged` surfacing, canonical-gap detection, and min-finished-height are preserved.  Known observability/timing deltas:

- `ExecutorEngine`: under a continuously-ready consensus stream `queued_outputs` may read marginally higher;  the closed-stream error logs once (not per poll).  The `if !consensus_closed` arm guard is required to avoid a busy-spin once the stream closes.
- `TnExExManager`: min-finished-height is recomputed per event rather than once per wake.  Benign: the watch receiver is currently dropped at the construction site.

## Test plan

- [x] build check: `tn-engine`, `tn-batch-builder`, `tn-exex`, `tn-node`
- [x] clippy with `--tests` (0 warnings)
- [x] nightly `fmt`
- [ ] run integration tests (they compile; not yet executed)

### DIFF
--- a/crates/engine/src/lib.rs
+++ b/crates/engine/src/lib.rs
@@ -12,14 +12,9 @@ mod payload_builder;
 use crate::metrics::ENGINE_METRICS;
 use error::EngineResult;
 pub use error::TnEngineError;
-use futures::{Future, StreamExt};
-use futures_util::FutureExt;
+use futures::{future::OptionFuture, StreamExt};
 pub use payload_builder::execute_consensus_output;
-use std::{
-    collections::VecDeque,
-    pin::{pin, Pin},
-    task::{Context, Poll},
-};
+use std::collections::VecDeque;
 use tn_reth::{payload::BuildArguments, RethEnv};
 use tn_types::{
     gas_accumulator::GasAccumulator, ConsensusOutput, EngineUpdate, Noticer, SealedHeader,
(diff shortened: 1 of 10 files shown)


crates/engine/src/lib.rs
```diff
@@ -189,97 +184,92 @@ impl ExecutorEngine {
...
-                    // only return if there are no current tasks and the queue is empty
-                    // otherwise, let the loop continue so any remaining tasks and queued output is
-                    // executed
-                    if this.pending_task.is_none() && this.queued.is_empty() {
-                        return Poll::Ready(Err(TnEngineError::ConsensusOutputStreamClosed));
-                    }
+            // check for shutdown signal
+            if self.rx_shutdown.noticed() {
+                info!(target: "engine", "received shutdown signal...");
+                // only return if there are no current tasks and the queue is empty
```
sstanfield, line 198: I think this is OK, but the containing task manager will probably not allow this a lot of time to clean up.  I think this is probably on the node level task manager, we might want to check or extend its shutdown time if we want to make sure tasks finish.  Would need to review all this though.

sstanfield (approved):
I am approving but I will spend some more time with exex/manager.  Also, Grant just merged that so he will need to review it for sure.  But in general I like removing the future implementations.

crates/engine/src/lib.rs
grantkee: I think this mirrors the original behavior, but worth a deeper dive. The restart flow should handle this (just more work in the unhappy path)

grantkee (approved):
Everything looks correct to me. Good work @MavenRain !

PR #810, opened 2026-06-30
feat(network): drop the legacy full-pack sync fallback (739)

MavenRain:
## What
  Makes the primary full epoch-pack fetch typed-only. `request_epoch_pack` now
  returns `Err` (the state-sync caller already retries) when no connected peer
  serves a pack over `/tn-primary-sync`, instead of falling back to the legacy
  `StreamEpoch` stream path.  Follows #797 (739 step 6), which cut the full-pack
  path to the typed protocol *with* a legacy fallback.

  ## Why
  The typed protocol already covers the full epoch pack, so the legacy fallback
  only lets un-upgraded peers keep syncing.  Removing it forces the upgrade, per
  the #797 discussion.

  ## Scope: narrow on purpose
  Typed `PrimarySyncRequest` has only `EpochPack` and `MissingCertificates`
  variants today, with no partial-prefix or single-output variant. So this
  removes only the full-pack legacy fallback and intentionally keeps:
  - Partial-prefix (`request_partial_epoch_pack` / `StreamEpochPartial`) on legacy.
  - Single consensus-output (`StreamConsensusOutput`) on legacy.
  - The legacy `StreamEpoch` responder, one release for backward compat (removing
    the request variant shifts BCS discriminants pinned by
  `request_tags_are_stable`, which needs a version bump, follow-up).

  A full "delete all legacy" today would regress partial-prefix and single-output
  sync. The `sync_capability` probe-dedup cache and `EpochPackAttempt::Unsupported`
  are retained; only stale "falls back to legacy" comments were reworded.

  ## Rollout precondition (operational, not code)
  Full-pack no longer falls back, so a node reachable only to un-upgraded peers
  hard-fails full-sync (bounded retry + requeue, logs "WE ARE STUCK", which is the
  intended force-upgrade behavior, not a hang).  The kept `StreamEpoch` responder
  protects old requesters fetching from upgraded nodes, but not a new node
  bootstrapping off an all-old fleet.  Upgrade seed/bootstrap peers first.

  ## Follow-ups
  1. Add typed `PrimarySyncRequest` partial-prefix + single-output variants; port
     those paths off legacy.
  2. Retire the legacy `StreamEpoch` responder + stream path (with a protocol-version bump).

  ## Testing
  clippy + fmt clean. Preserved-path regression tests pass
  (`test_send_partial_epoch_over_stream`,
  `test_send_consensus_output_over_stream`,
  `test_sync_epoch_pack_unavailable_denies`,
  `test_pending_epoch_stream_replacement_preserves_created_at`).  The new requester
  hard-fail behavior has no unit test (that branching path was untested pre-change;
  needs peer/stream mocking), flagged as follow-up.  The ignored e2e `test_epoch_sync`
  covers the typed full-pack path.

### DIFF
--- a/crates/consensus/primary/src/network/mod.rs
+++ b/crates/consensus/primary/src/network/mod.rs
@@ -85,7 +85,8 @@ const CONSENSUS_OUTPUT_STREAM_READ_TIMEOUT: Duration = Duration::from_secs(10);
 /// Timeout for the responder's first sync frame (`Ack`/`Deny`) after the epoch-pack
 /// request frame is written. A peer that negotiated the sync protocol but does not
 /// answer (a pre-cutover node that registered the protocol but reads the stream on
-/// the legacy digest path) trips this and the requester falls back to legacy.
+/// the legacy digest path) trips this and is cached unsyncable this epoch, so the
+/// probe moves on to the next peer.
 const SYNC_ACK_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Timeout for reading the opening request frame of an inbound sync stream, and the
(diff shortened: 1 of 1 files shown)


grantkee (approved):
LGTM!

sstanfield (approved):

PR #811, opened 2026-06-30
feat(network): cut the primary MissingCertificates path fully over to the typed sync protocol (739, step 7)

MavenRain:
## Summary

Step 7 of #739. Step 6 (#797) cut the primary's full epoch-pack transfer over to the typed sync-frame layer and declined an inbound `MissingCertificates` sync request with `Deny(Unavailable)`. This PR serves `MissingCertificates` over the `/tn-primary-sync` stream (`Ack` + incremental certificate `Data` frames + `End`) and **cuts the certificate fetch fully over to it . . . there is no legacy request-response fallback**.

The legacy `MissingCertificates` exchange was a single request-response round trip whose reply was one `RequestedCertificates` message, capped at the codec's fixed `max_response_size`.  After this PR:

- The **requester** fetches certificates over `/tn-primary-sync` only.  A peer that does not serve it is a failed peer; the staggered fan-out tries the next one.
- The **responder** no longer serves the legacy request-response `MissingCertificates`;  it answers a legacy request with a typed error.  Certificates are served only over sync.

Per the #739 rollout rule, **no BCS variant is removed**: `PrimaryRequest::MissingCertificates` and `PrimaryResponse::RequestedCertificates` are retained (variant indices ride on the wire) until the coordinated `/0.0.2` bump and legacy freeze in step 9.  No `SyncFrame` / `PrimarySyncRequest` variant is added, reordered, or removed (`frame_tags_are_stable` / `request_tags_are_stable` stay green).

### Requester: sync only (`network/mod.rs`)

- `PrimaryNetworkHandle::fetch_certificates` now opens a `/tn-primary-sync` stream, writes `Req(MissingCertificates { exclusive_lower_bound, skip_rounds })`, reads the `Ack`, and reassembles the streamed certificate `Data` frames via `read_sync_certificates`. Any failure (negotiation, `Deny`, transport, protocol, or timeout) returns an error so the staggered fan-out tries the next peer. The exchange is penalty-exempt; cancelling the future (the fan-out's cancel-on-first-success) drops the in-flight substream, resetting it.
- The post-`Ack` read is bounded by an overall `MISSING_CERTS_RESPONSE_TIMEOUT` (30 s) on top of the per-frame timeout, so a peer that `Ack`s then drips frames cannot pin the read loop and its substream (a slowloris the byte cap alone would not catch). This was surfaced by the pre-merge adversarial review.
- Removed: the legacy `fetch_certificates_legacy` request-response path, the per-peer `missing_certs_sync_capability` cache, and the `MissingCertsAttempt` classifier (with no fallback there is no Unsupported/Failed distinction — every non-served outcome is just an error).

### Responder: typed error on the legacy path (`network/mod.rs`, `network/handler.rs`)

- `process_request_for_missing_certs` answers a legacy request-response `MissingCertificates` with `PrimaryResponse::Error` ("served over the sync protocol only") instead of serving it.  `RequestHandler::retrieve_missing_certs` (the legacy serve) is removed.
- The sync responder (`process_sync_missing_certs_stream`, added here) is unchanged in shape from the step-6 epoch-pack responder: it admits via the shared `try_admit_sync` caps, builds a streaming `CertificateCollector`, and streams `Ack` + batched cert `Data` frames + `End` via `send_sync_certificates_over_stream`.  Errors are metrics-only during the rollout (logged, best-effort `SyncFrame::Err`, no penalty); a malformed `skip_rounds` is declined with `Err(Malformed)` before any `Ack`.

### Collector: streaming only (`state_sync/cert_collector.rs`)

- `CertificateCollector` is now constructed solely for streaming (`new(exclusive_lower_bound, skip_rounds, config)`).  The aggregate response-size cap existed only to fit the request-response codec; with the legacy serve gone it is removed, along with `MESSAGE_OVERHEAD` and the per-byte accounting.  The collector is bounded by the per-request DB-read time limit and the rounds in storage; the caller frames and byte-caps the wire response.

### Frame layer (`network/sync_codec.rs`)

- `send_sync_certificates_over_stream` (responder) batches certificates by encoded size (256 KiB target) into `Data` frames (each a BCS `Vec<Certificate>`); `read_sync_certificates` (requester) reassembles them via `try_decode` (a fallible decode of untrusted bytes).  Bounds: `MAX_SYNC_CERT_FRAME_SIZE` (512 KiB per-frame, both sides), `MAX_SYNC_MISSING_CERTS_RESPONSE_BYTES` (64 MiB honest responder cap; the requester accepts one frame beyond it and then errors).  The opening `Req` read in `process_inbound_sync_stream` uses `MAX_SYNC_REQUEST_FRAME_SIZE` (4 MiB) so a request carrying a skip-round bitmap per authority is admitted.

### Testability seam (`certificate_fetcher.rs`)

- Item 7 fetches certificates over a libp2p stream, which cannot be constructed in a unit test.  To keep the fetcher's orchestration (staggered fan-out, retries, cancel-on-first-success, validation) unit-tested, the per-peer fetch is abstracted behind a `MissingCertFetcher` trait (static dispatch, `Clone + Send + Sync + 'static`); `CertificateFetcher<DB, F = PrimaryNetworkHandle>` is generic over it.  Production passes the network handle; tests inject a `MockFetcher` that forwards `(peer, request, reply)` to the test . . . the same recv→assert→reply rhythm the command channel gave before, one level up. `certificate_fetcher.rs` no longer sets `max_response_size` (the sync path has no aggregate cap).

## Cross-version behaviour

|---|---|---|
| step 7 | step 7 | sync exchange, single open, streamed certs |
| step 7 | pre-step-7 | the sync open fails (or no `Ack`) → the fetch fails for that peer; the fan-out tries the next |
| pre-step-7 | step 7 | the legacy request-response `MissingCertificates` is answered with a typed error (not served) |

This is a deliberate, sharper cutover than steps 5/6 (which kept a per-peer legacy fallback).  It assumes the committee is on step 7+;  a peer that cannot speak sync cannot exchange certificates with us over this path.  The coordinated legacy freeze and `/0.0.2` bump remain step 9.

## Tests

- `sync_codec` cert round-trips: `sync_certs_round_trip_single_batch`, `sync_certs_round_trip_empty`, `sync_certs_round_trip_multi_batch` (spans multiple `Data` frames), and `read_sync_certificates_rejects_err_frame` / `read_sync_certificates_rejects_unexpected_frame`.
- The `certificate_fetcher` tests (8) are rewritten to drive the fetcher through the `MockFetcher` seam (injected canned certs / errors / dropped replies), exercising the fan-out, retries, cancellation, gc-round update, and bad-cert rejection without a live stream.
- `cert_collector`'s `test_certificate_iterator` is migrated to the streaming constructor (full, uncapped expected sets).
- The obsolete `primary_tests::test_fetch_certificates_handler` (which tested the removed legacy serve and, per its own note, duplicated the `state_sync` collector test) is removed.
- Coverage gap (acknowledged): the requester's actual stream exchange (open + frame I/O) needs a real libp2p stream, so it is covered by the `sync_codec` round-trips and is exercised end-to-end only by integration/e2e, not by these unit tests.

## Gates (all green, `-j 2`)

- `cargo +nightly-2026-03-20 fmt --check` (clean)
- `cargo build -p tn-primary` (0 errors, 0 warnings)
- `cargo clippy -p tn-primary --all-features --all-targets` (0 errors; 21 warn-level lints pre-existing at the base, none in this diff)
- `cargo test -p tn-primary` (0 failed, 167 passed, 1 ignored)
- `cargo doc --no-deps -p tn-primary` with `RUSTDOCFLAGS=-D warnings` (clean)
- `cargo build -p tn-node` (0 errors, 0 warnings — the `clear_sync_capability` call site)

BCS encodes enum variant indices: no `SyncFrame` / `PrimarySyncRequest` variant is reordered, removed, or added, and the legacy `PrimaryRequest::MissingCertificates` / `PrimaryResponse::RequestedCertificates` variant slots are retained (answered with a typed error) until the coordinated `/0.0.2` bump and legacy freeze in step 9.


sstanfield (approved):

grantkee (approved):
LGTM!

PR #812, opened 2026-06-30
fix(worker): do not ban gossip relayer for author-content faults

### LINKED ISSUE #801: Malformed/mis-topic gossip Fatal-bans the honest relayer
- File: `crates/consensus/worker/src/network/error.rs:120-126` (penalty map) + the gossip handler call site
- Category: peer-scoring

**Root cause.** `Bcs(_) | InvalidTopic | UnexpectedBatch | DuplicateBatch | RequestHashMismatch => Penalty::Fatal`. These are correct for direct request/response (the peer is the originator), but for gossip the penalized peer is the `propagation_source` (relayer), not the message author. A Byzantine author's malformed payload Fatal-bans whichever honest peer relayed it. Same class as the issues fixed in #743 / #785.

**Impact.** An attacker bans honest validators by publishing malformed gossip they then propagate; can partition the gossip mesh. High.

**Patch [directional].** Do not convert author-content faults into a TN ban on the relayer; report them to gossipsub so mesh scoring handles propagation:
```rust
// gossip handler (caller of process_gossip)
if let Err(e) = self.process_gossip(&topic, &msg).await {
    // author-content faults must not ban the relaying peer
    let acceptance = if e.is_author_content_fault() {
        MessageAcceptance::Reject  // gossipsub down-scores propagation, no TN ban
    } else {
        MessageAcceptance::Ignore
    };
    return Ok(acceptance); // do NOT apply e.penalty() to propagation_source
}
```
Add `WorkerNetworkError::is_author_content_fault()` covering `Bcs/InvalidTopic/UnexpectedBatch/DuplicateBatch/RequestHashMismatch`. First confirm whether the current gossip path applies `penalty()` to the source; if #785 already split request vs gossip, just ensure these variants land on the no-ban side.

MavenRain:
Closes #801.

## Problem

When the worker processes gossip, a processing error is converted to a peer
penalty and applied to the gossip's `relayer` (the `propagation_source`), not to
the message author. Several content faults map to `Penalty::Fatal`:

```rust
WorkerNetworkError::InvalidTopic
| WorkerNetworkError::Bcs(_)
| WorkerNetworkError::TooManyBatches { .. }
| WorkerNetworkError::UnexpectedBatch(_)
| WorkerNetworkError::DuplicateBatch(_)
| WorkerNetworkError::RequestHashMismatch => Some(Penalty::Fatal),
```

These are correct for direct request/response, where the peer in hand *is* the
originator of the content.  They are wrong for gossip: the network layer accepts
and forwards a message after only a shallow check (`verify_gossip` validates
message size, authorized publisher, and topic subscription, but does not BCS
decode the payload or match the inner gossip variant to the topic).  An honest
relayer therefore cannot screen out a payload that later fails the worker's deep
validation.  Banning that relayer hands a Byzantine (but authorized) author a way
to Fatal-ban honest validators by publishing malformed/mis-topic gossip, which
honest peers then propagate.  Cascading relayer bans can partition the gossip
mesh.

Of the Fatal content faults, only `Bcs` (failed BCS decode) and `InvalidTopic`
(decoded variant does not match the topic) are actually reachable from the
gossip handler today;  the rest arise on the request/response and stream paths.

## Fix

- `WorkerNetworkError::is_author_content_fault()` (new): an exhaustive
  classifier that returns `true` for faults determined purely by message content
  (the Fatal protocol-violation variants) and `false` for faults attributable to
  the transport peer's behavior or local to this node.  The exhaustive match means
  any new error variant forces an explicit classification decision.
- `process_gossip` now suppresses the relayer penalty for author-content faults:
  `relayer.zip(e.penalty()).filter(|_| !e.is_author_content_fault())`.  The
  message is still rejected and logged;  only the unjust ban of the relaying peer
  is removed. The request/response (`process_report_batch`) and stream
  (`process_request_batches_stream`) paths are unchanged and still penalize
  unconditionally, because there the peer is the content originator.

## Scope and follow-ups

This PR fixes the worker layer named in the issue.  Two structurally related items
are intentionally left out of scope:

- The network layer applies the same `Penalty::Fatal` to `propagation_source`
  when `verify_gossip` rejects a message
  (`crates/network-libp2p/src/consensus.rs`).  Because every node runs the same
  deterministic shallow check under gossipsub `ValidationMode::Strict`, an honest
  relayer that rejects a message never forwards it, so that path is a narrower
  vector (reachable mainly under epoch-boundary committee-view skew).  Different
  crate, different fault class . . . better tracked as its own issue.
- The Byzantine author is still not penalized for deep-validation faults: the
  network layer already reported `Accept` to gossipsub before the worker's deep
  check runs, so gossipsub's invalid-delivery scoring never fires.  Penalizing the
  author identity (rather than the relayer) is a separate enhancement.  Note: a
  *Mild* relayer penalty is deliberately **not** used here, since it would still
  accumulate against innocent downstream relayers and partition the mesh slowly.

## Testing

- New unit test `test_is_author_content_fault` locks the classification: the six
  Fatal content faults are author-content faults (and are exactly the ones that
  map to `Penalty::Fatal`), while transport/local faults . . . including the
  content-but-request/response-only `BatchValidation` and the no-penalty
  `BatchEpochMismatch` . . . are not.
- `cargo nextest run -p tn-worker` green (0 failed); `make fmt` clean.

### DIFF
--- a/crates/consensus/worker/src/network/error.rs
+++ b/crates/consensus/worker/src/network/error.rs
@@ -135,6 +135,55 @@ impl WorkerNetworkError {
             | WorkerNetworkError::Internal(_) => None,
         }
     }
+
+    /// Whether this fault is determined purely by the *content* of a message - its
+    /// encoding, declared topic, identity, or count - and is therefore attributable to
+    /// whoever authored that content rather than to the peer that delivered it.
+    ///
+    /// This distinction matters for gossip: the peer in hand is the relayer
+    /// (`propagation_source`), not the author, so an author-content fault must not
+    /// penalize it (see issue #801). Direct request/response paths penalize
+    /// unconditionally because there the peer *is* the originator of the content.
+    ///
+    /// These are the protocol-violation group that [`Self::penalty`] maps to
+    /// [`Penalty::Fatal`] (faults in a message's encoding, declared topic, identity, or
+    /// count). [`Self::BatchValidation`] is also `Fatal` for some subvariants but is
+    /// deliberately excluded: it arises only on the request/response path, where the
+    /// peer *is* the originator and penalizing it is correct. Of this set, only
+    /// [`Self::Bcs`] and [`Self::InvalidTopic`] are currently reachable from the gossip
+    /// handler; the rest are classified for forward-safety.
+    pub fn is_author_content_fault(&self) -> bool {
+        //
+        // explicitly match every error type so the classification is revisited with changes
+        //
+        match self {
+            // content-determined protocol violations
+            WorkerNetworkError::Bcs(_)
+            | WorkerNetworkError::InvalidTopic
+            | WorkerNetworkError::TooManyBatches { .. }
+            | WorkerNetworkError::UnexpectedBatch(_)
+            | WorkerNetworkError::DuplicateBatch(_)
+            | WorkerNetworkError::RequestHashMismatch => true,
+            // Not author-content faults on any current path. `BatchValidation` is
+            // content-determined, but only arises on the request/response path (where
+            // the peer is the originator, so its own penalty is correct) and never
+            // reaches the gossip handler; the rest are transport-peer behavior,
+            // timeouts, or faults local to this node.
+            WorkerNetworkError::BatchValidation(_)
+            | WorkerNetworkError::NonCommitteeBatch
+            | WorkerNetworkError::Internal(_)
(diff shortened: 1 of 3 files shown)


sstanfield (approved):

grantkee (approved):
LGTM! Good work keeping these as bite-sized PRs

PR #813, opened 2026-06-30
Use the pack consensus output bytes for sync.

sstanfield:
This builds on the last PR to start using the bytes from a pack file for a consensus output when syncing an individual consensus output.


crates/state-sync/src/lib.rs
```diff
@@ -298,44 +298,35 @@ async fn catch_up_consensus_from_to<DB: Database>(
...
-        let consensus_header = if let Ok(Some(header)) = db.get::<ConsensusHeaderCache>(&number) {
-            // Always check the cache first, even if on max_consesus_height so we evict this record
-            // if cached.
-            // Note that the consensus header at number must be consenstent (consensus was reached)
-            // so this is fine. In other words it would be a protocol violation to have
-            // different ConsensusHeaders at the same number.
-            remove_cache = true;
-            header
-        } else if number == max_consensus_height {
-            max_consensus.clone()
```
MavenRain, line 313: Dropping this could lead to silently skipping the last block, no?

MavenRain (dismissed):
A single question, likely borne of my ignorance on the topic

crates/state-sync/src/lib.rs
sstanfield: It should not.  In order for it to be set the full output should be in either the cache (db) or the consensus_chain.  This used to be a small optimization to avoid a DB lookup but since it is not a full output just a header this won't work anymore.  Maybe it could be a full output now but even so I don't think this will cause any skipping.

grantkee (dismissed):
LGTM!

grantkee (approved):

PR #814, opened 2026-06-30
fix(primary): report peer penalties for failed vote requests (#802)

### LINKED ISSUE #802: Vote RPC bypasses the peer-penalty pipeline
- File: `crates/consensus/primary/src/network/mod.rs:932-943` (`process_vote_request`)
- Category: peer-scoring

**Root cause.** The spawned task does `vote.into_response()` then `let _ = network_handle.handle.send_response(...)`. Any error from `request_handler.vote(...)` is collapsed into a response and the error is dropped, so it never reaches the error->penalty path other request handlers use. A Byzantine validator misbehaves on the consensus-critical vote RPC at zero reputational cost.

**Impact.** No peer scoring on votes. (Severity is High for reputation, possibly Medium overall if vote equivocation is independently slashed at the consensus layer; confirm that.)

**Patch [directional].** Report penalizable vote errors before collapsing to a response:
```rust
vote = request_handler.vote(peer, header, parents) => {
    if let Err(ref e) = vote {
        if let Some(penalty) = e.penalty() {
            network_handle.report_penalty(peer, penalty).await; // exact API per sibling handler
        }
    }
    let response = vote.into_response();
    let _ = network_handle.handle.send_response(response, channel).await;
}
```
Use whatever penalty-reporting call the missing-certs handler uses; if primary penalties are applied centrally on returned errors, route the vote error through that path instead of swallowing it in `into_response()`.

MavenRain:
Closes #802.

## Problem

`process_vote_request` was the only primary request handler that dropped penalizable
errors instead of routing them through the peer-penalty pipeline.  It evaluated a peer's
proposed header and then collapsed any error straight into a response via `into_response()`
and discarded it (`let _ = ...send_response(...)`).  A misbehaving validator could therefore
return penalizable errors on the consensus-critical vote RPC (voting for a header it did
not author, equivocation, malformed parents, ...) at zero reputational cost.

Every sibling handler already reports `(&err).into()` before building the response:
`process_request_for_missing_certs`, `process_consensus_output_request`,
`process_epoch_record_request`, `process_epoch_stream`, `process_gossip`, and
`process_inbound_stream`.  The vote handler was the sole omission.

## Changes

- **`crates/consensus/primary/src/network/mod.rs`**: in `process_vote_request`, report a
  penalty for a failed vote before collapsing the result into a response, mirroring the
  sibling handlers exactly.  The error is routed through the existing
  `From<&PrimaryNetworkError> for Option<Penalty>` table, which stays the single source of
  truth for which vote errors are penalizable and at what level (wrong author / unknown
  authority / equivocation -> `Fatal`).

- **`crates/consensus/primary/src/error/network.rs`**: reclassify two `HeaderError` arms
  in `penalty_from_header_error` from `Mild` to `None`.  Wiring votes into the pipeline newly
  activates these arms on the consensus-critical path, and both fire on local/transient
  conditions rather than peer misbehavior:
  - `Storage` — our own DB error; the peer is blameless.  Every other `Storage` variant in
    this table (`PrimaryNetworkError::Storage`, `CertificateError::Storage`,
    `CertManagerError::Storage`) already maps to `None`.
  - `UnknownExecutionResult` — returned only when our execution lags behind a peer that is
    merely ahead during catch-up;  the vote handler's own comment notes "this doesn't hurt",
    and the `*::Timeout` arms already map to `None`.

- **`crates/consensus/primary/src/tests/network_tests.rs`**: two regression tests:
  - `test_vote_byzantine_errors_are_penalizable`: the Byzantine vote errors (`PeerNotAuthor`,
    `UnknownAuthority`) map to `Some(Penalty::Fatal)` . . . the contract the wiring depends on.
  - `test_local_header_errors_are_not_penalized`: the reclassified `Storage` /
    `UnknownExecutionResult` errors map to `None`.

## Notes on scope and effect

- The response path is unchanged: the same `PrimaryResponse` is still returned to the peer;
  only the previously-dropped error is now also surfaced to the penalty pipeline.

- `penalty_from_header_error` is shared with the certificate path, so the two reclassified
  errors now map to `None` there as well . . . strictly less punitive and consistent with the
  table's existing intent (do not penalize a peer for our local conditions).

- **Equivocation severity (the issue's open question).** Equivocation maps to `Fatal`
  (`AlreadyVoted` / `AlreadyVotedForLaterRound`), so this restores reputational consequences
  for it.  Two caveats worth recording:
  - Vote RPCs are served to committee members, and exempt peers (committee validators /
    operator allowlist) bypass the score model in `apply_penalty` (`peers/peer.rs`):
    `Severe`/`Fatal` are suppressed to a `warn!` and not enforced.  So against an in-committee
    equivocator the practical bite of the restored penalty is a log line, not a ban . . . still
    an improvement over the prior silence, but not enforcement.
  - There is no economic slashing wired in the live Rust path (`Slash` / `applySlashes`
    exist only as a `sol!` / Solidity interface, with no Rust call site that builds a slash).
     The issue's hypothesis "Medium overall if vote equivocation is independently slashed at
    the consensus layer" is therefore not supported: reputation is the sole runtime
    consequence.  Severity stays High for reputation. Wiring economic deterrence (or applying
    enforced penalties to committee equivocators) is a reasonable follow-up, out of scope here.

## Verification

- `cargo clippy -p tn-primary --all-targets` — clean (0 warnings).
- `cargo test -p tn-primary` — the 16 `test_vote*` tests pass, plus the new
  `test_local_header_errors_are_not_penalized`.
- `cargo +nightly fmt` applied.

### DIFF
--- a/crates/consensus/primary/src/error/network.rs
+++ b/crates/consensus/primary/src/error/network.rs
@@ -147,10 +147,7 @@ impl From<&PrimaryNetworkError> for Option<Penalty> {
 fn penalty_from_header_error(error: &HeaderError) -> Option<Penalty> {
     match error {
         // mild
-        HeaderError::SyncBatches(_)
-        | HeaderError::TooNew { .. }
-        | HeaderError::Storage(_)
-        | HeaderError::UnknownExecutionResult(_) => Some(Penalty::Mild),
+        HeaderError::SyncBatches(_) | HeaderError::TooNew { .. } => Some(Penalty::Mild),
         // medium
         HeaderError::InvalidParents
         | HeaderError::WrongNumberOfParents(_, _)
@@ -171,8 +168,16 @@ fn penalty_from_header_error(error: &HeaderError) -> Option<Penalty> {
         | HeaderError::InvalidParentTimestamp { .. }
         | HeaderError::UnkownWorkerId
         | HeaderError::UnknownAuthority(_) => Some(Penalty::Fatal),
-        // ignore
+        // ignore (local/transient, not the peer's fault)
+        //
+        // Storage is our own DB error, and UnknownExecutionResult means a peer is merely
+        // ahead of our execution while we catch up (the vote handler notes "this doesn't
+        // hurt"). Penalizing either would punish an honest peer for a local condition, and
+        // contradicts the sibling PrimaryNetworkError::Storage and *::Timeout arms that
+        // already map to None.
         HeaderError::PendingCertificateOneshot
+        | HeaderError::Storage(_)
+        | HeaderError::UnknownExecutionResult(_)
         | HeaderError::TNSend(_)
         | HeaderError::InvalidEpoch { .. }
         | HeaderError::ClosedWatchChannel => None,
--- a/crates/consensus/primary/src/network/mod.rs
+++ b/crates/consensus/primary/src/network/mod.rs
@@ -1288,6 +1288,19 @@ where
         self.task_spawner.spawn_task(task_name, async move {
             tokio::select! {
                 vote = request_handler.vote(peer, header, parents) => {
+                    // report penalty if any
+                    //
+                    // votes are consensus-critical, so a peer that returns a penalizable
+                    // error must pay the same reputational cost as on every other request
+                    // handler. Route the error through the central PrimaryNetworkError →
+                    // Penalty table before collapsing it into a response, instead of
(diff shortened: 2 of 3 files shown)


sstanfield (approved):
Looks good.

grantkee (approved):
LGTM! Thanks @MavenRain

PR #815, opened 2026-06-30
fix: return error instead of panicking on non-validator vote path

### LINKED ISSUE #803: `expect("only validators can vote")` panics on a non-committee node
- File: `crates/consensus/primary/src/network/handler.rs:743`
- Category: error-handling

**Root cause.** `self.consensus_config.authority_id().expect("only validators can vote")` panics if `authority_id()` is `None`. This depends on local node identity, so if a non-committee node ever reaches the new-vote path while serving a vote request, a peer can crash it. Also violates the no-`expect`-on-runtime-conditions convention.

**Impact.** Remote panic / node crash if non-validators serve the vote RPC; otherwise a latent assertion. Confirm whether vote handling is gated to committee members upstream.

**Patch [precise].**
```rust
let authority_id = self
    .consensus_config
    .authority_id()
    .ok_or(HeaderError::NotCommitteeMember)?;   // add variant if absent
let vote = Vote::new(&header, authority_id, self.consensus_config.key_config());
```

MavenRain:
Fixes #803.

## Problem

`vote()` and `vote_inner()` in the primary network handler each called
`self.consensus_config.authority_id().expect("only validators can vote")` (three sites).
`authority_id()` returns `None` for any node that is not a member of the current committee
(an observer / non-validator). The `RequestHandler` that serves the inbound vote RPC is
constructed and spawned on every node, and there is no upstream gate that filters the vote
path by this node's own committee membership. A committee peer (misrouted or malicious) can
therefore drive a non-validator node into that `expect` and crash it: a remote panic / node
halt. Even where current routing makes this unlikely, an RPC handler that processes peer
input should never panic, and the `expect` also violates the no-`expect`-on-runtime-conditions
convention.

## Changes

- Add `HeaderError::NotCommitteeMember` ("This node is not a committee member and cannot
  vote").
- Replace the three `.expect("only validators can vote")` calls in
  `crates/consensus/primary/src/network/handler.rs` with
  `.ok_or(HeaderError::NotCommitteeMember)?`. A non-validator now returns a graceful error
  instead of aborting the process. The `?` converts to `PrimaryNetworkError` through the
  existing `InvalidHeader(#[from] HeaderError)` conversion, so no signature changes are
  needed.
- Classify `NotCommitteeMember` as no penalty in `penalty_from_header_error`. The failure is
  this node's own non-membership, so the honest requesting peer must not be penalized for it.
  Reusing the existing `UnknownAuthority` variant (which maps to `Penalty::Fatal`) would have
  wrongly banned the requesting peer.
- Add `test_vote_non_committee_member_returns_error`: it builds a handler whose
  `authority_id()` is `None` (an outsider key over the same committee) and asserts that a
  valid vote request returns `Err(NotCommitteeMember)` rather than panicking. This test aborts
  the process under the old code and passes only with the fix.

## Notes

- No behavior change for validators. When the node is in the committee, `.ok_or(..)?` yields
  the identical `AuthorityIdentifier` that `.expect()` returned.
- All three sites are guarded (the quick recast path, the same-round re-vote path, and the
  fresh-vote path), so any future caller of these paths is also panic-safe.

### DIFF
--- a/crates/consensus/primary/src/error/network.rs
+++ b/crates/consensus/primary/src/error/network.rs
@@ -175,6 +175,7 @@ fn penalty_from_header_error(error: &HeaderError) -> Option<Penalty> {
         HeaderError::PendingCertificateOneshot
         | HeaderError::TNSend(_)
         | HeaderError::InvalidEpoch { .. }
+        | HeaderError::NotCommitteeMember
         | HeaderError::ClosedWatchChannel => None,
     }
 }
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -401,7 +401,7 @@ where
             if vote_info.vote_digest == header.digest().into() {
                 let vote = Vote::new(
                     &header,
-                    self.consensus_config.authority_id().expect("only validators can vote"),
+                    self.consensus_config.authority_id().ok_or(HeaderError::NotCommitteeMember)?,
                     self.consensus_config.key_config(),
                 );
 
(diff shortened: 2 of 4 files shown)


sstanfield (approved):
Not sure how a non-validator would try to vote but I also have no issue with this patch.  I think it is safer just in case (and I did not run all the paths down about how it could happen). Definitely fixes a future foot gun even if not reachable now.

grantkee (approved):
Great to have defense in depth!

PR #817, opened 2026-06-30
Add the pack bytes number to gossiped consensus results.

### LINKED ISSUE #782: Add the bytes needed to encode the consensus output to the consensus result gossip.
NOTE: This is being addressed via a pack file data re-org vs then output bytes being added to consensus output so this is only for the metrics now.

This will allow nodes following this gossip to have a target size and close a lot potential issues a malicious peer could cause.

While we have this size add two metrics as well:
- One to track the consensus output size
- One to track the max seen consensus output size

sstanfield:
NOTE: No longer adds bytes to consents result.  This will be handled by a pack file format reorder.

This adds the bytes taken in the pack file to consensus results.  It is threaded through channels but not used yet.  Also makes the pack file level save return after write (we persist after the call anyway so this should not hurt performance and surfaces errors when they happen).

This will assume all nodes use the same compression, this is also tide to pack files being the same hash per epoch.

We will be able to use this for metrics and also when downloading a consensus output when an observer.

Closes https://github.com/Telcoin-Association/telcoin-network/issues/782

### DIFF
--- a/crates/storage/src/consensus_pack.rs
+++ b/crates/storage/src/consensus_pack.rs
@@ -29,18 +29,21 @@ use tokio::{
         oneshot,
     },
 };
-use tracing::{debug, error, warn};
+use tracing::{debug, error};
 
 use crate::archive::{
-    data_file::{create_dir_synced, fsync_directory},
+    data_file::create_dir_synced,
     digest_index::index::HdxIndex,
     error::{fetch::FetchError, open::OpenError},
     fxhasher::FxHasher,
     index::Index as _,
-    pack::{DataHeader, Pack, PackCompression, DATA_HEADER_BYTES},
+    pack::{write_value, DataHeader, Pack, PackCompression, DATA_HEADER_BYTES},
     pack_iter::AsyncPackIter,
     position_index::index::{PosIndexValue, PositionIndex},
 };
+
+/// Current version for new pack files.
+pub const PACK_VERSION: u16 = 1;
 
 /// Metadata for an Epoch.  Should always be the first record in a consensus pack.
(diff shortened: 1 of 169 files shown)


MavenRain (dismissed):
If receivers trust the signed `consensus_bytes` instead of recomputing, quorum can halt if validators ever produce different counts.  This would depend on byte-identical serialization and no future write site using a different `PackCompression`.  These would silently fragment quorum.

It might be worth adding an explicit invariant or test pinning compression and a cross-version determinism check, instead of leaving it implicit.

crates/storage/src/consensus_pack.rs
```diff
@@ -319,29 +304,23 @@ impl ConsensusPack {
         self.epoch
     }
 
-    /// Return a delayed error value.
-    /// Work is sent to a background thread and any errors are recorded.
-    pub fn get_error(&self) -> Result<(), PackError> {
```
grantkee, line 358: After a save failure, can a read/serve call now return partial/stale data where it previously errored? 

Might be worth incorporating a lightweight "pack is in a failed state" flag that the read/serve entrypoints check before serving, so a write-failed pack refuses to serve rather than serving a possibly-truncated prefix

grantkee (dismissed):
LGTM! One consideration for production hardening, but not a blocker

crates/storage/src/consensus_pack.rs
sstanfield: I right failure should cause swift node shutdown so I don't think we need this.

MavenRain (approved):

grantkee (approved):

PR #818, opened 2026-07-01
fix: stop remove_validator_ip from over-decrementing banned total (#809)

### LINKED ISSUE #809: `remove_validator_ip` over-decrements the banned-peer `total` (per IP), under-pruning the ban set
- File: `crates/network-libp2p/src/peers/banned.rs:65-78` (`BannedPeers::remove_validator_ip`), specifically the `self.total -= 1` at `banned.rs:75`;  called by `apply_committee_membership` at `crates/network-libp2p/src/peers/all_peers.rs:1213-1215`.
- Category: peer-scoring
- Severity: Medium (resource-bound erosion / slow memory growth; not fund loss, consensus, or active ban evasion).

## Root cause

`BannedPeers` tracks two things: `total: usize`, the count of banned **peers**, and `banned_peers_by_ip: HashMap<IpAddr, usize>`, a per-IP count.  `total` is incremented exactly once per peer in `add_banned_peer` (`banned.rs:82`) and decremented once per peer in `remove_banned_peer` (`banned.rs:43`).

`remove_validator_ip` is an **IP-map** cleanup: its documented job (`banned.rs:60-64`) is to clear a committee member's IPs from `banned_peers_by_ip` so the validator's IP is not blocked.  However, its loop body decrements the peer-level `total` **once for every IP entry it removes**:

```rust
for ip in ip_addresses {
    if let Some((_, _)) = self.banned_peers_by_ip.remove_entry(&ip) {
        warn!(...);
        self.total = self.total.saturating_sub(1); // banned.rs:75  <-- wrong
    }
}
```

Removing an IP-map entry unbans no peer, so this decrement is incorrect in **every** call path.  `total` is only ever written at `banned.rs:43`, `:75`, and `:82`; it is never re-derived from the live peer set, so the error is not self-correcting.

## Impact 1: a banned committee member is over-decremented by K on rotation

`apply_committee_membership` (`all_peers.rs:1173`) runs each committee rotation. For a member that currently holds `Banned` status:

1. The `Banned` arm calls `update_connection_status(.., Unbanned)` (`all_peers.rs:1199-1203`).
2. That dispatches to `handle_unbanned_transition`, whose `Banned` arm sets the peer to `Disconnected` and calls `remove_banned_peer` (`all_peers.rs:786`): this correctly does `total -= 1` once. It uses `get_mut` + `saturating_sub` and **does not remove zero-count keys** (`banned.rs:47-50`), so the member's IPs remain in the map at count 0.
3. Then, unconditionally for any member still in `self.peers`, line 1215 calls `remove_validator_ip(.., peer.known_ip_addresses())`. Its `remove_entry` finds each of those K lingering count-0 keys and does `total -= 1` for each (`banned.rs:73,75`).

Net result: `total` is decremented by `1 + K` instead of the correct `1`, where K is the number of the member's known IPs present in the map (IPs derive from the peer's own multiaddrs, `peer.rs:138-148`).

## Impact 2: an undercounted `total` under-prunes the banned set

`prune_banned_peers` sizes the prune **entirely** from `total()` and never consults the live count of `Banned` peers in `self.peers`:

```rust
let excess = self.banned_peers.total().saturating_sub(self.max_banned_peers); // all_peers.rs:1036
if excess > 0 { /* collect at most `excess` peers and evict */ }
```

So an undercounted `total` directly under-prunes.  The over-decrements accumulate monotonically across rotations, and `saturating_sub` floors `total` at 0, which is the worst case: `excess` is always 0, pruning never fires, and the banned-peer records grow unbounded, defeating the stated "prevent exhausting memory" purpose at `all_peers.rs:1034`.  This is a slow resource-bound erosion, not a correctness break of consensus.

A second, smaller manifestation hits **never-banned** members too: if a committee member shares an IP with a different still-banned peer, line 1215 wipes that shared IP entry and decrements `total` even though the member contributed nothing to it (see walkthrough 2).  Guarding only the `:1215` call site would therefore not fully fix the bug.

### 1. Banned committee member M with 2 known IPs (A, B), plus 3 unrelated background bans

Background bans are needed so `saturating_sub` does not floor at 0 and mask the over-decrement.

| Step | Call (file:line) | `total` | `banned_peers_by_ip` |
|------|------------------|---------|----------------------|
| 0 | P1,P2,P3 banned, `add_banned_peer` x3 (`banned.rs:82`) | 3 | {P1:1, P2:1, P3:1} |
| 1 | M reaches `Banned`, `add_banned_peer` (`all_peers.rs:722`) | 4 | + {A:1, B:1} |
| 2 | rotation, M `Banned` arm unbans, `remove_banned_peer` (`all_peers.rs:786`) | 3 (correct) | A:0, B:0 (keys remain) |
| 3a | same pass, `remove_validator_ip(M,[A,B])`, A `remove_entry` + `total-=1` (`banned.rs:73,75`) | 2 | {…, B:0} |
| 3b | …B `remove_entry` + `total-=1` (`banned.rs:73,75`) | **1** | {P1:1, P2:1, P3:1} |

Final `total = 1`; correct value is **3**. Over-decrement = 2 = K.

### 2. Never-banned committee member M co-located with a still-banned attacker A on IP X

| Step | Call (file:line) | `total` | `banned_peers_by_ip` | `ip_banned(X)` |
|------|------------------|---------|----------------------|----------------|
| 0 | A (non-committee) banned on X (`all_peers.rs:722`) | 1 | {X:1} | false (count not `> 1`, `banned.rs:114-116`) |
| 1 | rotation, M never banned, match arm does nothing (`all_peers.rs:1205-1208`) | 1 | {X:1} | false |
| 2 | same pass, `remove_validator_ip(M,[X])`, X `remove_entry` + `total-=1` (`banned.rs:73,75`) | **0** | {} | false |

Final `total = 0`; correct value is **1**.  Note: no **active** IP ban is dropped here (X was below the `> 1` block threshold), and A keeps its own per-peer ban (`remove_validator_ip` touches only `banned_peers_by_ip` and `total`, never status/reputation). The defect is the spurious `total` decrement, which feeds Impact 2.

### Test sketch (illustrative)

A regression test in the style of `crates/network-libp2p/src/tests/peers.rs` would drive walkthrough 1 through the real API and assert the corrected total:

```rust
#[test]
fn test_remove_validator_ip_does_not_corrupt_banned_total() {
    let mut all_peers = create_all_peers(None);

    // three unrelated background bans so saturating_sub does not floor the count
    for _ in 0..3 { ban_a_random_peer(&mut all_peers); } // Disconnecting{banned:true} -> Disconnected -> Banned

    // a committee member M with two known IPs, driven to Banned
    let (bls, net, peer_id) = make_peer_with_ips(&mut all_peers, 2); // upsert_peer(bls, net, vec![addr_a, addr_b])
    ban_peer(&mut all_peers, &peer_id);
    assert_eq!(all_peers.banned_peers.total(), 4);

    // rotate M into the committee -> apply_committee_membership unbans M and runs remove_validator_ip
    all_peers.update_committees(HashSet::new(), HashSet::from([bls]), HashSet::new());

    // M is unbanned (correct: -1), the 3 background bans remain -> expect 3.
    // On the current code this observes 1 (over-decremented by K = 2).
    assert_eq!(all_peers.banned_peers.total(), 3);
}
```

(Helpers `create_all_peers`, `update_connection_status`, `upsert_peer`, `update_committees`, `banned_peers.total()` already exist in the crate; `make_peer_with_ips`/`ban_peer` are thin wrappers over them.)

## Patch

Delete the `total` decrement from `remove_validator_ip`;  the method must only mutate `banned_peers_by_ip`. The peer-level unban accounting is already owned by `remove_banned_peer`.

```diff
 for ip in ip_addresses {
     debug!(target: "peer-manager", ?ip, "removing banned ip for validator");
-    if let Some((_, _)) = self.banned_peers_by_ip.remove_entry(&ip) {
+    if self.banned_peers_by_ip.remove(&ip).is_some() {
         warn!(target: "peer-manager", ?peer_id, "removed banned ip address associated with validator");
-        self.total = self.total.saturating_sub(1);
     }
 }
```

This fixes both Impact 1 (the banned-member double-decrement) and the never-banned co-located decrement, while preserving the intended validator-IP-unblocking behavior.  Guarding the call site at `all_peers.rs:1215` alone is insufficient, since it would not address the never-banned co-located path.  Optional hardening: have `remove_banned_peer` also drop count-0 keys (`banned.rs:47-50`) so the map does not retain stale zero entries.

## Reachability caveats

- Impact 1 requires a committee member to actually hold `Banned` status at a rotation.  In steady state that is rare, because validators are score-exempt, but it is reachable on the ban-then-rejoin path.  The co-located variant requires a real IP collision between a member and a banned peer.
- The per-event magnitude is often K == 1 for single-IP validators;  the consequential property is the uncorrected accumulation across rotations, not a single large jump.
- The within-crate call chain above is verified directly.  The external rotation drivers (`update_committees`, epoch start) are the evident callers of `apply_committee_membership` and were not separately re-read.

Found while fixing #799 (a separate, already-patched heap-direction bug in the same prune subsystem).

MavenRain:
## What

`BannedPeers::remove_validator_ip` is an IP-map cleanup: its job is to clear a
committee member's IPs from `banned_peers_by_ip` so a validator's IP is not blocked.
It also decremented the peer-level `total` once for **every IP entry** it removed
(`banned.rs:75`).   Removing an IP-map entry unbans no peer, so that decrement is wrong
in every call path.

`total` is only ever written by `add_banned_peer` (+1) and `remove_banned_peer` (-1);
the bug added a spurious third write in `remove_validator_ip` (pre-fix `banned.rs:75`)
that this PR removes.  `total` is never re-derived from the live peer set, so the error
is not self-correcting.   It feeds
`prune_banned_peers`, which sizes the prune **entirely** from `total()`
(`all_peers.rs:1033`).   Across committee rotations the over-decrements accumulate
monotonically and `saturating_sub` floors `total` at 0, so `excess` is always 0,
pruning never fires, and the banned-peer set grows unbounded, defeating the stated
"prevent exhausting memory" purpose.

Two ways the spurious decrement is reached:
1. A banned committee member at a rotation is unbanned by `remove_banned_peer` (correct
   `-1`), then `apply_committee_membership` (`all_peers.rs:1212`) runs
   `remove_validator_ip` over its lingering count-0 IP keys, decrementing `total` by an
   extra `K` (the member's known-IP count).
2. A never-banned member co-located on an IP with a still-banned peer: clearing the
   member's IP wipes the shared entry and decrements `total` even though the member
   contributed nothing.   (This is why guarding only the call site is insufficient.)

## Change

- `crates/network-libp2p/src/peers/banned.rs`: delete the `self.total -= 1` from
  `remove_validator_ip`;  the method now only mutates `banned_peers_by_ip`
  (`remove_entry` -> `remove`, since both halves were discarded).   Doc comment updated to
  record the invariant: `total` is owned by `remove_banned_peer` and must not be touched
  here.
- `crates/network-libp2p/src/tests/banned_peers.rs`: two regression tests -
  `test_remove_validator_ip_does_not_decrement_total` (mirrors the rotation path: ban,
  `remove_banned_peer`, then `remove_validator_ip`; asserts `total` unchanged) and
  `test_remove_validator_ip_for_never_banned_member_preserves_total` (the co-located
  variant).  Both fail on the pre-fix code and pass after.

## Notes

- Independent of #807 (the #799 heap-direction fix): that PR touches
  `collect_excess_peers` in `all_peers.rs`;  this one touches only `banned.rs` plus its
  unit tests, so there is no file overlap or conflict.
- Optional follow-up (not included, to keep this minimal): have `remove_banned_peer`
  drop count-0 keys so the map does not retain stale zero entries.

Closes #809.

### DIFF
--- a/crates/network-libp2p/src/peers/banned.rs
+++ b/crates/network-libp2p/src/peers/banned.rs
@@ -62,6 +62,11 @@ impl BannedPeers {
     /// This method is a precaution to ensure a committee member's IP address won't become
     /// blocked during initial connection. It's possible the IP address becomes banned again,
     /// but the validator won't be disconnected because it is a trusted peer.
+    ///
+    /// This only mutates `banned_peers_by_ip`. The peer-level `total` is owned by
+    /// [`Self::remove_banned_peer`] and is deliberately left untouched here: removing an IP-map
+    /// entry unbans no peer, so decrementing `total` per entry over-counts the unban and, across
+    /// committee rotations, floors `total` at 0 and silently disables ban-set pruning (issue #809).
     pub(super) fn remove_validator_ip(
         &mut self,
         peer_id: &PeerId,
@@ -70,9 +75,8 @@ impl BannedPeers {
         debug!(target: "peer-manager", ips=?self.banned_peers_by_ip, "filtering banned ips for validators");
         for ip in ip_addresses {
             debug!(target: "peer-manager", ?ip, "removing banned ip for validator");
-            if let Some((_, _)) = self.banned_peers_by_ip.remove_entry(&ip) {
+            if self.banned_peers_by_ip.remove(&ip).is_some() {
                 warn!(target: "peer-manager", ?peer_id, "removed banned ip address associated with validator");
-                self.total = self.total.saturating_sub(1);
             }
         }
     }
--- a/crates/network-libp2p/src/tests/banned_peers.rs
+++ b/crates/network-libp2p/src/tests/banned_peers.rs
@@ -2,7 +2,7 @@
 
 use super::*;
 use crate::common::{ensure_score_config, random_ip_addr};
-use libp2p::{multiaddr::Protocol, Multiaddr};
+use libp2p::{multiaddr::Protocol, Multiaddr, PeerId};
 use std::net::Ipv4Addr;
 
 /// Helper function to create a peer with specific IP addresses.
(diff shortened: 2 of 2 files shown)


sstanfield (approved):
LGTM!

grantkee (approved):
Nice find!

PR #820, opened 2026-07-01
refactor: push non-validator txns to committee via RPC instead of gossip

### LINKED ISSUE #804: refactor: Update Non-Validator Gossip for Transaction Submission
- instead of gossip, use KAD RPC to push to committee validators

MavenRain:
## Summary

Closes #804. Non-committee (non-CVV) workers no longer **gossip** the transactions they accept.  Instead they **push them directly to the committee validators over the existing request/response RPC** (routed to each validator by BLS key via KAD-discovered addresses), replacing the fire-and-forget `worker_txn` gossip topic with targeted, acked, retried delivery.

Inclusion semantics are unchanged: each committee member still keeps only the transactions in its own shard (`submit_txn_if_mine`, `sender_addr % committee_size == slot`), so an account's transactions still all land on one validator and nonce ordering is preserved.

## Changes

- **New RPC message** (`worker/network/message.rs`): `WorkerRequest::ReportTxns { transactions }` + `WorkerResponse::ReportTxns` ack.  Both are **appended at the end** of their enums so the change does not shift the serialized (BCS) variant discriminants of the existing variants — batch-sync / peer-exchange / report-batch RPC stay wire-compatible across versions.  Removed `WorkerGossip::Txn` (the gossip enum is now batch-only).
- **Send side** (`worker/worker.rs`): `disburse_txns` fans the accepted transactions out to every committee validator on a background task (`push_txns_to_committee`), one concurrent RPC per member, so batch production is never stalled by a slow/unreachable member.  Each per-member send retries transient errors with exponential backoff (`push_txns_to_peer`, mirroring `QuorumWaiter`), and does not retry a permanent rejection.  Committee push targets are captured once per epoch at construction (mirroring how `QuorumWaiter` captures the committee);  a non-CVV targets all members, a CVV excludes itself.
- **Handle** (`worker/network/handle.rs`): `publish_txn` → `report_txns(peer, transactions)`, mirroring `report_batch`.
- **Receive side** (`worker/network/handler.rs`, `mod.rs`): new `process_report_txns` + dispatcher mirroring `process_report_batch`.  Only committee members act; each re-derives its own slot and routes each txn through `submit_txn_if_mine`.  A request whose transaction count exceeds `MAX_TXNS_PER_REPORT` (well above any gas-bounded batch) is rejected as an invalid request and the peer is penalized, so peer scoring can throttle a flooder (parity with the gossip mesh's scoring).  Non-committee recipients ignore the request.
- **Removed gossip wiring**: `LibP2pConfig::worker_txn_topic` (`config/network.rs`) and its `subscribe(...)` in `start_epoch.rs`.  The batch gossip topic is untouched.

## Tests

- New unit tests for the receive path: a committee member offers every pushed txn to `submit_txn_if_mine` exactly once, in order, with its own committee size/slot;  an empty push is a no-op;  an over-cap push is rejected.
- Updated `test_batch_gossip_topics` (retired txn-topic cases removed;  batch topic enforcement kept).
- Full `tn-worker` suite green;  `make clippy` clean on the touched crates;  the e2e `test_restarts_observer` (observer submits a txn → a validator confirms it) passes on the new path.

## Deployment note

Existing worker RPC (batch sync, peer exchange, report batch) is wire-compatible across versions because the new variants are appended.  During a rolling upgrade the **new** non-validator→committee txn push is simply unavailable between a mixed pair (an old node can't decode `ReportTxns`;  a new node no longer subscribes to the txn gossip topic), so non-validator transaction submission may be degraded until nodes are upgraded (e.g. at an epoch boundary).  Consensus itself is unaffected.  I'd be happy to keep the gossip path for one release as a transition bridge if preferred.

## Possible follow-up (kept out to keep this a clean transport swap)

- Per-slot targeting: the sender could compute each txn's owning slot and push only to that one validator (1 RPC instead of N) rather than pushing to all members and letting each filter.

### DIFF
--- a/crates/consensus/worker/src/network/handler.rs
+++ b/crates/consensus/worker/src/network/handler.rs
@@ -28,15 +28,6 @@ use tracing::{debug, warn};
 /// Set for 500MB through emerging market worse-case 20MB/s upload
 const SEND_STREAM_TIMEOUT: Duration = Duration::from_secs(200);
 
-/// Maximum number of transactions accepted in a single [`WorkerRequest::ReportTxns`].
-///
-/// A legitimately built batch is gas-bounded to far fewer transactions than this
-/// (`max_batch_gas` / a transaction's intrinsic gas), so this only rejects abusive
-/// requests that pack the RPC frame with transactions to amplify signature-recovery work
-/// on the recipient. Over-limit requests are rejected and penalized so peer scoring can
-/// throttle a flooder (parity with the gossip mesh's peer scoring).
-const MAX_TXNS_PER_REPORT: usize = 5_000;
-
 /// The type that handles requests from peers.
 ///
 /// An instance is cloned for each request and used in a spawned task.
@@ -137,37 +128,6 @@ where
                             tracing::error!(target: "worker:network", "failed to get gossipped batch {batch_hash}: {e}");
                         }
                     }
-                }
-            }
-        }
-
-        Ok(())
-    }
-
-    /// Process transactions pushed by a peer for committee inclusion.
-    ///
-    /// This is the RPC replacement for the retired `WorkerGossip::Txn` path (issue #804):
-    /// only committee members act, and each keeps only the transactions that fall in its
-    /// own committee shard via `submit_txn_if_mine`. A non-committee recipient (or a member
-    /// that cannot locate its own slot) simply ignores the request.
-    pub(super) async fn process_report_txns(
-        &self,
-        transactions: Vec<Vec<u8>>,
-    ) -> WorkerNetworkResult<()> {
-        ensure!(
-            transactions.len() <= MAX_TXNS_PER_REPORT,
-            WorkerNetworkError::InvalidRequest(format!(
-                "ReportTxns exceeds max transactions: {} > {MAX_TXNS_PER_REPORT}",
-                transactions.len()
-            ))
-        );
-        if let Some(authority) = self.consensus_config.authority() {
-            let committee = self.consensus_config.committee();
-            let authorities = committee.authorities();
-            let size = authorities.len();
-            if let Some(slot) = authorities.into_iter().position(|auth| &auth == authority) {
-                for tx_bytes in &transactions {
-                    self.validator.submit_txn_if_mine(tx_bytes, size as u64, slot as u64);
                 }
             }
         }
@@ -363,15 +323,6 @@ where
         msg: &GossipMessage,
     ) -> WorkerNetworkResult<()> {
         self.process_gossip(msg).await
-    }
-
-    /// Publicly available for tests.
-    /// See [Self::process_report_txns].
-    pub async fn pub_process_report_txns(
-        &self,
-        transactions: Vec<Vec<u8>>,
-    ) -> WorkerNetworkResult<()> {
-        self.process_report_txns(transactions).await
     }
 
     /// Publicly available for tests.
--- a/crates/consensus/worker/src/network/handler.rs
+++ b/crates/consensus/worker/src/network/handler.rs
@@ -28,6 +28,15 @@ use tracing::{debug, warn};
 /// Set for 500MB through emerging market worse-case 20MB/s upload
 const SEND_STREAM_TIMEOUT: Duration = Duration::from_secs(200);
 
+/// Maximum number of transactions accepted in a single [`WorkerRequest::ReportTxns`].
+///
+/// A legitimately built batch is gas-bounded to far fewer transactions than this
+/// (`max_batch_gas` / a transaction's intrinsic gas), so this only rejects abusive
+/// requests that pack the RPC frame with transactions to amplify signature-recovery work
+/// on the recipient. Over-limit requests are rejected and penalized so peer scoring can
+/// throttle a flooder (parity with the gossip mesh's peer scoring).
+const MAX_TXNS_PER_REPORT: usize = 5_000;
+
 /// The type that handles requests from peers.
 ///
 /// An instance is cloned for each request and used in a spawned task.
@@ -130,23 +139,35 @@ where
                     }
                 }
             }
-            WorkerGossip::Txn(tx_bytes) => {
-                ensure!(
-                    topic.to_string().eq(&tn_config::LibP2pConfig::worker_txn_topic(
-                        self.consensus_config.chain_id()
-                    )),
-                    WorkerNetworkError::InvalidTopic
-                );
-                if let Some(authority) = self.consensus_config.authority() {
-                    let committee = self.consensus_config.committee();
-                    let authorities = committee.authorities();
-                    let size = authorities.len();
-                    for (slot, auth) in authorities.into_iter().enumerate() {
-                        if &auth == authority {
-                            self.validator.submit_txn_if_mine(&tx_bytes, size as u64, slot as u64);
-                            break;
-                        }
-                    }
+        }
+
+        Ok(())
+    }
+
+    /// Process transactions pushed by a peer for committee inclusion.
+    ///
+    /// This is the RPC replacement for the retired `WorkerGossip::Txn` path (issue #804):
+    /// only committee members act, and each keeps only the transactions that fall in its
+    /// own committee shard via `submit_txn_if_mine`. A non-committee recipient (or a member
+    /// that cannot locate its own slot) simply ignores the request.
+    pub(super) async fn process_report_txns(
+        &self,
+        transactions: Vec<Vec<u8>>,
+    ) -> WorkerNetworkResult<()> {
+        ensure!(
+            transactions.len() <= MAX_TXNS_PER_REPORT,
+            WorkerNetworkError::InvalidRequest(format!(
+                "ReportTxns exceeds max transactions: {} > {MAX_TXNS_PER_REPORT}",
+                transactions.len()
+            ))
+        );
+        if let Some(authority) = self.consensus_config.authority() {
+            let committee = self.consensus_config.committee();
+            let authorities = committee.authorities();
+            let size = authorities.len();
+            if let Some(slot) = authorities.into_iter().position(|auth| &auth == authority) {
+                for tx_bytes in &transactions {
+                    self.validator.submit_txn_if_mine(tx_bytes, size as u64, slot as u64);
                 }
             }
         }
@@ -342,6 +363,15 @@ where
         msg: &GossipMessage,
     ) -> WorkerNetworkResult<()> {
         self.process_gossip(msg).await
+    }
+
+    /// Publicly available for tests.
+    /// See [Self::process_report_txns].
+    pub async fn pub_process_report_txns(
+        &self,
+        transactions: Vec<Vec<u8>>,
+    ) -> WorkerNetworkResult<()> {
+        self.process_report_txns(transactions).await
     }
 
     /// Publicly available for tests.
--- a/crates/tn-reth/src/forward.rs
+++ b/crates/tn-reth/src/forward.rs
@@ -12,13 +12,38 @@
 //! [`submit_txn_if_mine`]: tn_types::BatchValidation::submit_txn_if_mine
 
 use crate::recover_raw_transaction;
-use alloy::providers::{Provider as _, RootProvider};
+use alloy::{
+    providers::{Provider as _, RootProvider},
+    transports::{RpcError, TransportErrorKind},
+};
 use std::{
     collections::{BTreeMap, BTreeSet},
     sync::{Arc, Mutex},
+    time::Duration,
 };
 use tn_types::{BlsPublicKey, RpcInfo, TaskSpawner, TxnForwarder};
+use tokio::time::timeout;
 use tracing::{debug, warn};
+
+/// Bounds a single validator's `eth_sendRawTransaction` round-trip so one unresponsive endpoint
+/// cannot stall the fallback chain; on timeout the forward tries the next validator.
+const FORWARD_SEND_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Bounds the total time spent forwarding one transaction across all its fallback validators, so a
+/// whole unresponsive committee cannot make a single transaction cost `committee_size` back-to-back
+/// [`FORWARD_SEND_TIMEOUT`]s. When it elapses the transaction is left unforwarded and the next one
+/// proceeds.
+const FORWARD_TX_BUDGET: Duration = Duration::from_secs(15);
+
+/// JSON-RPC error codes reth returns for a validator-local, transient condition (a full pool,
+/// `-32003`, or an internal/IO error, `-32603`) rather than a verdict on the transaction itself.
+/// The forward falls through to the next advertised validator on these, since another validator
+/// may still accept the transaction.
+const TRANSIENT_RPC_CODES: [i64; 2] = [-32003, -32603];
+
+/// Substring of reth's `eth_sendRawTransaction` error message when the transaction is already in a
+/// validator's pool (`code -32000`). Treated as a successful delivery, not a failure to retry.
+const ALREADY_KNOWN_MESSAGE: &str = "already known";
 
 /// Forwards observer transactions to validators over their advertised JSON-RPC endpoints.
 #[derive(Clone)]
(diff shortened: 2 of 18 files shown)


crates/consensus/worker/src/network/handler.rs
```diff
@@ -344,6 +365,15 @@ where
         self.process_gossip(msg).await
     }
 
+    /// Publicly available for tests.
```
sstanfield: Shouldn't this have a cfg(test)?  If not then not sure what the point of this vs just using pub (no super) on the function above is?

sstanfield (dismissed):
This will require every single observer that accepts transactions over RPC to potentially connect to every single committee member.  Are we sure this is a good idea/will work?  I guess the req/res will connect and disconnect so maybe it is ok?  Also it may be possible for validators to restrict their allowed libp2p connections for security, if so that would break this.  The intent of using gossip initially was so the TXNs would just be "out there" and find there way to committee members without then having to connect with tons for non-committee members (potentially).  Of course the direct send is more efficient.  I am just not sure this will not have unforeseen consequences in the near future when we are dealing with more observers and trying to figure out the best validator security stance.  This may be a layer in a bigger plan, for instance if a validator was protected it would need at least one proxy to act on it's behalf on the wider network.

crates/consensus/worker/src/network/handler.rs
MavenRain: It does have it: this fn is inside the `impl` block that starts at line 354, which is gated with `#[cfg(any(test, feature = "test-utils"))]`.  It's easy to miss in the diff since the new shim was appended into that existing block alongside `pub_process_gossip_for_test` and friends.

The reason for a shim rather than `pub` on `process_report_txns` itself: the callers are the IT tests in `tests/it/network_tests.rs`, which compile the crate as an external dependency, so `cfg(test)` is not active there and the entry point has to be genuinely `pub`, but only under the `test-utils` feature (enabled via the self dev-dependency in Cargo.toml).  An unconditional `pub` on `process_report_txns` would put the raw handler into the production API surface.

MavenRain:
> This will require every single observer that accepts transactions over RPC to potentially connect to every single committee member. Are we sure this is a good idea/will work? I guess the req/res will connect and disconnect so maybe it is ok? Also it may be possible for validators to restrict their allowed libp2p connections for security, if so that would break this. The intent of using gossip initially was so the TXNs would just be "out there" and find there way to committee members without then having to connect with tons for non-committee members (potentially). Of course the direct send is more efficient. I am just not sure this will not have unforeseen consequences in the near future when we are dealing with more observers and trying to figure out the best validator security stance. This may be a layer in a bigger plan, for instance if a validator was protected it would need at least one proxy to act on it's behalf on the wider network.

Good questions . . . here's my Claude-assisted point-by-point:

**1. Observers already connect to committee members today; this PR rides that pattern rather than creating it.** On the primary network, an observer catches up by sending `PrimaryRequest::ConsensusHeader` RPCs to validators (`state-sync`'s `get_consensus_header` -> `PrimaryNetworkHandle::request_consensus`), and the serving handler in `primary/src/network/mod.rs` even notes "observers legitimately request not-yet-served headers".  On the worker network, observers fetch batches they're missing from committee workers (`process_sync_batches_stream` / `process_request_batches_stream` serve any peer) and sat in the same batch gossip mesh. Committee bootstrap servers are registered as known peers for every node role at epoch init, and the consensus fetch tasks are spawned unconditionally regardless of role.  So observer->validator req/res is an existing operating assumption of the observer model; this PR adds a message type to it, not a new class of connection.  It's also not hypothetical that it works: the restarts-observer e2e (observer submits a txn, a validator confirms it) passes on the new path.

**2. On "connect and disconnect": usually it's even cheaper than that.** libp2p multiplexes everything over one connection per peer pair, and request-response reuses any established connection regardless of which side originally dialed (the protocol is registered `ProtocolSupport::Full`, and the send path picks from existing connections with no direction check).  It only dials when the peer is cold, using the KAD-cached addresses from `auth_to_peer`, and the swarm's `max_idle_connection_timeout` (65s default) reaps the connection once it goes idle.  Since an observer is typically already talking to validators on both networks (header/output sync on the primary side, batch sync on the worker side), the common case is zero new connections, just an occasional extra stream;  the worst case is a short-lived dial that closes about a minute after the push.

**3. On validators restricting connections: I checked what exists today, and a committee-only stance would break much more than this PR.** The inbound gate (`handle_established_inbound_connection`) rejects only self-connections, banned peers, and banned IPs;  `PeerConfig` has count/timing/score knobs but no identity allowlist (the `operator_allowlisted` flag is purely permissive and isn't wired to any config or CLI).  If committee-only inbound were added, observers would lose header sync and batch sync (both serving paths are ungated, unlike `process_report_batch`, which is already committee-gated), and the node's KAD server duty would starve too.  So any future restriction plan already has to include an observer carve-out, and whatever carve-out lets a validator serve headers and batches to observers also lets it receive their txn pushes.  It's worth noting the middle stance also works: a validator that stops accepting inbound but still dials outbound keeps both gossip and this push alive, because request-response happily sends over a validator-initiated connection.

**4. The "txns just out there" property of gossip was real, but it had a cost.** `tn-txn` was the only permissionless-publish topic in the system: subscribed with `publishers: None`, so `verify_gossip` accepted any signed message on it (primary and batch topics are committee-gated via `subscribe_with_publishers`), with no gossipsub-native peer scoring, no rate limiting, and only the 12KB size cap;  a rejected message penalizes the propagating peer, not the origin, so a flooder wasn't even attributable. Every worker in the network subscribed and re-forwarded, so any peer could make the whole network carry arbitrary bytes. The push flips those properties: the recipient sees exactly which peer is submitting, there's a hard per-request cap (`MAX_TXNS_PER_REPORT`), and abuse is penalized against the actual sender.  If validators do end up locking things down, an attributable, gateable RPC seems like the better substrate, not a worse one: gating `process_report_txns` on an allowlist later is a three-line mirror of `process_report_batch`'s committee gate, whereas an open gossip topic is hard to restrict without killing submission entirely.

**5. The scenario where gossip is genuinely more robust is narrower than it looks, and it converges on your proxy idea.** Under a strict committee-only stance (no connections in either direction with non-committee peers), gossip also stops delivering: mesh links only exist over established connections, so the committee's txn mesh would contain only committee members and outside txns would have no entry point.  Both designs need at least one bridging node there, which is exactly the proxy layer you describe, and I'd argue the RPC composes better with it: an acked request gives the proxy (and the observer behind it) delivery feedback that fire-and-forget gossip never had, and an application-level forwarder for a plain RPC is trivial.  Nothing relay-shaped is compiled today (no libp2p-relay/autonat/dcutr), so a protected-validator story is new work under either transport.  The residual case where gossip wins is a validator that refuses observer dials but keeps outbound non-committee connections;  if that stance is likely before a proxy tier exists, I'm happy to keep the gossip topic for one release as a bridge, per the note in the description.

**6. Scale, concretely.** Committee is 4 in the current mainnet config (`chain-configs/mainnet/committee.yaml`) and the e2e defaults, with ~100 as the planning envelope in the SyncConfig docs.  Per observer, that's at most committee-size connections, mostly reused or reaped.  Per validator, observer fan-in is already bounded by the peer manager: it accepts up to ~39 connections and prunes back to the target 30 every heartbeat, with committee members exempt.  A pruned observer just gets re-dialed on its next push, and the push itself is bounded and non-blocking (3 attempts with 200ms/400ms backoff, permanent rejections never retried, sealing never waits on delivery).  If fan-out ever matters at large committee sizes, the per-slot targeting follow-up in the description cuts each observer's push from N members to 1, and the committee's aggregate receive load by the same factor.

In summary: it works (and is exercised end to end in the observer e2e), it doesn't create a new connectivity class, and it leaves validators better positioned for a restrictive security stance than the open topic did, because it gives them an attributable, gateable submission channel.

grantkee (changes requested):
I think this is likely problematic from a scalability perspective. The req/res protocol requires a peer to fully connect, but this has a lot of overhead for just submitting a single transaction. The intention with including RPC in a validator node record was to support this these types of queries

The goal is to forward to the validator's RPC so users can have a complete RPC experience

https://github.com/Telcoin-Association/telcoin-network/blob/main/crates/network-libp2p/src/types.rs#L478-L487

sstanfield:
> > This will require every single observer that accepts transactions over RPC to potentially connect to every single committee member. Are we sure this is a good idea/will work? I guess the req/res will connect and disconnect so maybe it is ok? Also it may be possible for validators to restrict their allowed libp2p connections for security, if so that would break this. The intent of using gossip initially was so the TXNs would just be "out there" and find there way to committee members without then having to connect with tons for non-committee members (potentially). Of course the direct send is more efficient. I am just not sure this will not have unforeseen consequences in the near future when we are dealing with more observers and trying to figure out the best validator security stance. This may be a layer in a bigger plan, for instance if a validator was protected it would need at least one proxy to act on it's behalf on the wider network.
> 
> Good questions . . . here's my Claude-assisted point-by-point:
> 

> 
> In summary: it works (and is exercised end to end in the observer e2e), it doesn't create a new connectivity class, and it leaves validators better positioned for a restrictive security stance than the open topic did, because it gives them an attributable, gateable submission channel.

I agree it will work and is currently fine.  It also matches current patterns although sync is changing in ways that should not require any direct validator connections.  I just think we will outgrow this quickly.  For instance what if there are 1000 active validators (all accepting TXNs) trying to use 20 committee members?  Your last point indicates this will still work but will lead to a lot of connection churn.  Also, I don't know how we will protect validators in the future but I suspect using a firewall to control who can connect may be a thing (or maybe not).  It does sound like using gossip is also not the way forward...  We almost need some sort of "router" to send an indirect message to validator so it will not need to be slammed with connections.  We may also want to use dedicated observers to front to a validator vs it being wide open.  Anyway, I am just worried about the scalability but I think this need some planning.  I don't think this PR is wrong or anything and it certainly using the existing mechanisms we may just need to extend it to handle a sprawling network with far more observers (of unknown maliciousness) than validators.  This is something I am trying to keep in mind on the new sync stuff, i.e. a syncing node really does not care where it gets it's data it all has to be verified anyway.

sstanfield:
> > This will require every single observer that accepts transactions over RPC to potentially connect to every single committee member. Are we sure this is a good idea/will work? I guess the req/res will connect and disconnect so maybe it is ok? Also it may be possible for validators to restrict their allowed libp2p connections for security, if so that would break this. The intent of using gossip initially was so the TXNs would just be "out there" and find there way to committee members without then having to connect with tons for non-committee members (potentially). Of course the direct send is more efficient. I am just not sure this will not have unforeseen consequences in the near future when we are dealing with more observers and trying to figure out the best validator security stance. This may be a layer in a bigger plan, for instance if a validator was protected it would need at least one proxy to act on it's behalf on the wider network.
> 
> Good questions . . . here's my Claude-assisted point-by-point:
> 

> 
> In summary: it works (and is exercised end to end in the observer e2e), it doesn't create a new connectivity class, and it leaves validators better positioned for a restrictive security stance than the open topic did, because it gives them an attributable, gateable submission channel.

I agree it will work and is currently fine.  It also matches current patterns although sync is changing in ways that should not require any direct validator connections.  I just think we will outgrow this quickly.  For instance what if there are 1000 active validators (all accepting TXNs) trying to use 20 committee members?  Your last point indicates this will still work but will lead to a lot of connection churn.  Also, I don't know how we will protect validators in the future but I suspect using a firewall to control who can connect may be a thing (or maybe not).  It does sound like using gossip is also not the way forward...  We almost need some sort of "router" to send an indirect message to validator so it will not need to be slammed with connections.  We may also want to use dedicated observers to front to a validator vs it being wide open.  Anyway, I am just worried about the scalability but I think this need some planning.  I don't think this PR is wrong or anything and it certainly using the existing mechanisms we may just need to extend it to handle a sprawling network with far more observers (of unknown maliciousness) than validators.  This is something I am trying to keep in mind on the new sync stuff, i.e. a syncing node really does not care where it gets it's data it all has to be verified anyway.

Also, if we replace libp2p RPC with http RPC we will have similar issues.  Using a lot more observers spread out to accept TXNs should still consolidate some of the traffic (although should be use batching to consolidate it more?).  So I don't think this is something new but you're pushing the boundaries now and this has been an issue waiting to crop up.

MavenRain:
What changed:

- A non-committee ("observer") worker no longer opens a `WorkerRequest::ReportTxns` req/res stream to committee members. Instead, when it accepts a transaction, it forwards the raw bytes to the JSON-RPC endpoint the owning validator advertised on its worker `NodeRecord` (`RpcInfo.http`), discovered over kademlia via `get_all_validator_rpcs`.  That wires those discovery primitives into a real consumer for the first time (they were test-only before).

- Routing is preserved end to end: a transaction is forwarded to the validator whose committee slot owns the sender (`u64::from_le_bytes(sender[0..8]) % committee_size`), exactly the rule `submit_txn_if_mine` used receiver-side, so all transactions from one account still converge on a single validator and nonce ordering holds. If the owning validator hasn't advertised an endpoint (or is momentarily unreachable), it falls back to any other advertised validator; consensus dedupes by nonce, so the fallback is safe.

- The whole worker-layer push apparatus is gone: `WorkerRequest/Response::ReportTxns`, the `report_txns` handle method, the `disburse_txns` fan-out with its retry/backoff, the receiver-side `process_report_txns` + `MAX_TXNS_PER_REPORT`, and the per-epoch `committee_txn_targets`.  Forwarding lives in a small `WorkerRpcForwarder` (in `tn-reth`, where the HTTP client and `recover_raw_transaction` already are), injected into the worker like the batch validator, so the consensus crate stays free of the HTTP/alloy stack.

Two things worth flagging:

1. This only delivers to validators that advertise `worker.rpc`, which is opt-in operator config today and defaults to `None`. So the forward degrades gracefully (logs and skips when nothing is advertised), but for it to actually deliver in a deployment the committee validators need to advertise their endpoints.  The e2e observer tests now wire this in explicitly (`advertise_worker_rpc` writes each test validator's endpoint into its `node-info.yaml` between the ceremony and node start), so delivery is exercised end to end on this path.  I'd be happy to follow up with auto-deriving the advertisement from the node's configured RPC address if you want it on by default for operators too.

2. This also answers Steve's approval-comment concern: because ingress now rides a separate public HTTP surface that the libp2p swarm never dials, a validator can restrict its libp2p connections to committee-only without breaking observer submission.  The observer effectively becomes the submission proxy he described.

Observers already execute the full chain, so reads (receipts, nonce, `eth_call`) are served locally from their synced state;  this change closes the remaining gap, which was getting the submitted transaction to a validator for inclusion with real RPC feedback rather than a fire-and-forget push.

Let me know if the forwarding scope here matches what you had in mind, or if you want it pushed further toward a full method-proxying gateway.

crates/tn-reth/src/forward.rs
```diff
@@ -0,0 +1,224 @@
+//! Forward transactions accepted by a non-committee ("observer") worker to the committee.
```
sstanfield, line 1: Nit and not sure it matters but should this live in tn-reth?  Maybe it does.

sstanfield (dismissed):
LGMT!

crates/tn-reth/src/forward.rs
MavenRain: It's a good question.  The short version: it's fine here, and the reason it isn't next to
   its consumer is a dependency cycle.
  
   `WorkerRpcForwarder` is mostly a plain alloy HTTP client, but it has exactly one
   hard tie to reth: `owning_validator` calls `recover_raw_transaction` to recover
   the signer and pick the owning committee slot (mirroring `submit_txn_if_mine`).
    That helper pulls `reth-rpc-eth-types`, on which only `tn-reth` depends, so the
   file genuinely needs something only `tn-reth` owns.
  
   One natural instinct is to move it next to `disburse_txns` in `tn-worker`, but
   that cycles: `tn-reth` already depends on `tn-worker`, and the forwarder needs
   `recover_raw_transaction` back from `tn-reth`.  Moving it into `tn-types` cycles for
   the same reason (`tn-types` is the leaf crate).  So the only cycle-free
   relocations are `tn-node` (which just constructs it, no real gain) or a dedicated crate.
  
   The dedicated-crate option is the one with precedent: it's exactly how the
   sibling `BatchValidation` trait is handled (trait in `tn-types`,
   `BatchValidator`/`NoopBatchValidator` impl in `tn-batch-validator`, injected as
   `Arc<dyn Trait>`).  I'd prefer to keep it here, but holler and I can extract this to a
    tn-txn-forwarder crate to match the symmetry.

crates/tn-reth/src/forward.rs
```diff
@@ -0,0 +1,224 @@
...
+                let mut tried = BTreeSet::new();
+                let mut forwarded = false;
+                for key in ordered {
+                    if !tried.insert(key) {
+                        continue;
+                    }
+                    let Some(provider) = providers.get(&key) else {
+                        continue;
+                    };
+                    match provider.send_raw_transaction(tx.as_slice()).await {
```
grantkee: This should have a timeout to prevent stalling

crates/tn-reth/src/forward.rs
```diff
@@ -0,0 +1,224 @@
...
+                        Ok(_) => {
+                            forwarded = true;
+                            break;
+                        }
+                        Err(err) => {
```
grantkee: It might be worth matching on these errors in case there is a problem with the tx (no RPC will accept after epoch boundary fork?) Do the RPCs return err for duplicate transaction?

crates/tn-reth/src/forward.rs
MavenRain: Done, at two levels. Each send is wrapped in `tokio::time::timeout(FORWARD_SEND_TIMEOUT, ...)` (5s), matching the idiom already used in `tn-metrics` and `node/health.rs`.  Awaiting `send_raw_transaction` in alloy is the actual `eth_sendRawTransaction` round-trip (we drop the returned `PendingTransactionBuilder` and never await a receipt), so the timeout bounds exactly the network call. On elapse the transaction falls through to the next advertised validator.

Since the fallback chain can be as long as the committee, I also put a total per-transaction budget around the whole chain (`FORWARD_TX_BUDGET`, 15s), so a fully unresponsive committee can't make one transaction cost `committee_size` back-to-back 5s timeouts before the next transaction proceeds. Both values are named consts, easy to tune.

One thing this does *not* yet bound is the number of in-flight forward tasks: `disburse_txns` spawns the forward as fire-and-forget per sealed batch, so under sustained sealing against an unresponsive committee those tasks could still accumulate.  The per-transaction budget caps how long each one lives (so they now drain in bounded time rather than hanging), but true backpressure (a single-flight/semaphore on the forward path) is a separate, pre-existing hardening item . . . happy to take it in a follow-up alongside the auto-advertise work if you want it.

crates/tn-reth/src/forward.rs
MavenRain: Yes . . . reth returns an error for a duplicate: JSON-RPC **code `-32000`, message `"already known"`** (its pool maps `AlreadyImported` to `AlreadyKnown` at the RPC boundary).  So the old "retry on any error" arm was treating an already-delivered tx as a failure and pushing it at every other validator.

The `Err` arm now classifies each response instead of blindly falling through:

- **Already in a pool** (`"already known"`) → treated as **delivered**, stop.  Answers your duplicate question directly and stops the thrash.
- **Considered rejection of the tx itself** (`-32000`: nonce too low, underpriced, replacement-underpriced, invalid, exceeds gas, wrong fork after an epoch boundary) → **stop, don't re-send to other validators**. Every validator shares consensus state, so the rejection repeats everywhere . . . that's your "no RPC will accept after the fork" case.  It also protects the single-validator-convergence invariant the module documents: re-sending a rejected tx to a non-owner would split one account's nonce stream across two pools.
- **Transient / endpoint-local** (full pool `-32003`, internal error `-32603`, transport failure, or the new timeout) → **keep falling through** to the next validator, since a healthy one may still accept it.

On discriminants: `-32003` (full pool) is the only pool rejection reth gives a distinct code, so that transient case is matched by **code**; `"already known"` has no distinct code (it shares `-32000` with the terminal rejections), so it's matched by **message**.  I added a unit test pinning the three representative cases (`already known` → delivered, `txpool is full` → try next, `nonce too low` → stop).

It's worth noting the stop-on-rejection also fixes a latent convergence bug in the old "retry on any error" path: a `replacement transaction underpriced` from the owning validator would previously get re-sent to another validator that lacks the original pending tx and would accept it . . . landing one account's same-nonce txns on two different validators, which is exactly the single-validator convergence this module exists to preserve.  Stopping keeps the account on its owner.

grantkee (approved):
Good work @MavenRain !

PR #823, opened 2026-07-01
# feat(worker-gateway): stateless reverse proxy crate skeleton + proxy core (#712)

MavenRain:
## What this adds

A new binary crate `bin/worker-gateway/` (package `tn-worker-gateway`, binary `worker-gateway`): a stateless reverse proxy that fronts a worker's JSON-RPC endpoint so it can be scaled horizontally.  Every instance is identical and stateless, so any replica can serve any request.

This PR is the **crate skeleton + proxy core**.  Edge protections (PR3) and observability/deployment (PR4) build on top.

### Scope implemented here

- **clap v4 CLI + env-var fallbacks**, and a **YAML upstream config** loaded through `ConfigTrait` (`crates/config`).  Upstreams are `{ worker_id, rpc_url, readiness_url }`.
- **Verbatim HTTP forwarding** of the full JSON-RPC surface (`eth_*` / `net_*` / `web3_*` / `tn_*`) via `reqwest`: method, body and content-type in, upstream status/body/content-type out.
- **Readiness polling** of each upstream's `GET /health/workers` (PR1 contract) on a configurable interval/timeout. A worker counts as ready only when its entry reports `accepting_transactions: true`;  unreachable / timed-out / malformed / absent all mark it **not-ready (fail closed)**. Unknown fields and newer envelope versions are tolerated (forward compatible).
- **JSON-RPC error gating**: when no upstream is ready (`503`), unreachable (`502`), or timed out (`504`), the client gets a well-formed JSON-RPC 2.0 error (codes `-32000/-32001/-32002`) with the request `id` echoed . . . never a bare connection reset.
- **Gateway's own endpoints**: `GET /health` (liveness, always `200`) and `GET /ready` (`200`/`503` gated by upstream availability), on the same listener as the proxy.
- **Graceful SIGTERM drain** via `tn_types::TaskManager` (`join_until_exit` installs the SIGTERM/ctrl-c handler) + axum `with_graceful_shutdown`, bounded by `--graceful-shutdown-timeout`.
- Per-request upstream **connect + request deadlines**.
- A `README.md` documenting config/env/YAML, the readiness contract, endpoints, failure behaviour, and graceful shutdown.

### Reuses

`clap` (derive+env), `tokio`, `tracing`/`tracing-subscriber`, `serde`/`serde_json`, `humantime`, `url`, `eyre`, `reqwest`, and `tn_types` `TaskManager`/`TaskSpawner`/`Notifier` + `tn_config` `ConfigTrait` — all existing workspace deps.

### One new workspace dependency

`axum = { version = "0.8.6", default-features = false }` added to `[workspace.dependencies]`.  axum isn't currently a direct workspace dep on `main` (the existing healthcheck/metrics servers are hand-rolled raw-TCP).  For a reverse proxy, axum + reqwest is the right tool and matches the HTTP-library decision we agreed on in the proposal.

## Not in scope (deferred, per the plan)

- WebSocket / `eth_subscribe` pass-through (v1 is HTTP-only;  subscriptions are per-connection stateful and break the stateless-scaling invariant).
- Method-aware routing (the config models a worker *list* and forwards to the first ready worker; v1 implements no selection beyond that).
- Rate/size limits + `eth_sendRawTransaction` sanity checks → **PR3**.
- Prometheus metrics, Dockerfile, k8s/HPA manifests → **PR4**.
- TLS termination, auth/API keys, dynamic discovery: out of scope per the issue.

## Testing

22 tests (unit + in-process end-to-end against a mock upstream): readiness parsing across the full contract (ready/not-accepting/absent/empty/malformed/newer-version), first-ready preference ordering, config YAML load + missing-file error, CLI resolution (inline resolve, both-URLs-required, `https` rejected as HTTP-only), JSON-RPC error rendering + id recovery, and live server tests for `/health`, `/ready` state transitions, happy-path proxy forwarding, the no-upstream-ready `503` JSON-RPC error, and the oversized-body `413` JSON-RPC error.

Gates: `cargo check`, `cargo clippy -D warnings` (0/0), `cargo test` (18 pass), `cargo +nightly fmt --check` — all clean.


bin/worker-gateway/README.md
```diff
@@ -0,0 +1,113 @@
+# worker-gateway
+
+A stateless reverse proxy that fronts a Telcoin Network worker's JSON-RPC
+endpoint. It forwards the full JSON-RPC surface (`eth_*` / `net_*` / `web3_*` /
+`tn_*`) verbatim to a ready upstream worker, gates traffic on a polled
+per-worker readiness signal, and exposes its own liveness and readiness
+endpoints so an orchestrator can route around it.
+
+Because every instance is stateless and identical, the gateway can be scaled
+horizontally: any replica can serve any request. This is PR2 of the epic
```
grantkee, line 12: This is more of a PR comment than a README but not a blocker for this initial skeleton

grantkee (dismissed):
Good work! I flagged a PR-specific comment in the README, but not a blocker

I think leaving it in the bin/worker-gateway is a good solution for rapid development but this might be better off in a separate repo. Let's discuss this early next week

grantkee (approved):
LGTM!

PR #824, opened 2026-07-02
fix(network): attribute gossip faults to the accountable peer across the reject, worker, and primary paths (#819)

### LINKED ISSUE #819: network-libp2p: the gossip Reject path Fatal-bans the relaying peer for author-attributable faults (the #801/#785 fix never reached it), and author-content faults caught by the worker's deep validation now penalize no one
## Summary

#801 fixed the worker layer so a gossip *content* fault (`Bcs`, `InvalidTopic`) no longer Fatal-bans the relaying peer (`propagation_source`) instead of the message author.  Two adjacent facets of the same author-vs-relayer misattribution remain, both in the gossip path but outside the worker layer #801 touched:

1. **The network-layer `Reject` branch still Fatal-bans the relayer.** When `verify_gossip` rejects a message, `process_gossip_event` applies `Penalty::Fatal` to `propagation_source`.  Every `Reject` reason is author-attributable (oversized payload, author not authorized, author BLS unresolved), yet the penalty lands on the forwarder.  This is the exact #801 class, one layer up.  It is narrower than the original because the peer-manager exemption gate spares current-committee relayers with a resolved BLS, but two honest relayers survive that gate and can be banned.

2. **After #801, a deep author-content fault penalizes no one.** The worker's deep validation runs *after* the network layer already reported `Accept` to gossipsub, and #801 (correctly) stopped charging the relayer.  Nothing charges the author either, even though the author is cryptographically identified.  A Byzantine-but-authorized author can emit malformed/mis-topic gossip with zero peer-scoring consequence.  This is an accountability gap, not a liveness threat.

Note on mechanism: this codebase configures **no gossipsub peer score** (no `with_peer_score` / `PeerScoreParams` anywhere). The only consequence channel for either facet is TN's custom `PeerManager` penalty, so there is no built-in mesh-scoring backstop to fall back on.

Line numbers are against `main` at `f3a8d592` (the network-libp2p and node files below are unchanged by the #801 worker-only patch).

---

### Where: `crates/network-libp2p/src/consensus.rs`

`verify_gossip` (1155-1176) rejects on three conditions, all determined by the *author*, not the forwarder:

```rust
fn verify_gossip(&self, gossip: &GossipMessage) -> GossipAcceptance {
    // (a) oversize: the payload was authored upstream
    if gossip.data.len() > self.config.max_gossip_message_size {
        return GossipAcceptance::Reject;
    }
    let GossipMessage { topic, .. } = gossip;
    // (b)/(c) author not authorized for topic, or author BLS not yet resolvable
    if gossip.source.is_some_and(|id| {
        let bls_key = self.swarm.behaviour().peer_manager.peer_to_bls(&id);
        self.authorized_publishers.get(topic.as_str()).is_some_and(|auth| {
            auth.is_none()
                || (bls_key.is_some()
                    && auth.as_ref().expect("is some").contains(&bls_key.expect("is some")))
        })
    }) { GossipAcceptance::Accept } else { GossipAcceptance::Reject }
}
```

`process_gossip_event` then unconditionally Fatal-penalizes `propagation_source` on `Reject` (917-920):

```rust
} else {
    self.metrics.record_gossip_rejected();
    let GossipMessage { source, topic, .. } = message;
    warn!(target: "network", author = ?source, ?topic,
        "received invalid gossip - applying fatal penalty to propagation source: {:?}",
        propagation_source);
    self.swarm.behaviour_mut().peer_manager
        .process_penalty(propagation_source, Penalty::Fatal);   // <-- relayer, not author
}
```

### Why an honest relayer can be the victim (reachability)

Gossipsub is configured `ValidationMode::Strict` + `validate_messages()`, so a peer only forwards what *it* accepted.  For an honest relayer A to forward a message that this node B then rejects, A and B must **disagree** on `verify_gossip`. Two real disagreement windows exist:

- **(b) authorized-publisher view skew at an epoch boundary.** The gossip topics' publisher set is only the *current* `committee_keys` (`crates/node/src/manager/node/start_epoch.rs:686-690` for `worker_batch_topic`; primary topics similarly).  During epoch rotation the publisher set and the peer-manager's committee slots are installed by separate calls, so an author authorized under A's view but not yet under B's makes A accept+relay and B reject.
- **(c) author-BLS resolution race.** `verify_gossip` requires `bls_key.is_some()` (1168): if B has not yet resolved the author `PeerId -> BLS` (the "Connected before `NodeRecord` resolves" race the code documents on the Accept path at 878-887), B rejects a message A already validated.

The oversize check (a) is deterministic from a shared static constant (`max_gossip_message_size = 12_000`, `crates/config/src/network.rs:184`), so it cannot diverge between honest nodes and is *not* an honest-disagreement vector.

The Fatal only *bans* when the relayer is not exempt.  `apply_penalty` (`crates/network-libp2p/src/peers/peer.rs:154-178`) makes `Severe`/`Fatal` a no-op `warn!` for exempt peers; exemption (`all_peers.rs:852-872`) requires a BLS Confirmed in the prev/current/next committee.  Two honest relayers fall outside it:

1. **Non-committee NVV relayers.** They legitimately subscribe to and relay the batch/certificate gossip topics but hold no committee slot, so any `Reject` they relayed bans them outright.
2. **Committee relayers whose BLS is still unresolved at B.** `identity_for` (`all_peers.rs:107-113`) returns `Unidentified` until the BLS resolves, so the Validator exemption does not apply and the ban lands.

### The #785 inconsistency

Case 2 is the *same* unresolved-identity race that #785 explicitly handled on the **Accept** branch (`consensus.rs:888-897`: skip the consensus penalty and still deliver when the relayer is unresolved).  The **Reject** branch received no equivalent guard, so the race #785 neutralized for accepted gossip is still live (and ban-bearing) for rejected gossip.

---

### Where

`process_gossip_event` reports `Accept` to gossipsub (`consensus.rs:868`) *before* the worker runs its deep validation.  The worker's deep check (`crates/consensus/worker/src/network/handler.rs:60`) discards the author identity it is handed:

```rust
pub(super) async fn process_gossip(&self, msg: &GossipMessage) -> WorkerNetworkResult<()> {
    let GossipMessage { data, source: _, sequence_number: _, topic } = msg;  // author dropped
    let gossip = try_decode(data)?;  // -> Bcs on garbage
    // ... ensure!(topic == expected, InvalidTopic) ...
}
```

After #801, `mod.rs:341-369` suppresses the relayer penalty for these `is_author_content_fault` variants.  The author is never charged, so a `Bcs`/`InvalidTopic` fault is consequence-free.

### Why this is only low

- **No honest victim.** That is exactly what #801 fixed; this is residual author *impunity*, not a relayer ban.
- **The author is reliably attributable.** Under `MessageAuthenticity::Signed` + `ValidationMode::Strict`, `GossipMessage.source` is the cryptographically-bound publisher, and on the batch topic `verify_gossip` has already proven `peer_to_bls(source)` is a current-committee key.  The information to charge the author exists at the network layer; it is simply thrown away (`handler.rs` `source: _`) and unreachable from the worker handle (`report_penalty` takes a `BlsPublicKey`, and the worker holds no `PeerManager`).
- **Not a DoS amplifier.** The cost per malformed message is one failed `bcs` decode or one topic-string compare, plus gossipsub re-propagation of an already-accepted message, bounded by `max_gossip_message_size` and gossipsub's own duplicate cache and mesh degree.  The only amplifier in this code (`request_batches` fan-out, `handler.rs:99-100`) sits behind a *successful* decode + correct topic + matching epoch + local cache-miss, i.e. a well-formed `Batch(current_epoch, bogus_digest)` that returns `Ok(())` and is not a fault at all, a separate hardening item, out of scope here.
- Scope of author: the batch topic requires committee membership;  the transaction topic is open (`worker_txn_topic` is subscribed without a publisher restriction, `start_epoch.rs:680`), so any peer with a `source` can author there.

---

## Impact

- **Finding 1 (medium):** a Byzantine author can craft gossip that is authorized under some honest relayers' committee view (or that arrives during the BLS-resolution lag) but not under a victim node's view, causing the victim to Fatal-ban the honest forwarder (`process_penalty -> Reputation::Banned -> NewConnectionStatus::Banned`;  the ban also taints the peer's IPs).  Repeated across the mesh this evicts honest non-committee NVV relayers and honest-but-unresolved committee relayers: the #801 attack class relocated to the network layer. Liveness / peer-graph health only; no consensus-safety impact.  Narrowed by the exemption gate and by the transient nature of the disagreement windows.
- **Finding 2 (low):** loss of the accountability/peer-scoring signal against a misbehaving staked validator, plus minor wasted work.  No liveness or safety threat.

## Suggested fix

**Finding 1:** mirror the #785/#801 reasoning on the `Reject` branch (`consensus.rs:907-921`):
- Resolve the relayer first;  if `peer_to_bls(propagation_source)` is `None`, skip the penalty and log, exactly as the Accept path does at 888-897;  this closes the unresolved-committee-relayer window and restores symmetry between the branches.
- Do not charge an author-attributable `Reject` to the forwarder: either penalize the resolved *author* (`gossip.source -> BLS -> peer`), or downgrade the relayer penalty to non-fatal for these shallow author faults, reserving `Fatal` for genuinely relayer-attributable misbehavior.

**Finding 2:** keep #801's relayer-suppression, but additionally attribute the fault to the author.  Cheapest: since `verify_gossip` already resolved `peer_to_bls(message.source)` on the batch topic, carry the author BLS forward alongside the relayer `Option` into the accepted-gossip event, so the worker can `report_penalty(author, Fatal)` on an `is_author_content_fault`.  Guard with the same unresolved-identity skip used for the relayer (the author can be unresolved only on the open txn topic; on the batch topic it is guaranteed resolved).

## Relationship to prior work

Follow-up to #801 (worker deep-validation layer, fixed) and #785 (Accept-path unresolved-relayer delivery, fixed).  Finding 1 is the Reject-path analogue of both;  Finding 2 is the accountability residue #801 necessarily left open.  Related silent-drop/`PeerMissing` issue on the same handler: #743.

MavenRain:
fix(network): attribute gossip faults to the accountable peer across the reject path, worker, and primary (#819)

## Summary

Closes #819.  Follow-up to #801/#812 (worker deep-validation layer) and #785 (Accept-path unresolved-relayer delivery).  #812 stopped the worker Fatal-banning the *relaying* peer for a message *author's* content fault, but three adjacent facets of the same author-vs-relayer misattribution remained in the gossip path:

1. the network-layer `Reject` branch still Fatal-banned the relayer for author-attributable rejects;
2. after #812 a deep author-content fault penalized *no one* (the author went uncharged);
3. the **primary** gossip path still Fatal-banned the relayer for author-content faults (`Decode` / `InvalidTopic`) — the same #801 bug on the sibling path.

Facet 3 is beyond #819's original text (which scoped the worker + network-reject paths) but the fix threads the author identity through the shared gossip event and touches the primary consumer, so leaving the identical honest-relayer ban on the adjacent path would be a half-fix;  it is included here.

This codebase configures **no gossipsub peer score** (no `with_peer_score` / `PeerScoreParams`), so the only consequence channel is TN's custom `PeerManager` penalty . . . there is no built-in mesh-scoring backstop.

## Finding 1: network-layer `Reject` branch Fatal-banned the relayer for author faults (medium)

`verify_gossip` rejects for oversized payload, unauthorized author, or unresolved-author (all determined by the *author*, not the forwarder) and `process_gossip_event` unconditionally Fatal-banned `propagation_source`.  Under gossipsub `Strict` validation a relayer only forwards what it accepted, so an honest relayer can only forward a message this node rejects when the two *disagree*: that happens for the author-authorized / author-BLS checks (epoch-boundary committee-view skew;  BLS-resolution race), but **not** for oversize, which is deterministic from a shared constant.

**Fix.** `verify_gossip` returns a typed `RejectReason`;  `RejectReason::penalty(relayer_resolved, author_resolved)` (unit-tested) charges an oversized payload from a *resolved* relayer to the forwarder (an honest peer never forwards a deterministically oversized payload), and charges an unauthorized author to the resolved *author* rather than the relayer, skipping either penalty while that peer's identity is unresolved (mirroring the #785 Accept path).  The relaying peer is never Fatal-banned for an author fault, and an honest author authorized under a neighbouring committee view is spared by the committee exemption in the peer manager.

## Finding 2: deep author-content fault penalized no one (low)

After #812 the worker stopped charging the relayer for `Bcs`/`InvalidTopic`/… faults from its deep validation, but nothing charged the author.  The accepted-gossip event now carries the author's resolved BLS alongside the relayer's, and the worker charges the accountable peer via the shared `is_author_content_fault` classifier . . . the resolved author for a content fault (guaranteed resolved on the restricted batch topic), the relayer otherwise.

## Primary path: same misattribution on the sibling gossip path (medium)

The primary's `process_gossip` charged the relayer for every fault, and `PrimaryNetworkError::Decode`/`InvalidTopic` map to `Fatal`.  On the open `consensus_output_topic` (any author accepted) a Byzantine authorized publisher's malformed/mis-topic gossip gets an honest non-committee relayer Fatal-banned.  Fixed by a new `PrimaryNetworkError::is_author_content_fault` (restricted to the two gossip-envelope faults authored by the source . . . `Decode` and `InvalidTopic`; embedded certificate/consensus faults are signed by a party inside the payload, not the gossip source, so their existing attribution is untouched) and the same author/relayer routing, using the author now plumbed through the event.

## Notable design choices

- **`NetworkEvent::Gossip` is now a named struct variant** `{ message, relayer, author }` (was `(GossipMessage, Option<BlsPublicKey>)`).  Charging a content fault to the wrong peer is the entire bug class, so the two identities are named rather than positional and cannot be transposed.
- **Attribution decisions are explicit, exhaustively-matched classifiers**: `RejectReason` / `RejectReason::penalty` at the network layer, and `is_author_content_fault` on each of `WorkerNetworkError` / `PrimaryNetworkError`.  Each `process_gossip` then selects author-vs-relayer from that single tested predicate — no bare boolean decides who gets banned in more than one place.

## Testing

- `RejectReason::penalty`: an `UnauthorizedAuthor` reject never penalizes the relayer (resolved or not);  an oversized reject is `Fatal` only for a resolved relayer.
- `accepted_gossip_event` carries the resolved author identity (new), alongside the existing relayer-carry regressions.
- `WorkerNetworkError::is_author_content_fault` (existing #801 test, unchanged) and a new `PrimaryNetworkError::is_author_content_fault` test pin exactly `Decode`/`InvalidTopic` as author faults.
- `cargocho build --tests` + targeted suites for `tn-network-libp2p`, `tn-worker`, `tn-primary` are clean (0 failed); fmt + clippy clean (no new warnings).

## Base

Branched off `origin/main` at `c8f7b062`. (The immediately prior HEAD `3f809525` had an unrelated compile error in `tn-primary` — a `.is_some()` on an already-unwrapped `u64` in `request_epoch_pack_inner` . . . that `c8f7b062` fixed by dropping the dead branch. This PR sits on the fixed base.)

## Relationship to prior work

Finding 1 is the reject-path analogue of #801/#785;  Finding 2 is the accountability residue #812 left open;  the primary fix extends #801's relayer-suppression to the sibling primary gossip path.  Related silent-drop/`PeerMissing` handling: #743.

### DIFF
--- a/crates/network-libp2p/src/types.rs
+++ b/crates/network-libp2p/src/types.rs
@@ -202,6 +202,12 @@ impl<Res> ResponseChannel<Res> {
 }
 
 /// Events created from network activity.
+// The `Gossip` variant carries the payload plus two resolved BLS identities (relayer + author)
+// for penalty attribution, making it the largest variant. `NetworkEvent` is a short-lived message
+// moved across a small-capacity channel, so the per-slot size is immaterial, whereas boxing a
+// field would add a heap allocation to the hot gossip-delivery path; the peer-manager event enum
+// takes the same allowance (see `peers/types.rs`).
+#[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
 pub enum NetworkEvent<Req, Res> {
     /// Direct request from peer.
@@ -215,16 +221,27 @@ pub enum NetworkEvent<Req, Res> {
         /// The oneshot channel if the request gets cancelled at the network level.
         cancel: oneshot::Receiver<()>,
     },
-    /// Gossip message received, with the relaying peer's BLS identity when it
-    /// has resolved.
+    /// Gossip message received, with the relaying peer's and the author's BLS identities
+    /// when they have resolved.
     ///
-    /// The identity is `None` during the brief window after a peer joins the
-    /// gossipsub mesh and relays a message before its signed `NodeRecord`
-    /// (which carries the BLS key) has been resolved. The payload is delivered
-    /// regardless, because the message author is already authenticated during
-    /// gossip verification; the relayer identity is used only for penalty
-    /// attribution and a log label, never for the payload itself.
-    Gossip(GossipMessage, Option<BlsPublicKey>),
+    /// Both identities are `Option` because a peer is `Connected` — and can relay or author
+    /// gossip — before its signed `NodeRecord` (which carries the BLS key) has been resolved.
+    /// The payload is delivered regardless, because the message author is already authenticated
+    /// during gossip verification; the identities are used only for penalty attribution and a
+    /// log label, never for the payload itself.
+    ///
+    /// `relayer` (the forwarding `propagation_source`) and `author` (the publishing
+    /// `GossipMessage::source`) are distinct peers, and charging a content fault to the wrong
+    /// one bans honest peers (see issues #801/#819); they are named rather than positional so
+    /// the two cannot be transposed.
+    Gossip {
+        /// The gossip message payload.
+        message: GossipMessage,
+        /// BLS identity of the relaying peer (`propagation_source`), when resolved.
+        relayer: Option<BlsPublicKey>,
+        /// BLS identity of the message author (`GossipMessage::source`), when resolved.
+        author: Option<BlsPublicKey>,
+    },
     /// Send an error back the requester.
     Error(String, ResponseChannel<Res>),
     /// An inbound stream was established by a peer.
--- a/crates/consensus/primary/src/error/network.rs
+++ b/crates/consensus/primary/src/error/network.rs
@@ -100,7 +100,6 @@ impl PrimaryNetworkError {
             | PrimaryNetworkError::Storage(_)
             | PrimaryNetworkError::InvalidRequest(_)
             | PrimaryNetworkError::Internal(_)
-            | PrimaryNetworkError::UnknownConsensusHeaderDigest(_)
             | PrimaryNetworkError::UnknownConsensusHeaderCert(_)
             | PrimaryNetworkError::UnknownConsensusOutput(_)
             | PrimaryNetworkError::PeerNotInCommittee(_)
--- a/crates/consensus/primary/src/error/network.rs
+++ b/crates/consensus/primary/src/error/network.rs
@@ -73,6 +73,51 @@ pub(crate) enum PrimaryNetworkError {
     ConsensusChainError(ConsensusChainError),
 }
 
+impl PrimaryNetworkError {
+    /// Whether this fault is determined purely by the *content* of a gossip message — its BCS
+    /// encoding or its declared topic — and is therefore attributable to the message author
+    /// (`GossipMessage::source`) rather than the peer that relayed it.
+    ///
+    /// Mirrors `WorkerNetworkError::is_author_content_fault` (tn-worker) and is consulted only on
+    /// the gossip path (`PrimaryNetwork::process_gossip`): the network layer accepted and forwarded
+    /// the message after a shallow check, so an honest relayer could not have screened a
+    /// content-determined fault, and charging the relayer for it lets a Byzantine author ban honest
+    /// forwarders (see issues #801/#819). The request/response paths penalize the peer
+    /// unconditionally because there the peer *is* the originator.
+    ///
+    /// Restricted to the two `Fatal` faults authored by the gossip source and reachable from the
+    /// gossip handler: [`Self::Decode`] (`try_decode` of the payload) and [`Self::InvalidTopic`]
+    /// (the payload variant was published to the wrong topic). Every other variant is either
+    /// benign on the gossip path or concerns embedded certificate/consensus content whose signer
+    /// is carried inside the payload and is not necessarily the gossip source, so its existing
+    /// relayer/no-penalty attribution is left unchanged.
+    pub(crate) fn is_author_content_fault(&self) -> bool {
+        //
+        // explicitly match every variant so the classification is revisited with changes
+        //
+        match self {
+            PrimaryNetworkError::Decode(_) | PrimaryNetworkError::InvalidTopic => true,
+            PrimaryNetworkError::InvalidHeader(_)
+            | PrimaryNetworkError::Certificate(_)
+            | PrimaryNetworkError::StdIo(_)
+            | PrimaryNetworkError::Storage(_)
+            | PrimaryNetworkError::InvalidRequest(_)
+            | PrimaryNetworkError::Internal(_)
+            | PrimaryNetworkError::UnknownConsensusHeaderDigest(_)
+            | PrimaryNetworkError::UnknownConsensusHeaderCert(_)
+            | PrimaryNetworkError::UnknownConsensusOutput(_)
+            | PrimaryNetworkError::PeerNotInCommittee(_)
+            | PrimaryNetworkError::UnavailableEpoch(_)
+            | PrimaryNetworkError::UnavailableEpochDigest(_)
+            | PrimaryNetworkError::InvalidEpochRequest
+            | PrimaryNetworkError::Timeout(_)
+            | PrimaryNetworkError::UnknownStreamRequest(_)
+            | PrimaryNetworkError::StreamUnavailable(_)
+            | PrimaryNetworkError::ConsensusChainError(_) => false,
+        }
+    }
+}
+
 impl From<&PrimaryNetworkError> for Option<Penalty> {
     fn from(val: &PrimaryNetworkError) -> Self {
         //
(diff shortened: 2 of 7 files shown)


crates/network-libp2p/src/types.rs
```diff
@@ -202,6 +202,12 @@ impl<Res> ResponseChannel<Res> {
 }
 
 /// Events created from network activity.
+// The `Gossip` variant carries the payload plus two resolved BLS identities (relayer + author)
+// for penalty attribution, making it the largest variant. `NetworkEvent` is a short-lived message
+// moved across a small-capacity channel, so the per-slot size is immaterial, whereas boxing a
+// field would add a heap allocation to the hot gossip-delivery path; the peer-manager event enum
+// takes the same allowance (see `peers/types.rs`).
+#[allow(clippy::large_enum_variant)]
```
grantkee, line 210: Should we create an issue to refactor this into smaller sub-enums?

grantkee (approved):
LGTM - no blockers. There are some residual primary gossip path that still charges the for malformed signed payloads I shared in a separate channel in case you prefer to address those in a follow up

MavenRain:
> LGTM - no blockers. There are some residual primary gossip path that still charges the for malformed signed payloads I shared in a separate channel in case you prefer to address those in a follow up

Yes . . . will definitely cover these in an immediate follow-up later this morning

crates/network-libp2p/src/types.rs
MavenRain: We absolutely should.  I will create the issue draft shortly.

PR #825, opened 2026-07-02
feat(network): move the goodbye exchange to a dedicated peer-exchange protocol with embedded fallback (739, step 8)

MavenRain:
Step 8 of #739.  The peer-exchange goodbye moves off the consensus request enums onto a dedicated small request-response protocol:

- `/tn-primary-peer-exchange-{chain_id}/0.0.1`
- `/tn-worker-{id}-peer-exchange-{chain_id}/0.0.1`

Goodbyes prefer the dedicated protocol and fall back to the `PeerExchange` variant embedded in the legacy request enum when the peer has not upgraded yet (outbound `UnsupportedProtocols`, which is already penalty-exempt).  The change is entirely inside `tn-network-libp2p`: zero consumer-crate edits, and no BCS enum variant is added, removed, or reordered.

## Why

The goodbye is the one message that forces the `From<PeerExchangeMap>` + `peer_exchange_msg()` bounds onto every consumer request/response enum through `TNMessage`, and the interception special-case in the req-res handler.  Extracting it to its own protocol is the prerequisite for dropping those bounds and the embedded variants at the coordinated `/0.0.2` bump (step 9).  It also gives the exchange its own wire identity so its traffic and failures are observable apart from consensus RPCs.

## Wire shape

Both directions carry a bare BCS `PeerExchangeMap` through the same hardened length-prefixed snappy codec (`TNCodec<PeerExchangeMap, PeerExchangeMap>`), so the goodbye path keeps the bounded-read guarantees from #742.  The request is the disconnecting node's exchange map; the response is an ack (empty today: the symmetric shape leaves room for a reciprocal exchange).  `impl TNMessage for PeerExchangeMap` satisfies the codec bounds; `From<PeerExchangeMap>` holds reflexively.

## Outbound path

`PeerEvent::DisconnectPeerX` now calls `send_goodbye`, which opens the dedicated protocol and tracks the attempt in a new `pending_goodbyes` map.  This map is deliberately separate from `pending_px_disconnects`: `OutboundRequestId`s are scoped to the behaviour that issued them, so ids from the dedicated behaviour could collide with legacy req-res ids in a shared map.  Each entry retains the exchange map (for the fallback resend) and a oneshot to the disconnect-deadline task.

Resolution matrix (disconnect discipline unchanged from the legacy path):

- **Ack received** → disconnect immediately (mirrors the legacy response arm).
- **`UnsupportedProtocols`** → re-send the exchange as the embedded legacy variant via `send_legacy_goodbye` (byte-for-byte the previous `DisconnectPeerX` body, including its own full `px_disconnect_timeout` window and the existing response/failure handling), and the dedicated-path deadline task stands down (`GoodbyeOutcome::FellBack`) so it cannot disconnect underneath the in-flight legacy attempt.
- **Any other failure** (dial / connection-closed / IO / timeout) → no ack is coming;  dropping the notify sender wakes the deadline task, which disconnects.  No penalty anywhere on the goodbye path — same class as the existing early-return exemption for px failures.

Both maps share the `max_px_disconnects` budget through `goodbyes_in_flight()`, which also feeds the existing pending-px gauge, so the combined bound and the metric keep their original meaning.

## Inbound path

A dedicated handler mirrors the legacy intercept exactly: process the exchange through the peer manager, send the empty ack, initiate the reciprocal disconnect.  The legacy intercept in `process_reqres_event` stays untouched so goodbyes from not-yet-upgraded peers keep working until step 9 removes them.

## Testing

- `test_network_type_protocol_names` pins the chain-namespaced protocol names (peer-compatibility contract).
- `test_peer_exchange_codec_round_trip` round-trips the exchange request and the empty ack through the dedicated codec.
- **`test_goodbye_falls_back_to_embedded_exchange_for_legacy_peer`** (new e2e, and the first cross-version harness from the issue's rollout rules): a raw libp2p swarm advertising only the legacy `/tn-primary-{chain}/0.0.1` req-res protocol, the exact wire surface of a not-yet-upgraded node, dials a target whose peer slots are filled with prune-protected committee validators.  The target prunes it, the goodbye attempt on the dedicated protocol fails `UnsupportedProtocols`, and the test asserts the goodbye arrives as the embedded legacy variant (only reachable through the fallback), carries a non-empty exchange map, and is followed by a prompt disconnect.
- `test_peer_exchange_with_excess_peers` (existing) now exercises the dedicated-protocol happy path end to end, since both sides are upgraded nodes.

## Rollout

Dedicated-protocol goodbyes and embedded goodbyes coexist;  the fallback needs no capability cache because a goodbye is a single terminal exchange (the failed probe answers the question the cache would).  Step 9 stops sending the embedded variant, drops the `From<PeerExchangeMap>` / `peer_exchange_msg()` bounds from `TNMessage`, and removes the variants at the coordinated `/0.0.2` bump . . . BCS encodes variant indices, so no slot moves before then.

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -3,7 +3,7 @@
 //! This network is used by workers and primaries to reliably send consensus messages.
 
 use crate::{
-    codec::{TNCodec, TNMessage},
+    codec::{PeerExchangeCodec, TNCodec, TNMessage},
     error::NetworkError,
     kad::KadStore,
     metrics::{PeerManagerMetrics, SwarmMetrics},
@@ -57,6 +57,7 @@ mod network_tests;
 /// - `peer_manager`: Connection management and peer scoring
 /// - `gossipsub`: Flood publishing for certificates and batches
 /// - `req_res`: Point-to-point request-response messages
+/// - `peer_exchange`: Dedicated request-response protocol for the goodbye exchange
 /// - `kademlia`: Distributed hash table for peer discovery
 /// - `stream`: Stream-based bulk data transfer for state sync
 ///
@@ -76,6 +77,13 @@ where
     pub(crate) gossipsub: gossipsub::Behaviour,
     /// The request-response network behavior.
     pub(crate) req_res: request_response::Behaviour<C>,
+    /// Dedicated request-response behavior for the peer-exchange goodbye.
+    ///
+    /// Preferred over the [`PeerExchangeMap`] variants embedded in the consensus
+    /// request enums; goodbyes fall back to the embedded variant when the peer
+    /// has not upgraded yet. The embedded variants stay on the wire until the
+    /// coordinated `/0.0.2` protocol bump.
+    pub(crate) peer_exchange: request_response::Behaviour<PeerExchangeCodec>,
     /// Used for peer discovery.
     pub(crate) kademlia: kad::Behaviour<KadStore<DB>>,
     /// Stream-based sync behavior for bulk data transfer.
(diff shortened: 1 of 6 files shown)


crates/network-libp2p/src/consensus.rs
```diff
@@ -88,22 +96,52 @@ where
...
+
+/// How a goodbye on the dedicated peer-exchange protocol resolved.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum GoodbyeOutcome {
+    /// The peer acked the exchange: safe to disconnect immediately.
+    Acked,
+    /// The peer does not support the dedicated protocol; the goodbye was re-sent
+    /// on the legacy embedded path, which owns the disconnect from here.
+    FellBack,
+}
```
grantkee, line 143: nit: move these to `types.rs`?

crates/network-libp2p/src/consensus.rs
```diff
@@ -76,6 +77,13 @@ where
     pub(crate) gossipsub: gossipsub::Behaviour,
     /// The request-response network behavior.
     pub(crate) req_res: request_response::Behaviour<C>,
+    /// Dedicated request-response behavior for the peer-exchange goodbye.
+    ///
+    /// Preferred over the [`PeerExchangeMap`] variants embedded in the consensus
+    /// request enums; goodbyes fall back to the embedded variant when the peer
+    /// has not upgraded yet. The embedded variants stay on the wire until the
+    /// coordinated `/0.0.2` protocol bump.
+    pub(crate) peer_exchange: request_response::Behaviour<PeerExchangeCodec>,
```
grantkee, line 86: This is probably the correct order since we're using the [derive macro](https://docs.rs/libp2p/latest/libp2p/swarm/trait.NetworkBehaviour.html#custom-networkbehaviour-with-the-derive-macro) - wanted to flag @MavenRain since we haven't discussed the hierarchy before

`PeerManager` first is by design since it should intercept any peer connect/disconnects

grantkee (approved):
LGTM! I like the premise. If/when we implement our own stream behavior to replace request/response, I like the idea of including this as part of the `ConnectionHandler` closing disconnect message so we don't have to poll at all for receipt

sigp/lighthouse does something similar for their custom RPC behavior: https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/handler.rs#L252-L253

crates/network-libp2p/src/consensus.rs
sstanfield: @grantkee You sure?  These appear to be private so that may just clutter up the public interface.

sstanfield (approved):
LGTM!

crates/network-libp2p/src/consensus.rs
MavenRain: Agree with @sstanfield on this point . . . will keep private for now

Issue #822, opened 2026-07-01
Review of PR #811 (MissingCertificates sync cutover): out-of-bounds panic (High) plus 5 findings


MavenRain:
## Code review (verified): MissingCertificates sync cutover

Review of **PR #811** (branch `739-missing-certificates-sync-cutover`, step 7 of #739).

- **Date**: 2026-06-30
- **Scope**: 3 commits on top of merge-base `770b8fb`: `875148b3` (step 7 cutover), `79a078db` (stream-fold responder), `b0e49792` (stream-fold requester).
- **Changed files (8)**: `certificate_fetcher.rs`, `network/handler.rs`, `network/mod.rs`, `network/sync_codec.rs`, `state_sync/cert_collector.rs`, and 3 test files.
- **Method**: raw findings from a diff review, then each non-informational finding independently re-derived by a separate agent that saw only the Claim, Key Question, and Relevant Files, never the original reasoning.

### Summary

All four non-informational findings independently CONFIRMED (0 false positives, all HIGH confidence).  One High: a pre-existing out-of-bounds index panic in the requester invalid-cert path that this PR promotes to the sole code path.  Findings 3 and 4 had their verifier-proposed severities corrected back down to Low on honest impact calibration (facts confirmed, impact over-rated).

| # | Title | Severity (initial) | Verified severity | Verdict | Confidence |
|---|-------|--------------------|-------------------|---------|------------|
| 1 | Out-of-bounds index panic in fetch peer-error logging | High | High | CONFIRMED | HIGH |
| 2 | Per-response accept cap raised ~64x with unbounded requester fan-out | Medium | Medium | CONFIRMED (design decision: confirm/document) | HIGH |
| 3 | Each certificate BCS-encoded twice on the responder hot path | Low | Low | CONFIRMED | HIGH |
| 4 | `get_bounds` round-arithmetic overflow on untrusted input (adjacent) | Low | Low | CONFIRMED | HIGH |
| 5 | Vestigial `max_response_size` field on `MissingCertificatesRequest` | Informational | Informational | Confirmed (observation) | n/a |
| 6 | No test coverage for the invalid-cert fan-out path | Informational | Informational | Confirmed (observation) | n/a |

---

### 1. Out-of-bounds index panic in fetch peer-error logging

- **Verified severity**: High. **Verdict**: CONFIRMED. **Confidence**: HIGH.
- **Category**: Error handling / Denial of Service
[...]

PR #826, opened 2026-07-02
fix(primary): out-of-bounds peer index when logging invalid fetched certs (#822)

MavenRain:
**Addresses**: #822 (correctness findings 1 and 4; adds the finding-6 regression test)

## Title

fix(primary): out-of-bounds peer index when logging invalid fetched certs (#822)

## Description

#811 (the MissingCertificates sync cutover) is now on `main`, so the findings from its
review (#822) are live.  This PR fixes the two correctness defects and adds the missing
regression coverage:

- [x] **Out-of-bounds panic in `fetch_from_peers` error logging** (#822 finding 1, High).
  The invalid-cert error closure logged `peers[peer_index]`, but `peer_index` is the index
  of the *next peer to spawn*, not the responder.  With a single fetchable peer it already
  equals `peers.len()` when the first response arrives, so a peer returning a genesis-state
  certificate panicked the fetch task, and certificate catch-up stalled on every retry.
  Even in bounds the logged peer was the wrong one.  The responding `BlsPublicKey` is now
  threaded through the results channel and logged directly; the index access is gone.
- [x] **`get_bounds` skip-round overflow on untrusted input** (#822 finding 4, Low).
  `exclusive_lower_bound + r as Round` on a peer-supplied RoaringBitmap wrapped in release
  and panicked in debug.  Now `checked_add`, surfacing as `PrimaryNetworkError::InvalidRequest`
  through the existing malformed-request path (and dropping the `as` cast).
- [x] **Regression tests** (#822 finding 6).
  - `test_invalid_cert_from_sole_peer_does_not_panic` drives `fetch_from_peers` (via the
    existing `MockFetcher` seam) with a single peer returning a genesis-state certificate:
    pre-fix it panics on `peers[peer_index]`;  post-fix it stays pending and the test's
    timeout elapses cleanly.
  - `test_missing_certs_request_skip_round_overflow` feeds a skip round that overflows the
    `u32` round space and asserts `InvalidRequest` rather than a wrap/panic.

### Findings deliberately not in this PR

- **Finding 2** (per-response 64.5 MiB accept cap x concurrent requester fan-out, Medium):
  a design decision about the requester's memory envelope, not a clear bug.  Best resolved by
  the maintainers (document the worst case, or cap concurrent in-flight fetches).  I leave it for a
  separate discussion to keep this PR to unambiguous correctness fixes.
- **Finding 3** (each served certificate BCS-encoded twice on the responder hot path, Low):
  a safe optimization in `sync_codec.rs`;  worth a focused follow-up rather than mixing an
  optimization into a correctness fix.
- **Finding 5** (`max_response_size`): on `main` the sync path now sets it to `0` with an
  explanatory comment, so it is benign;  no change needed.

### DIFF
--- a/crates/consensus/primary/src/certificate_fetcher.rs
+++ b/crates/consensus/primary/src/certificate_fetcher.rs
@@ -476,7 +476,7 @@ async fn fetch_from_peers<F: MissingCertFetcher>(
 
     loop {
         tokio::select! {
-            Some(result) = rx_results.recv() => {
+            Some((peer, result)) = rx_results.recv() => {
                 if let Ok(certificates) = result {
                     if !certificates.is_empty() {
                         debug!(target: "primary::cert_fetcher",
@@ -488,7 +488,7 @@ async fn fetch_from_peers<F: MissingCertFetcher>(
                             validate_fetched_certificate(cert).map_err(|e| {
                                 error!(
                                     target: "primary::cert_fetcher",
-                                    peer=?peers[peer_index],
+                                    peer=?peer,
                                     "cert fetch received invalid genesis cert from peer"
                                 );
                                 e.into()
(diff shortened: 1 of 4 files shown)


grantkee (approved):
Great solution

PR #837, opened 2026-07-03
(feat) CLI keytool set rpc

### LINKED ISSUE #901: feat: CLI `keytool` has no way to set a node's advertised JSON-RPC endpoint
## Problem

`NodeInfo` already carries an optional worker JSON-RPC endpoint (`p2p_info.worker.rpc`, type `Option<RpcInfo>`) that the node reads and validates at startup, then hands to the worker network so peers can discover it over Kademlia. This lets wallets and dapps find where to submit transactions without out-of-band configuration.

Nothing populates that field, though. There's no `keytool` support for setting it — not at key-generation time and not afterward. An operator who wants to advertise an RPC endpoint has to hand-edit `node-info.yaml` directly, with no validation against the scheme checks the node applies at startup (`http`/`https` for the HTTP endpoint, `ws`/`wss` for the optional WebSocket one). A typo or wrong scheme there surfaces only when the node starts and the peers reject the advertisement, not when the operator makes the change.

Separately, every `keytool` subcommand currently prompts for the BLS key passphrase, even ones that never touch the BLS key. A command that only edits public config in `node-info.yaml` doesn't need it, and prompting anyway is friction with no security benefit.

## Solution

Introduce `keytool set-rpc` to set or clear the worker RPC endpoint in an existing `node-info.yaml`. It should be a config-only edit — no keys read, no passphrase required — that loads the existing file, applies the same scheme validation node startup runs, and writes the result back. Since it depends on an existing identity, it should fail with a clear "generate keys first" hint if `node-info.yaml` isn't present yet, rather than silently creating one with a default identity.

`keytool generate validator|observer` should expose equivalent `--rpc-http`/`--rpc-ws` flags so an operator standing up a new node can set the endpoint in the same step as key generation, instead of running `generate` followed by a separate `set-rpc`. Both entry points should share one validation routine so the CLI enforces identical rules regardless of which command is used, and `generate` should validate before minting any keys so a bad endpoint fails fast without leaving partial state on disk.

Because `set-rpc` never touches the BLS key, the CLI's passphrase-gating logic needs to distinguish config-only keytool subcommands from key-touching ones (`generate validator|observer`, `generate pop`) and skip the prompt for the former.

grantkee:
- Add `keytool set-rpc --http URL [--ws URL] | --clear` to set or clear `p2p_info.worker.rpc` in an existing `node-info.yaml`. Errors with a "generate keys first" hint if `node-info.yaml` is missing, instead of creating one.
- Add `--rpc-http`/`--rpc-ws` to `keytool generate validator|observer` so new nodes can set the endpoint during key generation in one step. Validation runs before any keys are minted, so a bad scheme leaves no `node-info.yaml` behind.
- Both commands validate through a shared `build_worker_rpc` helper (`http`/`https` for `--http`, `ws`/`wss` for `--ws`), matching the check node startup applies, so a bad scheme is caught at CLI time instead of at peer advertisement.
- `set-rpc` is a config-only edit — no keys read, BLS passphrase not required. `KeyArgs::needs_passphrase()` now distinguishes config-only keytool subcommands from key-touching ones (`generate validator|observer`, `generate pop`), and the passphrase-source gating in `passphrase.rs` checks it before prompting.
- Box `Commands::Keytool`'s payload to keep the `Commands` enum size down now that `KeyArgs` carries more variants.
- README: document `set-rpc` usage and the `--clear` behavior.

closes #901

### DIFF
--- a/crates/telcoin-network-cli/src/keytool/set_rpc.rs
+++ b/crates/telcoin-network-cli/src/keytool/set_rpc.rs
@@ -27,7 +27,7 @@ use url::Url;
 /// Build a validated worker `RpcInfo` from the `--http`/`--ws` (or `--rpc-http`/`--rpc-ws`) flags.
 ///
 /// Returns `Ok(None)` when no HTTP endpoint was provided (nothing to advertise). Applies the same
-/// validation node startup runs (`node.rs:624`), so a bad scheme/length fails at CLI time. Shared by
+/// validation node startup runs (`node.rs`), so a bad scheme/length fails at CLI time. Shared by
 /// `set-rpc` and `generate validator|observer`.
 pub(crate) fn build_worker_rpc(
     http: Option<Url>,
--- a/crates/telcoin-network-cli/src/keytool/set_rpc.rs
+++ b/crates/telcoin-network-cli/src/keytool/set_rpc.rs
@@ -96,7 +96,7 @@ impl SetRpcArgs {
         } else {
             // clap guarantees --http is present unless --clear.
             let rpc = build_worker_rpc(self.http.clone(), self.ws.clone())?
-                .expect("clap requires --http unless --clear");
+                .ok_or_else(|| eyre::eyre!("clap requires --http unless --clear"))?;
             if node_info.p2p_info.worker.rpc.is_some() {
                 warn!(target: "tn::keytool", "overwriting existing worker RPC endpoint");
             }
--- a/crates/telcoin-network-cli/src/keytool/set_rpc.rs
+++ b/crates/telcoin-network-cli/src/keytool/set_rpc.rs
@@ -1,0 +1,98 @@
+//! `set-rpc` subcommand: set or clear the worker JSON-RPC endpoint advertised
+//! in `node-info.yaml`.
+//!
+//! A node advertises an optional JSON-RPC endpoint to peers over kademlia so
+//! wallets/dapps can discover where to submit transactions. That endpoint lives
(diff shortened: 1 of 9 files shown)


grantkee:
Related to #820

crates/telcoin-network-cli/src/keytool/set_rpc.rs
```diff
@@ -0,0 +1,119 @@
...
+        if self.clear {
+            if node_info.p2p_info.worker.rpc.take().is_some() {
+                warn!(target: "tn::keytool", "cleared existing worker RPC endpoint");
+            } else {
+                info!(target: "tn::keytool", "no worker RPC endpoint set; nothing to clear");
+            }
+        } else {
+            // clap guarantees --http is present unless --clear.
+            let rpc = build_worker_rpc(self.http.clone(), self.ws.clone())?
+                .expect("clap requires --http unless --clear");
```
MavenRain: We're within an `eyre::Result<()>` expression, so we should be able to write
```
.ok_or_else(|| eyre::eyre!("--http is required unless --clear"))?
```
instead

MavenRain (dismissed):
One suggestion

MavenRain (approved):

crates/telcoin-network-cli/src/keytool/set_rpc.rs
grantkee: Good catch. I agree and think this should be consistent

PR #840, opened 2026-07-03
feat(node):  seed base fee from chain on sync/restart instead of recomputing

### LINKED ISSUE #839: Basefee must be seeded from chain on sync and restart
## Problem

Base fee is consensus-affecting: every node must arrive at the same per-worker base fee for a given epoch, or block validation diverges across the network. The current epoch-boundary logic computes the next epoch's base fee locally from each worker's accumulated gas usage (EIP-1559 style adjustment) whenever an epoch closes.

That computation is only valid for the node that lived through the epoch as the active producer and accumulated a complete gas usage record for it. A node that is syncing from genesis, replaying missed consensus, or restarting mid-epoch does not have that record — its `GasAccumulator` starts empty. If such a node runs the same forward computation, it derives a different base fee than the one already committed to the chain by the original producer, silently diverging from the canonical value at epoch entry.

This is a correctness gap in the interaction between epoch transition and the recently-introduced EIP-1559 base-fee adjustment in #796 : the boundary logic doesn't yet distinguish "I lived through this epoch and can compute the next fee" from "I'm entering an epoch whose blocks already exist on-chain." 

## Solution

At epoch entry, the node should check whether the canonical chain tip is already inside the epoch being entered. If it is, the node is catching up (syncing or restarting) rather than producing live, so it should adopt each worker's most recent on-chain `base_fee_per_gas` directly from the executed blocks in that epoch's range, rather than recomputing anything.

If the tip is still in the previous epoch, the node is the live producer crossing the boundary in real time; in that case the value computed by the existing forward adjustment at epoch close should be kept as-is, since no chain data exists yet to seed from.

This requires identifying which worker produced a given block (recoverable from a field already used to encode worker id) and locating the block range for the epoch being entered, so per-worker fees can be read back from the chain. Workers that produced no block in that range are left with whatever value they already hold. The distinction between "live producer" and "catching-up node" should be made from context already available in the epoch-close path, without adding new state.

The design preserves a single base-fee-from-chain invariant: only the node that just lived through an epoch computes the next value forward; every other node — sync, restart, replay — reads it from the chain.

grantkee:
- At epoch entry, compare the canonical tip's epoch to the epoch being entered. If they match, the node is catching up (sync/restart) and adopts each worker's latest on-chain `base_fee_per_gas` instead of recomputing; if they differ, it's the live producer crossing the boundary and keeps the value `adjust_base_fees` already computed.
- Added `worker_id_from_header`/`latest_base_fee_per_worker` in `crates/node/src/manager/node.rs` to recover per-worker base fee from executed headers (worker id is read from the low 16 bits of `difficulty`, matching `GasAccumulator::inc_block` encoding).
- `close_epoch` now only calls `adjust_base_fees` on the live boundary path (`shutdown_consensus.is_some()`); restart/leftover-drain paths (`None`) skip it and rely on the new epoch-start seeding instead.
- Documented the base-fee-from-chain invariant in `crates/types/src/gas_accumulator.rs` module docs.
- New unit tests cover: seed-adopts-chain-fee when tip is in the entered epoch, keep-computed-value for the live producer, and untouched-for-workers-with-no-blocks-this-epoch.

closes #839

### DIFF
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -179,8 +179,7 @@ pub async fn catchup_accumulator(
 
             // difficulty contains the worker id and batch index:
             // `U256::from(payload.batch_index << 16 | payload.worker_id as usize)`
-            let lower64 = current.difficulty.into_limbs()[0];
-            let worker_id = (lower64 & 0xffff) as u16;
+            let worker_id = worker_id_from_header(&current);
             gas_accumulator.inc_block(worker_id, gas, limit);
         }
 
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -21,7 +21,7 @@ use crate::{
 };
 use eyre::{eyre, WrapErr as _};
 use state_sync::{request_missing_packs, spawn_fetch_consensus, spawn_fetch_recent_consensus};
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use tn_config::{Config, ConfigFmt, ConfigTrait as _, KeyConfig, NetworkConfig, TelcoinDirs};
 use tn_network_libp2p::{types::NetworkEvent, ConsensusNetwork};
 use tn_primary::{network::PrimaryNetworkHandle, ConsensusBusApp, NodeMode, QueChannel};
@@ -30,8 +30,8 @@ use tn_storage::{consensus::ConsensusChain, open_db, DatabaseType};
 use tn_types::{
     deconstruct_nonce, gas_accumulator::GasAccumulator, BlsPublicKey, BootstrapServer, Committee,
     ConsensusHeader, ConsensusHeaderDigest, ConsensusNumHash, ConsensusOutput,
-    Database as TNDatabase, EngineUpdate, Epoch, Notifier, TaskError, TaskManager, TaskSpawner,
-    TimestampSec, DEFAULT_WORKER_ID, MIN_PROTOCOL_BASE_FEE,
+    Database as TNDatabase, EngineUpdate, Epoch, Notifier, SealedHeader, TaskError, TaskManager,
+    TaskSpawner, TimestampSec, WorkerId, DEFAULT_WORKER_ID, MIN_PROTOCOL_BASE_FEE,
 };
 use tn_worker::{WorkerNetworkHandle, WorkerRequest, WorkerResponse};
 use tokio::sync::mpsc;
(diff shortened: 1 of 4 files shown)


grantkee:
This is a stacked PR for review. I will rebase onto main once the upstream PR #796 is merged

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -573,12 +618,37 @@ fn adjust_base_fees(gas_accumulator: &GasAccumulator) {
...
+    entered_epoch: Epoch,
+    tip_epoch: Epoch,
+    chain_fees: &HashMap<WorkerId, u64>,
+    gas_accumulator: &GasAccumulator,
+) {
+    if tip_epoch != entered_epoch {
+        // new epoch - calculate fees from accumulated epoch gas
+        return;
+    }
+    for worker_id in 0..gas_accumulator.num_workers() as u16 {
```
MavenRain, line 645: A node that restarts or replays at an epic boundary will fall to here also?

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -137,6 +141,38 @@ where
...
+                let chain_fees = if tip_epoch == committee.epoch() {
+                    // The range END (`tip.number`) comes from `finalized_header()` above, while the
+                    // range START (`epoch_info.blockHeight`) comes from the canonical-tip epoch
+                    // state. These reference the same block under this node's instant BFT finality
+                    // (the payload builder finalizes the header it just made canonical), so the
+                    // range is well-formed. If finality ever lags the canonical tip, pin both ends
+                    // to one header (the same dual-reference pattern lives in catchup_accumulator).
+                    let epoch_state = reth_env.epoch_state_from_canonical_tip()?;
+                    let blocks = reth_env
+                        .blocks_for_range(epoch_state.epoch_info.blockHeight..=tip.number)?;
```
MavenRain, line 163: This pair is a canonical-tip start with a finalized-tip end.  If the canonical-tip moves to the next epoch while the finalized tip lags, so that `blockHeight > tip.number`, then `blocks_for_range` returns empty, `chain_fees` is empty, and then every worker's seeding is skipped silently.

crates/node/src/manager/node.rs
```diff
@@ -196,6 +195,33 @@ pub async fn catchup_accumulator(
...
+///
+/// `headers` are the executed reth blocks for the current epoch (epoch-start height..=tip), in
+/// ascending block-number order, so the last header seen for a worker is its latest block. The
+/// worker id is read from each header's `difficulty` field (lower 16 bits, matching how
+/// [`GasAccumulator::inc_block`] callers encode it). Workers that produced no block in the range
+/// are absent from the returned map.
+///
+/// Used to seed per-worker base fees from the chain on sync and restart, preserving the
+/// base-fee-from-chain invariant (see the [`tn_types::gas_accumulator`] module docs).
+pub(crate) fn latest_base_fee_per_worker(headers: &[SealedHeader]) -> HashMap<WorkerId, u64> {
```
MavenRain, line 214: Assumes ascending order (last `Some(_)` wins), which could be a footgun when fees begin to vary intra-epoch

MavenRain (dismissed):
I don't _think_ anything is blocking.

crates/node/src/manager/node/run_epoch.rs
grantkee: Yes, this is a temp/bridging solution. Ultimately, the worker configuration is managed on chain and the closing epoch state dictates the next epoch. This works because the protocol only supports one worker for now

crates/node/src/manager/node/run_epoch.rs
grantkee: Good feedback. I will address this in a follow up. This might be an existing bug

crates/node/src/manager/node.rs
grantkee: I will revisit in a follow up

MavenRain (approved):
Looking good

PR #842, opened 2026-07-03
feat(node): restore base fee per worker on catchup

### LINKED ISSUE #841: Catchup only restores worker 0's base fee, breaking the base-fee-from-chain invariant for other workers
## Problem

`catchup_accumulator` runs at node startup to rebuild in-memory gas-accumulator state (base fees, gas stats, leader counts) from the on-chain record after a restart. Today it restores the base fee for a single hardcoded worker (worker 0) from the finalized reth header, and leaves every other worker's base fee at whatever default the accumulator was constructed with.

This was fine when the network only ran one worker per primary, but it breaks the base-fee-from-chain invariant established for multi-worker configurations: each worker's EIP-1559 base fee must be derived from that worker's own chain history, not shared across workers. After a node restart in a multi-worker setup, only worker 0 recovers its correct base fee — every other worker resumes with a stale or default value instead of the value implied by its latest executed block. This can cause a restarted node's fee computation to diverge from what it would have computed had it never restarted, which risks disagreement with peers that stayed online.

## Solution

`catchup_accumulator` should restore the base fee for every worker, not just worker 0. For the range of reth blocks being replayed during catchup, the code should determine each worker's most recent block and read that block's base fee, then seed each worker's entry in the gas accumulator accordingly. Workers that produced no block in the replayed range keep their existing (default/minimum) value, since there is no on-chain data to restore from.

This requires walking the block range once to group blocks by worker (using the existing worker-id-from-header extraction already used for gas-stat accumulation) and keeping only the latest base fee seen per worker, then applying those per-worker base fees to the accumulator before the gas-stat accumulation loop runs. The change is confined to the catchup path in the node manager and does not affect steady-state fee computation.

grantkee:
- Add `latest_base_fee_per_worker`, which walks the replayed block range, extracts each block's worker id via `worker_id_from_header`, and keeps the latest base fee seen per worker
- `catchup_accumulator` now applies this per-worker map to the gas accumulator instead of only setting worker 0's base fee from the finalized header
- Adds a regression test (`test_catchup_accumulator`) that poisons the recovered accumulator's base fee before catchup and asserts catchup overwrites it with the chain-derived value, rather than leaving the stale one in place

closes #841

### DIFF
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -140,8 +140,8 @@ pub(crate) struct EpochManager<P, DB> {
 /// [`tn_types::gas_accumulator`] for the full picture). It runs once at startup, before
 /// execution resumes, and performs the following:
 ///
-/// 1. **Base fee** — restores each worker's base fee from its latest on-chain block this epoch
-///    (via [`latest_base_fee_per_worker`]), preserving the base-fee-from-chain invariant.
+/// 1. **Base fee** — restores each worker's base fee from its latest on-chain block this epoch (via
+///    [`latest_base_fee_per_worker`]), preserving the base-fee-from-chain invariant.
 /// 2. **Gas stats** — iterates every reth block from the epoch's start height through the finalized
 ///    tip, extracting the worker id from each block's `difficulty` field and calling
 ///    [`GasAccumulator::inc_block`] to rebuild per-worker gas totals.
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -31,7 +31,7 @@ use tn_types::{
     deconstruct_nonce, gas_accumulator::GasAccumulator, BlsPublicKey, BootstrapServer, Committee,
     ConsensusHeader, ConsensusHeaderDigest, ConsensusNumHash, ConsensusOutput,
     Database as TNDatabase, EngineUpdate, Epoch, Notifier, SealedHeader, TaskError, TaskManager,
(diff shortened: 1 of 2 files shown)


grantkee:
This is a stacked PR but I will rebase to main after upstream PR #840 merged

crates/node/src/manager/node.rs
```diff
@@ -160,25 +160,24 @@ pub async fn catchup_accumulator(
...
         let nonce: u64 = block.nonce.into();
         let (last_executed_epoch, last_executed_round) = deconstruct_nonce(nonce);
 
         let blocks =
             reth_env.blocks_for_range(epoch_state.epoch_info.blockHeight..=block.number)?;
 
+        // Restore each worker's base fee from its latest on-chain block this epoch (the
+        // base-fee-from-chain invariant). Workers that produced no block keep the default (MIN).
+        // Generalizes the previous worker-0-only restore now that there can be multiple workers.
+        for (worker_id, base_fee) in latest_base_fee_per_worker(&blocks) {
```
MavenRain, line 172: Would there be any issue also seeding workers that have not produced a block in the current epoch?

MavenRain (dismissed):
Just a question

crates/node/src/manager/node.rs
grantkee: Yes, good catch. I kept the naive approach in this PR to keep it small since there can only be 1 worker for now but this is definitely a gap in logic. Addressed in #852

MavenRain (approved):
LGTM

PR #844, opened 2026-07-03
feat(node): activate per-worker base fees from WorkerConfigs

### LINKED ISSUE #843: Activate per-worker base fees from WorkerConfigs at epoch close
## Problem

The per-worker base fee adjustment at epoch close is currently wired to a hardcoded `u64::MAX` gas target, which makes the mechanism permanently inert: since the target is unreachable, the EIP-1559-style formula can only ratchet fees downward, and every worker floors at `MIN_PROTOCOL_BASE_FEE`. This was intentional scaffolding — the deterministic seam (reading gas usage, computing the next fee, writing it back at the epoch boundary) was built first, ahead of the governance-configurable piece.

With only a hardcoded target, there is no way for governance to actually set a fee strategy per worker. Workers can't be given a real gas target for EIP-1559-style adjustment, nor can they be pinned to a static fee — both are legitimate strategies operators may want, and neither is reachable today.

## Solution

`close_epoch` should read each worker's fee strategy from the `WorkerConfigs` contract at the canonical tip (the epoch's closing block, since this runs after `wait_for_consensus_execution`) and apply it when computing the next base fee, instead of always using the `u64::MAX` placeholder.

The fee strategy should be represented as an enum with two variants to start: an EIP-1559 variant carrying a target gas value (feeding the existing adjustment formula), and a static variant that pins the fee to a governance-set value regardless of gas usage. The number of configs read should be validated against the on-chain worker count, so a mismatch between what the node expects and what the contract reports is caught rather than silently misapplied.

Because a contract read can fail (revert, or a worker-count mismatch), the failure mode matters: this should propagate as an error out of epoch close rather than being swallowed or guessed around, since guessing a fee risks divergence across committee members. On any existing chain, the contract's default strategy is the same `Eip1559` target used today, so the change is inert until governance actively sets a different per-worker target — no behavior changes for chains that don't touch the new config.

grantkee:
- `adjust_base_fees` now reads one `WorkerFeeConfig` per worker via `RethEnv::get_worker_fee_configs` and dispatches through the new `next_base_fee_for_config` helper: `Eip1559 { target_gas }` feeds `compute_next_base_fee_eip1559` with the real target instead of the old hardcoded `u64::MAX`; `Static { fee }` pins the fee directly.
- Genesis/default strategy is still `Eip1559 { target_gas: u64::MAX }`, so existing chains see no behavior change until governance sets real per-worker targets — activation is inert by default.
- Fail-open on contract-read failure or `numWorkers()` drift: falls back to a fixed `FALLBACK_BASE_FEE` (1 GWEI) for worker 0 instead of aborting epoch close, since every committee member hitting the same failure independently lands on the same deterministic value.
- Tests reworked around `next_base_fee_for_config` directly (inert-at-min, gas-vs-target movement, static pin) plus a fixed-value check on `FALLBACK_BASE_FEE` for regression

closes #843

### DIFF
--- a/crates/node/src/manager/node/run_epoch.rs
+++ b/crates/node/src/manager/node/run_epoch.rs
@@ -27,7 +27,7 @@ use tn_primary::ConsensusBus;
 use tn_reth::RethEnv;
 use tn_storage::{certificate_pack::CertificatePack, tables::OurNodeBatchesCache};
 use tn_types::{
-    gas_accumulator::{compute_next_base_fee_eip1559, GasAccumulator},
+    gas_accumulator::{compute_next_base_fee_eip1559, GasAccumulator, WorkerFeeConfig},
     BlsPublicKey, Committee, ConsensusHeaderDigest, ConsensusOutput, Database as TNDatabase, Epoch,
     EpochRecord, Notifier, TaskJoinError, TaskManager, TaskSpawner, TnReceiver, WorkerId,
 };
@@ -208,7 +208,7 @@ where
             if let Some(target_hash) = replay.take_epoch_close_hash() {
                 // If things go down at exactly the wrong time we might have to replay the epoch end
                 // so account for that.
-                self.close_epoch(None, &gas_accumulator, target_hash).await?;
+                self.close_epoch(None, engine, &gas_accumulator, target_hash).await?;
                 self.clear_consensus_db_for_next_epoch()?;
                 return Ok(RunEpochMode::NewEpoch);
             }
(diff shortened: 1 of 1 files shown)


grantkee:
This is a stacked PR for review - I will rebase on main once upstream PR #842 is merged

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -597,30 +603,73 @@ where
...
+                base_fee.set_base_fee(next_base_fee);
+            }
+        }
+        Err(e) => {
+            warn!(
+                target: "epoch-manager",
+                ?e,
+                "failed to read worker fee configs at epoch close - falling back to {FALLBACK_BASE_FEE} wei for worker 0"
+            );
+            gas_accumulator.base_fee(0).set_base_fee(FALLBACK_BASE_FEE);
```
MavenRain, line 653: Why should a multi-worker config only pin worker 0 when the read fails?

MavenRain (dismissed):
Just a question

crates/node/src/manager/node/run_epoch.rs
grantkee: Worker 0 is the only one with basefee adjustments, but I've since opted for removing this fallback. If the worker configs fail to read on-chain, the protocol will use the same settings for the next epoch in #846

MavenRain (approved):
Good to go

PR #846, opened 2026-07-04
feat(node): sync GasAccumulator worker count from on-chain `WorkerConfigs`

### LINKED ISSUE #845: GasAccumulator worker count must track on-chain WorkerConfigs, not a fixed startup value
## Problem

The `GasAccumulator` that tracks per-worker gas totals and base fees is currently sized once at node startup (`GasAccumulator::new(1)`) and never resized. Worker count, however, is governance-controlled through the `WorkerConfigs` contract, which already supports `setNumWorkers`. Once governance grows or shrinks the worker set, the in-memory accumulator silently disagrees with the chain: new workers have no fee/gas slot to write into, and `adjust_base_fees` at epoch close reads worker fee configs while asserting the caller's stale worker count against `numWorkers()` on-chain, erroring out entirely on any drift.

This leaves worker-count changes effectively unusable in practice — any committee member that restarts, syncs from genesis, or crosses an epoch boundary after `setNumWorkers` is called ends up with a `GasAccumulator` that no longer matches the contract, either panicking on an out-of-range worker id during block execution or hitting the drift error at epoch close. There's no path for a running network to safely add or remove workers.

Additionally, the previous fail-open behavior at epoch close (pinning worker 0 to a hardcoded 1 GWEI fallback when the `WorkerConfigs` read fails) doesn't account for a multi-worker accumulator — it only ever touches worker 0's fee, leaving other workers' fees stale and doesn't address the underlying count-drift problem.

## Solution

`GasAccumulator` should support resizing its worker slot list in place, so that all existing clones (held by the engine, and the EVM config's rewards counter) observe the new size without needing to be reconstructed. This requires making the inner slot storage lock-based rather than a plain `Arc<Vec<_>>`, since resizing must be safe against concurrent reads from `inc_block`/`get_values`/`base_fee`.

The node should read the on-chain worker count and sync the accumulator to it at two points:

- **Startup**, before catchup recovery runs, so that catchup's per-worker restore writes land in correctly sized slots.
- **Epoch entry**, before fee seeding and before replay drives `inc_block`, using the count as of the entering epoch's first block's parent (the previous epoch's closing block) — the same block every node (live producer, restarting node, syncing node) converges on regardless of timing, and one that is immune to mid-epoch `setNumWorkers` writes since those only take effect at the next boundary.

At epoch close, `adjust_base_fees` should resize the accumulator to the worker count returned by the contract read itself, rather than asserting it against a caller-supplied count — this lets newly added workers get their configured fee strategy applied immediately instead of defaulting to the protocol minimum until the next epoch-entry sync.

Both sync points should fail open: if the contract read fails (older chains without the contract deployed, or a transient read error), the accumulator keeps its current size and per-worker state rather than aborting the epoch close or crashing at startup. Because every committee member fails open identically off the same on-chain state, this preserves determinism without requiring a fixed fallback fee.

grantkee:
- `GasAccumulator`'s slot list is now `Arc<RwLock<Vec<Accumulated>>>` instead of `Arc<Vec<_>>`, with a new `set_num_workers` that resizes in place (grow appends default slots, shrink truncates) so existing clones (engine, rewards counter) see the change without reconstruction
- Added `sync_num_workers_from_chain`, called at startup (before `catchup_accumulator`) and at every epoch entry (before fee seeding/replay), reading the `WorkerConfigs` count at the entering epoch's first block's parent
- `adjust_base_fees` now resizes the accumulator to the count returned by `get_worker_fee_configs()` directly instead of asserting a caller-supplied count against on-chain `numWorkers()` — new workers get their configured fee applied immediately at close rather than waiting for the next epoch-entry sync
- Removed `get_worker_fee_configs`'s `num_workers` drift-check parameter and the hardcoded 1 GWEI fallback for worker 0; both sync points now fail open by keeping current state on any read failure
- Added IT tests covering: accumulator grow/shrink to on-chain count, fail-open behavior when `WorkerConfigs` is absent, and worker-count reads landing on the correct epoch-boundary block

closes #845

### DIFF
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -194,6 +194,75 @@ pub async fn catchup_accumulator(
     Ok(())
 }
 
+/// Resize the [`GasAccumulator`] to the on-chain worker count for the epoch whose first block is
+/// `epoch_first_block`.
+///
+/// The `WorkerConfigs` contract is the absolute source of truth for the worker count, and the
+/// count for epoch E is its state at block `epoch_first_block - 1` - E's first block's parent,
+/// i.e. the previous epoch's closing block (`saturating_sub` makes epoch 0 read genesis state).
+/// That block is identical for a live producer at the boundary, a restarting node, and a
+/// mid-epoch syncing node, and it is immune to mid-epoch `setNumWorkers` writes, which by design
+/// only take effect at the next boundary.
+///
+/// FAIL-OPEN: any read failure (contract absent on older chains, state unavailable) logs a
+/// warning and keeps the accumulator's current size, so the node behaves exactly as it did
+/// before worker counts were chain-derived.
+///
+/// Callers must only invoke this while no consensus output is executing (startup before
+/// [`catchup_accumulator`], epoch entry before replay) - see
+/// [`GasAccumulator::set_num_workers`].
+pub fn sync_num_workers_from_chain(
+    reth_env: &RethEnv,
+    gas_accumulator: &GasAccumulator,
+    epoch_first_block: u64,
+) {
+    let read_block = epoch_first_block.saturating_sub(1);
+    let header = match reth_env.sealed_header_by_number(read_block) {
+        Ok(Some(header)) => header,
+        Ok(None) => {
+            warn!(
+                target: "epoch-manager",
+                read_block,
+                "no header at epoch start parent while syncing worker count - keeping current count"
+            );
+            return;
+        }
+        Err(e) => {
+            warn!(
+                target: "epoch-manager",
+                ?e,
+                read_block,
+                "failed to read header while syncing worker count - keeping current count"
+            );
+            return;
+        }
+    };
+
+    match reth_env.get_worker_fee_configs_at_block(header.hash()) {
+        Ok((num_workers, _configs)) => {
+            let current = gas_accumulator.num_workers();
+            if current != num_workers {
+                info!(
+                    target: "epoch-manager",
+                    current,
+                    on_chain = num_workers,
+                    read_block,
+                    "syncing GasAccumulator worker count to on-chain WorkerConfigs"
+                );
+            }
+            gas_accumulator.set_num_workers(num_workers);
+        }
+        Err(e) => {
+            warn!(
+                target: "epoch-manager",
+                ?e,
+                read_block,
+                "failed to read WorkerConfigs while syncing worker count - keeping current count"
+            );
+        }
+    }
+}
+
 /// Worker id encoded in a header's `difficulty` (low 16 bits of `batch_index << 16 | worker_id`).
 pub(crate) fn worker_id_from_header(header: &SealedHeader) -> WorkerId {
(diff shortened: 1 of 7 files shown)


grantkee:
This is a stacked PR but I will rebase onto main when #844 is merged

crates/node/src/manager/node.rs
```diff
@@ -194,6 +194,75 @@ pub async fn catchup_accumulator(
...
+/// i.e. the previous epoch's closing block (`saturating_sub` makes epoch 0 read genesis state).
+/// That block is identical for a live producer at the boundary, a restarting node, and a
+/// mid-epoch syncing node, and it is immune to mid-epoch `setNumWorkers` writes, which by design
+/// only take effect at the next boundary.
+///
+/// FAIL-OPEN: any read failure (contract absent on older chains, state unavailable) logs a
+/// warning and keeps the accumulator's current size, so the node behaves exactly as it did
+/// before worker counts were chain-derived.
+///
+/// Callers must only invoke this while no consensus output is executing (startup before
```
MavenRain, line 211: This safety invariant is only mentioned in code comments across node startup, epoch entry, and epoch close.  Now, `set_num_workers` can shrink the worker vector in place.

Consider refactor to not panic on an out-of-bounds access upon shrink, or add a synchronization primitive to supplement the commented invariant.

crates/node/src/manager/node.rs
```diff
@@ -194,6 +194,75 @@ pub async fn catchup_accumulator(
...
+    };
+
+    match reth_env.get_worker_fee_configs_at_block(header.hash()) {
+        Ok((num_workers, _configs)) => {
+            let current = gas_accumulator.num_workers();
+            if current != num_workers {
+                info!(
+                    target: "epoch-manager",
+                    current,
+                    on_chain = num_workers,
```
MavenRain, line 248: `set_num_workers` clasps via `max(1)`.  If a contract every reports 0 workers (e.g. due to malformed `WorkerConfigs`), then the log will claim 0 while the accumulator is set to 1.

Consider a distinct warning on the 0 case before clamping.

MavenRain (dismissed):
No blockers . . . a couple of suggestions

crates/node/src/manager/node.rs
grantkee: Agreed, this should never happen but a warning is worth logging for debugging

crates/node/src/manager/node.rs
grantkee: Good call - I addressed this in the final stacked PR https://github.com/Telcoin-Association/telcoin-network/pull/852/changes/c6a80d5564627dd8a6f47daf11bebec48766e5dd

I opted for fail-open to prioritize liveness

grantkee: I think it's acceptable as is for this PR because the protocol has strict 1 worker limitations in other parts of the code (ie - network-types, node/manager, etc)

sstanfield (approved):
LGTM!  Sad we had to add a lock but it looks like we don't have a choice.

PR #848, opened 2026-07-04
refactor(cli): extract BLS passphrase handling into telcoin-network-cli

### LINKED ISSUE #847: refactor: BLS passphrase handling is trapped in the stock binary's main.rs
## Problem

The logic that resolves a validator's BLS keystore passphrase lives entirely inside `bin/telcoin-network/src/main.rs`. Two helpers, `get_bls_passphrase_from_env` and `read_passphrase`, plus the branching that decides when a passphrase is required, are private to that binary.

Any external binary that embeds the TN CLI has to reuse this logic. An ExEx host binary (for example, an explorer indexer that runs the node in-process) needs the same three steps the stock binary performs: preload the passphrase from the environment, resolve it against the requested command, then run the launcher. Today the only way to get that behavior is to copy `main.rs`. Copied code drifts. The passphrase path has enough quirks that drift is a correctness risk: stdin overrides the environment variable, an empty stdin line yields `Some("")` rather than `None`, the `ask` and `db`/`genesis` commands still preload from the environment, and `db` is exempt from the passphrase requirement. A second copy that misses one of these behaves differently from the canonical node.

## Solution

Move the passphrase logic out of the binary and into the `telcoin-network-cli` crate so every embedder shares one implementation.

Introduce a public `passphrase` module in `telcoin-network-cli` holding `get_bls_passphrase_from_env` and `read_passphrase`, and add a `Cli::resolve_bls_passphrase` method that encapsulates the resolve step. The stock binary then becomes the canonical template: preload the environment, call `resolve_bls_passphrase`, run with the launcher. External binaries call the same method instead of reimplementing it.

Behavior should stay byte-identical. Prompt and error strings must match the current output exactly, and both fatal paths must still print the same stderr and exit with code 1. The two `eprintln!` + `exit` sites can become `eyre` errors whose `Display` reproduces today's text, which keeps the observable output the same while letting the logic live in a library that returns errors instead of terminating the process. The existing quirks (stdin overriding env, the empty-line `Some("")` case, env preload for `ask`, the `db` exemption) should be preserved deliberately, not incidentally.

grantkee:
- Move `get_bls_passphrase_from_env` and `read_passphrase` from `bin/telcoin-network/src/main.rs` into a new public `passphrase` module in `telcoin-network-cli`, so binaries that embed the TN CLI reuse the logic instead of copying `main.rs`.
- Add `Cli::resolve_bls_passphrase`; the stock binary is now the three-step template (env preload, resolve passphrase, run with launcher).
- Behavior is unchanged: prompt and error strings are byte-identical, and both fatal paths still print the same stderr and exit 1. The two `eprintln!` + `exit` sites become `eyre` errors whose `Display` matches the current output.
- Preserved quirks: stdin overrides env and yields `Some("")` on an empty line; `ask` and `db`/`genesis` keep the env preload; `db` is exempt from the passphrase requirement.

closes #847

### DIFF
--- a/crates/telcoin-network-cli/src/passphrase.rs
+++ b/crates/telcoin-network-cli/src/passphrase.rs
@@ -1,0 +1,159 @@
+//! BLS key passphrase acquisition, shared by the stock `telcoin-network` binary
+//! and external binaries that embed the TN CLI (e.g. ExEx host binaries).
+
+use crate::cli::{Cli, Commands, PassSource};
+use eyre::bail;
+use std::fmt;
+
+/// Environment variable read (and cleared) at process start for the BLS key passphrase.
+pub const BLS_PASSPHRASE_ENVVAR: &str = "TN_BLS_PASSPHRASE";
+
+/// Read the bls key passphrase from then incoming environment if set.
+/// This also will remove the key once read to avoid leaks in future.
+/// This is meant to be called once at the very beginning of program
+/// start before any threads exists.  It will only return the passphrase
+/// on the first call (it clears the env if it is set).
+///
+/// Call this exactly once, as the FIRST statement of `main`, before any
+/// threads exist: it mutates the process environment, which is not thread safe.
+pub fn get_bls_passphrase_from_env() -> Option<String> {
+    if let Ok(passphrase) = std::env::var(BLS_PASSPHRASE_ENVVAR) {
+        if !passphrase.is_empty() {
+            // Clear then remove the passphrase from the env.
+            // NOTE: This is probably not doing much but is an attempt to make the var "more
+            // deleted". This will depend on the underlying platform/libc but should
+            // worst case does nothing. Note on safety, these need calls need to happen
+            // to avoid any leaks of the passphrase if set and they are unsafe.  They
+            // are unsafe because they are not thread safe and we only call this
+            // function once at the beginning of startup so no threads should exist yet.
+            unsafe {
+                std::env::set_var(BLS_PASSPHRASE_ENVVAR, "");
+                std::env::remove_var(BLS_PASSPHRASE_ENVVAR);
+            }
+            Some(passphrase)
+        } else {
+            None
+        }
+    } else {
+        None
+    }
+}
+
+/// Prompt (twice) for a new passphrase when generating keys.
+fn read_passphrase() -> Option<String> {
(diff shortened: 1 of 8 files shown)


crates/telcoin-network-cli/src/passphrase.rs
```diff
@@ -0,0 +1,159 @@
...
+    ) -> eyre::Result<Option<String>> {
+        let mut passphrase = preloaded;
+        match self.bls_passphrase_source {
+            PassSource::Env => {} // Already have the env var if provided.
+            PassSource::Stdin => {
+                let mut buffer = String::new();
+                if let Err(err) = std::io::stdin().read_line(&mut buffer) {
+                    bail!("Error reading BLS passphrase from stdin: {err:?}");
+                }
+                passphrase = Some(buffer.trim_end().to_string());
```
MavenRain: A user can enter an empty string via `Stdin` to bypass the passphrase requirement.

Recommendation:
```
passphrase = Some(buffer.trim_end().to_string()).filter(|s| !s.is_empty());
```

crates/telcoin-network-cli/src/passphrase.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    #[test]
+    fn db_subcommand_is_exempt() {
+        let cli = cli(&["tn", "db", "stats"]);
+        assert_eq!(cli.resolve_bls_passphrase(None).unwrap(), None);
+    }
+
+    /// The ONLY test allowed to touch the environment: `cargo test` runs tests
+    /// concurrently in one process, and the env is process-global.
```
MavenRain, line 186: The documentation here is definitely helpful, but we could enforce this via `#[serial_test::serial]` or using a mutex.

MavenRain (changes requested):
Just held solely on the empty passphrase issue

MavenRain (approved):
Looks GTG now

crates/telcoin-network-cli/src/passphrase.rs
grantkee: Good catch. Made this change and updated tests

crates/telcoin-network-cli/src/passphrase.rs
grantkee: We can serialize tests if/when we have any more that compete for env. Leaving this as an optional follow that

PR #850, opened 2026-07-04
feat(tn-reth): expand RethEnv read API for ExEx consumers

### LINKED ISSUE #849: feat: RethEnv lacks the reads an out-of-process ExEx needs to serve queries
## Problem

`RethEnv` is the read gateway into execution state for code running inside the node, but its API stops short of what an out-of-process ExEx consumer needs. An explorer indexer that runs as an ExEx wants to answer the same queries a block explorer serves — look up a transaction by hash with its block metadata, fetch a receipt, list a block's receipts, walk the chain by transaction number, read an account's code, and make read-only contract calls — without standing up a JSON-RPC endpoint just to reach data the node already holds.

Some of these reads have no home in the current API at all. Others exist only in test utilities (`retrieve_account`) or in shapes that don't fit an external caller (`read_contract_state`). A consumer that needs a chain-wide transaction feed also has to skip empty blocks efficiently rather than visiting every block. And the current error type, `RegistryReadError`, is named for registry reads, which no longer describes a surface that now covers arbitrary contract reads with revert-versus-internal-error distinctions.

Without these reads on `RethEnv`, an ExEx consumer either reaches around the abstraction into raw reth types (coupling it to reth internals) or runs a redundant RPC server.

## Solution

Expand the `RethEnv` read API to cover the queries an out-of-process consumer needs, and give external crates the type re-exports to name the new signatures without depending on reth directly.

Add reads for transaction lookup with metadata (`transaction_by_hash_with_meta`), receipts (`receipt_by_hash`, a per-transaction gas variant, and `receipts_by_block`), a total transaction count (`total_transactions`), and a chain-wide `TxNumber` feed (`transactions_by_tx_range_with_meta`) that skips empty blocks instead of visiting them. Add `account_code` for reading deployed bytecode, and generic read-only contract calls (`read_contract` / `read_contract_at_block`) with `eth_call` semantics: a zero-address caller, no state commit, and the inherent 30M system-call gas cap.

Rename `RegistryReadError` / `RegistryReadResult` to `EvmReadError` / `EvmReadResult`, since the revert-versus-internal split now applies to arbitrary contract reads rather than just the registry, and add a `SignerRecovery` error variant for the sender-recovery fallback. Move `retrieve_account` verbatim from test utilities into the production API, and supersede `read_contract_state` with `read_contract_at_block`, rewiring its single caller. Re-export `TxNumber`, `Log`, `LogData`, `TransactionMeta`, and `Account` from `tn_types` so external crates can name every type in the new signatures.

follow up to #770

grantkee:
- Add the reads an out-of-process ExEx (e.g. an explorer indexer) needs to serve queries without JSON-RPC: `transaction_by_hash_with_meta`, `receipt_by_hash` (plus a per-transaction gas variant), `receipts_by_block`, `total_transactions`, `account_code`, and `read_contract` / `read_contract_at_block`.
- `transactions_by_tx_range_with_meta` is a chain-wide `TxNumber` feed that skips empty blocks rather than visiting them.
- Read-only calls use `eth_call` semantics: zero-address caller, no state commit, inherent 30M system-call gas cap.
- Rename `RegistryReadError` / `RegistryReadResult` to `EvmReadError` / `EvmReadResult` now that the revert-vs-internal split covers arbitrary contract reads; add `TnRethError::SignerRecovery` for the sender-recovery fallback.
- `retrieve_account` moves verbatim from test-utils into the production API; `read_contract_state` is superseded by `read_contract_at_block` (its one caller rewired).
- `tn_types` re-exports `TxNumber`, `Log`, `LogData`, `TransactionMeta`, and `Account` so external crates can name the new signatures without depending on reth directly.

closes #849

### DIFF
--- a/crates/tn-reth/src/lib.rs
+++ b/crates/tn-reth/src/lib.rs
@@ -37,7 +37,7 @@ use alloy::{
 use alloy_evm::Evm;
 use clap::Parser;
 use dirs::path_to_datadir;
-use error::{RegistryReadError, RegistryReadResult, TnRethError, TnRethResult};
+use error::{EvmReadError, EvmReadResult, TnRethError, TnRethResult};
 use evm::TnEvmConfig;
 use eyre::OptionExt;
 use jsonrpsee::Methods;
@@ -80,12 +80,14 @@ use reth_node_builder::{
     DEFAULT_SPARSE_TRIE_MAX_STORAGE_TRIES, DEFAULT_SPARSE_TRIE_PRUNE_DEPTH,
 };
 use reth_node_core::node_config::DEFAULT_CROSS_BLOCK_CACHE_SIZE_MB;
+use reth_primitives_traits::SignerRecoverable as _;
 use reth_provider::{
-    providers::BlockchainProvider, BlockIdReader as _, BlockNumReader, BlockReader,
-    CanonChainTracker, CanonStateSubscriptions as _, Chain, ChainStateBlockReader,
-    ChainStateBlockWriter, DBProvider, DatabaseProviderFactory, ExecutionOutcome,
-    HeaderProvider as _, ProviderFactory, ReceiptProvider as _, StateProviderBox,
-    StateProviderFactory, TransactionVariant,
+    providers::BlockchainProvider, AccountReader as _, BlockBodyIndicesProvider as _,
+    BlockIdReader as _, BlockNumReader, BlockReader, CanonChainTracker,
+    CanonStateSubscriptions as _, Chain, ChainStateBlockReader, ChainStateBlockWriter, DBProvider,
+    DatabaseProviderFactory, ExecutionOutcome, HeaderProvider as _, ProviderFactory,
+    ReceiptProvider as _, StateProvider as _, StateProviderBox, StateProviderFactory,
+    TransactionVariant, TransactionsProvider as _,
 };
 use reth_revm::{
     cached::CachedReads,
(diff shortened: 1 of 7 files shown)


crates/tn-reth/src/lib.rs
```diff
@@ -1121,6 +1144,190 @@ impl RethEnv {
...
+            self.inner.blockchain_provider.transaction_by_hash_with_meta(hash)?
+        else {
+            return Ok(None);
+        };
+
+        // prefer the sender persisted at execution time (two point reads, no ecrecover)
+        let sender = match self.inner.blockchain_provider.transaction_id(hash)? {
+            Some(id) => self.inner.blockchain_provider.transaction_sender(id)?,
+            None => None,
+        };
```
MavenRain, line 1163: It's possible that a reorg between these two reads could lead to a situation where we silently produce a `Recovered` with a mismatched signature instead of erroring, since we trust the resulting sender without re-deriving from the signature.

MavenRain (approved):
Just one thing I think I noticed

sstanfield (approved):

crates/tn-reth/src/lib.rs
grantkee: That is true, but thankfully the nature of TN's BFT consensus makes reorgs impossible. Engine only executes subdag commits, which never go backwards. Thanks for flagging

PR #852, opened 2026-07-04
Feat/basefee 08 e2e

### LINKED ISSUE #867: feat: e2e coverage for per-worker base fees, closing epoch-boundary and restart correctness gaps
## Problem

The per-worker base-fee stack (PRs 01-07) has no end-to-end test proving fees behave correctly across real epoch boundaries and validator restarts — only unit and integration coverage of individual pieces exists. Closing `issues/7-e2e-coverage.md` requires that e2e coverage.

Writing it surfaces that several boundary, restart, and genesis paths are unsound once a real (non-`u64::MAX`) fee target is configured:

- A restart or CVV-resync landing exactly on an epoch boundary can skip both forward fee computation and chain-seeding, leaving the node on stale fees for a full epoch. Exact-match batch validation then rejects its peers, and vice versa. Tracked separately in `issues/blocker-restart-boundary-recovery.md`, still open at HEAD.
- Fail-open paths (worker-count sync, per-worker config read) mask node-local failures as committee-deterministic ones, risking downstream panics or divergence.
- Genesis construction doesn't check the `WorkerConfigs` constructor result, so a ceremony error can silently deploy an unowned, unconfigurable contract.

None of this is currently reachable in production because genesis ships an effectively-infinite gas target — but a single governance transaction arms all of it.

## Solution

Add single-worker e2e coverage for the three core fee behaviors: static-fee pinning, EIP-1559 fee movement across epoch boundaries, and fee recovery after a mid-epoch validator restart.

While closing that coverage, harden the boundary-window recovery path so a node that misses fee computation at epoch close also can't silently skip re-seeding at epoch entry — the invariant should be enforced by the code, not just documented in a comment. Fail-open paths in worker-count sync and per-worker config reads should be tightened to warn rather than silently no-op when they hit an uninitialized or mismatched state. Genesis construction should check the result of pre-genesis constructor calls, starting with `WorkerConfigs`, instead of committing state unconditionally on a ceremony error.

Test assertions that were structurally unable to fail — a final restart check skipped on timeout, an EIP-1559 series checked only by inequality — should be tightened so they can actually catch a regression in the fee path they're named for.

grantkee:
- Adds three e2e tests: static fee pinning, EIP-1559 fee rise across epoch boundaries, and fee recovery after a mid-epoch restart
- Tightens those tests so they can actually fail on regression: mandatory final restart check instead of skip-on-timeout, exact per-boundary fee-step pinning instead of inequality checks, measured (not assumed) mid-epoch restart positioning, receipt-based tx attribution instead of tip inference
- Boundary/restart correctness: entered-epoch and absent-worker base fees now derive from prior-epoch/closing-block chain state instead of defaulting to MIN; catchup/seeding block ranges pinned to a single header source
- Fail-open hardening: warn (not silently no-op) when `set_worker_base_fee` targets an uninitialized worker; diagnosable panic message on worker-id/accumulator mismatch
- Genesis: fail construction on a pre-genesis constructor revert instead of committing state unconditionally; validate CLI worker configs
- Docs: correct base-fee recovery invariant comments to match the F1/F6 fixed behavior

closes #867


grantkee:
This is a final stack for implementing basefees for single worker at the epoch boundary. I will rebase onto main once upstream PR #846 is merged

crates/types/src/gas_accumulator.rs
```diff
@@ -315,7 +331,19 @@ impl GasAccumulator {
     /// Return the shared [`BaseFeeContainer`] for `worker_id`. Mutations are visible to all
     /// holders of the returned clone.
     pub fn base_fee(&self, worker_id: WorkerId) -> BaseFeeContainer {
-        self.inner.read().get(worker_id as usize).expect("valid worker id").base_fee.clone()
+        let inner = self.inner.read();
+        inner
+            .get(worker_id as usize)
+            .unwrap_or_else(|| {
+                panic!(
```
MavenRain, line 338: Just flagging this as a new panic on the production path

crates/types/src/gas_accumulator.rs
```diff
@@ -284,7 +292,15 @@ impl GasAccumulator {
             return;
         }
         let inner = self.inner.read();
-        let mut guard = inner.get(worker_id as usize).expect("valid worker id").gas.lock();
+        let accumulated = inner.get(worker_id as usize).unwrap_or_else(|| {
+            panic!(
```
MavenRain, line 296: Just flagging as another potential new panic on the production path

crates/node/src/engine/inner.rs
```diff
@@ -179,14 +179,28 @@ impl ExecutionNodeInner {
...
             block_info.pending_basefee = base_fee;
             pool.set_block_info(block_info);
+        } else {
+            warn!(
+                target: "tn::execution",
+                worker_id,
+                initialized_workers = self.workers.len(),
+                "set_worker_base_fee: dropping base-fee update for uninitialized worker"
+            );
+            debug_assert!(
```
MavenRain, line 199: Probably by-design, but just flagging this new debug assert on the non-test path

MavenRain (approved):
I don't think anything is blocking . . . just noticed a few new potential non-test-path panics

crates/types/src/gas_accumulator.rs
grantkee: I'm open to other ideas. This was an `expect` but changed to `panic` so more information is included for diagnosing. This would cause  a batch builder to propose the wrong basefees. Maybe this one isn't worth shutting down but no transactions would go through - shouldn't happen though unless there are problems with db

crates/types/src/gas_accumulator.rs
grantkee: This would cause a fork and is likely worth shutting down. I'm open to alternative approaches

crates/node/src/engine/inner.rs
grantkee: My intention was to fail a test if this ever happened in an e2e test but maybe this should always shut the node down?

sstanfield: Seems like node will fork if this happens (?).  If so I vote for a shutdown.

sstanfield (approved):

crates/node/src/engine/inner.rs
grantkee: I will add this as a shutdown in the next PR. Thank you both for the input

PR #854, opened 2026-07-05
test: add persistence barrier before iterating in db_simp_bench (#831)

### LINKED ISSUE #831: test: add a persistence barrier before iterating in db_simp_bench
## Summary

`db_simp_bench` commits a write transaction and then immediately iterates the DB with no happens-before barrier against the background writer thread. When the interleaving lines up, the forward-iterate assertion observes a key duplicated across the on-disk and in-memory layers and fails. This is a real but timing-dependent (rare) flake.

## Affected test

`test_compositedb_dbsimpbench` (`crates/storage/src/composite_db.rs:437`). The mdbx backend path always runs; the redb path runs under the `redb` feature. Both drive the shared `db_simp_bench` helper (`crates/storage/src/lib.rs:231`).

## Root cause

`db_simp_bench` writes 50k rows, commits, and iterates back to back:

`crates/storage/src/lib.rs:245`
```rust
txn.commit().unwrap();
```
`crates/storage/src/lib.rs:255`
```rust
for (k, v) in db.iter::<TestTable>() {
    assert_eq!(k, i);
    assert_eq!(v, i.to_string());
    i += 1;
}
```

`TestTable::HINT` is `TableHint::Cache` (`crates/storage/src/lib.rs:225`), which routes to the `cache_db` layer opened with `full_memory = false` (`crates/storage/src/composite_db.rs:27`). For a non-full-memory layer, `commit()` is asynchronous: it only enqueues a message to the background `std::thread` spawned in `LayeredDatabase::open` (`crates/storage/src/layered_db.rs:222`).

`crates/storage/src/layered_db.rs:84`
```rust
fn commit(self) -> eyre::Result<()> {
    self.tx.send(DBMessage::CommitTxn).map_err(|_| eyre::eyre!("DB thread gone, FATAL!"))?;
    Ok(())
}
```

Writes are staged both in the mem layer and on the background on-disk txn. Each `insert` writes to `mem_db` synchronously and queues a `DBMessage::Insert` (`crates/storage/src/layered_db.rs:277`). The background thread handles `CommitTxn` by committing the on-disk txn and only then draining the mem layer:

`crates/storage/src/layered_db.rs:113`
```rust
DBMessage::CommitTxn => {
    if let Some((current_txn, count)) = txn.take() {
        if count <= 1 {
            if let Err(e) = current_txn.commit() { /* ... */ }
            if let Some(mem_db) = mem_db.as_ref() {
                for insert in committed_inserts.drain(..) {
                    insert.clear_insert_mem(mem_db);
                }
            }
        }
    }
}
```

For a non-full-memory layer, `iter()` chains the on-disk iterator with the mem iterator, and its doc comment explicitly warns it may return duplicates during in-flight writes:

`crates/storage/src/layered_db.rs:309`
```rust
fn iter<T: Table>(&self) -> DBIter<'_, T> {
    if self.full_memory {
        Box::new(self.mem_db.iter::<T>())
    } else {
        Box::new(self.db.iter::<T>().chain(self.mem_db.iter::<T>()))
    }
}
```

## Why it flakes

`txn.commit()` returns as soon as `CommitTxn` is queued, then the main thread calls `db.iter()` concurrently with the background writer. The failing window is narrow but real: the on-disk `current_txn.commit()` and the `clear_insert_mem` drain live in the same handler and run back to back on the background thread, so there is a brief interval (the length of the 50k-entry drain loop) in which the on-disk txn is committed but the mem layer is not yet cleared. If the main thread constructs and consumes the chained iterator during that interval, the on-disk half yields `0..50_000` and then the mem half re-yields keys still present in mem. After the disk half advances `i` to `50_000`, the mem half emits a key like `0`, so `assert_eq!(k, i)` at `crates/storage/src/lib.rs:256` fails.

In the more common orderings the test passes: if the background thread has not yet committed the on-disk txn, the disk iterator is empty and only the mem layer is observed; if it has fully committed and drained, the mem layer is empty. The failure depends purely on hitting the commit-done-drain-pending window, which is why it is intermittent.

A synchronous barrier already exists and provides the needed happens-before edge. Because the mpsc channel is FIFO and `CaughtUp` is sent after `CommitTxn`, the background thread finishes the commit and the drain before it replies to `CaughtUp`:

`crates/storage/src/layered_db.rs:362`
```rust
fn sync_persist(&self) {
    let (tx, mut rx) = oneshot::channel();
    let r = self.tx.send(DBMessage::CaughtUp(tx))
        .map_err(|_| eyre::eyre!("DB thread gone, FATAL!"));
    // ... polls until the background thread replies ...
}
```

`CompositeDatabase::sync_persist` (`crates/storage/src/composite_db.rs:116`) already fans this out across all layers, so it is callable directly from the test.

## Suggested fix

- [ ] In `db_simp_bench`, call `db.sync_persist()` after `txn.commit()` and before the first `db.iter()` so the background writer has drained through `CommitTxn`.
- [ ] Consider adding the same barrier before the subsequent `reverse_iter` and read-txn loops. For a non-full-memory layer `reverse_iter` reads the on-disk layer only (`crates/storage/src/layered_db.rs:325`), so before the background commit lands it can observe a not-yet-committed txn (here it would silently under-iterate rather than assert, so this is a lesser concern).
- [ ] Optionally document that `commit()` on a non-full-memory `LayeredDatabase` is asynchronous, and that callers needing a read-your-writes guarantee for iteration must `sync_persist()` (or `await persist()`) first.

MavenRain:
## Summary

`db_simp_bench` (`crates/storage/src/lib.rs`) writes 50k rows, commits, and iterates back to
back.  On a non-full-memory `LayeredDatabase`, `commit()` only queues `DBMessage::CommitTxn` to
the background writer thread and returns.  The handler commits the on-disk txn and only then
drains the mem layer, so there is a brief window (the length of the 50k-entry drain loop) in
which a key is present in both layers.  `iter()` on a non-full-memory layer chains the on-disk
and mem iterators, so an iterator consumed inside that window re-yields keys and the
`assert_eq!(k, i)` ordering assertion fails.  `test_compositedb_dbsimpbench` hits this because
`TestTable::HINT` is `TableHint::Cache`, which routes to the composite cache layer opened with
`full_memory = false`.  Rare, timing-dependent flake; full trace in #831.

## Fix

- Call `db.sync_persist()` in `db_simp_bench` after the first commit and before iterating.
  `CaughtUp` is FIFO-ordered behind `CommitTxn` on the same channel, so when it returns the
  on-disk commit has landed and the mem layer is drained.
- A single barrier covers the forward iterate, `reverse_iter`, and the read loops that follow,
  because no further writes occur until the `clear_table`.  This also closes the `reverse_iter`
  under-iteration case noted in the issue (on a non-full-memory layer it reads only the on-disk
  layer, which the barrier guarantees is committed).
- The barrier sits outside the printed timings on purpose: the commit number keeps measuring
  the (async) `commit()` call itself.  `sync_persist()` is a default no-op on the `Database`
  trait, so the raw MDBX / ReDB / MemDatabase callers of the shared helper are unaffected.
- Document on `LayeredDbTxMut::commit` that the commit is asynchronous for a non-full-memory
  layer and that callers needing a read-your-writes guarantee for iteration must
  `sync_persist()` (or await `persist()`) after committing.

Not changed here: `test_layereddb_dbsimpbench` still runs only the full-memory configuration
("Otherwise iterators, while correct, may not work the test").  With the barrier now in the
shared helper it could also cover the non-full-memory path; happy to extend it in this PR or a
follow-up if wanted.

## Tests

- `cargo test -p tn-storage dbsimpbench`: 5 passed (mdbx, memdb, layered, composite).
- `cargo test -p tn-storage --features redb dbsimpbench`: 5 passed.
- The pre-fix failure is timing-dependent (the commit-done, drain-pending window), so there is
  no deterministic fail-before repro; the barrier removes the window by construction.

## Checklist

- [x] `db.sync_persist()` barrier after commit, before the first `iter()` (issue box 1)
- [x] the same happens-before edge covers `reverse_iter` and the read-txn loops (issue box 2)
- [x] `commit()` async semantics documented on the layered impl (issue box 3)
- [x] `cargo +nightly-2026-03-20 fmt` clean
- [x] `cargo +nightly-2026-03-20 clippy -p tn-storage --features redb,test-utils --all-targets`
  clean (note: `-p tn-storage --all-features` fails to compile on current main because the
  crate's `adiri = []` feature does not forward to `tn-types/adiri`; unrelated to this diff)
- [x] `tn-storage` dbsimpbench tests pass (default and `redb`)

Closes #831.

### DIFF
--- a/crates/storage/src/layered_db.rs
+++ b/crates/storage/src/layered_db.rs
@@ -81,6 +81,13 @@ impl<DB: Database> DbTxMut for LayeredDbTxMut<DB> {
         Ok(())
     }
 
+    /// Queue the commit to the background writer thread and return.
+    ///
+    /// For a non-full-memory layer the commit is therefore asynchronous: when this
+    /// returns, the data may not be committed on-disk yet and may still be in the mem
+    /// layer, so an immediate `iter()` can observe a key in both layers.  Callers that
+    /// need a read-your-writes guarantee for iteration must call
+    /// [`Database::sync_persist`] (or await [`Database::persist`]) after committing.
     fn commit(self) -> eyre::Result<()> {
         self.tx.send(DBMessage::CommitTxn).map_err(|_| eyre::eyre!("DB thread gone, FATAL!"))?;
         Ok(())
--- a/crates/storage/src/lib.rs
+++ b/crates/storage/src/lib.rs
@@ -251,6 +251,15 @@ mod test {
             start.elapsed().as_secs_f64()
         );
 
+        // `commit()` on a non-full-memory layered DB only queues the commit to the
+        // background writer, so wait for that write path to drain before iterating.
+        // Without this barrier the chained disk+mem iterator below can briefly observe
+        // a key in both layers and fail the ordering assertions.  A single barrier
+        // covers all the reads below since no writes occur until the clear_table.
+        // Deliberately untimed: the commit number above measures the (async) commit
+        // call itself, not durability.
+        db.sync_persist();
+
         let start = std::time::Instant::now();
         let mut i = 0;
         #[allow(clippy::explicit_counter_loop)]


grantkee (approved):
LGTM

PR #855, opened 2026-07-05
test: drive test_all_possible_error_outcomes off the rebuild ordering instead of sleep(1s)

### LINKED ISSUE #833: test: drive test_all_possible_error_outcomes off the rebuild event instead of sleep(1s)
## Summary

`test_all_possible_error_outcomes` uses a fixed `sleep(1s)` plus an assumed ordering between the batch builder's 1ms rebuild interval and a freshly submitted pending transaction. The per-iteration assert checks an exact accumulating transaction count with zero slack, so a rebuild that snapshots the pool before the new tx lands makes the test fail.

## Affected test(s)

- `test_all_possible_error_outcomes` (`crates/batch-builder/src/lib.rs:553`)

The test uses the default `#[tokio::test]` runtime, which is the current-thread flavor with a real (non-paused) clock.

## Root cause

The builder under test is constructed with a 1ms `max_delay`:

`crates/batch-builder/src/lib.rs:562`
```rust
let batch_builder = BatchBuilder::new(
    &reth_env,
    txpool.clone(),
    to_worker,
    address,
    Duration::from_millis(1),
```

Each loop iteration receives a batch and asserts an exact, accumulating count:

`crates/batch-builder/src/lib.rs:660`
```rust
// all 3 transactions present
assert_eq!(sealed_batch.batch().transactions().len(), 3 + subdag_index);
```

After acking a non-fatal error, the loop submits one new tx, triggers a canonical update, then sleeps a fixed second before looping back to the assert:

`crates/batch-builder/src/lib.rs:688`
```rust
// sleep to ensure canonical update received before ack
let _ = tokio::time::sleep(Duration::from_secs(1)).await;
```

On a non-fatal error the builder returns an empty `mined_transactions`, which leaves the previously pending txs in the pool and only resets the 1ms interval before breaking:

`crates/batch-builder/src/lib.rs:289`
```rust
// NOTE: empty vec returned for non-fatal error during block proposal
if mined_transactions.is_empty() {
    // reset interval to prevent immediate re-wake from stale tick
    this.max_delay_interval.reset();
    let _ = this.max_delay_interval.poll_tick(cx);

    // return pending and wait for canonical update to wake up again
    break;
}
```

When the builder next wakes, the still-pending txs are non-empty, so it spawns another build that reads whatever is in the pool at that instant:

`crates/batch-builder/src/lib.rs:259`
```rust
if this.pending_task.is_none() {
    // check for pending transactions
    if this.pool.pending_transactions().is_empty() {
        // reset interval to wake up after some time
        //
        // only need to reset here if there is no pending block being built
        this.max_delay_interval.reset();

        // tick interval to ensure it advances
        let _ = this.max_delay_interval.poll_tick(cx);

        // nothing pending
        break;
    }

    // start building the next batch
    this.pending_task = Some(this.spawn_execution_task());
```

## Why it flakes

The builder future only wakes on its `max_delay_interval` (1ms) or the canonical `state_changed` stream. The wake sources are the struct's `pending_task`, `max_delay_interval`, and `state_changed` fields (`crates/batch-builder/src/lib.rs:60`); there is no pool pending-tx notification subscription. The builder discovers new txs only by polling `pool.pending_transactions()` when it happens to wake.

Sequence per iteration:
1. Test acks the non-fatal error. The builder observes the empty `mined_transactions`, resets its 1ms interval, and breaks with the prior txs still pending.
2. Test submits one new pending tx (`create_and_submit_eip1559_pool_tx(...).await`). The pool add is async (it awaits validation), so it yields before the tx reaches the pending set.
3. Test triggers an empty canonical update, then `sleep(1s)`.

On the current-thread runtime the builder was woken by the ack in step 1, so it is polled while the test is parked in step 2's internal await. That poll only rebuilds if the 1ms interval is already ready. Normally the interval is not ready (only microseconds have elapsed since the reset), so the builder re-parks and the eventual rebuild during the sleep sees the new tx. But if more than 1ms of wall-clock has elapsed at that poll (scheduling jitter, a single OS preemption of the process, or general CI load), the interval fires and the builder snapshots the pool before the new tx is added. That short batch is sent to the worker and sits in the `from_batch_builder` channel, so the next iteration's `assert_eq!(len, 3 + subdag_index)` observes one fewer tx than expected. The `sleep(1s)` does not prevent that early rebuild, and the exact-count assert has no tolerance for it.

This is a real but rare ordering hazard: it does not reproduce reliably on an unloaded machine and surfaces intermittently under CI load.

## Suggested fix

- [ ] Replace the fixed `sleep(Duration::from_secs(1))` with a deterministic wait keyed to observable state (for example, wait until `txpool.pool_size().pending` reflects the newly submitted tx before proceeding).
- [ ] Assert against the rebuilt batch only after the new tx is guaranteed to be in the pool, or drain `from_batch_builder` until a batch with the expected count arrives, instead of asserting on the first received batch.
- [ ] Alternatively, submit the new tx and trigger the canonical update before acking the non-fatal error so the builder cannot rebuild an intermediate batch that misses the new tx.
- [ ] Consider a larger `max_delay` in this test so the interval does not race the pool update, if a timing-sensitive path still remains.

## Severity / notes

Test-only flakiness. The exact-count assert has zero slack and the failure is timing-dependent, so it surfaces intermittently under CI load rather than deterministically. Production builder behavior is unchanged; this is a test-harness ordering issue.

MavenRain:
Closes #833.

`test_all_possible_error_outcomes` raced its own batch builder.   After acking a
non-fatal error the test submitted a new pending tx, and a fixed `sleep(1s)` was
the only thing between the builder's 1ms rebuild interval and that pool update.
Under CI load the builder could wake and snapshot the pool before the new tx
landed, producing a short batch that fails the zero-slack
`assert_eq!(len, 3 + subdag_index)`.

Changes (test-only, no production code touched):

- Submit the new pending tx and execute the canonical update *before* sending the
  non-fatal ack.  The builder cannot start another build while `pending_task` is
  still awaiting the ack, so the first possible rebuild happens only after the
  ack resolves - by which point the awaited (synchronous) pool submit guarantees
  the new tx is in the pending set.  Every rebuild now sees the new tx regardless
  of when the 1ms interval fires.  This mirrors the ordering
  `test_pool_updates_after_txs_mined` already uses.
- Remove the `sleep(Duration::from_secs(1))`;  with the reorder there is no timing
  window left for it to paper over, and the test no longer depends on wall-clock
  timing.

Asserted batch counts are unchanged (3, 4, 5, 6, then 7).  Verified with the fixed
test run 5x plus the full `tn-batch-builder` suite (7 passed);  nightly fmt clean.

### DIFF
--- a/crates/batch-builder/src/lib.rs
+++ b/crates/batch-builder/src/lib.rs
@@ -672,10 +672,9 @@ mod tests {
             // all 3 transactions present
             assert_eq!(sealed_batch.batch().transactions().len(), 3 + subdag_index);
 
-            // send non-fatal error
-            let _ = ack.send(Err(error));
-
-            // submit another tx to pool
+            // submit another tx to pool before acking so the builder's next rebuild
+            // (whenever it wakes) is guaranteed to see the new tx - the builder can't
+            // rebuild while it still awaits this ack
             tx_factory
                 .create_and_submit_eip1559_pool_tx(
                     chain.clone(),
@@ -697,8 +696,8 @@ mod tests {
             // update values for next loop
             parent = final_header;
 
-            // sleep to ensure canonical update received before ack
-            let _ = tokio::time::sleep(Duration::from_secs(1)).await;
+            // send non-fatal error
+            let _ = ack.send(Err(error));
         }
 
         // wait for next block


grantkee (approved):
LGTM!

PR #856, opened 2026-07-05
fix(network-libp2p): bound `published_to_peers` with an LRU (closes #828)

### LINKED ISSUE #828: network-libp2p: `published_to_peers` grows with every distinct peer and is never cleaned on disconnect (RAM-capped nodes will OOM)
## Summary

`published_to_peers` is a per-process-lifetime de-dup gate that is inserted into once per distinct `PeerId` ever connected and is never removed, including on disconnect.  Under a hard cap, this monotonic growth becomes a guaranteed eventual OOM-kill plus restart loop.  Confirmed against `origin/main` at `0f07dcca`; line numbers are for that commit.

This is one of three related unbounded peer-map findings.  Companion issues: `known_peers` (`peers/manager.rs:54`) and `banned_peers_by_ip` (`peers/banned.rs:28`).  A separate, higher-privilege finding covers `consensus_certs` in `consensus/primary`.

## Details

**Field:** `crates/network-libp2p/src/consensus.rs:183`
```rust
183:  published_to_peers: HashSet<PeerId>,
```

**Sole insert:** `consensus.rs:1292`, in the `PeerEvent::PeerConnected` arm:
```rust
1292:  if self.published_to_peers.insert(peer_id) {
```

**No eviction exists.** The set has exactly four references (decl `183`, init `399`, doc `484`, insert `1292`); it is never removed from. The `PeerDisconnected` handlers clean only `connected_peers`, not this set. Its own doc comment (`consensus.rs:181`) calls it "per-process-lifetime." It is a de-dup gate whose purpose is to avoid re-pushing our local kad store to a peer we have already pushed to, so it grows once per distinct `PeerId` ever connected.

**Trigger.** `PeerId` is a peer-minted cryptographic identity. An attacker mints fresh Ed25519 keypairs and connects once each; each admitted connection inserts a permanent entry that survives the subsequent disconnect. The concurrent-connection admission gate (`peers/behavior.rs`) caps live connections but not the set, because entries persist across disconnect, so a connect / insert / disconnect / repeat loop grows it without bound. Benign long-uptime peer churn does the same, more slowly.

**Rough 16GB math.** At ~64 to 80 B per entry (a `PeerId` multihash), ~13 to 16M distinct peers is ~1GB.  Growth is throttled by the libp2p handshake cost and the need for a free admission slot per new id, so this is a slow leak; but it is monotonic with no ceiling and adds to the total alongside the other peer maps.

## Fix

Do **not** naively remove on `PeerDisconnected`: that re-enables the exact kad re-push amplification on flapping peers that this set was added to prevent.  Instead bound it: replace the unbounded `HashSet<PeerId>` with a fixed-capacity LRU (or a `PeerId -> last_pushed_at` map with TTL eviction).  A peer evicted from the LRU would at worst be re-pushed to once if it reconnects after a long absence, which is acceptable and self-limiting.

**Fix risk.** Medium if done wrong (a plain remove-on-disconnect reintroduces the amplification).  Low with an LRU/TTL, which preserves the de-dup intent for actively churning peers while capping resident size.

## Checklist
- [ ] Replace the unbounded `HashSet<PeerId>` with a capacity-bounded LRU or a TTL-evicted map
- [ ] Confirm the de-dup intent is preserved for actively churning peers (no re-push storm)
- [ ] Add a resident-size regression test: connect/disconnect many fresh peer ids, assert the set stays bounded

MavenRain:
## Problem

`published_to_peers` (`crates/network-libp2p/src/consensus.rs`) is a per-process-lifetime de-dup
gate recording every peer we have pushed our `NodeRecord` to.  It is inserted into once per distinct
`PeerId` ever connected (`PeerEvent::PeerConnected`) and **never removed** . . . the `PeerDisconnected`
handlers clean only `connected_peers`, not this set.  Because a `PeerId` is a peer-minted
cryptographic identity, a churn of fresh identities (or benign long-uptime peer churn) grows the set
without bound.  On a RAM-capped node this monotonic growth is a slow but guaranteed OOM-kill plus
restart loop.

Removing entries on disconnect is **not** a valid fix: it re-enables the exact kad re-push
amplification on flapping peers that this gate exists to prevent.

## Fix

Replace the unbounded `HashSet<PeerId>` with a capacity-bounded `lru::LruCache<PeerId, ()>`
(`MAX_PUBLISHED_TO_PEERS = 10_000`, a generous multiple of the ~33 live-peer target, ~64–80 B/entry
so well under 1 MB resident):

- The de-dup decision is factored into `mark_published_to_peer(peer_id) -> bool`, which returns
  `true` the first time a peer is seen (a push is warranted) and `false` otherwise.  It is backed by
  `LruCache::put(peer_id, ()).is_none()`, which is behaviourally identical to the old
  `HashSet::insert(peer_id)` for the de-dup decision.
- A hit **promotes** the peer to most-recently-used, so an actively (re)connecting peer is never the
  eviction victim and is never re-pushed to.  Only a peer absent long enough to fall out of the LRU
  is re-pushed on its eventual return . . . at worst once, which is self-limiting.  This preserves the
  de-dup intent while capping resident memory.

`lru` was already an (unused) workspace dependency pinned at a stale `0.10`;  bumped to `0.16` to
match the version already resolved transitively (`0.16.3`), so no new lockfile version is
introduced.

## Changes

- [x] Replace the unbounded `HashSet<PeerId>` with a capacity-bounded LRU (`MAX_PUBLISHED_TO_PEERS`)
- [x] Preserve the de-dup intent for actively churning peers via LRU promotion (no re-push storm)
- [x] Add resident-size regression test: a flood of distinct fresh peer ids keeps the set pinned at
      capacity (`test_published_to_peers_is_bounded`)
- [x] Add an anti-amplification regression test: an actively reconnecting peer survives eviction and
      keeps de-duping under a fresh-peer flood (`test_published_to_peers_keeps_active_peer_resident`)
- [x] `lru` workspace dependency `0.10` → `0.16`; add `lru = { workspace = true }` to
      `tn-network-libp2p`

Companion issues from the same unbounded-peer-map audit (separate PRs): `known_peers`
(`peers/manager.rs`), `banned_peers_by_ip` (`peers/banned.rs`), `consensus_certs`
(`consensus/primary`).

Closes #828

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11699,6 +11699,7 @@ dependencies = [
  "eyre",
  "futures",
  "libp2p",
+ "lru",
  "metrics",
  "metrics-util",
  "rand 0.9.2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -201,7 +201,7 @@ ouroboros = "0.17"
 proc-macro2 = "1.0.47"
 uint = "0.10"
 cfg-if = "1.0.0"
-lru = "0.10"
+lru = "0.16"
 tokio-stream = { version = "0.1.14", features = ["sync", "net"] }
 serde_yaml = "0.8.26"
 byteorder = "1.4.3"
--- a/crates/network-libp2p/Cargo.toml
+++ b/crates/network-libp2p/Cargo.toml
@@ -35,6 +35,7 @@ snap = { workspace = true }
 # peer manager
 rand = { workspace = true }
 serde_with = { workspace = true }
+lru = { workspace = true }
 
 # metrics
 metrics = { workspace = true }
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -31,9 +31,11 @@ use libp2p::{
     swarm::{NetworkBehaviour, SwarmEvent},
     Multiaddr, PeerId, StreamProtocol, Swarm, SwarmBuilder,
 };
+use lru::LruCache;
 use std::{
     collections::{HashMap, HashSet, VecDeque},
     io::ErrorKind,
+    num::NonZeroUsize,
     time::Duration,
 };
 use tn_config::{KeyConfig, LibP2pConfig, NetworkConfig, PeerConfig};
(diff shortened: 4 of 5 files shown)


sstanfield (dismissed):
LGTM!

grantkee (approved):

PR #857, opened 2026-07-06
fix(network-libp2p): bound known_peers against unbounded peer injection (#827)

### LINKED ISSUE #827: network-libp2p: `known_peers` grows without bound and accepts entries for arbitrary keys from any peer (RAM-capped nodes will OOM)
## Summary

`PeerManager.known_peers` is only ever inserted into and never pruned: authorities rotated out of the committee are never removed, and any remote peer can inject entries for arbitrary keys via signed kad records.  Under a hard cap, this monotonic growth stops being a slow leak and becomes a guaranteed eventual OOM-kill plus restart loop.  Confirmed against `origin/main` at `0f07dcca`; line numbers are for that commit.

This is one of three related unbounded peer-map findings.  Companion issues: `published_to_peers` (`consensus.rs:183`) and `banned_peers_by_ip` (`peers/banned.rs:28`).  A separate, higher-privilege finding covers `consensus_certs` in `consensus/primary`.

## Details

**Field:** `crates/network-libp2p/src/peers/manager.rs:54`
```rust
54:  known_peers: HashMap<BlsPublicKey, NetworkInfo>,
```

**Inserts:** `manager.rs:139` (bootstrap / authority) and `manager.rs:753` (`add_known_peer`), the latter reached from the kad discovery path at `consensus.rs:1596` (inbound `PutRecord`) and `consensus.rs:1686` (query result).

**No eviction exists.** Every reference to `known_peers` in the crate is `insert` / `get` / `contains_key` / `iter` / `len` (manager.rs lines 54, 112, 139, 298, 721, 753, 769, 782, 787, 795). There is no `remove`, `retain`, `clear`, TTL, or capacity anywhere. Authorities rotated out of the committee are never removed, so the map is the union of every authority ever seen plus every bootstrap and kad-discovered peer.

**Remote injection path (no committee check).** Inbound record handling in `consensus.rs`:
```rust
1586:  if let Some((key, value)) = self.peer_record_valid(&record) {
1588:      if self.is_newer_record(&record) {
1593:          .put(record)  // kad store, bounded by capacity + TTL
1594:          .map_err(|e| NetworkError::StoreKademliaRecord(e.to_string()))?;
1596:          self.swarm.behaviour_mut().peer_manager.add_known_peer(key, value.info);
```
`peer_record_valid` only checks that the record is signature-valid and that the publisher matches the record's network key. A peer can generate a fresh `(BlsPublicKey, network keypair)` pair, sign its own `NodeRecord` (publisher = its own key), and pass this check. `key` is the attacker-chosen `BlsPublicKey`; nothing verifies it belongs to any committee. Valid records incur no penalty (only invalid ones get `Penalty::Fatal` at `consensus.rs:1608`), so the sender is not banned for doing this.

**Why it is unbounded despite the kad store cap.** The kad store `put` at `consensus.rs:1593` is bounded (capacity plus `kad_record_ttl`, set at `consensus.rs:300`), and `add_known_peer` is only reached when `put` succeeds.  However, `known_peers` entries never expire. When a kad record expires from the store, its slot frees, yet the `known_peers` entry it created remains forever.  Each TTL window an attacker can therefore add up to ~(kad store capacity) new distinct entries with fresh keys, and cumulative `known_peers` size grows without bound across windows. Benign committee rotation adds to this at a slower, also-unbounded rate.

**Rough 16GB math.** At ~350 B per entry (BLS G2 public key plus `NetworkInfo`), ~3M entries is ~1GB and ~48M is ~16GB. The kad-store-capacity-per-TTL throttle makes this a sustained-attack timescale rather than minutes, but there is no ceiling, so the cap is reached eventually.

## Fix

Bound `known_peers`. Options, cheapest first:
- On committee rotation (`update_committees` path), drop `known_peers` entries whose key is not in the current or next committee and is not a configured bootstrap/allowlisted peer.
- Additionally cap the discovery-inserted (non-committee, non-bootstrap) subset with an LRU or a TTL tied to `kad_record_ttl`, so kad-discovered records that are never promoted to committee members age out with their kad record.

**Fix risk.** Low. `known_peers` is read via `get` / `contains_key` on demand;  evicting a non-committee entry only forces a re-discovery if that peer later becomes relevant, which is the intended kad flow anyway.

## Checklist
- [ ] Prune `known_peers` on committee rotation (keep current/next committee + bootstrap/allowlisted)
- [ ] Add TTL/LRU eviction for discovery-inserted, non-committee entries
- [ ] Add a resident-size regression test: flood signed records for fresh keys, assert `known_peers` stays bounded

MavenRain:
Fixes #827.

## Summary

`PeerManager.known_peers` was insert-only and unbounded.  Any remote peer could grow it without limit by publishing signature-valid kad records for endless fresh `BlsPublicKey`s, and rotated-out committee members were never removed.  On a RAM-capped node this is an eventual OOM-kill plus restart loop.

This bounds the map with a hard structural invariant:

> `known_peers` ⊆ operator-pinned keys ∪ (previous ∪ current ∪ next committee)

Both sets are bounded (node config and committee size), so the map can no longer grow without bound: no soft LRU/TTL cap needed.

## How

The single `add_known_peer` served two very different callers, and the fix separates them by trust:

- **Operator/local path** (`add_known_peer`, unchanged callers: startup restore, `AddExplicitPeer`, `AddBootstrapPeers`, and the trusted-peer dial).  These are config-bounded and not attacker-reachable, so they stay uncapped but are now **pinned** . . . recorded in a new `pinned_peers` set so rotation never evicts them.
- **Kad discovery path** (new `add_discovered_peer`, wired to the two kad sites in `consensus.rs`: the inbound `PutRecord` handler and the query-result handler).  This is the attacker-reachable path.  A record is cached **only if its key is a tracked committee member**.  `known_peers` exists solely to resolve committee members' network info, so a record for a key in no committee slot is either stale or forged and is never read: dropping it costs nothing.  Legitimate discovery is unaffected because kad lookups are only ever triggered for current/next committee members whose info is missing (`trigger_missing_authorities`).

To also close the slower leak (members that legitimately rotate out), `prune_known_peers` runs at the end of `update_committees` and retains only pinned or currently-committee keys.  A member still in the `previous` slot is kept one more epoch by design (late gossip), then dropped.

Supporting change: `AllPeers::is_committee_member` centralizes the previous∨current∨next slot check (also now used by `apply_membership_if_committee`).

## Why this is safe

Committee membership is set every epoch from authoritative consensus state, never from a network record, so an attacker cannot make `is_committee_member` return true for a key they invented. Evicting a non-committee entry only forces a fresh kad lookup if that peer later becomes committee-relevant — which is the intended discovery flow (see the issue's "Fix risk: low").  All readers of `known_peers` (`get_rpc`, `all_rpcs`, `auth_to_peer`, `find_authorities`, `trigger_missing_authorities`) query by authority/committee key, so non-committee entries were never usefully read.

## Tests

- `test_discovered_peers_bounded_to_committee_membership`: floods 1,000 signature-valid records for fresh non-committee keys via the discovery path and asserts `known_peers` stays empty, while a genuine committee member's record on the same path is cached.
- `test_known_peers_pruned_on_rotation_but_pinned_survive`: a discovered committee member is pruned once it rotates out of every slot, while an operator-pinned peer survives rotation.

## Checklist

- [x] Prune `known_peers` on committee rotation (keep current/next/previous committee + pinned)
- [x] Bound the discovery-inserted subset (reject non-committee kad records — tighter than the LRU/TTL the issue suggested; gives an exact structural bound)
- [x] Resident-size regression test: flood signed records for fresh keys, assert `known_peers` stays bounded

### DIFF
--- a/crates/network-libp2p/src/peers/manager.rs
+++ b/crates/network-libp2p/src/peers/manager.rs
@@ -52,6 +52,14 @@ pub(crate) struct PeerManager {
     /// This is used for bootstrapping and to make sure we know the network settings of committee
     /// members.
     known_peers: HashMap<BlsPublicKey, NetworkInfo>,
+    /// BLS keys whose `known_peers` entry is pinned and never evicted by committee rotation.
+    ///
+    /// Populated by the locally provisioned insertion paths — trusted/bootstrap/explicit peers and
+    /// records restored from persistence at startup — whose count is bounded by node
+    /// configuration. Every other `known_peers` entry is a kad-discovered committee record kept
+    /// only while its key sits in a tracked committee slot, so [`Self::prune_known_peers`] can
+    /// drop rotated-out members without touching operator-provisioned peers.
+    pinned_peers: HashSet<BlsPublicKey>,
     /// A queue of events that the `PeerManager` is waiting to produce.
     events: VecDeque<PeerEvent>,
     /// A queue of peers to dial.
@@ -110,6 +118,7 @@ impl PeerManager {
             heartbeat,
             peers,
             known_peers: Default::default(),
+            pinned_peers: Default::default(),
             events: Default::default(),
             dial_requests: Default::default(),
             temporarily_banned,
@@ -136,6 +145,8 @@ impl PeerManager {
         if self.temporarily_banned.remove(&peer_id) {
             warn!(target: "peer-manager", ?peer_id, "removed trusted peer from temporarily banned list");
         }
+        // trusted peers are operator-provisioned; pin so committee rotation never evicts them
+        self.pinned_peers.insert(bls_key);
         self.known_peers.insert(bls_key, info);
 
         self.dial_peer(peer_id, multiaddr, Some(reply));
@@ -685,6 +696,10 @@ impl PeerManager {
         self.forgive_temporarily_banned(&next);
         let unban_actions = self.peers.update_committees(previous, current, next);
         self.apply_unban_actions(unban_actions);
+        // bound `known_peers`: with the new committee slots in place, drop every entry that is
+        // neither pinned nor a current committee member so rotated-out members and stale discovered
+        // records cannot accumulate across epochs (issue #827).
+        self.prune_known_peers();
     }
 
     /// Pre-dial recovery: forgive bans for a committee so a subsequent dial loop can connect,
@@ -696,6 +711,21 @@ impl PeerManager {
         self.forgive_temporarily_banned(&committee);
         let unban_actions = self.peers.mark_committee_for_dial(committee);
         self.apply_unban_actions(unban_actions);
+    }
+
+    /// Drop cached network info for peers no longer worth tracking, bounding `known_peers`.
+    ///
+    /// Run after every committee rotation. An entry survives only if it is pinned (an
+    /// operator-provisioned trusted/bootstrap/explicit peer) or its key still sits in a tracked
+    /// committee slot, so members that rotated out — and any stale kad-discovered records — cannot
+    /// accumulate across epochs. `known_peers` is thereby bounded by the operator-configured set
+    /// plus the three committee slots. Evicting a still-relevant peer only forces a fresh kad
+    /// lookup if it becomes a committee member again, which is the intended discovery flow.
+    fn prune_known_peers(&mut self) {
+        let pinned = &self.pinned_peers;
+        let peers = &self.peers;
+        self.known_peers
+            .retain(|bls_key, _| pinned.contains(bls_key) || peers.is_committee_member(bls_key));
     }
 
     /// Lift any already-known committee members out of the manager's temporary-ban cache.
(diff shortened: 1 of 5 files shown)


crates/network-libp2p/src/peers/manager.rs
```diff
@@ -698,6 +713,21 @@ impl PeerManager {
...
 
+    /// Drop cached network info for peers no longer worth tracking, bounding `known_peers`.
+    ///
+    /// Run after every committee rotation. An entry survives only if it is pinned (an
+    /// operator-provisioned trusted/bootstrap/explicit peer) or its key still sits in a tracked
+    /// committee slot, so members that rotated out — and any stale kad-discovered records — cannot
+    /// accumulate across epochs. `known_peers` is thereby bounded by the operator-configured set
+    /// plus the three committee slots. Evicting a still-relevant peer only forces a fresh kad
+    /// lookup if it becomes a committee member again, which is the intended discovery flow.
+    fn prune_known_peers(&mut self) {
```
sstanfield, line 724: Will this create pressure on validators at epoch boundaries by causing observers to forget each other and hammer committee members until they rebuild a mesh of observer peers?  Maybe not, if they don't disconnect.

sstanfield (approved):
Looks good.  Have a question about the eviction policy effecting the network of non-validators.

crates/network-libp2p/src/peers/manager.rs
MavenRain: Your hedge "if they don't disconnect" is the right intuition.
 
 The short answer is: no herd, and no observer forgets another observer.
  
 Prune never disconnects.  `prune_known_peers` is one retain on the `known_peers` map (crates/network-libp2p/src/peers/manager.rs:727-728);  the only other mutations to that map are two inserts. It writes no swarm, kademlia, gossipsub, connected_peers, or discovery_peers state, so evicting an entry drops a cached record and severs no live link.  That is precisely your "if they don't disconnect" case.
  
Observers are never held in `known_peers`.  It is a BlsPublicKey -> NetworkInfo cache.  Live observer connectivity lives in the swarm behaviours: on connect we register request-response, `kademlia.add_address`, gossipsub (FromSwarm::ConnectionEstablished), and `connected_peers.push_back` (consensus.rs:1283-1302), with no known_peers write.  Reconnection dialing draws from the PeerId-keyed `discovery_peers` pool, not `known_peers` (manager.rs:~983).  Also, the write paths keep observers out by design: `add_discovered_peer` drops non-committee keys (:788-795) and `add_self_advertised_peer` keeps a live nvv/observer identity in the peer store but not in known_peers (:807).  So the observer mesh was never stored here for a prune to forget.
  
 So there is no re-bootstrap storm at the boundary.  For the one map prune does touch, it is read only to reach committee members (`auth_to_peer` feeds `DialBls`/`SendRequest`/`OpenStream`), and it only evicts keys that are neither pinned nor in any committee slot, so a member you'd actually dial is never pruned.  Missing-authority discovery is emitted only for current/next committee members with missing info and runs before the prune (manager.rs:691-692, then :702).  Nothing aims observers at committee members.

PR #859, opened 2026-07-06
test(network-libp2p): replace fixed real-time sleeps in network_tests  with condition polling

### LINKED ISSUE #835: test: replace fixed real-time sleeps in network_tests.rs with condition polling
## Summary

Several tests in `network_tests.rs` wait a fixed wall-clock interval for libp2p state (gossipsub mesh formation, connected-peer set, or connection-close plus pending-request cleanup) to converge, then assert exact state. The margins are generous on loopback but the shape is timing-dependent and can flake under CI CPU starvation. Replace each fixed sleep with a bounded poll for the actual condition.

## Affected test(s)

- `test_publish_to_one_peer`
- `test_msg_verification_ignores_unauthorized_publisher`
- `test_multi_peer_mesh_formation`
- `test_valid_req_res_connection_closed_cleanup`

## Root cause

Each test sleeps for a fixed duration as a proxy for asynchronous swarm convergence, then asserts.

The two gossip publish tests use a 500ms sleep as a stand-in for mesh formation before publishing:

`crates/network-libp2p/src/tests/network_tests.rs:798`
```rust
// sleep for gossip connection time lapse
tokio::time::sleep(Duration::from_millis(500)).await;
```

The same pattern repeats in `test_msg_verification_ignores_unauthorized_publisher`:

`crates/network-libp2p/src/tests/network_tests.rs:861`
```rust
// sleep for gossip connection time lapse
tokio::time::sleep(Duration::from_millis(500)).await;
```

`test_multi_peer_mesh_formation` sleeps for two heartbeats (`TEST_HEARTBEAT_INTERVAL = 1`, so 2s) and then asserts an exact connected-peer count:

`crates/network-libp2p/src/tests/network_tests.rs:1385`
```rust
// Wait for connections to stabilize
tokio::time::sleep(Duration::from_secs(TEST_HEARTBEAT_INTERVAL * 2)).await;
// Verify all peers are connected to target
let connected_peers = target_peer.network_handle.connected_peer_ids().await?;
assert_eq!(connected_peers.len(), other_peers.len(), "All peers should be connected to target");
```

(`TEST_HEARTBEAT_INTERVAL` is defined at `crates/network-libp2p/src/tests/common.rs:16`.)

`test_valid_req_res_connection_closed_cleanup` aborts peer2, sleeps 500ms, then asserts the pending-request count dropped to zero:

`crates/network-libp2p/src/tests/network_tests.rs:361`
```rust
// allow peer1 to process disconnect
tokio::time::sleep(Duration::from_millis(500)).await;
// peer1 removes pending requests
let count = peer1.get_pending_request_count().await?;
assert_eq!(count, 0);
```

The pending-request count is `self.outbound_requests.len()` (`crates/network-libp2p/src/consensus.rs:721`), and the entry is only removed when the request-response behaviour surfaces an outbound failure for the closed connection:

`crates/network-libp2p/src/consensus.rs:1029`
```rust
// try to forward error to original caller
let _ = self.outbound_requests.remove(&(peer, request_id)).map(|ack| {
    let _ = ack.send(Err(error.into()));
});
```

That removal is driven by the swarm connection-close. The peer manager also observes the same close independently at `crates/network-libp2p/src/peers/behavior.rs:128` (`FromSwarm::ConnectionClosed`), but that path is its own bookkeeping and is not what clears `outbound_requests`. The 500ms sleep is a guess at how long the full chain (abort task, QUIC teardown, swarm event, outbound-failure delivery) takes.

## Why it flakes

The sleep does not synchronize with the event it is waiting for; it just bets the event lands first. Under CPU contention the ordering can invert:

- Gossip tests: if mesh grafting or subscription propagation has not completed by the time the sleep expires, `publish` reaches no mesh peers and the subsequent `timeout(...recv())` never receives the message.
- `test_multi_peer_mesh_formation`: the `assert_eq!` demands the exact final peer count. If any of the dials or heartbeats has not settled within 2s, the count is short and the equality fails. (The later mesh-membership checks are already written defensively with `assert!(!...is_empty())`, but the connected-count assertion is not.)
- `test_valid_req_res_connection_closed_cleanup`: if the aborted-peer teardown plus outbound-failure delivery takes longer than 500ms, `get_pending_request_count()` still returns 1.

These are wall-clock races. The margins are wide enough that they pass locally and are rare, but they remain a possible intermittent failure on a busy CI runner.

## Suggested fix

Poll for the real condition with a deadline instead of a single fixed sleep. The file already has a precedent helper, `wait_for_peer_discovery` (`crates/network-libp2p/src/tests/network_tests.rs:212`), which loops on a condition until a deadline; mirror that shape (or prefer awaiting the relevant swarm/network event where one is exposed).

- [ ] `test_publish_to_one_peer` (`:799`): poll until the publisher observes the subscriber in the mesh / subscription for `TEST_TOPIC` before publishing, instead of `sleep(500ms)`.
- [ ] `test_msg_verification_ignores_unauthorized_publisher` (`:862`): same condition poll as above before the first publish.
- [ ] `test_multi_peer_mesh_formation` (`:1385`): poll `connected_peer_ids()` until it reaches `other_peers.len()` (with a deadline) before the `assert_eq!` at `:1389`.
- [ ] `test_valid_req_res_connection_closed_cleanup` (`:362`): poll `get_pending_request_count()` until it reaches 0 (with a deadline), or await the observable disconnect signal, instead of `sleep(500ms)`.
- [ ] Consider a small shared `wait_until(deadline, || cond)` helper alongside `wait_for_peer_discovery` so each test keeps a bounded timeout that fails with a clear message rather than a bare equality mismatch.

## Severity / notes

Medium: these are test-only correctness issues (no production code change), but they are a possible source of intermittent CI failures and retries. Converting fixed sleeps to bounded condition polls removes the race while keeping an upper bound on runtime.

MavenRain:
## What
  Replaces the four fixed `tokio::time::sleep(...)` "wait for convergence" calls in
  `crates/network-libp2p/src/tests/network_tests.rs` with bounded condition polls.

  ## Why
  Each sleep is a stand-in for an async swarm event and just bets the event lands
  before the timer fires. Under CI CPU starvation the ordering can invert and the
  following assertion runs against not-yet-settled state, producing rare
  intermittent failures. Polling the real condition removes the race while keeping
  an upper bound on runtime.

  ## Changes
  - New `wait_until(timeout, description, condition)` helper alongside the existing
    `wait_for_peer_discovery` precedent: polls every 10ms until the condition holds
    or the deadline elapses, failing with a descriptive message.
  - `test_valid_req_res_connection_closed_cleanup`: poll
  `get_pending_request_count()`
    to 0 instead of `sleep(500ms)`.
  `test_msg_verification_ignores_unauthorized_publisher`:
    poll `all_peers()` until the publisher has observed a `TEST_TOPIC` subscriber
    before publishing. The publisher is not subscribed to the topic, so its
    `mesh_peers` set is empty; `all_peers()` is the correct fanout precondition.
  - `test_multi_peer_mesh_formation`: poll `connected_peer_ids()` until it reaches
    `other_peers.len()` before the exact-count `assert_eq!`, instead of
    `sleep(2 * heartbeat)`.

  Test-only; no production code changed. Other sleeps in the file belong to tests
  outside this issue's scope and are untouched.

  Closes #835

### DIFF
--- a/crates/network-libp2p/src/tests/network_tests.rs
+++ b/crates/network-libp2p/src/tests/network_tests.rs
@@ -228,6 +228,35 @@ async fn wait_for_peer_discovery<Req: TNMessage, Res: TNMessage>(
     }
 }
 
+/// Poll an async `condition` until it returns `true`, or fail once `timeout_duration` elapses.
+///
+/// A bounded, self-synchronizing replacement for a fixed `sleep` that only guesses
+/// at how long an asynchronous swarm event (gossip mesh grafting, subscription
+/// propagation, connection teardown plus pending-request cleanup) takes. It waits
+/// for the real observable condition and fails with a clear message on timeout
+/// rather than asserting against not-yet-settled state. Mirrors the poll shape of
+/// [`wait_for_peer_discovery`] but is generic over the polled condition.
+async fn wait_until<F, Fut>(
+    timeout_duration: Duration,
+    description: &str,
+    mut condition: F,
+) -> eyre::Result<()>
+where
+    F: FnMut() -> Fut,
+    Fut: std::future::Future<Output = eyre::Result<bool>>,
+{
+    let deadline = tokio::time::Instant::now() + timeout_duration;
+    loop {
+        if condition().await? {
+            return Ok(());
+        }
+        if tokio::time::Instant::now() >= deadline {
+            return Err(eyre!("timed out waiting for condition: {description}"));
+        }
+        tokio::time::sleep(Duration::from_millis(10)).await;
+    }
+}
+
 #[tokio::test]
 async fn test_valid_req_restt() -> eyre::Result<()> {
     // start honest peer1 network
@@ -359,8 +388,15 @@ async fn test_valid_req_res_connection_closed_cleanup() -> eyre::Result<()> {
     peer2_network_task.abort();
     assert!(peer2_network_task.await.unwrap_err().is_cancelled());
 
-    // allow peer1 to process disconnect
-    tokio::time::sleep(Duration::from_millis(500)).await;
+    // wait for peer1 to observe the disconnect and drain its pending request,
+    // instead of guessing at how long the teardown plus outbound-failure delivery takes
+    let handle = &peer1;
+    wait_until(
+        Duration::from_secs(5),
+        "peer1 clears its pending request after peer2 disconnects",
+        move || async move { Ok(handle.get_pending_request_count().await? == 0) },
+    )
+    .await?;
 
     // peer1 removes pending requests
     let count = peer1.get_pending_request_count().await?;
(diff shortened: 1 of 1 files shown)


grantkee (approved):
Awesome! This is a much better approach than sleeping all over the place. We should continue to expand on this approach. Good thinking

PR #861, opened 2026-07-06
test: reserve test ports from per-nextest-slot windows (#830)

### LINKED ISSUE #830: Ephemeral-port TOCTOU flakes parallel e2e tests
## Summary
`get_available_tcp_port` binds `127.0.0.1:0`, records the OS-assigned port in a process-local set, closes the listener, and returns the bare port number. The node binds that port later from a spawned thread, so nothing owns the port during the gap. Under nextest the e2e test-group runs each test in a separate process, so the in-process dedup cannot stop two processes from being handed the same freed port, and the OS is free to reassign it inside the close-then-rebind window.

The existing comment on the dedup set already acknowledges the hazard ("avoid TOCTOU issues with port generation"), but the guard is per-process only, so it does not coordinate across nextest processes.

## Affected test(s)
Multi-node e2e tests that allocate ephemeral ports and hand them to spawned nodes.

Run by default in the `e2e` test-group (both `#[tokio::test]`, both go through `spawn_local_testnet`):
- `test_faucet_drip_tel_and_xyz_e2e` (faucet.rs:22)
- `test_genesis_with_consensus_registry_accounts` (genesis_tests.rs:91)

`test_precompile_genesis_accounts` (genesis_tests.rs:53) does not spawn nodes or allocate ports, so it is not affected.

Not affected under the intended invocation: `test_epoch_boundary` (epochs.rs:407) and `test_epoch_sync` (epochs.rs:445) allocate ephemeral ports the same way, but both carry `#[ignore = "only run independently from all other it tests"]`. They are excluded from the default run and are explicitly meant to run alone, so the cross-process race does not apply when that annotation is honored.

The `#[ignore]`d restart/sync suite (`test_restartstt`, `test_restarts_observer`, `test_observer_late_join_catchup`, `test_observer_reconnect_after_pause` in restarts.rs; `test_observer_pack_imports_after_epoch_close` in sync.rs) allocates ports through the same helper directly in the test bodies. It hits the same race, but only when its separate ignored step runs two of these tests concurrently under `max-threads = 2`.

## Root cause

The helper binds `:0`, reads the assigned port, and then drops the listener before returning, so the port is free the instant it hands the number back.

`crates/types/src/helpers.rs:55`
```rust
fn get_ephemeral_port(host: &str, socket_type: SocketType) -> std::io::Result<u16> {
    match socket_type {
        SocketType::Tcp => {
            let listener = TcpListener::bind((host, 0))?;
            Ok(listener.local_addr()?.port())
        }
```

Dedup is a single process-local static, so it only protects allocations within one process.

`crates/types/src/helpers.rs:10`
```rust
static USED_PORTS: Mutex<Vec<u16>> = Mutex::new(Vec::new());
```

`crates/types/src/helpers.rs:38`
```rust
let mut used_ports = USED_PORTS.lock();
if !used_ports.contains(&port) {
    used_ports.push(port);
    return Ok(port);
}
```

The testnet helper (`spawn_local_testnet`) allocates two ports per validator up front, then the node binds them later inside a spawned thread.

`crates/e2e-tests/src/lib.rs:179`
```rust
let rpc_port = tn_types::get_available_tcp_port("127.0.0.1")
    .expect("Failed to get an ephemeral rpc port");

let ws_port = tn_types::get_available_tcp_port("127.0.0.1")
    .expect("Failed to get an ephemeral ws port");
```

`crates/e2e-tests/src/lib.rs:211`
```rust
std::thread::spawn(move || {
    let runtime = tokio::runtime::Builder::new_multi_thread()
        // ...
    runtime.block_on(async move {
        match command.execute(dir, key_config, /* ... */) {
```

## Why it flakes
1. Process A calls the helper: OS assigns port P, A records P in its `USED_PORTS`, the listener is dropped, P is now free.
2. Process B (a second e2e test, allowed by `max-threads = 2`) calls the helper. B has its own fresh `USED_PORTS` static, so the cross-process guard is absent. The OS can hand B the same free port P. (Within a single process this cannot recur: `USED_PORTS.contains(&P)` is already true, so the retry loop skips P.)
3. Some milliseconds later both spawned node threads reach `command.execute`, each trying to bind P. One wins, the other fails to bind, and the test times out or errors.

`.config/nextest.toml` runs the group at up to two concurrent processes:
```toml
[test-groups.e2e]
max-threads = 2
```

and the ci profile currently masks the intermittent failures:
```toml
[profile.ci]
retries = 2
```

## Notes on likelihood
The race requires two port-allocating e2e tests to run concurrently in separate processes. In the default run that is exactly the faucet and genesis pair under `max-threads = 2`, so it is a real, reproducible flake rather than a purely theoretical one. Single-test, run-alone cases (the `#[ignore]`d epoch tests when honored) only expose the generic close-then-rebind ephemeral-port window against unrelated OS processes, which is real but rare.

## Suggested fix
- [ ] Hold the bound listener open through node startup and pass the fd/listener (not just the port number) into the node, so there is no close-then-rebind gap.
- [ ] Or have the node bind `:0` itself and report the chosen address back to the test, removing the pre-allocate step entirely.
- [ ] Or reserve ports from a cross-process-coordinated allocator (file lock or a fixed per-worker range keyed on the nextest slot) instead of the process-local `USED_PORTS`.
- [ ] Document that `USED_PORTS` is a per-process guard only and does not coordinate across nextest processes.
- [ ] Once the race is closed, drop the e2e `retries` back toward 0 so real regressions are not hidden.

MavenRain:
## Summary

`get_available_port` (`crates/types/src/helpers.rs`) binds `:0`, records the kernel-assigned
port in the process-local `USED_PORTS` set, drops the listener, and returns the bare number.
The node binds that port later from a spawned thread, so nothing owns the port during the gap.
nextest runs each test in its own process, and each process starts with a fresh `USED_PORTS`,
so the guard cannot stop two concurrent e2e tests (`max-threads = 2`) from being handed the
same freed port inside the close-then-rebind window.  Both node threads then race to bind the
same port; one loses and the test errors or times out.  Full trace in #830.

## Fix

Cross-process coordination via the nextest slot (issue suggested-fix option 3):

- When `NEXTEST_TEST_GLOBAL_SLOT` is set (nextest exports it to every test process, unique
  among tests running at the same time), `get_available_port` draws from a 256-port window
  keyed on the slot: `slot_window(slot) = 16384 + (slot % 64) * 256`, half-open, inside the
  fixed band `16384..32768`.
- Windows are disjoint per slot, so two concurrent test processes can never probe the same
  port.  The band sits deliberately below every OS ephemeral range (Linux 32768..60999,
  macOS/IANA 49152..65535), so the kernel never auto-assigns a band port to an unrelated
  process; only an explicit bind could collide, and nothing else in the test run binds there.
- Ports in the window are probed in order: the per-process guard is checked first, so ports
  already handed out are skipped without touching a socket (a transient probe bind could
  otherwise collide with the earlier port's consumer binding for real), then `can_bind` +
  `claim`.  A fully saturated window falls back to the historical ephemeral allocator instead of failing,
  so a heavily loaded host degrades to the old behavior.  Because degrading reopens the
  cross-process race while `retries = 0` assumes it closed, every degraded edge (window
  exhausted, slot beyond band capacity, `NEXTEST_TEST_GLOBAL_SLOT` present but unparsable)
  is announced on stderr, once per process (the slot lookup is memoized): a failure one
  causes is attributable in the captured test output instead of surfacing as an unexplained
  flake.  Saturation needs a single test process to claim more than 256 ports (nextest runs
  one test per process), so in practice these messages should never appear.
- Slot windows are disjoint within one nextest run, but a second nextest invocation running
  concurrently on the same host reuses the same slot numbers.  So each run probes its window
  cyclically from a starting offset hashed from `NEXTEST_RUN_ID` (fixed within a run,
  different across runs): concurrent runs walk separate stretches of a shared window and can
  only collide after one run's allocations wrap into the other's stretch, instead of racing
  deterministically for the window's first port.
- No `NEXTEST_TEST_GLOBAL_SLOT` (plain `cargo test`, non-test callers): kernel-assigned
  ephemeral ports, exactly the historical behavior.
- `USED_PORTS` is kept as the intra-process dedup and documented as a per-process guard only
  (issue box 4).
- With the race closed, the ci-profile e2e `retries` drop from 2 to 0 (issue box 5) so real
  regressions surface instead of being silently retried away.

Band capacity: the band holds 64 windows, and nextest global slots are dense small integers
assigned from actual concurrency, so exceeding capacity requires 65 simultaneously running
port-allocating test processes (the e2e group, where the port-heavy tests live, caps at
`max-threads = 2`).  A slot past capacity is rejected up front and degrades loudly to the
ephemeral path rather than wrapping onto a window another process is using, where the two
would probe the same ports deterministically.  If CI ever runs that wide, `SLOT_WINDOW` can
shrink (128 ports x 128 slots fills the same band); 256 ports per window comfortably exceeds
the few dozen ports a single e2e test allocates across its nodes.

Not changed here: the issue's alternative fixes (holding the bound listener open through node
startup, or having the node bind `:0` and report back) would remove the gap entirely but
require plumbing listeners/addresses through node startup; the slot allocator closes the
cross-process race without touching node code.  One latent hazard is documented rather than
coded around: spawned node children inherit `NEXTEST_TEST_GLOBAL_SLOT`, so a child code path
that ever called this allocator would compute its parent's window; no such path exists today,
and the doc on `nextest_global_slot` says to clear `NEXTEST_*` at the spawn site if one is
added.

## Tests

- `cargo test -p tn-types`: 92 passed, including 6 new unit tests (window disjointness across
  all slot pairs, band containment, wrap totality, out-of-capacity slot rejection, windowed
  reservation lands in its slot's window regardless of run offset, saturated window falls
  back to an out-of-window port).
- `cargo nextest run -p e2e-tests -E 'test(test_genesis_with_consensus_registry_accounts)'`
  passes with the allocator active (nextest sets `NEXTEST_TEST_GLOBAL_SLOT`, so the e2e path
  draws from the slot band end to end).
- The two port-heaviest ignored tests pass when run concurrently, two processes with the
  allocator active on distinct slots: `cargo nextest run -p e2e-tests --run-ignored all -E
  'test(test_restartstt) or test(test_restarts_observer)'` (2 passed, ~7 min each side by
  side under `max-threads = 2`).
- The pre-fix failure needs two e2e processes to draw the same port in a millisecond window,
  so there is no deterministic fail-before repro; the disjoint windows remove the overlap by
  construction.

Found while verifying, unrelated to this diff: `test_faucet_drip_tel_and_xyz_e2e` (behind
`--features faucet`, not enabled by the CI `nextest run --workspace` invocation) currently
fails during pre-genesis setup at `faucet.rs:224` ("faucet address missing from bundle
state"), before `spawn_local_testnet` allocates any port.  Happy to file it separately.

## Checklist

- [x] cross-process-coordinated allocator keyed on the nextest slot (issue box 3)
- [x] `USED_PORTS` documented as per-process only (issue box 4)
- [x] e2e `retries` back to 0 (issue box 5)
- [x] `cargo +nightly fmt` clean
- [x] `cargo clippy -p tn-types --all-targets` clean (2 pre-existing warnings in
  `committee.rs` / `epoch.rs`, untouched by this diff)
- [x] `cargo test -p tn-types` 92 passed
- [x] concurrent e2e pair green (`test_restartstt` + `test_restarts_observer`, 2/2)

Closes #830.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -24,7 +24,13 @@ filter = "package(e2e-tests)"
 test-group = "e2e"
 slow-timeout = { period = "720s", terminate-after = 3 }
 
+# The cross-process slot allocator (tn_types::helpers) closes the ephemeral-port race,
+# so e2e tests no longer need retries to mask it. Real regressions now surface instead of
+# being silently retried away. The allocator's degraded edges (window exhausted, slot
+# beyond the band's capacity) announce themselves on stderr, so a failure they cause is
+# attributable in the captured output. Other packages keep the profile-wide retries = 2.
 [[profile.ci.overrides]]
 filter = "package(e2e-tests)"
 test-group = "e2e"
+retries = 0
 slow-timeout = { period = "720s", terminate-after = 3 }
--- a/crates/types/src/helpers.rs
+++ b/crates/types/src/helpers.rs
@@ -2,12 +2,43 @@
 
 use parking_lot::Mutex;
 use secp256k1::PublicKey;
-use std::net::{TcpListener, UdpSocket};
+use std::{
+    collections::hash_map::DefaultHasher,
+    hash::{Hash, Hasher},
+    net::{TcpListener, UdpSocket},
+    sync::OnceLock,
+};
 
+/// Number of probe attempts before giving up on finding a free port.
 const MAX_RETRIES: u32 = 1000;
 
-/// Track any ports we have given out already to avoid TOCTOU issues with port generation.
+/// Ports this process has already handed out.
+///
+/// This guard is **per-process only**. `cargo nextest` runs each test in its own
+/// process, and each process starts with a fresh, empty `USED_PORTS`, so this set
+/// cannot stop two concurrent test processes from being handed the same freed port.
+/// Cross-process disjointness is provided separately by [`slot_window`], which carves
+/// a fixed port band into non-overlapping windows keyed on the nextest slot. Keep both:
+/// `USED_PORTS` dedups *within* a process, the slot window dedups *across* processes.
 static USED_PORTS: Mutex<Vec<u16>> = Mutex::new(Vec::new());
+
+/// Lowest port of the cross-process coordination band.
+///
+/// The band sits deliberately **below** the operating system's ephemeral range (Linux
+/// 32768-60999, macOS/IANA 49152-65535). The kernel never auto-assigns a port from this
+/// band to an unrelated process, so once a test process picks a port here nothing else
+/// on the host will race it into the close-then-rebind gap.
+const SLOT_BAND_START: u16 = 16384;
+
+/// One past the highest port of the coordination band (exclusive): `16384..32768`.
+const SLOT_BAND_END: u16 = 32768;
+
+/// Ports reserved per nextest slot. `256 * 64` slots exactly fills the band, and 256
+/// comfortably exceeds the few dozen ports a single e2e test allocates across its nodes.
+const SLOT_WINDOW: u16 = 256;
+
+/// Number of disjoint slot windows the band is divided into.
+const SLOT_COUNT: u16 = (SLOT_BAND_END - SLOT_BAND_START) / SLOT_WINDOW;
 
 /// Represents the type of socket to create
 #[derive(Debug, Clone, Copy)]
@@ -31,18 +62,15 @@ pub enum PortError {
     NoPortsAvailable,
 }
 
-/// Get an available port with the specified configuration
+/// Get an available port with the specified configuration.
+///
+/// When running under `cargo nextest` the port is drawn from this process's slot window
+/// (see [`slot_window`]) so concurrent test processes never collide. Otherwise the port
+/// is a kernel-assigned ephemeral one, preserving the historical behavior for plain
+/// `cargo test` and non-test callers.
 pub fn get_available_port(config: &PortConfig) -> Result<u16, PortError> {
-    for _ in 0..config.max_retries {
-        if let Ok(port) = get_ephemeral_port(&config.host, config.socket_type) {
-            let mut used_ports = USED_PORTS.lock();
-            if !used_ports.contains(&port) {
-                used_ports.push(port);
-                return Ok(port);
-            }
-        }
-    }
-    Err(PortError::NoPortsAvailable)
+    nextest_global_slot()
+        .map_or_else(|| reserve_ephemeral_port(config), |slot| reserve_windowed_port(config, slot))
 }
 
 impl From<std::io::Error> for PortError {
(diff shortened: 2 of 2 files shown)


sstanfield (approved):

grantkee (approved):
Good work. This was infrequent for me, but very annoying. This should go a long way to reduce port collisions during testing

Issue #834, opened 2026-07-03
test: use an injectable clock for BannedPeerCache expiry tests


MavenRain:
## Summary

`BannedPeerCache` expiry tests drive time with real `std::thread::sleep` against a 100ms TTL, so a scheduling stall on a single sleep can push a peer past its expiry boundary a heartbeat early and fail an exact `expired.len()` assertion. The tests are plain synchronous `#[test]`s (no async runtime, no threads, no barriers), so the only source of nondeterminism is `sleep` overshoot.

## Affected test(s)

- `crates/network-libp2p/src/tests/cache_peers.rs::test_remove_expired` (primary, tightest margin)
- `crates/network-libp2p/src/tests/cache_peers.rs::test_remove_expired_with_reinsertions`
- `crates/network-libp2p/src/tests/cache_peers.rs::test_remove_expired_ordering`

## Root cause

Expiry is measured against `Instant::now()` at insertion and again at heartbeat, so the tests can only control timing with real sleeps.

`crates/network-libp2p/src/peers/cache.rs:53`
```rust
self.list.push_back(Element { key, inserted: Instant::now() });
```

`crates/network-libp2p/src/peers/cache.rs:89`
```rust
let now = Instant::now();
```

`crates/network-libp2p/src/peers/cache.rs:93`
```rust
if element.inserted + self.duration > now {
    self.list.push_front(element);
    break;
}
```

`test_remove_expired` uses a 100ms TTL, inserts `peer3` after a 25ms sleep, then sleeps 76ms before the deciding heartbeat.

`crates/network-libp2p/src/tests/cache_peers.rs:30`
```rust
let mut cache = BannedPeerCache::new(Duration::from_millis(100));
```

`crates/network-libp2p/src/tests/cache_peers.rs:40`
```rust
std::thread::sleep(Duration::from_millis(25));
```

`crates/network-libp2p/src/tests/cache_peers.rs:53`
```rust
std::thread::sleep(Duration::from_millis(76));
```

`crates/network-libp2p/src/tests/cache_peers.rs:58`
```rust
let expired = cache.heartbeat();
assert_eq!(expired.len(), 1, "Peer1 should be expired");
```

## Why it flakes (real but rare)

[...]

PR #862, opened 2026-07-06
test: use an injectable clock for BannedPeerCache expiry tests

MavenRain:
## Summary

Fixes the flaky `BannedPeerCache` expiry tests reported in #834 by removing their dependence on wall-clock `std::thread::sleep`.

The tests measured a 100ms TTL with real sleeps, so a scheduling stall on a single sleep could push a peer past its expiry boundary a heartbeat early and fail an exact `expired.len()` assertion.  Because `sleep` only ever overshoots, a ~24ms overshoot concentrated on one sleep was enough to flip `test_remove_expired`.

## Changes

- Add a `Clock` trait (`now() -> Instant`) that timestamps insertions and drives expiry in `BannedPeerCache`.
- Wire production through a defaulted type parameter `BannedPeerCache<Key, C = SystemClock>`, where `SystemClock` returns `Instant::now()`. `manager.rs` (`BannedPeerCache<PeerId>` / `BannedPeerCache::new`) needs no changes and production behavior is identical.
- Add a `#[cfg(test)]` `ManualClock` (a cheaply cloneable `Rc<Cell<Instant>>`) and a `#[cfg(test)] with_clock` constructor, so the test clock adds zero production API surface.
- Rewrite `test_remove_expired`, `test_remove_expired_with_reinsertions`, and `test_remove_expired_ordering` to advance the fake clock deterministically and drop every `std::thread::sleep`. The tests now run instantly and no longer depend on scheduler timing.

## Verification

- `cargo nextest run -p tn-network-libp2p` cache tests: 5 passed, 0 failed
- `cargo clippy -p tn-network-libp2p --all-targets --all-features` (stable + nightly-2026-03-20): 0 warnings
- `cargo +nightly-2026-03-20 fmt`: no changes

### DIFF
--- a/crates/network-libp2p/src/peers/cache.rs
+++ b/crates/network-libp2p/src/peers/cache.rs
@@ -9,6 +9,56 @@ use std::{
 #[path = "../tests/cache_peers.rs"]
 mod cache_peers;
 
+/// Source of the current [`Instant`] used to timestamp insertions and evaluate expiry.
+///
+/// Production uses [`SystemClock`], which reads the real monotonic clock. Tests inject a manually
+/// advanced clock so expiry is deterministic and does not depend on wall-clock sleeps.
+pub(super) trait Clock {
+    /// Return the current instant.
+    fn now(&self) -> Instant;
+}
+
+/// The real monotonic clock backed by [`Instant::now`].
+#[derive(Debug)]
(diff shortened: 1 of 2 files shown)


grantkee (approved):
Great solution!

PR #864, opened 2026-07-07
test: remove wall-clock timing assumptions from test_fetch_cancellation_on_success

### LINKED ISSUE #832: test: remove wall-clock timing assumptions from test_fetch_cancellation_on_success
## Summary

`test_fetch_cancellation_on_success` couples its main assertion to real wall-clock timing against the fetcher's staggered per-peer spawn timer, so it can flake under loaded CI scheduling. It should await an explicit cancellation signal instead of racing the 100ms spawn interval, and poll the store instead of assuming a fixed post-reply delay.

## Affected test(s)

- `test_fetch_cancellation_on_success` (`crates/consensus/primary/src/tests/certificate_fetcher_tests.rs:450`, `#[tokio::test(flavor = "current_thread")]`, real clock: `sleep`/`timeout` come from `tokio::time` and the test does not set `start_paused`).

## Root cause

The test shortens the staggered spawn interval from its 5s default to 100ms:

`crates/consensus/primary/src/tests/certificate_fetcher_tests.rs:452`
```rust
.with_consensus_parameters(Parameters {
    parallel_fetch_request_delay_interval: Duration::from_millis(100),
    ..Default::default()
})
```

Inside `fetch_from_peers`, that value flows in as `fallback_delay` (`certificate_fetcher.rs:268`, `let fallback_delay = self.parallel_fetch_request_delay_interval;`). The first peer is spawned immediately, then a real-clock timer governs when the next peer is spawned:

`crates/consensus/primary/src/certificate_fetcher.rs:442`
```rust
// timer for spawning next peer request
let mut spawn_timer = Box::pin(sleep(fallback_delay));
```

`crates/consensus/primary/src/certificate_fetcher.rs:486`
```rust
_ = &mut spawn_timer, if peer_index < peers.len() => {
    // time to spawn next peer fetch
    spawn_peer_fetch(/* ... */);
    peer_index += 1;
    // reset timer for next spawn
    spawn_timer = Box::pin(sleep(fallback_delay));
}
```

The default committee has 4 authorities (`crates/test-utils-committee/src/builder.rs:38`, `committee_size: NonZeroUsize::new(4)`), and the peer list excludes self (`crates/consensus/primary/src/certificate_fetcher.rs:370-374`, `others_primaries_by_id`), so `peers.len() == 3` and the `peer_index < peers.len()` guard is armed. The test therefore leans on two independent real-clock timing assumptions.

### (a) The "no second request" assertion races the 100ms spawn timer

The test replies to the first peer, then asserts no second request is sent:

`crates/consensus/primary/src/tests/certificate_fetcher_tests.rs:522`
```rust
reply
    .send(Ok(NetworkResponseMessage {
        peer,
        result: PrimaryResponse::RequestedCertificates(all_certificates.clone()),
    }))
    .unwrap();
```

`crates/consensus/primary/src/tests/certificate_fetcher_tests.rs:533`
```rust
let timeout_result = timeout(Duration::from_millis(500), fake_receiver.recv()).await;
// should timeout - no second request should be made after success
assert!(timeout_result.is_err(), "No additional requests should be made after success");
```

For this assert to hold, the reply must be delivered and drained by `rx_results.recv` (which fires `cancel_signal.notify()`, `certificate_fetcher.rs:468`) before the 100ms `spawn_timer` elapses. If the timer wins, `spawn_peer_fetch` runs, a second `SendRequest` lands in `fake_receiver` (buffer 1000), the `recv()` at :533 returns `Ok`, and the `is_err()` assert fails. Cancellation firing afterward does not un-send the already-buffered request.

### (b) The persistence check follows a fixed sleep rather than the write

`crates/consensus/primary/src/tests/certificate_fetcher_tests.rs:539`
```rust
sleep(Duration::from_millis(100)).await;
for cert in &round1_certs {
    assert!(
        certificate_store.read(cert.digest()).unwrap().is_some(),
        "Round 1 certificates should be stored after fetch"
    );
}
```

This is a separate hardcoded 100ms (not the `parallel_fetch_request_delay_interval`). The store write is not synchronous with the reply: `process_fetched_certificates_in_parallel` (`cert_validator.rs:233`) sig-verifies via `verify_collection` and then hands off to `forward_verified_certs` (`cert_validator.rs:131`), which awaits the certificate manager over a oneshot before the write lands:

`crates/consensus/primary/src/state_sync/cert_validator.rs:216`
```rust
let (reply, res) = oneshot::channel();
self.consensus_bus
    .certificate_manager()
    .send(CertificateManagerCommand::ProcessVerifiedCertificates { certificates, reply })
    .await?;
// await response from certificate manager
res.await.map_err(|_| CertManagerError::CertificateManagerOneshot)?
```

This pipeline runs on the fetcher's own spawned task and is never awaited by the test.

## Why it flakes

Everything past the reply runs off-test on spawned tasks under a single-threaded `current_thread` runtime with a real clock.

- (a) is the primary and most plausible flake. The full loop of spawn to network-send to test-recv to reply to `tx_results` to `rx_results.recv` cancel must complete within 100ms of the first spawn. On an idle machine this is microseconds, but a ~100ms scheduling stall on a loaded CI runner lets the `spawn_timer` fire first, emitting a real second request and tripping the `is_err()` assert.
- (b) is real but rare. The effective budget for the store write is not 100ms: the assert at :536 requires the `timeout(500ms, ..)` at :533 to elapse in full (no second request), so verify plus manager handoff plus write get roughly 500ms + 100ms of wall clock after the reply before :542 reads. It only fails if that pipeline exceeds ~600ms, which is a much larger stall than (a) needs.

Both windows are fixed real-clock durations racing async work, which is the classic recipe for a scheduling-dependent flake, with (a) as the tight one.

## Suggested fix

- [ ] Drive the "no second request" check off an explicit cancellation event rather than the `spawn_timer` not firing: have the fetcher expose (or the test observe) a notify/oneshot that fires once one peer's response cancels the others, and await that before asserting no additional `SendRequest`.
- [ ] Replace `sleep(Duration::from_millis(100))` before the persistence check with polling `certificate_store.read(..)` under a generous deadline (retry until present or timeout), or await a store-write completion signal.
- [ ] Optionally decouple the test from the exact `parallel_fetch_request_delay_interval` value so the (a) window is not pinned to that 100ms constant, and consider `tokio::test(start_paused = true)` so the spawn timer advances deterministically rather than against the real clock.

## Severity / notes

Medium, trending low. Test-only flakiness, no production impact. The underlying cancel-on-success and persistence behavior appears correct; the test synchronizes on wall-clock timing rather than on the events it means to observe. In practice (a) is the realistic flake and (b) is unlikely given the ~600ms effective budget.

MavenRain:
Closes #832.

`test_fetch_cancellation_on_success` synchronized on wall-clock durations rather
than on the events it means to observe, so it could flake under loaded CI.  This
makes it deterministic without changing any production code.

## What was flaky

- **The "no second request" assertion raced the spawn timer.** The test set
  `parallel_fetch_request_delay_interval` to 100ms, replied to the first peer,
  then asserted (via a 500ms `timeout` on `recv`) that no second request
  followed. For that to hold, the reply had to propagate and fire
  `cancel_signal.notify()` before the 100ms staggered-spawn timer elapsed.  On a
  loaded runner the timer could win, spawn the next peer, and land a real second
  request in the channel, tripping the assert.
- **The persistence check followed a fixed `sleep(100ms)`.** The store write
  happens off-test on the fetcher's spawned task (sig-verify, then a certificate
  manager handoff over a oneshot, then the write).  A hardcoded sleep is a bet on
  that pipeline finishing in time.

## Fix (test-only)

- [ ] Push `parallel_fetch_request_delay_interval` beyond the test's lifetime
      (`3600s`) so the fallback spawn timer cannot fire and race the
      cancellation.  This test exercises cancel-on-success, not the staggered
      fan-out, so the first peer's success now deterministically returns before
      any second peer is ever queued.
- [ ] Replace the fixed `sleep` before the persistence check with polling the
      certificate store under a generous deadline (10s), so a slow runner waits
      longer instead of failing.  Reaching the stored state also proves the fetch
      task returned (i.e. cancellation fired).
- [ ] Assert "no additional request" via `fake_receiver.try_recv()` returning
      `Empty` (matching the existing idiom in this file) instead of a 500ms
      timeout window.  With the spawn timer disabled this is exact, not a race.

There are no production code changes;  the underlying cancel-on-success and persistence
behavior is unchanged and already correct.

## Verification

- `cargo nextest run -p tn-primary -E 'test(test_fetch_cancellation_on_success)'`
- `make fmt` / `make clippy` clean on the touched crate.

### DIFF
--- a/crates/consensus/primary/src/tests/certificate_fetcher_tests.rs
+++ b/crates/consensus/primary/src/tests/certificate_fetcher_tests.rs
@@ -415,7 +415,11 @@ async fn test_fetch_certificates_basic() {
 async fn test_fetch_cancellation_on_success() {
     let fixture = CommitteeFixture::builder(MemDatabase::default)
         .with_consensus_parameters(Parameters {
-            parallel_fetch_request_delay_interval: Duration::from_millis(100),
+            // This test exercises cancel-on-success, not the staggered fan-out. Push the
+            // fallback spawn interval well beyond the test's lifetime so the timer that
+            // would spawn the *next* peer can never fire and race the cancellation. That
+            // race against a 100ms interval was the wall-clock flake in issue #832.
+            parallel_fetch_request_delay_interval: Duration::from_secs(3600),
             ..Default::default()
         })
         .randomize_ports(true)
@@ -473,7 +477,7 @@ async fn test_fetch_cancellation_on_success() {
     // should be pending due to missing parents
     assert!(result.is_err(), "Should fail due to missing parents");
 
-    // expecxt a fetch request for the missing certificates
+    // expect a fetch request for the missing certificates
     if let Some((_, _, reply)) = timeout(Duration::from_secs(2), fake_receiver.recv())
         .await
         .expect("Should get fetch request")
@@ -484,19 +488,34 @@ async fn test_fetch_cancellation_on_success() {
         panic!("Expected a fetch request for missing certificates");
     }
 
-    // wait for potential second request (should not happen due to cancellation)
-    let timeout_result = timeout(Duration::from_millis(500), fake_receiver.recv()).await;
-
-    // should timeout - no second request should be made after success
-    assert!(timeout_result.is_err(), "No additional requests should be made after success");
-
-    // verify certificates were actually stored
-    sleep(Duration::from_millis(100)).await;
-    for cert in &round1_certs {
+    // Synchronize on the observable effect of a successful fetch (the round-1
+    // certificates landing in the store) instead of assuming a fixed post-reply delay.
+    // Poll under a generous deadline so a slow CI runner simply waits longer rather than
+    // failing. Reaching this point also proves the fetch task returned, which is when the
+    // fan-out is cancelled.
+    let deadline = tokio::time::Instant::now() + Duration::from_secs(10);
+    loop {
+        let all_stored = round1_certs
+            .iter()
+            .all(|cert| certificate_store.read(cert.digest()).unwrap().is_some());
+        if all_stored {
+            break;
+        }
         assert!(
-            certificate_store.read(cert.digest()).unwrap().is_some(),
-            "Round 1 certificates should be stored after fetch"
+            tokio::time::Instant::now() < deadline,
+            "Round 1 certificates should be stored after a successful fetch"
         );
+        sleep(Duration::from_millis(10)).await;
+    }
+
+    // No further peer is queried after success. With the fallback spawn interval pushed
+    // beyond the test's lifetime, the only request ever emitted was the one answered
+    // above, so anything still queued here is a real cancel-on-success regression rather
+    // than a timing artifact.
+    match fake_receiver.try_recv() {
+        Err(TryRecvError::Empty) => {}
+        Ok(_) => panic!("No additional requests should be made after success"),
+        Err(TryRecvError::Disconnected) => panic!("Certificate fetcher unexpectedly shut down"),
     }
 }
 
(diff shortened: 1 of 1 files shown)


grantkee (approved):
I'm all for more precise tests. LGTM

PR #868, opened 2026-07-07
Fork consensus registry bls precompile

### LINKED ISSUE #869: In-protocol upgrade path for the live adiri `ConsensusRegistry` to the BLS-precompile ABI
## Problem

adiri's already-deployed `ConsensusRegistry` predates the native BLS precompile (`42f74b33`/`44dc5fff`) and the integration ABI now on `main`. We can't simply redeploy the contract on the live chain: doing so would either lose the existing validator storage or break
re-execution of pre-fork history for anyone resyncing from genesis. There is currently no mechanism for moving a live, already-deployed contract onto new bytecode at a coordinated point in the chain's history while keeping historical epoch closes byte-identical.

## Solution

Introduce an in-protocol, epoch-boundary code swap for the `ConsensusRegistry` account:

- Pin the exact pre-fork bytecode hash (`CONSENSUS_REGISTRY_PRE_FORK_CODE_HASH`) so the swap only
  ever fires over the known, currently-deployed bytecode — fail closed on anything else rather  than silently mutating unexpected code.
- At the epoch boundary, swap the account's runtime code in place while preserving its balance,  nonce, and existing storage.
- Immediately run a one-time on-chain migration (`migrateValidatorSets()`, invoked as a system  call) that back-fills the new per-status validator sets and cached eligible-validator count from  the preserved `currentStatus` data.
- Gate the committee-eligible-pool read by the registry's current code hash, so every epoch close
  before the fork continues to speak the legacy ABI (keeping full historical resync byte-identical), while closes at/after the fork speak the new ABI.

The fork fires as the first step of the epoch-closing block that concludes
`CONSENSUS_REGISTRY_FORK_EPOCH - 1`, before rewards/conclude processing, so the new committee read and eligible-count guard observe migrated state within the same block.

The target epoch itself is deliberately deferred: `crates/types/src/forks.rs` currently pins
`CONSENSUS_REGISTRY_FORK_EPOCH = u32::MAX` as a placeholder pending a follow-up PR that sets the real value once rollout timing is decided.

This is scoped entirely to the adiri testnet's specific historical situation, not a general
contract-upgrade mechanism — all of it is `#[cfg(feature = "adiri")]`-gated.

grantkee:
- Code-hash-gated dual-ABI committee read: `read_committee_eligible_pool` picks legacy or `getValidatorsInfo` based on the registry's code hash at each epoch close, so pre-fork closes stay byte-identical for historical resync.
- Fail-closed bytecode swap in `apply_consensus_registry_fork`: swaps the registry's code while preserving balance/nonce/storage, refusing to proceed if the pre-fork code hash doesn't match the pinned value.
- Swap runs `migrateValidatorSets()` as a system call immediately after, back-filling per-status validator sets and cached eligible count from the preserved `currentStatus` data.
- Sequencing: swap + migration happen first in the epoch-closing block that concludes `CONSENSUS_REGISTRY_FORK_EPOCH - 1`, before rewards/conclude, so downstream reads see migrated state in the same block.
- `extra_data` decode hardened in `config.rs`: distinguishes 0 bytes (normal) from 32 bytes (closing-epoch digest) and errors on anything else instead of panicking.
- Accepted gap: `isValidator` legacy-key lookups have a documented post-fork gap — RPC-only, not consensus-critical.
- `CONSENSUS_REGISTRY_FORK_EPOCH = u32::MAX` is a placeholder; setting the real target epoch is a deferred follow-up PR (`forks.rs` documents the pre-deploy checklist).
- `adiri` feature wiring added through Cargo.toml, Makefile, and Dockerfile so this only compiles for the testnet build.

closes #869

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,8 +147,9 @@ docker-login:
 	gcloud auth configure-docker us-docker.pkg.dev ;
 
 # build and push latest adiri image for amd64 and arm64
+# CARGO_FEATURES=adiri compiles in the testnet fork logic (ConsensusRegistry fork, ADIRI_DUP_BATCH_EPOCH).
 docker-adiri:
-	docker buildx build -f ./etc/Dockerfile --platform linux/amd64,linux/arm64 --no-cache -t us-docker.pkg.dev/telcoin-network/tn-public/adiri:$(TAG) . --push ;
+	docker buildx build -f ./etc/Dockerfile --build-arg CARGO_FEATURES=adiri --platform linux/amd64,linux/arm64 --no-cache -t us-docker.pkg.dev/telcoin-network/tn-public/adiri:$(TAG) . --push ;
 
 # push local adiri:latest to the gcloud artifact registry
 docker-push:
--- a/bin/telcoin-network/Cargo.toml
+++ b/bin/telcoin-network/Cargo.toml
@@ -14,5 +14,12 @@ exclude.workspace = true
 telcoin-network-cli = { workspace = true }
 tn-node = { workspace = true }
 
+[features]
+# Adiri-testnet build. Propagates the `adiri` feature through the whole node graph so the
+# testnet-only fork logic (`tn_types::forks`, the ConsensusRegistry fork, `ADIRI_DUP_BATCH_EPOCH`)
+# is actually compiled into the binary. Enabled for the `docker-adiri` image via a Docker build arg;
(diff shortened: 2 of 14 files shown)


MavenRain (dismissed):
LGTM

sstanfield (dismissed):

grantkee:
Rebased and fixed conflicts. Attestation passed

PR #876, opened 2026-07-07
patch: storage memory management

### LINKED ISSUE #875: patch: txn-drop wedge: no abort path stops persistence AND leaks
`crates/storage/src/layered_db.rs`

`DBMessage` has only `StartTxn`/`CommitTxn`. If any `write_txn()` handle is dropped without
`commit()`, the background thread's txn count is permanently off — **no commit ever fires again**
on that DB (the count oscillates 1↔2): writes stop persisting to disk and `committed_inserts`
grows forever (even on `cache_db`). Triggers exist on **normal** paths:

- `CertificateStore::delete` early-returns `Ok(())` with the txn open when the cert is missing.
- Error `?`-returns with a live txn in `certificate_store.rs` (write/write_all),
  `batch_fetcher.rs`, and `worker/src/network/primary.rs`.

**Fix:** an abandoned txn ends as **commit + warn**. Writes are already visible in `mem_db`
immediately, there is no rollback machinery, and overlapped logical txns share one physical txn —
an abort would discard innocent writes. Implemented with a shared `TxnGuard` (Arc) whose `Drop`
sends a new `EndTxn` message iff `commit()` was never called; exactly one `StartTxn`/end pair per
logical txn even across clones.

grantkee:
- stop LayeredDatabase retaining every insert on full-memory DBs
- end dropped write txns so persistence never wedges

closes #875
closes #876


crates/storage/src/layered_db.rs
```diff
@@ -222,6 +306,24 @@ impl<DB: Database> LayeredDatabase<DB> {
...
+    ///
+    /// Processed in queue order, so it reflects all operations sent before this call.
+    pub fn stats(&self) -> eyre::Result<LayeredDbStats> {
+        let (tx, mut rx) = oneshot::channel();
+        self.tx.send(DBMessage::Stats(tx)).map_err(|_| eyre::eyre!("DB thread gone, FATAL!"))?;
+
+        // Can not use rx.blocking_recv() because it will be called from some tokio tests and that
+        // will panic.
+        loop {
+            match rx.try_recv() {
```
MavenRain, line 327: Consider an encompassing deadline as a matter of future-proofing against an infinite livelock.

MavenRain (approved):
Nothing blocking

sstanfield (approved):

PR #880, opened 2026-07-07
fix(engine): bound the consensus-output execution backlog

### LINKED ISSUE #879: patch: Engine backlog unbounded `queued: VecDeque<ConsensusOutput>`
`crates/engine/src/lib.rs`

The engine loop drains the consensus-output channel into `queued` unconditionally while executing
one output at a time. If execution lags consensus (state grows over weeks; prior incidents of
execution-height divergence exist), `queued` grows without bound — each item carries a full
`CommittedSubDag` + batches. The `tn_engine_queued_outputs` gauge already tracks this.

**Fix:** bound the queue (`MAX_QUEUED_OUTPUTS = 8`) by gating the stream poll, and shrink the
`to_engine` channel (1000 → 64). Backpressure analysis: blocking the `to_engine` send stalls only
the EpochManager forwarder task, not DAG progress; `close_epoch` already waits for execution
catch-up, so no deadlock. BUT beyond the 100-slot `consensus_output` broadcast, a blocked
forwarder gets `Lagged` → warn-and-skip → **silent execution gap**. Backpressure therefore ships
with gap detection: a continuity check on forwarded consensus numbers in
`wait_for_epoch_boundary` that skips stale (already-forwarded) output and errors on a gap so the
restart path replays from the consensus DB (`replay_missed_consensus`) instead of diverging
silently.

grantkee:
The engine drained its consensus stream into queued: VecDeque unconditionally while executing one output at a time, so whenever execution lags consensus the backlog grows without bound - each entry a full CommittedSubDag plus batches. Combined with the 1000-slot to_engine channel this let a lagging engine buffer over a thousand outputs.

- engine: stop polling the stream while queued holds MAX_QUEUED_OUTPUTS (8). A pending execution task always exists while gated, and each completion re-opens the gate, so backpressure propagates to the to_engine channel instead of memory. tn_engine_queued_outputs already tracks occupancy.
- node: shrink to_engine 1000 -> 64 (TO_ENGINE_CAPACITY); a deep channel would just buffer the backlog one hop upstream.

Backpressure changes lag behavior: beyond the 100-slot consensus_output broadcast, a blocked epoch-manager forwarder gets Lagged -> warn-and-skip, which was previously a silent execution gap. It now ships with detection:

- wait_for_epoch_boundary classifies each output number against the last number actually forwarded to the engine (check_output_continuity): stale outputs (already forwarded, e.g. replayed from the DB) are skipped instead of double-executed, and a gap is a hard error so the restart path replays the missed output from the consensus DB (every output is saved before it is broadcast) instead of diverging silently. The replay and leftover-drain paths legitimately re-forward and are not checked.
- prime last_forwarded_consensus_number from last_consensus_parent (pack ground truth, the same anchor the subscriber numbers new output from) instead of the slot-file hint, which can be stale after a hard crash and would trip the continuity check with a spurious gap on the first live output.

Tests: engine backpressure IT (gate holds at the bound on first poll - fails with 'drained 2' without the gate - then drains to completion with every output executed exactly once, in order) and unit coverage for check_output_continuity including the genesis and u64::MAX edges.

closes #879

### DIFF
--- a/crates/engine/src/lib.rs
+++ b/crates/engine/src/lib.rs
@@ -28,6 +28,15 @@ use tracing::{debug, error, info, warn};
 /// `SealedHeader`.
 type PendingExecutionTask = oneshot::Receiver<EngineResult<SealedHeader>>;
 
+/// Maximum number of consensus outputs buffered for execution.
+///
+/// Each queued output holds a full `CommittedSubDag` plus batches, so an unbounded queue lets a
+/// lagging engine consume all memory while consensus keeps committing. Once the queue is full
+/// the engine stops polling the consensus stream, pushing backpressure into the `to_engine`
+/// channel (and ultimately the EpochManager forwarder) instead of buffering here. The
+/// `tn_engine_queued_outputs` gauge tracks occupancy.
+pub const MAX_QUEUED_OUTPUTS: usize = 8;
+
 /// The TN consensus engine is responsible executing state that has reached consensus.
 ///
 /// The engine makes no attempt to track consensus. It's only purpose is to receive output from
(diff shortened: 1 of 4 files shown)


MavenRain (dismissed):
LGTM!

sstanfield (dismissed):
Looks good.  Out of scope but we might need to revisit how we prime the last output.  We should probably looks at the last executed consensus and then skip until we have a new execution base.  Although that might also have issues and just using the last executed could leave us with a lot of extra empty executions (or maybe we skip all the empty consensus).  I think there may be room for improvement but this PR seems to be a strict improvement.

grantkee:
Rebased onto main and need another review

sstanfield (approved):

PR #882, opened 2026-07-08
fix(state-sync): request missing range from peers on observer catch-up stall

### LINKED ISSUE #836: fix: re-drive state-sync request on observer catch-up timeout instead of waiting for gossip
## Summary

The observer consensus-header catch-up loop in `spawn_stream_consensus_headers` polls the local DB cache on a fixed 100ms sleep, gives up after ~60s, then falls back to waiting for the next watch (gossip) update, which may never arrive on a slow or independently running node. On timeout it should actively re-drive a state-sync request for the missing range instead of blocking passively.

## Affected test(s)

`test_epoch_sync` and `test_epoch_boundary` in `crates/e2e-tests/tests/it/epochs.rs`, both `#[ignore = "only run independently from all other it tests"]`, so they do not run in the default CI e2e group. The observer catch-up path they exercise runs in production.

## Root cause

The catch-up retry loop caps at `MAX_NO_PROGRESS = 600` iterations of a hardcoded 100ms poll (~60s), then breaks:

`crates/state-sync/src/lib.rs:216`
```rust
let mut no_progress_count = 0u32;
const MAX_NO_PROGRESS: u32 = 600; // 600 * 100ms = 60 seconds max wait
```

`crates/state-sync/src/lib.rs:240`
```rust
if no_progress_count >= MAX_NO_PROGRESS {
    warn!(target: "state-sync", ?epoch, last_consensus_height, target=pending_header.number,
        "could not catch up to consensus target after retries, waiting for next gossip update");
    break;
}
```

Each retry only sleeps on a fixed interval:

`crates/state-sync/src/lib.rs:245`
```rust
tokio::select! {
    _ = tokio::time::sleep(Duration::from_millis(100)) => {}
    _ = &rx_shutdown => return Ok(()),
}
```

Once the inner loop breaks, control returns to the outer `select!` which just awaits the next watch update:

`crates/state-sync/src/lib.rs:259`
```rust
tokio::select! {
    _ = rx_last_consensus_header.changed() => {
        // ...
        pending_header = rx_last_consensus_header.borrow_and_update().clone().unwrap_or_default();
    }
```

The retry loop itself does not issue a network request for a missing header. `catch_up_consensus_from_to` reads the header from `ConsensusHeaderCache` / `ConsensusChain`, and on a miss it returns early, relying on the concurrent backward traversal in `spawn_track_recent_consensus` to have filled the cache:

`crates/state-sync/src/lib.rs:313`
```rust
} else {
    // ...
    // We should have all the required headers in local storage by now...
    return Ok(result_header);
};
```

Note that the function's own doc comment overstates what it does:

`crates/state-sync/src/lib.rs:273`
```rust
/// Queries peers for latest height and downloads and executes any missing consensus output.
```

The function body contains no peer query, download, or request call. It only reads local storage (`ConsensusHeaderCache` and `ConsensusChain`) and blocks on `wait_for_execution`.

## Why it flakes

The retry loop depends on a separate async backward traversal populating `ConsensusHeaderCache` within the ~60s budget. There is no barrier or join forcing that traversal to complete before catch-up runs, so the two race. If the traversal is slow (loaded machine, single-node/independent run, or a header that was never gossiped for an older epoch), the 600 polls exhaust and the loop breaks with no progress. Control then blocks in the outer `select!` on `rx_last_consensus_header.changed()`. While no further watch update is sent (the case flagged by the warn string "waiting for next gossip update"), the observer stays parked at `last_consensus_height < pending_header.number`.

A later watch update re-enters the retry loop with a fresh ~60s budget, so recovery is possible once updates resume and the cache is populated. The stall is unbounded only while updates are absent. Under default CI the covering epoch tests are `#[ignore]`d, so this interleaving is only hit on manual/independent runs and in production.

## Suggested fix

- [ ] On catch-up timeout, actively re-drive a state-sync request for the missing consensus-header range rather than breaking to a passive `changed()` wait.
- [ ] In the missing-header branch of `catch_up_consensus_from_to` (`crates/state-sync/src/lib.rs:313`), request the header from peers instead of returning early and depending solely on the backward traversal (this would also make the behavior match the doc comment at `:273`).
- [ ] Make the poll interval (currently `Duration::from_millis(100)`) and the give-up bound (currently `MAX_NO_PROGRESS = 600`) configurable rather than hardcoded.
- [ ] Consider un-`#[ignore]`ing or adding a fast, deterministic variant of `test_epoch_sync` / `test_epoch_boundary` so the catch-up path has default-CI coverage.

## Severity / notes

Medium. No consensus-safety impact, but a stalled observer stops advancing until a new update happens to arrive. Real but rare: it requires the backward traversal to miss its ~60s budget and no follow-up watch update. The gossip-wait framing is taken directly from the code's own warn message and the outer `changed()` await; the exact upstream sender of `last_consensus_header()` was not traced, so that attribution rests on the code's own wording.

MavenRain:
## Summary

Fixes #836.  The observer consensus-header catch-up loop
(`spawn_stream_consensus_headers`) polled the local `ConsensusCache` on a fixed
100ms interval, gave up after ~60s of no progress, then fell back to waiting for
the next `last_consensus_header` (gossip) update.  On a slow or independently
running node that update may never arrive, so the observer could stay parked
below the consensus target indefinitely.  On timeout it should actively request
the missing range from peers instead of blocking passively.

## Changes

- [x] On the catch-up no-progress timeout, directly request the missing range
  from peers: walk backward from the pending header down to
  `last_consensus_height + 1` (`get_consensus_header_range`), populating the
  `ConsensusCache` that the next catch-up poll consumes.  This does **not** rely
  on the gossip-driven backward traversal (`spawn_track_recent_consensus` ->
  `spawn_fetch_recent_consensus`), which only runs when a gossip update arrives
  and floors its walk at an internal `last_number` watermark, so it cannot
  recover a gap at the bottom of the catch-up range.  The request is bounded by a
  fetch budget (so a peer that never serves the range cannot wedge the loop;  
  partial fetches are cached and persisted, so the next re-drive resumes) and
  guarded by `rx_shutdown`, and the loop keeps polling instead of parking.  This
  threads the existing `PrimaryNetworkHandle` through `spawn_state_sync` ->
  `spawn_stream_consensus_headers` (issue item 2).
- [x] Make the catch-up poll interval and no-progress bound configurable via
  `SyncConfig` (`consensus_header_catch_up_poll_interval`,
  `consensus_header_catch_up_max_no_progress`), defaulting to the previous
  100ms / 600 (60s) behavior. Adds a `sync_config_mut` accessor on `NetworkConfig`.
- [x] Add a `tn_state_sync.catch_up_redrives_total` metric so re-drives are
  observable in production.
- [x] Correct the `catch_up_consensus_from_to` doc comment, which claimed it
  "queries peers";  it only reads local storage (`ConsensusCache`, then
  `ConsensusChain`), leaving the caller to request the missing range.

## Testing

- New unit test `state_sync::tests::redrives_state_sync_request_on_catch_up_timeout`:
  with an empty cache and chain (so catch-up can never progress) and a tiny tuned
  budget, it asserts the stall branch issues an outbound peer request on a mock
  `PrimaryNetworkHandle` within the budget (the loop performs no other network
  I/O, so any command proves the direct re-drive fired).
- Extended the `metrics` test to cover the new counter.
- `cargo clippy -p tn-config -p state-sync -p tn-executor --all-targets --all-features`
  clean; `cargo test -p state-sync` green; `make fmt` clean.

## Notes

- Happy-path behavior is unchanged: when the async backward traversal completes
  within the budget (the common case), the loop makes progress by polling and no
  peer request is issued.  The direct request is the safety net for the no-gossip
  / bottom-of-range-gap case, throttled to at most once per no-progress budget
  and shutdown-aware.
- The remaining #836 checklist item (default-CI coverage via a fast deterministic
  variant of `test_epoch_sync` / `test_epoch_boundary`) is deferred: those tests
  spin up multiple node processes and are `#[ignore]`d for independent runs.  The
  new unit test covers the re-drive decision deterministically without that
  harness.

### DIFF
--- a/crates/state-sync/src/consensus.rs
+++ b/crates/state-sync/src/consensus.rs
@@ -365,7 +365,7 @@ pub async fn spawn_fetch_consensus(
 
 /// Retrieve a consensus headers from a peer.
 /// Start at number/hash and work backwards to end number.
-async fn get_consensus_header_range<DB: TNDatabase>(
+pub(crate) async fn get_consensus_header_range<DB: TNDatabase>(
     number: u64,
     hash: ConsensusHeaderDigest,
     db: &DB,
--- a/crates/state-sync/src/consensus.rs
+++ b/crates/state-sync/src/consensus.rs
@@ -2,7 +2,7 @@
 
 use std::{
     sync::{
-        atomic::{AtomicI32, Ordering},
+        atomic::{AtomicBool, AtomicI32, Ordering},
         Arc,
     },
     time::Duration,
@@ -29,6 +29,31 @@ enum ConsensusHeaderResult {
     Done,
     Continue(u64, ConsensusHeaderDigest),
     Retry,
+}
+
+/// RAII accounting for a single in-flight gap-fill walk.
+///
+/// Constructing one marks the gap walk active (single-flight) and adds it to the shared `tasks`
+/// throttle so the gossip-driven backfill sees the load. `Drop` releases both, so the slot is
+/// freed even if the walk task panics or is cancelled — otherwise a lost walk would wedge gap
+/// recovery permanently.
+struct GapWalkGuard {
+    active: Arc<AtomicBool>,
+    tasks: Arc<AtomicI32>,
+}
+
+impl GapWalkGuard {
+    fn new(active: Arc<AtomicBool>, tasks: Arc<AtomicI32>) -> Self {
+        tasks.fetch_add(1, Ordering::Relaxed);
+        Self { active, tasks }
+    }
+}
+
+impl Drop for GapWalkGuard {
+    fn drop(&mut self) {
+        self.tasks.fetch_sub(1, Ordering::Relaxed);
+        self.active.store(false, Ordering::Relaxed);
+    }
 }
 
 /// Retrieve a verified consensus OUTPUT (header + batches) for `number` and cache it.
@@ -530,6 +555,10 @@ pub async fn spawn_fetch_recent_consensus<DB: TNDatabase>(
     let mut first_gossipped_epoch = None; // Track the first epoch we see via gossip.
     let mut last_number = None;
     let tasks = Arc::new(AtomicI32::new(0));
+    // At most one gap-fill walk runs at a time (single-flight); further requests coalesce onto the
+    // watch until it clears. Independent of the gossip watermark, so it can reach a gap below it.
+    let gap_walk_active = Arc::new(AtomicBool::new(false));
+    let mut rx_consensus_gap = consensus_bus.consensus_gap_request().subscribe();
     // This loop will track current consensus as well as try to backfill from current.
     // This task backfills the current epoch records as well as requesting entire pack files
     // be downloaded for missing historic epochs.
@@ -552,6 +581,46 @@ pub async fn spawn_fetch_recent_consensus<DB: TNDatabase>(
                     &mut last_number,
                     &mut current_fetch_epoch,
                 ).await;
+            }
+
+            // The observer catch-up loop stalled on a bottom gap the gossip-driven walk cannot
+            // reach (its floor only ever rises). Fill exactly that range here, in the task that
+            // owns the fetch accounting, walking back from the verified tip down to `floor`.
+            gap = rx_consensus_gap.changed() => {
+                if gap.is_err() {
+                    // Bus dropped: the node is shutting down and this critical task ends with it.
+                    return;
+                }
+                let request = *rx_consensus_gap.borrow_and_update();
+                if let Some((epoch, number, hash, floor)) = request {
+                    // Single-flight: skip if a gap walk is already in flight (it will re-signal).
+                    if number >= floor && !gap_walk_active.swap(true, Ordering::Relaxed) {
+                        let guard = GapWalkGuard::new(gap_walk_active.clone(), tasks.clone());
+                        let db = db.clone();
+                        let consensus_bus = consensus_bus.clone();
+                        let network = network.clone();
+                        let consensus_chain = consensus_chain.clone();
+                        task_spawner.spawn_task(
+                            format!("gap-fill epoch {epoch} consensus {floor}..={number}"),
+                            async move {
+                                // Held for the walk's lifetime; Drop frees the single-flight slot
+                                // and the task count even on panic/cancellation.
+                                let _guard = guard;
+                                get_consensus_header_range(
+                                    number,
+                                    hash,
+                                    &db,
+                                    &consensus_bus,
+                                    &network,
+                                    &consensus_chain,
+                                    floor,
+                                )
+                                .await;
+                                Ok(())
+                            },
+                        );
+                    }
+                }
             }
 
             _ = &rx_shutdown => {
--- a/crates/state-sync/src/consensus.rs
+++ b/crates/state-sync/src/consensus.rs
@@ -27,6 +27,11 @@ enum ConsensusHeaderResult {
 }
 
 /// A single consensus-height range `[floor..=ceil]` currently being walked by a fetch task.
+///
+/// Ranges carry no epoch tag even though the tracker is app-scoped (a walk from one epoch can
+/// linger into the next). This is sound ONLY because consensus numbers are globally monotonic
+/// across epochs, so a `(floor, ceil)` pair names a unique height range regardless of epoch. If
+/// numbering ever became per-epoch, two epochs' ranges could alias and a walk could wrongly defer.
 struct WalkRange {
     id: u64,
     floor: u64,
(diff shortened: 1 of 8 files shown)


sstanfield (changes requested):
I think there is some useful stuff here but calling get_consensus_header_range() from spawn_stream_consensus_headers() is not correct IMO.  That is correctly used in spawn_fetch_recent_consensus() where more accounting for in flight requests is done.  Also that call could take a while and is intended to run on its own task. spawn_stream_consensus_headers() should just process consensus in order once the gaps are filled by other tasks.  If waiting for the next consensus gossip is causing issues then adding a simple timeout to that bottom select would probably help (?).  But stepping on other tasks functions I don't think is the proper response, we can discuss tomorrow in scrum though- maybe I am missing something. If you have found a bug where gaps are never filled then the fix should live in spawn_fetch_recent_consensus() most likely (that is it's job).

MavenRain:
> I think there is some useful stuff here but calling get_consensus_header_range() from spawn_stream_consensus_headers() is not correct IMO. That is correctly used in spawn_fetch_recent_consensus() where more accounting for in flight requests is done. Also that call could take a while and is intended to run on its own task. spawn_stream_consensus_headers() should just process consensus in order once the gaps are filled by other tasks. If waiting for the next consensus gossip is causing issues then adding a simple timeout to that bottom select would probably help (?). But stepping on other tasks functions I don't think is the proper response, we can discuss tomorrow in scrum though- maybe I am missing something. If you have found a bug where gaps are never filled then the fix should live in spawn_fetch_recent_consensus() most likely (that is it's job).

I think you're right, and digging into why has actually sharpened the fix.  Calling `get_consensus_header_range()` from `spawn_stream_consensus_headers()` reaches across the task boundary the design draws, and the real defect isn't even where my PR put the patch.

The actual mechanism behind #836: when a backward walk eventually caches the missing outputs, `get_consensus_output` calls `send_last_consensus_header_if_newer`, which only fires the `last_consensus_header` watch when `header.number > current` (consensus_bus.rs:416).  The backfilled heights are all `<= pending_header.number`, so the send is a no-op and the stream task's parked `rx_last_consensus_header.changed()` select never wakes even though the data is now local.  So the observer parks not because gossip stops, but because a silent cache-fill never signals the consumer.  That's exactly why your "simple timeout on the bottom select" instinct is the right first move: on wake it re-enters `catch_up_consensus_from_to`, the cache now hits, and it drains to the target.

The one gap a bare timeout leaves is the range a gossip-triggered walk can't reach: `manage_new_consensus` advances the walk floor to `number+1` at spawn time (consensus.rs:478), monotone and never re-lowered, so a hole below the watermark (wedged walk, or a fault that clears later) stays unfetched.  That fill genuinely is `spawn_fetch_recent_consensus`'s job, as you said.  So I'd like to rework it as two halves along the same seam you drew:

- CONSUME (stream task): add the timeout arm you suggested to the bottom select, sized off the existing catch-up knobs.
- FILL (fetch task): on a stall the stream task publishes its executed height as data on a new `tx_consensus_gap` bus watch;  `spawn_fetch_recent_consensus` gains a select arm on it and issues a single-flight, budget-bounded `get_consensus_header_range` walk floored at executed+1, using the accounting it already owns (the `tasks` counter) and without touching `last_number`.

That keeps the ordered-processing loop pure, puts the fetch in the task that owns the in-flight bookkeeping, and still recovers a bottom gap that no gossip walk will.  I'll keep the config knobs, the re-drive metric, and the doc-comment fix from this PR, and drop the `network` handle I threaded into the stream task.

crates/state-sync/src/consensus.rs
```diff
@@ -365,7 +390,7 @@ pub async fn spawn_fetch_consensus(
 
 /// Retrieve a consensus headers from a peer.
 /// Start at number/hash and work backwards to end number.
-async fn get_consensus_header_range<DB: TNDatabase>(
+pub(crate) async fn get_consensus_header_range<DB: TNDatabase>(
```
sstanfield: Don't need pub(crate) anymore?

crates/state-sync/src/consensus.rs
```diff
@@ -554,6 +583,46 @@ pub async fn spawn_fetch_recent_consensus<DB: TNDatabase>(
                 ).await;
             }
 
+            // The observer catch-up loop stalled on a bottom gap the gossip-driven walk cannot
+            // reach (its floor only ever rises). Fill exactly that range here, in the task that
+            // owns the fetch accounting, walking back from the verified tip down to `floor`.
+            gap = rx_consensus_gap.changed() => {
```
sstanfield, line 684: This is a big improvement but I am almost certain this will still conflict with/step on the tasks started in manage_new_consenus().  A properly working node can take a long time to manually backfill the current epoch (depending on how deep it is) and this gap code will likely trigger while the original large backfill is still in flight.  I think it will work but would lead to tasks racing to download consensus.  You might be able to make GapWalkGuard smarter and use it for all the output download tasks to avoid any overlap.  For instance the last_number is already tracked to avoid tasks walking into other tasks.  If a gap walk manager that tracked the bookkeeping (with drop updates from GapWalkGuard) were added it could probably clean all this up- manage the tasks allowed in flight, handle tasks crashing before done, etc.  Anyway, I don't think this is quite right yet (but will probably work and pass tests) but I think is on the brink of cleaning up this entire gap filling code quite a bit.  Just need all the ways to walk unified and understand some walk tasks will look like stalls to the consensus execute loops.

sstanfield (changes requested):
Still have a comment we need to discuss or address.  I think this code is unlikely to be needed but if it is it could save a node restart for sure (so worth having).  Specifically most of the issues would require a stalled network (no more gossip) to trigger but still worth covering.  Except if a walk task had an issue- that would be a major problem and this work will close that.  But I don't think it is quite there however it is really close to making the gap filling a lot cleaner and smoother with a little more work (I think).

grantkee (approved):
LGTM

sstanfield (approved):
Looks great, thanks.

PR #884, opened 2026-07-09
feat(network): cut the primary partial epoch-pack path over to the typed sync protocol with legacy fallback (739)

MavenRain:
## Summary

Cuts the primary **partial epoch-pack** stream path (`StreamEpochPartial`) over to the typed `/tn-primary-sync` protocol, with a legacy fallback for peers that do not yet serve the new sync variant.  This is the direct partial-prefix sibling of the full-pack cutover (#797) and its fallback removal (#810), and it removes one of the two remaining users of the raw `/tn-stream` behaviour (the other, single-consensus-output streaming, is a follow-up).  It is part of finishing the #739 migration before the final legacy freeze.

Additive and wire-compatible: no request/response or `PrimarySyncRequest` variant is reordered or removed; the legacy `StreamEpochPartial` request-response + `/tn-stream` responder stay in place as the fallback and are removed in the coordinated `/0.0.2` bump (item 9).

## What changes

**New typed request variant.** `PrimarySyncRequest::EpochPackPartial { epoch, last_consensus_number }` is appended (BCS discriminant 2); `request_tags_are_stable` pins it and a round-trip test covers it.  It folds the legacy `StreamEpochPartial` ack-plus-digest handshake into the opening `SyncFrame::Req`, exactly as `EpochPack` did for the full pack.

**Requester (dual-stack, sync-first).** `request_partial_epoch_pack` now tries the sync protocol first over the same per-peer probe as the full pack (`request_epoch_pack_sync`, reused with a `last_consensus_number: Option<u64>` selector), then falls back to the existing legacy `/tn-stream` loop when no connected peer serves the partial variant over sync.  The sync probe is penalty-exempt.  On `Ack` the reassembled `Data`/`End` stream is imported via `ConsensusChain::import_partial_to_staging` (a side staging dir), preserving the legacy split that keeps the partial prefix from racing the in-order build of the current epoch's main pack.  Bad-pack import errors on the sync path are metrics-only (no penalty) during rollout, matching the full-pack sync path.

**Capability-cache isolation.** The shared `sync_capability` probe-dedup cache is not poisoned by a partial `Unsupported`: only a *full*-pack `Unsupported` authoritatively proves a peer does not speak `/tn-primary-sync`.  A partial `Unsupported` is ambiguous: the peer may be on an earlier item that serves full packs but cannot decode the newer `EpochPackPartial` variant, so it must not cache `false` and hide that peer from full-pack sync.  Partial still honors a `false` a full probe wrote and caches `true` on success, so it dedups without cross-contamination.

**Responder.** `process_inbound_sync_stream` gains an `EpochPackPartial` arm; `send_sync_epoch_pack_over_stream` is generalized with `stop_number: Option<u64>` and bounds the source reader to the verifiable `[0, output_end(k))` prefix (`get_partial_epoch_stream` + `AsyncReadExt::take`), reusing the same `Ack` + chunked `Data` + `End` framing, admission caps, and bounded writes as the full pack.

## Conventions

No two-branch pattern-match on `Option`: the request variant is selected with `Option::map_or`, the source reader/cap with `Option::map_or_else` over boxed futures, and the import target with `last_consensus_number.is_some()` + a boxed `if` (never a `match`/`if let` on the `Option`).

## Tests

- `primary_partial_epoch_pack_request_round_trips` + extended `request_tags_are_stable` (pins discriminant 2).
- `test_sync_partial_epoch_pack_over_stream`: the sync serve writes `Ack` then streams exactly the verifiable prefix bytes; the reassembled `Data` frames equal the data file truncated at `output_end(k)`, and a later cutoff streams strictly more with the smaller prefix a true prefix of it (the sync mirror of `test_send_partial_epoch_over_stream`).

The requester-side import-target choice (`import_partial_to_staging` for partial vs `stream_import` for full) is not unit-tested because the exchange runs over a real `libp2p::Stream`, which is unmockable, so the epoch-pack requester has no trait seam (same limitation as the already-merged full-pack requester;  the streamed happy path is covered by the ignored observer-pack-import e2e test).  The choice mirrors the tested legacy `StreamEpochPartial` requester, which also stages, and the in-code comment documents why partial must not import in place.

## Gates

`-j 2` throughout (OOM-safe).  fmt (nightly-2026-03-20);  clippy `-p tn-network-libp2p --all-features --all-targets` and `-p tn-primary` clean;  `tn-network-libp2p` + `tn-primary` tests pass;  `cargo doc -D warnings`.

## Rollout / follow-ups

Dual-stack per the #739 rollout rules (`Severe` suppressed for the new-protocol negotiation failure while the fallback exists).  Remaining before the epic closes:
- Single consensus-output stream (`StreamConsensusOutput`) cutover: the last legacy `/tn-stream` user.
- Then the legacy freeze: make the worker-batch and these paths sync-only (mirroring #810), answer the deprecated request-response variants with a typed error, and delete the raw `/tn-stream` open path, digest pending maps, and prune tasks.
- Finally the coordinated `/0.0.2` bump (telemetry + epoch-boundary gated): delete the deprecated BCS variants and drop the `From<PeerExchangeMap>` / `peer_exchange_msg()` bounds from `TNMessage`.

### DIFF
--- a/crates/consensus/primary/src/network/mod.rs
+++ b/crates/consensus/primary/src/network/mod.rs
@@ -12,7 +12,9 @@ use std::{
 use crate::{
     proposer::OurDigestMessage, state_sync::StateSynchronizer, ConsensusBus, ConsensusBusApp,
 };
-use futures::{AsyncReadExt as _, AsyncWriteExt as _, StreamExt as _};
+use futures::{
+    AsyncReadExt as _, AsyncWriteExt as _, FutureExt as _, StreamExt as _, TryFutureExt as _,
+};
 use handler::RequestHandler;
 pub use message::{MissingCertificatesRequest, PrimaryRequest, PrimaryResponse};
 use message::{PrimaryGossip, PrimaryRPCError};
@@ -690,7 +692,9 @@ impl PrimaryNetworkHandle {
     /// Full-pack fetch is typed-only (no legacy fallback): this returns `Err`, and
     /// the caller retries, when no connected peer successfully serves a pack over
     /// the sync protocol (no peers, all unsupported, or all failed to import). See
-    /// [`Self::request_epoch_pack_sync`] for the per-peer probe behavior.
+    /// [`Self::request_epoch_pack_sync`] for the per-peer probe behavior. The partial
+    /// prefix ([`Self::request_partial_epoch_pack`]) reuses the same probe over the
+    /// sync protocol but keeps a legacy fallback.
     pub async fn request_epoch_pack(
         &self,
         epoch_record: &EpochRecord,
@@ -736,8 +740,9 @@ impl PrimaryNetworkHandle {
     /// Shared implementation for full ([`Self::request_epoch_pack`]) and partial
     /// ([`Self::request_partial_epoch_pack`]) epoch pack streaming. A full pack
     /// (`last_consensus_number` is `None`) is fetched over the typed sync protocol
-    /// only; a partial prefix (`Some`) still negotiates over the legacy stream path,
-    /// which has no typed sync request variant yet.
+    /// only; a partial prefix (`Some`) is fetched over the sync protocol first and
+    /// falls back to the legacy stream path for peers that do not yet serve the
+    /// `EpochPackPartial` sync variant.
     async fn request_epoch_pack_inner(
         &self,
         epoch_record: &EpochRecord,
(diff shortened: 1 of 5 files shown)


crates/consensus/primary/src/network/mod.rs
```diff
@@ -236,6 +238,27 @@ enum EpochPackAttempt {
...
+/// exchange). Bundles what the exchange needs so each probe hop takes the peer
+/// plus this instead of six loose parameters.
+#[derive(Clone, Copy)]
+struct EpochPackSyncRequest<'a> {
+    /// The epoch whose pack is requested.
+    epoch: Epoch,
+    /// `None` requests the full pack; `Some(n)` requests a verifiable prefix
+    /// stopping after consensus number `n`.
+    last_consensus_number: Option<u64>,
+    /// Record of the epoch being fetched; the streamed pack verifies against it.
```
sstanfield, line 252: IIF you make any changes before merge it would be worth mentioning in this comment that epoch_record can (and will) be synthetic for the current epoch partial pack since it does not exist yet.

sstanfield (approved):
Looks good.

grantkee (approved):
LGTM

PR #886, opened 2026-07-09
perf(primary): resolve remaining #822 review findings (missing-certs sync)

MavenRain:
## Summary

Follow-up to the #822 review, addressing the remaining findings that PR #826 deliberately deferred.  PR #826 (merged) fixed the High out-of-bounds panic (finding 1), the `get_bounds` overflow (finding 4), and added the regression test (finding 6).  This PR closes out the three remaining findings — **2 (memory envelope)**, **3 (double BCS encode)**, and **5 (vestigial `max_response_size`)** — each kept minimal and wire-safe.

Fixes are scoped to be provably behavior-preserving where the finding is efficiency/cleanup, and to respect the deliberately-retained legacy wire shape.

## Finding 2 (Medium) — document the requester-side aggregate memory envelope

Each per-peer fetch accepts up to `MAX_SYNC_MISSING_CERTS_ACCEPT_BYTES` (~64.5 MiB), and `fetch_from_peers` staggers fetches while cancelling only on the first success, so several exchanges can buffer concurrently with no joint byte budget.  The reviewer flagged this as a design decision to confirm with the author.

This PR takes the documentation route (the reviewer's option 1): the worst-case envelope — `min(peers, ceil(per_fetch_timeout / fallback_delay))` concurrent exchanges × the per-fetch cap — is now spelled out on both `fetch_from_peers` and the accept-cap doc-comment, so the per-exchange cap can no longer be mistaken for the process-wide bound.

I did **not** change the fan-out behavior.  A concurrent-fetch cap (option 2) was rejected because it changes observable fan-out that two existing behavior tests assert — it deadlocks `test_timeout_scenario` (which drops every reply, so a live-task cap never releases a slot and stalls below `num_peers`) and alters `test_network_failure_keeps_trying`.  A shared byte budget across in-flight fetches (option 3) is the only bound that caps peak memory independent of committee size without breaking the all-peers-requested test premise, but it is a real design change (threading a shared budget into every `spawn_peer_fetch` and enforcing it in the streaming reader) — left to the maintainers' call.  I'd be happy to follow up with option 3 if wanted.

## Finding 3 (Low) — drop the throwaway per-certificate encode on the responder hot path

`send_sync_certificates_over_stream` computed each certificate's byte size with `encode(&cert).len()`, allocating and discarding a full `Vec<u8>` per certificate purely to read its length;  the batch is then encoded again whole in `write_cert_batch_frame`.

- Added `tn_types::encoded_size`, which returns the BCS-encoded byte count via `bcs::serialized_size` **without** allocating the encoding — the same value `encode(obj).len()` produces.
- Replaced `encode(&cert).len()` with `encoded_size(&cert)`, threading its (unreachable, for locally-read certs) error through the existing `PrimaryNetworkResult` rather than panicking.

The per-certificate `size` still drives the `MAX_SYNC_MISSING_CERTS_RESPONSE_BYTES` cap accounting (`scan`) and the `SYNC_CERT_BATCH_TARGET_SIZE` flush threshold (`CertBatch::push`) with byte-for-byte identical values, so frame boundaries and the streamed bytes are unchanged.  The existing `sync_codec` round-trip test passing unchanged proves the equivalence.

## Finding 5 (Informational) — clarify the vestigial `max_response_size` dummy

`max_response_size` is unused on the sync fetch path (the reply is streamed and ended explicitly).  `CertificateCollector::new` rebuilds a `MissingCertificatesRequest` locally with a dummy `max_response_size: 0` solely to reuse `get_bounds()`.

The field is **not** removed: `MissingCertificatesRequest` is embedded whole in `PrimaryRequest::MissingCertificates`, the `/0.0.1` request-response variant that is deliberately retained for wire compatibility until the coordinated `/0.0.2` bump (item 9 of #739).  Removing the field would change that variant's BCS encoding and desynchronize decoding against `/0.0.1` peers.

Instead, the bare literal `0` is replaced with a named constant and a comment explaining it is a wire-compat dummy the sync path ignores — no wire-byte or behavior change.

## Testing

- `cargo clippy -p tn-types -p tn-primary --all-targets` — 0 errors.
- `cargo test -p tn-primary --lib sync_codec certificate_fetcher` — 18 passed, 0 failed.  The
  `sync_codec` round-trip tests derive their frame boundaries from `encode(&cert).len()`, so
  passing unchanged proves finding 3's `encoded_size` is byte-for-byte equivalent.
- `cargo +nightly-2026-03-20 fmt` — clean.

No wire-format change, no behavior change: `encoded_size` produces the same byte count as
`encode(..).len()`, and findings 2 and 5 are documentation / a named-constant rename.

### DIFF
--- a/crates/consensus/primary/src/certificate_fetcher.rs
+++ b/crates/consensus/primary/src/certificate_fetcher.rs
@@ -443,6 +443,15 @@ async fn fetch_and_process_certificates<DB: Database, F: MissingCertFetcher>(
 /// Try to fetch certificates from multiple peers with staggered requests.
 /// Sends requests to peers one at a time with ~5 second intervals between each.
 /// Returns as soon as one peer provides a non-empty response.
+///
+/// Memory envelope: fetches are cancelled only on the first success, so slow, empty, or
+/// invalid responders keep buffering until their per-fetch timeout elapses. Each in-flight
+/// fetch accepts up to the per-exchange cap (`MAX_SYNC_MISSING_CERTS_ACCEPT_BYTES`, ~64.5 MiB),
+/// and up to `min(peers.len(), ceil(per_fetch_timeout / fallback_delay))` run concurrently, so
+/// the requester-side peak is that many times the per-fetch cap (a few hundred MiB with the
+/// defaults, scaling with committee size up to the overlap bound). There is no joint byte
+/// budget shared across in-flight fetches: capping the concurrent fan-out or threading a shared
+/// budget is left as a design decision (issue #822, finding 2).
 async fn fetch_from_peers<F: MissingCertFetcher>(
     network: F,
     peers: Vec<BlsPublicKey>,
--- a/crates/consensus/primary/src/network/sync_codec.rs
+++ b/crates/consensus/primary/src/network/sync_codec.rs
@@ -23,7 +23,7 @@ use futures::{
 use std::time::Duration;
 use tn_network_libp2p::{read_frame, write_frame, DenyReason, PrimarySyncRequest, SyncFrame};
 use tn_storage::consensus::ConsensusChain;
-use tn_types::{encode, try_decode, BlsPublicKey, Certificate, Epoch};
+use tn_types::{encode, encoded_size, try_decode, BlsPublicKey, Certificate, Epoch};
 use tokio::io::AsyncRead as TokioAsyncRead;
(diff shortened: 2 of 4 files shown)


sstanfield (dismissed):
LGTM!

grantkee (dismissed):

grantkee (approved):

PR #888, opened 2026-07-09
fix: reject round-0 headers on the vote path (header.round() - 1 underflow)

### LINKED ISSUE #789: [Security][Resiliency] Subtraction underflow on attacker-controlled header round (`header.round() - 1`) in `check_for_missing_parents`
**Severity:** Low / Resiliency (correctness wart on the release binary; not a crash). Disposition pending the assumed Byzantine count *f*, since the attacker must be a committee member.
**Attacker model:** A Byzantine committee member (the Vote path is committee-gated; a member crafts its own round-0 header).

## Summary
`check_for_missing_parents` computes `header.round() - 1` with no `saturating_sub` and no prior guarantee that `round >= 1`. A committee member can craft a self-consistent round-0 header that reaches this subtraction. On the mainnet (release) binary, overflow-checks are off, so `0u32 - 1` wraps to `u32::MAX` rather than panicking, leaving a stale `(u32::MAX, digest)` requested-parent entry: a correctness/hygiene wart, not an exploitable crash.

## Mechanism
1. **Reachable on the Vote path.** `vote()` is committee-gated (`crates/consensus/primary/src/network/handler.rs:347-355`: the header author must be a committee authority and the peer-id must match), so a malicious committee member crafts its *own* round-0 header (no forgery needed). With an empty `parents` vec, `vote_inner` routes to `check_for_missing_parents` (`handler.rs:515-517`).
2. **No guard establishes `round >= 1`.** `Header::validate` (`types/.../header.rs:100-132`) does not reject round 0. The age guard `ensure!(limit <= header.round())` (`handler.rs:741`) passes when `limit == 0`; `limit = primary_round().saturating_sub(max_proposed_header_age_limit)` (default 3) is 0 at startup / early epoch (`primary_round` initializes to 0). `identify_unkown_parents` special-cases only `round == 1`; for round 0 it falls through to a DB lookup and returns the bogus digests as unknown.
3. **The subtraction.** `handler.rs:765-773`: `unknown_certs.retain(|digest| { let key = (header.round() - 1, *digest); ... })` evaluates `0u32 - 1` (`Round = u32`).

## Behavior by build profile
- **Release (mainnet):** overflow-checks default off — confirmed, the repo has no `[profile.release]` override. `0u32 - 1` wraps to `u32::MAX`. The `(u32::MAX, digest)` key is inserted, the bogus digest is retained, and the peer receives a benign `MissingParents` response. The stale entry is not reached by the GC loop at `handler.rs:756-762`. No crash, OOM, hang, ban, or consensus-safety impact (a round-0 header cannot satisfy the parent-quorum checks, so it never produces a certificate).
- **Debug/CI (overflow-checks on):** panics, but contained — the vote request runs in a non-critical `spawn_task` (`mod.rs`), and the task manager ignores (`continue`) a non-critical task that errors/panics, so no node shutdown. CI is not the attack target.

## Why existing guards do not prevent it
- `validate()` and the `TooOld`/`TooNew` round checks do not establish `round >= 1` before the subtraction; the only relevant guard (`limit <= header.round()`) is satisfiable at round 0 when `limit == 0`.
- Unlike the `limit` computation two lines above, the subtraction here does not use `saturating_sub`.

## Impact
On the mainnet binary: a silent-wrap correctness wart (a stale `(u32::MAX, digest)` requested-parent entry that GC does not reach), self-inflicted by a malicious committee member with no leverage over other nodes. Not a crash/hang/OOM/ban. Worth fixing for correctness and defense-in-depth.

## Suggested fix
- Use `header.round().saturating_sub(1)` (matching the saturating style already used for `limit`), or reject `round == 0` / `header.round() < 1` early in `vote_inner` before parent tracking.
- Consider enabling `overflow-checks = true` for the release profile so silent wraps cannot mask logic errors elsewhere.

## Acceptance criteria
- A round-0 header on the Vote path does not produce a `u32::MAX` key or a stale requested-parent entry.
- Regression test: a committee-authored round-0 header with empty parents is handled without underflow and without leaving stale state.

## References
`primary/network/handler.rs:741`, `:756-773`, `:347-355`, `:515-517`; `types/.../header.rs:100-132`.

MavenRain:
Closes #789.

## Problem

`check_for_missing_parents` (`crates/consensus/primary/src/network/handler.rs`) computes
`header.round() - 1` with no `saturating_sub` and no prior guarantee that `round >= 1`:

```rust
let key = (header.round() - 1, *digest);
```

The Vote path is committee-gated (`vote()` verifies the header author is a committee
authority and the peer id matches), so a **Byzantine committee member** can craft its own
self-consistent **round-0** header . . . no forgery needed.  With an empty `parents` vec,
`vote_inner` routes to `check_for_missing_parents`.  `Header::validate` does not reject round 0,
and the age guard `ensure!(limit <= header.round())` passes when `limit == 0` (true at
startup / early epoch), so the subtraction is reached at `round == 0`.

- **Release (mainnet):** `overflow-checks` default off — `0u32 - 1` wraps to `u32::MAX`,
  inserting a stale `(u32::MAX, digest)` requested-parent entry that the GC loop
  (`round < limit.saturating_sub(1)`) never reaches.  No crash/OOM/ban/consensus-safety
  impact (a round-0 header can never satisfy the parent-quorum checks), but a silent
  correctness/hygiene wart.
- **Debug/CI:** panics, but contained — the vote request runs in a non-critical task the
  task manager ignores.

## Changes

- [x] Add `HeaderError::InvalidRound(HeaderDigest)` (`crates/types/src/error.rs`).
- [x] Reject round-0 headers early in `vote_inner`, right after `Header::validate`, before
      parent tracking.  Proposed headers are always `round >= 1` (round 0 is reserved for
      genesis certificates and is never voted on), so this makes the implicit invariant
      explicit.
- [x] Classify `InvalidRound` as `Penalty::Fatal` in `penalty_from_header_error`, alongside
      the other structurally-malformed-header faults (`InvalidGenesisParent`, `UnkownWorkerId`,
      ...).  The exhaustive match forces this to be revisited on any future variant change.
- [x] Change the subtraction to `header.round().saturating_sub(1)` as defense-in-depth so the
      helper is underflow-safe independent of its caller, matching the `saturating_sub` style
      already used for `limit` two lines above.
- [x] Add `test_request_vote_round_zero_rejected` (`crates/consensus/primary/src/tests/primary_tests.rs`):
      a committee-authored round-0 header with empty parents is rejected with `InvalidRound`,
      so no underflow occurs and no stale requested-parent state is left behind.

## Acceptance criteria (from the issue)

- [x] A round-0 header on the Vote path does not produce a `u32::MAX` key or a stale
      requested-parent entry (the header is rejected before parent tracking runs).
- [x] Regression test for a committee-authored round-0 header with empty parents.

## Not included (deferred)

The issue also suggested enabling `overflow-checks = true` for the release profile so silent
wraps cannot mask logic errors elsewhere.  That is a repo-wide change with performance
implications and its own review surface, so it is intentionally left out of this targeted fix.

### DIFF
--- a/crates/consensus/primary/src/error/network.rs
+++ b/crates/consensus/primary/src/error/network.rs
@@ -204,6 +204,7 @@ fn penalty_from_header_error(error: &HeaderError) -> Option<Penalty> {
         | HeaderError::UnknownNetworkKey(_)
         | HeaderError::PeerNotAuthor
         | HeaderError::InvalidGenesisParent(_)
+        | HeaderError::InvalidRound(_)
         | HeaderError::ParentMissingSignature
         | HeaderError::InvalidParentTimestamp { .. }
         | HeaderError::UnkownWorkerId
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -529,6 +529,12 @@ where
 
         // validate header
         header.validate(committee)?;
+
+        // A proposed header is always at round >= 1; round 0 is reserved for genesis certificates
+        // and is never voted on. Reject a round-0 header here, before parent tracking, so
+        // `check_for_missing_parents` never evaluates `header.round() - 1` at round 0 (see #789).
+        ensure!(header.round() != 0, HeaderError::InvalidRound(header.digest()).into());
+
         let max_round =
             self.consensus_bus.committed_round() + self.consensus_config.parameters().gc_depth;
         // Make sure the header is not unreasonable in the future.
@@ -822,7 +828,9 @@ where
 
         // filter out parents that were already requested and new ones
         unknown_certs.retain(|digest| {
-            let key = (header.round() - 1, *digest);
+            // `saturating_sub` matches the style used for `limit` above and keeps this helper
+            // underflow-safe on its own; `vote_inner` already rejects round 0 (see #789).
+            let key = (header.round().saturating_sub(1), *digest);
             if let std::collections::btree_map::Entry::Vacant(e) = current_requests.entry(key) {
                 e.insert(header.author().clone());
                 true
--- a/crates/consensus/primary/src/tests/primary_tests.rs
+++ b/crates/consensus/primary/src/tests/primary_tests.rs
@@ -92,6 +92,79 @@ async fn test_request_vote_too_new() {
     );
 }
 
+/// Regression for #789: a committee member can craft its own round-0 header (the Vote path is
+/// committee-gated). Before the fix, `check_for_missing_parents` evaluated `header.round() - 1`
+/// at round 0, wrapping to `u32::MAX` on the release binary and leaving a stale
+/// `(u32::MAX, digest)` requested-parent entry. The header is now rejected before parent tracking,
+/// so no underflow occurs and no stale state is left behind.
+#[tokio::test(flavor = "current_thread", start_paused = true)]
+async fn test_request_vote_round_zero_rejected() {
+    const NUM_PARENTS: usize = 10;
+    let fixture = CommitteeFixture::builder(MemDatabase::default)
+        .randomize_ports(true)
+        .committee_size(NonZeroUsize::new(NUM_PARENTS).unwrap())
+        .build();
+    let target = fixture.authorities().next().unwrap();
+    let author = fixture.authorities().nth(2).unwrap();
+    let author_id = author.id();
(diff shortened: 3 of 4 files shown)


sstanfield (approved):
LGTM!

grantkee (approved):
Good catch

PR #889, opened 2026-07-09
fix: charge signed gossip content faults to the author, not the relayer

### LINKED ISSUE #871: Primary gossip path charges the relayer for content-determined faults in signed payloads
## Context

PR #819 (branch `819-gossip-reject-author-attribution`) continues the work of #801 / #785 to move gossip fault attribution from the relaying peer to an accountable party. During review of PR #819 we confirmed a follow-up gap: for the primary's signed gossip payloads, several content-determined faults are still charged to the relayer rather than the party responsible for the bad content.  This issue records that finding so it can be tracked separately from the merged PR.

## Problem

`is_author_content_fault` returns `true` for only two variants, `Decode` and `InvalidTopic`, so exactly those two faults are attributed to the message author. Every other content-determined fault the primary gossip handler can raise is classified `false` and therefore charged to the relayer.  Because deep validation of signed payloads runs after gossipsub has already accepted and forwarded the message, an honest relayer cannot screen these faults before relaying them. That is the same #801 / #819 failure mode the PR set out to fix, still open for the primary's signed certificate, vote, and consensus payloads.

## Details / Evidence

The classifier is an exhaustive match over `PrimaryNetworkError`:

- `crates/consensus/primary/src/error/network.rs:91` defines `pub(crate) fn is_author_content_fault(&self) -> bool`.
- `crates/consensus/primary/src/error/network.rs:96` is the only `true` arm: `PrimaryNetworkError::Decode(_) | PrimaryNetworkError::InvalidTopic => true`.
- `crates/consensus/primary/src/error/network.rs:97-112` are the `false` arms, including `InvalidHeader(_)`, `Certificate(_)`, and `UnknownConsensusHeaderCert(_)`, all of which are charged to the relayer.

The charge site selects author vs relayer from that classifier and dispatches the penalty:

- `crates/consensus/primary/src/network/mod.rs:1560`: `let charged = if e.is_author_content_fault() { author } else { relayer };`
- `crates/consensus/primary/src/network/mod.rs:1561-1562`: `if let Some((peer, penalty)) = charged.zip((&e).into()) { network_handle.report_penalty(peer, penalty).await; }` (the `zip` skips an unresolved peer or a `None` penalty).

The penalty mapping itself is `impl From<&PrimaryNetworkError> for Option<Penalty>` at `crates/consensus/primary/src/error/network.rs:117` (the `mod.rs` charge site only invokes it via `(&e).into()`).

Gossip-reachable content faults that are currently charged to the relayer:

1. Certificate faults, reachable post-Accept on the gossip Certificate path:
   - `crates/consensus/primary/src/network/handler.rs:231`: `cert.validate_received().map_err(CertManagerError::from)?;`
   - `crates/consensus/primary/src/network/handler.rs:244`: `self.state_sync.process_peer_certificate(&mut cert).await?;`
   - Map to `Penalty::Fatal`: `CertificateError::{RecoverBlsAggregateSignatureBytes, Unsigned, Inquorate, InvalidSignature}` at `crates/consensus/primary/src/error/network.rs:134-137`, and `CertManagerError::UnverifiedSignature(_)` at `crates/consensus/primary/src/error/network.rs:144`.
2. `EpochVote` with a bad signature -> `InvalidHeader(HeaderError::PeerNotAuthor)`:
   - `crates/consensus/primary/src/network/handler.rs:369-372`: `ensure!(vote.check_signature(), PrimaryNetworkError::InvalidHeader(HeaderError::PeerNotAuthor));`
   - `Penalty::Fatal` via `HeaderError::PeerNotAuthor` at `crates/consensus/primary/src/error/network.rs:205`.
3. `EpochVote` from a non-committee key -> `InvalidHeader(HeaderError::UnknownAuthority(...))`:
   - `crates/consensus/primary/src/network/handler.rs:377-383`: `ensure!(epoch_rec.committee.contains(&vote.public_key), PrimaryNetworkError::InvalidHeader(HeaderError::UnknownAuthority(...)));`
   - `Penalty::Fatal` via `HeaderError::UnknownAuthority` at `crates/consensus/primary/src/error/network.rs:210`.
4. `ConsensusResult` with a bad signature -> `UnknownConsensusHeaderCert(...)`:
   - `crates/consensus/primary/src/network/handler.rs:316-320`: `ensure!(signature.verify_secure(&to_intent_message(consensus_result_hash), &key), PrimaryNetworkError::UnknownConsensusHeaderCert(hash));`
   - `Penalty::Mild` at `crates/consensus/primary/src/error/network.rs:167` (variant defined at `crates/consensus/primary/src/error/network.rs:40` as `UnknownConsensusHeaderCert(ConsensusHeaderDigest)`).

Note on the current variant: the fault that maps a bad gossiped `ConsensusResult` signature to `Mild` is `UnknownConsensusHeaderCert`. An earlier name (`UnknownConsensusHeaderDigest`) no longer exists as an enum variant in the code; it survives only in a stale `crates/network-libp2p/src/peers/penalty.md` doc table (which still lists it mapping to `None`, itself out of date since the code now maps `UnknownConsensusHeaderCert` to `Mild`).  The citation above reflects the current tree.

The doc comment above the classifier frames the remaining faults as benign or as embedded content whose attribution is left unchanged:

- `crates/consensus/primary/src/error/network.rs:87-90`: "benign on the gossip path or concerns embedded certificate/consensus content whose signer is carried inside the payload ... so its existing relayer/no-penalty attribution is left unchanged."

That wording understates the residual exposure, since "left unchanged" here means honest relayers remain subject to Fatal (or Mild) penalties for content they could not screen.

## Impact

Severity: Medium (Consensus Safety / Bugs).

A Byzantine committee member that broadcasts a malformed certificate or vote can cause honest relayers of that gossip message to be penalized for content they cannot validate before relaying (validation runs after gossipsub accept/forward).  The exposure is bounded by an existing mitigation and is pre-existing:

- Committee-member relayers are spared. `crates/network-libp2p/src/peers/peer.rs:154-174` shows `apply_penalty` bypasses the score model entirely for exempt peers (previous/current/next committee validators, or operator allowlist entries, via `TrustBasis`): when an exemption is present, no penalty of any level is applied to the peer's score.  The `matches!(penalty, Penalty::Severe | Penalty::Fatal)` arm only emits a `warn!` operational signal, it does not gate the suppression, so exempt peers are spared every penalty level, `Mild` included.  Honest validator relayers therefore do not get Fatal-banned, nor do they accrue `Mild`, for these faults.
- Residual real exposure is non-committee relayers (observers / NVVs, which remain future work): they receive the `Fatal` penalty for relaying a malformed certificate or vote (and `Mild` for relaying a bad `ConsensusResult`).
- This is pre-existing, not a regression introduced by PR #819. Before the PR the relayer was charged for all faults;  PR #819 fixes 2 of at least 5 gossip-reachable content faults (`Decode` and `InvalidTopic`).  It is nonetheless in scope for the PR's stated goal of fixing this failure mode.

Author attribution is correctly not applied to these faults today, because the offending signer is embedded in the payload (`vote.public_key`, the certificate aggregate signers, the `ConsensusResult` validator/key) and is generally not `message.source`, so charging `author` would penalize the wrong party.  A correct remedy needs a distinct embedded-signer attribution path.

## Proposed fix

- Add a third attribution path, distinct from author vs relayer, that charges cert / vote / consensus faults to the embedded signer (the certificate aggregate signers, `vote.public_key`, or the `ConsensusResult` validator), rather than to the relayer.
- In the interim, amend the `is_author_content_fault` doc comment (`crates/consensus/primary/src/error/network.rs:87-90`) to explicitly acknowledge the residual relayer exposure (non-committee relayers can still be Fatal- or Mild-penalized; committee members are exempt from the score model) instead of implying the attribution is complete.

## Acceptance criteria

- [ ] A malformed gossiped certificate (`RecoverBlsAggregateSignatureBytes` / `Unsigned` / `Inquorate` / `InvalidSignature` / `UnverifiedSignature`) attributes the penalty to the embedded certificate signer(s), not the relaying peer.
- [ ] A gossiped `EpochVote` failing signature or committee-membership checks (`PeerNotAuthor` / `UnknownAuthority`) attributes the penalty to `vote.public_key`, not the relayer.
- [ ] A gossiped `ConsensusResult` with a bad signature (`UnknownConsensusHeaderCert`) attributes the penalty to the embedded validator/key, not the relayer.
- [ ] Non-committee relayers (observers / NVVs) are no longer Fatal-penalized for relaying content-determined faults they could not screen.
- [ ] If the full embedded-signer path is deferred, the `is_author_content_fault` doc comment at `crates/consensus/primary/src/error/network.rs:87-90` is updated to state the residual relayer exposure explicitly.
- [ ] Test coverage exercises each reachable variant (cert faults, `EpochVote` faults, `ConsensusResult` bad signature) and asserts the charged party.

MavenRain:
Closes #871.

## Problem

On the primary gossip path, `PrimaryNetworkError::is_author_content_fault` attributes a
fault to the message author for only `Decode` and `InvalidTopic`.  Every other
content-determined fault the gossip handler raises is classified `false` and charged to
the **relaying peer**.  Deep validation of a signed payload runs *after* gossipsub has
already accepted and forwarded the message, so an honest relayer cannot screen these
faults before relaying them.  That is the #801/#785/#819 failure mode, still open for the
primary's signed certificate, epoch-vote, and consensus-result payloads.

## Change

- [x] Move `InvalidHeader(_)`, `Certificate(_)`, and `UnknownConsensusHeaderCert(_)` to
      the author arm of `is_author_content_fault`, so these gossip content faults charge
      the authenticated gossip **author** (`GossipMessage::source`) instead of the honest
      relayer.
- [x] Rewrite the classifier doc comment: the author is the accountable **publisher**;
      explain why the embedded payload signer is deliberately *not* charged (framing
      risk, below).
- [x] Update the now-stale sibling test doc comment; add
      `embedded_signed_payload_faults_are_author_content` asserting the charged party for
      each reachable variant (cert `Unsigned`/`InvalidSignature`, `EpochVote`
      `PeerNotAuthor`/`UnknownAuthority`, `ConsensusResult` `UnknownConsensusHeaderCert`).

This is a gossip-path-only classifier;  the request/response and stream paths penalize the
peer through a separate, untouched mapping, so RPC/stream penalties are unchanged.

## Why the author, and not the payload-embedded signer

The issue proposed a third attribution path that charges the embedded signer (cert
aggregate signers / `vote.public_key` / `ConsensusResult.validator`).  Charging the
authenticated author is both simpler and strictly safer:

1. **The author is authenticated.** Gossipsub runs `MessageAuthenticity::Signed` +
   `ValidationMode::Strict`, so an accepted message's `source` is bound to the publisher's
   key and cannot be spoofed.  Honest nodes only gossip content they produced and
   self-signed (`publish_certificate` / `publish_consensus` / `publish_epoch_vote`), and
   gossipsub forwarding preserves the original `source`, so a content fault is the
   publisher's own.  `verify_cert` Fatal failures are logged, not propagated, and
   `process_peer_certificate` runs only after verification succeeds, so an honest author
   cannot trip the Fatal certificate arms.

2. **Charging the embedded signer would open a framing vector.** These faults are exactly
   signature-verification failures, so the key named inside the payload is unauthenticated
   at the point of fault.  An attacker could gossip
   `ConsensusResult { validator: <victim>, .. }` (or an `EpochVote` naming any committee
   key) with a bad signature and have the innocent victim penalized.  `HeaderError::
   UnknownAuthority` also carries only a formatted `String`, so the key is not even cleanly
   recoverable at the charge site.  The authenticated author carries no such risk.

## Acceptance criteria

- [x] Non-committee relayers (observers / NVVs) are no longer Fatal- or Mild-penalized for
      relaying content-determined faults they could not screen.
- [x] A malformed gossiped certificate, an `EpochVote` failing signature/committee checks,
      and a `ConsensusResult` with a bad signature are charged to the authenticated
      publisher rather than the relayer.
- [x] Test coverage exercises each reachable variant and asserts the charged party.

The remaining criteria ("attribute to the embedded certificate signer(s) / `vote.
public_key` / embedded validator") are intentionally satisfied by charging the
authenticated publisher instead of the embedded signer, for the framing reason above.  The
`is_author_content_fault` doc comment now states the residual exposure explicitly rather
than implying attribution is complete.

## Known bounded residual

`Certificate(_)` also covers `CertificateError::TooOld`, a **Mild** penalty judged against
the *receiver's* GC round. An honest node that re-publishes an old-but-valid certificate
(e.g. the startup re-publish of its last certificate) to a caught-up peer is now charged
that Mild penalty as the author, where previously it fell on the honest relayer.  This
replaces one honest-party Mild charge with another; it is self-limiting (a single ~-1.0
nick per distinct certificate, ~5-minute decay half-life, de-duplicated by storage and
gossipsub message-id, far below the ban threshold) and strictly better than charging an
arbitrary forwarder.  Splitting `TooOld` off to a no-penalty gossip outcome would require a
gossip-vs-RPC split in the shared penalty mapping and is left out of scope.

## Testing

- `cargo test -p tn-primary error::network` (classifier unit tests, incl. the new one).
- `cargo +<nightly> fmt` clean; `cargo clippy -p tn-primary` clean.

### DIFF
--- a/crates/consensus/primary/src/error/network.rs
+++ b/crates/consensus/primary/src/error/network.rs
@@ -302,4 +302,59 @@ mod tests {
         assert!(PrimaryNetworkError::UnknownConsensusHeaderCert(ConsensusHeaderDigest::default())
             .is_author_content_fault());
     }
+
+    /// Finding #871 attribution guard. `process_gossip` charges the accountable peer with
+    /// `if fault.is_author_content_fault() { author } else { relayer }.zip((&fault).into())`, so
+    /// for each fault this PR moved to the author arm the charged party is the authenticated
+    /// author and the penalty is the fault's own. Both halves are asserted here: a future edit
+    /// that re-routes a variant to the relayer, or drops its penalty (which `zip` would silently
+    /// swallow, charging no one), fails this test. The mirror of the selection is exercised by the
+    /// closing control case, and the honest relayer's protection is what the change exists for.
+    #[test]
+    fn moved_gossip_faults_charge_the_author_not_the_relayer() {
+        // the penalty the gossip handler charges the *author*, or `None` if the fault is not
+        // author-content (so the relayer would be charged instead) or carries no penalty
+        let author_penalty = |fault: &PrimaryNetworkError| {
+            fault.is_author_content_fault().then(|| Option::<Penalty>::from(fault)).flatten()
+        };
+
+        // certificate: unsigned / bad-signature -> Fatal, charged to the author
+        assert!(matches!(
+            author_penalty(&PrimaryNetworkError::Certificate(CertManagerError::Certificate(
+                CertificateError::Unsigned
+            ))),
+            Some(Penalty::Fatal)
+        ));
+        assert!(matches!(
+            author_penalty(&PrimaryNetworkError::Certificate(CertManagerError::Certificate(
+                CertificateError::InvalidSignature
+            ))),
+            Some(Penalty::Fatal)
+        ));
+        // epoch vote: bad signature (PeerNotAuthor) / non-committee key (UnknownAuthority) -> Fatal
+        assert!(matches!(
+            author_penalty(&PrimaryNetworkError::InvalidHeader(HeaderError::PeerNotAuthor)),
+            Some(Penalty::Fatal)
+        ));
+        assert!(matches!(
+            author_penalty(&PrimaryNetworkError::InvalidHeader(HeaderError::UnknownAuthority(
+                "key not in committee".to_string()
+            ))),
+            Some(Penalty::Fatal)
+        ));
+        // consensus result: bad signature -> Mild
+        assert!(matches!(
+            author_penalty(&PrimaryNetworkError::UnknownConsensusHeaderCert(
+                ConsensusHeaderDigest::default()
+            )),
+            Some(Penalty::Mild)
+        ));
+
+        // control: a non-content fault is not author-charged, so the relayer arm is taken and the
+        // author penalty is `None` even though the fault itself carries one
+        let relayer_fault = PrimaryNetworkError::InvalidEpochRequest;
+        assert!(!relayer_fault.is_author_content_fault());
+        assert!(author_penalty(&relayer_fault).is_none());
+        assert!(matches!(Option::<Penalty>::from(&relayer_fault), Some(Penalty::Medium)));
+    }
 }
--- a/crates/consensus/primary/src/error/network.rs
+++ b/crates/consensus/primary/src/error/network.rs
@@ -71,36 +71,60 @@ pub(crate) enum PrimaryNetworkError {
 }
 
 impl PrimaryNetworkError {
-    /// Whether this fault is determined purely by the *content* of a gossip message — its BCS
-    /// encoding or its declared topic — and is therefore attributable to the message author
-    /// (`GossipMessage::source`) rather than the peer that relayed it.
+    /// Whether this gossip fault is determined by the *content* of the message and is therefore
+    /// charged to its authenticated publisher (the gossip source `GossipMessage::source`, resolved
+    /// to `author`), rather than to the peer that relayed it.
     ///
-    /// Mirrors `WorkerNetworkError::is_author_content_fault` (tn-worker) and is consulted only on
-    /// the gossip path (`PrimaryNetwork::process_gossip`): the network layer accepted and forwarded
-    /// the message after a shallow check, so an honest relayer could not have screened a
-    /// content-determined fault, and charging the relayer for it lets a Byzantine author ban honest
-    /// forwarders (see issues #801/#819). The request/response paths penalize the peer
-    /// unconditionally because there the peer *is* the originator.
+    /// Consulted only on the gossip path (`PrimaryNetwork::process_gossip`), whose deep validation
+    /// of a signed payload runs *after* gossipsub has already accepted and forwarded the message.
+    /// An honest relayer therefore cannot screen a content-determined fault before relaying it, and
+    /// charging the relayer for one lets a Byzantine author ban honest forwarders (#801/#785/#819).
+    /// The request/response paths never consult this classifier: there the peer *is* the originator
+    /// and is penalized unconditionally.
     ///
-    /// Restricted to the two `Fatal` faults authored by the gossip source and reachable from the
-    /// gossip handler: [`Self::Decode`] (`try_decode` of the payload) and [`Self::InvalidTopic`]
-    /// (the payload variant was published to the wrong topic). Every other variant is either
-    /// benign on the gossip path or concerns embedded certificate/consensus content whose signer
-    /// is carried inside the payload and is not necessarily the gossip source, so its existing
-    /// relayer/no-penalty attribution is left unchanged.
+    /// The publisher is the accountable party (not the relayer, and not a signer embedded in the
+    /// payload) for two reasons:
+    ///
+    /// 1. The source is authenticated. Gossipsub runs with `MessageAuthenticity::Signed` and
+    ///    `ValidationMode::Strict` (`crates/network-libp2p/src/consensus.rs`), so an accepted
+    ///    message carries a source signed by that peer's key; an attacker cannot publish under a
+    ///    victim's identity. Honest nodes only gossip content they produced and (for consensus
+    ///    results / epoch votes) self-signed (`publish_certificate` / `publish_consensus` /
+    ///    `publish_epoch_vote`), so a content-determined fault is the publisher's own.
+    ///
+    /// 2. Charging the *embedded* signer would be unsafe. A certificate is a multi-party aggregate,
+    ///    and the vote / consensus-result faults here are exactly signature-verification failures,
+    ///    so the key named in the payload is unauthenticated at the point of fault. An attacker
+    ///    could gossip a `ConsensusResult { validator: <victim>, .. }`, or an `EpochVote` naming
+    ///    any committee key, with a bad signature and have the innocent victim penalized: a framing
+    ///    vector the authenticated publisher carries no risk of.
+    ///
+    /// Parallels `WorkerNetworkError::is_author_content_fault` (tn-worker) at the mechanism level,
+    /// though the primary additionally covers the deep embedded-signature faults its signed
+    /// payloads carry, which the worker has no equivalent for.
+    ///
+    /// True for the envelope faults [`Self::Decode`] / [`Self::InvalidTopic`] and for the embedded
+    /// signed-payload content faults reachable from the gossip handler: [`Self::Certificate`] (a
+    /// malformed / unsigned / inquorate / bad-signature certificate), [`Self::InvalidHeader`] (an
+    /// `EpochVote` failing its signature or committee-membership check), and
+    /// [`Self::UnknownConsensusHeaderCert`] (a `ConsensusResult` with a bad signature). The
+    /// remaining arms are transport/local/benign and keep their relayer / no-penalty attribution;
+    /// [`Self::PeerNotInCommittee`] is content-determined but already maps to no penalty, so the
+    /// relayer was never charged for it.
     pub(crate) fn is_author_content_fault(&self) -> bool {
         //
         // explicitly match every variant so the classification is revisited with changes
         //
         match self {
-            PrimaryNetworkError::Decode(_) | PrimaryNetworkError::InvalidTopic => true,
-            PrimaryNetworkError::InvalidHeader(_)
+            PrimaryNetworkError::Decode(_)
+            | PrimaryNetworkError::InvalidTopic
+            | PrimaryNetworkError::InvalidHeader(_)
             | PrimaryNetworkError::Certificate(_)
-            | PrimaryNetworkError::StdIo(_)
+            | PrimaryNetworkError::UnknownConsensusHeaderCert(_) => true,
+            PrimaryNetworkError::StdIo(_)
             | PrimaryNetworkError::Storage(_)
             | PrimaryNetworkError::InvalidRequest(_)
             | PrimaryNetworkError::Internal(_)
-            | PrimaryNetworkError::UnknownConsensusHeaderCert(_)
             | PrimaryNetworkError::UnknownConsensusOutput(_)
             | PrimaryNetworkError::PeerNotInCommittee(_)
             | PrimaryNetworkError::UnavailableEpoch(_)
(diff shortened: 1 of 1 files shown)


sstanfield (dismissed):
LGTM!

grantkee (commented):
Can an "honest" peer push a maliciously published message?

This may be a consequence of propagating a message that is loosely validated (message signature only - not payload quality). Is it worth validating fully before propagating? This may delay gossip, so this may be best to filter by topic hash. For example, primary `Consensus` gossip message can just have signature validation by current committee member to keep propagation swift

As it stands now, validators are excluded from penalties and the only nodes allowed to produce gossip. We want to ensure honest validators that rotate out of the current committee (tracked in `prev_committee`) aren't penalized for late gossip near an epoch boundary.

MavenRain:
> Can an "honest" peer push a maliciously published message?
> 
> This may be a consequence of propagating a message that is loosely validated (message signature only - not payload quality). Is it worth validating fully before propagating? This may delay gossip, so this may be best to filter by topic hash. For example, primary `Consensus` gossip message can just have signature validation by current committee member to keep propagation swift
> 
> As it stands now, validators are excluded from penalties and the only nodes allowed to produce gossip. We want to ensure honest validators that rotate out of the current committee (tracked in `prev_committee`) aren't penalized for late gossip near an epoch boundary.

Short version: this PR is deliberately narrow.  It only fixes *who* gets charged for the faults that deep validation surfaces *after* gossipsub has already forwarded a message, and it composes correctly with the committee exemption you're asking about.  Your propagation-gating idea is partly already in place and partly a good separate improvement.  Details per point:

**1. "Can an honest peer push a maliciously published message?"**

Yes, and that is exactly the case for this PR.  In the mesh an honest peer *relays* (pushes) content authored by someone else, and it forwards after only the shallow `verify_gossip` check (size + authorized-publisher), before the primary's deep payload validation runs. So an honest relayer genuinely cannot screen a signature/quorum fault before relaying it. Today those faults are charged to that relayer, which is the #801/#785/#819 ban-the-forwarder vector.  This PR moves them to the authenticated author.

What an honest peer *cannot* do is become the authenticated `source`/author of a malicious payload.  Gossipsub runs `MessageAuthenticity::Signed` + `ValidationMode::Strict` (`network-libp2p/src/consensus.rs:313,326`), so `GossipMessage::source` is bound to the publisher's own key and cannot be spoofed onto a victim.  That is also why the PR charges the authenticated `source` (`network/mod.rs:1572`, `author` resolved from `message.source` at `consensus.rs:964`) and never the signer named *inside* the payload: at a signature-verification failure the embedded key is unauthenticated, so charging it would let an attacker gossip `ConsensusResult { validator: <victim>, .. }` (or an `EpochVote` naming any committee key) with a bad signature and frame the victim. The classifier only decides the penalty *magnitude*;  the *target* is always `message.source`, so there is no embedded-signer charge path.

**2. "Is it worth validating fully before propagating? ... primary Consensus gossip can just have committee-member signature validation to keep propagation swift."**

Partly we are already there, partly there is a real improvement worth a separate change.  Every message is signature-authenticated by gossipsub `Strict` + `Signed` before it is delivered, and `verify_gossip` (`consensus.rs:1421`) runs *before* `report_message_validation_result` (the call that actually decides propagation), so the swift pre-propagation gate exists.  However, the committee-publisher restriction is only wired on some topics:

- Certificate gossip (`primary_topic`) is already restricted: `start_epoch.rs:510` subscribes it via `subscribe_with_publishers(..., committee_keys)`, so a non-committee author is rejected and never propagated.  This is exactly your suggestion, already in place.
- The consensus-output and epoch-vote topics, however, are currently subscribed *open*: `node.rs:461` and `node.rs:457` call plain `.subscribe(...)` with no publisher set, so `verify_gossip` accepts any signed source on them. Those are the "Consensus" payloads you're describing (`ConsensusResult` -> `UnknownConsensusHeaderCert`, `EpochVote` -> `InvalidHeader`), so restricting them to committee publishers the way `primary_topic` already is would be a genuine hardening, not a no-op.

Full *deep* validation before propagating (BLS aggregate verify, quorum, epoch-vote committee check) is intentionally deferred to the app layer (`consensus.rs:957-963`), because doing it pre-forward would delay gossip, which is the cost you flagged.  This PR keeps that deferral and just corrects the attribution of the resulting penalty.  I'd suggest tackling the open-topic restriction as its own follow-up (see the boundary caveat in point 3, since a naive "current committee only" restriction would break late epoch-boundary gossip).  I'd be happy to file that issue / take that PR if you want it.

**3. "Ensure honest validators that rotate out of the current committee (prev_committee) aren't penalized for late gossip near an epoch boundary."**

This is already guaranteed by the peer manager's exemption, it is topic-independent, and this PR composes with it. `AllPeers::exemption` (`all_peers.rs:851`) grants `TrustBasis::Validator` to any peer whose key is in `previous_committee || current_committee || next_committee`, and `Peer::apply_penalty` (`peer.rs:154`) has exempt peers bypass the score model entirely: the score is never touched, so `process_penalty` returns `NoAction` and the ban/disconnect branches are unreachable (Severe/Fatal are surfaced as a `warn!` for ops only).  The exemption is evaluated live at penalty time from the three slots, so it cannot drift out of sync with rotation, and `previous_committee` is refreshed every epoch from on-chain state (`start_epoch.rs:155`) precisely so the outgoing committee is not pruned or penalized during the late-gossip window (`consensus.rs:841` note; `RejectReason::UnauthorizedAuthor` doc at `consensus.rs:1964` states the same for the reject path).

Every penalty path funnels through this single exemption check, so a rotated-out validator in `previous_committee` is covered in all cases:
- Accept path (this PR): a deep-validation fault is charged to the author via `report_penalty` -> `process_penalty` -> `exemption`; an exempt author's score is untouched.
- Reject path (pre-existing): an `UnauthorizedAuthor` reject's `FatalAuthor` charge goes through the same `process_penalty` -> `exemption`.

So this PR introduces no new penalty for `previous_committee` validators.  The one honest-party residual I noted in the description (`CertificateError::TooOld` re-publish charged Mild to the author) lands only on the restricted certificate topic, where the author is by construction a committee member, so it too is exempt.

The rotated-out case is already regression-tested at the exemption layer: `test_update_committees_in_window_peer_stays_trusted` (`tests/peers.rs:1500`) asserts a validator that rotates from `current` into the `previous` slot keeps `TrustBasis::Validator`, `test_committee_rotation_revokes_validator_exemption` (`tests/peers.rs:852`, issue #715) asserts an in-window member's fatal penalty is suppressed, and `test_update_committees_demotes_peer_that_exits_window` (`tests/peers.rs:1522`) pins the boundary where the exemption ends. What those don't cover is the new wiring specific to this PR, that an `is_author_content_fault` on the primary path routes to `author` (`GossipMessage::source`) and thence into that exemption.  If you'd like belt-and-suspenders on the new attribution, I'm happy to add a primary-side test asserting the author (not the relayer) is the charged party for each newly-moved variant, so a future change can't silently regress it.  I can fold it into this PR, if you'd like.

grantkee (approved):

PR #891, opened 2026-07-09
patch: catchup finalized watermark inconsistency on restart

### LINKED ISSUE #890: patch: Align catchup's scan watermark with replay's (or self-heal the finalized marker at startup)
**Status**: open — follow-up recommended during the closing-block seeding refactor (`5c332b6c`)
**Severity**: medium (silent mid-epoch gas/leader undercount after a specific crash window; operator-halt on the boundary-crossing variant)
**Scope**: `catchup_accumulator` / startup recovery only. Epoch-entry fee/count seeding is NOT affected — it derives from the closing block resolved via `getEpochInfo(entered).blockHeight − 1`, independent of the finalized marker.

## Problem

Startup recovery uses two different watermarks for "what has been executed":

1. **Catchup** (`catchup_accumulator`, `crates/node/src/manager/node.rs:176`) rebuilds the
   `GasAccumulator`'s gas stats and leader counts by scanning
   `epoch_info.blockHeight ..= finalized.number`, pinned to the **finalized marker**
   (`reth_env.finalized_header()`, node.rs:182, scan at node.rs:220). Leader counting is bounded
   by `last_executed_round` decoded from the **finalized header's nonce** (node.rs:249-251).
2. **Replay** (`replay_missed_consensus` → `get_missing_consensus`,
   `crates/state-sync/src/lib.rs:148`) resumes from the **latest executed canonical block**
   (`consensus_bus.last_executed_consensus_block` → `recent_blocks().latest_execution_block()`,
   `crates/consensus/primary/src/consensus_bus.rs:653`), i.e. the persisted execution tip — NOT
   the finalized marker. It only re-forwards consensus output whose blocks were never committed.

These two markers are persisted in **separate DB transactions**:
`finish_executing_output` commits the round's blocks (`save_blocks` + `commit`,
`crates/tn-reth/src/lib.rs:832`), and only afterwards does the engine-update path call
`finalize_block`, which persists `save_finalized_block_number` in a **second**
`database_provider_rw` transaction (`crates/tn-reth/src/lib.rs:802-815`).

A crash in the window between those two commits leaves, on restart:

```
finalized marker  <  persisted canonical tip
```

The blocks in the gap `(finalized.number, canonical_tip.number]` are then in no one's
jurisdiction:

- **Catchup skips them** — its scan and its leader-count bound end at the finalized marker.
- **Replay skips them** — its watermark is the canonical execution tip, and re-forwarding
  already-executed output would double-execute (explicitly forbidden, see the
  `replay_missed_consensus` doc in `crates/node/src/manager/node/start_epoch.rs:56-66`).

### A. Mid-epoch crash window (pre-existing, silent)

Restart with a one-round lag inside an epoch: catchup rebuilds gas stats and leader counts
**missing the gap round(s)** and the node runs the rest of the epoch on an undercounted
accumulator. Consequences at the epoch close:

- **Leader counts feed the rewards counter** consumed by the close-time system calls
  (incentives). An undercounting node diverges from peers that counted the round live.
- **Gas totals feed `adjust_base_fees`** at the live close. The next-epoch fee this node
  computes from undercounted gas can differ from the committee's. (The entry derivation
  re-derives fees from chain state at the next entry, so fee divergence self-corrects at the
  entry seam — but the close-window value and anything derived from the accumulator before then
  is wrong.)

This is silent: no error, no warning, just slightly-wrong per-epoch accounting. It exists on
`main` as well as this branch; the current integration test
`catchup_guards_finality_lag_across_epoch_boundary` (crates/node/tests/it/main.rs) pins the
mid-epoch lag as *legal* and even asserts the scan is bounded by the finalized header — i.e. the
undercount is currently tested-in behavior.

### B. Boundary-crossing crash window (now an intentional halt)

If the gap spans an epoch boundary (crash after committing the closing block but before the
finalize commit), the finalized marker points into epoch E−1 while the canonical tip reports
epoch E. Since `5c332b6c`, catchup's cross-view guard (node.rs, right after
`epoch_state_at_header`) detects this and **hard-errors at startup** with an operator hint,
because every pinned input (scan range, worker-count read block, leader-count bound) would
describe the wrong epoch. That guard is correct — before it, this shape silently rebuilt against
the wrong epoch's state — but it converts a self-inflicted persistence artifact into a
manual-intervention halt. The node's DB is fine; it just never advanced the marker.

### Option 1 (preferred): self-heal the finalized marker at startup

Before `catchup_accumulator` runs (call site `crates/node/src/manager/node.rs:842`), compare the
persisted canonical tip with the finalized marker. If the tip is ahead, advance the marker via
`reth_env.finalize_block(canonical_tip_header)`. This is sound because in this chain **every
committed block is consensus-final by construction** — blocks only enter the canonical DB from
committed consensus output (`finish_executing_output`); there is no speculative/reorgable
segment. The lag is purely a two-transaction persistence artifact.

Effects: catchup then scans the true executed range (fixing A), the cross-view guard can never
fire on this artifact (fixing B — it remains as a tripwire for genuine corruption), and both
watermarks converge by construction. Also heals the `finalized` tag served over RPC.

Care points:
- Do the heal from the **persisted** canonical tip (DB), not any in-memory view, since this runs
  before the engine processes anything.
- `finalize_block` also calls `set_safe` and `remove_persisted_blocks`; verify those are
  harmless this early in startup (they should be — the same call runs on every live round).
- Order: heal → `epoch_state_from_canonical_tip` read (node.rs:837) → catchup. The existing
  canonical-epoch guard parameter stays.

### Option 2: repoint catchup's watermark at replay's

Change `catchup_accumulator` to pin its scan, its epoch state, and its leader-count bound to the
**persisted canonical tip header** instead of the finalized header (keeping the one-header
discipline: all inputs from that single sealed header). This aligns the two recovery stages on
one watermark without touching the marker itself.

Trade-offs vs option 1: the stale finalized marker remains stale (RPC `finalized` tag, and the
standalone `derive_idle_worker_fee` seam at node.rs:561 also reads `finalized_header()`); the
cross-view guard would be re-anchored or removed (canonical-vs-canonical can't disagree), losing
its corruption-tripwire value.

## Acceptance criteria

1. A mid-epoch finalized lag at startup (harness exists in
   `catchup_guards_finality_lag_across_epoch_boundary`) results in gas stats and leader counts
   **including the gap blocks** — flip the current assertion that the scan stops at the
   finalized header; assert equality with a live accumulator that saw every round.
2. A boundary-crossing finalized lag at startup recovers cleanly (node starts, catchup succeeds
   against the entered epoch's state) instead of the guard halt. The guard itself stays
   (option 1) and a test proves it no longer fires on the healed artifact.
3. Replay double-execution is impossible: a test (or invariant argument in the PR) showing the
   healed/aligned watermark never causes `replay_missed_consensus` to re-forward output whose
   blocks are already committed.
4. Existing suites stay green: `cargo nextest run -p tn-node -p tn-types -p tn-reth` (baseline
   228/228) and the basefee e2e tests.

## Code references

| What | Where |
| --- | --- |
| Catchup scan pinned to finalized marker | `crates/node/src/manager/node.rs:176-260` (pin :182, scan :220, leader bound :249) |
| Cross-view guard (boundary halt) | `crates/node/src/manager/node.rs` (immediately after `epoch_state_at_header`) |
| Startup call site | `crates/node/src/manager/node.rs:835-842` |
| Blocks commit (txn 1) | `crates/tn-reth/src/lib.rs:832` `finish_executing_output` |
| Finalize commit (txn 2) | `crates/tn-reth/src/lib.rs:802` `finalize_block` |
| Replay watermark | `crates/state-sync/src/lib.rs:148` `get_missing_consensus`; `crates/consensus/primary/src/consensus_bus.rs:653` |
| Mid-epoch-lag test (asserts current undercount behavior) | `crates/node/tests/it/main.rs` `catchup_guards_finality_lag_across_epoch_boundary` |

## Background docs

- `issues/dual-header-read-robustness.md` — the earlier finalized-vs-canonical inconsistency
  work this follows from.
- Commit `5c332b6c` ("derive worker count and base fees from closing-block state at entry") —
  removed the finalized-marker dependency from entry seeding and added the startup guard; this
  issue is the recommended follow-up named in its review.

grantkee:
- Block persistence and the finalized/safe marker commit happen in separate DB transactions. A crash between them leaves the marker lagging the persisted canonical tip, which silently hides the gap blocks from gas-accumulator catchup, tx-pool seeding, and RPC finalized/safe tags. `RethEnv::heal_finalized_to_persisted_tip` (`crates/tn-reth/src/lib.rs`) advances the marker to the tip at startup, before anything reads it.
- If the marker is found *ahead* of the tip instead — meaning the execution DB lost blocks this node already attested as final — startup now hard-fails with `TnRethError::FinalizedMarkerAheadOfTip` rather than silently re-executing past attested state.
- `EpochManager::run` calls the heal before the canonical-tip epoch read and `catchup_accumulator`, so every startup reader sees a marker consistent with the persisted tip. The cross-view epoch-boundary guard in `catchup_accumulator` stays, with inverted meaning: since the benign crash-window lag is healed first, a firing guard now means genuine DB inconsistency, not a restart artifact.
- `set_worker_base_fee` used to drop a base-fee update for an uninitialized worker with just a warning and a `debug_assert!` (a no-op in release builds). Per PR #852 feedback, it now returns a real `eyre::Result` error that propagates through `start_epoch.rs` and shuts the node down — silently keeping a stale fee would let the worker admit underpriced transactions peers reject.
- Test coverage: `catchup_heals_finality_lag_across_epoch_boundary` reworked to cover both mid-epoch lag and boundary-crossing lag through the heal; new `heal_errors_when_finalized_marker_ahead_of_tip` and `heal_noop_on_fresh_genesis`; `test_catchup_accumulator` extended to prove a healed-then-caught-up run reproduces identical gas and per-leader reward counts as the live accumulator.

closes #890

### DIFF
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -236,13 +236,13 @@ impl ExecutionNodeInner {
     /// This returns the hash of the last executed ConsensusHeader on the consensus chain.
     /// since the execution layer is confirming the last executing block.
     pub(super) fn last_executed_output(&self) -> eyre::Result<ConsensusHeaderDigest> {
-        // NOTE: The payload_builder only extends canonical tip and sets finalized after
-        // entire output is successfully executed. This ensures consistent recovery state.
-        //
-        // For example: consensus round 8 sends an output with 5 blocks, but only 2 blocks are
-        // executed before the node restarts. The provider never finalized the round, so the
-        // `finalized_block_number` would point to the last block of round 7. The primary
-        // would then re-send consensus output for round 8.
+        // The payload builder commits an output's blocks and the finalized marker in separate
+        // transactions; a crash between them leaves the marker lagging the persisted tip.
+        // Startup heals that lag (`RethEnv::heal_finalized_to_persisted_tip`) before anything
+        // reads the marker, so the finalized block read here is the last block of the last
+        // consensus output whose execution was fully committed — the digest recovered below
+        // names exactly the last executed output, and the primary re-requests everything after
+        // it.
         //
         // recover finalized block's nonce: this is the last subdag index from consensus (round)
         let finalized_block_num = self.reth_env.last_finalized_block_number()?;
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -3,7 +3,7 @@
 //! This module contains the logic for execution.
 
 use crate::error::ExecutionError;
-use eyre::OptionExt;
+use eyre::{eyre, OptionExt};
 use jsonrpsee::http_client::HttpClient;
 use std::{net::SocketAddr, sync::Arc};
 use tn_batch_builder::BatchBuilder;
@@ -180,10 +180,14 @@ impl ExecutionNodeInner {
     /// Called every epoch so the pool charges the base fee the accumulator currently holds for
     /// the worker. This covers the respawn path, where `initialize_worker_components` is skipped
     /// but the base fee for the new epoch must still take effect. If the worker's components have
-    /// not been initialized, the update is dropped with a warning (and a debug assertion): the
+    /// not been initialized, the update is dropped with a warning and forces node shutdown: the
     /// worker's pool would keep charging the previous epoch's base fee, admitting underpriced
-    /// transactions into batches.
-    pub(super) fn set_worker_base_fee(&self, worker_id: WorkerId, base_fee: u64) {
+    /// transactions into batches that are rejected by peers.
+    pub(super) fn set_worker_base_fee(
+        &self,
+        worker_id: WorkerId,
+        base_fee: u64,
+    ) -> eyre::Result<()> {
         if let Some(worker) = self.workers.get(worker_id as usize) {
             let pool = worker.pool();
             let mut block_info = pool.block_info();
@@ -196,12 +200,12 @@ impl ExecutionNodeInner {
                 initialized_workers = self.workers.len(),
                 "set_worker_base_fee: dropping base-fee update for uninitialized worker"
             );
-            debug_assert!(
-                false,
+            return Err(eyre!(
                 "set_worker_base_fee: worker {worker_id} uninitialized ({} workers initialized)",
                 self.workers.len()
-            );
+            ));
         }
+        Ok(())
     }
 
     /// Respawn any tasks on the worker network when we get a new epoch task manager.
--- a/crates/node/src/engine/mod.rs
+++ b/crates/node/src/engine/mod.rs
@@ -165,9 +165,13 @@ impl ExecutionNode {
     ///
     /// Called every epoch so the pool charges the accumulator's current base fee for the worker,
     /// including on the respawn path where [`Self::initialize_worker_components`] is skipped.
-    pub async fn set_worker_base_fee(&self, worker_id: WorkerId, base_fee: u64) {
+    pub async fn set_worker_base_fee(
+        &self,
+        worker_id: WorkerId,
+        base_fee: u64,
+    ) -> eyre::Result<()> {
         let guard = self.internal.read().await;
-        guard.set_worker_base_fee(worker_id, base_fee);
+        guard.set_worker_base_fee(worker_id, base_fee)
     }
 
     /// Respawn any tasks on the worker network when we get a new epoch task manager.
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -160,12 +160,15 @@ pub(crate) struct EpochManager<P, DB> {
 ///
 /// Every chain-derived input (the block scan range's start and end, the worker-count read block,
 /// and the epoch used to bound leader counting) is pinned to the single finalized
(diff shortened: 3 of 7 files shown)


crates/node/src/engine/inner.rs
```diff
@@ -180,10 +180,14 @@ impl ExecutionNodeInner {
...
+    /// not been initialized, the update is dropped with a warning and forces node shutdown: the
     /// worker's pool would keep charging the previous epoch's base fee, admitting underpriced
-    /// transactions into batches.
-    pub(super) fn set_worker_base_fee(&self, worker_id: WorkerId, base_fee: u64) {
+    /// transactions into batches that are rejected by peers.
+    pub(super) fn set_worker_base_fee(
+        &self,
+        worker_id: WorkerId,
+        base_fee: u64,
+    ) -> eyre::Result<()> {
```
grantkee, line 190: Address feedback in #852

MavenRain: Many thanks

MavenRain (approved):
LGTM!

sstanfield (approved):
LGTM!

PR #892, opened 2026-07-09
fix: enforce the gossip size bound as a protocol constant on both paths (#872)

### LINKED ISSUE #872: `TooLarge` gossip rejects Fatal-ban the relayer on a per-node size bound that is not an enforced protocol constant
## Context

PR #819 (`819-gossip-reject-author-attribution`) reworks gossip fault handling into accountable-peer attribution, continuing the work tracked under #801 / #785 / #819.  This follow-up was surfaced during review of that PR and does not block it.

## Problem

On the gossip reject path, an oversized message (`RejectReason::TooLarge`) whose relaying peer can be resolved is mapped to `RejectPenalty::FatalRelayer`, which applies a `Penalty::Fatal` ban to the propagation source.  The code justifies charging the relayer with the rationale that `max_gossip_message_size` is a "shared constant", so that under gossipsub `Strict` validation an honest peer computes the same verdict and never forwards an oversized message.

That rationale is overstated. `max_gossip_message_size` is a per-node `#[serde(default)]` config field (default `12_000`), not an enforced protocol constant, and libp2p's `max_transmit_size` is never configured (so libp2p's default of `65536` applies and does not back-stop the 12k bound).  Under heterogeneous operator config, an honest relayer whose own configured limit is larger than ours can be Fatal-banned for forwarding a payload that is under its limit but over the local node's limit.  Nothing on the reject path reconciles against the relaying peer's own configured size limit.

## Details / Evidence

All citations are `file:line` in the current tree.

- `crates/network-libp2p/src/consensus.rs:1207-1210` — the size check that emits the reject: `if gossip.data.len() > self.config.max_gossip_message_size { return GossipAcceptance::Reject(RejectReason::TooLarge); }`.  The comparison is against the per-node config field, not a protocol constant.
- `crates/network-libp2p/src/consensus.rs:1763-1767` — the doc comment on `RejectReason::TooLarge` (variant at line 1767) that carries the "shared constant" rationale: "The bound is a shared constant, so under gossipsub `Strict` validation an honest peer computes the same verdict and never forwards the message".
- `crates/network-libp2p/src/consensus.rs:1793-1801` — the `RejectReason::penalty` doc repeating the same "size bound is a shared constant / `Strict` validation" reasoning.
- `crates/network-libp2p/src/consensus.rs:1804-1805` — the penalty mapping: `RejectReason::TooLarge if relayer_resolved => RejectPenalty::FatalRelayer,` else `=> RejectPenalty::Skip,`.
- `crates/network-libp2p/src/consensus.rs:934-944` — the application site: `FatalRelayer` calls `process_penalty(propagation_source, Penalty::Fatal)`, i.e. the actual Fatal ban of the relayer, with no cross-check against the relayer's own size limit.
- `crates/network-libp2p/src/consensus.rs:263-278` — the gossipsub `ConfigBuilder` sets `ValidationMode::Strict` (line 267) but never calls `.max_transmit_size()`, so libp2p's default (`65536`) applies and does not enforce the 12k bound. (`max_transmit_size` does not appear anywhere in the tree.)
- `crates/config/src/network.rs:109` — `pub max_gossip_message_size: usize`, a plain config field.
- `crates/config/src/network.rs:100` — `#[serde(default)]` on `struct LibP2pConfig`, the struct that houses the field (this is a struct-level attribute, not a per-field one at 109).
- `crates/config/src/network.rs:184` — the default value, `max_gossip_message_size: 12_000`.
- `crates/network-libp2p/src/peers/penalty.md:41-42` — penalty-table rows echoing the same "the size bound is deterministic network-wide" rationale (TooLarge + resolved relayer -> Fatal).  Note: these rows internally cite stale `consensus.rs:940/942` line numbers; the current application site is `consensus.rs:934-944`.
- `crates/network-libp2p/src/tests/network_tests.rs:2766-2778` — `oversized_reject_penalizes_only_a_resolved_relayer` and its doc, which restate "the size bound is deterministic network-wide, so under `Strict` validation an honest peer never forwards one" and assert `TooLarge -> FatalRelayer` / `Skip`.

The gap in plain terms: the attribution is sound only if every node computes the identical size verdict, but the bound is a per-node configuration value (default `12_000`) rather than an enforced protocol constant, and there is no lower-layer back-stop (`max_transmit_size` is unset), so the "shared constant" premise the comment relies on does not hold across heterogeneous operator config.

## Impact

Severity: Low.

- This is pre-existing behavior. Older code also relayer-banned oversized messages, so PR #819 does not introduce the ban; the change of concern is that the new comment overstates "shared constant" as the justification.
- The Fatal ban is mitigated for validators by the committee / allowlist `TrustBasis` short-circuit in `Peer::apply_penalty`, so a validator relayer is spared the Fatal outcome.
- Residual exposure is limited to non-committee relayers running config that differs from the local node's `max_gossip_message_size`.  Such a peer forwarding a payload under its own limit but over the local 12k bound can receive an unwarranted Fatal ban.

## Proposed fix

- Promote the bound to a genuine protocol constant, or document the uniform-config requirement and enforce it (for example, also set libp2p's `max_transmit_size` to the same bound so the transport layer back-stops it network-wide).
- At minimum (interim / doc-only step), soften the rationale in the doc comments and docs from "shared constant" to "shared by convention" so the attribution rationale is not overstated.  Affected text: `consensus.rs:1763-1766`, `consensus.rs:1793-1801`, `peers/penalty.md:41-42`, and `tests/network_tests.rs:2766-2778`.

## Acceptance criteria

- [ ] `max_gossip_message_size` is either a compile-time protocol constant, or the uniform-config requirement is documented and enforced (e.g. libp2p `max_transmit_size` set to the same bound as a transport-layer back-stop).
- [ ] The "shared constant" / "deterministic network-wide" language in `consensus.rs`, `peers/penalty.md`, and `network_tests.rs` is corrected so it matches actual enforcement (genuinely enforced, or reworded to "shared by convention").
- [ ] The `TooLarge -> FatalRelayer` attribution rationale in code and docs is consistent with whichever enforcement approach is chosen.
- [ ] The stale `consensus.rs:940/942` line references in `peers/penalty.md:41-42` are updated to the current application site.
- [ ] `oversized_reject_penalizes_only_a_resolved_relayer` still passes, and its doc comment reflects the corrected rationale.

MavenRain:
Closes #872.

## Problem

On the gossip reject path, an oversized message (`RejectReason::TooLarge`)
whose relaying peer resolves is mapped to `RejectPenalty::FatalRelayer`, which
Fatal-bans the propagation source.  The doc comments justified this with the
claim that `max_gossip_message_size` is a "shared constant", so under gossipsub
`Strict` validation an honest peer computes the same verdict and never forwards
an oversized message.

That was overstated: `max_gossip_message_size` was a per-node
`#[serde(default)]` config field (default `12_000`), not an enforced protocol
constant.  Under heterogeneous operator config, an honest relayer whose own
configured limit is larger than ours could be Fatal-banned for forwarding a
payload that is under its own limit but over the local node's.

## Fix

Promote the bound to a genuine compile-time protocol constant,
`tn_config::MAX_GOSSIP_MESSAGE_SIZE` (`= 12_000`), and enforce it on **both**
gossip paths so the attribution is genuinely sound:

- **Receive path**: `verify_gossip` checks `MAX_GOSSIP_MESSAGE_SIZE` instead
  of `self.config.max_gossip_message_size`.  Because the bound is now identical
  on every honest node and gossipsub runs `Strict` with `validate_messages()`,
  a peer that deems a message `TooLarge` reports `Reject` and never forwards it,
  so a forwarded oversized payload is genuine relayer misbehavior.
- **Publish path**: the `NetworkCommand::Publish` handler refuses a message
  larger than `MAX_GOSSIP_MESSAGE_SIZE` locally, returning
  `PublishError::MessageTooLarge` to the caller.

### Why the publish-side guard is part of this fix

`Strict` validation only gates *forwarding* of received messages; it never runs
on a node's own originated message.  Without a publish-side check the enforcement
is asymmetric: gossipsub's own cap is the unset `65_536`-byte default, so a
non-committee node, which uses the non-CVV `publish_txn` path
(`worker/src/network/handle.rs`), and whose transactions reth's pool allows up
to 128 KiB — could originate a `12_001..65_535`-byte gossip message.  Its honest
first-hop neighbours would then reject it as `TooLarge` and Fatal-attribute it
to the propagation source, which on the first hop is the honest originator.  That
is the same class of honest-node-ban the issue is about, relocated from
heterogeneous config to a missing origination/forwarding symmetry.  Guarding the
publish path closes it: origination and forwarding now apply the identical
bound, so the "genuine relayer misbehavior" rationale in the docs is literally
true, and an honest node that tries to publish an oversized message gets a
clear local error instead of a silent network-wide drop plus a peer-applied ban.

This behaviour is strictly safer than before: a `> 12_000`-byte message was
already un-gossipable (every receiver rejects it) . . . the guard just surfaces that
locally and immediately instead of after a failed fan-out.

### Changes

- [x] Add `tn_config::MAX_GOSSIP_MESSAGE_SIZE` (`= 12_000`), documented as a
      protocol constant enforced on both paths.
- [x] `verify_gossip` and the `NetworkCommand::Publish` handler both enforce
      `MAX_GOSSIP_MESSAGE_SIZE`.
- [x] Remove the `max_gossip_message_size` field from `LibP2pConfig` (and its
      default);  drop the now-nonexistent key from the partial-yaml test fixture.
      Existing config files that still carry the key keep loading unchanged —
      `LibP2pConfig` does not deny unknown fields.
- [x] Correct the "shared constant" / "deterministic network-wide" language in
      `consensus.rs`, `peers/penalty.md`, and `network_tests.rs`.
- [x] Update the stale `consensus.rs:940/942` line references in
      `peers/penalty.md` §3.1 to the current application sites (`994` / `1020`).
- [x] `oversized_reject_penalizes_only_a_resolved_relayer` still passes (only
      its doc changed); add `oversized_publish_is_refused_locally` covering the
      publish guard.

## Why not set libp2p `max_transmit_size`?

The issue listed setting libp2p's `max_transmit_size` to the same bound as an
alternative transport-layer back-stop.  I did not take that path: `max_transmit_size`
caps the whole protobuf-encoded RPC frame, not a single message's `data` field,
so pinning it at `12_000` would refuse a legitimate `12_000`-byte message (which
`verify_gossip` accepts, since its check is `> MAX_GOSSIP_MESSAGE_SIZE`) once
protobuf framing and any piggybacked gossipsub control are added.  The frame and
per-message limits are at different granularities, so `max_transmit_size` cannot
cleanly enforce the `12_000`-byte payload bound.  The compile-time constant plus
symmetric application-layer enforcement makes the attribution sound without it.

## Notes / out of scope

- `peers/penalty.md` carries other `consensus.rs:<line>` references (in §3.2–§3.5)
  that also drifted after the #819 merge (same ~200-line offset).  This PR updates
  only the two rows the issue names (§3.1).  A full line-number refresh, or a
  switch to stable anchors (function / match-arm names) since raw line numbers
  drift chronically here, is a separate cleanup.

## Testing

- `cargo clippy -p tn-config -p tn-network-libp2p --all-targets` — clean.
- `cargo test -p tn-config` (network config round-trip / partial-yaml / defaults).
- `cargo test -p tn-network-libp2p` — `oversized_reject_penalizes_only_a_resolved_relayer`,
  `author_attributable_reject_charges_the_author_never_the_relayer`, and the new
  `oversized_publish_is_refused_locally`.

### DIFF
--- a/crates/config/src/network.rs
+++ b/crates/config/src/network.rs
@@ -95,18 +95,28 @@ impl NetworkConfig {
     }
 }
 
+/// The maximum size, in bytes, of a gossip message payload (`GossipMessage::data`).
+///
+/// This is a protocol constant, not a per-node tunable, and is deliberately identical on every
+/// node. The gossip reject path Fatal-bans the *relaying* peer for delivering an oversized payload
+/// (`RejectReason::TooLarge` -> `RejectPenalty::FatalRelayer`), and that attribution is sound only
+/// if every honest node applies the identical bound: under gossipsub `Strict` validation a peer
+/// that deems a message `TooLarge` reports `Reject` and never forwards it, so a delivered oversized
+/// payload is genuine relayer misbehavior. The bound is enforced symmetrically on both the publish
+/// path (the network layer refuses to originate a larger message) and the receive path, so an
+/// honest node never emits one for its peers to reject. A per-node bound would instead let an
+/// honest relayer configured with a larger limit be Fatal-banned for forwarding a payload that is
+/// under its own limit but over ours, so the bound is fixed here network-wide rather than read from
+/// operator config. The largest legitimate gossip message is a `Certificate`; 12,000 bytes is sized
+/// to accommodate it.
+pub const MAX_GOSSIP_MESSAGE_SIZE: usize = 12_000;
+
 /// Configurations for libp2p library.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(default)]
 pub struct LibP2pConfig {
     /// Maximum message size between request/response network messages in bytes.
     pub max_rpc_message_size: usize,
-    /// Maximum message size for gossipped network messages in bytes.
-    ///
-    /// The largest gossip message is likely a `Certificate`.
-    /// The default of 12,000 bytes should be enough for legit gossip.
-    /// Large messages should not be gossipped
-    pub max_gossip_message_size: usize,
     /// The maximum duration to keep an idle connection alive between peers.
     ///
     /// The strategy for TN is to rely on QUIC to send keep alive messages instead of adding
@@ -181,7 +191,6 @@ impl Default for LibP2pConfig {
     fn default() -> Self {
         Self {
             max_rpc_message_size: 1024 * 1024,                    // 1 MiB
-            max_gossip_message_size: 12_000,                      // 12kb
             max_idle_connection_timeout: Duration::from_secs(65), // same as quic handshake
             max_px_disconnects: 10,
             px_disconnect_timeout: Duration::from_secs(3),
@@ -503,7 +512,6 @@ mod tests {
         let yaml = r#"
 libp2p_config:
   max_rpc_message_size: 2097152
-  max_gossip_message_size: 12000
   max_idle_connection_timeout:
     secs: 65
     nanos: 0
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -19,8 +19,8 @@ use crate::{
 use futures::StreamExt as _;
 use libp2p::{
     gossipsub::{
-        self, Event as GossipEvent, IdentTopic, Message as GossipMessage, MessageAcceptance, Topic,
-        TopicHash,
+        self, Event as GossipEvent, IdentTopic, Message as GossipMessage, MessageAcceptance,
+        PublishError, Topic, TopicHash,
     },
     kad::{self, store::RecordStore, Mode, QueryId},
     request_response::{
@@ -36,7 +36,7 @@ use std::{
     io::ErrorKind,
     time::Duration,
 };
-use tn_config::{KeyConfig, LibP2pConfig, NetworkConfig, PeerConfig};
+use tn_config::{KeyConfig, LibP2pConfig, NetworkConfig, PeerConfig, MAX_GOSSIP_MESSAGE_SIZE};
 use tn_types::{
     encode, now, BlsPublicKey, BlsSigner, Database, NetworkKeypair, NetworkPublicKey, TaskSpawner,
     TnSender, WorkerId,
@@ -720,8 +720,18 @@ where
                 send_or_log_error!(reply, peer_id, "LocalPeerId");
             }
             NetworkCommand::Publish { topic, msg, reply } => {
-                let res =
-                    self.swarm.behaviour_mut().gossipsub.publish(TopicHash::from_raw(topic), msg);
+                // Enforce `MAX_GOSSIP_MESSAGE_SIZE` at origination, symmetrically with the
+                // receive-side check in `verify_gossip`. Honest peers reject an oversized payload
+                // as `RejectReason::TooLarge` and Fatal-attribute it to the
+                // relaying peer; on the first hop that relayer is the originator,
+                // so a node that published an oversized message would be banned by
+                // its own neighbours. Refuse locally with a clear error instead, so
+                // origination and forwarding apply the identical bound.
+                let res = if msg.len() > MAX_GOSSIP_MESSAGE_SIZE {
+                    Err(PublishError::MessageTooLarge)
+                } else {
+                    self.swarm.behaviour_mut().gossipsub.publish(TopicHash::from_raw(topic), msg)
+                };
                 if res.is_ok() {
                     self.metrics.record_gossip_published();
                 }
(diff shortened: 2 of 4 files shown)


crates/config/src/network.rs
```diff
@@ -95,18 +95,28 @@ impl NetworkConfig {
...
 #[serde(default)]
 pub struct LibP2pConfig {
     /// Maximum message size between request/response network messages in bytes.
     pub max_rpc_message_size: usize,
-    /// Maximum message size for gossipped network messages in bytes.
-    ///
-    /// The largest gossip message is likely a `Certificate`.
-    /// The default of 12,000 bytes should be enough for legit gossip.
-    /// Large messages should not be gossipped
-    pub max_gossip_message_size: usize,
```
sstanfield, line 109: We will have existing config files with this set which I believe will cause a node to error out on startup?

sstanfield (approved):
Looks good except for the config file issue I called out.  Pretty sure we will have to address that. I don't think it will just work and ignore the old value as it stands- it would be good if I am wrong about that...  Could tweak a test to find out (for instance where you removed it from a test config.

crates/config/src/network.rs
grantkee: That's right. We broke an upgrade by removing content from the config before. We should create a legacy config for testing and ensure it doesn't brick a node. The script `etc/local-testnet.sh` off main will generate the correct file you can use for testing. We can discuss this strategy as well. 

I think a fallback function that migrates the config would work, then we can remove it after the next network upgrade?

grantkee (approved):
This is a correct change. The only concern is handling existing configs on deployed testnet nodes

PR #893, opened 2026-07-10
Save consensus in pack files with the header then the batches instead of batches then the header.

sstanfield:
This will make validating while downloading easier and will help close some potential dos attacks based on bogus consensus output data.

### DIFF
--- a/crates/storage/src/archive/pack.rs
+++ b/crates/storage/src/archive/pack.rs
@@ -19,7 +19,7 @@ use std::{
     fmt::Debug,
     fs::{self, File},
     hash::Hasher as _,
-    io::{self, ErrorKind, Read, Seek, SeekFrom, Write},
+    io::{self, Read, Seek, SeekFrom, Write},
     marker::PhantomData,
     path::Path,
 };
@@ -448,8 +448,7 @@ where
     W: ?Sized + std::io::Write,
 {
     value_buffer.clear();
-    encode_into_buffer(value_buffer, value)
-        .map_err(|e| std::io::Error::new(ErrorKind::Other, e.to_string()))?;
+    encode_into_buffer(value_buffer, value).map_err(|e| std::io::Error::other(e.to_string()))?;
     let buffer = match compression {
         PackCompression::None => value_buffer,
         PackCompression::ZStd => {
--- a/crates/storage/src/archive/pack.rs
+++ b/crates/storage/src/archive/pack.rs
@@ -19,7 +19,7 @@ use std::{
     fmt::Debug,
     fs::{self, File},
     hash::Hasher as _,
-    io::{self, Read, Seek, SeekFrom, Write},
+    io::{self, ErrorKind, Read, Seek, SeekFrom, Write},
     marker::PhantomData,
     path::Path,
 };
@@ -44,8 +44,9 @@ where
         uid_idx: u64,
         read_only: bool,
         compression: PackCompression,
+        version: u16,
     ) -> Result<Self, OpenError> {
-        Ok(Self { inner: PackInner::open(path, uid_idx, read_only, compression)? })
+        Ok(Self { inner: PackInner::open(path, uid_idx, read_only, compression, version)? })
     }
 
     /// Length of the Pack file.
@@ -188,9 +189,11 @@ where
         uid_idx: u64,
         read_only: bool,
         compression: PackCompression,
+        version: u16,
     ) -> Result<Self, OpenError> {
-        let (data_file, header) = Self::open_data_file(path, uid_idx, read_only, compression)
-            .map_err(OpenError::DataFileOpen)?;
+        let (data_file, header) =
+            Self::open_data_file(path, uid_idx, read_only, compression, version)
+                .map_err(OpenError::DataFileOpen)?;
         Ok(Self {
             header,
             data_file,
@@ -233,33 +236,14 @@ where
     /// Do the actual insert so the public function can rollback easily on an error.
     fn append_inner(&mut self, value: &V) -> Result<u64, AppendError> {
         let record_pos = self.data_file.len();
-        self.value_buffer.clear();
-        encode_into_buffer(&mut self.value_buffer, value)
-            .map_err(|e| AppendError::SerializeValue(e.to_string()))?;
-        let buffer = match self.header.compression {
-            PackCompression::None => &self.value_buffer,
-            PackCompression::ZStd => {
-                self.compression_buffer.clear();
-                let mut compressor =
-                    zstd::stream::write::Encoder::new(&mut self.compression_buffer, 0)?;
-                compressor.write_all(&self.value_buffer)?;
-                compressor.finish()?;
-                &self.compression_buffer
-            }
-        };
 
-        let mut crc32_hasher = crc32fast::Hasher::new();
-        // Once we have written to write_buffer, it needs to be rolled back before returning an
-        // error. Space for the value length.
-        let value_size = (buffer.len() as u32).to_le_bytes();
-        self.data_file.write_all(&value_size)?;
-        crc32_hasher.update(&value_size);
-
-        self.data_file.write_all(buffer)?;
-        crc32_hasher.update(buffer);
-        let crc32 = crc32_hasher.finalize();
-        self.data_file.write_all(&crc32.to_le_bytes())?;
-
+        write_value(
+            value,
+            &mut self.data_file,
+            &mut self.value_buffer,
+            &mut self.compression_buffer,
+            self.header.compression,
+        )?;
         Ok(record_pos)
     }
 
@@ -332,17 +316,19 @@ where
         uid_idx: u64,
         ro: bool,
         compression: PackCompression,
+        version: u16,
     ) -> Result<(DataFile, DataHeader), LoadHeaderError> {
         let mut data_file = DataFile::open(path, ro)?;
         let file_end = data_file.data_file_end();
 
         let header = if file_end == 0 {
-            let header = DataHeader::new(uid_idx, compression);
+            let header = DataHeader::new(uid_idx, compression, version);
             header.write_header(&mut data_file)?;
             header
         } else {
             let header = DataHeader::load_header(&mut data_file, uid_idx)?;
-            if header.version() != 0 {
+            if header.version() > version {
+                // Do not allow a newer version than we request but allow an older.
                 return Err(LoadHeaderError::InvalidVersion);
             }
             if header.appnum() != 1 {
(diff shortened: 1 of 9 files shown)


crates/storage/src/archive/pack.rs
```diff
@@ -332,17 +316,19 @@ where
...
 
         let header = if file_end == 0 {
-            let header = DataHeader::new(uid_idx, compression);
+            let header = DataHeader::new(uid_idx, compression, version);
             header.write_header(&mut data_file)?;
             header
         } else {
             let header = DataHeader::load_header(&mut data_file, uid_idx)?;
-            if header.version() != 0 {
+            if header.version() > version {
```
MavenRain, line 330: With this change, is it the case that the type system no longer presents a v1 record order out of a v0 file?

crates/storage/src/consensus_pack.rs
```diff
@@ -1329,10 +1383,39 @@ pub async fn bytes_to_output<R: AsyncRead + Unpin>(
...
+    iter_to_output_legacy(&mut stream_iter, timeout, committee).await
+}
+
+/// Private helper to read the next record from a pack iterator or timeout if it takes
+/// longer than timeout.
+async fn next_output_record<R: AsyncRead + Unpin>(
+    iter: &mut AsyncPackIter<PackRecord, R>,
+    timeout: Duration,
+) -> Result<Option<PackRecord>, PackError> {
+    match tokio::time::timeout(timeout, iter.next()).await {
```
MavenRain, line 1413: Nit: Typically, matching on `Result<Option<Result<_, _>>, _>` could be simplified with a basic combination of `and_then` and `transpose`

MavenRain: I've noticed that several LLMs have an appetite for doing this when tokio timeouts are involved.

crates/storage/src/consensus_pack.rs
```diff
@@ -979,41 +1041,17 @@ impl Inner {
...
+            if self.version() == 0 { self.save_consensus_batches(consensus)? } else { None };
         // Now save the consensus header.
         let consensus_digest = consensus.consensus_header_hash();
         let position = self
             .data
             .append(&PackRecord::Consensus(Box::new(consensus.consensus_header())))
             .map_err(|e| PackError::Append(e.to_string()))?;
+        if self.version() > 0 {
+            self.save_consensus_batches(consensus)?;
+        };
```
MavenRain, line 1054: I guess that, now, unlike previously, if `save_consensus_batches` errors midway, anyone could look at a consensus header that points to non-existent batches.

MavenRain: But that is the point, I now realize . . . we shouldn't be satisfied with just seeing the header, but wait until we see all the batches that the header calls-out . . . got it

MavenRain (dismissed):
I dig it . . . nothing blocking

crates/storage/src/pack_validate.rs
grantkee, line 1: ### `validate_pack_file` cannot validate header-first (v1) packs

- **Category**: Bugs / Correctness (validation at a diagnostic boundary)
- **Location**: `crates/storage/src/pack_validate.rs:178-179` (opens accepting v0 and v1);
  `crates/storage/src/pack_validate.rs:234-311` (single-pass validation loop, no version branch);
  `crates/storage/src/pack_validate.rs:469-487` (`build_records` test helper);
  `crates/storage/src/pack_validate.rs:490-498` (`write_records` stamps `PACK_VERSION`)
- **Sole caller**: `crates/telcoin-network-cli/src/db.rs:95` (`tn db validate`, read-only operator diagnostic)

**Claim** (as submitted): `validate_pack_file` opens with `PACK_VERSION` (value 1), accepted by the
`> version` gate for both v0 and v1 files, but its single-pass loop only implements the legacy
batches-before-header ordering. On a genuine v1 pack every header is checked against an empty (or
previous-group) `batches` set, so every referenced digest is recorded as `MissingBatch` and every
non-first group's batches as `ExtraBatch`, yielding `Verdict::Invalid` for a well-formed pack. The
tests miss it because `build_records` emits batches-before-header (v0 order) into a
`PACK_VERSION`-stamped file.

**Evidence:**

- Write side is header-first for v1. `Inner::save_consensus_output` is version-branched at
  `consensus_pack.rs:1044-1054`: for v0 it writes batches first (`save_consensus_batches`, then the
  `Consensus` record); for **v1 it appends the `Consensus` header record first** (`:1048-1051`)
  **then the batches** (`:1052-1053`). `PACK_VERSION = 1` (`consensus_pack.rs:46`), so every pack
  written today is v1 / header-first.
- The read path agrees and is version-dispatched: `stream_import` picks the decoder by version
  (`consensus_pack.rs:960-969`) — `iter_to_output` (v1) requires the **first** record to be the
  `Consensus` header (`:1428-1440`) and reads the batches that follow; `iter_to_output_legacy` (v0)
  accumulates batches first, then the header (`:1567+`). A well-formed v1 pack therefore decodes
  cleanly.
- The validator does **not** branch on version. Its loop accumulates every `Batch` into `batches`
  (`pack_validate.rs:242-249`); on a `Consensus` record it requires each referenced digest to
  **already be present** in `batches` (`:286-298`), treats any leftover as `ExtraBatch`
  (`:300-305`), then clears the per-group sets (`:307-310`). That is exactly the batches-first (v0)
  semantic — it mirrors `iter_to_output_legacy`, not the v1 reader — and the module doc states the
  assumption outright (`:22-28`).
- **Per-record walk of a header-first v1 pack** (`EpochMeta`, `Consensus(h1)`, `B1…`,
  `Consensus(h2)`, `B2…`, …): at `Consensus(h1)` the `batches` set is empty → `MissingBatch` for
  every digest h1 references; at `Consensus(h2)` `batches` still holds h1's batches (not h2's) →
  `MissingBatch` for every h2 digest **and** `ExtraBatch` for every h1 batch; and so on. The final
  group's batches never get an `ExtraBatch` (no successor header). Post-loop, every missing digest
  is present somewhere in `all_batch_digests`, so all are reclassified `Misordered` (`:319-327`).
  Result: `Verdict::Invalid`.
- **Version gate does not save it.** `archive/pack.rs:330` rejects only `header.version() > version`;
  validate requests `PACK_VERSION = 1`, a v1 file reports `version() == 1`, `1 > 1` is false →
  accepted, loop runs.
- **Why the tests miss it.** `build_records` pushes each `Batch` (`:477-481`) then the `Consensus`
  header (`:483`) — batches-first (v0) content — and `write_records` stamps the file with
  `PACK_VERSION = 1` (`:492`). So the suite writes a **v1-stamped file carrying v0-ordered content**,
  the one arrangement the loop expects; no test feeds a genuine `save_consensus_output` v1 pack.
- **Empirical reproduction (executed, then reverted — not left in the tree).** A temporary
  `#[tokio::test]` built a real v1 pack via `ConsensusPack::open_append` +
  `save_consensus_output` (5 chained outputs, 4 batches each) + `persist`, then called
  `validate_pack_file(<dir>/epoch-0/data, 0, None)`:
  ```
  EMPIRICAL_VERDICT=INVALID
  EMPIRICAL_ISSUES_TOTAL=36
  EMPIRICAL_MISORDERED=20 EMPIRICAL_ABSENT=0 EMPIRICAL_EXTRA=16
  ```
  20 = 5 headers × 4 referenced-but-unseen digests; 16 = 4 non-first groups × 4 orphaned
  previous-group batches. A well-formed pack is reported `INVALID` with 36 spurious issues, matching
  the trace exactly.

**Severity calibration.** An independent verifier assessed this **High** on the grounds that the
verdict is inverted for the *only* production on-disk format and the test suite gives false
confidence. Held at **Medium** here: the sole caller is the read-only operator diagnostic
`tn db validate` (`db.rs:95`); it gates no sync/import/consensus path, cannot corrupt state, and the
node itself is unaffected. The defect is certain to manifest but its blast radius is confined to a
diagnostic aid — "guaranteed to manifest" is a likelihood, not an impact, multiplier. Not Low/Info
because it inverts the result on 100% of current packs and is actively misleading during incident
response. Reasonable reviewers could argue Medium vs High; the fix is the same either way.

**Proposed fix.** Make the validation loop version-aware, mirroring the read path's own v0/v1 branch
(`consensus_pack.rs:960-969`). Read the on-disk version from the opened pack and, for v1, group
header-first (finalize a group when the next header arrives, and the last group at EOF):

```rust
// after the existing open (pack_validate.rs:178-179):
let version = pack.version(); // Pack::version() -> u16 (archive/pack.rs:94)

// shared helper — keeps both branches in lockstep:
fn finalize_group(
    number: u64,
    referenced: &HashSet<BlockHash>,
    batches: &HashSet<BlockHash>,
    issues: &mut Vec<PackIssue>,
) {
    for digest in referenced {
        if !batches.contains(digest) {
            issues.push(PackIssue::MissingBatch { number, digest: *digest, class: BatchClass::Absent });
        }
    }
    for digest in batches {
        if !referenced.contains(digest) {
            issues.push(PackIssue::ExtraBatch { number, digest: *digest });
        }
    }
}

// v0 keeps the existing loop (batches-first). v1 (header-first) branch:
let mut pending: Option<(u64, HashSet<BlockHash>)> = None; // (number, referenced digests)
let mut batches: HashSet<BlockHash> = HashSet::new();      // batches seen in the pending group

for record in iter {
    match record? {
        PackRecord::EpochMeta(_) => issues.push(PackIssue::EpochMetaMismatch {
            detail: "epoch meta data found more than once".to_string(),
        }),
        PackRecord::Batch(batch) => {
            batch_count += 1;
            let digest = batch.digest();
            batches.insert(digest);
            all_batch_digests.insert(digest);
        }
        PackRecord::Consensus(consensus_header) => {
            // Close the previous group now that all its batches have been read.
            if let Some((prev_number, referenced)) = pending.take() {
                finalize_group(prev_number, &referenced, &batches, &mut issues);
            }
            batches.clear();

            consensus_count += 1;
            let number = consensus_header.number;
            first_consensus_number.get_or_insert(number);
            last_consensus_number = Some(number);

            // Order-independent checks — identical to the v0 path (pack_validate.rs:264-281).
            let expected_number = start_consensus_number + (consensus_count - 1);
            if number != expected_number {
                issues.push(PackIssue::NonSequentialConsensusNumber { expected: expected_number, found: number });
            }
            if let Some(parent) = expected_parent {
                if consensus_header.parent_hash != parent {
                    issues.push(PackIssue::ChainBreak {
                        number, expected_parent: parent, found_parent: consensus_header.parent_hash,
                    });
                }
            }
            expected_parent = Some(consensus_header.digest());

            // Open the new group: record the digests this header references.
            let mut referenced = HashSet::new();
            for header in consensus_header.sub_dag.headers() {
                for (digest, _) in header.payload().iter() {
                    referenced.insert(*digest);
                }
            }
            pending = Some((number, referenced));
        }
    }
}
// Finalize the last group at EOF (its batches have no successor header to close them).
if let Some((number, referenced)) = pending.take() {
    finalize_group(number, &referenced, &batches, &mut issues);
}
// The deferred Absent/Misordered class-resolution pass (pack_validate.rs:319-327) is unchanged.
```

**Collateral (must ship with the fix).** The current test helpers hide the bug *and* would break
under a naive fix: `write_records` stamps `PACK_VERSION = 1` while `build_records` emits v0-ordered
content. Once the loop dispatches on version, that v1-stamped/v0-ordered file would be (correctly)
flagged as misordered, breaking `test_validate_clean_pack` etc. Reconcile by making the helper's
on-disk version match its record ordering — either stamp the batches-first helper as version 0
(`open_append_version`/`Pack::open(..., 0)` already allow this, `consensus_pack.rs:225`), or add a
header-first v1 helper — and add a genuine end-to-end v1 regression test built by the real writer
(the empirical reproduction above, asserting `Verdict::Valid` post-fix):

```rust
#[tokio::test]
async fn test_validate_clean_v1_pack() {
    let (temp_dir, committee, chain) = setup();
    let previous_epoch = EpochRecord {
        epoch: 0,
        committee: committee.bls_keys().iter().copied().collect(),
        next_committee: committee.bls_keys().iter().copied().collect(),
        ..Default::default()
    };
    let pack = ConsensusPack::open_append(temp_dir.path(), previous_epoch, committee.clone())
        .expect("open pack");
    for output in make_outputs(&committee, chain, 5) {
        pack.save_consensus_output(output).await.expect("queue output");
    }
    pack.persist().await.expect("persist");

    let data_path = temp_dir.path().join("epoch-0").join("data");
    let report = validate_pack_file(&data_path, 0, None).expect("validate");
    assert_eq!(report.verdict, Verdict::Valid, "genuine v1 pack must validate clean: {:?}", report.issues);
    assert_eq!(report.consensus_count, 5);
}
// (Pre-fix, this exact test fails with Verdict::Invalid / 36 issues — the reproduction run above.)
```

grantkee (dismissed):
The code LGTM. There is one outstanding issue with the CLI db `validate` subcommand. I asked claude to provide more context and shared the finding below. I think it's okay as a follow up PR if you think it's outside the scope of this PR @sstanfield

crates/storage/src/archive/pack.rs
sstanfield: Basically it makes sure an old version of the software does not try to open a newer pack than it can read.  That should be it.

crates/storage/src/consensus_pack.rs
sstanfield: I may be missing something but the Result have different Error types so I don't think that will work (?).

MavenRain (approved):

grantkee (approved):

PR #895, opened 2026-07-10
cut the primary consensus-output path fully over to the typed sync protocol (739)

MavenRain:
## What

Item 9b of #739.  Cuts the primary's single-consensus-output fetch (`PrimaryNetworkHandle::request_consensus_output`, the observer/state-sync catch-up path) **fully** over from the legacy `/tn-stream` digest-correlation transfer to the typed `/tn-primary-sync` protocol.  This is a hard cutover with **no legacy fallback**: the requester is sync-only, the responder answers the legacy request-response with a typed error, and the legacy consensus-output `/tn-stream` serve is removed.  Mirrors the item-7 MissingCertificates full cutover.

## Why

`StreamConsensusOutput` was the last primary `/tn-stream` opener that still carried a live requester.  It streams one output's opaque, pack-file-encoded bytes (returned as a `Vec<u8>` the caller decodes with the epoch committee and verifies against the already-known header digest).  A hard cutover removes the RPC-ack-plus-digest handshake, the correlation-digest machinery, and the legacy serve for this path in one step, rather than leaving a fallback to unwind later.

## How

**Base + discriminant.** Based directly on `origin/main`. `origin/main`'s `PrimarySyncRequest` is `{EpochPack(0), MissingCertificates(1)}`, so this appends `ConsensusOutput { number }` at BCS discriminant **2**. `request_tags_are_stable` pins it (`Some(&2)`) and a round-trip test covers the wire.  Append-only;  no existing variant is reordered or removed.

**Requester (sync-only).** `request_consensus_output` is rewritten to probe up to `MAX_EPOCH_SYNC_PROBES` connected peers over `/tn-primary-sync`, returning the reassembled bytes from the first peer that serves, otherwise `Err` (the caller, `get_consensus_output` in state-sync, already retries).  The structure mirrors the full-epoch-pack `request_epoch_pack_sync`: `filter` skips peers cached unsyncable, `then` classifies each exchange as `ConsensusOutputAttempt::{Fetched, Unsupported, Failed}`, and a pinned `filter_map(ready).next()` short-circuits on the first success.  `try_sync_consensus_output_exchange` opens the stream, writes `Req(ConsensusOutput { number })`, reads the `Ack`, and reassembles the `Data`/`End` frames into a `Vec<u8>` bounded by the `MAX_CONSENSUS_OUTPUT_STREAM_BYTES` (512 MiB) size cap and the per-frame `EPOCH_PACK_FRAME_TIMEOUT` inside the reader.

**Cache-poisoning guard.** A `ConsensusOutput` `Unsupported` (a peer that speaks `/tn-primary-sync` but cannot decode the newer discriminant-2 variant) does NOT write `false` to the shared per-peer `sync_capability` cache — unlike a full-pack `Unsupported`, which is authoritative — so a peer that still serves full epoch packs is not hidden from the full-pack path.

**Responder.** `process_inbound_sync_stream` gains a `ConsensusOutput` dispatch arm (the exhaustive `SyncFrame::Req(PrimarySyncRequest::…)` match forces it) routing to the new `RequestHandler::process_sync_consensus_output_stream`, which resolves the bytes via the existing `consensus_output_bytes` (with its brief catch-up wait) and serves them through the new `sync_codec::send_sync_consensus_output_over_stream`: `Ack` + `Data`* + `End`, or `Deny(Unavailable)` on a benign miss.  Reuses the epoch-pack chunk framing (`write_pack_data_frames`, `MAX_SYNC_PACK_FRAME_SIZE`) so the requester reassembles with the same `sync_pack_reader`.  Errors are metrics-only (no penalty), like the epoch-pack sync path.  Admission goes through the existing shared `try_admit_sync`.

**Legacy removal.** The request-response `PrimaryRequest::StreamConsensusOutput` is retained for wire compatibility (BCS request variant indices stay until the coordinated `/0.0.2` bump, item 9c) but is now answered with a typed `PrimaryResponse::Error` (mirroring `process_request_for_missing_certs`) instead of served.  Deleted: `request_consensus_output_from_peer`, `stream_consensus_output`, `send_consensus_output_over_stream`, the `StreamRequestKind::ConsensusOutput` variant and its `stream_request_digest`/`process_request_epoch_stream` arms, and the `CONSENSUS_OUTPUT_STREAM_READ_TIMEOUT` constant.

**Tests.** Request round-trip + tag-stability (`request.rs`);  `sync_codec` unit tests for the serve helper (round-trip, empty, `Deny(Unavailable)` on `None`, over-cap rejection).  The legacy `test_send_consensus_output_over_stream` (which drove the removed serve) is deleted;  `test_retrieve_consensus_output` (covering the retained `consensus_output_bytes`) stays.  As with the epoch-pack requester, `try_sync_consensus_output_exchange` and `process_sync_consensus_output_stream` run over an unmockable `libp2p::Stream`, so those are covered by the codec unit tests and the ignored observer e2e rather than a mocked-stream unit test.

## Compatibility / rollout

- Hard cutover (matches item 6 full-pack and item 7 missing-certs): a `request_consensus_output` that finds no peer serving the variant fails and the caller retries, forcing peers to upgrade.  There is no legacy fallback.
- A peer on an older binary that requests consensus output over the legacy request-response path receives a typed error and retries elsewhere.
- No wire-breaking change: the `ConsensusOutput` sync variant is append-only (disc 2);  the legacy `PrimaryRequest::StreamConsensusOutput` request variant is retained (answered-with-error) until the item-9c `/0.0.2` bump.  BCS encodes enum variant indices, so no deprecated variant is deleted before that coordinated bump.

## Gates

`-j2` throughout (OOM-safe). fmt (nightly-2026-03-20);  clippy `-p tn-network-libp2p -p tn-primary --all-features --all-targets` clean; `tn-network-libp2p` + `tn-primary` tests pass.  `cargo doc -D warnings` adds zero new warnings (pre-existing links only;  no cargo-doc CI gate).  Consumer crates compile with zero edits (`request_consensus_output` signature unchanged).

### DIFF
--- a/crates/consensus/primary/src/network/mod.rs
+++ b/crates/consensus/primary/src/network/mod.rs
@@ -91,9 +91,6 @@ pub const MAX_PENDING_REQUESTS_PER_PEER: usize = 2;
 /// TODO- replace with a size from consensus.  See Issue 782.
 const MAX_CONSENSUS_OUTPUT_STREAM_BYTES: usize = 512 * 1024 * 1024;
 
-/// Per-read timeout while draining a consensus-output stream (slow-peer guard).
-const CONSENSUS_OUTPUT_STREAM_READ_TIMEOUT: Duration = Duration::from_secs(10);
-
 /// Timeout for the responder's first sync frame (`Ack`/`Deny`) after the epoch-pack
 /// request frame is written. A peer that negotiated the sync protocol but does not
 /// answer (a pre-cutover node that registered the protocol but reads the stream on
@@ -152,18 +149,15 @@ pub enum StreamRequestKind {
         /// The final (inclusive) consensus header number to stream up to.
         last_consensus_number: u64,
     },
-    /// Stream the raw bytes for a single consensus output (by consensus chain number).
-    ConsensusOutput(u64),
 }
 
-/// Correlation digest for a single consensus-output stream request.
+/// Correlation digest for an epoch-pack stream request.
 ///
 /// Each kind is domain-separated so distinct requests never collide in the pending map:
 /// - `EpochPack(epoch)` hashes only the epoch — byte-for-byte the original full-stream scheme, so
 ///   existing full-stream clients are unaffected.
 /// - `EpochPackPartial { epoch, n }` additionally mixes in the stop number, giving it a distinct
 ///   digest from the full-epoch request.
-/// - `ConsensusOutput(n)` is tagged so output `N` never collides with epoch `N`.
 fn stream_request_digest(kind: &StreamRequestKind) -> B256 {
     let mut hasher = tn_types::DefaultHashFunction::new();
     match kind {
@@ -175,10 +169,6 @@ fn stream_request_digest(kind: &StreamRequestKind) -> B256 {
             hasher.update(b"epoch-pack-partial");
             hasher.update(&epoch.to_le_bytes());
             hasher.update(&last_consensus_number.to_le_bytes());
-        }
-        StreamRequestKind::ConsensusOutput(number) => {
-            hasher.update(b"consensus-output");
-            hasher.update(&number.to_le_bytes());
         }
     }
     B256::from_slice(hasher.finalize().as_bytes())
@@ -231,6 +221,24 @@ enum EpochPackAttempt {
     Imported,
     /// The peer did not answer the sync exchange; cache it unsyncable this epoch
     /// and skip it on later probes (it must upgrade to be syncable again).
+    Unsupported,
+    /// The peer answered but the exchange failed; try the next peer.
+    Failed(NetworkError),
+}
+
+/// The outcome of a single-consensus-output fetch attempt over the typed sync
+/// protocol.
+///
+/// Mirrors [`EpochPackAttempt`] but carries the fetched bytes on success (an output
+/// is reassembled into a `Vec<u8>` for the caller to decode and verify, not imported
+/// into the chain).
+enum ConsensusOutputAttempt {
+    /// The peer served the output's bytes.
+    Fetched(Vec<u8>),
+    /// The peer did not answer the sync exchange. Unlike a full-pack `Unsupported`
+    /// this is NOT cached unsyncable, because a peer that speaks `/tn-primary-sync`
+    /// but cannot decode this newer `ConsensusOutput` variant still serves full epoch
+    /// packs and must not be hidden from the full-pack path.
     Unsupported,
     /// The peer answered but the exchange failed; try the next peer.
     Failed(NetworkError),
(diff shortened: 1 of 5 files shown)


crates/consensus/primary/src/network/mod.rs
```diff
@@ -529,139 +537,209 @@ impl PrimaryNetworkHandle {
...
+    /// cached unsyncable, opening a stream whose opening frame carries
+    /// [`PrimarySyncRequest::ConsensusOutput`]; returns the reassembled bytes from the
+    /// first peer that serves, otherwise `Err` (the caller retries). Returns the
+    /// pack-file encoded output (batches + consensus header); the caller deserializes it
+    /// with the epoch committee and verifies the decoded header digest. A peer that
+    /// serves proves it speaks the sync protocol (cached `true`); a peer that does not
+    /// answer is NOT cached (an ambiguous `Unsupported`; see `ConsensusOutputAttempt`).
+    /// The probe is penalty-exempt.
+    pub async fn request_consensus_output(&self, number: u64) -> NetworkResult<Vec<u8>> {
+        let peers = self.handle.connected_peers().await?;
```
sstanfield, line 576: I like this pattern.  I am slightly concerned that this will lead to always hitting the same peer for all requests vs moving around who we request from.  This may be fine though and is out of scope for this PR just something we may want to decide if it's an issue or not.

crates/consensus/primary/src/network/mod.rs
```diff
@@ -1531,18 +1609,34 @@ where
...
+    /// `/tn-primary-sync` protocol, so this request-response path is no longer served:
+    /// peers fetch a single consensus output over sync. The request variant is retained
+    /// for wire compatibility (BCS variant indices stay until the coordinated `/0.0.2`
+    /// bump, item 9c); a legacy requester is answered with a typed error rather than
+    /// served.
     fn process_consensus_output_stream(
         &self,
         peer: BlsPublicKey,
-        number: u64,
+        _number: u64,
```
sstanfield: Any reason to keep this?

sstanfield (dismissed):
Good deal moving this to new streaming.  We may need to move some of this to work with ConsensusOutput vs bytes just because we will be able to do some early checks if we decode to filter out bogus bytes.  Out of scope for now, I'll look at it once this and my recent PR land.

crates/consensus/primary/src/network/mod.rs
MavenRain: Nope.  I dropped it.  The handler no longer serves this request (it just returns the typed error now), so the number was dead.  I removed the param and bound the dispatch as `StreamConsensusOutput { .. }`.  The request variant keeps its `number` field for wire compatibility until the `/0.0.2` bump (item 9c).

crates/consensus/primary/src/network/mod.rs
MavenRain: Good catch.  Right now this probes `connected_peers()` in order and short-circuits on the first peer that serves, so a stable peer list does mean the same peer keeps answering (the full epoch-pack path `request_epoch_pack_sync` has the identical shape).  It's fine for correctness since any serving peer is authoritative and the caller verifies the decoded header digest, but you're right that it doesn't spread load.  Leaving it out of scope here;  I'd be happy to add a rotation/shuffle across both sync probe paths as a follow-up under 739 if we decide it's worth it.

MavenRain:
> Good deal moving this to new streaming. We may need to move some of this to work with ConsensusOutput vs bytes just because we will be able to do some early checks if we decode to filter out bogus bytes. Out of scope for now, I'll look at it once this and my recent PR land.

That makes sense.  Today the network layer serves the opaque pack-encoded bytes and the caller decodes plus verifies against the epoch committee and the known header digest, so bogus bytes get caught one layer up rather than filtered at receive.  Decoding at the boundary to reject garbage earlier is a clean follow-up.  I'd be happy to coordinate once this and your PR land so we only touch the serve/reader path once.

grantkee (approved):
LGTM!

Issue #894, opened 2026-07-10
`test`: cut targeted test-suite runtime ~25% by memoizing per-proptest-case fixtures and sharing crypto contexts (no coverage loss)


MavenRain:
## Summary

A batch of ~56 unit/integration tests across `tn-primary`, `tn-types`,
`tn-storage`, `tn-reth`, `tn-network-libp2p`, `tn-worker`, and
`tn-worker-gateway` spends most of its wall time rebuilding expensive,
input-invariant fixtures inside hot loops: BLS-keyed `CommitteeFixture`s rebuilt
once per proptest case, and a fresh `secp256k1` context built per test
transaction. Hoisting/memoizing those, plus replacing a few fixed wall-clock
`sleep`s with event-driven waits, cuts the targeted suite by ~25% **without
changing what any test asserts and without lowering any proptest case count or
benchmark iteration count.**

This is a test-only change with no production-code behavior change.

## Baseline (how the time was measured)

`cargo nextest run` of the 56 targeted tests, warm cache, on worktree
`990d2127` (2 commits behind `origin/main` `141a0919`;  line numbers below should
be re-confirmed against `main` before the PR).  Sum of individual test durations
≈ **1143s**; 25% ≈ **286s** to remove.  The time is concentrated:

| Cluster | Tests | ~seconds | Root cause |
|---|---|---|---|
| `tn-reth::it pipeline_tel_precompile_props` | 12 | 407 | real reth pipeline ×120 (see "Not worth doing") |
| `tn-primary::it consensus_props` | 6 | 186 | `CommitteeFixture` rebuilt per proptest case |
| `tn-primary certificate_tests::test_certificate_verification` | 1 | 69 | `CommitteeFixture` rebuilt per proptest case |
| `tn-storage consensus_pack::test_consensus_pack` | 1 | 69 | `Secp256k1::new()` per test tx (~20k) |
| `tn-storage archive...test_archive_hdx_index` | 1 | 86 | 4M index ops (see "Not worth doing") |
| `tn-storage consensus::{race, db_stream, ...}` | 7 | 84 | `Secp256k1::new()` per test tx |
| `tn-storage *_dbsimpbench` | 3 | 65 | microbenchmarks-as-tests |
| `tn-types::it committee_props` | 5 | 82 | `BlsKeypair::generate` ×256 cases, keys unused |
| network / worker-gateway / task_manager | ~14 | 40 | mostly fixed `sleep` pacing |

[...]

PR #899, opened 2026-07-10
test: cut targeted suite runtime by memoizing fixtures and sharing crypto contexts

MavenRain:
## Summary

Implements the Tier B test-suite speedup proposed in #894.  A batch of targeted
unit and property tests spent most of their wall time rebuilding
input-invariant fixtures inside hot loops.  This memoizes or hoists those
fixtures and replaces a few fixed wall-clock sleeps with event-driven waits,
cutting the targeted suite without changing what any test asserts and without
lowering any proptest case count or benchmark iteration count.

All changes are test-only; there is no production-code behavior change.

## Changes

**Memoize `CommitteeFixture` by committee size.** Building a `CommitteeFixture`
generates one BLS keypair per authority, which dominated these tests.  Each was
rebuilding the fixture inside every proptest case:

- `consensus/primary/tests/it/consensus_props.rs`: 6 `prop_*` fns at proptest's
  default 256 cases each rebuilt a fixture per case.  A file-local
  `committee_for_size` now caches `(Committee, Vec<AuthorityIdentifier>)` by size
  in a `LazyLock<Mutex<HashMap<..>>>`, so each of the ~11 distinct sizes builds
  once per test process.
- `types/tests/it/committee_props.rs`: `create_test_committee` now memoizes the
  `Committee` by size behind a `LazyLock<Mutex<HashMap<..>>>`;  the signature is
  unchanged so callers are untouched.
- `consensus/primary/src/tests/certificate_tests.rs`:
  `test_certificate_verification` (sizes 4..35) now memoizes the full
  `CommitteeFixture` per size in a `thread_local` cache (proptest runs every case
  on the single libtest thread here, so a `thread_local` is sound), since
  certificate signing needs the authority keypairs, not just the committee.

The generated-input range and case count of every property test are unchanged.
The only behavioral difference is that key material is reused across cases for a
given size; no test asserts on per-case key identity, so this preserves coverage.

**Share one secp256k1 context in `TransactionFactory`**
(`tn-reth/src/test_utils.rs`).  The three factory constructors each built a fresh
`Secp256k1::new()`, which precomputes the EC multiplication tables.  They now use
the crate's global `SECP256K1` context (the `global-context` feature is already
enabled on the workspace dependency), which is built once lazily and yields
byte-identical keys and signatures.  The seeded constructor still produces the
same documented address.

**Replace fixed sleeps with event-driven waits.**

- `storage/src/consensus_pack.rs`: the fixed 2s sleep after `stream_import` is
  replaced with a bounded `wait_for` helper that polls until the last imported
  output is readable (the exact condition the following assertion checks),
  capped at 10s so a regression still fails loudly.
- `types/src/task_manager.rs`: the fixed 1s post-drop sleep is replaced with a
  bounded poll (capped 5s) on the killed tasks' ping channels closing.
- `consensus/primary/src/tests/certifier_tests.rs`: `propose_header_failure` and
  `propose_header_scenario_with_bad_sigs` gain `start_paused = true` on their
  `#[tokio::test]` attribute.  These tests' 5s negative-path timeouts are pure
  dead waits proving no certificate arrives over in-memory channels;  under paused
  time tokio auto-advances the clock once the runtime is idle, so the windows
  elapse instantly with identical semantics.  Two same-crate tests already use
  this pattern.

## Not done (documented dead ends from #894)

`pipeline_tel_precompile_props` (the largest bucket) and `test_archive_hdx_index`
have no lever that preserves coverage;  both only shrink by cutting case/element
volume, which is out of scope here.  See #894 for the full analysis.

## Verification

All targeted tests pass, and the same tests were timed on the same machine
before and after the change (`cargo nextest run`, warm cache, same
`-E` filter, run as an isolated subset).  Per-test durations below.

`tn-primary` (fixture memoization + `start_paused`):

| Test | Before | After |
|---|---|---|
| `certifier_tests::propose_header_failure` | 5.039s | 0.031s |
| `certifier_tests::propose_header_scenario_with_bad_sigs` | 10.137s | 0.208s |
| `certificate_tests::test_certificate_verification` | 10.457s | 2.659s |
| `consensus_props::prop_leader_deterministic` | 5.241s | 0.243s |
| `consensus_props::prop_round_robin_covers_all` | 4.503s | 0.206s |
| `consensus_props::prop_no_consecutive_leaders` | 4.496s | 0.265s |
| `consensus_props::prop_leader_schedule_cycles` | 4.107s | 0.190s |
| `consensus_props::prop_leader_is_committee_member` | 3.987s | 0.118s |
| `consensus_props::prop_swap_deterministic` | 3.349s | 0.143s |
| `consensus_props::test_bad_node_gets_swapped` | 0.026s | 0.028s |
| **wall time (2 binaries)** | **27.01s** | **2.67s** |

`tn-types` / `tn-storage` (committee memoization + shared secp + sleep removal):

| Test | Before | After |
|---|---|---|
| `consensus_pack::test_consensus_pack` | 7.922s | 5.307s |
| `task_manager::test_task_manager` | 1.021s | 0.031s |
| `committee_props::prop_two_quorums_intersect` | 1.840s | 0.271s |
| `committee_props::prop_validity_threshold_greater_than_one_third` | 1.687s | 0.270s |
| `committee_props::prop_quorum_threshold_greater_than_two_thirds` | 1.669s | 0.265s |
| `committee_props::prop_quorum_greater_than_validity` | 1.484s | 0.270s |
| `committee_props::prop_byzantine_fault_tolerance` | 1.303s | 0.254s |
| `consensus_pack::test_pack_save_wrong_epoch_rejected` | 0.163s | 0.151s |

Across both groups the affected-test wall time drops from ~37.1s to ~8.0s
(~78%) on this machine.

Note on the absolute numbers: these per-test times are lower than the
full-suite figures in #894 because that baseline ran the whole ~56-test set in
parallel, where CPU contention inflates each test roughly 3x.  Running the
affected tests as an isolated subset here understates the absolute time removed
but cleanly isolates each change's effect;  on the contended full suite the
wall-clock gain is larger.

- `cargo nextest run` of all affected tests: 23 passed, 0 failed (both groups).
- `cargo clippy -p tn-primary -p tn-types -p tn-storage -p tn-reth --all-targets`:
  0 errors, and 0 new warnings (the pre-existing workspace warnings are all at
  untouched lines).
- `cargo +nightly fmt`: no changes.

### DIFF
--- a/crates/consensus/primary/src/tests/certificate_tests.rs
+++ b/crates/consensus/primary/src/tests/certificate_tests.rs
@@ -1,7 +1,7 @@
 //! Certificate tests
 
 use rand::{rngs::StdRng, SeedableRng};
-use std::num::NonZeroUsize;
+use std::{cell::RefCell, collections::HashMap, num::NonZeroUsize, rc::Rc};
 use tn_storage::mem_db::MemDatabase;
 use tn_test_utils_committee::CommitteeFixture;
 use tn_types::{AuthorityIdentifier, BlsKeypair, Certificate, SignatureVerificationState, Vote};
@@ -113,14 +113,29 @@ fn test_unknown_signature_in_certificate() {
     assert!(Certificate::new_unverified(&committee, header, signatures).is_err());
 }
 
+// Proptest runs every case for this test on the single libtest thread (no fork is
+// configured), so a thread_local cache is sound. Memoizing one CommitteeFixture per
+// committee size avoids regenerating BLS keypairs for every case; reusing the same keys
+// across cases changes neither the generated input range nor any assertion.
+thread_local! {
+    static FIXTURE_CACHE: RefCell<HashMap<usize, Rc<CommitteeFixture<MemDatabase>>>> =
+        RefCell::new(HashMap::new());
+}
+
 proptest::proptest! {
     #[test]
     fn test_certificate_verification(
         committee_size in 4..35_usize
     ) {
-        let fixture = CommitteeFixture::builder(MemDatabase::default)
-            .committee_size(NonZeroUsize::new(committee_size).unwrap())
-            .build();
+        let fixture = FIXTURE_CACHE.with(|cache| {
+            Rc::clone(cache.borrow_mut().entry(committee_size).or_insert_with(|| {
+                Rc::new(
+                    CommitteeFixture::builder(MemDatabase::default)
+                        .committee_size(NonZeroUsize::new(committee_size).unwrap())
+                        .build(),
+                )
+            }))
+        });
         let committee = fixture.committee();
         let header = fixture.header_from_last_authority();
 
--- a/crates/consensus/primary/src/tests/certifier_tests.rs
+++ b/crates/consensus/primary/src/tests/certifier_tests.rs
@@ -88,7 +88,7 @@ async fn propose_header_to_form_certificate() {
     ));
 }
 
-#[tokio::test(flavor = "current_thread")]
+#[tokio::test(flavor = "current_thread", start_paused = true)]
 async fn propose_header_failure() {
     let fixture = CommitteeFixture::builder(MemDatabase::default).randomize_ports(true).build();
     let committee = fixture.committee();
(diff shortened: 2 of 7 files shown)


sstanfield (approved):
LGTM!

PR #900, opened 2026-07-10
test(e2e): cut runtime of the four ignored e2e tests (#897)

### LINKED ISSUE #897: `test`: cut the four ignored e2e tests (epoch boundary/sync, delayed restarts) 25-50% via de-contention, event-driven restart waits, and a decoupled epoch-duration cut (no coverage loss)
## Summary

The four heaviest `#[ignore]`d e2e integration tests in `crates/e2e-tests`
(`epochs::test_epoch_boundary`, `epochs::test_epoch_sync`,
`restarts::test_restarts_delayed`, `restarts::test_restarts_lagged_delayed`)
spend most of their wall time on things that are not the behavior under test:
CPU contention from co-scheduling, fixed wall-clock `sleep`s that outlast the
consensus condition they are waiting for, and retry budgets spent re-confirming
nodes that are already dead.  Removing those cuts each test by **25 to 50 percent
without changing what any test asserts, without lowering any epoch count,
iteration count, or assertion, and without weakening the restart follow-path
coverage.**

This is disjoint from #894, which targets unit/integration tests in `tn-primary`,
`tn-types`, `tn-storage`, `tn-reth`, `tn-network-libp2p`, `tn-worker`, and
`tn-worker-gateway`.  None of the four e2e tests here are in that scope.  Finding 2,
Finding 5, and part of Finding 6 apply the same "replace a fixed sleep with an
event-driven wait" technique #894 mentions, but to a different set of tests.

`cargo nextest run -p e2e-tests --run-ignored`, warm cache, on branch
`bump-e2e-slow-timeout-720s` (HEAD `3ef12ce3`, which is `main` plus the CI
slow-timeout bump; the four test files are untouched by that commit, so line
numbers below should still be re-confirmed against `main` before the PR).

| Test | Reported (contended) | Solo | ~seconds | Dominant cost |
|---|---|---|---|---|
| `epochs::test_epoch_sync` | 527s | ~165s | 15 epoch boundaries at 10s + a mid-test kill/restart | epoch-duration cadence + contention |
| `epochs::test_epoch_boundary` | 431s | ~112s | ~13 epoch boundaries at 10s (early-break loop) | epoch-duration cadence + contention |
| `restarts::test_restarts_lagged_delayed` | 135s | n/a | 70s down-delay + 5s settle + dead-node probes | fixed sleeps |
| `restarts::test_restarts_delayed` | 131s | n/a | 70s down-delay + dead-node probes | fixed sleeps |

Note on contention: the reported per-test numbers are inflated roughly 3.5 to 4x
by CPU contention when the e2e group co-schedules two of these at once.  Prior
isolated measurement put `test_epoch_boundary` at ~112s (vs 431s reported) and
`test_epoch_sync` at ~165s (vs 527s reported).  So relieving contention and the
CPU-heavy epoch cadence also frees cores for the rest, and the real wall-clock
improvement exceeds the per-test arithmetic below.

## Root cause 1: the four tests co-schedule despite being marked "run independently"

All four carry `#[ignore]` attributes that say they must run alone:

- `crates/e2e-tests/tests/it/epochs.rs:404,442`: `#[ignore = "only run independently from all other it tests"]`
- `crates/e2e-tests/tests/it/restarts.rs:403,412`: `#[ignore = "should not run with a default cargo test, run restart tests as seperate step"]`

But the nextest e2e group still allows two at a time:

- `.config/nextest.toml:19-30`: `[test-groups.e2e] max-threads = 2`, applied to `package(e2e-tests)`.
- `crates/e2e-tests/tests/it/common.rs:36`: `const MAX_CONCURRENT_TESTS: usize = 2;` (the in-test semaphore, `:38`).

Each of these tests spawns four to six node processes.  Two running at once is
twelve-plus node processes contending for cores, which is the 3.5 to 4x inflation
above.  Giving these ignored tests a dedicated `max-threads = 1` test-group, or a
separate serial nextest invocation, drops `test_epoch_boundary` from 431s toward
~112s and `test_epoch_sync` from 527s toward ~165s, and takes 15 to 25 percent
off each restart test.  This is a scheduling change, not a reduction in CPU work,
and it is the isolation the `#[ignore]` messages already ask for.

## Root cause 2: the 70s restart down-delay is a fixed sleep ~1.7x the threshold it waits for

Both delayed restart tests keep a node down for a hard-coded 70 seconds:

- `crates/e2e-tests/tests/it/restarts.rs:406`: `do_restarts(70, false, "restarts_delayed")`
- `crates/e2e-tests/tests/it/restarts.rs:415`: `do_restarts(70, true, "restarts_lagged_delayed")`
- `restarts.rs:74` and `:142`: `std::thread::sleep(Duration::from_secs(delay_secs));`

The purpose (comment at `restarts.rs:400-401`) is to keep the node down long
enough that it does not rejoin consensus but instead follows the chain, as
opposed to `test_restartstt` (`restarts.rs:311-314`, delay 2s) which rejoins.
That threshold is a round count, not a wall-clock value: a restarted node is
demoted to the follow path (`NodeMode::CvvInactive`) only when the network has
advanced more than `gc_depth - 10 = 40` rounds past its recovered round
(`crates/consensus/primary/src/network/handler.rs:136,140,172`;  `gc_depth`
default 50 in `crates/config/src/node.rs`). The e2e testnet ticks headers every
500 to 1000ms (`crates/e2e-tests/src/lib.rs:117-120`, `max_header_delay=1000ms`),
so the ~41-round threshold is reached in ~41 seconds of downtime. 70s is roughly
a 1.7x margin of dead wall-clock.

Replace the fixed sleep with an event-driven wait: record a live node's block
height at kill time and poll sub-second until peers have advanced at least 45
rounds past it, then restart.  This pins the greater-than-40-round invariant
directly, is immune to any future header-cadence change, and terminates as soon
as the condition holds (~41 to 45s). Saves ~25 to 29s per restart test.

Because neither test asserts `NodeMode`, this is also the moment to add an
assertion that the restarted node actually entered `CvvInactive`, so the
follow-path coverage can never silently regress (see Not worth doing).

## Root cause 3: retries are spent re-confirming nodes that are already dead

Each restart test asserts a killed node is unreachable, but passes `retries = 5`
to the RPC helper, which sleeps one second between attempts:

- `crates/e2e-tests/tests/it/restarts.rs:77` (delayed) and `:145` (lagged): `get_balance(&client_urls[2], ..., 5).is_ok()` on the node just killed.
- `restarts.rs:289-292`: four more `get_balance(..., 5).is_err()` probes after `guard.kill_all()`.
- `crates/e2e-tests/tests/it/common.rs:246-254`: the retry loop does `tokio::time::sleep(Duration::from_secs(1))` per attempt.

That is five probes each spending ~5s retrying a negative assertion that cannot
change with retries (a dead node stays dead).  Passing `retries = 0` at these five
call sites leaves the `is_err()` / `is_ok()` assertions byte-for-byte unchanged
and removes ~25s per restart test.  Zero risk: connection-refused is instant and
deterministic.

## Root cause 4: epoch duration is 10s but the consensus floor is 5s (with one caveat)

`epoch_boundary = epoch_start + epochDuration` and closes on the first consensus
commit whose timestamp crosses it
(`crates/node/src/manager/node/run_epoch.rs:135,474-477`;
`committed_at` is whole-second, `crates/types/src/primary/output.rs:173-176`).
`EPOCH_DURATION` is fed to genesis via `--epoch-duration-in-secs`
(`epochs.rs:36,589-590`).  At the configured 500 to 1000ms header cadence a 10s
epoch spans ~10 to 20 rounds;  a 5s epoch still spans ~5 to 10 rounds, more than
enough to commit and close cleanly.  So `EPOCH_DURATION = 5` roughly halves the
per-epoch wait (~38 percent off `test_epoch_boundary`, ~48 percent off
`test_epoch_sync`), and every epoch count and assertion is preserved verbatim
(the 25-iteration loop, `epochs.rs:104`;  the 15 verified boundaries in
`loop_epochs`, `epochs.rs:320,331,339`).

The caveat that makes this safe: not every deadline in the tests scales with the
constant.  Two epoch-record poll deadlines are sized to clear a hard-coded ~62s
absolute timeout, not a multiple of epoch length:

- `epochs.rs:359-363`: `let deadline = Instant::now() + Duration::from_secs(EPOCH_DURATION * 6);`, with the comment "when a new validator joins the committee mid-test, its epoch vote quorum collection can time out (25 x 2.5s = ~62s) before the failed-quorum P2P fallback runs.  The spawn_epoch_record_collector retries every 5s independently, so 60s gives enough time for it to succeed."
- `epochs.rs:150`: the `EPOCH_DURATION * 3` deadline in the boundary record loop, which runs only after `shuffled == true` (`epochs.rs:142`), that is, exactly when a mid-test validator join can trigger the same 62s vote-collection window.

At `EPOCH_DURATION = 5` those become 30s and 15s, below the 62s timeout, which
reintroduces the flake the comment says was fixed (and it fires in the common
case, since `test_epoch_sync` shuffles a new validator over ~16 epochs).  So this
change is: set `EPOCH_DURATION = 5`, keep the parallel `PACK_IMPORT` copy in
`crates/e2e-tests/tests/it/sync.rs` in lockstep, and floor those two deadlines
absolutely, for example `Duration::from_secs((EPOCH_DURATION * 6).max(65))` and
`(EPOCH_DURATION * 3).max(35)`.  Apply after Finding 1, so the shorter `* 4`
epoch-change deadlines (`epochs.rs:106,267`) keep comfortable margin over
un-contended epoch turnaround.

## Root cause 5: each test process rebuilds the node binary

`TELCOIN_BINARY` is a per-process `OnceLock` (`crates/e2e-tests/src/lib.rs:22`),
and nextest runs one process per test, so all four tests re-invoke a `cargo`
build through escargot on startup
(`get_telcoin_network_binary`, `lib.rs:305-322`): ~3 to 10s of warm
metadata/freshness checking each, serialized across co-scheduled tests on cargo's
target-dir lock, and unbounded on a cold cache.  Build the binary once before the
nextest invocation (in `etc/test-and-attest.sh`) and have
`get_telcoin_network_binary` honor an env var such as `TN_BIN_PATH`, falling back
to escargot when unset.  Saves ~3 to 10s per test warm, minutes for whichever test
runs first cold, with no hazard beyond a stale pointer (avoided by building in the
same script that sets the var).

## Root cause 6: one-second polling and a fixed 5s settle in the restart path

- `crates/e2e-tests/tests/it/common.rs:246-254`: the shared RPC retry loop polls at 1s.  `get_balance` / balance-with-retry / block-null-retry helpers built on it wait in 1s steps against live nodes during the restart catch-up.  Dropping the interval to ~250ms recovers ~3 to 5s per restart test and ~1 to 2s in `test_epoch_sync`'s per-endpoint `get_block` verification, with no change to the retry count or the values asserted.
- `crates/e2e-tests/tests/it/restarts.rs:150-154`: the lagged variant sends a transaction and then `std::thread::sleep(Duration::from_millis(5000))`.  Replacing that unconditional 5s settle with an event wait for the nonce-1 transaction to be included on a live node saves ~3 to 4s in the lagged test.

## Not worth doing

These were checked and rejected:

- **A bare `70 -> 50s` cut of the restart down-delay.** The delay is the only
  thing separating these tests from `test_restartstt` (which rejoins), and neither
  test asserts `NodeMode`.  During downtime the surviving quorum is exactly three
  of four validators (2f+1, zero slack), so under contention a slowed round can
  push effective downtime below the ~41-round threshold: the node silently rejoins
  `CvvActive`, the test still passes, and the follow-path coverage is gone with no
  failure to signal it.  50s also straddles both the 41-round demotion edge and the
  50-round hard cert-reject edge (`gc_round = committed - gc_depth`,
  `crates/consensus/primary/src/consensus/state.rs`), the racy region the 10-round
  buffer exists to avoid.  Finding 2 (measure the round delta directly) is the safe
  way to get the same time back.
- **Dropping the epoch-change poll interval from 1s.** The ~1s of slop per epoch
  overlaps the 10s epoch cadence the loop is already blocked on, so it saves ~0s.
  The `EPOCH_DURATION` cut in Finding 4 is where the epoch-loop time actually is.
- **Deleting the `sleep(Duration::from_secs(2))` "advancing" pauses**
  (`restarts.rs:34,126,183`).  These are load-bearing: the comment states they
  exist so that a lagging node fails the immediately following check rather than
  passing spuriously.

## Acceptance

- Every one of the four tests still passes, asserting the same conditions.
- No `EPOCH_DURATION`-independent epoch count, `for` iteration count, or assertion
  is changed (coverage preserved).
- The restart follow-not-rejoin path is still exercised, and ideally now asserted
  directly (`CvvInactive` entered) so it cannot silently regress.
- The targeted `cargo nextest run -p e2e-tests --run-ignored` set is at least 25
  percent faster on the same machine (each of the four individually, and in
  aggregate).

## Measurement method

`cargo nextest run -p e2e-tests --run-ignored` (process-per-test), warm cache,
branch `bump-e2e-slow-timeout-720s` (`3ef12ce3`).  Per-test durations are inflated
by mutual CPU contention (isolated `test_epoch_boundary` ~112s vs 431s
co-scheduled), so the wall-clock gain from Finding 1 compounds beyond the
per-test seconds shown, and Findings 2 to 6 are per-test CPU reductions that hold
regardless of scheduling.

MavenRain:
## Summary

Closes #897.  Test-only optimization of the four ignored e2e integration tests
(`test_epoch_boundary`, `test_epoch_sync`, `test_restarts_delayed`,
`test_restarts_lagged_delayed`) that cuts their wall time by removing
non-essential overhead.  **No test assertions, epoch counts, or iteration counts
are changed, and restart follow-path coverage is preserved.**  The shared-infra
changes (nextest de-contention, prebuilt binary, RPC backoff) also speed up the
other ignored restart/observer/sync e2e tests for free.

## Changes

**Finding 1 — CPU de-contention (`.config/nextest.toml`)**
- Every e2e test spawns 4-6 node processes, and the heavy `#[ignore]`d ones
  (restart, epoch, observer, basefee, metrics, sync) run for minutes.  Under
  `max-threads = 2` two of them co-scheduled, putting 12+ node processes on the
  box and inflating wall time ~3.5-4x.
- Capped the whole `e2e` test-group at `max-threads = 1` (both `default` and
  `ci` profiles), so no two e2e tests ever run concurrently.  A single group is
  deliberately chosen over a name-matched "heavy" subset: it also covers heavy
  tests added later (there are already `basefee`/`metrics` ones that a
  restart/epoch/observer name filter would miss) and, being one group, cannot
  overlap itself.  The 720s slow-timeout and ci `retries = 0` overrides are
  preserved.  `cargo nextest show-config test-groups` confirms `e2e (max threads = 1)`
  applies to every e2e test.

**Finding 2 — event-driven restart downtime (`restarts.rs`)**
- Replaced the fixed 70s post-kill sleep in `run_restart_tests1` /
  `run_restart_tests_lagged1` with `wait_for_downtime`, which polls the live
  validators' consensus chain and returns once a floor has elapsed **and** the
  peers have actually advanced.
- The delayed tests now pass `60` instead of `70`.  Demotion to the
  follow/catch-up path happens once a node falls outside the GC window
  (`crates/consensus/primary/src/network/handler.rs`:
  `effective_exec_round + (gc_depth - 10) < round`, i.e. > 40 rounds behind with
  the default `gc_depth = 50`).  At the tests' `--max-header-delay-ms 1000`
  cadence (~1 round/s) the 60s floor clears that ~40-round / ~40s threshold with
  margin, so the restarted node still exercises the follow path.  The floor is the
  actual guarantee (valid because `max_header_delay` bounds the round period and
  Finding 1 now runs the e2e group serially, so nothing else steals cores);  the
  peer-advancement check is a stall-guard so the wait can't end while the network
  is wedged, and a fail-safe cap prevents hangs.  (An earlier draft used a 50s
  floor; review flagged it as too tight under load, so it was raised to 60 —
  still 10s/test faster than the old 70s sleep.)

**Finding 3 — no retries on dead-node checks (`restarts.rs`)**
- The six balance checks that assert a *known-shut-down* node is unreachable
  used `retries = 5` (~5s of 1s-spaced retries each).  A dead node cannot recover,
  so these now use `retries = 0` and fail immediately.  `kill_child`/`kill_all`
  wait for process exit before the checks run, so the connection is already
  refused.

**Finding 4 — shorter epoch, floored deadlines (`epochs.rs`)**
- `EPOCH_DURATION` 10s -> 5s (the consensus minimum).  It feeds genesis via
  `--epoch-duration-in-secs`, so `assert_eq!(epochInfo.epochDuration, EPOCH_DURATION)`
  still holds and epoch counts are unchanged.
- The two `tn_epochRecord` certificate-availability polls are floored at 30s and
  60s absolute (`(EPOCH_DURATION * 3).max(30)`, `(EPOCH_DURATION * 6).max(60)`).
  Certificate production is a fixed async quorum-voting cost (~62s worst case
  when a new validator joins mid-test), so those deadlines must not shrink with
  the epoch cadence.  All other epoch-scaled deadlines (epoch-change polls,
  first-epoch offset, tx-confirm timeout) legitimately track the cadence and
  still hold at 5s.
- The separate `EPOCH_DURATION`/`PACK_IMPORT_EPOCH_DURATION` constants in
  `sync.rs` and `basefee.rs` are intentionally left at 10s (those tests are
  outside the four-test scope);  their comments are updated to note the
  deliberate divergence from `epochs.rs` rather than silently going stale.

**Finding 5 — build the node binary once (`lib.rs`, `Makefile`)**
- `get_telcoin_network_binary` now honors `TN_BIN_PATH`: when set it uses the
  prebuilt binary and runs no per-process cargo build (escargot otherwise costs
  ~3-10s of overhead per test process).  When unset it falls back to the existing
  escargot build, so `cargo nextest`/`cargo test` still work with no setup.
- Added a `build-e2e-bin` make target
  (`cargo build --bin telcoin-network --features tn-storage/test-utils`) and
  threaded `TN_BIN_PATH=$(E2E_BIN)` through `test-restarts`, `test-e2e`, and
  `public-tests`.

**Finding 6 — tighter polling (`common.rs`, `restarts.rs`)**
- RPC retry backoff in `call_rpc_inner` 1s -> 250ms (these retries only mask
  brief unavailability such as a node mid-restart).
- The lagged-restart transaction settle is now event-driven: poll a live peer
  until the transfer lands (`get_balance_above_with_retry`) instead of a blind
  5s sleep before restarting the lagged node.

## Acceptance criteria

- [x] All four tests keep identical assertions — only timing constants and the
      wait mechanism changed.
- [x] No epoch-count or iteration-count reductions (the `0..25` / `loop_epochs`
      counts are untouched; only `EPOCH_DURATION` changed).
- [x] Restart follow-path coverage preserved — the 60s downtime floor still
      exceeds the ~40-round GC-window demotion threshold at the test cadence.
- [ ] **≥25% aggregate speedup** — expected from the de-contention (heavy tests
      no longer contend), the epoch-duration halving (~38-48% per epoch test),
      the 70->60s and 5s->event-driven restart waits, and the build-once binary.
      Please confirm the aggregate number from a CI run / a full
      `make public-tests` on the reference machine — see the note below.

## Verification

- `cargo +nightly-2026-03-20 fmt --check` clean.
- `cargo check -p e2e-tests --tests` compiles clean (0 warnings; the whole `it`
  test binary, including `basefee`/`metrics`, builds against the `TestBinary`
  refactor).
- `cargo +nightly-2026-03-20 clippy -p e2e-tests --all-features -- -D warnings`
  passes (0 warnings) — this matches CI's `cargo clippy --all --all-features --
  -D warnings`, which gates the lib; `TestBinary` derives `Debug` so it does not
  trip the workspace `missing_debug_implementations` lint under `-D warnings`.
- `cargo build --bin telcoin-network --features tn-storage/test-utils` (the
  `build-e2e-bin` invocation) builds successfully → the prebuilt binary path is
  valid.
- `cargo nextest show-config test-groups` reports `e2e (max threads = 1)` applied
  to every e2e test;  `make -n test-restarts` / `test-e2e` confirm `build-e2e-bin`
  runs first and `TN_BIN_PATH` is threaded correctly.
- Reviewed with a multi-agent adversarial pass; its findings led to three
  hardening changes already folded in here: the nextest de-contention was made a
  single `max-threads = 1` group (the earlier name-matched filter missed the
  `basefee`/`metrics` heavy tests), `TestBinary` got `#[derive(Debug)]`, and
  `build-e2e-bin` pins `--target-dir` (so a set `CARGO_TARGET_DIR` can't make
  `TN_BIN_PATH` point at a missing file).
- The heavy e2e suite was not run end-to-end locally (multi-process, resource
  heavy);  the **≥25% aggregate-speedup** AC should be validated in CI or on the
  maintainer's machine via `make public-tests`.

### DIFF
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -18,8 +18,32 @@ use tn_types::{test_utils::CommandParser, Address, Genesis, GenesisAccount};
 use tracing::{error, info};
 // unused deps warnings
 
-/// Only compile main bin once for all tests.
-pub static TELCOIN_BINARY: OnceLock<CargoRun> = OnceLock::new();
+/// A telcoin-network binary for the e2e tests.
+///
+/// Building the node binary via escargot costs ~3-10s of cargo overhead per test process.
+/// Setting `TN_BIN_PATH` to a prebuilt binary (see `make build-e2e-bin`) skips that
+/// entirely; when it is unset we fall back to an escargot build so `cargo nextest` and
+/// `cargo test` still work with no extra setup.
+#[derive(Debug)]
+pub enum TestBinary {
+    /// Prebuilt binary located via the `TN_BIN_PATH` environment variable.
+    Prebuilt(PathBuf),
+    /// Binary built on demand by escargot.
+    Cargo(CargoRun),
+}
+
+impl TestBinary {
+    /// Build a [`std::process::Command`] that runs this binary.
+    pub fn command(&self) -> std::process::Command {
+        match self {
+            TestBinary::Prebuilt(path) => std::process::Command::new(path),
+            TestBinary::Cargo(run) => run.command(),
+        }
+    }
+}
+
+/// Resolve the main bin once for all tests.
+pub static TELCOIN_BINARY: OnceLock<TestBinary> = OnceLock::new();
 
 /// RPC endpoints for a single node across all transports.
 #[derive(Debug)]
@@ -352,23 +376,35 @@ pub fn setup_log_dir(
     test_dir
 }
 
-/// Helper to retrieve and build the main project binary.
-pub fn get_telcoin_network_binary() -> &'static CargoRun {
-    info!("building main binary for e2e tests");
+/// Retrieve the main project binary, resolving it once for the whole test process.
+///
+/// Honors `TN_BIN_PATH`: when set, the prebuilt binary at that path is used and no
+/// per-process cargo build runs. Otherwise the binary is built once via escargot.
+pub fn get_telcoin_network_binary() -> &'static TestBinary {
     TELCOIN_BINARY.get_or_init(|| {
+        if let Ok(prebuilt) = std::env::var("TN_BIN_PATH") {
+            let path = PathBuf::from(&prebuilt);
+            assert!(path.is_file(), "TN_BIN_PATH is set to {prebuilt:?} but no file exists there");
+            info!("using prebuilt telcoin-network binary from TN_BIN_PATH: {prebuilt}");
+            return TestBinary::Prebuilt(path);
+        }
+
+        info!("building main binary for e2e tests");
         let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR not set");
         let path = PathBuf::from(manifest_dir);
         let workspace_root =
             path.parent().and_then(|p| p.parent()).expect("Cannot find workspace root");
 
-        CargoBuild::new()
-            .bin("telcoin-network")
-            .features("tn-storage/test-utils")
-            .manifest_path(workspace_root.join("Cargo.toml"))
-            .target_dir(workspace_root.join("target"))
-            .current_target()
-            .run()
-            .expect("Failed to build telcoin-network binary")
+        TestBinary::Cargo(
+            CargoBuild::new()
+                .bin("telcoin-network")
+                .features("tn-storage/test-utils")
+                .manifest_path(workspace_root.join("Cargo.toml"))
+                .target_dir(workspace_root.join("target"))
+                .current_target()
+                .run()
+                .expect("Failed to build telcoin-network binary"),
+        )
     })
 }
 
--- a/crates/e2e-tests/tests/it/epochs.rs
+++ b/crates/e2e-tests/tests/it/epochs.rs
@@ -32,8 +32,12 @@ const INITIAL_STAKE_AMOUNT: &str = "1_000_000";
 const MIN_EPOCHS_TO_TEST: usize = 6;
 // Epoch init creates HDX index files per epoch (open_epoch_pack → new_epoch →
 // ConsensusPack::open_append). With test-utils, these are ~1.3MB each (vs ~130MB in prod).
-// 10s provides margin for parallel test execution and CI load variance.
-const EPOCH_DURATION: u64 = 10;
+// 5s is the consensus minimum epoch duration; halving it from 10s roughly halves the
+// wall time of each epoch test. The two `tn_epochRecord` certificate-availability polls
+// below are floored to an absolute minimum (`.max(..)`) rather than scaling with this
+// constant, because certificate production is a fixed async quorum-voting cost that does
+// not shrink with the epoch cadence.
+const EPOCH_DURATION: u64 = 5;
 
 async fn test_epoch_boundary_inner(
     genesis: Genesis,
@@ -147,7 +151,9 @@ async fn test_epoch_boundary_inner(
         for ep in endpoints {
             let provider = ProviderBuilder::new().connect_http(ep.http_url.parse()?);
             for epoch in 0..=latest_epoch {
-                let deadline = Instant::now() + Duration::from_secs(EPOCH_DURATION * 3);
+                // Certificate availability is a fixed async quorum-voting cost, so floor
+                // this deadline at 30s instead of letting it shrink with EPOCH_DURATION.
+                let deadline = Instant::now() + Duration::from_secs((EPOCH_DURATION * 3).max(30));
                 let (epoch_rec, cert) = loop {
                     match provider
                         .raw_request::<_, (EpochRecord, EpochCertificate)>(
@@ -406,11 +412,12 @@ async fn test_epoch_sync_inner(
                 .join("data");
             let pack_file_exists = std::fs::exists(file_test).unwrap_or_default();
             assert!(pack_file_exists, "Missing an epoch pack file for {val_name} on epoch {epoch}");
-            // Use 6× epoch duration: when a new validator joins the committee mid-test,
-            // its epoch vote quorum collection can time out (25 × 2.5s = ~62s) before the
-            // failed-quorum P2P fallback runs. The spawn_epoch_record_collector retries
-            // every 5s independently, so 60s gives enough time for it to succeed.
-            let deadline = Instant::now() + Duration::from_secs(EPOCH_DURATION * 6);
+            // When a new validator joins the committee mid-test, its epoch vote quorum
+            // collection can time out (25 × 2.5s = ~62s) before the failed-quorum P2P
+            // fallback runs. The spawn_epoch_record_collector retries every 5s
+            // independently, so 60s gives enough time for it to succeed. This is a fixed
+            // cost, so floor the deadline at 60s rather than scaling it with EPOCH_DURATION.
+            let deadline = Instant::now() + Duration::from_secs((EPOCH_DURATION * 6).max(60));
             let (epoch_rec, cert) = loop {
                 match provider
                     .raw_request::<_, (EpochRecord, EpochCertificate)>(
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -14,10 +14,13 @@ retries = 2
 fail-fast = true
 test-threads = "num-cpus"
 
-# E2E tests use dynamic ports and can run in parallel.
-# Limit concurrency since each test spawns 4-6 node processes.
+# E2E tests each spawn 4-6 node processes, and the heavy (ignored) ones run for minutes.
+# Running two at once puts 12+ node processes on the box and inflated wall time ~3.5-4x,
+# so cap the whole e2e group at one test at a time (was max-threads = 2). Using a single
+# group rather than a name-matched "heavy" subset also covers heavy tests added later
+# (basefee, metrics, ...) without enumerating them, and avoids two groups overlapping.
 [test-groups.e2e]
-max-threads = 2
+max-threads = 1
 
 [[profile.default.overrides]]
 filter = "package(e2e-tests)"
--- a/Makefile
+++ b/Makefile
@@ -107,13 +107,22 @@ test-cargo:
 test-faucet:
 	cargo nextest run --package telcoin-network --features faucet --test it ;
 
+# Build the node binary once so e2e test processes reuse it via TN_BIN_PATH instead of
+# rebuilding it through escargot (~3-10s of cargo overhead per test process).
+.PHONY: build-e2e-bin
+build-e2e-bin:
+	cargo build --bin telcoin-network --features tn-storage/test-utils --target-dir $(CURDIR)/target ;
+
+# Location of the binary built by build-e2e-bin, passed to the e2e tests via TN_BIN_PATH.
+E2E_BIN := $(CURDIR)/target/debug/telcoin-network
+
 # run restart integration tests
-test-restarts:
-	cargo nextest run --run-ignored all test_restarts ;
+test-restarts: build-e2e-bin
+	TN_BIN_PATH=$(E2E_BIN) cargo nextest run --run-ignored all test_restarts ;
 
 # run e2e tests
-test-e2e:
-	cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features ;
+test-e2e: build-e2e-bin
+	TN_BIN_PATH=$(E2E_BIN) cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features ;
 
 # run tests with coverage (using llvm-cov + nextest)
 coverage:
@@ -181,10 +190,10 @@ revert-submodule:
 	cd tn-contracts && git checkout $(COMMIT_SHA)
 
 # workspace tests that don't require faucet credentials
-public-tests:
-	cargo nextest run --workspace --exclude tn-faucet --no-fail-fast ;
-	cargo nextest run -p e2e-tests --test it --run-ignored all test_epoch ;
-	cargo nextest run --run-ignored all test_restarts ;
+public-tests: build-e2e-bin
+	TN_BIN_PATH=$(E2E_BIN) cargo nextest run --workspace --exclude tn-faucet --no-fail-fast ;
+	TN_BIN_PATH=$(E2E_BIN) cargo nextest run -p e2e-tests --test it --run-ignored all test_epoch ;
+	TN_BIN_PATH=$(E2E_BIN) cargo nextest run --run-ignored all test_restarts ;
 
 # local checks to ensure PR is ready
 pr:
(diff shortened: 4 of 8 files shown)


crates/e2e-tests/tests/it/epochs.rs
```diff
@@ -32,8 +32,12 @@ const INITIAL_STAKE_AMOUNT: &str = "1_000_000";
...
 // Epoch init creates HDX index files per epoch (open_epoch_pack → new_epoch →
 // ConsensusPack::open_append). With test-utils, these are ~1.3MB each (vs ~130MB in prod).
-// 10s provides margin for parallel test execution and CI load variance.
-const EPOCH_DURATION: u64 = 10;
+// 5s is the consensus minimum epoch duration; halving it from 10s roughly halves the
+// wall time of each epoch test. The two `tn_epochRecord` certificate-availability polls
+// below are floored to an absolute minimum (`.max(..)`) rather than scaling with this
+// constant, because certificate production is a fixed async quorum-voting cost that does
+// not shrink with the epoch cadence.
+const EPOCH_DURATION: u64 = 5;
```
sstanfield, line 40: Only concern is this is still long enough for the epoch tests.  I think it should be with appropriate 1s (?) block times but I assume you confirmed this.

sstanfield (approved):
Look good and good work cutting these times down.  One question is all.

crates/e2e-tests/src/lib.rs
```diff
@@ -18,8 +18,32 @@ use tn_types::{test_utils::CommandParser, Address, Genesis, GenesisAccount};
...
+/// Setting `TN_BIN_PATH` to a prebuilt binary (see `make build-e2e-bin`) skips that
+/// entirely; when it is unset we fall back to an escargot build so `cargo nextest` and
+/// `cargo test` still work with no extra setup.
+#[derive(Debug)]
+pub enum TestBinary {
+    /// Prebuilt binary located via the `TN_BIN_PATH` environment variable.
+    Prebuilt(PathBuf),
+    /// Binary built on demand by escargot.
+    Cargo(CargoRun),
+}
```
grantkee, line 33: Great idea

crates/e2e-tests/tests/it/epochs.rs
grantkee: @MavenRain we tried this in the past and it seemed to have issues. However, that was based on other timing-based expectations throughout the tests. I ran this branch 18 times successfully before I had a TCP connection error because my machine fell asleep. I'm okay with this change. Worst case scenario, we bump it back up

grantkee (approved):
LGTM! Thank you for taking the time to improve e2e tests. I really appreciate it

PR #902, opened 2026-07-11
fix(consensus): make the consensus-output batch-count bound symmetric across writer and reader

### LINKED ISSUE #896: Consensus output batch-count asymmetry: uncapped live execution/write vs. 1000-batch reconstruction cap makes a deep sub-DAG permanently undecodable (sync wedge + restart crash-loop)
## Summary

A `ConsensusOutput` is produced, executed, and written to the pack file with no bound on how many batches it may contain, but the code that reads a `ConsensusOutput` back caps it at `MAX_BATCHES_PER_OUTPUT` (1000 on non-test builds) and returns `PackError::TooManyBatches` past that.  The producing and reconstructing halves disagree, and nothing at consensus-commit time keeps a sub-DAG under the reader's limit.

If a single committed sub-DAG references more than 1000 unique batches, every live node executes it normally (the active path never checks the count), but no node can ever reconstruct that output.  Two things follow, both bad.  A node that restarts after committing such a block cannot replay it and fails fatally on every boot, because the block stays in its own pack file: an unrecoverable crash-loop.  Any observer or rejoining CVV that tries to sync that block decodes it, gets `TooManyBatches`, and retries forever without advancing past that height.  This needs no attacker: the trigger is an ordinary leader stall or partition under load, given the real mainnet parameters.  It is a consensus-availability / liveness defect, and because the divergence is between "executed live" and "cannot be read at all," it is deterministic rather than probabilistic.

## The defect

**Locations:** the reader cap and its enforcement at `crates/storage/src/consensus_pack.rs:1317-1318` and `:1360-1375`; the uncapped writer at `crates/storage/src/consensus_pack.rs:982-993`; the uncapped active build path at `crates/consensus/executor/src/subscriber.rs:456-468`.

**Reader is capped (1000, not adiri-gated)** (`consensus_pack.rs:1317-1322`):

```rust
#[cfg(not(test))]
const MAX_BATCHES_PER_OUTPUT: usize = 1_000;
/// Lowered in tests so the cap can be exercised cheaply; legitimate test outputs use only a
/// handful of batches, well under this.
#[cfg(test)]
const MAX_BATCHES_PER_OUTPUT: usize = 50;
```

Enforced inside `iter_to_output`, the sole decode path for a `ConsensusOutput` from pack bytes (`consensus_pack.rs:1366-1375`):

```rust
PackRecord::Batch(batch) => {
    // Bound how many batch records a (possibly hostile) stream can deliver before the
    // terminating Consensus record arrives.  ...
    // A legitimate ConsensusOutput references far fewer batches than this.
    batch_records += 1;
    if batch_records > MAX_BATCHES_PER_OUTPUT {
        return Err(PackError::TooManyBatches(MAX_BATCHES_PER_OUTPUT));
    }
```

**Writer imposes no such bound** (`consensus_pack.rs:982-993`), and its comment says so explicitly:

```rust
let mut batches = BTreeMap::new();
// We want to make sure batches are saved to the pack in a deterministic order, so
// collect them in a BTreeMap.  We probably don't actually need this but this
// means we do not impose any extra restrictions on consensus output.
for cert_batch in consensus.batches() {
    for batch in &cert_batch.batches {
        let digest = batch.digest();
        // Should not have duplicate batches across output.
        // Will work if we do but will save batches more than once in a pack.
        batches.insert(digest, batch.clone());
    }
}
```

**The active build path is also uncapped, and its safety note is stale** (`subscriber.rs:459-468`):

```rust
for header in sub_dag.headers() {
    for (digest, _) in header.payload().iter() {
        batch_set.insert(*digest);
        batch_digests.push_back(*digest);
    }
}

// SAFETY: 10-node committees * 6-round commit max * 5 batch max = 300 max batch digests
// possible 32bytes * 300 = 9.6 kb => well within 1MB max message size
let mut fetched_batches = self.fetch_batches_from_peers(batch_set).await?;
```

The "6-round commit max" figure is wrong (see reachability below), and even if it were right the count is never actually checked here or at write time.  So the writer accepts what the reader will later reject.

## Reachability (no attacker required)

The mainnet parameters make a >1000-batch sub-DAG reachable through normal fault conditions.  `chain-configs/mainnet/parameters.yaml` sets `max_header_num_of_batches: 10` and `gc_depth: 50` (matching the defaults in `crates/config/src/node.rs:290-291` and `:303-304`).

A single `CommittedSubDag` is the flattened causal history of one leader certificate.  `order_dag` (`crates/consensus/primary/src/consensus/utils.rs:10-22`) descends from the leader down to `gc_round + 1`, where `gc_round = leader.round().saturating_sub(state.gc_depth)`:

```rust
let gc_round = leader.round().saturating_sub(state.gc_depth);
...
if x.round() == gc_round + 1 {
    // Do not try to order parents of the certificate, since they have been GC'ed.
    continue;
}
```

So one sub-DAG can span up to `gc_depth` (50) rounds.  During a leader outage no intermediate leaders are elected while certificates keep accumulating; one later leader then commits a deep multi-round sub-DAG in a single output.  The unique batch count is on the order of (rounds spanned) x (committee authors) x (up to `max_header_num_of_batches` = 10 batches per header).  A stall of a few tens of rounds under sustained load crosses 1000.  In steady state a sub-DAG spans only one to a few rounds (tens of batches, safely under the cap), so the bug bites precisely during a partition or leader stall, which is the worst time to lose the ability to sync or restart nodes.  A faulty leader set can also force a deep commit deliberately.

## Impact

Once such a block is committed, live nodes execute it (the active path never fails on count), but reconstruction fails everywhere:

- **Producer restart crash-loop (unrecoverable).**  On restart, `EpochManager::replay_missed_consensus` (`crates/node/src/manager/node/start_epoch.rs:68`) re-reads each unexecuted output through the capped decoder and propagates the error with `?` (`start_epoch.rs:86-87`):

  ```rust
  let consensus_output =
      self.consensus_chain.get_consensus_output_current(consensus_header.number).await?;
  ```

  `get_consensus_output_current` (`crates/storage/src/consensus.rs:731`) reads from the local pack via `current_pack().get_consensus_output(...)`, which funnels into `iter_to_output`, so a node reads its own uncapped-written block back through the 1000-cap and gets `TooManyBatches`.  `run_epoch.rs` calls `self.replay_missed_consensus(..).await?`, so the epoch task returns `Err` and the node shuts down.  The oversized block is still in the pack, so every subsequent boot re-hits it: the validator can never come back up.

- **Observer / rejoining-CVV permanent wedge.**  Sync decodes each fetched output and, on any decode error, returns `ConsensusHeaderResult::Retry` (`crates/state-sync/src/consensus.rs:73-79`):

  ```rust
  let output = match consensus_chain.decode_consensus_output(epoch, bytes).await {
      Ok(output) => output,
      Err(e) => {
          // Includes the case where we do not yet hold this epoch's committee/pack.
          warn!(target: "tn::observer", ?e, ?number, "failed to decode consensus output, will retry");
          return ConsensusHeaderResult::Retry;
      }
  };
  ```

  On the wedge path the caller loop (`consensus.rs:385-408`) only ever hits the `Retry` arm, which has no exit: it increments a retry counter (capped at 5), sleeps up to 5 seconds, and loops again (`consensus.rs:401-406`).  The loop exits only via `Done` (`:400`) or, on the success path, a `Continue` that has advanced below `end_number` (`:391-392`), neither of which a permanent decode failure can reach.  Because the underlying block genuinely has more than 1000 batches, the decode fails identically forever, so the node never advances past that height.

- **Honest server penalized.**  A peer that serves the legitimate bytes is scored `Penalty::Severe`, since `PackError::TooManyBatches(_)` maps to `Penalty::Severe` in `consensus_chain_error_to_penalty` (`crates/consensus/primary/src/network/mod.rs:1123-1125`).

## Root cause

`MAX_BATCHES_PER_OUTPUT` is a reader-only guard whose comment assumes "a legitimate ConsensusOutput references far fewer batches than this," but no invariant at consensus-commit time enforces that assumption.  The write path and the active execution path apply no batch-count bound, so the network can produce a legitimate, fully-executed output that its own reconstruction layer will reject.  A cap that is not symmetric across producing and reconstructing cannot be a safety bound; it is a divergence source.

## Suggested fix

Make the batch-count bound a consensus-level invariant that the producer and the reader share:

- Preferred: bound the sub-DAG at commit time so an oversized `ConsensusOutput` is never produced.  This must be deterministic across all nodes (for example, split a deep commit into multiple bounded outputs, or reject/limit above a protocol constant derived from `gc_depth * committee_size * max_header_num_of_batches`), so writer and reader agree by construction.
- Or enforce the same `MAX_BATCHES_PER_OUTPUT` at write/execute time (`fetch_batches` in `subscriber.rs` and `save_consensus_output` in `consensus_pack.rs`) so a block that would be undecodable is never committed or executed.  Note this converts the condition into a commit-time failure, which is only acceptable if it is deterministic and handled as a protocol rule, not a local error.
- At minimum, raise the reader cap above the true maximum the active path can ever produce and correct the stale `subscriber.rs:466` "300 max" comment and the "far fewer batches than this" claim at `consensus_pack.rs:1370-1371`.

Whichever path is chosen, a reader-only cap cannot remain the only guard, because that asymmetry is the bug.

## Suggested test design

- A unit test in `crates/storage` that builds a `ConsensusOutput` with more than 1000 unique batches, writes it via `save_consensus_output`, then reads it back via the pack decoder, and asserts the round-trip succeeds (or, if the design chooses commit-time rejection, that both write and read reject it identically).  Today this round-trip fails only on the read side.
- An integration test that drives `order_dag` / the commit path through a multi-round leader stall under load to produce a deep sub-DAG, then exercises both restart replay (`replay_missed_consensus`) and observer sync (`decode_consensus_output`) on the resulting block, asserting neither wedges nor fatally errors.

## Acceptance criteria

- A `ConsensusOutput` that is executed live can always be reconstructed from the pack (or is deterministically never produced), on every node, so write and read never disagree on batch count.
- A node that commits a deep sub-DAG can restart and replay it without a fatal error, and an observer or rejoining CVV can sync past it.
- The `MAX_BATCHES_PER_OUTPUT` bound (if retained) is derived from and enforced against the actual maximum the commit path can produce, and the stale batch-count comments in `subscriber.rs` and `consensus_pack.rs` are corrected.

MavenRain:
## Summary

Fixes #896.  A `ConsensusOutput` was produced, executed, and written to the pack with no bound on how many batches it may contain, but the pack reader capped it at a fixed `MAX_BATCHES_PER_OUTPUT = 1000` and returned `PackError::TooManyBatches` past that.  The producing and reconstructing halves disagreed, and nothing at consensus-commit time kept a sub-DAG under the reader's limit.

Under real mainnet parameters (`gc_depth = 50`, `max_header_num_of_batches = 10`, committee = 5) a single committed sub-DAG can reference ~2500 batches during an ordinary leader stall, which every node executes live but no node can reconstruct.  The consequences (no attacker required):

- **Restart crash-loop.** `replay_missed_consensus` re-reads the node's own oversized block through the capped decoder, propagates `TooManyBatches` with `?`, and shuts the node down.  The block stays in the pack, so every boot re-hits it.
- **Observer / rejoining-CVV wedge.** Sync decodes the fetched output, gets `TooManyBatches`, and returns `Retry` forever — the catch-up loop has no permanent-failure exit, so the node never advances past that height.
- **Honest peers penalized.** A peer serving the legitimate bytes is scored `Penalty::Severe`.

## Root cause

`MAX_BATCHES_PER_OUTPUT` was a reader-only guard whose comment assumed "a legitimate ConsensusOutput references far fewer batches than this," but no invariant at commit time enforced that.  A cap that is not symmetric across producing and reconstructing is a divergence source, not a safety bound.

## Fix

Make the batch-count bound a shared consensus invariant that the producer and the reader agree on by construction.

- **Reader bound is derived, not guessed.** `iter_to_output` now bounds buffered batches at `committee.size() * MAX_GC_DEPTH * MAX_HEADER_NUM_OF_BATCHES`, using the committee it already holds.  That is exactly the maximum a legitimately committed output can reference (a sub-DAG spans at most `gc_depth` rounds, one certificate per authority per round, at most `max_header_num_of_batches` batches per header), so an honest deep sub-DAG is never rejected — while an unauthenticated peer flood is still bounded (the sub-DAG is only authenticated *after* decode, so the cap remains load-bearing against OOM).
- **Producer side is enforced against the same maximum.** New `MAX_GC_DEPTH` / `MAX_HEADER_NUM_OF_BATCHES` protocol constants in `tn-types`;  `Parameters::validate` rejects a node configured above them (called from `ConsensusConfig::new_with_committee`, so every startup and epoch transition is checked); and `Header::validate` rejects any inbound header referencing more than `MAX_HEADER_NUM_OF_BATCHES` batches, so a certified header can never push a committed sub-DAG past the reader bound.
- **Stale comments corrected** in `subscriber.rs` (the "10-node committees * 6-round commit max * 5 batch max = 300" note) and `consensus_pack.rs` (the "far fewer batches than this" claim).

## Changes

- `crates/types/src/primary/mod.rs` — add `MAX_GC_DEPTH` and `MAX_HEADER_NUM_OF_BATCHES` protocol constants.
- `crates/types/src/error.rs` — add `HeaderError::TooManyBatches`.
- `crates/types/src/primary/header.rs` — `Header::validate` rejects an over-ceiling payload.
- `crates/config/src/node.rs` — `gc_depth` / `max_header_num_of_batches` defaults reference the constants; add `Parameters::validate`; tests.
- `crates/config/src/consensus.rs` — validate parameters in `new_with_committee`.
- `crates/storage/src/consensus_pack.rs` — derive the reader bound (`max_batches_per_output`); correct the comment; update the cap test; add a >1000-batch round-trip regression test.
- `crates/consensus/executor/src/subscriber.rs` — correct the stale batch-count comment on the active build path.
- `crates/consensus/primary/src/error/network.rs` — map `HeaderError::TooManyBatches` to `Penalty::Fatal` (malformed header, like `TooManyParents`).
- `crates/consensus/primary/src/tests/proposer_tests.rs` — `Header::validate` ceiling test.

## Tests

- `tn-storage`: a `ConsensusOutput` referencing 1100 batches (over the old fixed cap, within the derived bound) now round-trips through `save_consensus_output` -> pack decode; an unauthenticated flood past the derived bound is still rejected with `TooManyBatches`.
- `tn-primary`: a header at the batch ceiling validates;  one batch over is rejected with `TooManyBatches`.
- `tn-config`: default parameters validate; `gc_depth` / `max_header_num_of_batches` over the ceiling are rejected.

## Notes on the design choices

- **Why not reject at commit/write time?** Rejecting an already-executed output at write time converts the liveness bug into a different one (a node that cannot persist a block it already executed still halts). The bound is enforced *before* commit, at header validation and config load, so an oversized output is never produced in the first place.
- **Why constants rather than persisting the values in the pack?** The reader already carries the committee;  deriving the bound from two protocol constants avoids any pack-format/serialization change while keeping writer and reader in agreement. `gc_depth` and `max_header_num_of_batches` are network-wide protocol parameters (nodes must agree on them or they would compute different sub-DAGs), so treating their ceilings as shared constants is sound; raising a ceiling is a deliberate, coordinated change (the doc comments call this out).

## Deployment note

The new `Header::validate` batch-count check is a consensus-rule tightening, so it should be rolled out the way any consensus change is — at an epoch/version boundary rather than piecemeal within a live epoch.  Concretely: `Header::validate` now rejects (as `Penalty::Fatal`) any header referencing more than `MAX_HEADER_NUM_OF_BATCHES` (10) batches.  Both `mainnet` and `testnet` already run `max_header_num_of_batches: 10`, so no header any honest node produces today exceeds the ceiling and there is no behavioural change for the current networks.  The only exposure is an operator who had raised `max_header_num_of_batches` above 10 (previously unvalidated): during a mixed-version window an upgraded node would reject in-GC-window headers that a non-upgraded peer still accepts.  Such a node also now fails `Parameters::validate` at startup, so the intended migration is to lower the parameter to `<= 10` before upgrading.  If preferred, the `Header::validate` enforcement can be split into a follow-up gated on a version/epoch boundary while landing the storage-side reader bound and config validation first.

### DIFF
--- a/crates/config/src/node.rs
+++ b/crates/config/src/node.rs
@@ -288,7 +288,7 @@ impl Parameters {
     }
 
     fn default_max_header_num_of_batches() -> usize {
-        10
+        tn_types::MAX_HEADER_NUM_OF_BATCHES
     }
 
     fn default_max_header_delay() -> Duration {
@@ -301,7 +301,7 @@ impl Parameters {
 
     /// The default gc depth for consensus.
     pub fn default_gc_depth() -> u32 {
-        50
+        tn_types::MAX_GC_DEPTH
     }
 
     fn default_sync_retry_delay() -> Duration {
@@ -371,6 +371,32 @@ impl Default for Parameters {
 }
 
 impl Parameters {
+    /// Validate protocol-critical parameters against the ceilings the rest of the protocol depends
+    /// on.
+    ///
+    /// `gc_depth` and `max_header_num_of_batches` together bound how many unique batches a single
+    /// committed [`ConsensusOutput`](tn_types::ConsensusOutput) can reference. The consensus-pack
+    /// reader derives its reconstruction bound from [`tn_types::MAX_GC_DEPTH`] and
+    /// [`tn_types::MAX_HEADER_NUM_OF_BATCHES`], so a node configured above those ceilings could
+    /// commit an output that no node (including itself, on restart) can later reconstruct.
+    /// Rejecting such a configuration at startup keeps the producing and reconstructing halves
+    /// in agreement.
+    pub fn validate(&self) -> eyre::Result<()> {
+        eyre::ensure!(
+            self.gc_depth <= tn_types::MAX_GC_DEPTH,
+            "gc_depth {} exceeds the protocol maximum {}",
+            self.gc_depth,
+            tn_types::MAX_GC_DEPTH,
+        );
+        eyre::ensure!(
+            self.max_header_num_of_batches <= tn_types::MAX_HEADER_NUM_OF_BATCHES,
+            "max_header_num_of_batches {} exceeds the protocol maximum {}",
+            self.max_header_num_of_batches,
+            tn_types::MAX_HEADER_NUM_OF_BATCHES,
+        );
+        Ok(())
+    }
+
     /// Tracing::info! for [Self].
     pub fn tracing(&self) {
         info!("Header number of batches threshold set to {}", self.header_num_of_batches_threshold);
@@ -382,5 +408,33 @@ impl Parameters {
         info!("Sync retry nodes set to {} nodes", self.sync_retry_nodes);
         info!("Max batch delay set to {} ms", self.max_batch_delay.as_millis());
         info!("Max concurrent requests set to {}", self.max_concurrent_requests);
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::Parameters;
+
+    #[test]
+    fn default_parameters_are_within_protocol_ceilings() {
+        Parameters::default().validate().expect("default parameters must validate");
+    }
+
+    #[test]
+    fn parameters_reject_gc_depth_over_ceiling() {
+        let params = Parameters { gc_depth: tn_types::MAX_GC_DEPTH + 1, ..Default::default() };
+        assert!(params.validate().is_err(), "gc_depth over the protocol ceiling must be rejected");
+    }
+
+    #[test]
+    fn parameters_reject_max_header_num_of_batches_over_ceiling() {
+        let params = Parameters {
+            max_header_num_of_batches: tn_types::MAX_HEADER_NUM_OF_BATCHES + 1,
+            ..Default::default()
+        };
+        assert!(
+            params.validate().is_err(),
+            "max_header_num_of_batches over the protocol ceiling must be rejected"
+        );
     }
 }
--- a/crates/storage/src/consensus_pack.rs
+++ b/crates/storage/src/consensus_pack.rs
@@ -20,7 +20,7 @@ use tn_types::{
     gas_accumulator::RewardsCounter, AuthorityIdentifier, Batch, BlockHash, BlockNumHash,
     BlsPublicKey, CertifiedBatch, CommittedSubDag, Committee, ConsensusHeader,
     ConsensusHeaderDigest, ConsensusNumHash, ConsensusOutput, Epoch, EpochRecord, Hash as _, Round,
-    B256,
+    B256, MAX_GC_DEPTH, MAX_HEADER_NUM_OF_BATCHES,
 };
 use tokio::{
     io::{AsyncRead, BufReader},
@@ -1311,15 +1311,21 @@ pub(crate) fn verify_epoch_meta(
     Ok(())
 }
 
-/// Upper bound on how many `Batch` records `iter_to_output` will buffer before the
-/// terminating `Consensus` record.  This caps memory use when reading a (possibly hostile)
-/// peer stream; a legitimate ConsensusOutput references far fewer batches than this.
-#[cfg(not(test))]
-const MAX_BATCHES_PER_OUTPUT: usize = 1_000;
-/// Lowered in tests so the cap can be exercised cheaply; legitimate test outputs use only a
-/// handful of batches, well under this.
-#[cfg(test)]
-const MAX_BATCHES_PER_OUTPUT: usize = 50;
+/// Upper bound on how many `Batch` records `iter_to_output` will buffer before the terminating
+/// `Consensus` record, derived from the committee that produced the output.
+///
+/// A single `CommittedSubDag` spans at most [`MAX_GC_DEPTH`] rounds (`order_dag` descends only to
+/// `gc_round + 1`), with at most one certificate per authority per round and at most
+/// [`MAX_HEADER_NUM_OF_BATCHES`] batches per header (the proposer self-limits and
+/// `Header::validate` rejects oversized inbound headers).  So a legitimately committed output
+/// references at most `committee.size() * MAX_GC_DEPTH * MAX_HEADER_NUM_OF_BATCHES` unique batches.
+/// Bounding the reader at exactly that maximum keeps the writer and reader in agreement by
+/// construction — every executed output can be reconstructed — while still capping an
+/// unauthenticated peer flood of `Batch` records (the sub-DAG is only authenticated *after* decode,
+/// and the per-record size cap does not bound their count).
+fn max_batches_per_output(committee: &Committee) -> usize {
+    committee.size().saturating_mul(MAX_GC_DEPTH as usize).saturating_mul(MAX_HEADER_NUM_OF_BATCHES)
+}
 
 /// Take an async stream of bytes that in pack file representation of ConsensusOutput and return the
 /// ConsensusOutput.
@@ -1358,6 +1364,7 @@ async fn iter_to_output<R: AsyncRead + Unpin>(
     let mut available_batches = HashMap::new();
     let mut referenced_batches = HashSet::new();
     let mut batch_records = 0_usize;
+    let max_batches = max_batches_per_output(committee);
     while let Some(record) = next(stream_iter, timeout).await? {
         match record {
             PackRecord::EpochMeta(_epoch_meta) => {
@@ -1367,11 +1374,13 @@ async fn iter_to_output<R: AsyncRead + Unpin>(
                 // Bound how many batch records a (possibly hostile) stream can deliver before the
                 // terminating Consensus record arrives.  Without this an `EpochMeta`/`Consensus`
                 // -less flood of Batch records would grow `available_batches` until OOM; the
-                // per-record size cap (MAX_RECORD_SIZE) only bounds individual records.  A
-                // legitimate ConsensusOutput references far fewer batches than this.
+                // per-record size cap (MAX_RECORD_SIZE) only bounds individual records.  The bound
+                // is the maximum a legitimately committed output for this committee
+                // can reference (see `max_batches_per_output`), so an honest deep
+                // sub-DAG is never rejected.
                 batch_records += 1;
-                if batch_records > MAX_BATCHES_PER_OUTPUT {
-                    return Err(PackError::TooManyBatches(MAX_BATCHES_PER_OUTPUT));
+                if batch_records > max_batches {
+                    return Err(PackError::TooManyBatches(max_batches));
                 }
                 let batch_digest = batch.digest();
                 available_batches.insert(batch_digest, batch);
@@ -1656,9 +1665,9 @@ pub(crate) mod test {
     use tn_reth::RethChainSpec;
     use tn_test_utils::CommitteeFixture;
     use tn_types::{
-        test_genesis, BlockHash, Certificate, CertifiedBatch, CommittedSubDag, Committee,
-        ConsensusHeader, ConsensusHeaderDigest, ConsensusOutput, EpochRecord, Hash, HeaderBuilder,
-        ReputationScores,
+        test_genesis, Batch, BlockHash, Certificate, CertifiedBatch, CommittedSubDag, Committee,
+        ConsensusHeader, ConsensusHeaderDigest, ConsensusOutput, EpochRecord, ExecHeader, Hash,
+        HeaderBuilder, ReputationScores,
     };
 
     use crate::{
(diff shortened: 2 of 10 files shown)


crates/storage/src/consensus_pack.rs
```diff
@@ -1311,15 +1311,21 @@ pub(crate) fn verify_epoch_meta(
...
+/// `gc_round + 1`), with at most one certificate per authority per round and at most
+/// [`MAX_HEADER_NUM_OF_BATCHES`] batches per header (the proposer self-limits and
+/// `Header::validate` rejects oversized inbound headers).  So a legitimately committed output
+/// references at most `committee.size() * MAX_GC_DEPTH * MAX_HEADER_NUM_OF_BATCHES` unique batches.
+/// Bounding the reader at exactly that maximum keeps the writer and reader in agreement by
+/// construction — every executed output can be reconstructed — while still capping an
+/// unauthenticated peer flood of `Batch` records (the sub-DAG is only authenticated *after* decode,
+/// and the per-record size cap does not bound their count).
+fn max_batches_per_output(committee: &Committee) -> usize {
+    committee.size().saturating_mul(MAX_GC_DEPTH as usize).saturating_mul(MAX_HEADER_NUM_OF_BATCHES)
```
sstanfield, line 1386: I think this calc is OK but this comment is misleading.  MAX_GC_DEPTH is NOT the max depth for a subdag.  I believe it is actually 6 in the worst case (every other round commits and should not be possible to miss three commits).  @grantkee can fact check me on this.  This is why the old const was 1000, that should have been plenty.  I think this PR is useful though, the read and write should be symmetrical.  I just don't think this comment that presents GC depth in an inaccurate way should remain but using it for the calc is probably OK to produce a plenty large enough value.

sstanfield (changes requested):
This is good but the comment presenting GC depth as the max depth for a sub dag is inaccurate and misleading and should be fixed.  Adding symmetry to this seems like a win otherwise.

crates/storage/src/consensus_pack.rs
MavenRain: It's a good catch.  You're right that MAX_GC_DEPTH isn't the sub-DAG depth: order_dag stops at gc_round + 1 and skips already-committed rounds, so a commit only reaches back a couple of rounds in practice (leaders are even-round only).  I reworded all three comments to present MAX_GC_DEPTH as the GC horizon and the batch bound as a deliberately loose over-estimate, and kept it in the calc as you suggested since it yields a safely large value.  Also updated the matching comment in subscriber.rs and the MAX_GC_DEPTH doc in tn-types so they don't repeat the same framing.

sstanfield (approved):

crates/config/src/node.rs
```diff
@@ -371,6 +371,32 @@ impl Default for Parameters {
...
+    /// on.
+    ///
+    /// `gc_depth` and `max_header_num_of_batches` together bound how many unique batches a single
+    /// committed [`ConsensusOutput`](tn_types::ConsensusOutput) can reference. The consensus-pack
+    /// reader derives its reconstruction bound from [`tn_types::MAX_GC_DEPTH`] and
+    /// [`tn_types::MAX_HEADER_NUM_OF_BATCHES`], so a node configured above those ceilings could
+    /// commit an output that no node (including itself, on restart) can later reconstruct.
+    /// Rejecting such a configuration at startup keeps the producing and reconstructing halves
+    /// in agreement.
+    pub fn validate(&self) -> eyre::Result<()> {
```
grantkee, line 384: nit: this could also validate min `gc_depth > 0` but not a blocker for this PR. We should probably move these to protocol constants as well, since in deviation between committee validators would cause issues

grantkee (approved):
LGTM - one comment for your consideration

crates/config/src/node.rs
MavenRain: Writing a protocol constants Github issue so I can follow-up in a subsequent PR

PR #904, opened 2026-07-11
Use validator RPC from the current committee

### LINKED ISSUE #903: Worker RPC gateway can route to rotated-out or non-validator peers
## Problem

The network command that backs the worker RPC gateway, `GetAllValidatorRpcs`, returns a snapshot of every peer in `PeerManager`'s `known_peers` cache that has advertised an RPC endpoint. That cache isn't scoped to the active validator set: `prune_known_peers` keeps an entry if it's pinned (operator bootstrap, trusted, or explicit peers) or if the peer sits in any of the three tracked committee slots — previous, current, or next.

A load balancer or gateway built on this snapshot to discover validators' worker RPC endpoints and route client traffic across them can therefore receive endpoints for validators that already rotated out of the committee, validators in the next committee that aren't active yet, and pinned peers that were never committee members. Routing live client traffic to any of these means requests can land on a node that isn't currently authorized to serve them, or miss the current committee members entirely if the snapshot is taken mid-rotation.

## Solution

The snapshot should scope strictly to the current committee slot, not the full `known_peers` cache. `PeerManager` should expose a method that filters to current-committee members only, backed by a new accessor on the committee-tracking type that exposes just that slot.

For current-committee members with no known RPC record yet, the same call should re-trigger the kad discovery path used at epoch start, so a caller that polls the snapshot periodically converges on the full current-committee set as records arrive, rather than silently omitting members whose records haven't resolved.

This changes what `GetAllValidatorRpcs` guarantees: not "every known validator with an advertised RPC," but "the current committee's advertised RPCs, discovered on a best-effort basis." Callers that need a complete set should poll rather than treat a single response as final.

Related to #820

grantkee:
# Scope validator RPC snapshot to the current committee

- Replaced `all_rpcs()` (all of `known_peers`) with `current_committee_rpcs()`, filtered through a new `AllPeers::current_committee()` accessor.
- The snapshot call re-triggers kad discovery for current-committee members with no known record, so later polls pick up their entries as discovery converges.
- Updated the doc comments on `NetworkHandle::get_all_validator_rpcs` and `NetworkCommand::GetAllValidatorRpcs` to describe the new scoping and discovery behavior.
- Added test coverage for committee scoping, exclusion of previous/next-only and non-committee members, and discovery triggering for missing current-committee records.

closes #903 
blocked by #820

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -935,7 +935,7 @@ where
                 send_or_log_error!(reply, rpc, "GetValidatorRpc");
             }
             NetworkCommand::GetAllValidatorRpcs { reply } => {
-                let rpcs = self.swarm.behaviour().peer_manager.all_rpcs();
+                let rpcs = self.swarm.behaviour_mut().peer_manager.current_committee_rpcs();
                 send_or_log_error!(reply, rpcs, "GetAllValidatorRpcs");
             }
             NetworkCommand::OpenStream { peer, kind, reply } => {
--- a/crates/network-libp2p/src/peers/all_peers.rs
+++ b/crates/network-libp2p/src/peers/all_peers.rs
@@ -1138,6 +1138,12 @@ impl AllPeers {
         self.apply_committee_membership(committee)
     }
 
+    /// The current committee slot, as set from authoritative epoch state by
+    /// [`Self::update_committees`].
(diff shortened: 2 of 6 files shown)


grantkee:
I will attest once the upstream PR is merged into main and this branch is rebased

MavenRain (approved):
Looks clean

PR #906, opened 2026-07-11
Harden known peers restore

### LINKED ISSUE #905: `known_peers` cache can be pinned past its committee bound and can regress to stale data
## Problem

The `known_peers` cache sits downstream of two data paths: the persisted kad store (restored at startup) and live kad queries (merged in during discovery). Both paths currently bypass guarantees that the cache is supposed to enforce.

On restart, persisted kad records are restored into `known_peers` through the same insertion path used for bootstrap peers, which pins the entry against the committee-rotation pruning introduced for issue #827. But the persisted store legitimately holds arbitrary signature-valid third-party records — a normal consequence of the node's DHT storage duty — not just peers this node has vetted. Since the store can hold up to `MaxRecords` entries, an attacker able to fill it with valid-but-arbitrary records can, purely by forcing or waiting for a restart, convert a large share of `known_peers` into permanently pinned entries. That defeats the committee bound #827 was meant to enforce.

Separately, `known_peers` is not timestamp-monotonic the way the kad store is. Query results reach the cache via `close_kad_query -> add_discovered_peer`, a path that never touches the store's `is_newer_record` staleness check. A stale-but-still-signature-valid record, served as the only response to a query, can overwrite a fresher cached entry — regressing the discovery cache to older peer info with no store-side protection to catch it.

Both gaps stem from the same root cause: `known_peers` is meant to be a validated read-model over kad, but two of its write paths don't carry forward the invariants (pinning scope, monotonicity) that the store enforces on its own writes.

## Solution

Startup restoration of persisted kad records should use a distinct, cache-only insertion path that does not pin — entries restored this way stay subject to committee-rotation pruning like any other discovered peer. The node's own record should be skipped during restore, and bootstrap peers should go through a separate always-pin path that never overwrites an existing richer record, preserving the current pinning guarantee for the peers it's actually meant to protect.

For monotonicity, `known_peers` writes reachable from kad query results should apply the same staleness rule the store already applies internally, rejecting a candidate record whose timestamp doesn't advance the cached one. This affects the discovered-peer path and the committee branch of self-advertised peer handling. Pinned (operator-provisioned) entries should remain exempt, since their timestamp reflects local provisioning rather than a peer-signed observation and isn't comparable on the same axis.

Together, these changes narrow `known_peers` to only pin entries whose trust is actually earned (committee membership, bootstrap configuration) and make its updates resistant to regression from stale records — bringing its write paths in line with the guarantees the underlying kad store already provides.

grantkee:
- Restore persisted kad records via a new cache-only `add_restored_peer` instead of `add_known_peer`, so restart no longer pins up to `MaxRecords` attacker-fillable records past the #827 committee bound. The node's own record is skipped during restore.
- Add a new `add_bootstrap_peer` for bootstrap peers specifically — always pins, never overwrites an existing richer record.
- Add a `kad_record_is_stale` guard mirroring the store's `is_newer_record` check, applied to `add_discovered_peer` and the committee branch of `add_self_advertised_peer`, so a stale record served as a query response can't regress a fresher cached entry. Pinned entries are exempt since their timestamp is a local provisioning stamp, not peer-signed.
- Document `known_peers` as the validated read-model of kad — why it's not redundant with the store (different lifecycles, post-validation values, sync hot-path lookup) and why `close_kad_query` deliberately never writes query results back to the store.

closes #905

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -434,9 +434,22 @@ where
             stream_protocols,
         );
 
-        // Promote the surviving records into the local peer cache.
+        // Promote the surviving records into the local peer cache. The store's contents are
+        // peer-fillable (arbitrary signature-valid third-party records held as DHT storage
+        // duty), so entries are restored UNPINNED and stay prunable at the first committee
+        // rotation. Our own record is skipped: both primary and worker key their record by
+        // the primary BLS key, and there is no point caching ourselves as a known peer.
+        let own_key = key_config.primary_public_key();
+        let mut restored: usize = 0;
         for (key, info) in known {
-            behavior.peer_manager.add_known_peer(key, info);
+            if key == own_key {
+                continue;
+            }
+            behavior.peer_manager.add_restored_peer(key, info);
+            restored += 1;
+        }
+        if restored > 0 {
+            info!(target: "network-kad", restored, "restored persisted kad records into the local peer cache");
         }
 
         let network_pubkey = keypair.public().into();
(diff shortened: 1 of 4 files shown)


grantkee:
Blocked by #904

MavenRain (dismissed):
LGTM

sstanfield (approved):
LGTM!

PR #908, opened 2026-07-12
feat(worker-gateway): edge protections — rate limits, size limit, tx screening (#712)

MavenRain:
> PR3 of the phased breakdown for #712.  Builds on the PR2 proxy core; adds the edge protections.  No compile dependency on any other PR . . . the gateway talks to workers over HTTP at runtime.

## What this adds

The three edge protections from the plan, all on the existing proxy path:

### 1. Rate limiting (per-IP + global)

Two token-bucket limiters shed load **before** a request is buffered or forwarded (the rate-limit layer is the outermost router layer, ahead of the body read):

- **Per-client-IP** (`--rate-limit-per-ip`, req/s, `--rate-limit-per-ip-burst`): stops any single source monopolizing the workers.
- **Global** (`--rate-limit-global` / `--rate-limit-global-burst`): caps aggregate throughput to roughly the workers' capacity.

Either is disabled with rate `0`;  a `0` burst derives 2×rate. An over-limit request gets a JSON-RPC `429` envelope (code `-32006`), never a bare reset.

Implementation is a small hand-rolled token bucket (std + tokio only: **no new dependency**), with an **injectable clock** so refill/GC behaviour is tested deterministically (mirrors the `BannedPeerCache` clock pattern in #834).  Per-IP state is bounded: idle buckets are reclaimed by a periodic sweep task (managed under the `TaskManager`, drains on SIGTERM) and the tracked-IP count is capped, so a wide source-IP spread cannot grow memory without limit.

The client identity is the **immediate TCP peer** (`ConnectInfo`, already injected per connection by PR2's accept loop). Documented caveat: run the gateway edge-facing; behind an untrusted L7 proxy the peer is that proxy, so trusting an inbound `X-Forwarded-For` would be a bypass . . . I did **not** trust it (see Q2).

### 2. Configurable request-size limit

`--max-request-bytes` (default 25 MiB, the previous hardcoded value) now feeds the `DefaultBodyLimit`;  oversized bodies still get the `-32003` "request too large" JSON-RPC error.

### 3. Shallow `eth_sendRawTransaction` screening

A single `eth_sendRawTransaction` call is decoded far enough to reject, at the edge, exactly what the worker would also reject, an **undecodable payload** (`-32007`) and an **EIP-4844 blob transaction** (`-32008`), front-running a wasted upstream round-trip. Key correctness property: the decode uses the same **pooled** wire format the worker's RPC accepts and **never recovers the signer**, so it is strictly more permissive than the worker and **cannot reject a transaction the worker would accept** (no false rejects).  A fast-path substring check means only bodies naming the method are parsed at all;  batches and every other method forward unchanged.

New JSON-RPC codes (gateway server-error range): `-32006` rate-limited (`429`), `-32007` invalid transaction (`400`), `-32008` unsupported transaction type (`400`).

## One-line cross-crate change

`crates/types/src/lib.rs` re-exports `alloy::eips::eip2718::Decodable2718` (alongside the already-exported `Encodable2718`) so the gateway decodes raw transactions version-aligned with reth's alloy, without a direct `alloy`/`tn-reth` dependency.

## Not in scope (deferred, per the plan)

- Prometheus metrics (incl. a rejections-by-reason counter and the in-flight HPA gauge), Dockerfile, k8s/HPA → **PR4**.
- TLS termination, auth/API keys, CORS, dynamic discovery: out of scope per the issue.

## Testing

**55 tests pass** (was 32 in PR2).  New coverage: rate-limiter unit tests (burst exhaustion, time-based refill, per-IP isolation, global cap, GC reclaims idle / spares busy buckets, full-table fail-open, both-disabled) driven by an injectable clock; an end-to-end `429`-envelope test through the real router;  transaction-screening tests (valid legacy tx forwarded, undecodable/non-hex/blob-typed rejected, other-method/batch/missing-params/non-JSON forwarded);  and CLI tests (defaults, `0`-disables, explicit burst, configurable size).

Gates: `cargo check`, `cargocho clippy --all-targets -D warnings` (0/0 on the gateway), `cargocho test` (56 pass), `cargo +nightly-2026-03-20 fmt --check` . . . all clean.  The binary's `--help` was smoke-tested (all new flags render with their defaults).

> Note: running clippy over `-p tn-types --all-targets` (which my one-line re-export touches) surfaces two **pre-existing** lints in `crates/types/src/{committee.rs, primary/epoch.rs}` . . . files this PR does not touch. They are on `origin/main` already, not introduced here.

Self-review: a 6-lens adversarial workflow (bounded ≤7 findings / ≤49 agents) over the diff surfaced only 3 low-severity items, all addressed . . . the rate-limit layer originally covered `/health` and `/ready` (now exempt, so an orchestrator's probes survive a flood; with a test), the limiter's control flow was moved onto combinators per the repo/style conventions, and this failure table gained the three new codes.  No correctness, security, or DoS finding survived verification.

### DIFF
--- a/bin/worker-gateway/src/ratelimit.rs
+++ b/bin/worker-gateway/src/ratelimit.rs
@@ -1,0 +1,452 @@
+//! Per-IP and global request rate limiting.
+//!
+//! Two token-bucket limiters shed load before a request reaches the proxy or
+//! has its body buffered: a gateway-wide bucket caps aggregate throughput to
+//! roughly what the upstream workers can absorb, and a per-client-IP bucket
+//! stops any single source monopolizing that budget. An over-limit request
+//! receives the gateway's JSON-RPC "rate limit exceeded" envelope (HTTP `429`),
+//! never a bare connection reset.
+//!
+//! The client identity is the immediate TCP peer address (`ConnectInfo`,
+//! injected per connection by the accept loop). The gateway is meant to run
+//! edge-facing; behind an untrusted L7 proxy the peer is that proxy, so the
+//! per-IP bucket would meter the proxy rather than the real client. Terminate
+//! such a proxy's client identity upstream, or run the gateway at the edge (see
+//! the crate README).
+
+use std::{
+    collections::HashMap,
+    net::{IpAddr, SocketAddr},
+    num::NonZeroU32,
+    sync::{Arc, Mutex, MutexGuard, PoisonError},
+    time::{Duration, Instant},
+};
+
+use axum::{
+    extract::{ConnectInfo, Request, State},
+    middleware::Next,
+    response::Response,
+};
+use tn_types::{Noticer, TaskError};
+use tokio::time::MissedTickBehavior;
+
+use crate::{
+    error::{error_response, GatewayError},
+    server::{HEALTH_PATH, READY_PATH},
+};
+
+/// How often the background sweep reclaims idle per-IP buckets.
+pub(crate) const GC_INTERVAL: Duration = Duration::from_secs(60);
+
+/// Default ceiling on tracked per-IP buckets. Bounds the per-IP map so a wide
+/// spread of source IPs cannot grow it without limit; ~10 MB at the cap, and
+/// idle buckets are reclaimed between sweeps by [`RateLimiters::gc`].
+pub(crate) const DEFAULT_MAX_PER_IP_ENTRIES: usize = 100_000;
+
+/// A source of monotonic time, injectable so the limiters can be driven
+/// deterministically in tests.
+pub(crate) trait Clock: Send + Sync + 'static {
+    fn now(&self) -> Instant;
+}
+
+/// The production clock: the process monotonic clock.
+#[derive(Debug, Clone, Copy, Default)]
+pub(crate) struct SystemClock;
+
+impl Clock for SystemClock {
+    fn now(&self) -> Instant {
+        Instant::now()
+    }
+}
+
+/// A resolved rate limit: a sustained `rate` (requests per second) with a
+/// `burst` ceiling (the token-bucket capacity).
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct RateLimit {
+    rate: NonZeroU32,
+    burst: NonZeroU32,
+}
+
+impl RateLimit {
+    pub(crate) fn new(rate: NonZeroU32, burst: NonZeroU32) -> Self {
+        Self { rate, burst }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn rate(&self) -> NonZeroU32 {
+        self.rate
+    }
+
+    #[cfg(test)]
+    pub(crate) fn burst(&self) -> NonZeroU32 {
+        self.burst
+    }
+
+    /// Sustained refill rate in tokens per second.
+    fn tokens_per_sec(&self) -> f64 {
+        f64::from(self.rate.get())
+    }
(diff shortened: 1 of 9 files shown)


sstanfield (approved):

bin/worker-gateway/src/ratelimit.rs
```diff
@@ -0,0 +1,452 @@
...
+        Instant::now()
+    }
+}
+
+/// A resolved rate limit: a sustained `rate` (requests per second) with a
+/// `burst` ceiling (the token-bucket capacity).
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct RateLimit {
+    rate: NonZeroU32,
+    burst: NonZeroU32,
```
grantkee, line 67: Please include doc comments for all struct fields. It may seem obvious, but if nothing else, it confirms assumptions quickly for maintainers, security researchers, and LLMs

bin/worker-gateway/src/ratelimit.rs
```diff
@@ -0,0 +1,452 @@
...
+        f64::from(self.burst.get())
+    }
+}
+
+/// A refilling token bucket. It starts full and refills at `rate` tokens per
+/// second up to `capacity`; each admitted request spends one token.
+#[derive(Debug)]
+struct Bucket {
+    tokens: f64,
+    last_refill: Instant,
```
grantkee, line 101: Flagging for doc comments

bin/worker-gateway/src/ratelimit.rs
```diff
@@ -0,0 +1,452 @@
...
+    fn is_idle(&self, now: Instant, rate: f64, capacity: f64) -> bool {
+        self.replenished(now, rate, capacity) >= capacity
+    }
+}
+
+/// The single gateway-wide bucket.
+#[derive(Debug)]
+struct GlobalLimiter {
+    limit: RateLimit,
+    bucket: Mutex<Bucket>,
```
grantkee, line 141: Doc comments

bin/worker-gateway/src/ratelimit.rs
```diff
@@ -0,0 +1,452 @@
...
+}
+
+/// Per-client-IP buckets, bounded in cardinality; idle buckets are reclaimed by
+/// [`RateLimiters::gc`].
+#[derive(Debug)]
+struct PerIpLimiter {
+    limit: RateLimit,
+    max_entries: usize,
+    buckets: Mutex<HashMap<IpAddr, Bucket>>,
```
grantkee, line 150: Doc comments

bin/worker-gateway/src/ratelimit.rs
```diff
@@ -0,0 +1,452 @@
...
+        // Table full: admit untracked (bounded memory over per-IP fairness).
+        true
+    }
+}
+
+/// The gateway's rate limiters. Either limiter may be disabled (`None`).
+pub(crate) struct RateLimiters<C: Clock = SystemClock> {
+    clock: C,
+    global: Option<GlobalLimiter>,
+    per_ip: Option<PerIpLimiter>,
```
grantkee, line 198: doc comments

grantkee (approved):
LGTM! I flagged some struct fields that were missing doc comments. I'd like us to include doc comments for all fields, types, structs, mods, etc. to maximize documentation as a best practice. Not a blocker for this PR

MavenRain:
Filing Github issues to follow-up with the doc comments

PR #909, opened 2026-07-13
feat(rpc): thread deadline through delegationDigest binding

Huwonk:
## Context

`ConsensusRegistry.delegationDigest` in tn-contracts gains a `uint256 deadline` argument (delegation signatures now carry an expiry; see tn-contracts #134). That changes the function selector, so the node's hand-written binding must match or `tn_delegationDigest` reverts at runtime.

## Change

- `system_calls.rs`: add `uint256 deadline` to the `delegationDigest` `sol!` binding.
- `rpc_ext.rs`: thread `deadline: U256` through the `tn_delegationDigest` trait method and its impl (`delegationDigestCall { ..., deadline }`). This adds a parameter to the public `tn_delegationDigest` JSON-RPC method.
- Add a unit test pinning the binding to the four-argument selector `delegationDigest(bytes,address,address,uint256)`. The RPC was previously untested, so a drifted binding would fail silently at runtime; this test fails at compile/test time instead.

## Verification

The binding and the selector test are self-contained (independent of the submodule artifact), so they compile and the unit test runs without a submodule bump - CI covers them here.

Not yet done, blocking full resolution:
- Bump the tn-contracts submodule to the merged commit from tn-contracts #134, so the deployed/genesis registry actually exposes the four-argument selector.
- Local build could not be run on Windows (pre-existing unrelated `ctrl_c` build failure in `tn-types`); relying on CI / WSL for confirmation.

Draft until tn-contracts #134 merges and the submodule is bumped.

### DIFF
--- a/crates/execution/tn-rpc/src/rpc_ext.rs
+++ b/crates/execution/tn-rpc/src/rpc_ext.rs
@@ -134,6 +134,7 @@ pub trait TelcoinNetworkRpcExtApi {
         bls_pubkey: Bytes,
         validator_address: Address,
         delegator: Address,
+        deadline: U256,
     ) -> TelcoinNetworkRpcResult<B256>;
     /// Return the BLS12-381 proof of possession message: `blsPubkey || validatorAddress`.
     ///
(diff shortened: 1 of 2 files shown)


Huwonk:
Merged into #1012 (feat/unified-conclude-epoch) so the delegationDigest deadline binding rides with the tn-contracts dc248c0 pin it depends on.

PR #911, opened 2026-07-13
Limit reth MDBX sizes for temp-chain tests

### LINKED ISSUE #910: `cargo test` fails with MDBX ENOMEM due to unbounded temp-chain map sizes
## Problem

`RethEnv::new_for_temp_chain` builds throwaway MDBX environments for tests without bounding the database geometry, so each environment inherits reth's default 8 TB map size. `cargo test` runs an entire test binary as threads within a single process, so every temp chain created during a run reserves its own 8 TB virtual address range in that one process's address space. Once enough temp chains are created concurrently, the cumulative virtual reservations exhaust the process's address space and MDBX aborts environment creation with ENOMEM ("Cannot allocate memory (12)"), causing `cargo test` to fail outright.

This does not affect `cargo nextest`, which spawns one process per test — each process only ever opens a small number of environments, so the 8 TB reservations never accumulate to the point of exhaustion. That difference in process model masked the bug in `make test` while leaving plain `cargo test` broken.

A throwaway temp chain used in tests never needs to hold a full node's history, making the 8 TB default both unnecessary and actively harmful in this context.

## Solution

`RethEnv::new_for_temp_chain` should cap the MDBX map size for temp-chain environments to a small ceiling sufficient for test workloads, and shrink the growth step accordingly so throwaway databases stay small on disk as well as in virtual memory. This keeps many concurrent temp-chain environments viable within a single process's address space, fixing `cargo test` without changing behavior for `cargo nextest` or production nodes, which are unaffected since they don't go through this constructor.

grantkee:
- Bound `RethEnv::new_for_temp_chain`'s MDBX geometry to a 2 GiB `max_size` / 4 MiB `growth_step` instead of reth's 8 TB default, since `cargo test` runs all tests as threads in one process and enough concurrent 8 TB virtual reservations exhaust address space, aborting env-open with ENOMEM. `cargo nextest` (one process per test) never hit this, which is why `make test` passed while `cargo test` didn't.
- Removed an unused `tokio::io::AsyncReadExt` import in `sync_codec.rs` tests.

closes #910

### DIFF
--- a/crates/consensus/primary/src/network/sync_codec.rs
+++ b/crates/consensus/primary/src/network/sync_codec.rs
@@ -554,7 +554,6 @@ where
 mod tests {
     use super::*;
     use tn_network_libp2p::SyncFrameError;
-    use tokio::io::AsyncReadExt as _;
 
     /// Frame `bytes` through [`write_pack_data_frames`] into an in-memory buffer.
     async fn frame_pack(bytes: &[u8]) -> Vec<u8> {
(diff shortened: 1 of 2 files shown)


MavenRain (approved):
GTG . . . nice find!

sstanfield (approved):

PR #920, opened 2026-07-14
test(e2e): cut base-fee e2e runtime via a decoupled epoch-duration cut (#907)

### LINKED ISSUE #907: `test`: cut the three ignored base-fee e2e tests ~40-50% via a decoupled epoch-duration cut (no coverage loss)
## Summary

The three `#[ignore]`d base-fee e2e tests in `crates/e2e-tests/tests/it/basefee.rs`
(`test_static_fee_applied_at_epoch_boundary`,
`test_eip1559_fee_rises_at_epoch_boundaries`,
`test_mid_epoch_restart_recovers_static_fee`) spend most of their wall time
crossing epoch boundaries at a fixed 10-second cadence and waiting out fixed
`sleep`s that outlast the consensus condition they gate.  Halving the configured
epoch duration to the 5-second consensus minimum (the same decoupled cut #897
applied to `epochs.rs`) roughly halves the epoch-boundary portion of each test
**without changing any assertion, any epoch/iteration count, or the fee oracle
being verified.**

It is disjoint from #897, which explicitly left `basefee.rs` at `EPOCH_DURATION = 10`
as out of its four-test scope, and from #894 (unit/integration tests in other
crates).  It reuses #897's two techniques, a decoupled epoch-duration cut with
absolutely-floored deadlines, and replacing a fixed sleep with an event-driven
wait, on the base-fee tests.

Measured from the per-node logs under `crates/e2e-tests/test_logs/basefee_*/`
(first-to-last node timestamp = the window in which the testnet is actually
doing work), on `main` with the four node processes prebuilt (see the note
below on why the *reported* nextest numbers are much larger).

| Test | Active network window | Epoch boundaries crossed | Dominant cost |
|---|---|---|---|
| `test_eip1559_fee_rises_at_epoch_boundaries` | ~42s | ~5 (`target_boundaries = 3`, starting epoch 1) | one tx + one boundary wait per epoch, at 10s |
| `test_mid_epoch_restart_recovers_static_fee` | ~27s | ~3 + a mid-epoch kill/restart | boundary waits + fixed sleeps + catch-up deadlines |
| `test_static_fee_applied_at_epoch_boundary` | ~19s | ~2 | two boundary waits at 10s |

**Important measurement note.**  A raw `cargo nextest run -p e2e-tests --run-ignored ...`
reports these tests at wildly larger numbers (e.g. `test_eip1559` at ~688s) because,
without `TN_BIN_PATH` set, `get_telcoin_network_binary()` shells out to escargot's
`cargo build` at test startup; the first test to run pays a full (re)build.  That
per-process build tax is what #897's `TN_BIN_PATH` / `make build-e2e-bin` change
removes - run these via `make test-e2e` / `make public-tests` and the reported time
collapses to the active-network windows above.  This issue is about the *remaining*
intrinsic cost once that tax is gone.

### Finding 1 - cut base-fee `EPOCH_DURATION` from 10s to 5s

`basefee.rs` defines its own `const EPOCH_DURATION: u64 = 10` and threads it into
genesis via `config_local_testnet_with_worker_fee_configs(..., Some(EPOCH_DURATION as u32), ...)`
in all three tests. 5s is the consensus minimum epoch duration (the same value
#897 used for `epochs.rs`).  Because every per-epoch wait scales with the on-chain
epoch duration, halving it roughly halves the boundary-waiting portion of each
test.  No test asserts on the literal value except through
`snap.epoch_duration`, which is read from chain, so it stays consistent.

Findings 2-4 are the caveats that make the cut safe;  they are the base-fee
analogue of #897's "two poll deadlines must be floored absolutely."

### Finding 2 - the mid-epoch phase window does not fit a 5s epoch (blocker)

`wait_for_mid_epoch` (used by `test_mid_epoch_restart_recovers_static_fee`)
returns the current epoch only when the measured phase satisfies:

```
phase >= MIN_PHASE (2s)  &&  phase + END_MARGIN (4s) <= epoch_duration
```

That requires `epoch_duration >= 7s` for any phase to qualify (and only a
1-second window even at 6s).  At a 5s epoch **no phase satisfies both bounds**, so
the loop runs to its `EPOCH_DURATION * 4` deadline and the test fails before the
restart under test happens.  The cut must shrink `MIN_PHASE` / `END_MARGIN` (e.g.
1s / 2s) or express the mid-epoch window as fractions of the epoch.  Their whole
purpose is to keep the kill away from both boundaries, which fractions preserve.

### Finding 3 - the blind `sleep(3s)` before landing a "mid-epoch" tx

`land_priced_tx_mid_epoch` and `land_cheap_tx_mid_epoch` both begin with a fixed
`tokio::time::sleep(Duration::from_secs(3))` so the transaction lands mid-epoch
rather than at a boundary.  Against a 5s epoch that positions the tx at phase 3 of
5 (only 2s before the next boundary) which is exactly the boundary-adjacent
landing the tests' own comments warn poisons the fee attribution.  Replace the
fixed sleep with the same phase measurement `wait_for_mid_epoch` already uses, or
scale it to the epoch, so the tx still lands safely mid-epoch at 5s.

### Finding 4 - floor the restart catch-up deadlines absolutely

`test_mid_epoch_restart` waits `EPOCH_DURATION * 6` (60s at 10s) and
`EPOCH_DURATION * 4` (40s at 10s) via `wait_for_block_fee` for the restarted node
to catch up to specific blocks.  Catch-up is a fixed sync cost (the node was down
only ~2s) not something that scales with the epoch cadence, so a naive halving
to 30s/20s needlessly narrows the margin.  Floor these like #897 floored the
`tn_epochRecord` polls (e.g. `(EPOCH_DURATION * 6).max(60)`, `(EPOCH_DURATION * 4).max(40)`),
so the deadlines keep their current absolute values while the epoch shrinks.

### Finding 5 - tighten the base-fee wait-loop poll intervals

`wait_for_epoch_at_least`, `wait_for_mid_epoch`, and `land_tx_and_read_fee` each
poll on a fixed 1-second cadence.  With 5-second epochs, 1s granularity is coarse
relative to the boundary it is trying to catch (up to ~1s of slop per boundary,
across several boundaries per test).  Dropping to ~250-500ms sharpens
boundary/confirmation detection - the same rationale #897 used to take
`call_rpc`'s retry backoff from 1s to 250ms, applied here to the base-fee loops.

### Finding 6 - event-drive (or trim) the fixed down-window sleep

`test_mid_epoch_restart` sleeps a fixed `2s` to "let the rest of the network
advance while the node is down" before restarting.  Small, but it is a fixed
sleep gating a consensus condition;  replacing it with a poll of a live peer's
block/consensus-header advance (or simply trimming it) removes another
non-essential wait.  Lowest priority of the six.

## Acceptance criteria

- All three base-fee tests continue to pass with identical assertions and the
  same fee-oracle checks.
- No reduction in the number of epochs crossed or transactions landed per test;
  only the epoch *duration* and the wait mechanisms change.
- The mid-epoch kill in `test_mid_epoch_restart` still lands strictly mid-epoch
  (Finding 2/3 done correctly), not at a boundary.
- A demonstrated ~40-50% reduction in the active-network window of the epoch-bound
  base-fee tests (`test_static_fee_*`, `test_eip1559_*`) on the same machine, run
  via `make test-e2e` so the escargot build tax is excluded.

## Notes

`metrics::test_metrics_endpoint_serves_tn_and_reth_metrics` is *not* in scope: it
runs on the default (8-hour) epoch and never waits on an epoch boundary, so its
intrinsic window is only a few seconds.

Line numbers should be re-confirmed against `main` before the PR (they are cited
from a worktree that has #897 staged, which shifts `basefee.rs` by a line or two
around the `EPOCH_DURATION` comment;  the helper functions are otherwise
untouched).

MavenRain:
## Summary

  Test-only optimization of the three `#[ignore]`d base-fee e2e tests in
  `crates/e2e-tests/tests/it/basefee.rs`, cutting the epoch-boundary portion of their
  active-network window by roughly half. No assertion, epoch or iteration count, transaction
  landed, or fee-oracle check changes; only the epoch duration and the wait
  mechanisms do.

  Disjoint from #897 (which explicitly left `basefee.rs` at `EPOCH_DURATION = 10` as out of
  its four-test scope) and from #894. Reuses #897's two techniques on the base-fee tests: a
  decoupled epoch-duration cut with absolutely-floored deadlines, and event-driven waits
  replacing fixed sleeps.

  ## Changes

  1. `EPOCH_DURATION` 10s -> 5s (the consensus minimum). Every per-epoch wait scales with it.
  2. `wait_for_mid_epoch`: `MIN_PHASE` 2s -> 1s, `END_MARGIN` 4s -> 2s, keeping the mid-epoch
     landing window `[MIN_PHASE, epoch_duration - END_MARGIN]` non-empty at a 5s epoch
     (`[1s, 3s]`). The old 2s/4s admits no phase at 5s, the blocker the issue calls out.
  3. `land_priced_tx_mid_epoch` / `land_cheap_tx_mid_epoch`: replace the blind 3s sleep with the same `wait_for_mid_epoch` phase measurement, so the tx still lands clear of a boundary.
  4. `test_mid_epoch_restart`: floor the restarted-node catch-up deadlines absolutely,
     `(EPOCH_DURATION * 6).max(60)` and `(EPOCH_DURATION * 4).max(40)`, because state-sync
     catch-up is a fixed cost that must not scale with the epoch cadence.
  5. Tighten the base-fee wait-loop poll cadence 1s -> 250ms in `wait_for_epoch_at_least`,
     `wait_for_mid_epoch`, and `land_tx_and_read_fee`.
  6. `test_mid_epoch_restart`: replace the fixed 2s post-kill down-window sleep with an
     event-driven `wait_out_downtime` (2s floor, then proceed once a live peer's consensus
     header advances, with a fail-safe cap), mirroring `restarts.rs::wait_for_downtime`.

  ## Testing

  - `cargo build -p e2e-tests --tests` compiles clean (0 errors, 0 warnings).
  - Acceptance criterion 4 (a *demonstrated* ~40-50% active-network-window
    confirmed via `make test-e2e` with the prebuilt node binary, so the escargot build tax is
    excluded per the issue's measurement note.

  Closes #907.

### DIFF
--- a/crates/e2e-tests/tests/it/basefee.rs
+++ b/crates/e2e-tests/tests/it/basefee.rs
@@ -42,14 +42,19 @@ use tokio::time::Instant;
 use tracing::info;
 
 use crate::common::{
-    address_from_word, call_rpc, get_balance, get_block, get_block_number, get_key, kill_child,
-    network_advancing, send_tel, start_validator, ProcessGuard,
+    address_from_word, call_rpc, get_balance, get_block, get_block_number, get_key,
+    get_latest_consensus_header_number, kill_child, network_advancing, send_tel, start_validator,
+    ProcessGuard,
 };
 
-/// Epoch duration (seconds) for base-fee tests. Held at 10s independently of
-/// `epochs.rs::EPOCH_DURATION` (which #897 cut to 5s): these base-fee tests are outside that
-/// four-test scope, and 10s keeps ample margin for their epoch-boundary assertions.
-const EPOCH_DURATION: u64 = 10;
+/// Epoch duration (seconds) for the base-fee tests. 5s is the consensus minimum epoch duration;
+/// halving it from 10s (the same decoupled cut #897 applied to `epochs.rs`) roughly halves the
+/// epoch-boundary portion of each test without changing any assertion or epoch/iteration count.
+/// Every per-epoch wait scales with this value, but the restarted-node catch-up deadlines in
+/// [`test_mid_epoch_restart_recovers_static_fee`] are floored to an absolute minimum (`.max(..)`)
+/// rather than shrinking with it, because state-sync catch-up is a fixed cost independent of the
+/// epoch cadence.
+const EPOCH_DURATION: u64 = 5;
 
 /// Static fee used by the deterministic tests. A clearly non-`MIN` value so any block produced
 /// under the static strategy is unmistakable.
@@ -206,7 +211,7 @@ async fn test_eip1559_fee_rises_at_epoch_boundaries() -> eyre::Result<()> {
         let nonce = i as u128;
         // Cheap gas price is fine: the EIP-1559 fee stays tiny (single/low double digits) across
         // these few epochs. A tx that misses its deadline is FATAL: skipping an epoch lets empty
-        // 10s epochs pass (each *decreasing* the fee against `target_gas = 1`), which would
+        // 5s epochs pass (each *decreasing* the fee against `target_gas = 1`), which would
         // poison the monotonic assertion below. Every recorded epoch must land exactly one tx.
         let (block, fee) = land_cheap_tx_mid_epoch(&client_urls[0], &funded_key, to, nonce)
             .await
@@ -350,8 +355,10 @@ async fn test_mid_epoch_restart_recovers_static_fee() -> eyre::Result<()> {
         kill_idx + 1
     );
 
-    // Let the rest of the network advance while the node is down.
-    tokio::time::sleep(Duration::from_secs(2)).await;
+    // Let the rest of the network advance while the node is down. Event-driven (see
+    // `wait_out_downtime`) instead of a fixed 2s sleep: hold a 2s downtime floor so the restart
+    // genuinely exercises catch-up, then proceed as soon as a live peer's consensus chain moves.
+    wait_out_downtime(&client_urls, kill_idx, 2).await?;
 
     // Restart the killed node on a fresh RPC port and re-register it.
     let new_rpc_port =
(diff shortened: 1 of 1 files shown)


grantkee (approved):
LGTM!

sstanfield (approved):

PR #922, opened 2026-07-14
Tolerate mid-epoch validator ejection in epoch records

### LINKED ISSUE #921: Mid-epoch validator ejection halts epoch boundary closing on running nodes
## Problem

Governance can eject a validator via `ConsensusRegistry::burn`, which retires the validator, confiscates its stake, and removes it from the stored current/next/subsequent committee arrays with a swap-and-pop. This can happen at any point during an epoch, not just at a boundary.

The node that produces each epoch's `EpochRecord` requires the committee it just read on-chain to be exactly equal to the `next_committee` promised by the previous epoch's record. A mid-epoch `burn` breaks that equality: the live committee read becomes a strict subset of what was promised. Every running node hits this check, errors out, and fails to close the next epoch boundary — the network stalls.

This is inconsistent with how a syncing node already behaves: `state-sync`'s epoch-record verifier tolerates a shrunken committee when a node is catching up from peers, using its own ad hoc bound. The codebase already assumed mid-epoch ejection was a survivable event, but the code path that actually produces records for a live network couldn't handle it. Any governance ejection during an epoch, rather than exactly at a boundary, currently stops the whole network's epoch progression — and the two independent acceptance checks (producer vs. sync verifier) drifting apart is a correctness risk beyond just this trigger.

## Solution

Introduce a single, shared predicate that both the epoch-record producer and the state-sync verifier delegate to, so they accept exactly the same committee shapes for a given previous-epoch promise: an unchanged set (in any stored order), or a bounded shrink that keeps at least four members, keeps a BFT-safe supermajority of the promised set, and draws every remaining member from that set. Anything else — growth, an unknown member swapped in, or a shrink past the tolerance floor — should remain a hard error, since producing a record that sync would reject only hides the divergence instead of surfacing it.

The epoch-record producer should use this predicate instead of strict equality, warning rather than failing when it detects a tolerated mid-epoch ejection. The epoch-vote collector should also enforce that only current-committee members' votes count toward quorum, and that a node ejected from the committee stops self-signing but still stores the certificate the remaining members produce, so it keeps following the chain as an observer instead of getting stuck.

To close the loop, the Rust `ConsensusRegistry` contract bindings should expose `burn` so tests, and any future governance tooling, can construct ejection transactions, and end-to-end coverage should exercise both ejection shapes: a validator ejected from a future-only committee slot, where no live committee changes shape, and a validator ejected from the currently active committee mid-epoch, the shape that halts today.

grantkee:
- Extract `EpochRecord::committee_compatible` as the single predicate for "is this committee an acceptable successor to what the previous record promised" — same set in any order, or a bounded shrink (≥4 members, BFT-safe supermajority, no unknown members). Both the epoch-record producer (`close_epoch.rs`) and `state-sync`'s `epoch_committee_valid` now delegate to it instead of carrying their own drifted logic.
- `close_epoch.rs`'s record-building logic is extracted into a standalone, unit-testable `build_epoch_record` function that warns instead of erroring when it detects a tolerated mid-epoch ejection.
- `epoch_votes.rs`'s vote collector now ignores votes from keys outside the record's committee and skips self-signing for a node that's been ejected, while still storing the resulting certificate.
- Add the `burn` binding to the Rust `ConsensusRegistry` interface (`system_calls.rs`) and a `governance_burn_tx` test helper.
- Add `crates/e2e-tests/tests/it/eject.rs` with two scenarios: ejection from a future-only committee slot, and ejection from the live current committee mid-epoch (the acceptance test for this fix — it hangs on pre-fix code).
- Add unit/integration coverage for `committee_compatible`, `build_epoch_record`, the vote collector's ejected-vote handling, `state-sync`'s record collection across an ejection epoch, and storage-layer committee lookups across an ejection epoch.
- Refactor `basefee.rs` and `epochs.rs` test helpers (epoch snapshot polling, mid-epoch phase detection, tx-receipt lookup, node/genesis bootstrapping) into shared helpers in `common.rs` so `eject.rs` and future epoch-boundary tests can reuse them instead of duplicating.

closes #921


MavenRain (dismissed):
No issues with this one

MavenRain (dismissed):

sstanfield (dismissed):
Looks good and seems clean to me.

PR #923, opened 2026-07-14
test: replace fixed real-time sleeps with condition polling (shared wait_until)

### LINKED ISSUE #883: test: replace fixed real-time sleeps with condition polling across the test suite (shared wait_until)
## Summary

PR #859 replaced four fixed `tokio::time::sleep(...)` "wait for convergence" calls in
`network-libp2p/src/tests/network_tests.rs` with a bounded condition poll
(`wait_until(timeout, description, condition)`).  On the review, @grantkee asked that we
"continue to expand on this approach."

A repo-wide survey (20 files, ~110 test-scope sleeps) shows the same anti-pattern is
widespread and, in most places, mechanically convertible against accessors that already
exist.  It also shows the polling helper is being reinvented independently in at least four
places.  This issue tracks the expansion and proposes a shared helper so we stop rewriting it.

## Two shapes of work

1. **Fixed sleep, then assert** (the exact #859 target): a `sleep(N)` guesses that an async
   event has landed, then the next line asserts on the settled state.  These are latent CI
   flakes;  converting removes the race.
2. **Hand-rolled poll loops** that already do the right thing but each reinvent `wait_until`
   inline (`wait_for_peer_discovery`, `wait_for_rpc`, `wait_for_execution`,
   `try_accept_or_sleep`, the epochs/sync deadline loops, ...).  These are not flaky, but they
   should fold onto one shared helper.

The helper (`wait_until`) currently lives only inside `network_tests.rs`.  There is no shared
poll helper in `tn-test-utils` or `e2e-tests`.

### 0. Lift `wait_until` into a shared test utility (prerequisite for the rest)
Promote a generic `wait_until` / `eventually` into `tn-test-utils` (or `e2e-tests` common) so
every crate can depend on one implementation.  Provide an async version and a blocking analogue,
because the `e2e-tests` helpers poll with `std::thread::sleep` / `block_on`, not `.await`.

### 1. Finish `network-libp2p/src/tests/network_tests.rs`
#859 explicitly left the rest of the file "outside this issue's scope." ~34 further top-level
sleeps in that same file are the identical pattern and poll against the same, already-present
`NetworkHandle` accessors:
- `connected_peer_ids()` (`types.rs:1069`), `connected_peers()` (`types.rs:686`) for
  connect / disconnect / ban settle before the containment assertion (lines 769, 889, 1167,
  1254, 1481, 1798, 1848, 1934, 2006, ...).
- `peer_score()` (`types.rs:1055`) for penalty / decay / unban tests (lines 548, 1494, 1504,
  1834, 1957, ...).

### 2. `consensus/primary` certificate-store write visibility
- `synchronizer_tests.rs` lines 179, 223, 240, 281: fixed sleeps after `try_accept_certificate`
  (all `start_paused = true`) waiting for the spawned synchronizer to persist.  Poll
  `CertificateStore::{contains,notify_read,read}` (`storage/src/stores/certificate_store.rs`).
- `certificate_fetcher_tests.rs` lines 343, 375: "allow time for fetch-reply processing" then
  assert store state.  Poll `certificate_store.read(...)`.  Also fold the existing hand-rolled
  loops at lines 62 and 502 onto the shared helper.

### 3. `node/src/health.rs` (3 sleeps)
`test_tcp_healthcheck`, `test_health_workers_reports_{ready,not_ready}` each `sleep(10ms)` with
"give server time to start listening."  The real predicate is a successful `TcpStream::connect`,
already used by the file-local `roundtrip()`.  Note `spawn()` binds the `TcpListener`
synchronously before returning the address, so these can likely just be deleted.

### 4. Batch layer
- `batch-builder/tests/it/build_batches.rs` lines: 3 fixed `sleep(1s)` for store-write /
  pool-maintenance notifications, each with an existing observable predicate.
- `batch-validator/src/validator.rs` lines 700, 761: existing poll loops over
  `tx_pool.pool_size().pending` (`tn-reth/src/txn_pool.rs`) to fold onto the shared helper.

### 5. `e2e-tests`
- Genuinely fixed pads to convert: `common.rs:502` `send_and_confirm()` `sleep(1000ms)` (the
  next line already polls `get_balance_above_with_retry`, so it is redundant); `faucet.rs:279`
  `sleep(15s)` network-converge pad, poll `eth_blockNumber` (the `wait_for_rpc` helper in
  `genesis_tests.rs` already does exactly this); `restarts.rs:571` observer-sync settle.
- Fold onto the shared helper: the hand-rolled deadline loops in `sync.rs` (4), `epochs.rs` (8),
  `common.rs::network_advancing`, and the `faucet.rs` `balanceOf` loop. Switch the blocking
  `std::thread::sleep` inside async fns (`sync.rs:138`, `common.rs`) to the async helper.

## Explicit non-goals (surveyed and deliberately excluded)

Converting these would be wrong; listing them so the scope stays honest:
- `consensus/executor/tests/it/main.rs` (4): a redundant yield between `recv().await`
  iterations. Fix is deletion or `yield_now()`, not a poll.
- `network-libp2p/src/tests/peer_manager.rs` (2): real score-decay / ban-window wall-clock.
- `consensus/primary/tests/it/recovery_tests.rs` (2): MDBX lock handoff on drop, no accessor to
  poll (only "retry the reopen").
- `worker/tests/it/quorum_waiter_tests.rs`, `worker/tests/it/network_tests.rs`,
  `certificate_fetcher_tests.rs:601`: the timeout firing IS the asserted behavior.
- `restarts.rs` lines 34, 126, 183, 332: deliberate "pause so a lagging node fails the check"
  settles; polling would defeat their purpose.
- `consensus/primary/src/network/handler.rs`: production code, out of scope by #859's own
  "test-only" framing.

## Acceptance criteria
- One shared `wait_until` (async + blocking) in a test-utils location, documented.
- `network_tests.rs` fully migrated (no top-level fixed convergence sleeps remain).
- At least the item-2 through item-5 fixed-sleep sites converted or deleted, each polling a
  named existing accessor with a bounded timeout and a descriptive failure message.
- Non-goal sleeps left untouched (or handled by their correct fix, e.g. deletion).

MavenRain:
Closes #883.

Adds a shared, bounded condition-poll helper and migrates the fixed
`sleep(N)` "wait for convergence" calls across the test suite to it,
following the pattern established by #859 and extending it repo-wide as
@grantkee requested.  Each converted site now polls a named accessor with
a bounded timeout and a descriptive failure message instead of guessing
how long an asynchronous event takes.

Item 0 (prerequisite): lift the helper into tn-test-utils
- New crates/test-utils/src/wait.rs exports `wait_until` (async) and
  `wait_until_blocking` (the synchronous analogue for the e2e helpers
  that poll with std::thread::sleep / block_on). Both fold over a bounded
  ~10ms poll sequence and carry the first resolved verdict forward, so
  they behave under tokio start_paused virtual time as well as wall clock.
- Added tn-test-utils as a dev-dependency to tn-primary, tn-batch-builder,
  tn-batch-validator, and e2e-tests (network-libp2p and tn-node already
  had it).

Item 1: network-libp2p/src/tests/network_tests.rs
- Deleted the file-local `wait_until` (now imported from tn-test-utils).
- Converted 18 fixed convergence sleeps: connection-establishment and
  disconnect settles now poll connected_peer_ids()/connected_peers(), and
  the penalty/decay/unban tests poll peer_score() for the score threshold
  the following assertion checks.
- Folded 3 hand-rolled poll loops (wait_for_peer_discovery plus two inline
  loops) onto the shared helper.
- Left the genuine wall-clock waits untouched: heartbeat-driven ban/prune
  windows, OS port reuse, kad record TTL, and dial-timeout / "confirm the
  connection was rejected" negative assertions.

Item 2: consensus/primary
- certificate_fetcher_tests.rs: folded the two positive-wait loops
  (verify_certificates_in_store and the round-1 timeout loop).  The six
  "wait, then assert the certificate is NOT in the store / no new request
  arrived" sleeps are deliberately left as fixed waits: polling for an
  absence returns immediately and would stop the test from giving the
  system a fair chance to (wrongly) act.

Item 3: node/src/health.rs
- Deleted the three "give server time to start listening" sleeps.
  HealthcheckServer::spawn binds and listens on the TcpListener
  synchronously before it returns the address, so the kernel accepts
  connections into the backlog before the accept loop runs;  the roundtrip
  helper already bounds connect and exchange with a 500ms timeout.

Item 4: batch layer
- batch-builder/tests/it/build_batches.rs: three store-write /
  pool-maintenance sleeps now poll the same accessor the following
  assertion checks (NodeBatchesCache read, pool_size().pending).
- batch-validator/src/validator.rs: folded the two pool_size().pending
  deadline loops onto the helper.

Item 5: e2e-tests
- common.rs: removed the redundant send_and_confirm 1s pad (the next line
  already polls get_balance_above_with_retry); folded network_advancing.
- faucet.rs: the 15s network-converge pad now polls block production;
  folded the stablecoin balanceOf loop.
- restarts.rs: folded the two observer catch-up loops.  Left the
  deliberate "pause so a lagging node fails the check" settles and the
  SIGSTOP pause untouched.
- sync.rs and epochs.rs: folded the epoch-boundary / tn_epochRecord
  deadline loops (poll the condition, then re-read the idempotent RPC
  value) and switched the block_on'd std::thread::sleep loop to the async
  helper.

Finding, out of scope here: crates/consensus/primary/src/tests/
synchronizer_tests.rs (item 2's other half) is orphaned dead code on main.
No source module declares it via #[path], `cargo nextest list` shows zero
tests from it, and wiring it in fails to compile against current APIs (it
still imports crate::synchronizer, tn_primary::CommitteeFixture,
tn_storage::traits, and BlsAggregateSignatureBytes, all since renamed or
removed).  Its sleeps therefore never build or run, so there is nothing to
de-flake; this change leaves the file untouched and flags it for the
maintainers to either repair-and-rewire or delete.

Deliberately deferred (noted for a follow-up, to keep this change honest):
the two epoch-roll read-consistency retry loops in epochs.rs (bounded
3-attempt guards, not convergence waits);  the get_block /
get_balance_above_with_retry internal loops in common.rs (they return a
value, folding would change their signatures);  one ambiguous max-capacity
dial in network_tests.rs;  and one pre-action pause in
certificate_fetcher_tests.rs with no observable predicate to poll.

Verification: cargo check --tests passes for every affected crate;
cargo +nightly fmt --check and clippy are clean (the change adds no new
lint warnings).  The node/health.rs tests (which lost their sleeps, 5 passed) and the
certificate_fetcher_tests suite (which exercises the folded store polls and
the deliberately-retained absence-assertion sleeps, 9 passed) were run and
pass.  The heavier network-libp2p and e2e suites were compile-verified
rather than run here.

Signed-off-by: Onyeka Obi <softwareengineerasaservant@isurvivable.cv>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3142,6 +3142,7 @@ dependencies = [
  "tn-config",
  "tn-node",
  "tn-reth",
+ "tn-test-utils",
  "tn-types",
  "tokio",
  "tracing",
@@ -11634,6 +11635,7 @@ dependencies = [
  "tn-network-types",
  "tn-reth",
  "tn-storage",
+ "tn-test-utils",
  "tn-types",
  "tn-worker",
  "tokio",
@@ -11645,9 +11647,11 @@ name = "tn-batch-validator"
 version = "0.1.0"
 dependencies = [
  "assert_matches",
+ "eyre",
  "rayon",
  "tempfile",
  "tn-reth",
+ "tn-test-utils",
  "tn-types",
  "tokio",
 ]
@@ -11860,6 +11864,7 @@ dependencies = [
  "tn-primary",
  "tn-reth",
  "tn-storage",
+ "tn-test-utils",
  "tn-test-utils-committee",
  "tn-types",
  "tokio",
@@ -11973,6 +11978,7 @@ version = "0.1.0"
 dependencies = [
  "clap",
  "eyre",
+ "futures",
  "indexmap 2.13.0",
  "telcoin-network-cli",
  "tempfile",
@@ -11981,6 +11987,7 @@ dependencies = [
  "tn-reth",
  "tn-test-utils-committee",
  "tn-types",
+ "tokio",
 ]
 
 [[package]]
--- a/crates/batch-builder/Cargo.toml
+++ b/crates/batch-builder/Cargo.toml
@@ -36,6 +36,7 @@ tn-storage = { workspace = true, features = ["test-utils"] }
 tn-batch-validator = { workspace = true }
 tn-engine = { workspace = true }
 tn-batch-builder = { workspace = true, features = ["test-utils"] }
+tn-test-utils = { workspace = true }
 
 [features]
 default = []
--- a/crates/batch-builder/tests/it/build_batches.rs
+++ b/crates/batch-builder/tests/it/build_batches.rs
@@ -15,6 +15,7 @@ use tn_reth::{
     RethChainSpec, RethEnv,
 };
 use tn_storage::{open_db, tables::NodeBatchesCache};
+use tn_test_utils::wait_until;
 use tn_types::{
     gas_accumulator::GasAccumulator, test_genesis, Address, Batch, BatchValidation, Bytes,
     Certificate, CertifiedBatch, CommittedSubDag, ConsensusHeaderDigest, ConsensusOutput, Database,
@@ -26,7 +27,7 @@ use tokio::time::timeout;
 use tracing::debug;
 
 #[tokio::test]
-async fn test_make_batch_el_to_cl() {
+async fn test_make_batch_el_to_cl() -> eyre::Result<()> {
     let tmp_dir = TempDir::new().expect("temp dir");
     let task_manager = TaskManager::default();
     //
@@ -138,17 +139,17 @@ async fn test_make_batch_el_to_cl() {
     //=== Test batch flow
     //
 
-    // wait for new batch
-    let mut sealed_batch = None;
-    for _ in 0..5 {
-        let _ = tokio::time::sleep(Duration::from_secs(1)).await;
-        // Ensure the batch is stored
-        if let Some((digest, wb)) = store.iter::<NodeBatchesCache>().next() {
-            sealed_batch = Some(SealedBatch::new(wb, digest));
-            break;
-        }
-    }
-    let sealed_batch = sealed_batch.unwrap();
+    // wait for new batch to be stored
+    wait_until(Duration::from_secs(5), "batch stored", || async {
+        Ok(store.iter::<NodeBatchesCache>().next().is_some())
+    })
+    .await?;
+    // re-fetch the stored batch once the store has converged
+    let sealed_batch = store
+        .iter::<NodeBatchesCache>()
+        .next()
+        .map(|(digest, wb)| SealedBatch::new(wb, digest))
+        .expect("batch in store after wait");
 
     // ensure batch validator succeeds
     let batch_validator =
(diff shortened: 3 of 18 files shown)


sstanfield (approved):
LGTM!

grantkee (approved):
Great initiative! LGTM

PR #925, opened 2026-07-15
backpressure when engine channel full instead of dropping output

### LINKED ISSUE #924: Engine backpressure should not be fatal
# Problem

EVM is single-threaded and a potential bottleneck at volume.

The payload builder `try_send` treats transient `Full` backpressure as fatal, returning `ChannelClosed` and  halting the engine

**Evidence**
- `crates/engine/src/payload_builder.rs:92-95`: `engine_update_tx.try_send(...).map_err(|e| ... TnEngineError::ChannelClosed)?` collapses **both** `TrySendError::Full` and `Closed` into a fatal error. `?` exits `execute_consensus_output`; the blocking task forwards it via the oneshot (`lib.rs:128-136`) and `run()` re-raises at `crates/engine/src/lib.rs:251`, ending the engine task permanently.
- Same defect one layer down on the non-empty path: `crates/tn-reth/src/lib.rs:869-878` (`finish_executing_output`) → `TnRethError::EngineUpdateChannelClosed` → `TnEngineError::Reth` (`error.rs:14-15`) → propagates from `payload_builder.rs:189`. Channel is `mpsc::channel(64)` (`node.rs:328`).

grantkee:
- use `blocking_send` in the spawned blocking task instead of erroring on backpressure

closes #924

### DIFF
--- a/crates/batch-builder/src/lib.rs
+++ b/crates/batch-builder/src/lib.rs
@@ -689,9 +689,13 @@ mod tests {
             // execute output to trigger canonical update
             let args = BuildArguments::new(reth_env.clone(), output.clone(), parent);
             let (engine_update_tx, _engine_update_rx) = tokio::sync::mpsc::channel(64);
-            let final_header =
-                execute_consensus_output(args, gas_accumulator.clone(), engine_update_tx)
-                    .expect("output executed");
+            // spawn blocking for payload_builder::blocking_recv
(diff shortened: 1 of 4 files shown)


MavenRain (approved):
LGTM!

sstanfield (approved):
Good catch.

PR #928, opened 2026-07-15
test: reuse the prebuilt e2e node binary on every path, not only make (#926)

### LINKED ISSUE #897: `test`: cut the four ignored e2e tests (epoch boundary/sync, delayed restarts) 25-50% via de-contention, event-driven restart waits, and a decoupled epoch-duration cut (no coverage loss)

### LINKED ISSUE #926: `test`: the ignored e2e suite still cold-builds the node binary in-test on every path except `make` (follow-up to #900)
## Summary

#900 (closing #897) added a `TN_BIN_PATH` escape hatch so e2e tests can reuse a
prebuilt `telcoin-network` binary instead of compiling it lazily inside the first
test.  That works, but **only when the run goes through `make test-e2e` /
`make test-restarts` / `make public-tests`**, which are the only callers that set
`TN_BIN_PATH`.  Every other way of running the ignored e2e suite still hits the
escargot fallback and cold-compiles the node binary inside the first test, which
under `nextest` output capture looks like the first test sitting frozen for
several minutes (observed on the order of 9 to 10 minutes) before it does
anything. The paths that still stall include:

- the project's own `etc/test-and-attest.sh` (line 102), which #900 did not rewire;
- the raw command a contributor would copy, `cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features`;
- IDE / editor test runners (VS Code, RustRover) that invoke `cargo nextest` / `cargo test` directly.

On top of that, the fallback is a *guaranteed* cold build even when a binary was
already compiled, because it still passes `.current_target()` and so builds into a
separate `target/<triple>/` tree that shares nothing with the `target/debug/`
tree that `make build-e2e-bin` and ordinary `cargo build` populate.

This is test-infrastructure only: no production-code change, and no change to any
test's assertions, counts, features, or coverage.  It closes the gap #900 left so
the sit disappears on the paths people actually run, not only through `make`.

## Immediate workaround (no code change)

Run the ignored suite through the target that already sets `TN_BIN_PATH`:

```
make test-e2e         # or: make test-restarts
```

or set it by hand:

```
cargo build --bin telcoin-network --features tn-storage/test-utils
TN_BIN_PATH=$(pwd)/target/debug/telcoin-network \
  cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features
```

Either way the first test no longer compiles the binary; the (still ~9-10 min at
`opt-level=0`) build happens once, visibly, up front.

## What you observe (without the workaround)

```
cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features
```

The first test to run shows `RUNNING` and produces no output for several minutes
before it begins asserting.  Because #900 also set the e2e group to
`max-threads = 1`, exactly one test runs at a time, so the whole suite is gated
behind that first-test build with nothing else progressing.

To see the build cost in isolation, from a clean `target/`:

```
time cargo build --bin telcoin-network --features tn-storage/test-utils --target "$(rustc -vV | sed -n 's/host: //p')"
```

(The ~9-10 minute figure is the observed wall-time of the silent sit, not a
number measured in a controlled run;  the mechanism below is confirmed from
source.  Pin the magnitude with the command above before quoting it in the PR.)

## Root cause 1: the escape hatch is only wired into `make`, not the documented/maintainer paths

`get_telcoin_network_binary()` short-circuits on `TN_BIN_PATH`, else falls back to
building via escargot:

- `crates/e2e-tests/src/lib.rs:391-397`: returns `TestBinary::Prebuilt(path)` when
  `TN_BIN_PATH` is set (`std::env::var("TN_BIN_PATH")` at `:393`).
- `crates/e2e-tests/src/lib.rs:406-413`: otherwise `TestBinary::Cargo(CargoBuild::new().bin("telcoin-network").features("tn-storage/test-utils")...run())`.

Only the `make` targets set the variable:

- `Makefile:113-114` `build-e2e-bin`: `cargo build --bin telcoin-network --features tn-storage/test-utils --target-dir $(CURDIR)/target`
- `Makefile:117` `E2E_BIN := $(CURDIR)/target/debug/telcoin-network`
- `Makefile:124-125` `test-e2e: build-e2e-bin` then `TN_BIN_PATH=$(E2E_BIN) cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features`

But the maintainer attest script was not updated by #900 and still runs the raw
command with no prebuild and no `TN_BIN_PATH`:

- `etc/test-and-attest.sh:102`: `cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features`

So anyone following the attest script, copying the raw command, or using an IDE
runner gets the cold in-test build.

## Root cause 2: the fallback is guaranteed cold because of `.current_target()`

The fallback passes `.current_target()` (`crates/e2e-tests/src/lib.rs:412`), which
adds `--target <host-triple>`, so it emits into `target/<triple>/debug/`.  Nothing
else in the repo uses a triple subdir (`.cargo/config.toml` has no
`[build] target = ...`), and `make build-e2e-bin` builds into plain
`target/debug/` (`Makefile:114`, no `--target`).  The two trees share zero
artifacts, so:

- even a developer who ran `cargo build --bin telcoin-network` (or `make build-e2e-bin`) beforehand still triggers a full cold rebuild in the fallback, because the fallback looks in a different directory;
- the fallback compiles a second full copy of the reth-based node into `target/<triple>/` that no other command reuses.

Dropping `.current_target()` makes the fallback's invocation identical to
`build-e2e-bin`, so it becomes a fast cargo freshness check whenever the binary
was already built into `target/debug/`.

## Root cause 3: the build is invisible under `nextest`, so it reads as a hang

`nextest` buffers each test's stdout/stderr until the test finishes, so the entire
`cargo build`, including the `info!("building main binary for e2e tests")` line,
is swallowed.  With `max-threads = 1` (`.config/nextest.toml:23`, lowered from 2 by
#900) and `slow-timeout = 720s` (`:28,:39`), the first test shows no output and
no "slow" marker until 12 minutes in, so a ~9-10 minute build looks like a dead
hang.

## Proposed fix

Make every invocation path get the prebuilt binary, and stop the fallback from
being guaranteed cold.  Both parts keep the `TestBinary` return type and all call
sites unchanged.

1. Rewire the maintainer path.  In `etc/test-and-attest.sh`, replace the raw
   ignored-e2e line (`:102`) with the `make` target that already prebuilds and
   sets `TN_BIN_PATH`:
   ```sh
   make test-e2e
   ```
   (or inline the `build-e2e-bin` + `TN_BIN_PATH=... cargo nextest ...` pair, to
   avoid a `make` dependency in the script).

2. De-cold the fallback.  In `crates/e2e-tests/src/lib.rs:412`, drop
   `.current_target()` (or add `[build] target = "<host-triple>"` to
   `.cargo/config.toml`) so the fallback builds into `target/debug/`, the same
   tree as `build-e2e-bin` and ordinary `cargo build`.  After this, a prior build
   makes the fallback a seconds-long freshness check instead of a cold compile.

3. (Optional, low-cost) Make the fallback self-explanatory: when `TN_BIN_PATH` is
   unset, emit an `eprintln!` before building, e.g.
   `"TN_BIN_PATH not set; building telcoin-network in-process (set it or run `make test-e2e` to skip this)"`,
   so the silent sit stops surprising people.  (`nextest` captures it, but it shows
   in the captured output on completion and in `--no-capture` runs.)

## Out of scope (separate issue)

This removes the *redundant/invisible* build from the non-`make` paths.  It does not
make the one-time node-binary build *faster*: at `opt-level = 0` that compile is
inherently several minutes.  Reducing its duration (an `[profile.e2e]` opt-level
bump for the escargot/prebuild binary) is tracked separately in the e2e
runtime-profile draft and stacks on top of this.

## Coverage / behavior preservation

No test logic, assertion, epoch or iteration count, feature set, or node
configuration changes.  The same `telcoin-network` binary is built with the same
`tn-storage/test-utils` feature;  it is just built once into the shared target tree
and reused, on every invocation path rather than only through `make`.

## File references (verified against `origin/main` @ `9a9c2f6f`, the #900 merge)

- `crates/e2e-tests/src/lib.rs:28-40` `TestBinary` enum (`Prebuilt` / `Cargo`); `:391` `get_telcoin_network_binary()`; `:393` `TN_BIN_PATH` check; `:406-413` escargot fallback; `:409` `features("tn-storage/test-utils")`; `:412` `.current_target()`
- Callers (all `#[ignore]`): `sync.rs:55`, `epochs.rs:632`, `restarts.rs:236,377,464,548`
- `Makefile:113-114` `build-e2e-bin`; `:117` `E2E_BIN`; `:124-125` `test-e2e`; `:120-121` `test-restarts`; `:194-197` `public-tests`
- `etc/test-and-attest.sh:102` raw ignored-e2e run (not rewired by #900)
- `.config/nextest.toml:23` `max-threads = 1` (was 2); `:28,:39` `slow-timeout = 720s`
- `.cargo/config.toml` (no `[build] target`)
- `bin/telcoin-network/Cargo.toml:2` `name = "telcoin-network"`; `crates/e2e-tests/Cargo.toml:14` depends on `telcoin-network-cli` (not the bin)

MavenRain:
## Summary

Follow-up to #900 (which closed #897). #900 added the `TN_BIN_PATH` escape hatch so the
ignored e2e suite can reuse a prebuilt `telcoin-network` binary instead of compiling it lazily
inside the first test, but it only wired that hatch into the `make` targets.  Every other way of
running the ignored suite still cold-builds the node binary inside the first test, which under
`nextest` output capture looks like the first test frozen for several minutes.

This closes the gap on the paths people actually run, with **no** change to any test's
assertions, counts, features, coverage, or node configuration.  Test-infrastructure only.

## What changed

**`crates/e2e-tests/src/lib.rs`** (`get_telcoin_network_binary()` escargot fallback):
- Removed `.current_target()`.  It passed `--target <host-triple>`, so the fallback built into
  `target/<triple>/debug/`, a tree that shares no artifacts with the plain `target/debug/` that
  `make build-e2e-bin` and ordinary `cargo build` populate.  That made the fallback a *guaranteed*
  cold rebuild even when a matching binary already existed.  Building into `target/debug/` lets
  cargo report the binary `Fresh` and skip the rebuild.
- Added a stderr diagnostic before the build.  The pre-existing `info!` goes through `tracing`,
  which has no subscriber at that point under `nextest`, so the multi-minute build was invisible.
  `eprintln!` is captured by nextest and shown on completion (and live under `--no-capture`), so
  the delay is attributable instead of an apparent hang.  The message points at `make test-e2e` /
  `TN_BIN_PATH` as the way to skip it.

**`etc/test-and-attest.sh`** (the maintainer attest script, which #900 did not rewire):
- The raw `cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features` line now
  prebuilds the binary and passes `TN_BIN_PATH`, mirroring `make test-e2e` exactly.  The script
  already `cd`s to the project root, so `$(pwd)/target/debug/telcoin-network` matches the
  Makefile's `$(CURDIR)/target/debug/telcoin-network`. `--target-dir "$(pwd)/target"` keeps the
  build location consistent with `TN_BIN_PATH` even if `CARGO_TARGET_DIR` is set in the env.

**`crates/e2e-tests/README.md`**:
- Added a "Running the ignored e2e suite" section documenting the cold build, the `make test-e2e`
  fast path, and the `TN_BIN_PATH` escape hatch.  This is the one persistent doc a contributor
  debugging a "frozen first test" would open, and it survives nextest's output capture (which
  otherwise buffers the stderr notice until the build finishes).  It also flags that `TN_BIN_PATH`
  must be an absolute path, since nextest runs each test with its cwd set to the package
  (`crates/e2e-tests`), not the workspace root, so a relative path would not resolve.

## Scope / safety

- The `make` targets (`test-e2e`, `test-restarts`, `public-tests`) set `TN_BIN_PATH` and so take
  the `Prebuilt` branch; they never reach the changed fallback and are behaviorally unchanged.
- PR CI (`.github/workflows/pr.yaml`) runs `cargo nextest run --workspace ... --profile ci` with
  no `--run-ignored`, so it never runs the ignored e2e tests and never touches this code path.
- No production code changes.  Same binary, same `tn-storage/test-utils` feature;  it is just built
  once into the shared target tree and reused on every invocation path rather than only via `make`.

## Out of scope (separate issue)

This removes the *redundant / invisible* build from the non-`make` paths.  It does not make the
one-time `opt-level = 0` node-binary build *faster*; that is tracked separately.

Closes #926.

### DIFF
--- a/crates/e2e-tests/README.md
+++ b/crates/e2e-tests/README.md
@@ -1,5 +1,32 @@
 # E2E Tests
 
+## Running the ignored e2e suite
+
+The heavy e2e tests (restart and epoch tests) are `#[ignore]`d and each needs the
+`telcoin-network` node binary. When `TN_BIN_PATH` is unset, the first test builds that binary
+in-process via `escargot`, which at `opt-level = 0` is a multi-minute compile. Under `nextest`'s
+default output capture that build is buffered, so the first test looks frozen for several minutes
+before it does anything.
+
+Point the suite at a prebuilt binary to avoid the in-test build:
+
+```
+make test-e2e          # builds the binary once, then runs the suite with TN_BIN_PATH set
+```
+
+or run the raw command yourself (for example from an IDE test runner):
+
+```
+cargo build --bin telcoin-network --features tn-storage/test-utils
+TN_BIN_PATH="$(pwd)/target/debug/telcoin-network" \
+  cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features
+```
+
+`TN_BIN_PATH` must be an absolute path: `nextest` runs each test with its working directory set
+to the package (`crates/e2e-tests`), not the workspace root, so a relative path would not resolve.
+If `TN_BIN_PATH` is left unset the suite still runs; add `--no-capture` to watch the one-time build
+instead of waiting on a silent first test.
+
 ## Test Log Output
 
 The e2e integration tests (restart and epoch tests) spawn multiple validator node processes. Each node's stdout is captured to a separate log file under `test_logs/` so that failures can be debugged without sifting through interleaved output from all nodes.
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -397,19 +397,35 @@ pub fn get_telcoin_network_binary() -> &'static TestBinary {
             return TestBinary::Prebuilt(path);
         }
 
+        // TN_BIN_PATH is unset, so build the node binary in-process via escargot. At
+        // opt-level 0 this is a multi-minute compile, and nextest captures stdout/stderr, so
+        // without this notice the first e2e test looks frozen until the build finishes. Announce
+        // it on stderr (shown on completion, and live under `--no-capture`) so the delay is
+        // attributable rather than an invisible hang.
+        eprintln!(
+            "e2e: TN_BIN_PATH not set; building telcoin-network via cargo before the first test. \
+             This can take several minutes at opt-level 0, and nextest hides it until the build \
+             finishes (add `--no-capture` to watch it). To skip it, run `make test-e2e`, or \
+             prebuild with `cargo build --bin telcoin-network --features tn-storage/test-utils` \
+             and export TN_BIN_PATH=\"$(pwd)/target/debug/telcoin-network\" from the workspace root."
+        );
         info!("building main binary for e2e tests");
         let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR not set");
         let path = PathBuf::from(manifest_dir);
         let workspace_root =
             path.parent().and_then(|p| p.parent()).expect("Cannot find workspace root");
 
+        // No `.current_target()`: passing `--target <triple>` would emit into
+        // `target/<triple>/debug/`, a tree that shares no artifacts with the plain
+        // `target/debug/` that `make build-e2e-bin` and ordinary `cargo build` populate, forcing
+        // a guaranteed cold rebuild. Building into `target/debug/` lets escargot reuse an already
+        // compiled binary (reported "Fresh" by cargo) instead.
         TestBinary::Cargo(
             CargoBuild::new()
                 .bin("telcoin-network")
                 .features("tn-storage/test-utils")
                 .manifest_path(workspace_root.join("Cargo.toml"))
                 .target_dir(workspace_root.join("target"))
-                .current_target()
                 .run()
                 .expect("Failed to build telcoin-network binary"),
         )
--- a/etc/test-and-attest.sh
+++ b/etc/test-and-attest.sh
@@ -98,8 +98,13 @@ echo "clippy for workspace: default and all features passed"
 #
 # default features
 cargo nextest run --workspace --no-fail-fast
-# run the e2e restart and epoch tests, they are seperate to avoid any port/node confusion
-cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features
+# run the e2e restart and epoch tests, they are seperate to avoid any port/node confusion.
+# Prebuild the node binary once into the shared target tree and hand it to the e2e tests via
+# TN_BIN_PATH (mirroring `make test-e2e`), so the ignored suite reuses it instead of cold-building
+# the binary inside the first test, which nextest capture would otherwise hide as a multi-minute hang.
+cargo build --bin telcoin-network --features tn-storage/test-utils --target-dir "$(pwd)/target"
+TN_BIN_PATH="$(pwd)/target/debug/telcoin-network" \
+    cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features
 
 echo "all checks passed - submitting attestation on-chain..."
 


grantkee (approved):
LGTM!

sstanfield (approved):

PR #929, opened 2026-07-15
fix(consensus/network): gate gossip cheap-first and set explicit per-topic publisher policy (#898)

### LINKED ISSUE #898: network/consensus: gossip handlers run expensive per-message crypto before cheap authorization/dedup gates; establish a uniform gate-ordering and explicit per-topic publisher policy
## Summary

Several gossip topics are subscribed "accept any publisher," and the handlers that process those topics do the *expensive*, attacker-supplied-input work (a full BLS verify, or an ecrecover plus a task spawn) *before* the *cheap*, attacker-independent gates that would discard the message (topic match, committee-membership, per-message dedup). The result is a class of message-processing paths where an off-committee peer's gossip is admitted into the network layer, re-propagated, and then charged full per-message crypto by every recipient, all before the message can be shown to be irrelevant.

This issue asks for two structural conventions, applied uniformly across every gossip topic:

1. **Uniform gate-ordering:** cheap, attacker-independent checks (topic, committee-membership, dedup) run *before* expensive crypto and *before* spawning a task.
2. **Explicit per-topic publisher policy:** each topic declares whether it accepts any publisher, and topics that should only carry committee traffic are subscribed with `subscribe_with_publishers` restricted to the current committee's resolvable keys, so non-committee sources are `Reject(UnauthorizedAuthor)` *before* re-propagation.

This is the same class already addressed once by the `consensus_certs` cap work (GHSA-2r5c-c4h7-gp5h, the "verify-then-bound" ordering fix on the `ConsensusResult` handler). Specific exploitable instances on individual topics are tracked privately in security advisories and are intentionally **not** reproduced here; this issue is about the shared architectural pattern and the hardening that closes the whole class.

### 1. "Accept any publisher" subscriptions bypass author authorization before re-propagation

The network layer offers two subscribe entry points (`crates/network-libp2p/src/types.rs:645-668`):

```rust
/// Subscribe to a topic with valid publishers.
pub async fn subscribe_with_publishers(
    &self,
    topic: String,
    publishers: HashSet<BlsPublicKey>,
) -> NetworkResult<bool> { /* publishers: Some(publishers) */ }        // types.rs:648-659

/// Subscribe to a topic, any publisher valid.
pub async fn subscribe(&self, topic: String) -> NetworkResult<bool> {
    /* publishers: None */                                            // types.rs:664-668
}
```

`verify_gossip` (the gossipsub validation callback) authorizes the source against the topic's registered publisher set. When that set is `None` (the plain `subscribe()` case), `auth.is_none()` short-circuits and any signed source is `Accept`ed (`crates/network-libp2p/src/consensus.rs:1430-1441`):

```rust
// ensure publisher is authorized
if gossip.source.is_some_and(|id| {
    let bls_key = self.swarm.behaviour().peer_manager.peer_to_bls(&id);
    self.authorized_publishers.get(topic.as_str()).is_some_and(|auth| {
        auth.is_none()                                                // consensus.rs:1433
            || (bls_key.is_some()
                && auth.as_ref().expect("is some").contains(&bls_key.expect("is some")))
    })
}) {
    GossipAcceptance::Accept
} else {
    GossipAcceptance::Reject(RejectReason::UnauthorizedAuthor)        // consensus.rs:1440
}
```

Because gossipsub only re-propagates messages that validate as `Accept`, a topic subscribed with `publishers: None` re-propagates any signed peer's message to the mesh *before* the application-layer handler runs. The `UnauthorizedAuthor` rejection at `consensus.rs:1440` is exactly the gate that never fires on an accept-any topic.

### 2. Handlers verify expensive crypto before the cheap gates, and spawn one task per message

Each inbound gossip message is dispatched to a *freshly spawned task* before the handler body runs. Primary side (`crates/consensus/primary/src/network/mod.rs:1549-1564`):

```rust
/// Process gossip from committee.
fn process_gossip(&self, msg: GossipMessage, relayer: Option<BlsPublicKey>, author: Option<BlsPublicKey>) {
    // ...
    let task_name = format!("ProcessGossip-{}-{relayer_label}", msg.topic);
    // spawn task to process gossip
    self.task_spawner.spawn_task(task_name, async move {          // mod.rs:1562
        if let Err(e) = request_handler.process_gossip(&msg).await {   // mod.rs:1563
            // ...
```

Worker side is the same shape (`crates/consensus/worker/src/network/mod.rs:341-369`, spawn at `mod.rs:353`). So the task-spawn is unconditional per message, ahead of any handler-level filtering.

Within the handlers, the cheap gates come *after* the expensive crypto:

**`EpochVote` (primary) verifies the signature before the committee-membership gate, and has no dedup** (`crates/consensus/primary/src/network/handler.rs:387-412`):

```rust
PrimaryGossip::EpochVote(vote) => {
    ensure!( /* topic matches epoch_vote_topic */ , PrimaryNetworkError::InvalidTopic);
    // Verify the BLS signature
    ensure!(
        vote.check_signature(),                                       // handler.rs:396  (expensive, first)
        PrimaryNetworkError::InvalidHeader(HeaderError::PeerNotAuthor)
    );
    // Verify committee membership if the epoch record is available
    if let Some((epoch_rec, _)) =
        self.consensus_chain.epochs().get_epoch_by_hash(vote.epoch_hash).await
    {
        ensure!(
            epoch_rec.committee.contains(&vote.public_key),           // handler.rs:404  (cheap gate, second)
            /* UnknownAuthority */
        );
    }
    // ...
}
```

The BLS `check_signature()` at `handler.rs:396` (which calls `signature.verify_raw`, `crates/types/src/primary/epoch.rs:126`) runs before the `committee.contains(&vote.public_key)` gate at `handler.rs:404`. There is no per-message dedup on this topic.

**`ConsensusResult` (primary) verifies per message before the tally-map bound** (`crates/consensus/primary/src/network/handler.rs:324-360`). This is the shape already hardened by the `consensus_certs` cap:

```rust
ensure!(
    signature.verify_secure(&to_intent_message(consensus_result_hash), &key),  // handler.rs:325
    PrimaryNetworkError::UnknownConsensusHeaderCert(hash)
);
// ...
// Hard-cap the tally map so a flood of validly-signed but non-quorum
// results cannot grow it without bound. ... See GHSA-2r5c-c4h7-gp5h.
if guard.len() >= super::MAX_CONSENSUS_CERTS                                    // handler.rs:349
```

The committee-membership gate here (`committee.contains(&key)` at `handler.rs:302`) *does* run before the verify, which is the correct ordering; but the map-bound at `handler.rs:349` is downstream of the verify, and that residual is exactly what GHSA-2r5c-c4h7-gp5h documents as the prior art for this class.

**Worker txn path does ecrecover and a second task-spawn on an accept-any topic.** The `worker_txn_topic` is subscribed with the plain `subscribe()` (`crates/node/src/manager/node/start_epoch.rs:689`):

```rust
network_handle.inner_handle()
    .subscribe(tn_config::LibP2pConfig::worker_txn_topic(consensus_config.chain_id()))  // start_epoch.rs:689
    .await?;
```

Contrast the batch topic subscribed just below, which *does* restrict publishers to the committee (`start_epoch.rs:695-698`):

```rust
network_handle.inner_handle()
    .subscribe_with_publishers(
        tn_config::LibP2pConfig::worker_batch_topic(consensus_config.chain_id()),
        committee_keys.into_iter().collect(),                                 // start_epoch.rs:697
    )
    .await?;
```

The `WorkerGossip::Txn` handler arm then calls `submit_txn_if_mine` (`crates/consensus/worker/src/network/handler.rs:133-151`), which runs `recover_raw_transaction` (an ecrecover) and, on a slot match, spawns yet another task (`crates/batch-validator/src/validator.rs:89-106`):

```rust
fn submit_txn_if_mine(&self, tx_bytes: &[u8], committee_size: u64, committee_slot: u64) {
    if let Ok(recovered) = recover_raw_transaction(tx_bytes) {            // validator.rs:90  (ecrecover, first)
        if let Some(tx_pool) = &self.tx_pool {
            // ...
            self.reth_env.get_task_spawner().spawn_task(task_name, async move {  // validator.rs:99
                let _ = tx_pool.add_recovered_transaction_external(recovered).await;
                Ok(())
            });
        }
    }
}
```

The committee check in the txn arm (`self.consensus_config.authority()` at `handler.rs:140`) gates on whether the *local* node is a committee member, not on the *sender's* authorization; on an accept-any topic the ecrecover at `validator.rs:90` is reached for any peer's gossiped bytes (on nodes that pass the local `authority()` gate).

### Topic inventory at HEAD

| Topic | Subscribe call | Publisher policy | Handler order concern |
|---|---|---|---|
| `epoch_vote_topic` | `crates/node/src/manager/node.rs:855` (`subscribe`) | accept-any | BLS verify (`handler.rs:396`) before committee gate (`handler.rs:404`); no dedup |
| `consensus_output_topic` | `crates/node/src/manager/node.rs:859` (`subscribe`) | accept-any | per-message `verify_secure` (`handler.rs:325`) before map-bound (`handler.rs:349`); prior art GHSA-2r5c-c4h7-gp5h |
| `worker_txn_topic` | `crates/node/src/manager/node/start_epoch.rs:689` (`subscribe`) | accept-any | ecrecover + task spawn (`validator.rs:90,99`) with no sender-authorization gate |
| `worker_batch_topic` | `crates/node/src/manager/node/start_epoch.rs:695` (`subscribe_with_publishers`) | committee-restricted | reference for the desired pattern |

## Impact

For any accept-any topic whose handler front-loads crypto, an off-committee peer's message is (a) re-propagated to the mesh before any application check, and (b) charged full per-message crypto plus a task spawn by every recipient, all before the message can be discarded as irrelevant. This inverts the intended cost asymmetry: the cheap, attacker-independent test that would drop the message is deferred behind the expensive, attacker-supplied-input operation. The `worker_batch_topic` already demonstrates the safer arrangement, so the fix is to make the rest of the topics match that convention rather than to invent new machinery.

## Root cause

Two independent defaults compound:

- The plain `subscribe()` default (`publishers: None`) makes accept-any the *easy* path, and `verify_gossip` silently downgrades to "any signed source" when a topic's publisher set is `None` (`consensus.rs:1433`). There is no per-topic record of *why* a topic is accept-any versus restricted.
- Each handler orders its own checks ad hoc, so "verify then filter" and "filter then verify" both exist in-tree, and the task-spawn wrapper (`mod.rs:1549-1564`) is unconditional, so even a message that a handler will immediately drop still costs a spawn.

## The ask (hardening, no exploit)

**(a) Adopt a uniform gate-ordering convention.** In every gossip handler, run the cheap, attacker-independent gates before the expensive crypto and before any task spawn:

1. topic match (already first in most arms),
2. committee-membership / publisher authorization,
3. per-message dedup,
4. only then BLS verify / ecrecover.

Where feasible, move the committee/dedup filter ahead of the `spawn_task` in `process_gossip` (`crates/consensus/primary/src/network/mod.rs:1549-1564` and the worker analogue at `crates/consensus/worker/src/network/mod.rs:341-369`) so a droppable message never costs a task.

**(b) Give each topic an explicit publisher-authorization policy.** Replace bare `subscribe()` with an explicit decision per topic: accept-any only where genuinely required (document why inline), otherwise `subscribe_with_publishers` restricted to the current committee's resolvable keys, so `verify_gossip` returns `Reject(UnauthorizedAuthor)` (`consensus.rs:1440`) for non-committee sources *before* re-propagation. The `worker_batch_topic` subscription (`start_epoch.rs:695`) is the template. Re-subscribe on each committee rotation so the key set tracks the live committee.

**(c) Add per-message dedup / rate-limiting ahead of verification.** For topics that must remain accept-any, add a cheap per-message dedup (and/or a per-source rate limit) that runs before the expensive verify, in the spirit of the `consensus_certs` bound already added for `ConsensusResult` (GHSA-2r5c-c4h7-gp5h). `epoch_vote_topic` currently has no dedup at all (`handler.rs:387-412`).

**Prior art.** The `consensus_certs` map cap (`crates/consensus/primary/src/network/handler.rs:332-360`, GHSA-2r5c-c4h7-gp5h) is the same class solved once for one topic. This issue generalizes that fix into a convention so new topics inherit it by default.

## Checklist

- [ ] Introduce an explicit per-topic publisher policy (a single place that maps each topic to accept-any vs committee-restricted, with a rationale comment for every accept-any entry).
- [ ] Migrate `epoch_vote_topic` (`node.rs:855`), `consensus_output_topic` (`node.rs:859`), and `worker_txn_topic` (`start_epoch.rs:689`) to `subscribe_with_publishers` with the current committee's keys, unless a documented reason requires accept-any.
- [ ] Re-subscribe on each committee rotation so the authorized-publisher set tracks the live committee.
- [ ] Reorder each gossip handler so committee-membership and dedup precede BLS verify / ecrecover (`EpochVote` arm at `handler.rs:387-412`; `ConsensusResult` map-bound relative to verify at `handler.rs:325,349`; `Txn` arm at `handler.rs:133-151` / `validator.rs:89-106`).
- [ ] Add per-message dedup (and consider per-source rate-limiting) ahead of verification, especially on `epoch_vote_topic` which has none.
- [ ] Where possible, move the committee/dedup filter ahead of the `spawn_task` in `process_gossip` (primary `mod.rs:1562`, worker `mod.rs:353`) so droppable messages never cost a task.
- [ ] Add a lint/test asserting no new gossip topic is subscribed with bare `subscribe()` without an explicit policy entry.

## Acceptance criteria

- Every gossip topic has an explicit, documented publisher policy; no topic is accept-any by accident of using the bare `subscribe()` default.
- On every committee-restricted topic, a non-committee source is `Reject(UnauthorizedAuthor)` at `verify_gossip` (`consensus.rs:1440`) before re-propagation, so no expensive crypto or task spawn runs for it.
- In every gossip handler, committee-membership and per-message dedup are evaluated before BLS verify / ecrecover; a message that fails a cheap gate never reaches the expensive operation.
- `epoch_vote_topic` performs a dedup check before `vote.check_signature()`.
- A regression test drives an off-committee source on each restricted topic and asserts rejection before verification, and a test asserts the gate-ordering (cheap-before-expensive) for each handler arm.

## Reachability caveats

- The re-propagation-before-checks property depends on gossipsub only forwarding `Accept`ed messages; that is the standard gossipsub validation contract and matches `verify_gossip`'s `Accept`/`Reject` return, but the exact forwarding timing was reasoned from the callback contract, not re-derived from the gossipsub internals.
- The `ConsensusResult` handler already gates committee-membership before the verify (`handler.rs:302` before `:325`); its residual is only the map-bound sitting downstream of the verify, which is the narrow slice GHSA-2r5c-c4h7-gp5h already tracks. It is included here for the ordering convention, not as a new unbounded-growth claim.
- The `worker_txn_topic` handler's `authority()` check (`handler.rs:140`) gates on the *local* node's committee status, so on a non-committee (non-CVV) node the ecrecover path is short-circuited by the `if let Some(authority)` arm; the concern is sharpest on committee nodes and on the re-propagation cost borne by all recipients. This asymmetry was read from the handler arm, not exhaustively traced through every caller.
- The magnitude of the per-message cost (one BLS verify or one ecrecover plus one or two task spawns) is small individually; the concern is the inverted ordering and the accept-any admission, not a single expensive call. Concrete exploitability of individual topics is assessed privately.

MavenRain:
Closes #898.

### Problem
Consensus gossip handlers ran expensive per-message cryptography before cheap, attacker-independent gates, and committee-only topics used accept-any subscriptions:
- `EpochVote` verified the BLS signature before checking committee membership and had no dedup; its membership check was skipped when the epoch record was unavailable.
- `epoch_vote_topic` and `consensus_output_topic` were subscribed with a plain `subscribe()` (any publisher), so non-committee messages were accepted and re-propagated at the network layer before being dropped.

### Changes
1. **EpochVote gate reorder** (`primary/src/network/handler.rs`): topic → committee membership → dedup → BLS verify → forward.  Membership resolves via `get_committee(vote.epoch)` (the same committee source `manage_epoch_votes` uses);  unknown-committee votes are dropped without writing request state from the unauthenticated epoch field.
2. **Bounded epoch-vote dedup cache** (`handler.rs` + `network/mod.rs`): keyed on `(author, epoch, epoch_hash)`, recorded only after the signature verifies.  The BLS signature does not cover `vote.epoch`, so including `epoch` in the key prevents a Byzantine committee publisher from censoring a real author's vote by replaying it with a tampered epoch.  LRU-capped at `MAX_EPOCH_VOTES`.
3. **Committee-restricted, rotation-correct subscriptions** (`node.rs` + `node/start_epoch.rs`): `epoch_vote_topic` and `consensus_output_topic` move from `run()` into `spawn_primary_network_for_epoch` (per-epoch, all node modes) and use `subscribe_with_publishers`.  `verify_gossip` now drops non-committee publishers before re-propagation.  The two boundary topics authorize **current ∪ previous** committee keys, since epoch-close votes and an epoch's final consensus output are authored by the outgoing committee and gossipped into the next epoch . . . a current-only allowlist would reject those in-flight messages during rotation and stall certification of the just-closed epoch.
4. **verify_gossip cleanup** (`network-libp2p/src/consensus.rs`): removes two guarded `.expect` in favor of `is_none_or`, preserving the exact absent→reject / open→accept / restricted→membership semantics.

The `worker_txn_topic` ecrecover concern the issue also flagged is moot on current main: #820 removed that gossip path (non-validator txns now go to the committee via RPC).

### Testing
- `cargo fmt` + `cargo clippy` (`-p tn-primary -p tn-network-libp2p -p tn-node`): 0 errors, 0 warnings.
- The two security-sensitive paths (dedup-poisoning resistance and rotation handover) were reviewed adversarially.

### Follow-up
- Handler-level unit tests for the epoch-vote dedup / poison-resistance.

Refs GHSA-2r5c-c4h7-gp5h.

### DIFF
--- a/crates/node/src/manager/node/start_epoch.rs
+++ b/crates/node/src/manager/node/start_epoch.rs
@@ -493,6 +493,15 @@ where
             .collect();
         let next_committee_keys: HashSet<BlsPublicKey> =
             consensus_config.next_committee_keys().iter().copied().collect();
+        // Publishers authorized for the epoch-boundary topics (`epoch_vote_topic`,
+        // `consensus_output_topic`): the current committee UNION the previous one. Epoch-close
+        // votes and an epoch's final consensus output are authored by the OUTGOING committee and
+        // gossipped into the next epoch, so a current-committee-only allowlist would reject those
+        // in-flight boundary messages during rotation (and stop re-propagating them), stalling
+        // certification of the just-closed epoch. Built before `previous_committee_keys` is moved
+        // into `init_network_for_epoch` below. Never-committee peers are still excluded.
+        let boundary_publishers: HashSet<BlsPublicKey> =
+            committee_keys.iter().chain(previous_committee_keys.iter()).copied().collect();
         Self::init_network_for_epoch(
             network_handle.inner_handle(),
             bootstrap_peers,
@@ -514,12 +523,32 @@ where
             network_handle.inner_handle().start_listening(primary_address).await?;
         }
 
-        // update the authorized publishers for gossip every epoch
+        // Update the authorized publishers for gossip every epoch. `primary_topic`,
+        // `epoch_vote_topic` and `consensus_output_topic` are all committee-only publish topics:
+        // restricting the publisher set makes the network layer (`verify_gossip`) drop messages
+        // from non-committee sources before re-propagation, and re-subscribing here every epoch
+        // refreshes the allowlist across committee rotation (the swarm overwrites the previous
+        // set). `primary_topic` uses the current committee; the two boundary topics use the
+        // current-union-previous set (see `boundary_publishers` above). See issue #898.
         network_handle
             .inner_handle()
             .subscribe_with_publishers(
                 tn_config::LibP2pConfig::primary_topic(consensus_config.chain_id()),
-                committee_keys.into_iter().collect(),
+                committee_keys,
+            )
+            .await?;
+        network_handle
+            .inner_handle()
+            .subscribe_with_publishers(
+                tn_config::LibP2pConfig::epoch_vote_topic(consensus_config.chain_id()),
+                boundary_publishers.clone(),
+            )
+            .await?;
+        network_handle
+            .inner_handle()
+            .subscribe_with_publishers(
+                tn_config::LibP2pConfig::consensus_output_topic(consensus_config.chain_id()),
+                boundary_publishers,
             )
             .await?;
 
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -62,6 +62,10 @@ type AuthEquivocationMap = HashMap<
 /// that have signed this result. See the `consensus_certs` field and GHSA-2r5c-c4h7-gp5h.
 type ConsensusCertTally = (u64, HashSet<BlsPublicKey>);
 
+/// Dedup key for an epoch vote at ingress: `(author, epoch, epoch-record hash)`. See the
+/// `epoch_votes_seen` field and issue #898.
+type EpochVoteKey = (BlsPublicKey, Epoch, EpochDigest);
+
 /// The type that handles requests from peers.
 #[derive(Clone, Debug)]
 pub(crate) struct RequestHandler<DB> {
@@ -83,6 +87,17 @@ pub(crate) struct RequestHandler<DB> {
     auth_last_vote: Arc<AuthEquivocationMap>,
     /// Track consensus headers until we hit a simple quorum then send on.
     consensus_certs: Arc<Mutex<HashMap<ConsensusResultDigest, ConsensusCertTally>>>,
+    /// Deduplicate epoch votes at ingress, keyed on `(author, epoch, epoch-record hash)`.
+    ///
+    /// The value is a monotonic recency sequence used to evict the least-recently-seen entry
+    /// once the map reaches [`MAX_EPOCH_VOTES`](super::MAX_EPOCH_VOTES), bounding memory
+    /// against a vote flood. A key is inserted only after its signature verifies, so this gate
+    /// skips repeated verifies of an already-accepted vote without letting a bad-signature vote
+    /// poison the slot of the real author. The `epoch` field is part of the key even though it is
+    /// not covered by the vote signature: this ensures a replay that tampers with `vote.epoch`
+    /// lands on a different key and cannot evict/pre-empt the real vote's slot. See the
+    /// `PrimaryGossip::EpochVote` handler and issue #898.
+    epoch_votes_seen: Arc<Mutex<HashMap<EpochVoteKey, u64>>>,
     /// Access to the consensus chain data.
     consensus_chain: ConsensusChain,
 }
@@ -111,6 +126,7 @@ where
             requested_parents: Default::default(),
             auth_last_vote: Arc::new(auth_last_vote),
             consensus_certs: Default::default(),
+            epoch_votes_seen: Default::default(),
             consensus_chain,
         }
     }
@@ -210,6 +226,38 @@ where
         } else {
             self.consensus_chain.epochs().get_committee_keys(epoch).await
         }
+    }
+
+    /// True if a vote for this `(author, epoch, epoch-record)` has already been verified and
+    /// forwarded.
+    ///
+    /// Cheap read used as a pre-verify gate: an accepted vote is re-gossiped by its author on a
+    /// timer and duplicated by the gossip mesh, so this caps repeated BLS verifies of the same
+    /// valid vote. See [`RequestHandler::record_epoch_vote`] and issue #898.
+    fn epoch_vote_seen(&self, author: BlsPublicKey, epoch: Epoch, epoch_hash: EpochDigest) -> bool {
+        self.epoch_votes_seen.lock().contains_key(&(author, epoch, epoch_hash))
+    }
+
+    /// Record a verified `(author, epoch, epoch-record)` vote so later replays are dropped before
+    /// the signature check.
+    ///
+    /// Callers MUST invoke this only after [`EpochVote::check_signature`] succeeds: recording a
+    /// bad-signature vote would let a Byzantine committee publisher poison the slot of the real
+    /// author and suppress their vote. Eviction is least-recently-seen once the map reaches
+    /// [`MAX_EPOCH_VOTES`](super::MAX_EPOCH_VOTES), so a flood of distinct keys cannot grow it
+    /// without bound; a false negative after eviction only costs one extra verify (the collector
+    /// dedups per signer downstream), never a dropped honest vote.
+    fn record_epoch_vote(&self, author: BlsPublicKey, epoch: Epoch, epoch_hash: EpochDigest) {
+        let key = (author, epoch, epoch_hash);
+        let mut guard = self.epoch_votes_seen.lock();
+        let next_seq = guard.values().copied().max().unwrap_or(0) + 1;
+        if guard.len() >= super::MAX_EPOCH_VOTES && !guard.contains_key(&key) {
+            let evict = guard.iter().min_by_key(|entry| *entry.1).map(|entry| *entry.0);
+            if let Some(evict) = evict {
+                guard.remove(&evict);
+            }
+        }
+        guard.insert(key, next_seq);
     }
 
     /// Process gossip from the committee.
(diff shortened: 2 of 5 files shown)


sstanfield (dismissed):
LGTM!

sstanfield (approved):

crates/node/src/manager/node/start_epoch.rs
```diff
@@ -493,6 +493,15 @@ where
...
         let next_committee_keys: HashSet<BlsPublicKey> =
             consensus_config.next_committee_keys().iter().copied().collect();
+        // Publishers authorized for the epoch-boundary topics (`epoch_vote_topic`,
+        // `consensus_output_topic`): the current committee UNION the previous one. Epoch-close
+        // votes and an epoch's final consensus output are authored by the OUTGOING committee and
+        // gossipped into the next epoch, so a current-committee-only allowlist would reject those
+        // in-flight boundary messages during rotation (and stop re-propagating them), stalling
+        // certification of the just-closed epoch. Built before `previous_committee_keys` is moved
+        // into `init_network_for_epoch` below. Never-committee peers are still excluded.
+        let boundary_publishers: HashSet<BlsPublicKey> =
```
grantkee, line 503: This is great

grantkee (approved):
Good work!

PR #932, opened 2026-07-15
Use new pack format to avoid issues with a malicious peer.

sstanfield:
Use the new pack ordering to verify the consensus header upfront when possible.  This sidesteps buffer sizes for consensus output bytes as well as making it harder for a peer to hang up a node with bad output.


crates/consensus/primary/src/network/mod.rs
```diff
@@ -86,13 +86,6 @@ pub(crate) const MAX_CONSENSUS_CERTS: usize = 20;
 /// Prevents a single malicious peer from filling all global slots.
 pub const MAX_PENDING_REQUESTS_PER_PEER: usize = 2;
 
-/// Maximum bytes the client will read from a single consensus-output stream.
-///
-/// Generous upper bound: comfortably above any realistic single output, but caps an
-/// unbounded or malicious stream.
-/// TODO- replace with a size from consensus.  See Issue 782.
-const MAX_CONSENSUS_OUTPUT_STREAM_BYTES: usize = 512 * 1024 * 1024;
```
MavenRain, line 121: Without this constant, there is only a per-record cap.  A committee member holding a real `max-batch-count` output can drive a large receive-side allocation.

crates/storage/src/consensus_pack.rs
```diff
@@ -1387,7 +1460,27 @@ pub async fn bytes_to_output<R: AsyncRead + Unpin>(
         AsyncPackIter::<PackRecord, R>::open_partial(stream, compression, PACK_VERSION)
             .await
             .map_err(|e| PackError::ReadError(e.to_string()))?;
-    iter_to_output(&mut stream_iter, timeout, committee).await
+    iter_to_output(&mut stream_iter, timeout, committee, HeaderExpectation::None).await
+}
+
+/// Take an async (v1, header-first) stream of pack-encoded ConsensusOutput bytes and return the
+/// ConsensusOutput, verifying the header's digest equals `expected_digest` the instant it is read —
+/// BEFORE any batch record is buffered. Used on the requested-output receive path, where the
```
MavenRain, line 1450: This claim is false for stream_import's v0 path.

The fix is either narrowing the doc claim to v1, or capping batch count before the header check.

crates/storage/src/consensus.rs
```diff
@@ -879,6 +879,38 @@ impl ConsensusChain {
...
+        &self,
+        epoch: Epoch,
+        reader: R,
+        expected_hash: ConsensusHeaderDigest,
+    ) -> Result<ConsensusOutput, ConsensusChainError> {
+        let pack = self.current_pack();
+        if epoch == pack.epoch() {
+            return Ok(pack.decode_output_stream(reader, expected_hash).await?);
+        }
+        if let Ok(pack) = self.get_static(epoch).await {
```
MavenRain, line 913: This can lead to a retry storm on errors that are not `NotFound`.

Falling through to `NoCurrentEpoch`, the caller treats as "not _yet_ decodable" and retries forever.

crates/storage/src/consensus_pack.rs
sstanfield: Not sure I follow, this function only works with v1 streams and will fail on a v0.  So I think the docs are correct unless I am missing something (?).  Individual consensus outputs will only be sent over the wire in v1 format even if read from a v0 pack file in order to allow the header first verification and to simplify the decode codepaths.

crates/consensus/primary/src/network/mod.rs
sstanfield: Not sure I follow.  Are you saying that a real large consensus output will cause a large allocation?  If so I don't think that can be avoided, there are plenty of places we have to have an inflated consensus output so the compressed byte encoding should not be a problem.  Am I missing something?  All this cap would do (I think) is add an unnecessary cap to the output bytes and stop a legit large output from being received if that ever came up.

crates/storage/src/consensus.rs
sstanfield: Good catch, working on this one.

sstanfield: @MavenRain I addressed this by checking for the pack upfront.  This will also pre-load the cache if needed however if we are doing this then we have to have a writable pack eventually so that's current or staging.  But this will error out early now if this happens vs loop forever on a local issue.

crates/consensus/primary/src/network/mod.rs
MavenRain: You're right, and my original note was off.  Re-reading the new receive path,
   the byte cap has nothing left to bound and re-adding one would only risk
   rejecting a legitimately large output.
  
   Two things make the 512MB constant moot now:
   - The success type went from Fetched(Vec<u8>) to Fetched(ConsensusOutput),
   and sync_pack_reader reassembles incrementally (an unfold + StreamReader
   yielding one Data frame at a time, each bounded by MAX_SYNC_PACK_FRAME_SIZE).
   There is no whole-output byte buffer for a byte cap to sit on anymore.
   - The real bound is header-first verification plus the batch-count cap, not a
   byte count.  `iter_to_output` reads the Consensus header first, runs
   `check_header_expectation` (Digest arm) at consensus_pack.rs:1509 before the
   batch loop, and rejects expected_digest_count > MAX_BATCHES_PER_OUTPUT at
   1535-1537 before buffering any batch.  On this path the header digest must
   equal the already-verified `expected_hash`, so a peer can't make you buffer
   more than the genuine output's batches.
  
   So my "only a per-record cap" was wrong: the enforced bound is (verified
   batch count, capped at MAX_BATCHES_PER_OUTPUT = 1000) times the per-record
   cap, and that count cap already predates this PR.  I agree the flat 512MB const
   (the Issue-782 placeholder) was the wrong tool;  dropping it is fine.
  
   One non-blocking aside, not for this PR: that product is the actual
   worst-case peak (1000 × MAX_RECORD_SIZE ≈ 16 GiB).  It predates this PR and
   isn't attacker-inflatable past the real output, so nothing to do here, but if
   peak memory ever needs tightening, those two caps are the knob rather than a byte cap.

crates/storage/src/consensus_pack.rs
MavenRain: That's my bad . . . you're correct.
   
   `bytes_to_verified_output` is genuinely v1-only: it decodes through
   `iter_to_output`, which rejects a leading Batch record (1498-1500) and runs the
   header check (1509) before buffering anything, so nothing is read out of
   order on this function.
  
   The v0 "batches before header" case I had in mind is real, but it lives in a
   different function, `iter_to_output_legacy` (it buffers batches and only checks
   the header when the trailing Consensus record arrives).  That path is
   wire-reachable only through `stream_import`, the whole-epoch-pack import, which
   doesn't make the "before any batch is buffered" claim.  As you said,
   single outputs are always v1 on the wire: get_consensus_output_bytes
   re-encodes a v0 pack header-first before transmission (442-457), so
   `bytes_to_verified_output` only ever sees v1 bytes, and
   `test_v0_output_served_as_v1_bytes` pins exactly that.  The doc is correct as
   written.
  
   Only nit, take it or leave it: the v0 rejection here is a consequence of
   record ordering (iter_to_output refusing a leading batch) plus the serve-side
   v1 re-encoding, not an explicit version gate in `open_partial` (which just
   stores the passed version).  That's fine today, although it might be worth a half-sentence in the
   doc so a future refactor that feeds unnormalized bytes in doesn't quietly lose the guarantee.

crates/storage/src/consensus.rs
MavenRain: Thanks, the upfront `contains_decode_epoch` check is a real improvement: it
   stops the inner probe from hammering `MAX_EPOCH_SYNC_PROBES` peers and failing
   to decode each time when we don't hold the pack, and it warms the cache as a bonus.
  
   One thing though . . . I don't think it fully closes the "retries forever" part,
   because the loop that actually spins is one level up.  `get_consensus_output`
   maps every Err from `request_consensus_output` (including the new hard
   RPCError) to `ConsensusHeaderResult::Retry` (state-sync/consensus.rs:74), and
   the loop in catch_up... only breaks on Continue/Done;  on Retry it just sleeps
   (backoff capped at 5s) and re-enters (375-397).  So a
   persistently-undecodable local pack still loops indefinitely, just more
   cheaply per iteration.  The early bail made each spin cheaper, not finite.
  
   The underlying thing my original note was getting at is still there too:
   `contains_decode_epoch` collapses any `get_static` error to false without
   inspecting it (~922-931), and get_static surfaces PackError, which has no
   NotFound variant (1839-1875), only IO, Open, CorruptPack, etc.  So a transient
   IO hiccup or a CorruptPack looks identical to genuine absence.  Practical
   effect: a transient local error produces a hard "we do not contain a pack
   file" that the outer loop then retries forever even though it might have
   self-healed, and a persistent corrupt pack becomes a permanent no-progress
   loop with a misleading message.
  
   If you want to close it cleanly: have `contains_decode_epoch` (or the caller)
   distinguish absence from transient/corrupt, e.g. surface the PackError kind,
   so genuine absence can bail hard while transient stays retryable, and/or give
   the outer Retry path a terminal condition so no local state can spin the
   loop forever.  I'd be happy to look at a follow-up, either here or in a subsequent PR.

MavenRain (dismissed):
Suggestions are not blocking . . . follow-up PRs can address

crates/consensus/primary/src/network/mod.rs
grantkee: This was an aggregate ceiling on one consensus-output download for defense in-depth. The new unverified prefix is one 16 MiB header record, and post-verification batches are digest-gated to the verified header, so a peer cannot inflate beyond the real output's size. Without this cap, the residual risk is a genuinely large (or Byzantine-authored-but-valid - aka theoretically impossible) output causing a memory spike with no backstop. I think this is probably okay to remove, although having a ceiling is a best practice

grantkee (dismissed):

grantkee (approved):

MavenRain (approved):

crates/storage/src/consensus.rs
sstanfield: @MavenRain I am going to merge (does not seem worse off) then look at this.  I don't disagree in general but one thing worth noting, if are trying to decode output for an epoch that is not writable (either the current pack or stash) then we have BIG problems that will manifest in other places as well.  It really only checks the static as a best effort thing so I don't think that is a big deal.  But I do agree we need to sort out this error condition but I don't think this will be some transient issue type thing but rather the node is in a very bad state sort of thing.

sstanfield: @MavenRain Actually the last comment might not be 100% accurate with consensus cache in play but I do think not being able to decode indicates a serious issue that may not resolve.  Have to dig into this deeper.

PR #936, opened 2026-07-16
test(e2e): floor the epoch-boundary cert deadline above the new-validator quorum window

### LINKED ISSUE #931: `test`: cut `epochs::test_epoch_sync` (and `test_epoch_boundary`) ~45% by isolating them and halving the epoch duration to the consensus floor (no coverage loss)
## Summary

`epochs::test_epoch_sync` is one of the two slowest e2e integration tests in
`crates/e2e-tests`.  Its wall time is almost entirely real network progression
(it drives 15 consensus epochs), not padded sleeps, so there are exactly two
places to get time back without changing what it asserts:

1. **Isolation.** The test is marked `#[ignore = "only run independently from all
   other it tests"]`, yet the e2e scheduler still lets two heavy tests run at
   once, so it co-schedules with `test_epoch_boundary` and the two fight for CPU.
2. **Epoch duration.** The per-epoch cadence is a `const EPOCH_DURATION = 10`
   seconds, but epoch close is wall-clock driven and the consensus floor is ~5s.
   Halving it roughly halves the 15-epoch progression.

Both changes are test-only, preserve every epoch count, iteration count and
assertion, and touch no production code.  `EPOCH_DURATION` is a file-level const
shared by both `epochs.rs` tests, so Change 2 also speeds `test_epoch_boundary`
and must carry that test's deadline floor too;  the scope below is honest about
that.

This is disjoint from #894 (unit/integration tests in other crates).  It is a
narrower, ship-ready subset of the broader e2e-speedup analysis: it takes only
the two highest-value, lowest-risk levers and leaves the restart-test levers out.

## Baseline

Branch `bump-e2e-slow-timeout-720s` (HEAD `3ef12ce3` = `main` plus the CI
slow-timeout bump;  the test file is untouched by that commit, so re-confirm line
numbers against `main` before the PR).

`test_epoch_sync` solo happy-path runtime is ~165 to 190s (~3 min), broken down
from `crates/e2e-tests/tests/it/epochs.rs`:

| Step | Cost | Line |
|---|---|---|
| RPC-ready wait (cap 20s, usually a few s) | ~few s | :303-315 |
| Initial first-epoch `sleep(EPOCH_DURATION + 1)` | 11s | :317 |
| Three `loop_epochs(_, 5)` = 15 real epoch boundaries at ~10s each | ~150s | :319, :330, :339 |
| One node kill + restart via `start_nodes` | ~few s | :333 |
| Final per-endpoint epoch-record verify (fast on happy path) | ~s | :347-398 |

The `EPOCH_DURATION * 6` (60s) and `EPOCH_DURATION * 4` (40s) deadlines are
failure/lag caps: they cost nothing on a green run.  The single dominant term is
`15 epochs x EPOCH_DURATION`, which is why the two levers below are duration and
contention.

Contended, the test has been observed far higher (~527s reported) because
co-scheduling with `test_epoch_boundary` puts up to `2 x 6 = 12` node processes
on the same cores.  Isolation returns it to the ~165 to 190s floor; the duration
cut then lowers that floor.

Numbers are source-derived cadence arithmetic plus a prior isolated measurement
(~165s solo, `test_epoch_boundary` ~112s solo);  confirm on the PR machine.

## Change 1: isolate the heavy epoch tests (scheduling only, zero coverage risk)

Both epoch tests already declare they must run alone:

- `crates/e2e-tests/tests/it/epochs.rs:404` and `:442`: `#[ignore = "only run
  independently from all other it tests"]`.

But two layers both still permit two concurrent heavy tests:

- `.config/nextest.toml:20`: `[test-groups.e2e] max-threads = 2` over
  `package(e2e-tests)` (slow-timeout `period = "720s", terminate-after = 3`,
  `:25`).
- `crates/e2e-tests/tests/it/common.rs:36`: `const MAX_CONCURRENT_TESTS: usize =
  2;`, the in-process semaphore (`:38`) every heavy test gates on via
  `acquire_test_permit()` (`epochs.rs:408,:446`).

**Fix:** set `MAX_CONCURRENT_TESTS = 1` in `common.rs:36`, and align the scheduler
with `max-threads = 1` in `.config/nextest.toml:20`.  The semaphore is the robust
one: it is process-internal, so it serializes the heavy set no matter how the
tests are launched.

This is the isolation the `#[ignore]` messages already ask for.  It only
serializes tests that call `acquire_test_permit()` (the node-spawning epoch,
restart, sync, genesis and staking tests);  the cheap `faucet` test does not
acquire a permit and is unaffected.  Trade-off to note in the PR: on a many-core
runner this gives up a little aggregate parallelism among the heavy tests in
exchange for per-test determinism, which is precisely what these tests were
annotated to want.

If a full-group serialization is undesirable, the equivalent runner-only form is
a dedicated serial invocation, e.g. `cargo nextest run -p e2e-tests
--run-ignored ignored-only -E 'test(test_epoch_sync)'`, with no source edit.

## Change 2: halve `EPOCH_DURATION` to the consensus floor, with mandatory deadline floors

`epoch_boundary = epoch_start + epochDuration` and closes on the first commit
whose timestamp crosses it (`crates/node/src/manager/node/run_epoch.rs:135,:507`;
`committed_at` is whole-second, `crates/types/src/primary/output.rs:174-175`).
The only duration validation anywhere is a Solidity `== 0` revert
(`tn-contracts/src/consensus/ConsensusRegistry.sol:1092`), so there is no
code-level floor above 5s.  Under the test's own header-cadence overrides
(`--max-header-delay-ms 1000 --min-header-delay-ms 500`, `epochs.rs:593-596`) a
5s epoch still spans ~5 to 10 consensus rounds, with multiple leader-commit
opportunities: enough to certify and close cleanly.

`EPOCH_DURATION` reaches genesis through `--epoch-duration-in-secs
&EPOCH_DURATION.to_string()` (`epochs.rs:589-590`), so the on-chain value and the
`assert_eq!(epochDuration, EPOCH_DURATION)` checks (`:282`) stay self-consistent.

**Primary edit:**

- `crates/e2e-tests/tests/it/epochs.rs:36`: `const EPOCH_DURATION: u64 = 10;` ->
  `= 5;`.

**Mandatory companion edits (the change is unsafe without them):** two deadlines
are sized to clear a fixed ~62.5s absolute timeout, not a multiple of epoch
length, so they must be floored absolutely:

- `epochs.rs:363` (sync path): `Duration::from_secs(EPOCH_DURATION * 6)` ->
  `Duration::from_secs((EPOCH_DURATION * 6).max(65))`.
- `epochs.rs:150` (boundary path, fires only after the mid-test validator
  shuffle): `Duration::from_secs(EPOCH_DURATION * 3)` ->
  `Duration::from_secs((EPOCH_DURATION * 3).max(35))`.

**Why the floor is not optional.** When a new validator joins the committee
mid-test, its epoch-vote quorum collection can run a fixed ~62.5s before the
failed-quorum P2P fallback succeeds, and that window does not scale with epoch
duration:

- `crates/node/src/manager/epoch_votes.rs:43`: `manage_epoch_votes` takes no
  duration parameter.
- `:77`: `let mut timeout = Duration::from_millis(2500);` (fixed 2.5s per attempt).
- `:132`: `if reached_quorum || timeouts > 24 {` (breaks only after 25 attempts,
  i.e. 25 x 2.5s = 62.5s).
- P2P fallback then retries via `EPOCH_COLLECT_RETRY_SECS = 5`
  (`crates/state-sync/src/epoch.rs:16`).

At `EPOCH_DURATION = 5` a bare `* 6` deadline is 30s, below 62.5s, so the
new-validator epoch record may legitimately not be certified yet when the
deadline fires.  That is exactly the flake the existing comment at `epochs.rs:355-361`
says the `* 6` sizing was added to prevent, and it is the core behavior
`test_epoch_sync` exists to verify, so dropping below the floor would silently
delete coverage while looking like a speedup.  There is already precedent for this
exact floor pattern in the tree: `crates/e2e-tests/tests/it/sync.rs:128` uses
`(PACK_IMPORT_EPOCH_DURATION * 6).max(60)`.

**Leave as-is:** the `EPOCH_DURATION * 4` epoch-change deadlines (`epochs.rs:106,
:267`). At 5s they become 20s, still ~4x a real ~5s epoch close;  they gate
epoch-change, not the vote-quorum window, so no floor is needed.  Applying Change 1
first keeps that 20s margin comfortable by removing contention.

**Consistency edit (recommended, lower priority):** `sync.rs:17-19` documents
`const PACK_IMPORT_EPOCH_DURATION: u64 = 10;` as a deliberate mirror of
`epochs.rs::EPOCH_DURATION` "so that epoch boundaries occur on the same cadence."
Cutting `EPOCH_DURATION` without touching it makes that comment stale. Either
mirror the cut (`sync.rs:19` -> `5`, floor its two `* 3` deadlines at `:99` and
`:148` the same way, `.max(35)`;  its `* 6` at `:128` is already floored at
`.max(60)`) which also speeds the PACK_IMPORT test, or update the lockstep comment
to reflect the divergence.

## Expected result

| Test | Contended today | Solo (Change 1) | Solo + duration (Change 2) |
|---|---|---|---|
| `test_epoch_sync` | ~527s | ~165-190s | ~90-110s |
| `test_epoch_boundary` | ~431s | ~112s | ~70s |

Change 2 is ~45% off `test_epoch_sync`'s solo floor (the 11s sleep drops to 6s and
each of 15 boundary waits drops ~10s to ~5s).  Change 1 is what makes that floor
reachable rather than a contended multiple of it.

## Not worth doing / do not do

- **Cutting `EPOCH_DURATION` without the `:363`/`:150` floors.** Identical apparent
  savings, but reintroduces the ~62.5s-window flake and breaks the new-validator
  sync coverage.  This is the whole reason the floors are mandatory, not optional.
- **Touching the node-side constants** (`epoch_votes.rs:77` `2500ms`, `:132`
  `timeouts > 24`, `state-sync/epoch.rs:16` `EPOCH_COLLECT_RETRY_SECS = 5`) to
  speed the test.  These are production consensus liveness parameters, not test
  knobs;  changing them alters real node behavior and is out of scope.
- **Lowering the 20s RPC-ready cap (`:304`) or the `* 4` epoch-change deadlines
  (`:106,:267`).** These are failure caps;  they save nothing on a green run and
  only make the test flakier under load.

## Acceptance

- `test_epoch_sync` and `test_epoch_boundary` still pass, asserting the same
  conditions, over the same epoch counts (`loop_epochs` iterations at
  `epochs.rs:319,330,339`) and the same per-epoch certificate + final-block
  verification (`:384-397`).
- No `EPOCH_DURATION`-independent epoch count or assertion changes;  only the
  wall-clock spacing between boundaries shrinks.
- The mid-test new-validator sync assertion is still exercised and does not flake:
  the `:363` verify deadline stays >= ~62.5s (`.max(65)`).
- On the same machine, run via `cargo nextest run -p e2e-tests --run-ignored
  ignored-only`, `test_epoch_sync` is at least ~40% faster.

## Measurement method

`cargo nextest run -p e2e-tests --run-ignored ignored-only` (process-per-test),
warm cache, branch `bump-e2e-slow-timeout-720s` (`3ef12ce3`).  Confirm the solo
floor and the post-change numbers on the PR machine, and specifically verify the
shrunk 20s `loop_epochs` deadline (`:267`) holds under CI contention (the comment
at `epochs.rs:34-35` cites CI margin as the reason for the original 10s).

MavenRain:
Closes #931 

## What

- Raise the `test_epoch_boundary` epoch-record certificate-availability deadline
  from `(EPOCH_DURATION * 3).max(30)` to `.max(65)`.
- Correct the misattributed comment on the analogous `test_epoch_sync` deadline
  (it described a new-validator committee join that only happens in
  `test_epoch_boundary`;  `test_epoch_sync` kills/restarts a node instead). Its
  60s floor is unchanged.

## Why

#900 halved `EPOCH_DURATION` 10 -> 5 and floored these two async
cert-availability polls so they no longer shrink with the epoch cadence.  The
sync-test poll got a 60s floor with a comment justifying it via the ~62.5s
new-validator vote-quorum window (25 x 2.5s in `manage_epoch_votes`, then the
`spawn_epoch_record_collector` 5s-cadence fallback). But that window applies to
the test that actually adds a validator to the committee (`test_epoch_boundary`)
 whose analogous poll was floored at only 30s.

`test_epoch_boundary` stakes + activates + shuffles a new validator in
(`generate_new_validator_txs`), then polls `tn_epochRecord` for the new-validator
epoch on every node including the freshly-joined one.  `super_quorum` for that
size-5 committee is `(5*2)/3 + 1 = 4` . . . exactly the established validators left
after the join . . . so the epoch has no signer redundancy: if a single established
vote is slow to reach the new node, that node falls back to its own local
vote-collection loop, which can run its full 25 x 2.5s = ~62.5s timeout before the
5s-cadence record collector back-fills the cert.  30s is below that, so under CI
contention the poll can fire before the cert is available -> latent flake. 65s
clears the window.

`test_epoch_sync` is not exposed to this: it kills and restarts a node rather than
adding one to the committee, so its certs come from the stable committee and its
60s floor is already conservative. Only its comment was wrong;  corrected.

These deadlines are failure-only ceilings (they cost nothing on a green run), so
raising the boundary floor removes tail flakiness without slowing the test.

## Test plan

Run the two ignored e2e tests on the CI/PR machine (each spawns 4-6 node
processes and takes minutes):

```
cargo nextest run -p e2e-tests --run-ignored ignored-only -E 'test(test_epoch_boundary)'
cargo nextest run -p e2e-tests --run-ignored ignored-only -E 'test(test_epoch_sync)'
```

Both should still pass, asserting the same epoch counts and per-epoch certificate
verification;  only the boundary poll's failure ceiling changes.

## Note on #931

Issue #931 proposed this speedup, but its substance (`EPOCH_DURATION` 10->5, the
`.max()` floors, nextest `max-threads=1`) already landed in #900.  The only piece
of #931 not covered there is this floor-value refinement, and #931's analysis was
inverted: it asked to raise the sync floor (not exposed to the window) and left
the boundary floor at 35s (still below ~62.5s).  This PR hardens the boundary floor,
 the one assertion genuinely exposed, and leaves the safely-conservative sync
floor alone. #931 can be closed as resolved by #900 + this follow-up.


grantkee (dismissed):
Merge conflicts, but I have no issues with the code

sstanfield (dismissed):

sstanfield (approved):

PR #937, opened 2026-07-16
fix(epoch): anchor recovered epoch records to the locally-trusted committee

### LINKED ISSUE #935: epoch: failed-quorum epoch-record recovery accepts a peer's record validated only against its own self-declared committee (missing the epoch_committee_valid anchor the state-sync path applies)
## Summary

When a node fails to gather a super-quorum on its own epoch-close record, it tries to recover by downloading an `EpochRecord` and certificate from a peer.  It accepts the downloaded record if `new_epoch_rec.verify_with_cert(&cert)` returns true.  That function verifies the certificate's aggregate signature against the public keys taken from the record's own `committee` field, so a record whose `committee` is a set of attacker keys, with a certificate self-signed by those keys, passes.  The recovery path never anchors the record's committee to the locally-trusted committee for that epoch.

The parallel path that ingests epoch records during normal state-sync does anchor it: it additionally checks `parents_match` and `epoch_committee_valid(&epoch_rec, &committee)`, where `committee` is derived from the locally-trusted `prev.next_committee`.  The recovery branch drops both of those checks and keeps only `verify_with_cert`. This is a validation asymmetry: the same class of artifact is accepted under weaker rules on the recovery path.

The blast radius is bounded and self-healing (detailed below), so this is a low-severity correctness / hardening issue rather than a fork, but the recovery path should apply the same committee anchor the state-sync path does.

## The defect

**Locations:** the recovery branch at `crates/node/src/manager/epoch_votes.rs:196-224`;  the weaker check it uses at `epoch_votes.rs:200`; the stronger check on the state-sync path at `crates/state-sync/src/epoch.rs:98-103`; the self-committee verification at `crates/types/src/primary/epoch.rs:59-89`;  the unauthenticated request at `crates/consensus/primary/src/network/mod.rs:768-787`;  the anchor helper at `crates/state-sync/src/epoch.rs:22-40`.

**The recovery branch accepts on `verify_with_cert` alone** (`epoch_votes.rs:196-217`):

```rust
// Try to recover by downloading the epoch record and cert from a peer
let mut got_epoch_record = false;
for _ in 0..5 {
    match network.request_epoch_cert(Some(epoch_rec.epoch), None).await {
        Ok((new_epoch_rec, cert)) => {
            if new_epoch_rec.verify_with_cert(&cert) {
                let new_epoch_hash = new_epoch_rec.digest();
                if new_epoch_hash != epoch_hash {
                    warn!(
                        target: "epoch-manager",
                        "Over wrote expected epoch record {epoch_hash} with verified epoch record {new_epoch_hash}",
                    );
                    save_and_persist_with_logs(&db, new_epoch_rec, cert).await;
                } else {
                    ...
                    save_and_persist_with_logs(&db, new_epoch_rec, cert).await;
                }
                got_epoch_record = true;
                break;
            }
        }
        ...
    }
}
```

**`verify_with_cert` checks the record's own embedded committee** (`crates/types/src/primary/epoch.rs:59-89`):

```rust
pub fn verify_with_cert(&self, cert: &EpochCertificate) -> bool {
    if self.digest() != cert.epoch_hash {
        // Record and cert don't match.
        return false;
    }

    let auth_indexes = cert.signed_authorities.iter().collect::<Vec<_>>();
    let mut auth_iter = 0;
    let pks: Vec<BlsPublicKey> = self
        .committee
        .iter()
        ...
```

The signing keys are drawn from `self.committee`, a field of the record the peer supplied.  A self-consistent record (attacker committee plus a certificate signed by that same attacker committee, reaching `self.super_quorum()`) passes.

**The request is answered by any peer, with no committee authorization** (`crates/consensus/primary/src/network/mod.rs:768-787`):

```rust
pub async fn request_epoch_cert(
    &self,
    epoch: Option<Epoch>,
    hash: Option<EpochDigest>,
) -> NetworkResult<(EpochRecord, EpochCertificate)> {
    let request = PrimaryRequest::EpochRecord { epoch, hash };
    for _ in 0..3 {
        let res = self.handle.send_request_any(request.clone()).await?;
        if let Ok(Ok(NetworkResponseMessage {
            peer: _,
            result: PrimaryResponse::EpochRecord { record, certificate },
        })) = res.await
        {
            return Ok((record, certificate));
        }
    }
    Err(NetworkError::RPCError("Could not get the epoch record!".to_string()))
}
```

**The state-sync ingest path anchors to the locally-trusted committee** (`crates/state-sync/src/epoch.rs:98-103`), which the recovery branch omits:

```rust
// Verify the epoch has the expected parent and committee and is signed by
// that committee.
let parents_match = parent_hash == epoch_rec.parent_hash;
let epoch_committee_valid = epoch_committee_valid(&epoch_rec, &committee);
let epoch_valid = epoch_rec.verify_with_cert(&cert);
if parents_match && epoch_committee_valid && epoch_valid {
```

Here `committee` is derived from the previous epoch record's `next_committee` (`epoch.rs:88-90`), that is, the locally-trusted committee.  `epoch_committee_valid` (`epoch.rs:22-40`) requires the record's committee to match / be contained in that trusted set with a reasonable size.  The recovery branch performs neither the `parents_match` nor the `epoch_committee_valid` check.

## Impact

An unanchored, attacker-committee epoch record plus certificate is accepted as verified and persisted to local epoch storage on the recovering node, and the node treats its epoch close as recovered from that artifact.  The blast radius is bounded by two in-code mitigations:

1. The node's own authoritative record for the epoch was already saved before entering this recovery branch, and the epoch-record save path is idempotent for an already-stored record, so the authoritative bytes and the honest certificate are not overwritten in the general case.
2. In-memory epoch indices are rebuilt from the stored records on the next open, so a corrupted in-memory view does not persist a forged committee across a restart.

So this is not a committee/fork rewrite and it self-heals on restart.  It is a validation asymmetry that accepts attacker-controlled epoch data transiently and should be closed, but it is low severity as it stands.

## Root cause

The recovery-fetch path reuses the same unauthenticated `request_epoch_cert` RPC as other epoch-record consumers, but validates the response with only `verify_with_cert` (which trusts the record's self-embedded committee) and drops the `parents_match` and `epoch_committee_valid` anchors that the state-sync path applies against the locally-trusted committee.  A record must be anchored to a trusted committee, not merely self-consistent.

## Suggested fix

In the recovery branch (`epoch_votes.rs:196-217`), before accepting a downloaded record, apply the same anchoring the state-sync path uses: derive the trusted committee for `epoch_rec.epoch` from the previous epoch record's `next_committee` (or genesis for the first epoch), and require `parents_match && epoch_committee_valid(&new_epoch_rec, &trusted_committee) && new_epoch_rec.verify_with_cert(&cert)`.  Factoring the state-sync validation into a shared helper and calling it from both sites would prevent the two paths from drifting again.

## Suggested test design

- A unit test that constructs an `EpochRecord` whose `committee` is an arbitrary attacker key set, with a certificate self-signed by that set reaching `super_quorum()`, and asserts the recovery branch rejects it (it is accepted today because `verify_with_cert` alone returns true).
- A test that a legitimate record whose committee matches `prev.next_committee` is still accepted by the recovery branch, so the added anchor does not break the recovery path.

## Acceptance criteria

- The failed-quorum recovery branch anchors a downloaded epoch record to the locally-trusted committee (from `prev.next_committee` / genesis) and checks `parents_match`, matching the state-sync ingest path, before persisting it.
- The two epoch-record ingest paths share one validation routine, so a downloaded epoch record cannot be accepted under weaker rules on the recovery path than on the state-sync path.

MavenRain:
## Summary

Fixes #935.

When a node fails to gather a super-quorum on its own epoch-close record, it recovers by
downloading an `EpochRecord` + certificate from a peer.  That recovery branch
(`crates/node/src/manager/epoch_votes.rs`) accepted the downloaded record on
`new_epoch_rec.verify_with_cert(&cert)` alone.  `verify_with_cert` only proves the record is
self-consistent with its *own embedded* committee, so a record whose `committee` is a set of
attacker keys, with a certificate self-signed by those keys reaching `super_quorum()`, passed.
The recovery path never anchored the record to the locally-trusted committee for that epoch.

The parallel state-sync ingest path (`crates/state-sync/src/epoch.rs`, `collect_epoch_records`)
already anchors: it additionally checks `parents_match` and `epoch_committee_valid` against the
committee derived from the locally-trusted `prev.next_committee` (or the genesis committee).  This
was a validation asymmetry, the same artifact accepted under weaker rules on one path than the
other.

## Fix

The two ingest paths now share **one** validation routine so they cannot drift again:

- **`crates/storage/src/epoch_records.rs`** (new): `EpochRecordDb::validate_downloaded_record(epoch,
  record, cert)` derives the trusted committee and expected parent hash from *local state only*,
  anchored to the **requested** `epoch` (the genesis committee for epoch 0, otherwise the previous
  epoch record's `next_committee`), and returns an `EpochRecordValidation` enum: `Valid`,
  `Invalid { epoch_matches, parents_match, committee_valid, cert_valid }`, or `NoAnchor` (no
  locally-trusted anchor available).  It runs the same checks the state-sync path used, plus an
  explicit `record.epoch == epoch` check: `epoch_matches && parents_match && epoch_committee_valid
  && verify_with_cert`.  Anchoring against the requested `epoch` rather than the record's
  self-declared epoch is deliberate: it stops a peer from answering a request for one slot with a
  self-consistent record for a different slot.  The private `epoch_committee_valid` helper (and its
  8 unit tests) moved here from state-sync, since `tn_storage` is the crate both ingest crates
  depend on (the node crate does not depend on `state-sync`).

- **`crates/state-sync/src/epoch.rs`**: `collect_epoch_records` now calls
  `validate_downloaded_record` and matches on the result, instead of deriving the committee and
  running the checks inline.  Behavior is preserved, with one intentional improvement: the old
  genesis branch used `.expect("always can retrieve epoch 0 committee")` (a panic).  The shared
  routine degrades to `NoAnchor` (retry) instead of panicking.

- **`crates/node/src/manager/epoch_votes.rs`**: the recovery branch now gates the save on
  `db.validate_downloaded_record(epoch_rec.epoch, &new_epoch_rec, &cert).await.is_valid()` and logs
  a warning when a peer returns an unanchored record.

`state-sync`'s `rand` / `rand_chacha` dev-dependencies were dropped (only the moved tests used
them).

## Tests

- `test_validate_downloaded_record_rejects_unanchored_committee` (regression): an attacker-committee
  record with a certificate self-signed by that same committee passes `verify_with_cert` (asserted)
  but is rejected by the shared anchor (`Invalid { committee_valid: false }`).
- `test_validate_downloaded_record_rejects_wrong_epoch`: a genuine, correctly-anchored record for
  epoch 1 offered as the answer to a request for epoch 2 is rejected (`epoch_matches: false`).
- `test_validate_downloaded_record_rejects_wrong_parent_hash`: a record with the right committee and
  a valid cert but the wrong parent hash is rejected (`parents_match: false`).
- `test_validate_downloaded_record_accepts_anchored_committee` /
  `test_validate_downloaded_record_accepts_genesis_epoch`: legitimate records (normal epoch and
  genesis) are accepted.
- `test_validate_downloaded_record_no_anchor_when_prev_missing`: with no previous epoch record
  stored, the record is not accepted (`NoAnchor`).
- The 8 relocated `epoch_committee_valid` unit tests keep their coverage in `tn_storage`.

All three crates pass `clippy --all-targets` clean; `tn-storage` (141) and `state-sync` (7) test
suites pass.

## Acceptance criteria

- [x] The failed-quorum recovery branch anchors a downloaded epoch record to the locally-trusted
      committee (from `prev.next_committee` / genesis) and checks `parents_match`, matching the
      state-sync ingest path, before persisting it.
- [x] The two epoch-record ingest paths share one validation routine, so a downloaded epoch record
      cannot be accepted under weaker rules on the recovery path than on the state-sync path.


sstanfield (dismissed):
Good catch, looks good.

sstanfield (approved):

PR #938, opened 2026-07-16
fix(worker): validate gossip-prefetched batches before caching (#933)

### LINKED ISSUE #933: worker: gossip-prefetched batches are cached without validate_batch, and the vote-path already-in-store short-circuit then skips validation, so a Byzantine committee member can get a semantically-invalid batch certified (execution halt)
## Summary

A worker caches a gossip-prefetched batch into `NodeBatchesCache` after checking only its epoch and digest binding, with no call to `BatchValidation::validate_batch`.  Later, when a primary must make that batch available before voting on a header that references it, `synchronize` sees the digest "already in store" and returns success without validating it: the only `validate_batch` call on the sync path runs solely for digests that were actually missing and had to be fetched.  The store is therefore used as a proxy for "validated and available," but the gossip-prefetch path violates that by inserting unvalidated bodies into the same store.

The result is a validity gap.  A Byzantine current-epoch committee member can gossip the digest of a structurally well-formed but invalid batch (transactions that do not decode, an EIP-4844 blob transaction, total gas over the limit, wrong base fee, or empty), serve that body over `request_batches`, and have every honest node cache it unvalidated.  When a header referencing that digest is proposed, honest primaries find the batch locally present, skip validation, and vote.  Once certified, the sub-DAG is executed on every node, where transaction recovery fails fatally and the engine loop exits.  Because the certified output is fixed, every honest node that committed it hits the same fatal decode, and each restart replays it and dies again: a deterministic, network-wide chain halt with a restart crash-loop.

This requires a Byzantine committee member, which is inside the fault model the protocol must tolerate (up to `f`).  Digest binding alone is not validation: a valid digest does not imply a valid batch.

## The defect

**Locations:** the unvalidated cache insert on the gossip-prefetch path at `crates/consensus/worker/src/network/handler.rs:201-208`; the validating sibling path (for contrast) at `handler.rs:243-246`;  the "already in store" short-circuit at `crates/consensus/worker/src/network/primary.rs:37-53`;  the sole sync-path validation gate at `primary.rs:105-111`; the shared read at `crates/consensus/worker/src/batch_fetcher.rs:32-54`;  the fatal execution recover at `crates/tn-reth/src/lib.rs:766-780`;  the engine termination at `crates/engine/src/lib.rs:262-264`.

**The gossip-prefetch path caches the fetched batch with no validation** (`handler.rs:201-208`, inside `process_gossip` handling `WorkerGossip::Batch`):

```rust
if batch.epoch == self.consensus_config.epoch() {
    store.insert::<NodeBatchesCache>(digest, batch).map_err(
        |e| {
            WorkerNetworkError::Internal(format!(
                "failed to write to batch store: {e}"
            ))
        },
    )?;
}
```

The only guarantees on the fetched body are digest match, uniqueness, and count, enforced by `request_batches`.  There is no `validate_batch` call here.

**Contrast the reported-batch path, which does validate before inserting** (`handler.rs:243-246`, `process_report_batch`):

```rust
// validate batch - log error if invalid
self.validator
    .validate_batch(sealed_batch.clone())
    .inspect_err(|_| self.metrics.record_batch_validation_failure())?;
```

**The vote-path sync treats "already in store" as done, without validating** (`primary.rs:37-53`, `synchronize`):

```rust
for digest in message.digests.iter() {
    // Check if we already have the batch.
    match self.batch_fetcher.fetch_local_batch(*digest).await {
        Ok(None) => {
            missing.insert(*digest);
            debug!("Requesting sync for batch {digest}");
        }
        Ok(Some(_)) => {
            trace!("Digest {digest} already in store, nothing to sync");
        }
        Err(e) => {
            return Err(eyre::eyre!("failed to read from batch store: {e:?}"));
        }
    };
}
if missing.is_empty() {
    return Ok(());
}
```

A prefetched digest takes the `Ok(Some(_))` arm and is never added to `missing`;  if all referenced digests are already cached, `synchronize` returns `Ok(())` before any validation.

**`validate_batch` runs only for digests that were missing** (`primary.rs:105-111`, `store_synced_batches`, invoked solely for the fetched-missing set):

```rust
for sealed_batch in sealed_batches_from_response.into_iter() {
    if !is_certified {
        // This batch is not part of a certificate, so we need to validate it.
        if let Err(err) = self.validator.validate_batch(sealed_batch.clone()) {
            return Err(eyre::eyre!("Invalid batch: {err}"));
        }
    }
```

**Both paths read the same table** (`batch_fetcher.rs:32-54`, `get_batch_local`, reached from `fetch_local_batch`):

```rust
if let Some(batch) = store
    .get::<NodeBatchesCache>(&batch_digest)
    ...
{
    Ok(Some(batch))
} else {
    Ok(consensus_chain.get_batches(epoch, [batch_digest].iter()).await.into_iter().next())
}
```

So the unvalidated body inserted by the prefetch path at `handler.rs:202` is exactly what satisfies the availability check at `primary.rs:43`, which is what causes validation to be skipped.

**What is bypassed.** `validate_batch` (`crates/batch-validator/src/validator.rs:39-82`) is the only place these checks run: digest match, worker id, epoch, total byte size and non-empty, per-transaction decode and signer recovery (`decode_transactions`), no EIP-4844 blob transactions, total gas limit under the block cap (with overflow check), and base fee.  Skipping it admits a batch that violates any of these.

## Why this halts the chain

Once the header is certified, the sub-DAG is executed. `build_block_from_batch_payload` (`crates/tn-reth/src/lib.rs:766-780`) recovers every transaction before executing any, and a recover failure is fatal:

```rust
let recovered_txs = transactions
    .par_iter()
    .map(|tx_bytes| {
        reth_recover_raw_transaction::<TransactionSigned>(tx_bytes)
            .inspect_err(|e| {
                error!(
                    target: "engine",
                    batch=?batch_digest,
                    ?tx_bytes,
                    "failed to recover signer: {e}"
                )
            })
            .map_err(TnRethError::from)
    })
    .collect::<TnRethResult<Vec<_>>>()?;
```

Note the asymmetry with the execution phase that follows: `execute_transaction` deliberately tolerates an `InvalidTx` by logging and `continue`ing (`lib.rs:783-796`), but the recover phase above has no such tolerance. One undecodable transaction returns `Err`.

That error propagates into the engine loop, which treats it as fatal (`crates/engine/src/lib.rs:262-264`):

```rust
Some(res) = OptionFuture::from(self.pending_task.as_mut()), if self.pending_task.is_some() => {
    debug!(target: "engine", ?res, "reciever for execution result polled ready");
    let finalized_header = res.map_err(Into::into).and_then(|res| res)?;
```

`run()` returns `Err` and the engine stops executing consensus output.  The certified output remains in the pack, so on restart replay the same output fails identically: an unrecoverable crash-loop.  Because certification commits the sub-DAG on every honest node, this is network-wide, not local.

## Reachability

The worker batch gossip topic is publisher-restricted to committee members, so the attacker must be a current-epoch committee member.  That is inside the standard consensus fault model (the protocol must remain safe and live with up to `f` Byzantine committee members).  The attacker gossips `WorkerGossip::Batch(epoch, digest)` for an invalid batch it can also serve over `request_batches`, waits for honest workers to prefetch and cache it, and then authors (or a colluder authors) a header referencing that digest.  The prefetch entry runs on every gossip receipt, so it populates `NodeBatchesCache` ahead of the header sync.  No downstream re-validation of the batch body exists on the certified execution path, so nothing catches the invalid batch after certification.

## Root cause

`NodeBatchesCache` is treated as "validated and available," but the gossip-prefetch path writes unvalidated bodies into it, and the vote-path sync uses presence in that store as a proxy for validity. Availability and validity are conflated.  The `validate_batch` gate is bypassable for any batch that reaches the cache through the prefetch path.

## Suggested fix

- Validate on the prefetch path before insert, mirroring `process_report_batch` (`handler.rs:243-246`): call `validate_batch` on the fetched body and drop it (do not cache, and score the server) if it fails.  This closes the gap at the source.
- Or keep prefetched-but-unvalidated batches in a separate store (or mark cache entries as unvalidated) so the vote-path `synchronize` cannot treat a prefetched entry as validated, and always runs `validate_batch` for a not-yet-validated digest even when it is locally present.
- Defense in depth: make the recover phase in `build_block_from_batch_payload` (`lib.rs:766-780`) consistent with the execute phase so that a single undecodable transaction in a certified batch cannot take the whole node down (for example, treat a certified-but-invalid batch as a defined, deterministic, non-fatal outcome rather than a fatal `?`).  This bounds the blast radius even if an invalid batch is ever certified by another route.

The primary fix is closing the validation bypass;  the recover-phase hardening prevents any future bypass from being a full halt.

## Suggested test design

- A worker unit test: feed an invalid batch (undecodable transaction, or wrong base fee) through the gossip-prefetch path (`process_gossip` handling `WorkerGossip::Batch`), then call `synchronize` for a header referencing that digest with `is_certified = false`, and assert the batch is validated and rejected rather than silently accepted.  Today the second step returns `Ok` without validating.
- An integration test that has a Byzantine committee member gossip and serve an invalid batch, then propose a header referencing it, and assert honest nodes refuse to vote for / certify the header (or, if certified by design in some variant, that execution does not fatally halt the engine).

## Acceptance criteria

- Every batch a node votes to certify has passed `validate_batch`, regardless of whether it entered the store via gossip prefetch, report, or sync.  Presence in `NodeBatchesCache` is never treated as evidence of validity for the vote path.
- An invalid batch cannot be admitted into a certified header by any node.
- A certified batch that fails transaction recovery does not fatally terminate the engine loop on every node (defense-in-depth criterion).

MavenRain:
Fixes #933.

## Summary

Closes the batch-validation bypass described in #933 and adds a defense-in-depth guard so a certified-but-invalid batch can never fatally halt execution.

**Root cause.**  `NodeBatchesCache` is used by the vote path as a proxy for "validated and available," but the gossip-prefetch path wrote *unvalidated* bodies into it.  `synchronize` then took the "already in store" short-circuit and voted without ever calling `validate_batch`.

## Changes

1. **Primary fix: validate before caching on the prefetch path** (`crates/consensus/worker/src/network/handler.rs`).  `process_gossip` now routes the fetched `(digest, batch)` through a new `validate_and_cache_prefetched_batch`, which seals the body with the fetched digest and calls `self.validator.validate_batch` (the same gate as `process_report_batch`) before inserting into `NodeBatchesCache`.  An invalid body is dropped and `record_batch_validation_failure` is recorded;  it is a content fault of the served body (not of the gossip relayer), so it is not propagated as an error.  Only a genuine storage-write error still propagates.

2. **Defense in depth: non-fatal transaction recovery on the execute path** (`crates/tn-reth/src/lib.rs`, `build_block_from_batch_payload`).  The signer-recovery phase changed from a fallible `map(...).collect::<Result>()?` (fatal on the first unrecoverable transaction) to `filter_map(...).ok()`, which now drops unrecoverable transactions and continues, exactly as the adjacent execute phase already tolerates `InvalidTx`.  The block already may contain a subset of a batch's transactions, so this preserves determinism (every honest node drops the same transactions from the same certified bytes).

## Why the primary fix is complete

Every production writer into `NodeBatchesCache` was audited:

| Writer | Validated? | Reachable by vote path pre-certification? |
|---|---|---|
| gossip prefetch (`handler.rs`) | **now yes** (this PR) | yes (the gap) |
| report batch (`handler.rs`) | yes (`validate_batch`) | n/a |
| sync during vote (`primary.rs` `store_synced_batches`) | yes when `!is_certified` | yes, already gated |
| own batch production (`worker.rs`) | valid by construction (own pool) | n/a |
| certificate-payload fetch (`batch_fetcher.rs` `fetch_for_primary`) | trusted-because-certified | no (post-certification) |

The gossip prefetch was the only unvalidated writer that could seed the cache before the vote, so closing it restores the "cached implies validated" invariant the vote path relies on.

## Tests

- `crates/consensus/worker/tests/it/network_tests.rs`
  - `test_gossip_prefetch_rejects_invalid_batch`: with a rejecting validator, a current-epoch prefetched body is dropped and never written to `NodeBatchesCache`.
  - `test_gossip_prefetch_caches_valid_batch`: the identical batch with an accepting validator *is* cached, isolating validation (not the epoch check) as the reason the first batch was dropped.
- `crates/tn-reth/tests/it/recover_drop_props.rs`
  - `undecodable_only_batch_builds_empty_block_instead_of_halting`: an undecodable-only batch builds an empty block instead of returning a fatal error.
  - `undecodable_tx_is_dropped_valid_tx_survives`: a batch mixing an undecodable transaction with a valid one drops only the undecodable one and still executes the valid transaction.

`cargo fmt`, `clippy` (tn-worker + tn-reth), and all four new tests pass.

## Acceptance criteria (from #933)

- [x] Every batch a node votes to certify has passed `validate_batch` regardless of how it entered the store;  presence in `NodeBatchesCache` is never treated as validity for the vote path.
- [x] An invalid batch cannot be admitted into a certified header via the gossip prefetch path.
- [x] A certified batch that fails transaction recovery no longer fatally terminates the engine loop (defense-in-depth).

### DIFF
--- a/crates/tn-reth/src/lib.rs
+++ b/crates/tn-reth/src/lib.rs
@@ -763,21 +763,31 @@ impl RethEnv {
         // Phase 1: Recover all transactions (ECDSA ecrecover) in parallel via rayon.
         // Always use par_iter — the slight overhead on small batches is negligible
         // compared to the savings on large ones, and avoids an extra code path.
+        //
+        // A transaction whose signer cannot be recovered cannot be executed, but a
+        // certified sub-DAG is fixed and identical on every honest node, so returning
+        // an error here would deterministically halt (and, on restart-replay,
+        // crash-loop) the whole network on a single undecodable transaction. Instead
+        // drop the unrecoverable transactions and continue, mirroring the `InvalidTx`
+        // tolerance of the execute phase below: every node drops the same
+        // transactions from the same certified bytes, so the resulting block stays
+        // deterministic (issue #933). The primary defense is validating batches before
+        // they can be certified; this only bounds the blast radius if one ever is.
         let recovered_txs = transactions
             .par_iter()
-            .map(|tx_bytes| {
+            .filter_map(|tx_bytes| {
                 reth_recover_raw_transaction::<TransactionSigned>(tx_bytes)
                     .inspect_err(|e| {
                         error!(
                             target: "engine",
                             batch=?batch_digest,
                             ?tx_bytes,
-                            "failed to recover signer: {e}"
+                            "failed to recover signer, dropping transaction: {e}"
                         )
                     })
-                    .map_err(TnRethError::from)
+                    .ok()
             })
-            .collect::<TnRethResult<Vec<_>>>()?;
+            .collect::<Vec<_>>();
 
         // Phase 2: Execute recovered transactions sequentially.
         for recovered in recovered_txs {
--- a/crates/consensus/worker/src/network/handler.rs
+++ b/crates/consensus/worker/src/network/handler.rs
@@ -23,8 +23,8 @@ use tn_network_libp2p::{
 use tn_network_types::{WorkerOthersBatchMessage, WorkerToPrimaryClient};
 use tn_storage::{consensus::ConsensusChain, tables::NodeBatchesCache};
 use tn_types::{
-    ensure, now, try_decode, BatchValidation, BlsPublicKey, Database, Epoch, SealedBatch, WorkerId,
-    B256,
+    ensure, now, try_decode, Batch, BatchValidation, BlsPublicKey, Database, Epoch, SealedBatch,
+    WorkerId, B256,
 };
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 use tracing::{debug, warn};
@@ -159,8 +159,6 @@ where
                 // most likely be removed before we can use it and will be fetched
                 // later when needed.
                 ensure!(my_epoch == epoch, WorkerNetworkError::BatchEpochMismatch(epoch, my_epoch));
-                // Retrieve the batch...
-                let store = self.consensus_config.node_storage();
                 // Since we are precaching Batches for the current epoch we only need to check if it
                 // is in the local cache. There should not have been an opertunity
                 // for it to be in the consensus chain yet.
@@ -191,29 +189,10 @@ where
                     let mut missing = BTreeSet::from([batch_hash]);
                     match self.network_handle.request_batches(&mut missing).await {
                         Ok(batches) => {
+                            // Note: retrieving this batch for no reason is wasteful, it should
+                            // only effect nodes catching up old epochs though...
                             if let Some((digest, batch)) = batches.first() {
-                                // Storing batches for future epochs can cause problems.  This might
-                                // open an attack for rogue
-                                // validator to fill disk space, the cache is cleared on
-                                // epoch boundaries anyway, etc.
-                                // Note: retrieving this batch for no reason is wasteful, it should
-                                // only effect nodes catching up old epochs though...
-                                if batch.epoch == self.consensus_config.epoch() {
-                                    store.insert::<NodeBatchesCache>(digest, batch).map_err(
-                                        |e| {
-                                            WorkerNetworkError::Internal(format!(
-                                                "failed to write to batch store: {e}"
-                                            ))
-                                        },
-                                    )?;
-                                } else {
-                                    debug!(
-                                        target: "worker:network",
-                                        batch_epoch = batch.epoch,
-                                        current_epoch = self.consensus_config.epoch(),
-                                        "gossipped batch epoch mismatch - discarding"
-                                    );
-                                }
+                                self.validate_and_cache_prefetched_batch(digest, batch)?;
                             }
                         }
                         Err(e) => {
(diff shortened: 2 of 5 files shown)


crates/tn-reth/src/lib.rs
```diff
@@ -763,21 +763,31 @@ impl RethEnv {
...
+        // crash-loop) the whole network on a single undecodable transaction. Instead
+        // drop the unrecoverable transactions and continue, mirroring the `InvalidTx`
+        // tolerance of the execute phase below: every node drops the same
+        // transactions from the same certified bytes, so the resulting block stays
+        // deterministic (issue #933). The primary defense is validating batches before
+        // they can be certified; this only bounds the blast radius if one ever is.
         let recovered_txs = transactions
             .par_iter()
-            .map(|tx_bytes| {
+            .filter_map(|tx_bytes| {
```
grantkee, line 778: Good catch. Should we also add a metric for this? There should never be unrecoverable transactions, but it's hard to confirm without reviewing logs for the error

grantkee (approved):
LGTM - one suggestion to add a metric

sstanfield (approved):
Looks good.  Also this prefetch code only makes sense for CVVs now.  Any nodes following consensus will download packs or full consensus output and prefetching will only waste bandwidth (harmless otherwise).  I don't think I made sure they no longer subscribe to this gossip.  Not for this PR, just something we should double check (and some of the comments are not entirely accurate with the new sync).

crates/tn-reth/src/lib.rs
MavenRain: I will create a follow-up issue to add a metric for this.

MavenRain:
> Looks good. Also this prefetch code only makes sense for CVVs now. Any nodes following consensus will download packs or full consensus output and prefetching will only waste bandwidth (harmless otherwise). I don't think I made sure they no longer subscribe to this gossip. Not for this PR, just something we should double check (and some of the comments are not entirely accurate with the new sync).

I will make a follow-up issue to address subscription cleanup here.

PR #939, opened 2026-07-16
feat(network): freeze and delete the legacy /tn-stream bulk path (739, step 9c)

MavenRain:
## Summary

With the worker batch, full and partial epoch pack, `MissingCertificates`, and consensus-output paths all migrated onto the typed sync protocol (items 4-9b), the legacy raw `/tn-stream/0.0.1` behaviour and its digest-correlation machinery are no longer used by any live path. This PR deletes them.

After this change every bulk transfer rides exactly one mechanism: a per-role sync protocol (`/tn-primary-sync-{chain}`, `/tn-worker-{id}-sync-{chain}`) whose stream carries the request in its first `SyncFrame`. The deprecated request-response variants that used to drive the ack-then-open-stream dance are retained on the wire (their BCS discriminants are positional and cannot move before a coordinated version bump) but are now answered with a typed error rather than served.

This is the "freeze" half of item 9; a later, telemetry-and-epoch-gated `/0.0.2` bump removes the now-dead request/response variants entirely.

## What is deleted

**`crates/network-libp2p` (the stream subsystem):**
- The `/tn-stream-{chain}/0.0.1` protocol builder (`types::stream_protocol` legacy form) and the dialer-preference `(bulk, sync)` protocol pair; `stream_protocol(network_type, chain_id)` now returns the single per-role sync protocol.
- The `StreamKind` enum (was `{Legacy, Sync}`) and every `kind` parameter it threaded through: `TNStreamProtocol` now advertises one protocol, `open_stream`/`OpenStream`/`InboundStream`/`OutboundOpen`/`PendingOpen` no longer carry a kind, and `upgrade_inbound` returns the raw stream (all inbound streams are sync). The negotiation-failure penalty exemption is preserved (an open that fails negotiation is still honest version/role skew).

**`crates/consensus/primary`:**
- `StreamRequestKind`, `stream_request_digest`, `PendingStreamRequest`, the `(peer, digest)` pending map and its 30s prune task, `accept_stream_request`, `send_stream_ack`, the legacy inbound reader `process_inbound_stream`, and the handler's `process_request_epoch_stream` + `send_epoch_over_stream`.
- `request_epoch_pack_inner` collapses to a single sync call (full and partial epoch packs are sync-only; a fetch hard-fails and the caller retries when no connected peer serves sync, rather than syncing over the legacy path).
- `PrimaryRequest::StreamEpoch` / `StreamEpochPartial` are now answered with `PrimaryResponse::Error` (mirroring how `StreamConsensusOutput` and `MissingCertificates` already behave).

**`crates/consensus/worker`:**
- `PendingBatchStream`, `LegacyStreamGuard`, `try_admit_legacy`, `begin_serving_legacy`, `peer_in_flight`, the pending map and its prune task, the legacy inbound reader `process_inbound_stream`, the legacy responder `process_request_batches_stream` + `send_batches_over_stream`, the legacy requester `request_batches_from_peer_legacy` + `read_and_validate_batches_with_timeout`, the chunk codec (`read_batch`, `read_chunk_count`), and `generate_batch_request_id`.
- The batch fetch is sync-only: `request_batches_from_peer` opens the sync protocol and, on a peer that does not answer, caches it unsyncable this epoch and lets the multi-peer driver try the next peer (no fallback). `WorkerRequest::RequestBatchesStream` is now answered with `WorkerResponse::Error`.

## What is intentionally kept (wire compatibility)

The following BCS variants are DEAD (never constructed) or DEPRECATED (answered with error) but are NOT removed here, because BCS encodes enum variants by position and removing them is a wire break that must be coordinated:
- `PrimaryRequest::{StreamEpoch, StreamConsensusOutput, StreamEpochPartial}`, `PrimaryResponse::StreamRequestAck`
- `WorkerRequest::RequestBatchesStream`, `WorkerResponse::RequestBatchesStream { ack }`

Their removal is the follow-up `/0.0.2` bump, gated on telemetry confirming no peers still open the legacy path and applied at an epoch boundary.

## Admission accounting

The per-peer and global concurrency caps used to be shared between the legacy and sync paths (a combined `peer_in_flight` count over pending + serving-legacy + sync). With the legacy half gone, `try_admit_sync` reduces to the sync per-peer counter alone (`try_admit_sync(&semaphore, &sync_peers, peer)`), and the global semaphore still bounds `MAX_CONCURRENT_*_STREAMS`. No cap is loosened.

## Behavior change to flag

Making the epoch-pack and worker-batch paths sync-only removes their legacy fallback: a fetch that previously could still succeed over `/tn-stream` against a pre-sync peer now fails that peer and moves on (retrying across peers / epochs). This is the intended endpoint of the migration and matches the step-6/step-7 precedent; it assumes the rollout has confirmed peers speak sync. A peer still on `/tn-stream` will get `UpgradeFailed` against a 9c node (penalty-exempt).

## Sync-requester robustness (transient-error classification)

Completing the sync-only cutover means there is no longer any legacy fallback to recover a fetch that a peer fails to answer, so the requester's classification of a failed exchange now has full weight. An adversarial review caught that both the worker batch requester and the primary epoch-pack requester classified *transient* post-negotiation failures — the Ack-read timeout and a mid-read `ConnectionReset`/`UnexpectedEof`/`BrokenPipe` — as `Unsupported`, which caches the peer unsyncable for the whole epoch. That was safe while a legacy fallback existed (the old code re-served the batch over `/tn-stream`), but under the sync-only path it can poison a momentarily-slow but genuinely sync-capable peer and, if the only holders of a needed digest are poisoned, wedge catch-up until the epoch rotates.

Fix: once negotiation has succeeded the peer is provably sync-capable, so every post-negotiation failure (Ack timeout or any read error) now maps to `Failed` — try the next peer and keep the peer sync-capable so it is re-probed on the next fetch. Only an open-time `UpgradeFailed` (multistream-select rejected the protocol) still maps to `Unsupported`. This corrects a real regression this PR would otherwise introduce for the worker batch path, and the same latent mis-classification on the primary full-epoch-pack path (present since the step-6 cutover, now system-critical with no fallback anywhere). The partial-epoch-pack and consensus-output paths were already guarded (their `Unsupported` never caches `false`).

## Testing

- Unit tests updated: legacy admission/pending/codec/serve tests removed; sync admission and sync-codec tests retained. `StreamKind`/protocol-ordering assertions updated to the single-protocol shape.
- Gates (OOM-safe `-j2`): fmt, clippy `--all-features --all-targets` (network-libp2p + worker) / `--lib` (primary), workspace lib+unit tests, and `cargo doc -D warnings` for the touched crates. Consumer crates compile with zero edits outside the three touched crates.


sstanfield (dismissed):
Looks good, love the high red count.

sstanfield (approved):

PR #940, opened 2026-07-16
fix(consensus): make anti-equivocation records durable before externalizing

### LINKED ISSUE #934: LastProposed (and Votes) are written to the async in-memory epoch DB and are not durable before the header is broadcast / the vote is returned, so a crash-restart makes an honest node equivocate
## Summary

The proposer persists its last-proposed header and then immediately broadcasts that header.  The persistence goes to the epoch database, which is a `LayeredDatabase` opened in full-memory mode: its `insert` writes an in-memory map and enqueues the durable disk write to a background thread, returning before the write is durable.  There is no durability barrier between "record the header we are about to propose" and "broadcast it."

If the node crashes in that window (after the header is broadcast but before the background thread flushes the `LastProposed` record), then on restart the anti-equivocation guard reads a missing record and builds a new, different header for the same round.  The pre-crash header is already on the network.  Two distinct, validly-signed headers from the same honest author at the same `(epoch, round)` now exist: proposer equivocation, produced by an ordinary crash with no Byzantine actor.  Equivocation by an authority is exactly the accountability fault that DAG-BFT relies on honest nodes never committing.

The same root cause has a second instance on the voter side: the vote handler returns its vote to the requesting peer before the `Votes` record is durable, and the restart recast-guard reads the same lost record.

## The defect

**Locations:** the write-then-broadcast ordering at `crates/consensus/primary/src/proposer.rs:262-274`;  the restart guard and the new-header branch at `proposer.rs:562-604`; the store write at `crates/storage/src/stores/proposer_store.rs:20-22`;  the table's epoch hint at `crates/storage/src/lib.rs:89`;  the full-memory epoch DB at `crates/storage/src/composite_db.rs:24-37`; the non-durable insert at `crates/storage/src/layered_db.rs:284-289`;  the voter-side instance at `crates/consensus/primary/src/network/handler.rs:439-451` and `:805-817`.

**The proposer persists, then broadcasts** (`proposer.rs:262-274`, `store_and_send_header`):

```rust
async fn store_and_send_header(
    header: &Header,
    proposer_store: DB,
    consensus_bus: &ConsensusBus,
) -> ProposerResult<()> {
    // Store the last header.
    proposer_store
        .write_last_proposed(header)
        .map_err(|e| ProposerError::StoreError(e.to_string()))?;

    // Send the new header to the `Certifier` that will broadcast and certify it.
    consensus_bus.headers().send(header.clone()).await.map_err(|e| Box::new(e).into())
}
```

**`write_last_proposed` is a plain insert into `LastProposed`** (`proposer_store.rs:20-22`):

```rust
fn write_last_proposed(&self, header: &Header) -> StoreResult<()> {
    let result = self.insert::<LastProposed>(&LAST_PROPOSAL_KEY, header);
```

**`LastProposed` is an epoch-hinted table** (`crates/storage/src/lib.rs:89`):

```rust
LastProposed;crate::LAST_PROPOSED_CF;TableHint::Epoch;<ProposerKey, Header>,  // Cleared every epoch
```

**The epoch DB is opened in full-memory mode** (`composite_db.rs:24-37`):

```rust
pub fn open(epoch_db: DB, kad_db: DB, cache_db: DB) -> Self {
    let epoch_db = LayeredDatabase::open(epoch_db, true);
    ...
}

fn get_db(&self, hint: TableHint) -> &LayeredDatabase<DB> {
    match hint {
        TableHint::Epoch => &self.inner.epoch_db,
        ...
    }
}
```

**A full-memory `insert` returns before the write is durable** (`layered_db.rs:284-289`):

```rust
fn insert<T: Table>(&self, key: &T::Key, value: &T::Value) -> eyre::Result<()> {
    self.mem_db.insert::<T>(key, value)?;
    let ins = Box::new(KeyValueInsert::<T> { key: key.clone(), value: value.clone() });
    self.tx.send(DBMessage::Insert(ins)).map_err(|_| eyre::eyre!("DB thread gone, FATAL!"))?;
    Ok(())
}
```

The in-memory map is updated and a `DBMessage::Insert` is enqueued to the background writer, then `insert` returns.  Nothing waits for the background thread to persist.  On restart the in-memory map is repopulated from disk only, so a record enqueued but not yet flushed at crash time is gone.

**On restart, a missing record leads to a new header for the same round** (`proposer.rs:562-604`):

```rust
// check if proposer store's last header is from this round
let last_proposed = self
    .proposer_store
    .get_last_proposed()
    .map_err(|e| ProposerError::StoreError(e.to_string()))?;
let possible_header_to_repropose =
    last_proposed.filter(|h| h.round() == current_round && h.epoch() == current_epoch);
...
match possible_header_to_repropose {
    // resend header
    Some(header) => { ... Proposer::repropose_header(header, ...) ... }
    // create new header
    None => {
        // collect values from &mut self for this header
        let num_of_digests = self.digests.len().min(self.max_header_num_of_batches);
        let digests: VecDeque<_> = self.digests.drain(..num_of_digests).collect();
        let parents = std::mem::take(&mut self.last_parents);
        ...
        // spawn tokio task to create, store, and send new header to certifier
        ...
    }
}
```

The `Some` arm re-proposes the identical header (safe).  The `None` arm, taken when the record for `current_round` is absent, drains fresh digests and parents and builds and signs a new header for the same round.  If the pre-crash header was already broadcast, that is a second distinct signed header at the same round.

**Voter-side instance (same root cause).** The vote handler returns its vote before the write is durable (`handler.rs:805-817`):

```rust
// this node hasn't voted yet
let vote = Vote::new(
    &header,
    self.consensus_config.authority_id().ok_or(HeaderError::NotCommitteeMember)?,
    self.consensus_config.key_config(),
);
...
// Update the vote digest store with the vote we just sent.
self.consensus_config.node_storage().write_vote(&vote)?;

Ok(PrimaryResponse::Vote(vote))
```

`write_vote` inserts into `Votes`, which is also `TableHint::Epoch`, so it is the same non-durable path.  The restart recast-guard depends on that record (`handler.rs:439-451`):

```rust
if let Ok(Some(vote_info)) = self.consensus_config.node_storage().read_vote_info(&auth_id) {
    // If we have already cast a vote for this header then just recast it quickly.
    if vote_info.vote_digest == header.digest().into() {
        ...
        return Ok(PrimaryResponse::Vote(vote));
    }
}
```

If the `Votes` record was lost in the crash window, the guard does not fire, and the voter can cast a second, different vote after restart.

## Impact

- **Proposer equivocation from a routine crash (no attacker).** An honest proposer that crashes in the write window emits two distinct signed headers at one `(epoch, round)`.  This is a signed accountability fault.  Depending on timing it can split honest votes for that round (no certificate forms, a liveness stall for that author's round) and, more seriously, undermines the anti-equivocation invariant the rest of consensus assumes holds for honest authorities.
- **Voter double-vote from a routine crash.** An honest voter that crashes in the write window can contribute two distinct votes at one round.  On its own this needs a proposer to have equivocated for the two votes to target different headers, so its severity is lower than the proposer instance, but it is the same defect and compounds it.

The window is short (the background writer flushes quickly), so the per-crash probability is low, but crashes are routine (OOM, restarts, deploys, power loss), and a single equivocation is a safety-relevant event, not a recoverable local error.

## Root cause

Consensus-critical anti-equivocation records (`LastProposed`, `Votes`) live in the full-memory epoch `LayeredDatabase`, whose `insert` is asynchronous with no durability barrier, and the proposer and vote handler externalize an irreversible action (broadcast the header, return the vote) before that record is durable.  The durability of the record and the externalization of the action are not ordered.

## Suggested fix

- Order durability before externalization: persist `LastProposed` durably before the header is handed to the Certifier for broadcast, and persist `Votes` durably before the vote is returned to the peer.  This likely means a synchronous or awaited durable write (a `persist` / sync barrier) on these specific writes, or a write path that confirms the background flush before proceeding.
- Or move `LastProposed` and `Votes` off the full-memory async epoch DB onto a synchronously-durable store, given they are anti-equivocation state rather than throughput-sensitive bulk data.

Either way, the invariant to restore is: an honest node never externalizes a header or vote whose anti-equivocation record is not yet durable.

## Suggested test design

- A `crates/storage` test that inserts a `LastProposed` record via the full-memory epoch `LayeredDatabase`, drops the handle without draining the background writer to simulate a crash, reopens from disk, and asserts (documents) that the record is absent, demonstrating the non-durability window that the proposer relies on.
- A proposer test: propose header H at round R, simulate a crash before the `LastProposed` flush, restart, drive `propose` for round R, and assert the proposer re-proposes H rather than building a new header (which it does today via the `None` branch).
- Symmetric vote-handler test for the recast guard across a simulated crash before the `Votes` flush.

## Acceptance criteria

- After a crash at any point after a header is broadcast, an honest node on restart never proposes a different header for that same `(epoch, round)`;  it either re-proposes the identical header or does not propose that round.
- After a crash at any point after a vote is returned, an honest node on restart never returns a different vote for the same round.
- `LastProposed` and `Votes` writes are durable before the corresponding header broadcast / vote return, verified by test.

MavenRain:
Closes #934 

## Problem

The proposer persists its `LastProposed` header and then immediately broadcasts it
(`store_and_send_header`, `crates/consensus/primary/src/proposer.rs`).  The vote handler writes its
`Votes` record and then returns the vote to the requesting peer (`vote_inner`,
`crates/consensus/primary/src/network/handler.rs`).

Both records are `TableHint::Epoch` tables, so they route to the epoch DB: a full-memory
`LayeredDatabase` whose `insert` updates an in-memory map and enqueues the durable disk write to a
background thread, returning **before** the write is durable.  There is no barrier between recording
what we are about to externalize and externalizing it.

If the node crashes in that window, on restart the in-memory map is repopulated from disk only, so a
record enqueued but not yet flushed at crash time is gone.  The anti-equivocation guard then reads
nothing and builds a *different* header (or signs a *different* vote) for the same `(epoch, round)`,
while the pre-crash header/vote is already on the network.  Two distinct, validly-signed messages
from the same honest author at the same round now exist: authority equivocation, produced by an
ordinary crash with no Byzantine actor.  Equivocation by an authority is exactly the accountability
fault DAG-BFT relies on honest nodes never committing.

## Why a plain persist barrier is not enough

The obvious fix, awaiting the existing `persist` barrier, does not fully close the window.  The
certificate store opens an epoch-DB write txn on essentially every round (`CertificateStore::write` /
`write_all`, all cert tables are `TableHint::Epoch`).  In the background writer (`db_run`), a bare
`insert` dequeued while a txn is open is **absorbed** into that open physical txn and only becomes
durable when it commits.  The current `CaughtUp` barrier (behind both `persist` and `sync_persist`)
acks unconditionally, so it can return while the absorbed record is still uncommitted.

Making `CaughtUp` wait for the open txn is not an option: callers that hold an open `write_txn`
across `sync_persist` would deadlock, and two existing tests
(`test_layereddb_overlapped_txn_drop_keeps_count_balanced`,
`test_layereddb_cloned_txn_commit_sends_exactly_one_end`) do exactly that on purpose.

## Fix

Add a dedicated durable barrier and leave `persist` / `sync_persist` byte-identical:

- **`tn-types`** — new provided `Database` method `persist_durable::<T>()`, a no-op by default (as
  `persist` is).  Synchronously-durable backends (raw MDBX / redb, `MemDatabase`) need no barrier;
  their inserts are already durable on return.
- **`tn-storage`** — new `DBMessage::DurableBarrier`.  The background thread acks it immediately
  when no write txn is open (every prior insert was written directly and durably);  otherwise it
  defers the ack and drains it once the physical txn that absorbed the write commits (the `CommitTxn`
  / `EndTxn` arms, guarded by `txn.is_none()`).  `txn` transitions to `None` only inside `end_txn`,
  which is reached only from those two arms, so no commit path can miss the drain.
  `LayeredDatabase` and `CompositeDatabase` implement it;  the composite routes by table hint, so an
  epoch-DB barrier only waits on epoch-DB txns.
- **`tn-primary`** — the proposer awaits `persist_durable::<LastProposed>()` after the store write
  and before broadcast;  the vote handler awaits `persist_durable::<Votes>()` after `write_vote` and
  before returning the vote.  The recast/resend paths only ever resend an already-stored record, so
  the single fresh-write site on each path is the only place a barrier is needed.

The MDBX backend commits durably (default `SyncMode::Durable`, fsync), so once the barrier resolves
the record is on disk.

## Restored invariant

An honest node never externalizes a header or vote whose anti-equivocation record is not yet durable.
After a crash at any point after the header is broadcast / the vote is returned, on restart the node
either re-proposes the identical header (re-signs the identical vote) or does not act for that round;
it never produces a different one.

## Changes

- [x] `persist_durable::<T>()` durable barrier on the `Database` trait (defaults to `persist`)
- [x] `DBMessage::DurableBarrier` with defer-until-commit drain in `db_run`; `LayeredDatabase` +
      `CompositeDatabase` impls (existing `persist` / `sync_persist` semantics unchanged)
- [x] Proposer awaits the barrier before broadcasting the header
- [x] Vote handler awaits the barrier before returning the vote
- [x] Storage tests: plain `persist` acks before an absorbed write is durable (documents the window);
      `persist_durable` stays pending until the absorbing txn commits and the record is then on disk
      (the fix);  with no open txn it resolves immediately against a durably-written record

## Testing

- `tn-storage` lib tests: 129 passed (3 new, plus the existing overlapped/cloned-txn tests that
  confirm the shared `CaughtUp` path is untouched)
- `tn-primary` proposer (3) and vote (30) tests pass
- `cargo fmt` + `clippy` clean on `tn-types`, `tn-storage`, `tn-primary`

### DIFF
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -18,7 +18,7 @@ use tn_config::ConsensusConfig;
 use tn_network_libp2p::{
     write_frame, GossipMessage, PrimarySyncRequest, Stream, SyncFrame, SyncFrameError,
 };
-use tn_storage::{consensus::ConsensusChain, CertificateStore, VoteDigestStore};
+use tn_storage::{consensus::ConsensusChain, tables::Votes, CertificateStore, VoteDigestStore};
 use tn_types::{
     ensure,
     error::{CertificateError, HeaderError, HeaderResult},
@@ -868,6 +868,14 @@ where
 
         // Update the vote digest store with the vote we just sent.
         self.consensus_config.node_storage().write_vote(&vote)?;
+
+        // Wait for the `Votes` record to be durable before returning the vote to the peer. The
+        // epoch DB persists asynchronously, so `write_vote` returns before the record hits disk;
+        // returning the vote first would let a crash in that window lose the record. On restart the
+        // recast guard would then read nothing and could sign a *different* vote for the same
+        // author/round while the pre-crash vote is already held by the peer: equivocation from an
+        // ordinary crash. See issue #934.
+        self.consensus_config.node_storage().persist_durable::<Votes>().await;
 
         Ok(PrimaryResponse::Vote(vote))
     }
--- a/crates/consensus/primary/src/proposer.rs
+++ b/crates/consensus/primary/src/proposer.rs
@@ -25,7 +25,7 @@ use std::{
     collections::{BTreeMap, VecDeque},
 };
 use tn_config::ConsensusConfig;
-use tn_storage::ProposerStore;
+use tn_storage::{tables::LastProposed, ProposerStore};
 use tn_types::{
     now, AuthorityIdentifier, BlockHash, Certificate, Committee, Database, Epoch, Hash as _,
     Header, Noticer, Round, TaskManager, TaskSpawner, TnReceiver, TnSender, WorkerId,
@@ -268,6 +268,14 @@ impl<DB: Database> Proposer<DB> {
         proposer_store
             .write_last_proposed(header)
             .map_err(|e| ProposerError::StoreError(e.to_string()))?;
+
+        // Wait for the `LastProposed` record to be durable before broadcasting. The epoch DB
+        // persists asynchronously, so `write_last_proposed` returns before the record hits disk;
+        // broadcasting first would let a crash in that window lose the record. On restart the
+        // anti-equivocation guard would then read nothing and build a *different* header for this
+        // same round while the pre-crash header is already on the network: equivocation from an
+        // ordinary crash. See issue #934.
+        proposer_store.persist_durable::<LastProposed>().await;
 
         // Send the new header to the `Certifier` that will broadcast and certify it.
         consensus_bus.headers().send(header.clone()).await.map_err(|e| Box::new(e).into())
--- a/crates/storage/Cargo.toml
+++ b/crates/storage/Cargo.toml
@@ -41,6 +41,8 @@ tn-reth = { workspace = true }
 tn-test-utils = { workspace = true }
 rand = { workspace = true }
 roaring = { workspace = true }
+# Async barrier tests for `persist_durable` need a runtime + timers.
+tokio = { workspace = true, features = [ "macros", "rt", "time" ] }
 
 [features]
 redb = []
--- a/crates/storage/src/composite_db.rs
+++ b/crates/storage/src/composite_db.rs
@@ -130,6 +130,10 @@ impl<DB: Database> Database for CompositeDatabase<DB> {
         self.get_db(T::HINT).persist::<T>()
     }
 
+    fn persist_durable<T: Table>(&self) -> impl Future<Output = ()> + Send {
+        self.get_db(T::HINT).persist_durable::<T>()
+    }
+
     fn sync_persist(&self) {
         self.inner.epoch_db.sync_persist();
         self.inner.kad_db.sync_persist();
--- a/crates/storage/src/layered_db.rs
+++ b/crates/storage/src/layered_db.rs
@@ -179,6 +179,11 @@ fn db_run<DB: Database>(db: DB, mem_db: Option<MemDatabase>, rx: Receiver<DBMess
     let mut txn = None;
     let mut last_compact = Instant::now();
     let mut committed_inserts: Vec<Box<dyn InsertTrait<DB>>> = Vec::with_capacity(1000);
+    // Durability-barrier acks deferred because a write txn was open when they arrived. A bare
+    // `Insert` enqueued while a txn is open rides that txn (see the `Insert` arm below), so its
+    // durability is only settled once the txn commits; these are drained the moment `txn` returns
+    // to `None`.
+    let mut pending_durable: Vec<oneshot::Sender<()>> = Vec::new();
     if let Err(e) = db.compact() {
         tracing::error!(target: "layered_db_runner", "DB ERROR compacting DB on startup (background): {e}");
     }
@@ -198,6 +203,11 @@ fn db_run<DB: Database>(db: DB, mem_db: Option<MemDatabase>, rx: Receiver<DBMess
             }
             DBMessage::CommitTxn => {
                 end_txn(&mut txn, &mut committed_inserts, mem_db.as_ref());
+                if txn.is_none() {
+                    pending_durable.drain(..).for_each(|tx| {
+                        let _ = tx.send(());
+                    });
+                }
             }
             DBMessage::EndTxn => {
                 tracing::warn!(
@@ -205,6 +215,11 @@ fn db_run<DB: Database>(db: DB, mem_db: Option<MemDatabase>, rx: Receiver<DBMess
                     "write txn dropped without commit; committing it to keep persistence alive"
                 );
                 end_txn(&mut txn, &mut committed_inserts, mem_db.as_ref());
+                if txn.is_none() {
+                    pending_durable.drain(..).for_each(|tx| {
+                        let _ = tx.send(());
+                    });
+                }
             }
             DBMessage::Insert(ins) => {
                 if let Some((txn, _)) = &mut txn {
@@ -246,6 +261,17 @@ fn db_run<DB: Database>(db: DB, mem_db: Option<MemDatabase>, rx: Receiver<DBMess
             }
             DBMessage::CaughtUp(tx) => {
                 let _ = tx.send(());
+            }
+            DBMessage::DurableBarrier(tx) => {
+                // If a write txn is open, any bare insert this barrier is ordered after has been
+                // absorbed into it and is not yet durable, so hold the ack until that txn commits
+                // (drained in the `CommitTxn`/`EndTxn` arms). With no txn open, every prior insert
+                // was written directly and durably, so ack now.
+                if txn.is_some() {
+                    pending_durable.push(tx);
+                } else {
+                    let _ = tx.send(());
+                }
             }
             DBMessage::Stats(tx) => {
                 let _ = tx.send(LayeredDbStats {
@@ -473,6 +499,19 @@ impl<DB: Database> Database for LayeredDatabase<DB> {
         }
     }
 
+    fn persist_durable<T: Table>(&self) -> impl Future<Output = ()> + Send {
+        let (tx, rx) = oneshot::channel();
+        let r = self
+            .tx
+            .send(DBMessage::DurableBarrier(tx))
+            .map_err(|_| eyre::eyre!("DB thread gone, FATAL!"));
+        async move {
+            if r.is_ok() {
+                let _ = rx.await;
+            }
+        }
+    }
+
     fn sync_persist(&self) {
         let (tx, mut rx) = oneshot::channel();
         let r = self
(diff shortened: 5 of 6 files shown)


sstanfield (approved):
Looks good but have a couple of notes.  One, not sure why we need to add persist_durable vs fix persist().  This is fixing a bug in persist() to my mind (and I originally wrote persist()) so I think it makes sense.  It was NOT intended to be sidestepped by open transactions, that is a bug.  Second, we need to review these tables and see if they should be raw MDBX vs layered with memory.  These persists calls may defeat the entire purpose of the memory layer depending on the overall usage patterns.

MavenRain:
> Looks good but have a couple of notes. One, not sure why we need to add persist_durable vs fix persist(). This is fixing a bug in persist() to my mind (and I originally wrote persist()) so I think it makes sense. It was NOT intended to be sidestepped by open transactions, that is a bug. Second, we need to review these tables and see if they should be raw MDBX vs layered with memory. These persists calls may defeat the entire purpose of the memory layer depending on the overall usage patterns.

Thanks, both notes are good.  Speaking on each:

**On `persist_durable` vs fixing `persist()` (note 1).** You're right, and this belongs in `persist()`.  Re-auditing to back that: the deadlock concern I wrote into the PR description is narrower than I made it sound.  There are no production callers of the `Database` trait methods `persist` / `sync_persist` at all;  every call site is under `#[cfg(test)]` or in `tests/it/` (the `crates/storage/src/lib.rs` sites sit inside the `#[cfg(test)]` module at `:216`, the `kad.rs` sites below `:570`, `composite_db.rs:472` below `:268`).  So the only things that would hang under wait-for-txn-close semantics are a few test sites, each a small rewrite, and the two overlapped-txn tests can substitute the `db.stats()` flush already on the next line.  `persist_durable` was the conservative additive choice under an existing approval rather than a claim that `persist()` can't be fixed.  I've drafted a follow-up (https://github.com/Telcoin-Association/telcoin-network/issues/962) to fold the durable semantics back into `persist()` as you describe, and to decide `sync_persist`'s fate at the same time.

**On raw MDBX vs layered (note 2).** Agreed these tables deserve that review.  One wrinkle worth flagging up front: the composite DB wraps every sub-DB in `LayeredDatabase` unconditionally (`composite_db.rs:33-45`, `get_db` returns `&LayeredDatabase`), so `TableHint` selects *which* layered DB, never layered-vs-raw.  Moving `LastProposed` / `Votes` to synchronously-durable MDBX is a composite-router change (field types + `get_db`'s return type, mirrored for redb), not a hint flip.  The tradeoff looks real in the direction you're pointing: `LastProposed` (one row) and `Votes` (one per authority) get little from the memory layer while paying the barrier's queue-coupled wait, whereas the certificate tables are the layer's actual beneficiaries and I'd keep those layered.  Drafted as https://github.com/Telcoin-Association/telcoin-network/issues/961 with a measurement step so the decision is made on numbers rather than by feel.

**One thing this PR does *not* cover, that the same review uncovered** Auditing the full write-then-externalize class, there are two more sites of the exact #934 shape left non-durable:

1. The certifier writes its own certificate then externalizes it with no barrier between: `insert::<ProposedCertificates>` at `certifier.rs:443`, then `process_own_certificate` at `:448` (which already sends on the parents bus) and gossip at `:454`.  Guard at `:418-430`, whose comment at `:421-425` names the stakes.
2. The vote-recast guard's certificate-existence check at `handler.rs:824-829` reads `read_by_index` -> `CertificateDigestByOrigin` + `Certificates`, both `TableHint::Epoch` and both non-durable.  A lost record there flips `cert_exists` to false and the node signs a different vote for the same round.

Both are pre-existing on `main` (this PR's diff doesn't touch `certifier.rs`), but this PR makes them *more* reachable, not less: once `LastProposed` / `Votes` are durable, a restart deterministically re-proposes the identical header and re-enters both guards, so these two records go from occasional backstops to load-bearing on every restart.  I've filed https://github.com/Telcoin-Association/telcoin-network/issues/963 as a mainnet blocker with the full write-up.

**Plan.** I'll land those in a follow-up PR right after this merges.  Site 1 there is a one-line barrier in this PR's own pattern;  Site 2 is the real decision and your read would help: fail the recast guard *closed* (refuse to re-vote on an unknown/errored certificate read, costing a round of liveness in the legitimate "proposer killed before quorum, restarted" case) versus making certificate acceptance durable, which reintroduces a barrier on the certificate hot path.

PR #945, opened 2026-07-16
fix(node): keep the healthcheck accept loop alive across transient accept errors

### LINKED ISSUE #943: node: the healthcheck accept loop is a critical task that exits on the first `accept()` error, so a transient error (e.g. fd exhaustion) shuts the whole node down
## Summary

The healthcheck endpoint (`--healthcheck <PORT>` / `HEALTHCHECK_TCP_PORT`) runs its accept loop as a **critical** task, but the loop is written as `while let Ok(..) = listener.accept().await { .. }`.  A single `Err` from `accept()` ends the `while let`, the task future returns `Ok(())`, and because the task is critical the `TaskManager` treats that `Ok` return (while the node is not already shutting down) as `CriticalExitOk` and notifies a **whole-node shutdown**.

`accept()` returns `Err` on ordinary, transient conditions, most importantly `EMFILE`/`ENFILE` (per-process or system-wide file-descriptor exhaustion), and also `ECONNABORTED`, `ENOBUFS`/`ENOMEM`, and `EINTR`.  So the liveness probe an operator enables to watch node health becomes the thing that takes the node down, and it does so precisely when the host is under fd/resource pressure (exactly when you least want the node to exit).  Because fd pressure often persists across a restart, it can re-trigger and turn into a restart loop.

The repo already handles this exact case correctly in the metrics server, whose accept loop swallows the error and continues, with a comment that says transient accept errors must not kill the node.  This issue asks the healthcheck loop to do the same.

The endpoint is off by default and only bites operators who enable it, but enabling a liveness/readiness probe is the standard production and Kubernetes setup, so the blast radius is "any operator following the documented monitoring path."

## The defect

**Locations:** `crates/node/src/health.rs:160-193` (the accept loop, spawned as a critical task), `crates/types/src/task_manager.rs:174-178` (`spawn_critical_task` marks the task `critical = true`) and `crates/types/src/task_manager.rs:452-486` (a critical task's `Ok` exit while not shutting down triggers node shutdown).  The production call site is `crates/node/src/manager/node.rs:1006`, which spawns it on the node's main task manager.

**Accept loop spawned as a critical task** (`health.rs:160-193`):

```rust
task_spawner.spawn_critical_task("healthcheck", async move {
    // accept loop - no connection limits; bounded read timeout per conn
    while let Ok((mut socket, _)) = listener.accept().await {
        // ... read request with a bounded timeout, route liveness/readiness,
        //     write the response, close the socket ...
    }
    Ok(())
});
```

A `while let Ok(..)` binds only on `Ok`;  the first `Err` from `accept()` falls out of the loop, so the closure runs off the end and returns `Ok(())`.

**`spawn_critical_task` marks the task critical** (`task_manager.rs:174-178`):

```rust
pub fn spawn_critical_task<F, S: ToString>(&self, name: S, future: F)
where
{
    self.create_task(name, future, true); // critical = true
}
```

**A critical task's `Ok` exit (while not shutting down) shuts the node down** (`task_manager.rs:452-486`):

```rust
Some(res) = self.tasks.next() => {
    match res {
        Ok(info) => {
            if !info.critical {
                continue;
            }
            // Ok exit is fine if we are shutting down.
            if !rx_shutdown.noticed() {
                result = Err(TaskJoinError::CriticalExitOk(info.name));
            }
        }
        Err((info, join_err)) => { /* ... */ }
    }
    break;
}
// ...
// No matter how we exit notify shutdown ...
shutdown.notify();
```

So: healthcheck `accept()` returns `Err` -> `while let` ends -> task returns `Ok(())` -> critical + `!rx_shutdown.noticed()` -> `CriticalExitOk("healthcheck")` -> `break` -> `shutdown.notify()` -> the whole node stops.

## Trigger conditions

The endpoint must be enabled (`--healthcheck <PORT>`, or `HEALTHCHECK_TCP_PORT`; `crates/telcoin-network-cli/src/node.rs:84-85`, `global = true`).  Given that, any single `accept()` error fells the node.  Realistic causes, none of which should be fatal:

- **File-descriptor exhaustion** (`EMFILE`/`ENFILE`): the node holds many peer connections, RocksDB handles, and sockets;  under a low `ulimit -n`, an fd leak, or a connection flood the accept syscall returns `EMFILE`.  This is the important case: the outage fires exactly under fd pressure, and can recur on restart while the pressure persists.
- **`ECONNABORTED`**: the peer reset the connection between the SYN and `accept()`;  transient and expected on a public listener.
- **`ENOBUFS`/`ENOMEM`**: transient kernel buffer pressure.
- **`EINTR`**: interrupted syscall.

tokio's `TcpListener::accept` surfaces these as `io::Error` rather than retrying, so the loop sees the `Err` directly.

## Root cause

The loop conflates "this connection failed" with "stop serving," and the endpoint's critical-task classification then escalates "stop serving" into "stop the node."  A liveness endpoint should be the most robust thing in the process: a failed `accept()` should be logged and skipped, never terminate the loop, and certainly never terminate the node.

## Suggested fix

Mirror the metrics server, which already gets this right (`crates/tn-metrics/src/server.rs:124-168`):

```rust
loop {
    match listener.accept().await {
        Ok((socket, _)) => { /* handle one connection */ }
        // transient accept errors (e.g. fd exhaustion) must not kill the node
        Err(e) => debug!(target: "tn::metrics", ?e, "metrics endpoint accept error"),
    }
}
```

Apply the same shape to `HealthcheckServer::spawn`:

```rust
task_spawner.spawn_critical_task("healthcheck", async move {
    loop {
        match listener.accept().await {
            Ok((mut socket, _)) => {
                // ... existing per-connection handling ...
            }
            // transient accept errors (e.g. fd exhaustion) must not kill the node
            Err(e) => {
                tracing::debug!(target: "epoch-manager", ?e, "healthcheck accept error");
                continue;
            }
        }
    }
});
```

Notes:

- The task future never needs to return `Ok(())` on its own: `create_task` already wraps it in `tokio::select! { _ = rx_shutdown => Ok(()), res = future => res }` (`task_manager.rs:189-198`), so node shutdown cancels the accept loop cleanly.  An unconditional `loop { match .. }` is therefore correct.
- Use the existing `"epoch-manager"` tracing target already used by this module (`health.rs:158`) or a dedicated `"tn::health"` target, whichever the team prefers.
- Optional hardening (separate from the fix): treat only a small set of truly-fatal errors as loop-ending (there aren't any in practice for a bound listener), and consider a tiny yield/backoff on a sustained error storm so a permanent `EMFILE` does not spin the log.  The metrics server does not bother, so matching it (plain `continue`) is the minimal, consistent change.

A one-line consistency check worth making in the same PR: audit other `spawn_critical_task` accept/serve loops for the same `while let Ok(..)`-over-`accept()` shape.

## Suggested test design

Add a test in `crates/node` that drives the healthcheck accept loop against a failing accept and asserts the node is not shut down:

- Bind the healthcheck server, then force an `accept()` error path (for example by simulating/injecting an accept error, or by exercising the loop with a listener that yields an error), and assert that the task does not resolve `Ok` and the manager's shutdown is not notified.
- Alternatively, a focused unit test that the loop continues past an `Err` from `accept()` and serves a subsequent successful connection.
- Regression intent: a single `accept()` error must leave the endpoint serving and the node running.

## Acceptance criteria

- A transient `accept()` error (fd exhaustion, `ECONNABORTED`, `EINTR`, `ENOBUFS`) is logged and skipped;  the accept loop keeps serving.
- With the healthcheck enabled, a failed `accept()` does not trigger `CriticalExitOk` or a node shutdown.
- A regression test covers the failed-accept case and asserts the node stays up.

---

Note: this defect surfaced during an AI-assisted availability review of the node startup and task-supervision paths, with the cited locations verified against current `main` (`0c59c15b`).  It is one instance of a broader pattern in which recoverable conditions are wired as node-fatal through the critical-task -> `until_exit` -> `shutdown.notify()` sink;  the metrics server (`server.rs:124-168`) is the in-repo precedent for the correct handling.  Trim or keep this note per the team's disclosure policy.

MavenRain:
Closes #943 

## Problem

The healthcheck endpoint (`--healthcheck <PORT>` / `HEALTHCHECK_TCP_PORT`) runs its accept loop as a
**critical** task (`crates/node/src/health.rs`), but the loop was written as
`while let Ok((mut socket, _)) = listener.accept().await { .. }` with `Ok(())` below it.

A `while let Ok(..)` binds only on `Ok`, so the first `Err` from `accept()` falls out of the loop and
the closure returns `Ok(())`.  Because the task is critical, `TaskManager` treats that `Ok` return
(while the node is not already shutting down) as `CriticalExitOk` and notifies a **whole-node
shutdown** (`crates/types/src/task_manager.rs`).

`accept()` returns `Err` on ordinary, transient conditions, most importantly `EMFILE`/`ENFILE`
(per-process or system-wide file-descriptor exhaustion), and also `ECONNABORTED`, `ENOBUFS`/`ENOMEM`,
and `EINTR`.  So the liveness probe an operator enables to watch node health becomes the thing that
takes the node down, and it does so precisely under fd/resource pressure, exactly when the node should
stay up.  Because that pressure often persists across a restart, it can turn into a restart loop.

The endpoint is off by default and only bites operators who enable it, but enabling a
liveness/readiness probe is the standard production and Kubernetes setup, so the blast radius is any
operator following the documented monitoring path.

## Fix

The repo already handles this exact case correctly in the metrics server
(`crates/tn-metrics/src/server.rs`), whose accept loop swallows transient accept errors and keeps
serving, with a comment saying so.  This change makes the healthcheck loop do the same.

The accept loop is extracted into a small `serve` seam over a stream of accepted connections:

- **`serve<S, F, Fut>(incoming: S, worker_ready: F)`** where `S: Stream<Item = io::Result<TcpStream>>
  + Unpin` runs `loop { match incoming.next().await { Some(Ok(socket)) => serve one connection,
  Some(Err(e)) => log at `debug` and skip, None => break } }`.  A transient accept error is logged and
  skipped instead of ending the loop, so the critical task never resolves `Ok` on it.  `worker_ready`
  is called inline in the loop body, so **no `Send`/`Sync` bound change** is needed on the public
  `spawn` signature.
- **Production wraps the real listener**: `serve(TcpListenerStream::new(listener),
  worker_ready).await`.  `TcpListenerStream` is infinite, so `None => break` is only reachable by the
  finite stream a test supplies.  Per-connection behavior (bounded read timeout, path routing, exact
  response bytes and headers, socket close) is byte-for-byte unchanged;  serving stays sequential.
- **`tokio-stream`** is added to the node crate for `TcpListenerStream`.  It is already a workspace
  dependency and already used in `engine` and `exex`, so this pulls in nothing new.

### Why the seam (rather than an inline `loop { match listener.accept() }`)

Wrapping the listener in a `Stream` lets a test substitute a stream that yields a failing accept
followed by real connections.  A raw `TcpListener` cannot be made to fail `accept()` on demand
without process-wide fd exhaustion, which is flaky and non-portable.  The seam is the minimum needed
to make the acceptance-criteria regression test deterministic.  `worker_ready` is driven inline via
`incoming.next()` (not a per-connection `for_each` closure) specifically so the fix needs no `Sync`
bound on the public API.

## Testing

- New regression test `test_accept_error_does_not_end_loop`: injects a failing `accept()` ahead of a
  real listener via `stream::once(Err).chain(TcpListenerStream::new(listener))`, then asserts the
  endpoint still serves the next connection (`200 OK`) and the serve task has not resolved.
- The test was mutation-checked: it **fails** if the `Some(Err(..))` arm is changed to end the loop
  (the pre-fix behavior), and passes with the fix.
- Full `crates/node` health suite passes under the repo-pinned `1.94` toolchain (6/6);  `cargo +1.94
  clippy` and nightly `cargo fmt` are clean.

## Consistency check (per the issue)

`health.rs` was the only critical task using the `while let Ok(..)`-over-`accept()` shape (`rg 'while
let Ok'` finds no other).  The remaining `spawn_critical_task` call sites run epoch-scoped `run(..)`
loops that return `Ok` deliberately at the epoch boundary, which the manager already treats as an
expected graceful exit.

## Note on behavior under persistent fd pressure

Like the metrics server precedent, the loop logs and immediately retries;  under a *persistent*
`EMFILE` it will spin at `debug` level until fds free up.  This is intentional parity with the metrics
server and keeps the diff minimal.  A shared, backed-off accept helper for both endpoints would be a
reasonable follow-up but is out of scope for this fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11835,6 +11835,7 @@ dependencies = [
  "tn-types",
  "tn-worker",
  "tokio",
+ "tokio-stream",
  "tracing",
 ]
 
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -15,6 +15,7 @@ authors = [
 [dependencies]
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
+tokio-stream = { workspace = true }
 futures = { workspace = true }
 tracing = { workspace = true }
 tn-executor = { workspace = true }
--- a/crates/node/src/health.rs
+++ b/crates/node/src/health.rs
@@ -14,14 +14,16 @@
 
 use std::{future::Future, net::SocketAddr, time::Duration};
 
+use futures::{Stream, StreamExt};
 use serde::Serialize;
 use tn_types::{TaskSpawner, WorkerId, DEFAULT_WORKER_ID};
 use tokio::{
     io::{AsyncReadExt, AsyncWriteExt},
-    net::TcpListener,
+    net::{TcpListener, TcpStream},
     time::timeout,
 };
-use tracing::info;
+use tokio_stream::wrappers::TcpListenerStream;
+use tracing::{debug, info};
 
 /// Request path that serves the per-worker readiness envelope.
 const WORKERS_PATH: &str = "/health/workers";
@@ -157,9 +159,43 @@ impl HealthcheckServer {
         let listen_on = listener.local_addr()?;
         info!(target: "epoch-manager", ?listen_on, "healthcheck listening");
 
+        // wrap the listener in a stream so the accept loop is a testable seam
+        // (`serve` below): production feeds it the real `TcpListenerStream`, and
+        // tests feed it a stream that injects a failing accept.
         task_spawner.spawn_critical_task("healthcheck", async move {
-            // accept loop - no connection limits; bounded read timeout per conn
-            while let Ok((mut socket, _)) = listener.accept().await {
+            serve(TcpListenerStream::new(listener), worker_ready).await;
+            Ok(())
+        });
+
+        Ok(listen_on)
+    }
+}
+
+/// Drive the accept loop over a stream of accepted connections.
+///
+/// Serves each connection synchronously (bounded per-connection read timeout),
+/// routing the workers path to readiness and everything else to liveness.
+///
+/// A transient `accept()` error (fd exhaustion `EMFILE`/`ENFILE`,
+/// `ECONNABORTED`, `EINTR`, `ENOBUFS`) is logged and skipped: the loop must
+/// keep serving. The caller spawns this as a *critical* task, so ending the
+/// loop would resolve the task `Ok` and notify a whole-node shutdown - exactly
+/// the outage this endpoint is supposed to warn about. This mirrors the metrics
+/// server (`tn_metrics::server`), whose accept loop swallows the same errors.
+async fn serve<S, F, Fut>(mut incoming: S, worker_ready: F)
+where
+    S: Stream<Item = std::io::Result<TcpStream>> + Unpin,
+    F: Fn() -> Fut,
+    Fut: Future<Output = bool>,
+{
+    // the loop survives a transient accept error because the `Some(Err(..))`
+    // arm below logs and skips instead of breaking. `worker_ready` is called
+    // inline in the loop body (rather than from a per-connection `for_each`
+    // closure) for a separate reason: it keeps a `Sync` bound off the public
+    // `spawn` signature.
+    loop {
+        match incoming.next().await {
+            Some(Ok(mut socket)) => {
                 // read the request with a bounded timeout so a slow or idle
                 // client cannot stall the loop; treat a timeout/error/empty
                 // read as "no path" and fall through to the liveness response.
@@ -190,10 +226,12 @@ impl HealthcheckServer {
                     let _ = socket.write_all(LIVENESS_RESPONSE).await;
                 }
             }
-            Ok(())
-        });
-
-        Ok(listen_on)
+            // transient accept errors (e.g. fd exhaustion) must not kill the node
+            Some(Err(e)) => debug!(target: "epoch-manager", ?e, "healthcheck accept error"),
+            // the production `TcpListenerStream` is infinite; a finite test
+            // stream ends here once drained.
+            None => break,
+        }
     }
 }
 
@@ -216,7 +254,8 @@ mod tests {
         net::TcpStream,
     };
 
-    use super::{readiness_json, request_path, HealthcheckServer};
+    use super::{readiness_json, request_path, serve, HealthcheckServer};
+    use futures::StreamExt;
     use tn_types::{get_available_tcp_port, TaskManager};
 
     /// Send `request` to the spawned server at `addr` and return the response.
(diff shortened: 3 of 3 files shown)


grantkee (approved):
LGTM! Good catch

sstanfield (approved):
Good catch.

PR #948, opened 2026-07-17
build(deps): gate rocksdb out of the default build (closes #941)

### LINKED ISSUE #941: `build`: cut cold/CI rocksdb (librocksdb-sys) compile time via a dev/test opt-level carve-out and an optional rocksdb feature gate
## Summary

`librocksdb-sys` compiles the entire RocksDB 10.4.2 C++ engine from source: 332
translation units on a clean build (the `.o` count under
`target/*/build/librocksdb-sys-*/out/`, matching `rocksdb_lib_sources.txt`).  It
is the single largest native compile in the tree, and it reaches us only
transitively.  The `reth` binary crate's `default` feature set includes
`rocksdb`, which flows `reth/rocksdb` -> `reth-ethereum-cli/rocksdb` ->
`reth-cli-commands/rocksdb` -> `reth-provider`'s (opt-in) `rocksdb` feature (=
`dep:rocksdb`) -> `librocksdb-sys`.  Nothing in the workspace enables it
directly: `cargo tree -e features -i rocksdb` roots solely at `reth feature
"default"`.

Two independent, low-risk changes cut how long we spend on it.  Both target the
*cold / cache-miss / CI* compile: the warm incremental Rust loop already skips
rocksdb (a leaf C++ dep, rebuilt only when a reth bump changes its inputs), and a
compiler cache already avoids repeated rocksdb compiles for anyone who has one
(see the closing note).  These changes reduce the cost of the misses a cache
cannot avoid (the first build, a toolchain/flag change that busts the key, a
cache eviction) and, for change 2, remove the compile entirely.

Change 1 (opt-level carve-out) is a pure config edit with no behavior change and
is safe to land as-is.  Change 2 (rocksdb feature gate) removes the compile
outright but is behavior-adjacent, and is presented with the analysis showing it
is inert for our current MDBX-only storage, plus a guardrail.

## Change 1: stop compiling `librocksdb-sys` at `-O2` in dev/test

`.cargo/config.toml` sets `[profile.dev.package."*"] opt-level = 2` (and the same
for `[profile.test.package."*"]`).  `librocksdb-sys`'s `build.rs` uses a bare
`cc::Build` with no explicit `.opt_level()`, so cargo's per-package `OPT_LEVEL`
flows straight through as `-O2` on all 332 RocksDB translation units.  Those TUs
are template-heavy and O2-codegen-bound.  Under our OOM-safe `cargo -j 2`
discipline they are also heavily serialized, so the per-TU cost lands almost
directly on wall-clock.

A more-specific per-package override beats the `"*"` glob (cargo profile
overrides are first-match-wins: a named package outranks the glob) and drops
those TUs to `-O1`, which keeps most of RocksDB's runtime throughput while
cutting a large fraction of the compile:

```toml
[profile.dev.package.librocksdb-sys]
opt-level = 1
[profile.dev.package.zstd-sys]
opt-level = 1

[profile.test.package.librocksdb-sys]
opt-level = 1
[profile.test.package.zstd-sys]
opt-level = 1
```

Details that make this correct:

- The `[profile.test...]` copy is required.  The test profile re-declares the
  `"*"` glob at `opt-level = 2`, so without a test-side override, `cargo test`
  keeps compiling rocksdb at O2.
- Include `zstd-sys` (also vendored and compiled from source): it honors
  `OPT_LEVEL`, so the override takes effect.  `bzip2-sys` and `libz-sys` honor it
  too but are tiny, so they are optional to add.
- The one exception is `lz4-sys`: its `build.rs:25` calls `.opt_level(3)`
  unconditionally, which shadows `OPT_LEVEL`, so a profile override on it is an
  inert no-op.  (`zstd-sys` only shadows `OPT_LEVEL` under its non-default `thin`
  feature, which is not enabled here, so it does honor the override.)
- `opt-level` is codegen only; `NDEBUG` is defined regardless, so there is no
  correctness or on-disk-format change.  Release/bench builds are unaffected and
  keep O2+.
- Cold / cache-miss win only: warm rebuilds already skip rocksdb, and this busts
  a compiler cache's key for these crates once, after which they re-cache at O1.

Measure:

```
cargo clean -p librocksdb-sys && time cargo build -p librocksdb-sys   # before vs after
```

## Change 2: gate the `rocksdb` feature off (optional;  removes the compile entirely)

reth v1.11.3 makes rocksdb optional.  `reth-provider` selects its backend by cfg
(`crates/storage/provider/src/providers/mod.rs`):

```rust
#[cfg_attr(all(unix, feature = "rocksdb"), path = "rocksdb/mod.rs")]
#[cfg_attr(not(all(unix, feature = "rocksdb")), path = "rocksdb_stub.rs")]
```

With the feature off, `reth_provider::providers::RocksDBProvider` resolves to
`rocksdb_stub.rs`, a zero-size placeholder whose `RocksDBProvider::new` returns
`Ok(Self)`, and `librocksdb-sys` is never built.  No reth patch is required, only
a workspace `Cargo.toml` edit.

The feature enters only through the `reth` binary crate's default set
(`bin/reth/Cargo.toml`):

```
default = ["jemalloc","otlp","otlp-logs","reth-revm/portable","js-tracer",
           "keccak-cache-global","asm-keccak","min-debug-logs","rocksdb"]
```

So the change is `default-features = false` on the `reth` dep, re-adding the
default set minus `rocksdb`.  One subtlety: `reth-revm/portable` is a
dependency-feature, not a feature of `reth`, so it cannot be re-added through
reth's own feature list.  Re-enable it on our existing `reth-revm` dep instead.

```toml
# Cargo.toml
reth = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3",
  default-features = false,
  features = ["jemalloc","otlp","otlp-logs","js-tracer","keccak-cache-global","asm-keccak","min-debug-logs"] }

reth-revm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3", features = ["portable"] }
```

Verify the edge is gone (this also confirms reth's default was the only enabler,
since `reth-provider` has no `default` feature that pulls rocksdb on its own):

```
cargo tree -e features -i rocksdb
# expect: error: package ID specification `rocksdb` did not match any packages
```

### Why this is runtime-safe for us today

`crates/tn-reth/src/lib.rs:681` constructs the provider and forwards it into
`ProviderFactory::new`:

```rust
let rocksdb_provider = reth_provider::providers::RocksDBProvider::new(datadir.data_dir())?;
// ...
let provider_factory = ProviderFactory::new(
    database, Arc::clone(&node_config.chain),
    StaticFileProvider::read_write(datadir.static_files())?,
    rocksdb_provider, runtime,
)?;
```

We call only `::new()` and hand the value on.  We never call any other
`RocksDBProvider` method, all of which the stub turns into silent no-ops.  Inside
reth, every read/write dispatch to the rocksdb provider is cfg-guarded on
`all(unix, feature = "rocksdb")` (per `rocksdb_stub.rs`'s header: "All method
calls are cfg-guarded in the calling code, so only type definitions are needed
here").  Our real storage is MDBX (`Arc<DatabaseEnv>`), and we do not enable
reth's storage-v2 path that routes history indices to rocksdb (no
`StorageSettings` / `storage_v2` use in the tree).  So with the feature off, the
stub is constructed, held, and never exercised.

### The guardrail

This holds only while we do not opt into reth's rocksdb-backed storage-v2, or
call a `RocksDBProvider` method beyond `::new()`.  Two ways to keep that honest:

- **Recommended:** expose it as a workspace feature (e.g. a `rocksdb` feature on
  `tn-reth`, off by default, that re-adds `reth/rocksdb`) so dev and CI build
  without it and anyone who needs storage-v2 flips one switch.  Production stays
  unchanged from today unless deliberately toggled.
- Or, if gating it off unconditionally, add a note/assert that reth storage-v2
  must not be enabled while rocksdb is gated out, since the stub silently no-ops
  those writes.

Maintenance cost: `default-features = false` forks us off reth's default set, so
every reth bump must re-reconcile this curated list.  A silently dropped
`jemalloc` / `asm-keccak` / `portable` regresses the allocator / keccak / revm
codegen without a compile error.  Adding the `cargo tree -e features -i rocksdb`
check to CI catches a future reth bump that silently re-introduces the dep.

## Recommended sequencing

Land change 1 now (pure config, zero behavior change).  Adopt change 2 as an
off-by-default `tn-reth` feature, with the `cargo tree` check wired into CI.

## Note: a compiler cache is the bigger structural lever, orthogonal to both

Because `librocksdb-sys` drives its C++ build through the `cc` crate, and `cc`
forwards a recognized `RUSTC_WRAPPER` (`sccache` / `cachepot` / `buildcache`) to
the C/C++ compiler as well as to `rustc` (see `cc`'s `rustc_wrapper_fallback`,
`VALID_WRAPPERS`), setting `RUSTC_WRAPPER=sccache` caches rocksdb's object files
across clean builds and fresh worktrees.  rocksdb's cache key is very stable
because reth is git-pinned, so it recompiles from source only on a reth bump, a
toolchain/flag change, or a cache eviction.  If CI is not already pointed at a
shared sccache backend (`SCCACHE_GHA_ENABLED=true`, or `SCCACHE_BUCKET=` for S3),
doing so eliminates the from-source rocksdb compile on CI runs regardless of
changes 1 and 2.

## Notes

- Counts and line numbers are from worktree `telcoin-network-926` against reth
  `v1.11.3` (checkout `d6324d6`).  Re-confirm against `main` before a PR.
- `edge` implies `rocksdb` (`edge = ["rocksdb"]` in both `reth` and
  `reth-provider`), so do not enable `edge` while rocksdb is gated out.
- 332 is the count of `.o` files compiled into the library (per
  `rocksdb_lib_sources.txt`).  The vendored source tree contains many more `.cc`
  files (tests, tools, benchmarks, Java bindings) that are not compiled into the
  library.

MavenRain:
## Problem

`librocksdb-sys` compiles the entire RocksDB 10.4.2 C++ engine from source: 332
translation units on a clean build, the single largest native compile in the tree.  It
reaches us only transitively, through the `reth` binary crate's `default` feature set:
`reth/rocksdb` -> `reth-ethereum-cli/rocksdb` -> `reth-cli-commands/rocksdb` ->
`reth-provider/rocksdb` (= `dep:rocksdb`) -> `librocksdb-sys`.  Nothing in the workspace
enables it directly (`cargo tree -e features -i rocksdb` roots solely at `reth feature
"default"`), and storage is MDBX-only.

This is the remaining half of #941: the dev/test `opt-level` carve-out landed via #913
(the `package."*"` glob is now `opt-level = 1`), and this change removes the rocksdb
compile from default builds outright.

## Why it is inert for current behavior

reth v1.11.3 selects the storage backend by cfg in
`crates/storage/provider/src/providers/mod.rs`:

```rust
#[cfg_attr(all(unix, feature = "rocksdb"), path = "rocksdb/mod.rs")]
#[cfg_attr(not(all(unix, feature = "rocksdb")), path = "rocksdb_stub.rs")]
```

With the feature off, `RocksDBProvider` resolves to a zero-size stub whose
`new(_path: impl AsRef<Path>)` returns `Ok(Self)`, signature-compatible with the real
provider.  Our sole touchpoint is `RocksDBProvider::new` in tn-reth's
`init_provider_factory` (`crates/tn-reth/src/lib.rs`), which just forwards the value into
`ProviderFactory::new`.  There is no `storage_v2` / `StorageSettings` usage in the tree,
so no rocksdb code path is reachable today.  No Rust changes are needed;  the diff is
manifests plus one CI job.

## Changes

- **Workspace `Cargo.toml`**: the `reth` dependency becomes `default-features = false`
  with the default set minus `rocksdb` (jemalloc, otlp, otlp-logs, js-tracer,
  keccak-cache-global, asm-keccak, min-debug-logs).  `reth-revm/portable`, the one
  default entry that is a dependency-feature rather than a reth feature key (and so
  cannot appear in a dependency's `features` list), moves to the workspace `reth-revm`
  entry;  tn-reth depends on reth-revm directly, so feature unification applies it to
  reth's copy in every build that contains reth.
- **Opt-in feature chain**: an off-by-default `rocksdb` feature propagates
  `tn-reth -> tn-node -> telcoin-network-cli -> telcoin-network`, mirroring the existing
  `adiri` chain, so `cargo build -p telcoin-network --features rocksdb` restores today's
  graph exactly.  It must be enabled before any storage-v2 / `StorageSettings`
  experimentation.
- **CI guard**: a `rocksdb-gate` job asserts rocksdb stays out of the default feature
  graph by counting rocksdb packages in the feature-resolved `cargo metadata` output (a
  structured check rather than scraping cargo's error wording;  a cargo or jq failure
  fails the job, so infrastructure errors fail closed rather than passing silently).  It
  resolves the dependency graph without compiling anything and is wired into
  `ci-success`.  This catches a future reth bump whose default set silently
  re-introduces the dep, and the `edge = ["rocksdb"]` implication as well.  Verified in
  both directions locally: 0 hits with default features, 1 hit with
  `--features telcoin-network/rocksdb`.

## Verification

All checks under the repo-pinned 1.94 toolchain, `--locked` throughout (Cargo.lock is
unchanged;  rocksdb stays locked for the opt-in path).

- `cargo tree --locked -e features -i rocksdb` now fails to match any package with
  default features, and restores the full chain
  (`rocksdb <- reth-provider <- reth <- tn-reth`) with `--features rocksdb`.
- Whole-graph equivalence, `cargo tree -f '{p} {f}'` rooted at `telcoin-network`, before
  vs after:
  - **With `--features rocksdb`**: identical to main's default graph except the vacuous
    `default` label on reth (all nine constituent features are enabled individually;  no
    code cfg-gates on `feature = "default"`) and the new tn `rocksdb` feature labels
    themselves.  The opt-in configuration is exactly what CI builds on main today, which
    is why no separate rocksdb-on compile is needed.
  - **With default features**: the only drops are the native compression pile
    (`librocksdb-sys`, `rocksdb`, `bzip2-sys`, `libz-sys`, `lz4-sys`, and `zstd-sys`
    entirely, since nothing else consumed it) plus `rocksdb` feature labels on reth
    crates.
- `cargo check --workspace --all-targets --locked` passes with the gate off (the
  unix-without-rocksdb stub path is the one genuinely new compile configuration).

## Behavioral note

`tikv-jemalloc-sys` loses `unprefixed_malloc_on_supported_platforms`, which entered the
graph only as `librocksdb-sys`'s jemalloc plumbing (so RocksDB's C++ allocations would
route through jemalloc).  With it gone, C-library allocations (e.g. MDBX's internal
mallocs;  the bulk of MDBX is mmap) revert to the system allocator.  The Rust global
allocator is unchanged.  If routing C allocations through jemalloc is considered
load-bearing, it is a one-line re-add on a direct `tikv-jemalloc-sys` entry;  flagging it
rather than deciding it here.

## Notes

- `cargo clippy --all --all-features` (CI clippy job) still compiles rocksdb by design:
  it keeps the opt-in configuration linting and compiling so the gated path cannot
  bit-rot, and its cost is absorbed by that job's dependency cache.  The default-features
  test job and every local dev/test build skip the 332-TU compile entirely.
- Every reth bump must re-reconcile the curated feature list against reth's default set;
  the `rocksdb-gate` job turns a silent re-introduction into a PR-time failure.

Closes #941.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -111,18 +111,55 @@ jobs:
       - name: Run tests
         run: cargo nextest run --locked --workspace --exclude tn-faucet --no-fail-fast --profile ci
 
+  # Guard: rocksdb (librocksdb-sys, the entire RocksDB C++ engine and the single largest
+  # native compile in the tree) must stay out of the default feature graph. It is gated
+  # behind the off-by-default `rocksdb` feature; this catches a future reth bump whose
+  # default set silently re-introduces it. Resolves the graph only - nothing is compiled.
+  rocksdb-gate:
+    needs: skip-maintainers
+    if: >-
+      needs.skip-maintainers.outputs.is_maintainer == 'false' &&
+      github.event.pull_request.draft != true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        run: rustup show active-toolchain || rustup toolchain install
+
+      - name: Cache dependencies
+        uses: Swatinem/rust-cache@f0deed1e0edfc6a9be95417288c0e1099b1eeec3
+        with:
+          shared-key: rocksdb-gate-cache
+          cache-on-failure: "true"
+
+      - name: Assert rocksdb absent from default feature graph
+        run: |
+          # Structured check: count rocksdb packages in the feature-resolved metadata
+          # graph instead of scraping cargo's error wording. A cargo/jq failure fails
+          # the assignment (bash -e -o pipefail), so infrastructure errors fail closed.
+          hits=$(cargo metadata --locked --format-version 1 \
+            | jq '[.resolve.nodes[].id | select(test("[/#]rocksdb@"))] | length')
+          if [[ "$hits" -ne 0 ]]; then
+            echo "librocksdb-sys re-entered the default dependency graph ($hits rocksdb package(s) resolved):"
+            cargo tree --locked -e features -i rocksdb || true
+            exit 1
+          fi
+          echo "rocksdb absent from the default feature graph"
+
   # Gate job for branch protection - depends on all checks
   ci-success:
     name: CI Success
     runs-on: ubuntu-latest
     if: always()
-    needs: [fmt, clippy, test]
+    needs: [fmt, clippy, test, rocksdb-gate]
     steps:
       - name: Check all jobs passed
         run: |
           if [[ "${{ needs.fmt.result }}" == "failure" || \
                 "${{ needs.clippy.result }}" == "failure" || \
-                "${{ needs.test.result }}" == "failure" ]]; then
+                "${{ needs.test.result }}" == "failure" || \
+                "${{ needs.rocksdb-gate.result }}" == "failure" ]]; then
             echo "One or more checks failed"
             exit 1
           fi
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,7 +96,24 @@ secp256k1 = { version = "0.31", default-features = false, features = [
 rpassword = "7.4.0"
 
 # execution
-reth = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
+#
+# Default features minus `rocksdb`: librocksdb-sys (the entire RocksDB C++ engine, 332
+# translation units, the single largest native compile in the tree) enters the graph only
+# through reth's default set, and our storage is MDBX-only (`RocksDBProvider` resolves to
+# reth's zero-size stub with the feature off). Re-enable explicitly via the off-by-default
+# `rocksdb` feature on tn-reth / tn-node / telcoin-network-cli / telcoin-network.
+# `reth-revm/portable` (a dependency-feature, not nameable in this list) is re-enabled on
+# the `reth-revm` entry below and reaches reth's copy through feature unification.
+# NOTE: reth's `edge` feature implies `rocksdb`; do not enable it while this gate is off.
+reth = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3", default-features = false, features = [
+    "jemalloc",
+    "otlp",
+    "otlp-logs",
+    "js-tracer",
+    "keccak-cache-global",
+    "asm-keccak",
+    "min-debug-logs",
+] }
 reth-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
 reth-primitives-traits = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
 reth-chainspec = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
@@ -115,7 +132,10 @@ reth-network-api = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3
 reth-network-peers = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
 reth-evm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
 reth-execution-types = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
-reth-revm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
+# `portable` replaces the `reth-revm/portable` entry dropped from the curated `reth`
+# feature list above (dependency-features cannot appear in a dependency's `features` key);
+# tn-reth depends on reth-revm directly, so unification applies it to reth's copy too.
+reth-revm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3", features = ["portable"] }
 reth-rpc = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
 reth-rpc-eth-api = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
 reth-rpc-eth-types = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
--- a/bin/telcoin-network/Cargo.toml
+++ b/bin/telcoin-network/Cargo.toml
@@ -20,6 +20,10 @@ tn-node = { workspace = true }
 # is actually compiled into the binary. Enabled for the `docker-adiri` image via a Docker build arg;
 # default/mainnet builds omit it. See `tn_types::forks::CONSENSUS_REGISTRY_FORK_EPOCH`.
 adiri = ["telcoin-network-cli/adiri", "tn-node/adiri"]
+# Opt back in to reth's RocksDB storage backend (librocksdb-sys). Off by default: storage
+# is MDBX-only today, so the node uses reth's zero-size `RocksDBProvider` stub and skips
+# compiling the RocksDB C++ engine. Must be enabled before any storage-v2 experimentation.
+rocksdb = ["telcoin-network-cli/rocksdb", "tn-node/rocksdb"]
 
 [lints]
 workspace = true
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -50,6 +50,8 @@ adiri = [
     "tn-storage/adiri",
     "tn-types/adiri",
 ]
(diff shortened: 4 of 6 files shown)


grantkee (approved):
LGTM! This should definitely speed up build times! There's one flag from CodeQL about token permissions, but I will follow up with that in a separate PR

sstanfield (approved):

PR #951, opened 2026-07-17
Finish the wait_until migration for eject/epoch e2e helpers

### LINKED ISSUE #950: Inconsistent, hand-rolled polling loops in e2e test helpers
## Problem

The e2e test helpers in `crates/e2e-tests/tests/it/` (`common.rs`, `epochs.rs`, `eject.rs`) each implement their own ad-hoc polling loop for waiting on asynchronous node state: a manual `Instant::now() + Duration` deadline, a `loop { ... sleep(1s) ... }` body, and a bespoke `assert!`/`eyre::ensure!` on timeout. Several of these loops are duplicated near-verbatim across files (e.g. waiting for a node's head to advance, waiting for an epoch to change, waiting for RPC availability).

This duplication makes the helpers harder to maintain and inconsistent in behavior: some loops sleep for a full second between polls regardless of how time-sensitive the check is, some don't name what they're waiting for in the failure message, and fixing a bug in one copy (e.g. adjusting poll cadence or improving the timeout error) doesn't propagate to the others. It also means new tests are tempted to hand-roll yet another copy of the same loop rather than reaching for a shared primitive.

Additionally, `ProcessGuard` had no way to wait for a child process to exit *on its own* (as opposed to being signaled) — a gap that blocks writing tests that assert a node reaches a designed, self-inflicted exit.

## Solution

Migrate the remaining hand-rolled deadline loops in the eject/epoch e2e helpers onto the shared `wait_until` (async) and `wait_until_blocking` (sync) primitives from `tn_test_utils`. Each call site should name what it's waiting for in the timeout message, so failures are self-explanatory instead of a generic "stuck" assertion.

Concretely: `loop_epochs`'s epoch-change poll, `wait_for_rpc`, and the eject test's "node head advanced" checks should all go through `wait_until`, consolidated into a single `wait_for_head_at_least` helper reused by both eject tests. `assert_epoch_reached` should move from being a private helper duplicated at the bottom of `eject.rs` into the shared `common.rs` module.

Separately, `ProcessGuard` should expose a `wait_for_natural_exit` method that polls `try_wait()` (via `wait_until_blocking`) for a specific child to exit without signaling it, clearing that child's slot once reaped so `kill_all`/`Drop` never re-signal a possibly-reused pid. This is a prerequisite for any e2e test that needs to assert a node terminates on its own with a specific exit code, rather than being killed by the test harness.

grantkee:
- Replace the remaining hand-rolled `Instant::now() + Duration` poll loops in `common.rs` (`loop_epochs`, `wait_for_rpc`) and `eject.rs` (post-burn head-advance checks) with `wait_until`/`wait_until_blocking`, each naming what it's waiting for.
- Add `wait_for_head_at_least` in `common.rs` and reuse it from both `test_validator_ejected_from_future_committee_only` and `test_validator_ejected_from_current_committee_mid_epoch`, replacing duplicated loop bodies.
- Move `assert_epoch_reached` from a private helper at the bottom of `eject.rs` into `common.rs` so it's reusable.
- Add `ProcessGuard::wait_for_natural_exit` — polls a specific child's `try_wait()` via `wait_until_blocking` and clears its slot once reaped, so `Drop`/`kill_all` never re-signal a reused pid.
- Add `burn_with_retry` in `eject.rs`: retries a governance burn tx once with the same nonce if its confirmation is lost to an epoch-boundary race.
- Strengthen `assert_epoch_records_verify` to compare the executed block's actual hash against the epoch record's committed hash, not just block existence — catches cross-node divergence that same-height-different-contents would otherwise miss.
- Remove now-redundant pre-pad `sleep(EPOCH_DURATION + 1)` calls in `epochs.rs` ahead of `loop_epochs`, since its first poll already waits for the epoch to change.
- New test: `test_network_halts_when_ejections_shrink_committee_below_tolerance` — burns two of five committee members in one epoch to prove every validator fail-stops (exit code 1, with the expected stderr) at the next epoch boundary when the committee shrinks below `committee_compatible` tolerance, and that no endpoint answers RPC afterward.

closes #950

### DIFF
--- a/crates/e2e-tests/tests/it/common.rs
+++ b/crates/e2e-tests/tests/it/common.rs
@@ -23,7 +23,7 @@ use nix::{
 use secp256k1::{Keypair, Secp256k1, SecretKey};
 use serde_json::Value;
 use std::{
-    cell::{Cell, RefCell},
+    cell::RefCell,
     collections::HashMap,
     fmt::Debug,
     ops::RangeInclusive,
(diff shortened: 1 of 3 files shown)


crates/e2e-tests/tests/it/eject.rs
```diff
@@ -588,19 +595,217 @@ async fn test_validator_ejected_from_current_committee_mid_epoch() -> eyre::Resu
...
+    // At the close of `halt_epoch` every validator's `build_epoch_record` reads the 3-member
+    // committee, rejects it against the previous record's promise, and the node exits through
+    // `main`'s error path. Exit code 1 distinguishes the designed error-exit from a signal
+    // death (`ExitStatus::code()` returns `None` for signals).
+    //
+    // All 5 must exit — the epoch cannot close for anyone. Known residual risk: a validator
+    // lagging at the boundary could in principle miss the closing commit once its peers exit
+    // first; the 4-epoch-duration bound per node absorbs realistic same-host lag, and a genuine
+    // stall should fail the test rather than be tolerated.
+    for idx in 0..nodes.len() {
```
MavenRain: Consider waiting on these concurrently by turning this into a stream.  Otherwise, the upper bound on the wait is the sum of the max timeout for each node times the number of nodes.

crates/e2e-tests/tests/it/common.rs
```diff
@@ -860,19 +901,15 @@ pub(crate) async fn loop_epochs(
...
-        };
+        wait_until(
+            Duration::from_secs(epoch_duration * 4),
+            &format!("epoch to change on iteration {i}"),
+            || async {
+                Ok(consensus_registry.getCurrentEpochInfo().call().await? != current_epoch_info)
+            },
+        )
+        .await?;
+        let new_epoch_info = consensus_registry.getCurrentEpochInfo().call().await?;
```
MavenRain: We should be able to reuse the result of the call made within the `wait_until` block, no?

MavenRain (dismissed):
Two suggestions (one strong one), but nothing technically blocking

crates/e2e-tests/tests/it/eject.rs
grantkee: Good idea - addressed in [d64e5a5](https://github.com/Telcoin-Association/telcoin-network/pull/951/commits/d64e5a55707df0cb875f335aa7c09b25d8b3de0e)

crates/e2e-tests/tests/it/common.rs
grantkee: Good idea - addressed in [d64e5a5](https://github.com/Telcoin-Association/telcoin-network/pull/951/commits/d64e5a55707df0cb875f335aa7c09b25d8b3de0e)

MavenRain (approved):
GTG

sstanfield (approved):

PR #952, opened 2026-07-17
Pin epoch-entry reads to the previous epoch's closing block

### LINKED ISSUE #968: Mid-epoch validator ejection can desync a restarting committee member's epoch view
## Problem

A node derives its view of the current epoch — committee membership, quorum threshold, leader schedule, and epoch-scoped fee/reward state — from a read of the `ConsensusRegistry` taken at the canonical execution tip. That's correct the first time a node enters an epoch, at the boundary. But a node can also *re-enter* the same epoch later: after a crash-restart, or through a mid-epoch ModeChange. On that path, a canonical-tip read no longer represents "the state at epoch start" — it represents whatever the chain has advanced to since.

A governance validator ejection (`ConsensusRegistry::burn`) swap-and-pops the ejected validator out of the *current* epoch's stored committee arrays immediately, mid-epoch. A node that re-enters after such an ejection and reads the canonical tip derives a smaller committee than the one the rest of the fleet is still running for that epoch — a different quorum threshold and a different round-robin leader schedule. It can't verify certificates produced under the real (pre-ejection) committee, so its consensus height stalls until the next epoch boundary. If several committee members restart during the same window, the network risks losing liveness mid-epoch.

It gets worse if the restarted node does still reach the epoch close under the wrong view. Reward tallying resolves each accumulated leader-count against whichever committee is currently set; an authority the committee no longer contains has its row silently dropped — not zeroed, not errored. A node that seeded its rewards counter from the post-ejection committee omits the ejected validator's leader-reward row when it builds the epoch-closing block's withdrawals, producing a different `withdrawals_root`, hence a different closing-block hash, hence a different epoch-record digest than every peer that kept the pre-ejection view. That's a silent, deterministic divergence between otherwise-correct nodes — a fork risk, not just a liveness hiccup.

This was surfaced during review of the preceding branch that introduced tolerance for mid-epoch committee shrinkage in epoch records; that review flagged the restart/re-entry path as an unverified gap worth a dedicated fix.

## Solution

The execution layer should expose a dedicated pinned read: given the current epoch, resolve the previous epoch's closing block (the block whose execution ran `concludeEpoch` and seated this epoch's committee) — genesis, for epoch 0 — and read the committee, `EpochInfo`, and epoch-start timestamp from that fixed header instead of the canonical tip. Because `concludeEpoch` writes those boundary scalars exactly once, the pin resolves to the identical header regardless of which tip is used to look it up, so every entry shape — a fresh boundary crossing, a crash-restart replay, a ModeChange re-entry — derives an identical epoch view no matter what happens on-chain after the boundary.

Every epoch-scoped entry read on the node side — committee construction, the next/next-plus-one committee prefetch, and epoch-entry base-fee derivation — should route through this pinned primitive exclusively. The engine's public API should stop exposing an unpinned committee read on the entry path altogether, so there's no ambient way for a future caller to reintroduce a tip-based read by accident; callers should have to supply the pin header explicitly. Epoch-CLOSE reads are a deliberate exception: the epoch record is supposed to commit whatever committee shape holds at the moment the epoch closes, ejections included, so the close path should keep reading at the epoch-closing block rather than the epoch-start pin — the two call sites read different headers on purpose, and that distinction should stay explicit in the code.

As a defense-in-depth check, the consensus-pack layer should compare a same-epoch re-entry's freshly derived committee against the committee snapshot already persisted for that epoch's pack. A mismatch can only mean an unpinned read crept back onto the entry path, so it's worth flagging loudly — a warning rather than a hard failure, since the persisted snapshot remains authoritative for decoding that epoch's output either way.

grantkee:
- Add `RethEnv::epoch_state_at_epoch_start` plus pinned `validators_for_epoch_at_block` / `bls_pubkeys_for_epoch_at_block` reads, and generalize the ConsensusRegistry EVM-read helpers to run against an arbitrary header instead of only the canonical tip.
- Route every epoch-entry read in `run_epoch`/`start_epoch` and the engine layer through the pinned primitive; the engine no longer exposes an unpinned committee read on the entry path, so it can't be reached by accident.
- Keep epoch-close reads (`close_epoch.rs`) pinned to the epoch-closing block on purpose — refactored to compute that header once up front (one-header discipline) rather than changing what it reads.
- Add a warn-only tripwire in the consensus pack's committee cross-check that flags a same-epoch re-entry whose derived committee disagrees with the pack's persisted epoch-start snapshot.
- New coverage: a `RewardsCounter` unit test isolating the row-drop semantics that make an unpinned re-entry dangerous, a four-leg `RethEnv` determinism test around a mid-epoch burn + re-entry, and a flagship e2e test that kills and restarts a committee member mid-epoch after a different validator's ejection — asserting the restart rejoins consensus and every node closes the epoch with hash-identical records.

closes #968

### DIFF
--- a/crates/tn-reth/src/lib.rs
+++ b/crates/tn-reth/src/lib.rs
@@ -1709,6 +1709,12 @@ impl RethEnv {
     /// - get current epoch info
     /// - getValidator token id by address
     /// - getValidator info by token id
+    ///
+    /// The committee arrays this returns mutate mid-epoch: a governance `burn` swap-and-pops the
+    /// ejected validator out of the CURRENT epoch's stored committees immediately, so tip reads
+    /// before and after the burn disagree. Epoch-scoped consensus reads (committee construction,
+    /// rewards seeding, quorum) must use [`Self::epoch_state_at_epoch_start`] instead; the tip
+    /// read remains correct for point-in-time queries.
     pub fn epoch_state_from_canonical_tip(&self) -> eyre::Result<EpochState> {
         let canonical_tip = self.canonical_tip();
         debug!(target: "engine", ?canonical_tip, "retrieving epoch state from canonical tip");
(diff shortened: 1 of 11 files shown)


crates/tn-reth/src/lib.rs
```diff
@@ -1756,6 +1762,71 @@ impl RethEnv {
...
+            "retrieving epoch state at epoch start"
+        );
+
+        let state = self.epoch_state_at_header(&pin_header)?;
+
+        // Tripwire: `concludeEpoch` executes INSIDE the closing block, so the registry at the
+        // closing header already reports the entered epoch. This can only fire if the registry's
+        // `blockHeight == closing block + 1` convention ever breaks — running a stale committee
+        // is a consensus-safety failure, so fail hard instead of returning the mismatched state.
+        debug_assert_eq!(
```
MavenRain: We should remove this `debug_assert_eq` and let the subsequent `return Err` underneath flow, particularly since we are within an `eyre::Result<(EpochState, SealedHeader)>` expression.

MavenRain (dismissed):
Just the one comment about the assert duplicating/causing dead code

sstanfield (dismissed):
LGTM!

crates/tn-reth/src/lib.rs
grantkee: Good catch - addressed in [3c91c17](https://github.com/Telcoin-Association/telcoin-network/pull/952/commits/3c91c17b34e15ed9b427d5130617e9b8cfe79d53)

MavenRain (approved):

PR #964, opened 2026-07-19
fix(consensus): close the two anti-equivocation guards #940 left non-durable (#963)

### LINKED ISSUE #963: anti-equivocation guard records still written non-durably after #940 (certifier own-certificate, vote-recast certificate check)
## Background

#934 established the hazard class: the epoch DB is a full-memory `LayeredDatabase` whose `insert` updates an in-memory map and enqueues the durable disk write to a background thread, returning before the write is durable (`crates/storage/src/stores/layered_db.rs:418-420`).  On crash the in-memory map is repopulated from disk only, so a record enqueued but not yet flushed is lost on restart.  If an anti-equivocation guard reads that lost record and consequently takes a *different* action than it did before the crash, an honest node equivocates.

#940 closes this for two records by awaiting a durable barrier before externalizing:

- `LastProposed`, barrier at `crates/consensus/primary/src/proposer.rs:278` before the header is broadcast.
- `Votes`, barrier at `crates/consensus/primary/src/network/handler.rs:878` before the vote is returned to the peer.

A full inventory of the primary consensus crate finds two more records in the same class that #940 leaves non-durable.  Both are `TableHint::Epoch` (`crates/storage/src/lib.rs:91,93,94`), so both route to the same full-memory layered DB and carry the identical crash-loss window.

Note that #940 makes both of these gaps *easier to reach*, not harder: by making `LastProposed` and `Votes` durable, a restarted node now deterministically re-proposes the identical header and reliably re-enters the guards below, promoting these two records from occasional backstops to load-bearing on every restart.

## Site 1: certifier own-certificate (`ProposedCertificates`)

`spawn_header_proposal` writes its own certificate, then externalizes it, with no barrier in between:

```rust
// crates/consensus/primary/src/certifier.rs
// :443  non-durable write of the guard record
if let Err(e) = self.config.node_storage().insert::<ProposedCertificates>(&header_digest, &certificate) { ... }
// :448  process_own_certificate -> process_certificate(cert, false) -> forward_verified_certs,
//       which sends on the parents bus (crates/consensus/primary/src/state_sync/cert_validator.rs:158)
//       and triggers certificate fetching. This is already an externalization point.
if let Err(e) = self.state_sync.process_own_certificate(&mut certificate).await { ... }
// :454  gossip publish
if let Err(e) = self.network.publish_certificate(certificate).await { ... }
```

The guard this record backs is at `certifier.rs:418-430`, whose own comment (`:421-425`) states the stakes: re-processing "could produce signature equivocation and destroy deterministic randomness (based on the leader signature)".  `ProposedCertificates` has exactly three uses in the tree: the guard read at `:419`, this write at `:443`, and the epoch clear at `crates/node/src/manager/node/close_epoch.rs:268`.  On crash between `:443` and the background commit, the record is lost;  on restart the guard at `:419` misses and the certifier re-runs `propose_header` for the same header.

Consequence: the re-run re-collects votes over an unordered channel and breaks at the first quorum (`certifier.rs:320`), so it can aggregate a *different* 2f+1 subset and produce a distinct aggregate signature for the same header.  Because `randomness = keccak256(leader.aggregated_signature())` feeds the committed-subdag digest and thence the consensus header, a distinct aggregate also perturbs deterministic randomness.  The distinct-aggregate step is plausible rather than proven (subset equality across two independent collection rounds is possible but not guaranteed);  the free variable is identified and unconstrained.

**Fix (clean, one line in #940's own pattern):** await the barrier immediately after the insert error check, *before* `:448` (not merely before `:454`, since `:448` already externalizes):

```rust
// crates/consensus/primary/src/certifier.rs, ~:446
self.config.node_storage().persist_durable::<ProposedCertificates>().await;
```

## Site 2: vote-recast certificate-existence check

The vote handler's recast guard decides whether it may sign a *second, different* vote for the same author and round.  It bases that decision on whether a certificate already exists:

```rust
// crates/consensus/primary/src/network/handler.rs:824-829
let cert_exists = self
    .consensus_config
    .node_storage()
    .read_by_index(header.author(), header.round())
    .unwrap_or(None)
    .is_some();
if cert_exists {
    // old vote contributed to a real certificate -> refuse: return AlreadyVoted (:836)
} else {
    // "old vote was never aggregated" -> fall through and sign a DIFFERENT vote (:840-847)
}
```

`read_by_index` reads `CertificateDigestByOrigin` then `Certificates` (`crates/storage/src/stores/certificate_store.rs:224-227`), both `TableHint::Epoch`, both written non-durably (via `write_all` at `crates/consensus/primary/src/state_sync/cert_manager.rs:196`, whose `txn.commit()` only enqueues `DBMessage::CommitTxn`).  The two legs share one physical txn, so they are lost together or not at all.

Failure mode: a certificate *was* formed (this node's earlier vote contributed to a real certificate on the network), the node crashes, the certificate record is lost on restart, so `cert_exists` reads `false`, the permissive branch runs, and the node signs a second, different vote for the same (author, round) while the first vote is already aggregated into a live certificate.  This is a *directly* distinct signed message, so the equivocation here is proven, not merely plausible.  Separately, `.unwrap_or(None)` at `:828` collapses a storage *error* into "no cert", i.e. it fails open into the same permissive branch.

#940 hardens the *other* leg of this same guard (the `Votes`/`VoteInfo` record that gets you into it at `handler.rs:787-816`) but not this certificate-existence leg.

**Fix (a real decision, not a one-liner):** the tables this leg reads (`Certificates`, `CertificateDigestByOrigin`) are the hot certificate tables that the memory layer actually benefits, so the raw-MDBX route contemplated in the table-placement follow-up is the wrong tool here.  Two viable options:

1. Fail **closed**: treat an unknown or errored `read_by_index` as `cert_exists = true` and refuse to re-vote.  Safe, cheap, but costs a round of liveness in the legitimate "proposer killed before quorum, restarted with a new header" case the comment at `:820-822` describes.
2. Make certificate acceptance durable before the node can act as if it has voted, so the guard record is never behind the vote.  Larger, and it reintroduces a barrier on the certificate hot path.

This tradeoff (safety versus a bounded liveness cost on that one recovery path) is the substance of the issue and should be decided explicitly.

## Scope is bounded

The inventory is complete for this class.  One nearby write is deliberately excluded: the inactive-CVV catch-up cache at `handler.rs:267` writes a peer certificate to the same tables but externalizes nothing and does not vote, so a lost cache entry causes only a re-fetch, not a distinct signed message.  It is not a hazard.

## Relationship to the other #940 follow-ups

- Site 1 could also be resolved by the table-placement follow-up if `ProposedCertificates` moves to synchronously-durable storage;  Site 2 cannot, because its tables stay layered.  This issue tracks the *safety gap*;  the table-placement issue tracks one possible *mechanism* for Site 1 only.  Neither gap should be parked behind that design investigation.
- Whichever barrier API the persist-convergence follow-up settles on, Site 1's fix uses it;  the barrier semantics exist under either name.

## Severity

Both sites are crash-recovery-only: the epoch DB is `full_memory = true` (`crates/storage/src/stores/composite_db.rs:33`), so the in-memory map holds each record for the process lifetime and neither guard can false-negative on a running node.  The exposure is a crash inside the write-to-externalize window.  Given the consequence (equivocation / distinct aggregate signature on a BFT consensus layer, undetectable because the certificate digest excludes the signature), this should be treated as a mainnet blocker and the fix opened as a follow-up PR the day #940 merges.

MavenRain:
Closes #963.  Follow-up to #940 (#934).

## Background

#940 closed the crash-loss equivocation window for two guard records by awaiting a
durable barrier before externalizing: `LastProposed` before the proposer broadcasts,
and `Votes` before the vote handler returns its vote.  Issue #963 inventories the
primary crate and finds two more records in the same hazard class.  Both are
`TableHint::Epoch`, so both route to the full-memory layered epoch DB whose `insert`
enqueues the durable disk write to a background thread and returns before it is on
disk.  A crash in the write-to-externalize window loses the record;  on restart the
guard reads nothing and an honest node can equivocate with no Byzantine actor.

#940 makes both gaps *easier* to reach, not harder: now that a restarted node
deterministically re-proposes the identical header, it reliably re-enters both
guards below on every restart.

## Changes

- **Site 1, certifier own-certificate (`ProposedCertificates`)**: add the #940
  durability barrier.  `spawn_header_proposal` inserts its own certificate and then
  calls `process_own_certificate` (already an externalization point: it forwards the
  certificate on the parents bus and triggers fetching) followed by
  `publish_certificate`, with no barrier in between.  This awaits
  `persist_durable::<ProposedCertificates>()` immediately after the insert and before
  `process_own_certificate`.  On crash-loss of the record, the guard at the top of the
  method would otherwise miss on restart and the certifier would re-propose the same
  header, re-collecting votes over an unordered channel and potentially aggregating a
  different `2f+1` subset into a distinct aggregate signature, which also perturbs the
  leader-signature randomness (`keccak256(leader.aggregated_signature())`).  This is a
  one-line change in #940's own pattern.

- **Site 2, vote-recast certificate-existence check**: fail closed.  The recast guard
  decided whether to sign a *second, different* vote for the same
  `(author, epoch, round)` by probing `read_by_index` for an existing certificate and
  re-voting when none was found.  That probe reads the same non-durable epoch tables
  (`Certificates` / `CertificateDigestByOrigin`), so a certificate that *was* formed
  from the earlier vote can be lost on restart and read back as absent, letting the
  node sign a distinct vote for a slot already aggregated into a live certificate: a
  directly distinct signed message.  The `.unwrap_or(None)` additionally failed a
  storage *error* open into the same permissive branch.  Since a `false`/errored read
  cannot prove the earlier vote was never certified, the guard now refuses (returns
  `AlreadyVoted`) whenever a different vote already exists for the slot.  The now-unused
  `read_by_index` probe is removed.

## The Site 2 decision (please weigh in)

Issue #963 flags Site 2 as a real design decision rather than a one-liner, with two
viable options:

1. **Fail closed** (this PR): treat an unknown or errored read as "cannot re-vote" and
   refuse.  Self-contained and cheap, with no barrier on the certificate hot path.  It
   costs a bounded, one-round liveness delay on exactly one recovery path: a proposer
   killed before quorum that restarts with a new header at the same round will not get
   this validator's vote and must advance a round.

2. **Durable certificate acceptance**: make certificate acceptance synchronously
   durable before the node can act as if it voted.  Larger, and it reintroduces a
   durability barrier on the certificate hot path (`write_all` in `cert_manager`).

I chose fail-closed to keep the fix contained to the primary crate and off the
certificate hot path, and because safety on a BFT layer outweighs a bounded,
single-path liveness cost.  Happy to switch to option 2 (or gate the strictness) if
you would rather preserve the permissive re-vote path.

## Severity

Both sites are crash-recovery-only: the epoch DB is `full_memory`, so the in-memory
map holds each record for the process lifetime and neither guard can false-negative on
a running node.  The exposure is a crash inside the write-to-externalize window.  The
consequence (equivocation or a distinct aggregate signature on a BFT consensus layer,
undetectable because the certificate digest excludes the signature) is why the issue
tracks these as mainnet blockers.

## Verification

- `cargo fmt` (nightly-2026-03-20) clean on the changed crate.
- `cargo check -p tn-primary` (1.94) clean.
- `cargo clippy` clean on the changed crate.
- New regression test `test_vote_recast_different_digest_fails_closed` (in
  `network_tests.rs`).  It simulates crash recovery: a durable `Votes` row for the author
  at the current `(epoch, round)` survives while the in-memory `auth_last_vote` guard is
  empty (fresh handler), so `vote_inner`'s storage-backed recast guard becomes load-bearing.
  With no certificate in storage, the recast of a different-digest header now returns
  `AlreadyVoted` instead of signing a second vote.  (On a running node the earlier
  in-memory guard catches this first;  only after a restart does Site 2 carry the weight,
  which is exactly the window #940 makes deterministic.)
- All `test_vote*` tests pass (18/18), including the existing equivocation tests
  `test_vote_different_digest_same_round_rejected` and `test_vote_older_round_rejected`.

## Scope note

One nearby write is deliberately left alone: the inactive-CVV catch-up cache in the
handler writes a peer certificate to the same tables but externalizes nothing and does
not vote, so a lost cache entry causes only a re-fetch, not a distinct signed message.
It is not a hazard, and its `CertificateStore::write` call keeps that import in use.

## Related observations (out of scope, flagged for a follow-up)

- Site 1's barrier trades the re-proposal equivocation window for a narrower, recoverable
  one: a crash between `persist_durable` and `process_own_certificate` leaves
  `ProposedCertificates` durable while the certificate was never processed into the local
  DAG, so on restart the guard re-publishes the cert on gossip but does not re-run
  `process_own_certificate`.  Nothing was externalized before the crash (the barrier
  precedes the first externalization), and the outcome is a local liveness edge the node
  recovers from by re-ingesting its own re-published cert, which is strictly preferable to
  a distinct aggregate signature.  The barrier must precede `process_own_certificate`
  because that call is itself an externalization point, so this trade is inherent.
- Site 1's *guard read* itself (`if let Ok(Some(cert)) = get::<ProposedCertificates>(..)`)
  still fails open on a storage read *error*, the same error-open shape Site 2 fixes.  On
  the full-memory epoch map a read error is not a realistic path, and hardening it is
  outside this issue's stated Site 1 scope (add the barrier), so it is left as is.  Happy
  to fail that read closed in a follow-up if you would like the symmetry.
- `persist_durable` acks a barrier even when the underlying epoch-DB write or commit
  errored and was logged-and-dropped in the background writer.  That is a property of
  #940's merged barrier primitive (used identically by the `LastProposed` / `Votes`
  barriers), not something this PR changes;  worth a separate look if best-effort durability
  on a disk error is a concern.

### DIFF
--- a/crates/consensus/primary/src/certifier.rs
+++ b/crates/consensus/primary/src/certifier.rs
@@ -445,6 +445,23 @@ impl<DB: Database> Certifier<DB> {
                             return Err(TaskError::from_message(e.to_string()));
                         }
 
+                        // Release the proposal lock now that the guard record is written.
+                        //
+                        // The lock exists to make the already-certified check at the top of this
+                        // method atomic with the insert above, and that pair is now complete: the
+                        // insert updates the in-memory layer synchronously, so any later proposal
+                        // reaching that check already observes this certificate. Nothing below
+                        // needs the exclusion.
+                        //
+                        // Holding it across the barrier would be actively harmful, because
+                        // `persist_durable` is a *whole-DB* barrier - the table parameter is unused
+                        // and only selects which of the three DBs to target - so it defers while
+                        // any write txn on the epoch DB is open and drains only at refcount zero,
+                        // with no timeout. Under catch-up or epoch close that is tens to hundreds
+                        // of milliseconds during which every other proposal would queue behind this
+                        // lock. Do not widen this critical section back out.
+                        drop(_guard);
+
                         // Wait for the `ProposedCertificates` record to be durable before
                         // externalizing. The epoch DB persists asynchronously, so the insert above
                         // returns before the record hits disk; `process_own_certificate` below already
--- a/crates/consensus/primary/src/certifier.rs
+++ b/crates/consensus/primary/src/certifier.rs
@@ -444,6 +444,18 @@ impl<DB: Database> Certifier<DB> {
                             error!(target: "primary::certifier", "error accepting own certificate, unable to save the certificate: {e}");
                             return Err(TaskError::from_message(e.to_string()));
                         }
+
+                        // Wait for the `ProposedCertificates` record to be durable before
+                        // externalizing. The epoch DB persists asynchronously, so the insert above
+                        // returns before the record hits disk; `process_own_certificate` below already
+                        // externalizes (it forwards the certificate on the parents bus and triggers
+                        // fetching), and the gossip publish follows. A crash in that window loses the
+                        // record, so on restart the guard at the top of this method misses and the
+                        // certifier re-proposes the same header, re-collecting votes over an unordered
+                        // channel, which can aggregate a different 2f+1 subset into a distinct aggregate
+                        // signature and thereby perturb the leader-signature randomness. See #934, #963.
+                        self.config.node_storage().persist_durable::<ProposedCertificates>().await;
+
                         // pass to state_sync for internal processing
                         if let Err(e) = self.state_sync.process_own_certificate(&mut certificate).await {
                             error!(target: "primary::certifier", "error accepting own certificate: {e}");
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -878,37 +878,28 @@ where
                         self.consensus_config.key_config(),
                     );
                     if vote.digest() != vote_info.vote_digest() {
-                        // Check if a certificate was already formed for this header author at this
-                        // round. If one exists, the old vote contributed to a real certificate, so
-                        // voting for a different header at the same round would be equivocation.
-                        // If no certificate exists, the old vote was never aggregated (e.g. the
-                        // proposer was killed before collecting enough votes, then restarted and
-                        // created a new header at the same round). In that case it is safe to
-                        // re-vote for the new header.
-                        let cert_exists = self
-                            .consensus_config
-                            .node_storage()
-                            .read_by_index(header.author(), header.round())
-                            .unwrap_or(None)
-                            .is_some();
-                        if cert_exists {
-                            warn!(
-                                "Authority {} submitted different header {:?} for voting",
-                                header.author(),
-                                header,
-                            );
-                            return Err(
-                                HeaderError::AlreadyVoted(header.digest(), header.round()).into()
-                            );
-                        }
-                        // No certificate was formed for the old vote — allow re-voting.
+                        // We have already signed a *different* vote for this author at
+                        // this (epoch, round); refuse to sign a second one.
+                        //
+                        // The previous behavior re-voted whenever `read_by_index` found no
+                        // certificate for the earlier vote. That probe is unsound across a
+                        // crash: the certificate tables are `TableHint::Epoch`, written
+                        // non-durably through the layered epoch DB, so a certificate that
+                        // *was* formed can be lost on restart and read back as absent,
+                        // letting the node sign a distinct vote for a slot already
+                        // aggregated into a live certificate. (`.unwrap_or(None)` also
+                        // failed a storage error open into the same branch.) A `false` or
+                        // errored read cannot prove the earlier vote was never certified,
+                        // so the only safe action is to refuse. See #934, #963.
                         warn!(
-                            "Authority {} re-proposing at round {} with a different header; \
-                         previous vote was for an uncertified header — allowing re-vote",
+                            "Authority {} submitted a different header {:?} for a round it was \
+                             already voted on; refusing to re-vote to avoid equivocation",
                             header.author(),
-                            header.round(),
+                            header,
                         );
-                        // Fall through to create and store the new vote below.
+                        return Err(
+                            HeaderError::AlreadyVoted(header.digest(), header.round()).into()
+                        );
                     } else {
                         debug!(
                             "Resending vote {vote:?} for {} at round {}",
--- a/crates/consensus/primary/src/primary.rs
+++ b/crates/consensus/primary/src/primary.rs
@@ -60,10 +60,26 @@ impl<DB: Database> Primary<DB> {
         leader_schedule: LeaderSchedule,
         task_manager: &TaskManager,
     ) {
-        // Probably don't need this for a non-committee member but it keeps channels drained and is
-        // not a problem.
-        self.state_sync.spawn(task_manager);
-
+        // Every `spawn` below subscribes to its channels synchronously and only then starts its
+        // task, so the order of these calls decides which channels are live before the first
+        // message is sent. `QueChannel::send` silently DROPS when nothing is subscribed (it returns
+        // `Ok(())` without enqueuing), so a sender that starts before its consumer subscribes loses
+        // messages with no error anywhere. Two such dependencies exist here:
+        //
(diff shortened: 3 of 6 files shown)


sstanfield (dismissed):

grantkee (approved):
LGTM! Thanks @MavenRain - good work

PR #965, opened 2026-07-19
test(storage): open the epoch data file, not the directory, in test_consensus_pack

### LINKED ISSUE #955: `test_consensus_pack` panics with `EINVAL` on tmpfs `/tmp`: it seeks a directory instead of the pack data file
## Summary

`storage::consensus_pack::test::test_consensus_pack` fails on any host whose system temp directory is **tmpfs** (the systemd >= 256 default for `/tmp`, e.g. Ubuntu 25.10, Fedora, Arch).  Two length-comparison opens in the test target the `epoch-0` **directory** instead of the `epoch-0/data` pack file, then call `seek(SeekFrom::End(0))` on it.  `lseek(SEEK_END)` on a directory fd is rejected with `EINVAL` by the libfs `dcache_dir_lseek` that tmpfs uses, whereas ext4/overlayfs (current CI runners, most dev machines, the repo's `rust:1.94-slim-bookworm` builder image) accept it.  The result is a hard `cargo nextest` failure that also blocks the local `make attest` flow.

## Panic

```
thread 'consensus_pack::test::test_consensus_pack' panicked at crates/storage/src/consensus_pack.rs:2171:43:
called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
```

## Location

`crates/storage/src/consensus_pack.rs` (origin/main `1adb486d`), inside `test_consensus_pack`, lines 2168-2173:

```rust
let mut f1 = File::open(temp_dir.path().join("epoch-0")).expect("log file");   // opens the DIRECTORY
let mut f2 = File::open(temp_dir2.path().join("epoch-0")).expect("log file");  // opens the DIRECTORY
assert_eq!(
    f1.seek(SeekFrom::End(0)).unwrap(),   // 2171:43 -- lseek(SEEK_END) on a directory fd
    f2.seek(SeekFrom::End(0)).unwrap(),
    "files not the same length"
);
```

## Root cause

Each epoch is laid out as a directory `epoch-{n}/` containing the data file `data`:

- `const DATA_NAME: &str = "data";` (`consensus_pack.rs:603`)
- `let base_dir = path.as_ref().join(format!("epoch-{epoch}"));` then `create_dir_synced(&base_dir);` (e.g. `725-726`), with the pack written to `base_dir.join(Self::DATA_NAME)`.

So `epoch-0` is a directory and the pack file is `epoch-0/data`.  Every other open in the test appends `.join(Inner::DATA_NAME)` to reach the data file;  the `f1`/`f2` opens above are the only two that stop at the directory.  A grep across the whole workspace confirms these are the **only** two `join("epoch-N")` uses that omit the data file:

```
crates/storage/src/consensus_pack.rs:2168:  let mut f1 = File::open(temp_dir.path().join("epoch-0")).expect("log file");
crates/storage/src/consensus_pack.rs:2169:  let mut f2 = File::open(temp_dir2.path().join("epoch-0")).expect("log file");
```

`File::open()` on a directory succeeds on Linux, so `.expect("log file")` passes.  `seek(SeekFrom::End(0))` then issues `lseek(fd, 0, SEEK_END)` on the directory fd, and the outcome is filesystem dependent:

- **tmpfs / ramfs / other libfs-backed directories** use `dcache_dir_lseek`, which returns `-EINVAL` for any `whence` other than `SEEK_SET`/`SEEK_CUR`.  `SEEK_END` -> `EINVAL` (errno 22).
- **ext4 / overlayfs** use `ext4_dir_llseek`, which accepts `SEEK_END` and returns the directory's size, so the seek succeeds and the assertion (accidentally) holds.

## Why CI passes but newer distros fail

`TempDir::with_prefix` creates its directory under `env::temp_dir()`, i.e. `/tmp`.  CI runners, the `rust:1.94-slim-bookworm` builder image, and most dev machines have a **disk-backed `/tmp`** (ext4/overlay), so `SEEK_END` on a directory succeeds and the test is green.  systemd >= 256 mounts `/tmp` as **tmpfs by default**, so on Ubuntu 25.10 / Fedora / Arch the same code path hits `dcache_dir_lseek` and panics.  Note the assertion is also semantically wrong even where it passes: it compares directory seek results, not the two pack files' lengths.

## Reproduction

```bash
# Any Linux host, no special distro needed: /dev/shm is tmpfs.
TMPDIR=/dev/shm cargo nextest run -p tn-storage consensus_pack::test::test_consensus_pack

# Or natively on a systemd >= 256 host where /tmp is already tmpfs (confirm with: findmnt /tmp):
cargo nextest run -p tn-storage consensus_pack::test::test_consensus_pack
```

## Impact

- `cargo nextest run` (and therefore `make attest`) fails on tmpfs-`/tmp` hosts, blocking local CI and on-chain attestation on current distros.
- No production impact: the defect is confined to test code, and the assertion does not verify what it intends even on ext4.

## Suggested fix

Open the actual `epoch-0/data` file on both sides, matching every other open in the test:

```diff
-            let mut f1 = File::open(temp_dir.path().join("epoch-0")).expect("log file");
-            let mut f2 = File::open(temp_dir2.path().join("epoch-0")).expect("log file");
+            let mut f1 = File::open(temp_dir.path().join("epoch-0").join(Inner::DATA_NAME))
+                .expect("log file");
+            let mut f2 = File::open(temp_dir2.path().join("epoch-0").join(Inner::DATA_NAME))
+                .expect("log file");
```

This makes `"files not the same length"` compare the two `data` files' lengths (the intent) and removes the directory seek entirely, fixing portability across filesystems.

## Workaround (until fixed)

Point the temp directory at a disk-backed path before running tests or `make attest`:

```bash
mkdir -p "$HOME/attest-tmp"
TMPDIR="$HOME/attest-tmp" make attest
```

MavenRain:
Closes #955.

## Problem

`storage::consensus_pack::test::test_consensus_pack` fails with `EINVAL` on any host
whose `/tmp` is **tmpfs** (the systemd >= 256 default: Ubuntu 25.10, Fedora, Arch). Two
length-comparison opens in the test target the `epoch-0` **directory** instead of the
`epoch-0/data` pack file, then call `seek(SeekFrom::End(0))` on it. `lseek(SEEK_END)` on a
directory fd is rejected with `EINVAL` by the libfs `dcache_dir_lseek` that tmpfs uses,
while ext4/overlayfs (CI runners, the `rust:1.94-slim-bookworm` builder image, most dev
machines) accept it. The failure hard-stops `cargo nextest` and therefore `make attest` on
current distros.

## Root cause

Each epoch is laid out as a directory `epoch-{n}/` holding the data file `data`
(`Inner::DATA_NAME`). Every open in the test appends `.join(Inner::DATA_NAME)` to reach the
pack file except these two, which stop at the directory. `File::open()` on a directory
succeeds on Linux, so `.expect("log file")` passes; the subsequent `seek(SeekFrom::End(0))`
is what diverges by filesystem. The assertion is also semantically wrong even where it
passes, since it compares two directory seek results rather than the two pack files'
lengths.

## Fix

Append `.join(Inner::DATA_NAME)` to both opens, matching every other open in the test.
`"files not the same length"` now compares the two `data` files' lengths (the intent) and
no directory seek remains, so the test is portable across filesystems.

Test-only change; no production code is touched.

## Validation

- `cargo +1.94 test -p tn-storage --lib consensus_pack::test::test_consensus_pack` passes
  under the repo-pinned toolchain.
- `cargo +nightly fmt --check` clean.

### DIFF
--- a/crates/storage/src/consensus_pack.rs
+++ b/crates/storage/src/consensus_pack.rs
@@ -2346,8 +2346,10 @@ pub(crate) mod test {
             assert!(pack.get_consensus_output(num_outputs as u64 * 2).await.is_ok());
             drop(pack);
 
-            let mut f1 = File::open(temp_dir.path().join("epoch-0")).expect("log file");
-            let mut f2 = File::open(temp_dir2.path().join("epoch-0")).expect("log file");
+            let mut f1 = File::open(temp_dir.path().join("epoch-0").join(Inner::DATA_NAME))
+                .expect("log file");
+            let mut f2 = File::open(temp_dir2.path().join("epoch-0").join(Inner::DATA_NAME))
+                .expect("log file");
             assert_eq!(
                 f1.seek(SeekFrom::End(0)).unwrap(),
                 f2.seek(SeekFrom::End(0)).unwrap(),


sstanfield (approved):
Good catch.

Issue #693, opened 2026-05-20
Public release pipeline with verifiable provenance


grantkee:
## Problem

There is no public, verifiable way to obtain a telcoin-network node binary.
The only documented distribution surface is `make docker-adiri`, which pushes to `us-docker.pkg.dev/telcoin-network/tn-public/adiri:latest` — a mutable tag, unsigned, with no provenance and no link back to a specific commit.
The README directs operators to `cargo build` from source.
Neither path produces something a third party can independently prove came from this repository.

For a BFT validator network, the binary an operator runs is a load-bearing trust assumption.
An operator who downloads a tarball or pulls an image today has no way to answer two basic questions: was this artifact built by CI from a tagged commit on this repo, and did the project maintainers approve the specific commit it was built from.
Without answers, supply-chain attacks against the build environment, the container registry, or an individual maintainer's account each become single points of compromise.

A second concern is internal: there is no mechanical guard against a single maintainer with credentials shipping a release on their own.
Tag protection prevents force-pushes, but a tagged commit on `main` plus push access to the release surface is sufficient for one person to ship a binary that pretends to be canonical.

## Solution

Introduce a public release channel under GitHub Releases plus a multi-arch image at `ghcr.io/telcoin-association/telcoin-network`, gated by a two-of-two maintainer signing requirement.

The pipeline should have three layers:

1. **CI build provenance.** A tag push (`v*.*.*` for mainnet, `v*.*.*-adiri` for testnet) triggers a GitHub Actions workflow that builds linux/amd64 and linux/arm64 binaries, packages tarballs with `SHA256SUMS`, and pushes a multi-arch image to `ghcr.io`. Every tarball and the image manifest carry SLSA build attestations via `actions/attest-build-provenance@v2`, which binds the artifact to the workflow file and the tagged commit through OIDC. Anyone can verify this layer with `gh attestation verify` against Sigstore — no key distribution required.

2. **Two-of-two maintainer countersignatures.** CI produces a *draft* release, never a published one. Each maintainer countersigns from their laptop using a YubiKey (PIV slot 9c, ECCP256, touch-policy=always) via cosign over PKCS#11. Each maintainer's public certificate is committed to `.github/release-keys/<handle>.pem` and pinned at the tag, so the trust anchor for a given release is the repo state at that commit, not any external key server. A maintainer signature applies to every  tarball, the `SHA256SUMS` file, and the image manifest digest.

[...]

PR #966, opened 2026-07-20
Makefile devnet build

grantkee:
- harden docker build for devnet
- restore faucet feature
- pas git sha for logging on node startup

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,6 +151,11 @@ docker-login:
 docker-adiri:
 	docker buildx build -f ./etc/Dockerfile --build-arg CARGO_FEATURES=adiri --platform linux/amd64,linux/arm64 --no-cache -t us-docker.pkg.dev/telcoin-network/tn-public/adiri:$(TAG) . --push ;
 
+# build and push latest devnet image for amd64 and arm64
+# CARGO_FEATURES=faucet compiles in the faucet support only
+docker-devnet:
+	docker buildx build -f ./etc/Dockerfile --build-arg CARGO_FEATURES=faucet --platform linux/amd64,linux/arm64 --no-cache -t us-docker.pkg.dev/telcoin-network/tn-public/adiri:$(TAG) . --push ;
+
 # push local adiri:latest to the gcloud artifact registry
 docker-push:
 	docker push us-docker.pkg.dev/telcoin-network/tn-public/adiri:$(TAG) ;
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -149,7 +153,12 @@ docker-login:
 # build and push latest adiri image for amd64 and arm64
 # CARGO_FEATURES=adiri compiles in the testnet fork logic (ConsensusRegistry fork, ADIRI_DUP_BATCH_EPOCH).
 docker-adiri:
-	docker buildx build -f ./etc/Dockerfile --build-arg CARGO_FEATURES=adiri --platform linux/amd64,linux/arm64 --no-cache -t us-docker.pkg.dev/telcoin-network/tn-public/adiri:$(TAG) . --push ;
+	docker buildx build -f ./etc/Dockerfile --build-arg CARGO_FEATURES=adiri --build-arg GIT_SHA=$(GIT_SHA) --platform linux/amd64,linux/arm64 --no-cache -t us-docker.pkg.dev/telcoin-network/tn-public/adiri:$(TAG) . --push ;
+
+# build and push latest devnet image for amd64 and arm64
+# CARGO_FEATURES=faucet compiles in the faucet support only
+docker-devnet:
+	docker buildx build -f ./etc/Dockerfile --build-arg CARGO_FEATURES=faucet --build-arg GIT_SHA=$(GIT_SHA) --platform linux/amd64,linux/arm64 --no-cache -t us-docker.pkg.dev/telcoin-network/tn-public/adiri:$(TAG) . --push ;
```
MavenRain, line 161: Need a double-check on my understanding here

On an equal `TAG`, `make docker-devnet` will overwrite the adiri testnet image with a faucet-without-adiri build . . . that is to say that it would drop the `ConsensusRegistry` fork logic and add the unrestricted mint precompile.

Should it, instead, be

```
-t us-docker.pkg.dev/telcoin-network/tn-public/devnet:$(TAG)
```
?

MavenRain (changes requested):
See Makefile comment

Makefile
grantkee: I agree that is a potential footgun, but this is a lot easier to manage right now from a single bucket because of the deployment scripts. I'm the only one publishing for now and append `*-adiri` and `*-devnet` to prevent collisions. The plan is to refactor devops pipeline with #693

MavenRain (approved):
Concerns addressed

PR #967, opened 2026-07-20
Add a pack file to store execution state at a specific block (the state tree).

sstanfield:
This adds the file format to store and recreate a specific reth execution state. This will allow sync to not have to start from genesis.

This is ONLY the on disk format.


crates/storage/src/archive/pack_iter.rs
```diff
@@ -49,9 +47,6 @@ where
     /// are returned in insert order.
     pub fn open(mut reader: R, uid_idx: u64) -> Result<Self, LoadHeaderError> {
         let header = DataHeader::load_header(&mut reader, uid_idx)?;
-        if header.version() > PACK_VERSION {
-            return Err(LoadHeaderError::InvalidVersion);
-        }
```
MavenRain, line 54: Assuming that this version-gate removal is intentional . . . wasn't immediately obvious from the diff where this gating may reappear . . . might be worth a comment somewhere as to why we're killing the gate here and the implied safety assurance

MavenRain (approved):
Just one comment on the version-gate removal

crates/storage/src/archive/pack_iter.rs
sstanfield: The issue is that pack_iter is generic over any pack.  The version belongs with the high level pack so enforcing it at a low level generic was always a bug I think.

grantkee (approved):

PR #969, opened 2026-07-20
Add reth code to export to or import from new exec state pack.

sstanfield:


crates/tn-reth/src/snapshot.rs
```diff
@@ -0,0 +1,1121 @@
...
+    /// [`import_chain_scaffold`](Self::import_chain_scaffold)). No JSONL, no ETL.
+    pub fn import_state(&self, reader: &mut ExecStatePackReader) -> eyre::Result<B256> {
+        let expected_root = reader.meta().state_root;
+        let b = reader.meta().block_number;
+
+        // pull the full account set out of the pack (address + genesis-shaped account)
+        let accounts: Vec<(Address, GenesisAccount)> = reader
+            .accounts()
+            .map(|res| res.map(|a| (a.address, a.account)))
+            .collect::<Result<_, _>>()?;
```
MavenRain, line 510: Unbounded in-memory hydration of an untrusted pack (no validation/pre-processing/size cap) . . . pretty spooky, no?

crates/tn-reth/src/snapshot.rs
```diff
@@ -0,0 +1,1121 @@
...
+/// True when `header` is a genuine worker batch block.
+///
+/// Mirrors the canonical `is_worker_batch_block` in `tn_node::manager::node`: excludes genesis
+/// (`number == 0`) and the synthetic empty-close block, whose `ommers_hash` (the batch-digest slot)
+/// is zero. Only genuine batch blocks carry a non-zero batch digest. Public so the snapshot
+/// uploader's fee-derivability precheck shares the exact attribution rule the restore-side
+/// [`SnapshotRestorer::derive_fee_precondition`] applies.
+pub fn is_worker_batch_block(header: &SealedHeader) -> bool {
+    header.number != 0 && header.ommers_hash != B256::ZERO
+}
```
MavenRain, line 666: `worker_id_from_header` and `is_worker_batch_block` are also in `tn_node::manager::node`.  It may be worth hoist the shared implementation into a single place to avoid them getting out-of-sync.

MavenRain (approved):
Two suggestions

grantkee (approved):

crates/tn-reth/src/snapshot.rs
sstanfield: Final PR goes into great depth on this and other large memory issues so this will be addressed before final landing.

crates/tn-reth/src/snapshot.rs
sstanfield: I noted this for a future PR, it seems like a good idea.  Will probably have to add to tn-types to avoid dep issues.

PR #970, opened 2026-07-21
feat(network): delete frozen legacy request-response variants, bump req-res to /0.0.2 (739)

MavenRain:
Final PR of the hybrid-streams migration (#739).  Steps 1 through 9c consolidated every
bulk request-response path onto the typed sync-stream protocol and froze the legacy
`/tn-stream` path.  Along the way, several request-response enum variants were kept on the
wire as dead-but-positional entries, each documented as "retained for wire compatibility;
removed at the coordinated `/0.0.2` bump".  This change performs that coordinated bump and
deletes them, closing the epic.

### What changed

Bumped the request-response wire protocol from `/0.0.1` to `/0.0.2`
(`/tn-primary-{chain}` and `/tn-worker-{id}-{chain}`) and deleted the eight frozen
variants plus their now-dead answer-with-error handlers:

- `PrimaryRequest`: `StreamEpoch`, `StreamEpochPartial`, `StreamConsensusOutput`,
  `MissingCertificates` (all four were answered with a typed error since their cutovers;
  handlers `process_epoch_stream`, `process_consensus_output_stream`, and
  `process_request_for_missing_certs` are removed).
- `PrimaryResponse`: `StreamRequestAck` and `RequestedCertificates` (neither had a
  producer left; `RequestedCertificates` was the reply paired with the deleted req-res
  `MissingCertificates`).
- `WorkerRequest` / `WorkerResponse`: `RequestBatchesStream` (handler
  `process_request_batches_stream` removed).

The surviving request-response variants are the live ones: primary `Vote`, `PeerExchange`,
`EpochRecord`;  primary responses `Vote`, `MissingParents`, `EpochRecord`, `PeerExchange`,
`Error`, `RecoverableError`;  worker `ReportBatch`, `PeerExchange`;  worker responses
`ReportBatch`, `PeerExchange`, `Error`, `RecoverableError`.

Also rewrote `crates/consensus/worker/src/network/README.md`, which still documented the
deleted legacy chunked-batch path (`send_batches_over_stream`, `PendingBatchStream`,
`LegacyStreamGuard`, the pending-request map and prune task).  It now describes the
sync-only batch path (request folded into the opening `SyncFrame::Req`, `try_admit_sync` /
`SyncStreamPermit` admission, `send_sync_batches_over_stream` serving `Ack` + `Data`* +
`End`).  This was the confirmed follow-up from the #939 review.

Also drops two now-dangling rows in `crates/network-libp2p/src/peers/penalty.md` that
documented the deleted `process_request_batches_stream` and `process_epoch_stream`
handlers (an adversarial review of this diff surfaced them;  penalty.md is not compiled or
`include_str!`d, so there is no build impact).  Other pre-existing line-number drift in
that file is left untouched.

### Why the version bump is mandatory and atomic with the deletion

Deleting a variant changes BCS discriminants.  In the worker enums the deletion is
mid-enum: removing `RequestBatchesStream` (discriminant 1) shifts `PeerExchange` (and, on
the response, `Error` / `RecoverableError`) down by one.  In the primary enums
`MissingCertificates` (discriminant 1) shifts the surviving variants too.  A `/0.0.1` peer
that still encodes one of these would be misdecoded by a node with the new layout.

Bumping the request-response protocol to `/0.0.2` in the same change fences this off: a
`/0.0.2` node never negotiates request-response with a not-yet-upgraded `/0.0.1` peer, so
the two never exchange a message under mismatched discriminants.  Only the request-response
protocol changed;  kad, sync, and peer-exchange stay at `/0.0.1` (their enums are
unchanged, and bumping them would needlessly fracture those subsystems).

### Rollout gate (important)

This is the "telemetry + epoch-gated" bump the migration deferred.  Because a `/0.0.2` node
stops speaking request-response to `/0.0.1` peers, it should land only once telemetry
confirms the legacy request-response variants see no live traffic and the rollout is
coordinated across the committee (an epoch boundary), exactly as the retained-variant
comments intended.  The code change is safe and self-contained;  the *deployment* is the
gated part.

### Scope notes / deliberately out of scope

- The `MissingCertificatesRequest` struct and its `max_response_size` field are kept: the
  struct is still used by the sync fetch path (`cert_collector` rebuilds it to reuse
  `get_bounds`).  Only the request-response variant that carried it on the wire is deleted;
  the field's comment is updated.  Removing the now-vestigial field is a possible follow-up.
- The embedded `PeerExchange` goodbye fallback (`send_legacy_goodbye` and the `PeerExchange`
  variants embedded in the request enums) is left in place.  That is a separate feature's
  deferred cleanup that was parked at the same `/0.0.2` coordination point;  this bump now
  unblocks it, but it touches the peer-manager goodbye path and belongs in its own PR.

### Gates

Toolchains: build/test stable 1.94;  fmt/clippy nightly-2026-03-20.  OOM-safe `-j 2`.

- check (`tn-network-libp2p`, `tn-worker`, `tn-primary`): clean.
- fmt `--check`: clean.
- clippy (`-D warnings`): `tn-primary --all-features --lib` clean.  `tn-network-libp2p` +
  `tn-worker --all-features --all-targets` adds no new warnings from this diff;  the only
  lint is a pre-existing test-only `collapsible_match` in `network_tests.rs`, byte-identical
  on origin/main (the same pre-existing test lint the earlier #739 PRs noted).  `tn-primary`
  uses `--lib` because `--all-targets` also hits a pre-existing `never_loop` in
  `certificate_fetcher_tests.rs`, not in this diff.
- tests: `tn-network-libp2p` 227, `tn-worker` 15 + 24, `tn-primary --lib` 158 (1 ignored).
  All pass.
- doc (`-D warnings`): `tn-network-libp2p` clean.  `tn-primary` / `tn-worker` have
  pre-existing private-intra-doc-link warnings on unchanged lines (there is no cargo-doc CI
  gate);  this diff adds zero new doc warnings.

No consumer-crate edits: the enums are internal to the consensus network modules.  The test
fixtures (`TestWorkerRequest` etc.) are independent test-only types decoded by their own
codec on both ends, so they are unaffected by the real-enum layout change.

### DIFF
--- a/crates/consensus/primary/src/network/message.rs
+++ b/crates/consensus/primary/src/network/message.rs
@@ -56,11 +56,6 @@ pub enum PrimaryRequest {
         /// missing them. The peer requires parent certs in order to vote.
         parents: Vec<Certificate>,
     },
-    /// Request for missing certificates.
-    MissingCertificates {
-        /// Inner type with specific helper methods for requesting missing certificates.
-        inner: MissingCertificatesRequest,
-    },
     /// Exchange peer information.
     ///
     /// This "request" is sent to peers when this node disconnects
@@ -77,35 +72,6 @@ pub enum PrimaryRequest {
         epoch: Option<Epoch>,
         /// Block hash requesting if not None.
         hash: Option<EpochDigest>,
-    },
-    /// Request to stream a pack file of the consensus data for epoch.
-    StreamEpoch {
-        /// The epoch we are requesting consensus data for.
-        epoch: Epoch,
-    },
-    /// Request to stream the raw (serialized) consensus output bytes for a consensus chain number.
-    ///
-    /// Unlike [`Self::ConsensusHeader`], this returns the full pack-file encoded output
-    /// (batches + consensus header) rather than just the header, so a peer can reconstruct
-    /// the [`tn_types::ConsensusOutput`] without separately fetching its batches.
-    ///
-    /// The output is streamed (rather than returned via request/response) because a single
-    /// output can exceed the request/response codec's message size limit.
-    StreamConsensusOutput {
-        /// The consensus chain number being requested.
-        number: u64,
-    },
-    /// Request to stream a verifiable PREFIX of an epoch's pack file, stopping after the consensus
-    /// output with `last_consensus_number`.
-    ///
-    /// Unlike [`Self::StreamEpoch`] (which streams a complete epoch), this streams the in-progress
-    /// current epoch up to an already-committed, verifiable point. The number is a chain consensus
-    /// header number, not a pack-relative index.
-    StreamEpochPartial {
-        /// The epoch we are requesting consensus data for.
-        epoch: Epoch,
-        /// The final (inclusive) consensus header number to stream up to.
-        last_consensus_number: u64,
     },
 }
 
@@ -289,8 +255,6 @@ impl From<PeerExchangeMap> for PrimaryRequest {
 pub enum PrimaryResponse {
     /// The peer's vote if the peer considered the proposed header valid.
     Vote(Vote),
-    /// The requested certificates requested by a peer.
-    RequestedCertificates(Vec<Certificate>),
     /// Missing certificates in order to vote.
     ///
     /// If the peer was unable to verify parents for a proposed header, they respond requesting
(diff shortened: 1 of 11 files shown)


sstanfield (approved):

grantkee (approved):
LGTM!

PR #973, opened 2026-07-21
Add a cli option and hook up the code to export exec state at the end of every epoch when enabled.

sstanfield:
Hooks up the ability to produce pack files of ending epoch state.

### DIFF
--- a/crates/node/src/manager/export_exec.rs
+++ b/crates/node/src/manager/export_exec.rs
@@ -1,0 +1,288 @@
+//! On-demand EVM execution-state snapshot export.
+//!
+//! [`ExecStateExporter`] runs state exports on a dedicated background thread so a (potentially
+//! slow) full plain-state walk never blocks the caller. An export is triggered by calling
+//! [`trigger_export`](ExecStateExporter::trigger_export), which enqueues a request and returns a
+//! receiver the caller may await for the result — or drop for fire-and-forget.
+//!
+//! The caller names the recent block to snapshot. Because the pinned plain-state reflects the
+//! persisted execution tip, the export succeeds only once the requested block IS that tip: each
+//! attempt pins a read-consistent view ([`RethEnv::pin_state_view`]) and confirms it matches the
+//! requested block
+//! ([`PinnedStateView::verify_tip`](tn_reth::snapshot::PinnedStateView::verify_tip)),
+//! retrying briefly while a just-agreed block is still catching up to persistence. It then writes
+//! the accounts plus that block's header and recent ancestors into an exec state pack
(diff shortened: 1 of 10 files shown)


MavenRain:
From Claude: In a worker overshoot, `verify_once` stays false forever (more specifically, past `block_N`, so that the 8 retries are futile).  Therefore, under sustained load, most epoch packs will be dropped silently.  A fix involves detecting the overshoot (tip > block) and fail loud with a metric instead of just warning.  Also reconsider the single-thread + 800ms budget vs a multi-second state walk.

crates/node/src/manager/node/close_epoch.rs
```diff
@@ -252,6 +252,85 @@ where
...
+        if final_dir.exists() {
+            debug!(target: "tn::snapshot", epoch, "epoch state already exported; skipping");
+            return Ok(());
+        }
+        // Export into a temp sibling and atomically rename it into place on success, so external
+        // tooling only ever observes a complete `epoch-{N}` directory. Clear any leftover temp from
+        // a prior failed attempt first.
+        let tmp_dir = export_root.join(format!("epoch-{epoch}.tmp"));
+        if tmp_dir.exists() {
+            if let Err(e) = std::fs::remove_dir_all(&tmp_dir) {
```
MavenRain, line 286: Consider this being fatal instead of just warning to prevent a complete export resulting in a stale+fresh mix.

crates/node/src/manager/export_exec.rs
```diff
@@ -0,0 +1,293 @@
...
+
+impl ExecStateExporter {
+    /// Spawn the background export worker. Each export supplies its own [`RethEnv`] (the node's
+    /// execution engine — and its `RethEnv` — is recreated per epoch), so the worker holds none.
+    pub fn spawn() -> Self {
+        let (tx, rx) = mpsc::channel(REQUEST_QUEUE);
+        let worker = std::thread::Builder::new()
+            .name("evm-state-exporter".to_string())
+            .spawn(move || run_worker(rx))
+            .expect("failed to spawn evm-state-exporter thread");
```
MavenRain, line 94: Expect on the non-test path?

grantkee (approved):
LGTM - discussed hardening in follow up #985

crates/node/src/manager/node/close_epoch.rs
sstanfield: Since snapshots are a side quest I am leaving as is for now.  I did drop a comment to this effect (next PR) and we may want to revisit later.

crates/node/src/manager/export_exec.rs
sstanfield: Given this is during startup an error will be more or less the same, however I agree it will be cleaner to handle it like any other early startup error.  Will be in the next PR.

PR #977, opened 2026-07-22
refactor(storage): converge persist and persist_durable into one durable barrier (#962)

### LINKED ISSUE #962: converge `persist()` and `persist_durable()` into a single durability-honoring barrier
## Background

#940 added `persist_durable::<T>()` as a new provided `Database` trait method (a no-op default, like `persist`) rather than changing `persist()` itself.  The reasoning in the PR body was that `persist()` / `sync_persist()` cannot simply be made to wait for an absorbing open transaction, because a caller holding an open `write_txn` across the barrier would deadlock, and two existing tests rely on that overlap.

@sstanfield, who wrote `persist()`, reads its open-transaction bypass as a bug in `persist()` rather than a property to design around, and would rather fix it than add a second barrier.  He is substantively right, and the PR body's deadlock rationale is weaker than it was stated to be.

## Why the deadlock argument is weaker than stated

There are **zero production callers** of the `Database` trait methods `persist` / `sync_persist`.  Every call site is under `#[cfg(test)]` or in `tests/it/`:

- `crates/storage/src/lib.rs` sites are all inside the `#[cfg(test)]` module (the module opens at `:216` and the first `sync_persist` call is well below it).
- `crates/network-libp2p/src/kad.rs` sites are all below the `#[cfg(test)]` at `:570`.
- `crates/storage/src/stores/composite_db.rs:472` is below the `#[cfg(test)]` at `:268`.
- `crates/consensus/primary/tests/it/` sites are integration tests.

So converging the API has no production blast radius.  The only things that would hang under wait-for-txn-close semantics are a handful of test call sites, each a small rewrite.

### Disambiguation (important before anyone greps)

There are six different `fn persist` in this tree.  Only the `Database` trait method is in scope here:

- `crates/types/src/database_traits.rs:135` (`persist`) and `:154` (`sync_persist`), with impls in `crates/storage/src/stores/composite_db.rs` and `layered_db.rs`.  **This is the one to converge.**

The others are unrelated same-named wrappers and **do** have production callers, so a naive "no production callers of persist" claim is false and will look careless:

- `EpochRecords::persist` (`crates/storage/src/stores/epoch_records.rs:341`), called at `crates/state-sync/src/epoch.rs:151`, `crates/node/src/manager/epoch_votes.rs:33`.
- `ConsensusPack::persist` (`consensus_pack.rs:527`), `CertificatePack::persist` (`certificate_pack.rs:188`), the latest-consensus channel `persist` (`consensus.rs:261`), each with their own callers.

## Suggested change

1. Make `persist()` defer its ack until an absorbing open transaction commits (i.e. give it the `DurableBarrier` semantics #940 built), or delete it in favor of `persist_durable`.  Land on **one** barrier, not two.
2. Rewrite the small number of test call sites that intentionally hold an open txn across the barrier.  In `layered_db.rs` the affected sites can substitute the `db.stats()` flush that already sits on the following line;  the PR's own old-semantics pin test documents the pre-fix behavior and would be updated or removed.
3. Decide `sync_persist`'s fate deliberately: its spin loop must never defer across a *same-thread* open txn, or it self-deadlocks.  Either keep it as the synchronous-only variant or remove it alongside `persist`.

## Sequencing

This is a mechanical API cleanup with no production blast radius, but it should follow the table-placement decision.  If the anti-equivocation tables move off the layered DB to synchronously-durable storage, several of the places that need any barrier at all disappear, and `persist_durable` may not need to exist under any name.  Settling the storage question first avoids converging an API that a later change deletes.

MavenRain:
Closes #962.

## Summary

#940 made the proposer's `LastProposed` header and the vote handler's `Votes` record durable before they are externalized, closing an anti-equivocation window (#934).  It did so by adding a *second* provided `Database` barrier, `persist_durable::<T>()`, rather than changing `persist()`, on the stated grounds that making `persist()` wait for an absorbing open transaction would deadlock callers that hold a `write_txn` across the barrier.

This PR converges the two barriers into one.  `persist()` gains the durable, defer-until-the-absorbing-txn-commits semantics that `persist_durable()` had, and `persist_durable()` is deleted.  `sync_persist()` is kept as the deliberately non-durable synchronous sibling.

## Why the second barrier was avoidable

The deadlock argument only applies to callers that hold an open `write_txn` across the barrier, and there are no production callers of the `Database` trait's `persist` / `sync_persist` at all.  Every call site is under `#[cfg(test)]` or in `tests/it/`:

- The `crates/storage/src/lib.rs`, `crates/storage/src/composite_db.rs`, and `crates/network-libp2p/src/kad.rs` sites are all below their crate's `#[cfg(test)]`.
- The `crates/consensus/primary/tests/it/` sites are integration tests.

The only production callers of the barrier family are the two `persist_durable` sites #940 added, and both want durability.  So giving `persist` the durable semantics has zero production blast radius, and the second method is redundant.

(Disambiguation: this is only about the `Database` trait method at `crates/types/src/database_traits.rs`.  The unrelated same-named wrappers `EpochRecords::persist`, `ConsensusPack::persist`, `CertificatePack::persist`, and the latest-consensus channel `persist` take no `Table` parameter, do have production callers, and are untouched.)

## Change

- [x] `crates/types/src/database_traits.rs`: `persist::<T>()` now documents the durability contract (resolves only once the write is durably committed, including the case where a concurrently open write txn has absorbed the caller's bare insert).  The `persist_durable` provided method is removed.  Synchronously-durable backends (raw MDBX / redb, `MemDatabase`) keep the no-op default, which is already correct for them.
- [x] `crates/storage/src/layered_db.rs`: the `LayeredDatabase::persist` impl now sends `DBMessage::DurableBarrier` (was `CaughtUp`);  the `persist_durable` impl is removed.  The background thread's defer-until-commit drain (`pending_durable`, drained in the `CommitTxn` / `EndTxn` arms) is unchanged from #940.
- [x] `crates/storage/src/composite_db.rs`: the `persist_durable` impl is removed;  `persist` already routes by `TableHint`, so a barrier still only waits on the relevant physical store.
- [x] `crates/consensus/primary/src/proposer.rs` and `crates/consensus/primary/src/network/handler.rs`: the two anti-equivocation sites call `persist::<T>()` again instead of `persist_durable::<T>()`.  No behavior change . . . `persist` is now what `persist_durable` was.

## sync_persist: kept, non-durable, on purpose

`sync_persist()` cannot adopt the deferring semantics.  It is synchronous and spins on the calling thread, so if that thread holds an open `write_txn` across it, deferring the ack until the txn commits would self-deadlock (the txn only commits after `sync_persist` returns).  It therefore stays on the immediate-ack `CaughtUp` path and is documented as a read-your-writes catch-up primitive, not a durability barrier, for tests that cannot `await`.  The two txn-count-balancing tests that hold an open txn across `sync_persist` (`test_layereddb_overlapped_txn_drop_keeps_count_balanced`, `test_layereddb_cloned_txn_commit_sends_exactly_one_end`) are unaffected precisely because `sync_persist` is left non-deferring.

## Restored invariant (unchanged from #940)

An honest node still never externalizes a header or vote whose anti-equivocation record is not yet durable.  After a crash at any point after the header is broadcast or the vote is returned, on restart the node either re-proposes the identical header (re-signs the identical vote) or does not act for that round;  it never produces a different one, so an ordinary crash (no attacker required) cannot manufacture authority equivocation.  The barrier backing that guarantee is now `persist`, not a separate method.

## Testing

Storage tests (`crates/storage/src/layered_db.rs`):

- The old-behavior pin `test_persist_acks_before_open_txn_flushes` (which asserted that a plain `persist` acked before an absorbed write was durable) is deleted . . . that window no longer exists.
- The two durability tests are renamed to target `persist`: `test_persist_waits_for_open_txn_commit` (the barrier must not resolve while the absorbing txn is open, then resolves once it commits with the record on disk) and `test_persist_acks_immediately_without_open_txn`.  `test_persist_waits_for_open_txn_commit` is non-vacuous: it fails if `persist` does not defer (the old `CaughtUp` `persist` resolved inside the 200ms window and tripped the `is_err()` assertion), which is the differential the deleted test previously pinned in the other direction.

Commands (worktree off `origin/main`, pinned toolchain):

- `cargo +1.94 check -p tn-types -p tn-storage -p tn-primary`: clean.
- `cargo +nightly-2026-03-20 fmt --check` on the three crates: clean.
- `cargo +1.94 clippy --all-targets` on the three crates: 0 warnings on the changed lines.
- `tn-storage` lib: 149 passed.  `tn-primary`: proposer 4, vote 30.  Integration `storage_tests` 12, `recovery_tests` 5 (every `.persist::<T>()` settle site runs after its writes commit, so it acks immediately and does not hang under the durable barrier).
- `make attest` proxy: `cargo +1.94 test --no-run --workspace` in an isolated target dir: green (30 test binaries built, 0 errors).
- The renamed `test_persist_waits_for_open_txn_commit` was confirmed non-vacuous by mutation: reverting `persist` to the non-deferring `CaughtUp` path makes it fail on the "durable barrier resolved while the absorbing txn was still open" assertion, then restoring the `DurableBarrier` path makes it pass again.

## Sequencing note

The issue flags that this cleanup should follow the table-placement decision: if the anti-equivocation tables move off the layered DB to synchronously-durable storage, several sites that need any barrier disappear.  This PR does not depend on that decision.  It removes the redundant second method without changing which tables live where, so it composes with either outcome (if the tables later move, the single `persist` degrades to the no-op default at those sites rather than leaving a second method to also delete).

### DIFF
--- a/crates/consensus/primary/src/network/handler.rs
+++ b/crates/consensus/primary/src/network/handler.rs
@@ -939,7 +939,7 @@ where
         // recast guard would then read nothing and could sign a *different* vote for the same
         // author/round while the pre-crash vote is already held by the peer: equivocation from an
         // ordinary crash. See issue #934.
-        self.consensus_config.node_storage().persist_durable::<Votes>().await;
+        self.consensus_config.node_storage().persist::<Votes>().await;
 
         Ok(PrimaryResponse::Vote(vote))
     }
--- a/crates/consensus/primary/src/proposer.rs
+++ b/crates/consensus/primary/src/proposer.rs
@@ -275,7 +275,7 @@ impl<DB: Database> Proposer<DB> {
         // anti-equivocation guard would then read nothing and build a *different* header for this
         // same round while the pre-crash header is already on the network: equivocation from an
         // ordinary crash. See issue #934.
-        proposer_store.persist_durable::<LastProposed>().await;
+        proposer_store.persist::<LastProposed>().await;
 
         // Send the new header to the `Certifier` that will broadcast and certify it.
         consensus_bus.headers().send(header.clone()).await.map_err(|e| Box::new(e).into())
--- a/crates/storage/Cargo.toml
+++ b/crates/storage/Cargo.toml
@@ -41,7 +41,7 @@ tn-reth = { workspace = true }
 tn-test-utils = { workspace = true }
 rand = { workspace = true }
 roaring = { workspace = true }
-# Async barrier tests for `persist_durable` need a runtime + timers.
+# Async barrier tests for the durable `persist` need a runtime + timers.
 tokio = { workspace = true, features = [ "macros", "rt", "time" ] }
 
 [features]
--- a/crates/storage/src/composite_db.rs
+++ b/crates/storage/src/composite_db.rs
@@ -130,10 +130,6 @@ impl<DB: Database> Database for CompositeDatabase<DB> {
         self.get_db(T::HINT).persist::<T>()
     }
 
-    fn persist_durable<T: Table>(&self) -> impl Future<Output = ()> + Send {
-        self.get_db(T::HINT).persist_durable::<T>()
-    }
-
     fn sync_persist(&self) {
         self.inner.epoch_db.sync_persist();
         self.inner.kad_db.sync_persist();
--- a/crates/storage/src/layered_db.rs
+++ b/crates/storage/src/layered_db.rs
@@ -490,19 +490,6 @@ impl<DB: Database> Database for LayeredDatabase<DB> {
         let (tx, rx) = oneshot::channel();
         let r = self
             .tx
-            .send(DBMessage::CaughtUp(tx))
-            .map_err(|_| eyre::eyre!("DB thread gone, FATAL!"));
-        async move {
-            if r.is_ok() {
-                let _ = rx.await;
-            }
-        }
-    }
-
-    fn persist_durable<T: Table>(&self) -> impl Future<Output = ()> + Send {
-        let (tx, rx) = oneshot::channel();
-        let r = self
-            .tx
             .send(DBMessage::DurableBarrier(tx))
             .map_err(|_| eyre::eyre!("DB thread gone, FATAL!"));
         async move {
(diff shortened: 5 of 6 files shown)


sstanfield (dismissed):
I like this since persist_durable basically fixed persist.

grantkee (approved):

PR #979, opened 2026-07-22
fix(consensus): fail-stop on epoch-DB commit error to prevent self-equivocation (#975)

### LINKED ISSUE #975: durability barrier acks a failed commit
## Summary

On a commit-time storage error (disk full, MDBX `MAP_FULL`, `EIO`, or a checksum failure), the anti-equivocation durability barrier still acknowledges as if the write succeeded, so the guard record (`LastProposed` / `Votes`) is silently lost while the header or vote is externalized to the network. After a restart the recast guard reads nothing and can produce a conflicting header/vote for the same `(author, round)`, i.e. self-inflicted equivocation with no Byzantine actor involved. Impact class: safety (durability). One-line trigger: any physical commit failure on the epoch DB while a barrier ack is pending. This is a hole in the barrier introduced (PR #940) to close issue #934's plain-crash window, so that remediation is incomplete.

## Defect

The `LayeredDatabase` background runner commits the shared physical write txn in `end_txn`, but a commit `Err` is only logged and then discarded. `crates/storage/src/layered_db.rs:160-164`:

```rust
if let Some((current_txn, count)) = txn.take() {
    if count <= 1 {
        if let Err(e) = current_txn.commit() {
            tracing::error!(target: "layered_db_runner", "DB TXN Commit: {e}")
        }
```

Because `txn.take()` already ran at `:160`, `txn` is left `None` whether or not the commit succeeded. The two arms that follow `end_txn` then drain the deferred durability acks purely on `txn.is_none()`, never on commit success. `crates/storage/src/layered_db.rs:205-210` (`CommitTxn`):

```rust
end_txn(&mut txn, &mut committed_inserts, mem_db.as_ref());
if txn.is_none() {
    pending_durable.drain(..).for_each(|tx| {
        let _ = tx.send(());
    });
}
```

The `EndTxn` arm's ack-drain is identical, `crates/storage/src/layered_db.rs:217-222`. The barrier itself, when a write txn is open, only defers the ack into `pending_durable`; it never records whether the eventual commit worked. `crates/storage/src/layered_db.rs:270-274`:

```rust
if txn.is_some() {
    pending_durable.push(tx);
} else {
    let _ = tx.send(());
}
```

The public barrier API cannot surface a failure: it awaits the oneshot and returns unit. `crates/storage/src/layered_db.rs:502-513`:

```rust
fn persist_durable<T: Table>(&self) -> impl Future<Output = ()> + Send {
    let r = self.tx.send(DBMessage::DurableBarrier(tx))...;
    async move {
        if r.is_ok() {
            let _ = rx.await;
        }
    }
}
```

Both consensus callers await this unit future and then externalize unconditionally. Proposer, `crates/consensus/primary/src/proposer.rs:268-281`:

```rust
proposer_store.write_last_proposed(header)...?;
// ... See issue #934.
proposer_store.persist_durable::<LastProposed>().await;
// Send the new header to the `Certifier` that will broadcast and certify it.
consensus_bus.headers().send(header.clone()).await...
```

Vote handler, `crates/consensus/primary/src/network/handler.rs:934-944`:

```rust
self.consensus_config.node_storage().write_vote(&vote)?;
// ... See issue #934.
self.consensus_config.node_storage().persist_durable::<Votes>().await;
Ok(PrimaryResponse::Vote(vote))
```

In both, `persist_durable().await` returns even when the underlying commit failed, so the header is broadcast to the `Certifier` and the vote is returned to the requesting peer while the corresponding guard record is not on disk.

## Reproduction

1. Build with the default epoch backend (MDBX; see Root cause). Configure the epoch DB with a small map size so it can be filled, or interpose a commit that returns `Err` (e.g. an mdbx `MAP_FULL`, an injected `EIO`, or a fault-injecting `Database` impl in a test build).
2. Drive the proposer to `store_and_send_header` for round `r`. `write_last_proposed` enqueues the insert into the open write txn; `persist_durable::<LastProposed>()` sends a `DurableBarrier` that is deferred into `pending_durable` because the txn is open (`layered_db.rs:270-271`).
3. Force the physical commit for that txn to fail. In `end_txn` the `current_txn.commit()` returns `Err`, which is only logged (`layered_db.rs:162-164`); `txn` becomes `None`.
4. Observe that `CommitTxn`/`EndTxn` then drains `pending_durable` and sends the ack (`layered_db.rs:206-209`), so `persist_durable().await` completes successfully even though nothing was persisted.
5. The proposer proceeds to broadcast the header for round `r` to the `Certifier` (`proposer.rs:281`). The header is now externalized; `LastProposed` on disk still holds the pre-`r` value (or is empty).
6. Restart the node. The anti-equivocation guard reads `LastProposed`, sees no record for round `r`, and permits building a different header for the same round while the pre-crash header is already on the network. The two conflicting artifacts for `(author, r)` are the self-inflicted equivocation. The same sequence reproduces via `handler.rs:934-944` for `Votes` returned to a peer.

## Root cause

The durability barrier acks on `txn == None` regardless of whether the physical commit succeeded (`layered_db.rs:160-173`, `:206-209`, `:218-221`), and the swallowed commit `Err` at `:162-164` never propagates. `persist_durable` returns `()` and structurally cannot report failure (`:502-513`), so callers treat a failed commit as a successful durability fence and externalize (`proposer.rs:278-281`, `handler.rs:942-944`).

Backend commit-error behavior (caveat resolved): the epoch tables are all `TableHint::Epoch` (`crates/storage/src/lib.rs:89-95`), routed to `epoch_db` (`crates/storage/src/composite_db.rs:41`). The default backend is MDBX; redb is only used when the `redb` feature overrides it (`crates/storage/src/lib.rs:111-115`, with `_open_mdbx` wiring `MdbxDatabase` for the `epoch` path at `:158`). Both concrete `commit()` impls return `eyre::Result` and propagate the underlying error via `?` rather than panicking: MDBX at `crates/storage/src/mdbx/database.rs:85-88` (`self.inner.commit()?`) and redb at `crates/storage/src/redb/database.rs:65-68` (`self.tx.commit()?`). Therefore, on `ENOSPC` / `MAP_FULL` / `EIO`, `commit()` returns `Err`, the runner takes the swallow-and-ack path, and the defect is the silent equivocation window (HIGH), not a fail-stop crash. There is no branch here where the backend panics and degrades the defect to fail-stop; both configured backends surface the error as a returned `Err` that is then dropped. (The `.expect(...)` calls in `_open_mdbx` are on DB open, not commit, so they do not change this.)

## Why not already covered

- Issue #934 (CLOSED) is the direct lineage: its fix (PR #940) added `persist_durable` before broadcast/return precisely to close the plain-crash async-persist window (see the in-code rationale at `proposer.rs:272-277` and `handler.rs:936-941`, both citing #934). This defect is a hole in that fix: the barrier it introduced still acks on a failed commit, leaving a residual commit-error window that the fix did not consider.
- Issues #961/#962/#963/#964 form the anti-equivocation durability cluster, but they operate one layer up and assume the barrier is sound: #961 asks whether the epoch tables should live on raw MDBX instead of the layered memory DB; #962 asks to converge `persist()` and `persist_durable()` into one barrier; #963/#964 add or relocate barrier call-sites for the certifier own-certificate and vote-recast guards. None of them change the fact that `end_txn` swallows the commit `Err` and the ack fires on `txn.is_none()`. Even after #961-#964 land, a failed commit still acks a durability barrier that did not persist.
- Not #415 (txpool) and not #959 (dropped-tx); those are unrelated subsystems.

## Suggested fix

- Make `end_txn` return the commit `Result` instead of logging and discarding it (`layered_db.rs:160-173`).
- In the `CommitTxn` / `EndTxn` arms, drain and ack `pending_durable` only on `Ok`; on `Err`, do NOT ack the pending barriers (`layered_db.rs:206-209`, `:218-221`). A failed barrier must not complete as success.
- Change `persist_durable` to return a `Result` (e.g. by having the runner send an ack payload that carries success/failure over the oneshot) so a failed commit is observable to callers (`layered_db.rs:502-513`).
- Update both callers to refuse to externalize on failure: the proposer must not `consensus_bus.headers().send(...)` and the handler must not return `PrimaryResponse::Vote(...)` if the barrier reports failure (`proposer.rs:278-281`, `handler.rs:942-944`).
- Because a commit failure on the epoch DB is unrecoverable for consensus safety, treat an epoch-DB commit `Err` as a fatal shutdown of the node rather than continuing. This both preserves the safety invariant and avoids the alternative failure mode where refusing to ack would leave `persist_durable().await` hanging forever; a clean fail-stop is preferable to a silent equivocation and to an indefinite await hang.

## Test design

- Commit-failure barrier test: build a `Database`/`TXMut` test impl whose `commit()` returns `Err` (fault injection). Open a write txn, enqueue an insert, issue `persist_durable`, then trigger the commit. Assert that the barrier does NOT complete as success (either the returned `Result` is `Err`, or the future does not resolve to a success ack), and assert that a simulated caller does not externalize (no header sent on the `headers` bus / no `PrimaryResponse::Vote` produced). This must be a confirm-by-mutation test: with the current code it should fail (the ack fires), proving the test is non-vacuous.
- Restart-guard test: with commit forced to fail for round `r`, run `store_and_send_header` (or `vote_for_header`), then reopen the epoch DB and invoke the anti-equivocation guard for the same `(author, r)`. Assert the node refuses to produce a second, conflicting artifact for that round (either because it fail-stopped before externalizing, or because the guard now sees a durable record). Include the positive path (commit succeeds -> barrier acks -> externalize allowed -> restart guard blocks a conflicting artifact) so the test distinguishes the fix from a blanket refusal.

## Acceptance criteria

- A commit `Err` in `end_txn` is propagated, not swallowed (`layered_db.rs:162-164`).
- `pending_durable` acks fire only when the physical commit succeeded; a failed commit never acks a `DurableBarrier` (`layered_db.rs:206-209`, `:218-221`).
- `persist_durable` can surface commit failure to its caller (`layered_db.rs:502-513`).
- Neither caller externalizes (broadcast header / return vote) when the barrier does not confirm durability (`proposer.rs:278-281`, `handler.rs:942-944`).
- An epoch-DB commit failure results in a clean fatal shutdown, not a silent continue and not an indefinite await hang.
- A confirm-by-mutation test injects a commit failure and asserts no false-durable ack and no externalized artifact; a restart-guard test asserts no conflicting `(author, round)` artifact is produced after the failure.

MavenRain:
## Summary

Fixes #975. On a commit-time storage error (disk full, MDBX `MAP_FULL`, `EIO`, or a checksum failure), the anti-equivocation durability barrier used to acknowledge as if the write had succeeded, so the guard record (`LastProposed` / `Votes`) was silently lost while the header or vote was externalized to the network. After a restart the recast guard read nothing and could produce a conflicting header/vote for the same `(author, round)`: self-inflicted equivocation with no Byzantine actor involved. This closes the residual commit-error hole left open by the plain-crash fix in #940 (issue #934).

The barrier is now honest about a failed commit, and, via a node-level poison latch, **every** externalization exit that depends on the epoch-DB guard fail-stops the node rather than externalizing a non-durable vote or header: the proposer broadcast, the first-vote reply, and both vote-recast replies.

## Problem

The `LayeredDatabase` background runner committed the shared physical write txn in `end_txn`, but a commit `Err` was only logged and then discarded, and the deferred durability acks fired purely on `txn.is_none()`, never on commit success:

- `end_txn` swallowed the commit error (`layered_db.rs`), leaving `txn == None` whether or not the commit succeeded.
- The `CommitTxn` / `EndTxn` arms drained `pending_durable` and acked on `txn.is_none()` alone.
- `persist_durable` awaited a unit oneshot and returned `()`, so it structurally could not report failure.
- The consensus callers awaited that unit future and then externalized unconditionally.

There are **four** externalization exits that depend on the epoch-DB guard being durable, and all four were unguarded:

1. The proposer broadcasts the header to the `Certifier` after `store_and_send_header`.
2. The vote handler returns a freshly created `PrimaryResponse::Vote` on the first-vote path.
3. The vote handler *recasts* a stored vote on the fast path, before taking the per-author lock (`handler.rs`, `vote`).
4. The vote handler *resends* a matching prior vote inside `vote_inner`.

Exits 3 and 4 do not write a fresh record; they read the existing guard from storage and return it. The epoch DB is opened with `full_memory = true`, so its mem layer is authoritative: `write_vote` lands the `Votes` record in memory synchronously and the recast reads it back even if the physical commit that was supposed to persist it later failed. A naive "await the barrier at the recast" does not close this, because with no write txn pending the barrier reflects only the *current* commit state and acks `Committed` immediately — it cannot see a commit that already failed.

Threat model: any physical commit failure on the epoch DB. No Byzantine actor is required; a single honest node with a full or failing disk can equivocate against itself on restart. Impact class: safety (durability).

## Root cause

The durability barrier acked on `txn == None` regardless of whether the physical commit succeeded, and the swallowed commit `Err` never propagated. Because `persist_durable` returned `()`, callers treated a failed commit as a successful durability fence and externalized. Both backends (`MdbxDatabase`, `redb`) surface `commit()` failures as a returned `Err` rather than a panic, so the failure mode is a silent equivocation window rather than a fail-stop crash. And because a failed commit can leave a guard record in the authoritative mem layer but never on disk, a later recast is a distinct externalization path that a per-call barrier alone cannot fence.

## Fix

- `end_txn` now returns the physical commit outcome (`Option<eyre::Result<()>>`) instead of logging and discarding it, and clears the cache-mode mirror only on a successful commit (on failure the values are not on disk, so the mem layer must keep serving them).
- The deferred durability acks now carry a `DurableAck` (`Committed` / `CommitFailed`) over the oneshot, driven by the real commit outcome. A failed commit never acks `Committed`, and a lost DB thread is surfaced as a failure rather than an indefinite hang.
- `persist_durable` now returns `impl Future<Output = eyre::Result<()>>` across the `Database` trait default, `CompositeDatabase`, and `LayeredDatabase`, so a failed commit is observable to callers.
- **Poison latch.** The `LayeredDatabase` background runner owns a sticky `commit_failed` flag, set the instant any physical commit fails. The flag is set at *every* physical-commit site the runner drives, not only the shared write-txn `commit`: a bare `insert` / `remove` / `clear` is, for the MDBX backend, itself a self-contained `write_txn` → op → `commit` (see `MdbxDatabase::insert`), and the guard writes `write_last_proposed` and `write_vote` take exactly this bare-insert path with no surrounding txn. Latching only the shared-txn commit would have left the real call sites unguarded. Both durability-ack paths consult the flag: a barrier deferred behind an open txn, and a barrier that arrives with no txn pending, both resolve to `CommitFailed` once it is set. The flag is a plain local, read and written only by the single runner thread in FIFO order with the commits themselves, so there is no cross-thread race and `persist_durable` never touches it directly . . . it simply maps the ack. This closes the past-failure case a per-call barrier misses: after the first failed commit every barrier fails, so every externalization exit fail-stops.
- The proposer refuses to broadcast the header on a barrier failure and returns a fatal `ProposerError::DurableBarrierFailed`; because the proposer runs as a critical task, this fail-stops the node cleanly.
- The vote handler now awaits the barrier at all three of its externalization exits: first vote, fast recast (pre-lock), and `vote_inner` resend . . . and refuses to return the vote on failure, initiating a node shutdown (`shutdown().notify()`). The error maps to `PrimaryNetworkError::Storage`, which carries no peer penalty: this is our own local disk failure, not the requesting peer's fault.

A commit failure on the epoch DB is unrecoverable for consensus safety, so the node fail-stops rather than continuing; a clean fail-stop is preferable both to a silent equivocation and to an indefinite `persist_durable().await` hang.

## Changes

- [x] `end_txn` returns the physical commit `Result`; cache mirror cleared only on success.
- [x] `DurableAck` enum + `ack_pending_durable` drive deferred acks off the real commit outcome.
- [x] `persist_durable` returns `eyre::Result<()>` (trait default, `CompositeDatabase`, `LayeredDatabase`).
- [x] `commit_failed` poison latch owned by the runner thread; set at every physical-commit site (shared-txn `commit` **and** bare no-txn `insert`/`remove`/`clear`, the guard-write path); both ack paths (deferred and no-txn) report `CommitFailed` once set, so a recast with no pending txn still fails after a past error.
- [x] Proposer returns fatal `ProposerError::DurableBarrierFailed` and does not broadcast on failure.
- [x] Vote handler awaits the barrier at all three vote exits (first vote + both recasts), fails the vote with a non-penalizing `Storage` error, and triggers node shutdown on failure.
- [x] Fault-injecting `CommitFailDb` test double + confirm-by-mutation tests; existing durable-barrier tests updated for the `Result` return.

## Testing

- `test_persist_durable_surfaces_failed_commit`: injects a failing physical commit via `CommitFailDb` while a txn is open and asserts the durable barrier resolves to `Err`, never a false-success ack. Confirmed non-vacuous by mutation: reverting the ack to always report `Committed` makes it fail.
- `test_persist_durable_latches_after_failed_commit` (new): after a faulted shared-txn commit, a *fresh* barrier with no open txn must still fail . . . the recast case. Confirmed non-vacuous by mutation: dropping the runner's latch on the commit path lets the second barrier ack success and the test fails.
- `test_persist_durable_latches_after_failed_bare_insert` (new): mirrors the actual guard-write path . . . a bare `Database::insert` with no surrounding txn (as in `write_last_proposed` / `write_vote`), whose direct physical commit is faulted, must trip the latch so a following no-txn barrier fails. Confirmed non-vacuous by mutation: dropping `commit_failed = true` from the runner's no-txn `Insert` arm lets the barrier ack success and the test fails, while the shared-txn latch test still passes (the two cover distinct runner paths).
- `test_cache_mode_retains_mem_on_failed_commit` (new): in cache mode (`full_memory = false`) a failed commit must not drain the mem mirror; asserts the retained insert survives and the value is still readable. Confirmed non-vacuous by mutation: reverting `end_txn` to drain unconditionally makes both assertions fail.
- `test_persist_durable_waits_for_open_txn_commit` / `test_persist_durable_acks_immediately_without_open_txn` (updated): positive controls, asserting the barrier reports success (`Ok`) so the fix is distinguished from a blanket refusal.
- `make attest` on the pinned 1.94 toolchain (fmt + clippy + workspace test-build).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -33,3 +33,31 @@ debug = false
 [profile.test.build-override]
 opt-level = 0
 codegen-units = 256
+
+# End-to-end test profile (see crates/e2e-tests). Selected with `--profile e2e` when
+# building the `telcoin-network` node binary the e2e suite spawns - through the Makefile's
+# `build-e2e-bin` target and, when `TN_BIN_PATH` is unset, the escargot build in
+# `get_telcoin_network_binary`. Inheriting `release` builds every crate at opt-level 2,
+# including the revm/reth/BLS dependencies where consensus and EVM runtime is spent; the
+# `dev`/`test` profiles above cap those dependencies at opt-level 1, so a plain dev build
+# would leave the hot code unoptimized. `debug-assertions` and `overflow-checks` are turned
+# back on - `release` disables them - so the suite still exercises the payload-builder and
+# header-validation invariants it asserts against. As a distinct profile its artifacts live
+# under `target/e2e/`, not shared with the dev/test `target/debug/` graph, so the dependency
+# closure compiles once more: the one-time build cost of running the suite optimized.
+[profile.e2e]
+inherits = "release"
+opt-level = 2
+debug = "line-tables-only"
+split-debuginfo = "unpacked"
+debug-assertions = true
+overflow-checks = true
+# release's default of 16, stated explicitly: codegen stays parallel (not serialized to 1 or
+# LTO) so the optimized rebuild above stays as cheap as possible.
+codegen-units = 16
+
+# Dependencies still compile at the profile's opt-level 2 but skip their debuginfo: it is
+# never inspected for third-party crates, and dropping it avoids the I/O cost on the large
+# reth/revm graph, mirroring the `dev`/`test` profiles' `package."*"` overrides.
+[profile.e2e.package."*"]
+debug = false
--- a/Makefile
+++ b/Makefile
@@ -108,13 +108,16 @@ test-faucet:
 	cargo nextest run --package telcoin-network --features faucet --test it ;
 
 # Build the node binary once so e2e test processes reuse it via TN_BIN_PATH instead of
-# rebuilding it through escargot (~3-10s of cargo overhead per test process).
+# rebuilding it through escargot (~3-10s of cargo overhead per test process). Built under the
+# optimized `e2e` profile (opt-level 2 with debug-assertions/overflow-checks kept on; see
+# `[profile.e2e]` in .cargo/config.toml) so the reused binary runs consensus and EVM at speed.
 .PHONY: build-e2e-bin
 build-e2e-bin:
-	cargo build --bin telcoin-network --features tn-storage/test-utils --target-dir $(CURDIR)/target ;
+	cargo build --profile e2e --bin telcoin-network --features tn-storage/test-utils --target-dir $(CURDIR)/target ;
 
 # Location of the binary built by build-e2e-bin, passed to the e2e tests via TN_BIN_PATH.
-E2E_BIN := $(CURDIR)/target/debug/telcoin-network
+# A named cargo profile emits into target/<profile>/, so the `e2e` profile binary lands here.
+E2E_BIN := $(CURDIR)/target/e2e/telcoin-network
 
 # run restart integration tests
 test-restarts: build-e2e-bin
--- a/crates/consensus/primary/src/certificate_fetcher.rs
+++ b/crates/consensus/primary/src/certificate_fetcher.rs
@@ -464,6 +464,12 @@ async fn fetch_from_peers<F: MissingCertFetcher>(
 
     // track requests per peer
     let mut peer_index = 0;
+    // Count completed per-peer fetches. Each spawned task sends exactly one result unless it is
+    // cancelled on another peer's success (after which we have already returned), so once every
+    // peer has been spawned and `results_received == peer_index`, all fetches have finished. This
+    // replaces `tx_results.is_closed()`, which never became true because this task holds
+    // `tx_results` for the whole loop, so the early-exit checks were dead (issue #866).
+    let mut results_received = 0usize;
 
     // spawn first peer fetch immediately
     if !peers.is_empty() {
@@ -486,6 +492,9 @@ async fn fetch_from_peers<F: MissingCertFetcher>(
     loop {
         tokio::select! {
             Some((peer, result)) = rx_results.recv() => {
+                // a spawned fetch reported back (success, empty, or error)
+                results_received += 1;
+
                 if let Ok(certificates) = result {
                     if !certificates.is_empty() {
                         debug!(target: "primary::cert_fetcher",
@@ -517,8 +526,14 @@ async fn fetch_from_peers<F: MissingCertFetcher>(
                         "Received empty certificate response, continuing with other peers");
                 }
 
-                // check if all peers have been tried and all tasks completed
-                if peer_index >= peers.len() && tx_results.is_closed() {
+                // fast-fail once every peer has been spawned and every spawned fetch has
+                // reported without success, instead of parking until the outer timeout fires.
+                // The caller (`handle_fetch_completion`) restarts immediately, so retry pacing for
+                // an unsatisfiable target comes from the staggered spawn above, a floor of
+                // ~`(peers.len() - 1) * fallback_delay`. With a single peer that stagger is absent,
+                // so pacing then relies on the per-peer fetch latency; real committees are
+                // `3f + 1` (>= 3 peers), where the stagger floor always applies.
+                if peer_index >= peers.len() && results_received >= peer_index {
                     debug!(target: "primary::cert_fetcher",
                         "All peers tried without success");
                     return Err(CertManagerError::NoCertificateFetched);
@@ -544,14 +559,14 @@ async fn fetch_from_peers<F: MissingCertFetcher>(
             }
 
             else => {
-                // all peers spawned, no results yet - wait for tasks to complete
-                if tx_results.is_closed() {
-                    debug!(target: "primary::cert_fetcher",
-                        "No peer could provide certificates");
-                    sleep(fallback_delay).await;
-                    return Err(CertManagerError::NoCertificateFetched);
-                }
-                // continue waiting for results
+                // Defensive terminal that keeps the `select!` total so it can never panic on an
+                // all-disabled poll. The `recv` branch only disables once every sender has
+                // dropped, which cannot happen while this task holds `tx_results`; ordinary
+                // completion is detected by the counter check above. If the channel is ever fully
+                // closed, report the same "nobody had them" verdict.
+                debug!(target: "primary::cert_fetcher",
+                    "No peer could provide certificates");
+                return Err(CertManagerError::NoCertificateFetched);
             }
         }
     }
--- a/crates/consensus/primary/src/certifier.rs
+++ b/crates/consensus/primary/src/certifier.rs
@@ -444,6 +444,49 @@ impl<DB: Database> Certifier<DB> {
                             error!(target: "primary::certifier", "error accepting own certificate, unable to save the certificate: {e}");
                             return Err(TaskError::from_message(e.to_string()));
                         }
+
+                        // Release the proposal lock now that the guard record is written.
+                        //
+                        // The lock exists to make the already-certified check at the top of this
+                        // method atomic with the insert above, and that pair is now complete: the
+                        // insert updates the in-memory layer synchronously, so any later proposal
+                        // reaching that check already observes this certificate. Nothing below
+                        // needs the exclusion.
+                        //
+                        // Holding it across the barrier would be actively harmful, because
+                        // `persist_durable` is a *whole-DB* barrier - the table parameter is unused
+                        // and only selects which of the three DBs to target - so it defers while
+                        // any write txn on the epoch DB is open and drains only at refcount zero,
+                        // with no timeout. Under catch-up or epoch close that is tens to hundreds
+                        // of milliseconds during which every other proposal would queue behind this
+                        // lock. Do not widen this critical section back out.
+                        drop(_guard);
+
+                        // Wait for the `ProposedCertificates` record to be durable before
+                        // externalizing. The epoch DB persists asynchronously, so the insert above
+                        // returns before the record hits disk; `process_own_certificate` below already
+                        // externalizes (it forwards the certificate on the parents bus and triggers
+                        // fetching), and the gossip publish follows. A crash in that window loses the
+                        // record, so on restart the guard at the top of this method misses and the
+                        // certifier re-proposes the same header, re-collecting votes over an unordered
+                        // channel, which can aggregate a different 2f+1 subset into a distinct aggregate
+                        // signature and thereby perturb the leader-signature randomness. See #934, #963.
+                        // If the barrier reports a failed commit (disk full, `EIO`, checksum), the
+                        // record is not on disk, so the internal processing and gossip publish below
+                        // would externalize a certificate whose guard record can be lost on restart -
+                        // exactly the re-proposal and leader-signature perturbation this barrier
+                        // exists to prevent. Refuse and fail the task (mirroring the insert failure
+                        // handled above) rather than externalize a non-durable certificate
+                        // (issue #975).
+                        self.config
+                            .node_storage()
+                            .persist_durable::<ProposedCertificates>()
+                            .await
+                            .map_err(|e| {
+                                error!(target: "primary::certifier", "durable barrier failed for own certificate, refusing to externalize: {e}");
+                                TaskError::from_message(e.to_string())
+                            })?;
+
                         // pass to state_sync for internal processing
                         if let Err(e) = self.state_sync.process_own_certificate(&mut certificate).await {
                             error!(target: "primary::certifier", "error accepting own certificate: {e}");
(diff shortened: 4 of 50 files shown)


sstanfield (dismissed):
Looks good but we need to evaluate some of these tables.  Now that we have more persist points for correctness it may be faster to just use the DB raw and skip the memory layer.  Or not, I guess it depends on how much time it will have for the background thread to catch up so may still be a win.  Either way correctness should win.

grantkee (dismissed):

MavenRain:
> Looks good but we need to evaluate some of these tables. Now that we have more persist points for correctness it may be faster to just use the DB raw and skip the memory layer. Or not, I guess it depends on how much time it will have for the background thread to catch up so may still be a win. Either way correctness should win.

Two things worth flagging before we act on it:

**1. It's an alternative fix strategy, not just a speed tweak.**  The whole durable-barrier + poison-latch machinery in this PR exists *only because* the guard writes are async-durable through the mem layer: `insert` returns before the background commit lands (`layered_db.rs` bare no-txn insert path), so a commit failure is only observable asynchronously on the DB thread.  If we wrote the guard tables synchronously-durable and read disk-first, the failure would surface inline at the write call, and the deferred barrier + cross-time latch would be unnecessary for the equivocation-safety property.  So "skip the mem layer" would close #975 a *different* way, not just make it faster.

**2. Heads-up on the lever: `full_memory=false` is not "skip the mem layer".**  Cache mode still buffers writes in mem and still serves a failed-commit value from mem (that is exactly what `test_cache_mode_retains_mem_on_failed_commit` pins down);  it only changes the retention policy (evict on a *successful* commit vs retain forever).  `get()` still reads mem-first in both modes.  So flipping the guard tables to `full_memory=false` would *not* close the window, the latch would still be required.  Truly skipping the layer is a new synchronous-durable + disk-read path that no current mode provides.

On the "evaluate some of these tables" part: the two guard tables (`LastProposed`, `Votes`) are cold.  `LastProposed` is a single-slot overwrite roughly once per proposal round, and `Votes` is one small record per author/round, bounded by committee size, so a synchronous durable write there is probably cheap.  The five bulk epoch tables (`Certificates` + its two digest indexes, `ProposedCertificates`, `Payload`) are where the async layer probably earns its keep (a cert write per authority per round, a payload token per batch), so I would keep those layered.  One wrinkle: `full_memory` is per-DB, not per-table, and all seven epoch tables share the one epoch DB, so isolating just the guard tables means splitting them into their own physical DB with a separate background thread, latch, and barrier, plus a divergence in iteration/ordering semantics.  It's likely worth measuring before committing to it.

Since correctness should win and this is a larger architectural change, I would keep it out of this PR and track it as a follow-up: https://github.com/Telcoin-Association/telcoin-network/issues/1030.  This PR stays the surgical safety fix for #975.

sstanfield (approved):

PR #980, opened 2026-07-22
test: build the e2e node binary under an opt-level-2 profile

### LINKED ISSUE #916: `test`: build the e2e node binary under an optimized profile (opt-level 2) to speed every e2e test (no coverage loss)
## Summary

The `telcoin-network` binary that every e2e test spawns and then runs for its
entire duration is built with the **unoptimized `dev` profile (`opt-level = 0`)**.
The e2e nodes therefore execute unoptimized consensus + EVM code for the whole
test.  Building that binary under an optimized profile (`opt-level = 2`) is a
test-only change that makes the node run materially faster with **identical
consensus/EVM results and no change to any assertion, epoch count, iteration
count, or timing constant.**

This is disjoint from #897 (which changes how *often* the binary is built via
`TN_BIN_PATH`/prebuild-once, but still builds it at `opt-level = 0`), from #907
and #894, and from #883 (test-side sleep→poll).  It changes the *runtime speed of
the built binary*, an axis none of those touch.

It is closely related to (but distinct from) the sibling issue "move the inert
`[profile]` tables out of `.cargo/config.toml` into `Cargo.toml`" — see
**Relationship** below.

## Root cause

`get_telcoin_network_binary()` builds the node with escargot and never selects a
profile, so cargo uses the default `dev` profile:

```rust
// crates/e2e-tests/src/lib.rs:305-322
pub fn get_telcoin_network_binary() -> &'static CargoRun {
    info!("building main binary for e2e tests");
    TELCOIN_BINARY.get_or_init(|| {
        ...
        CargoBuild::new()
            .bin("telcoin-network")
            .features("tn-storage/test-utils")
            .manifest_path(workspace_root.join("Cargo.toml"))
            .target_dir(workspace_root.join("target"))
            .current_target()
            .run()                          // <- no --profile => dev, opt-level = 0
            .expect("Failed to build telcoin-network binary")
    })
}
```

The root `Cargo.toml` has **no `[profile]` block at all**, and the `[profile.*]`
tables that exist live in `.cargo/config.toml`, which cargo does **not** read
profiles from (see the sibling issue).  Net effect: the first-party telcoin crates
*and* their dependencies compile at `opt-level = 0` for the e2e binary.

Baseline: on a `cargo nextest run -p e2e-tests --run-ignored` warm-cache run, the
CPU-bound tests dominate . . . `test_eip1559_fee_rises_at_epoch_boundaries` ~569s
(mostly the one-time build tax #897 removes, but its ~40s of real network work is
`opt-level=0` execution), `test_epoch_sync` ~98s, `test_restarts_lagged_delayed`
~98s, `test_restarts_delayed` ~95s.  The consensus DAG, BLS verification, and EVM
execution these run are exactly the code paths `opt-level` speeds up.

## Proposed change

1. Add a dedicated optimized profile to the **root `Cargo.toml`**:

```toml
[profile.e2e]
inherits = "release"
opt-level = 2
debug = 0
codegen-units = 16        # keep the build parallel
# Keep the real consensus invariants armed while still capturing the opt-level win:
debug-assertions = true
overflow-checks = true
```

2. Build the e2e binary under it (`escargot` supports passing raw cargo args):

```rust
// crates/e2e-tests/src/lib.rs, in get_telcoin_network_binary()
CargoBuild::new()
    .bin("telcoin-network")
    .features("tn-storage/test-utils")
    .manifest_path(workspace_root.join("Cargo.toml"))
    .target_dir(workspace_root.join("target"))
    .arg("--profile").arg("e2e")   // <- NEW
    .current_target()
    .run()
```

`--profile e2e` writes to `target/e2e/`;  escargot resolves the artifact path from
cargo's `--message-format=json` output, so keeping `.target_dir(.../target)` is
correct and no path change is needed.

`debug-assertions`/`overflow-checks` are kept **on** deliberately: the node has two
internal `debug_assert`s (in the payload builder and header validation) that guard
real invariants, and keeping them armed removes the only behavioral risk of the
profile switch while still capturing the dominant `opt-level 0 → 2` win.  If a
future measurement wants the last few percent, they can be turned off separately.

## Why this is new / disjoint

- **Not #897.** #897's Finding 5 (`TN_BIN_PATH` / prebuild-once) changes build
  *frequency*; the binary it prebuilds is still `opt-level = 0`.  This issue changes
  the *profile*, and in fact stacks with #897: prebuild once, at `opt-level = 2`.
- **Not the `.cargo/config.toml` sibling issue.** That issue fixes the *inert*
  `dev`/`test` profile tables (debuginfo + dependency opt) and leaves the binary at
  `opt-level = 0`.  This issue is specifically the runtime `opt-level` of the e2e
  node.
- **Not #883/#907/#894.** No sleep→poll, no epoch-duration/iteration change, no
  crypto-fixture memoization.

## Relationship to the `.cargo/config.toml` profile-fix issue

Both touch the root `Cargo.toml`.  They compose cleanly (`[profile.e2e]` is a
distinct profile from `[profile.dev]`/`[profile.test]`), but note: once this
`[profile.e2e]` is adopted, the e2e binary is optimized regardless of the sibling
fix; the sibling fix still matters for every *other* build (unit/integration
suite, local dev).  If you prefer one change, this one delivers the e2e runtime win
by itself.

## Acceptance criteria

- All e2e tests pass with identical assertions;  no epoch/iteration/timing constant
  changed.
- The e2e binary is built with `opt-level = 2` (confirm via `cargo build -v
  --profile e2e --bin telcoin-network` showing `-C opt-level=2` on first-party
  crates).
- Demonstrate the execution-bound wall-clock drop on a `make test-e2e` run (so the
  #897 build-once tax is excluded), e.g. on `test_epoch_sync` / a restart test.

## Risk

- One-time longer build on a cold cache (`opt-level = 2`);  amortized by #897's
  build-once and `codegen-units = 16`.  Net win requires stacking with #897 or a
  warm target dir.
- Keeping `debug-assertions`/`overflow-checks` on removes the behavioral-difference
  risk; the 49 explicit `checked_/wrapping_/saturating_` sites make arithmetic
  deterministic regardless.

*Line numbers are from `bump-e2e-slow-timeout-720s` (HEAD `3ef12ce3`);  re-confirm
against `main` before opening the PR.  `basefee`/`metrics` e2e tests referenced for
impact live on `main` (added by #867), not this branch.*

MavenRain:
Closes #916.

## Summary

The e2e suite spawns a real `telcoin-network` node process, and that binary was built with no profile flag, so it defaulted to the unoptimized dev profile (`opt-level = 0`).  The heavy e2e tests are CPU-bound on consensus and EVM execution, so running the node unoptimized dominates their wall-clock.

This adds a dedicated `e2e` build profile at `opt-level = 2` and points every path that produces the e2e binary at it.  Consensus and EVM results are unchanged:  `debug-assertions` and `overflow-checks` are kept on, so the invariants the suite relies on still fire, and only codegen (not semantics) changes.

## Changes

- [x] **`.cargo/config.toml`**:  new `[profile.e2e]` next to the existing `dev`/`test` profiles (the repo keeps all profile config here, not in `Cargo.toml`).  It `inherits = "release"` so every crate, including the revm/reth/BLS dependencies where the runtime is actually spent, is built at `opt-level = 2`.  The `dev`/`test` profiles cap dependencies at `opt-level = 1` via their `package."*"` overrides, so a plain dev build would leave the hot code unoptimized.  `debug-assertions` and `overflow-checks` are turned back on (release disables them).  `codegen-units = 16` is stated explicitly (release's default) to keep codegen parallel rather than serializing to LTO.  `[profile.e2e.package."*"] debug = false` drops third-party debuginfo (its I/O cost is what the `dev`/`test` profiles already avoid) while keeping deps at `opt-level = 2`.
- [x] **`Makefile`**:  `build-e2e-bin` now runs `cargo build --profile e2e`; `E2E_BIN` points at `target/e2e/telcoin-network` (a named profile emits into `target/<profile>/`).  This feeds `test-restarts`, `test-e2e`, and `public-tests` via `TN_BIN_PATH`.
- [x] **`etc/test-and-attest.sh`**:  the prebuild step (run by `make attest`) uses `--profile e2e` and sets `TN_BIN_PATH` to `target/e2e/telcoin-network`, so the built path and the consumed path stay in lockstep.
- [x] **`crates/e2e-tests/src/lib.rs`**:  the `TN_BIN_PATH`-unset fallback in `get_telcoin_network_binary()` passes `--profile e2e` to escargot's `CargoBuild`, matching the Makefile.  escargot resolves the artifact path from cargo's JSON output, so the `target/e2e/` location is found without assuming `target/debug/`.  Stale `opt-level 0` / `target/debug` wording in the surrounding comment and the operator-facing hint was corrected.
- [x] **`crates/e2e-tests/README.md`**:  the manual-run instructions now use `--profile e2e` and `target/e2e/telcoin-network`.

## Acceptance criteria

- [x] Binary is compiled with `opt-level = 2`:  `[profile.e2e]` sets it, `inherits = "release"` extends it to the dependency closure, and both the Makefile and escargot paths select `--profile e2e`.
- [x] No changes to test assertions, epoch counts, iteration counts, or timing constants:  the diff touches only build config, one escargot argument, and comments/docs.  Zero test files change.
- [x] Safety invariants preserved:  `debug-assertions = true` and `overflow-checks = true` keep `debug_assert!`/overflow panics and every `#[cfg(debug_assertions)]` path active, so consensus/EVM behavior is identical to the dev-profile binary.
- [ ] Measurable wall-clock reduction demonstrated:  captured by the reviewer's `make test-e2e` run (see Verification).  The optimized build is a one-time cost, so the win shows on the ignored suite, not on a single test.

## Verification

Confirmed in this change:

- `cargo check -p e2e-tests` builds clean, including `escargot` and the new `--profile e2e` argument.
- `cargo +nightly-2026-03-20 fmt --check` is clean (the `wrap_comments` reflow is applied).
- The profile resolves and emits to `target/e2e/` (verified against `cargo 1.96`;  a custom named profile defined in `.cargo/config.toml` is valid and builds into `target/<profile>/`).
- Every e2e-binary path moves together (`Makefile` `E2E_BIN`, `etc/test-and-attest.sh` `TN_BIN_PATH`, the escargot fallback, and the README), so the `is_file` assert in `get_telcoin_network_binary()` cannot be handed a stale `target/debug/` path.  The production `target/release/telcoin-network` references (README, Dockerfile) are intentionally untouched.

Wall-clock, to capture on a full checkout under the pinned toolchain:

```
make test-e2e     # builds --profile e2e once, then runs the ignored suite with TN_BIN_PATH set
```

The issue flags `test_eip1559_fee_rises_at_epoch_boundaries` and `test_epoch_sync` as the CPU-bound tests expected to improve.

## Notes / tradeoff

The `e2e` profile is a distinct optimization level, so its artifacts live under `target/e2e/` and do not share the `dev`/`test` `target/debug/` graph.  The dependency closure therefore compiles once more:  this is the inherent, one-time build cost of running the suite optimized, and is unavoidable for any profile that raises the dependencies above the dev/test `opt-level = 1`.  The issue accepts this and defers cold-build amortization to #897; `codegen-units = 16` keeps that extra compile parallel.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -33,3 +33,31 @@ debug = false
 [profile.test.build-override]
 opt-level = 0
 codegen-units = 256
+
+# End-to-end test profile (see crates/e2e-tests). Selected with `--profile e2e` when
+# building the `telcoin-network` node binary the e2e suite spawns - through the Makefile's
+# `build-e2e-bin` target and, when `TN_BIN_PATH` is unset, the escargot build in
+# `get_telcoin_network_binary`. Inheriting `release` builds every crate at opt-level 2,
+# including the revm/reth/BLS dependencies where consensus and EVM runtime is spent; the
+# `dev`/`test` profiles above cap those dependencies at opt-level 1, so a plain dev build
+# would leave the hot code unoptimized. `debug-assertions` and `overflow-checks` are turned
+# back on - `release` disables them - so the suite still exercises the payload-builder and
+# header-validation invariants it asserts against. As a distinct profile its artifacts live
+# under `target/e2e/`, not shared with the dev/test `target/debug/` graph, so the dependency
+# closure compiles once more: the one-time build cost of running the suite optimized.
+[profile.e2e]
+inherits = "release"
+opt-level = 2
+debug = "line-tables-only"
+split-debuginfo = "unpacked"
+debug-assertions = true
+overflow-checks = true
+# release's default of 16, stated explicitly: codegen stays parallel (not serialized to 1 or
+# LTO) so the optimized rebuild above stays as cheap as possible.
+codegen-units = 16
+
+# Dependencies still compile at the profile's opt-level 2 but skip their debuginfo: it is
+# never inspected for third-party crates, and dropping it avoids the I/O cost on the large
+# reth/revm graph, mirroring the `dev`/`test` profiles' `package."*"` overrides.
+[profile.e2e.package."*"]
+debug = false
--- a/Makefile
+++ b/Makefile
@@ -108,13 +108,16 @@ test-faucet:
 	cargo nextest run --package telcoin-network --features faucet --test it ;
 
 # Build the node binary once so e2e test processes reuse it via TN_BIN_PATH instead of
-# rebuilding it through escargot (~3-10s of cargo overhead per test process).
+# rebuilding it through escargot (~3-10s of cargo overhead per test process). Built under the
+# optimized `e2e` profile (opt-level 2 with debug-assertions/overflow-checks kept on; see
+# `[profile.e2e]` in .cargo/config.toml) so the reused binary runs consensus and EVM at speed.
 .PHONY: build-e2e-bin
 build-e2e-bin:
-	cargo build --bin telcoin-network --features tn-storage/test-utils --target-dir $(CURDIR)/target ;
+	cargo build --profile e2e --bin telcoin-network --features tn-storage/test-utils --target-dir $(CURDIR)/target ;
 
 # Location of the binary built by build-e2e-bin, passed to the e2e tests via TN_BIN_PATH.
-E2E_BIN := $(CURDIR)/target/debug/telcoin-network
+# A named cargo profile emits into target/<profile>/, so the `e2e` profile binary lands here.
+E2E_BIN := $(CURDIR)/target/e2e/telcoin-network
 
 # run restart integration tests
 test-restarts: build-e2e-bin
--- a/crates/e2e-tests/README.md
+++ b/crates/e2e-tests/README.md
@@ -4,7 +4,8 @@
 
 The heavy e2e tests (restart and epoch tests) are `#[ignore]`d and each needs the
 `telcoin-network` node binary. When `TN_BIN_PATH` is unset, the first test builds that binary
-in-process via `escargot`, which at `opt-level = 0` is a multi-minute compile. Under `nextest`'s
+in-process via `escargot`, which under the `e2e` profile (`opt-level = 2`) is a multi-minute
+compile. Under `nextest`'s
 default output capture that build is buffered, so the first test looks frozen for several minutes
 before it does anything.
 
@@ -17,8 +18,8 @@ make test-e2e          # builds the binary once, then runs the suite with TN_BIN
 or run the raw command yourself (for example from an IDE test runner):
 
 ```
-cargo build --bin telcoin-network --features tn-storage/test-utils
-TN_BIN_PATH="$(pwd)/target/debug/telcoin-network" \
+cargo build --profile e2e --bin telcoin-network --features tn-storage/test-utils
+TN_BIN_PATH="$(pwd)/target/e2e/telcoin-network" \
   cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features
 ```
 
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -399,17 +399,18 @@ pub fn get_telcoin_network_binary() -> &'static TestBinary {
             return TestBinary::Prebuilt(path);
         }
 
-        // TN_BIN_PATH is unset, so build the node binary in-process via escargot. At
-        // opt-level 0 this is a multi-minute compile, and nextest captures stdout/stderr, so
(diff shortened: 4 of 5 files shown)


sstanfield (approved):
I'm down.  I often run the e2e tests manually with release builds already...

PR #982, opened 2026-07-22
test(storage): open test-build MDBX envs in SafeNoSync to drop hot-path fsync

### LINKED ISSUE #917: `test`: open e2e MDBX databases in SafeNoSync under test cfg to remove hot-path fsync (no coverage loss)
## Summary

Every e2e node opens its MDBX databases in the **durable sync mode**, so consensus
commits and reth block writes `fsync` on the write path.  Under `#[cfg(test)]` /
`feature = "test-utils"` these nodes run on a throwaway `temp_path` that is deleted
at teardown, so the durability guarantee buys nothing — the tests only ever restart
a node by **killing and re-launching the process on the same host**, which
`MDBX_SAFE_NOSYNC` already survives.  Switching the test-build DB envs to
`SafeNoSync` removes the hot-path `fsync` latency with **no change to any assertion
and no loss of the restart-recovery coverage.**

Disjoint from the whole ledger: #897/#907/#894 never touch DB durability, and #883
is test-side sleep→poll, not node-side I/O.

**Caveat up front:** the win is disk-dependent.  On fast local NVMe with a warm page
cache it is near sub-second;  on a slow CI disk it is several seconds per long
epoch/restart test.  Treat this as a CI-runner optimization, not a guaranteed
local speedup.

## Root cause

Two DB open sites, both durable today.

**1. Consensus / epoch MDBX (this repo's own code).** `MdbxDatabase::open` builds
the environment with `write_map()` + geometry but sets **no sync mode**, so it
defaults to durable and `fsync`s at every `txn.commit()`:

```rust
// crates/storage/src/mdbx/database.rs:119-140
pub fn open<P: AsRef<Path>>(path: P, max_tables: usize, max_size: usize, growth_step: usize)
    -> eyre::Result<Self> {
    Environment::builder()
        .set_max_dbs(max_tables)
        .write_map()
        .set_geometry(Geometry { ... })
        // <- no sync-mode flag => durable; fsync at every commit (:157,180,187,194)
        .open(path.as_ref()).map(|env| MdbxDatabase { inner: env })
}
```

These envs are opened for the epoch/kad/cache DBs from a helper that **already**
branches on the test cfg (for DB sizing), which is the natural place to also pick
the sync mode:

```rust
// crates/storage/src/lib.rs:130-163  (_open_mdbx)
cfg_if::cfg_if! {
    if #[cfg(any(test, feature = "test-utils"))] {
        const EPOCH_MAX: usize = 16 * MEGABYTE; ...   // test sizes
    } else {
        const EPOCH_MAX: usize = 512 * MEGABYTE; ...  // prod sizes
    }
}
let epoch_db = MdbxDatabase::open(store_path.join("epoch"), 8, EPOCH_MAX, GROWTH)...
```

Crucially, the e2e binary is built with `.features("tn-storage/test-utils")`
(`crates/e2e-tests/src/lib.rs:315`), so the `feature = "test-utils"` cfg is **live
in the spawned node processes** . . . a cfg-gated sync mode reaches the e2e nodes with
**no CLI plumbing required.**

**2. reth block DB (optional, secondary).** reth's DB is opened durable via:

```rust
// crates/tn-reth/src/lib.rs:586-593
pub fn new_database<P: AsRef<Path>>(reth_config: &RethConfig, db_path: P) -> eyre::Result<RethDb> {
    ...
    Ok(Arc::new(init_db(db_path, reth_config.0.db.database_args())?))  // :592, durable
}
```

## Proposed change

**Consensus MDBX (primary, fully in-repo).** Thread a sync mode into
`MdbxDatabase::open` (or add a cfg branch inside it) so test builds use
`SafeNoSync`:

```rust
// database.rs, in open()
let mut builder = Environment::builder();
builder.set_max_dbs(max_tables).write_map().set_geometry(Geometry { ... });
    mode: Mode::ReadWrite { sync_mode: SyncMode::SafeNoSync },
    ..Default::default()
});
let env = builder.open(path.as_ref())?;
```

(Equivalently, select `SyncMode::SafeNoSync` vs `Durable` in the existing
`_open_mdbx` cfg_if at `storage/src/lib.rs:139-152` and pass it through `open`.)

**reth MDBX (optional).** If reth's `DatabaseArguments`
(`reth_config.0.db.database_args()`) exposes a sync-mode setter in the pinned reth
version, set `SafeNoSync` under the same test cfg;  otherwise skip this half . . . the
consensus DB is where every DAG round commits.  *(Confirm the reth
`DatabaseArguments` API before implementing this half.)*

**Hard constraint: `SafeNoSync`, never `UtterlyNoSync`.** With `write_map()` +
`SafeNoSync`, committed data survives a **process** crash/kill (it is in the file
-backed mmap / OS page cache and MDBX keeps a recoverable steady-state);  it is lost
only on an **OS/power** crash, which the tests never induce.  `UtterlyNoSync` can
leave the meta page inconsistent on process death and would break the all-nodes-down
restart-recovery assertions.  This distinction is the whole correctness argument —
do not weaken it.

## Why this is coverage-preserving

The restart tests (`test_restarts_delayed`, `test_restarts_lagged_delayed`,
`test_restarts_observer`, the observer catch-up/pause tests) restart a node by
killing its process and relaunching against the same on-disk datadir, then assert it
recovers committed state and rejoins.  `SafeNoSync` preserves exactly that state
across process death, so every restart assertion still holds. No test asserts
durability across a kernel crash or power loss.  All other assertions
(balances/fees/epoch records) are unaffected . . . only `fsync` latency is removed.

## Acceptance criteria

- All e2e tests pass with identical assertions, including every restart-recovery
  path.
- The test-build consensus MDBX env opens with `SafeNoSync` (verify via a debug log
  of the chosen mode, or an MDBX env-flags assertion in a test).
- Production builds are unchanged (still durable) . . . the cfg gate must not touch the
  non-test branch.
- Report the wall-clock delta on a slow-disk CI runner (where the win is real).

## Risk

- **Correctness:** only if `UtterlyNoSync` is used by mistake, enforced by review
  to `SafeNoSync`.
- **Savings variance:** near-zero on fast NVMe;  only material on slow CI disk. Worth
  it because it is a small, low-risk, stacking change (pairs with a tmpfs datadir
  for the same tests), not because it is a large guaranteed win.
- The reth half depends on the pinned reth `DatabaseArguments` API;  gate it behind a
  confirmation and ship the consensus half regardless.

*Line numbers from `bump-e2e-slow-timeout-720s` (HEAD `3ef12ce3`); re-confirm
against `main` before the PR.*

MavenRain:
Closes #917.

## Summary

Every e2e node opens its consensus MDBX databases (epoch / kad / cache) in the
default **durable** sync mode, so every `txn.commit()` on the consensus hot path
`fsync`s the meta and data pages.  Under `#[cfg(test)]` / `feature = "test-utils"`
those nodes run on a throwaway `temp_path` that is deleted at teardown, and the
suite only ever restarts a node by **killing and relaunching the process on the
same host** . . . a recovery that `MDBX_SAFE_NOSYNC` already survives.  This switches
the test-build consensus MDBX envs to `SafeNoSync`, which removes the hot-path
`fsync` with **no change to any assertion and no loss of restart-recovery
coverage.**  Production builds are untouched and stay fully `Durable`.

The win is disk-dependent: near sub-second on warm local NVMe, several seconds per
long epoch/restart test on a slow CI disk.  This is a low-risk, stacking CI-runner
optimization (it pairs with a tmpfs datadir), not a guaranteed local speedup.

## Changes

- `crates/storage/src/mdbx/database.rs`
  - [x] Add `BUILD_SYNC_MODE`, a cfg-gated const that is the **single source of
    truth** for the sync mode: `SafeNoSync` under `#[cfg(any(test, feature =
    "test-utils"))]`, and *not defined at all* in production builds (where the env
    keeps the default `Durable`).
  - [x] `MdbxDatabase::open` applies `BUILD_SYNC_MODE` via `set_flags(...)` inside
    the same cfg gate.  `write_map()` (env *kind*) and `set_flags` (env *flags*)
    are independent builder fields, so the only bit this changes versus today is
    the sync mode;  every other flag keeps its existing default.
  - [x] Add `test_open_uses_safe_no_sync_under_test_cfg`, which pins the exact
    compiled mode and proves the flag reaches MDBX (see Verification).

The gate is `any(test, feature = "test-utils")` on purpose.  `cfg(test)` covers
`tn-storage`'s own unit tests;  `feature = "test-utils"` is what reaches the e2e
nodes, which are spawned as separate processes built with
`--features tn-storage/test-utils` (`crates/e2e-tests/src/lib.rs`), where
`cfg(test)` is not live.  No CLI plumbing is required to select the mode.

## Acceptance criteria

- [x] The test-build consensus MDBX env opens in `SafeNoSync` (asserted from the
  live env, not just the source const).
- [x] `SafeNoSync`, never `UtterlyNoSync`.  `SafeNoSync` keeps the last steady
  commit's pages untouched, so it is corruption-proof on *any* crash (it can always
  roll back to that steady commit) and merely loses the last transactions on an
  OS/power crash;  a committed txn always survives a **process** kill+restart because
  the data is in the file-backed mmap / OS page cache and a relaunched process
  re-maps it.  `UtterlyNoSync` discards that steady commit for marginally faster
  writes and can corrupt the whole database on an **OS/power** crash, a needless risk
  for our access pattern.  Note the e2e restart suite would **not** reliably catch a
  drift to `UtterlyNoSync` (an OS-alive process restart recovers under either
  no-sync mode), so the guard is the `BUILD_SYNC_MODE` const and its `assert_eq!`,
  not the restart tests.  This documentation-accuracy point was raised and fixed in
  adversarial review.
- [x] Production builds are unchanged.  The `test-utils` feature is enabled only
  through `[dev-dependencies]` edges (`tn-types`, `state-sync`, `batch-builder`,
  `consensus/primary`, `network-libp2p`), never a normal `[dependencies]` edge, so
  a release `cargo build --bin telcoin-network` never sees it and the cfg block is
  compiled out.  `test-utils` is not in `default`.
- [x] All existing storage assertions are unaffected . . . only `fsync` latency is
  removed.

## Scope: consensus MDBX only (reth block DB deferred)

The issue names an optional secondary half: reth's block DB
(`crates/tn-reth/src/lib.rs::new_database` via `init_db` +
`DatabaseArguments::with_sync_mode`).  It is deliberately **not** included here.
The e2e node binary is built with only `--features tn-storage/test-utils`;  nothing
enables a `tn-reth` test feature, and `cfg(test)` is not live in the spawned
binary.  So the reth half cannot reach the e2e nodes without adding a new `tn-reth`
feature *and* changing the e2e build command . . . exactly the CLI plumbing the
issue asked to avoid.  The consensus MDBX env is where every DAG round commits, so
it is the primary hot path;  the reth half can follow as its own change if the
extra build wiring is judged worthwhile.

## Verification

Pinned toolchain (`rust-toolchain.toml` = 1.94).

- `cargo +1.94 test -p tn-storage --lib` . . . **151 passed;  0 failed**, including
  the new `test_open_uses_safe_no_sync_under_test_cfg`.
- **Confirm-by-mutation** (the regression test was *seen* to fail): flipping
  `BUILD_SYNC_MODE` to `UtterlyNoSync` fails the guard with
  `assertion left == right failed  left: UtterlyNoSync  right: SafeNoSync`;
  restored to `SafeNoSync`.
- The test asserts in two ways.  (1) `assert_eq!(BUILD_SYNC_MODE,
  SyncMode::SafeNoSync)` pins the exact mode (`SyncMode: PartialEq`), which is what
  catches an accidental `UtterlyNoSync`.  (2) A read-back of the opened env
  (`db.inner.info().mode()`) asserts the env is **not** `Durable`, proving the flag
  actually took effect.  Note: MDBX defines
  `MDBX_UTTERLY_NOSYNC = MDBX_SAFE_NOSYNC | <extra bit>`, and reth-libmdbx's
  `mode()` tests the UtterlyNoSync mask before the SafeNoSync one, so it reports
  `UtterlyNoSync` for a genuine `SafeNoSync` env . . . which is why the exact-mode
  guarantee is pinned by the const assert, not the read-back.
- `cargo +nightly fmt -- --check` clean;  `cargo +1.94 clippy -p tn-storage
  --all-targets --features test-utils` adds no new warnings (the 54 reported are
  pre-existing storage-test lints unrelated to this diff).
- `make attest` on 1.94 (fmt + clippy + `--workspace` test-build in an isolated
  target dir) is the final gate;  run it against the pushed hash.

The full e2e suite (every restart-recovery path) is the acceptance surface for the
runtime win and is best measured on a CI runner with a slow disk . . . the change is
assertion-preserving by construction, so those tests should pass unchanged.

### DIFF
--- a/crates/storage/src/mdbx/database.rs
+++ b/crates/storage/src/mdbx/database.rs
@@ -113,6 +113,14 @@ fn default_page_size() -> usize {
     os_page_size.clamp(min_page_size, libmdbx_max_page_size)
 }
 
+/// The MDBX sync mode compiled into `test` and `test-utils` builds. `SafeNoSync` removes the
+/// hot-path `fsync` while still surviving a process kill+restart (see [`MdbxDatabase::open`]);
+/// it must never be `UtterlyNoSync`, which risks whole-DB corruption on an OS/power crash. This
+/// is the single source of truth `open` applies, so a test can pin it exactly. Production builds
+/// enable neither cfg and this const does not exist for them (the env stays `Durable`).
+#[cfg(any(test, feature = "test-utils"))]
+const BUILD_SYNC_MODE: reth_libmdbx::SyncMode = reth_libmdbx::SyncMode::SafeNoSync;
+
 impl MdbxDatabase {
     /// Creates a new database at the specified path if it doesn't exist. Does NOT create tables.
     /// Check [`init_db`].
@@ -122,19 +130,51 @@ impl MdbxDatabase {
         max_size: usize,
         growth_step: usize,
     ) -> eyre::Result<Self> {
-        let env = Environment::builder()
-            .set_max_dbs(max_tables)
-            .write_map()
-            .set_geometry(Geometry {
-                // Maximum database size
-                size: Some(0..max_size),
-                // We grow the database in increments of 1 gigabyte
-                growth_step: Some(growth_step as isize),
-                // The database never shrinks
-                shrink_threshold: Some(0),
-                page_size: Some(PageSize::Set(default_page_size())),
-            })
-            .open(path.as_ref())?;
+        let mut builder = Environment::builder();
+        builder.set_max_dbs(max_tables).write_map().set_geometry(Geometry {
+            // Maximum database size
+            size: Some(0..max_size),
+            // We grow the database in increments of 1 gigabyte
+            growth_step: Some(growth_step as isize),
+            // The database never shrinks
+            shrink_threshold: Some(0),
+            page_size: Some(PageSize::Set(default_page_size())),
+        });
+
+        // Test and `test-utils` builds trade fsync durability for write speed: they open the
+        // env in `SafeNoSync` instead of the default `Durable`, which removes the meta+data
+        // `fsync` that MDBX performs at every `txn.commit()` on the consensus hot path.
+        //
+        // This is coverage-preserving. With `write_map()` + `SafeNoSync`, a committed
+        // transaction survives a *process* crash/kill+restart -- the only recovery the tests
+        // ever exercise (they restart a node by relaunching the process against the same
+        // on-disk datadir) -- because the data lives in the file-backed mmap / OS page cache
+        // and a relaunched process re-maps it. Durability is lost only on an *OS/power* crash,
+        // which no test induces.
+        //
+        // It must be `SafeNoSync`, never `UtterlyNoSync`. `SafeNoSync` keeps the last steady
+        // commit's pages untouched, so it is corruption-proof on *any* crash (it can always roll
+        // back to that steady commit) and merely loses the last transactions on an OS/power
+        // crash. `UtterlyNoSync` discards that steady commit for marginally faster writes and can
+        // corrupt the whole database on an OS/power crash -- a needless risk given our access
+        // pattern. Note the e2e suite would NOT reliably catch a drift to `UtterlyNoSync`: an
+        // OS-alive process restart recovers under either mode, so the `BUILD_SYNC_MODE` const and
+        // its `assert_eq!` are what actually enforce the choice, not the restart tests.
+        //
+        // The `feature = "test-utils"` arm is what reaches the e2e nodes: they are spawned as
+        // separate processes built with `--features tn-storage/test-utils`, where `cfg(test)`
+        // is not live, so no CLI plumbing is required to select the mode. Production builds
+        // enable neither cfg, so this block is compiled out and the default `Durable` stands.
+        #[cfg(any(test, feature = "test-utils"))]
+        {
+            use reth_libmdbx::{EnvironmentFlags, Mode};
+            builder.set_flags(EnvironmentFlags {
+                mode: Mode::ReadWrite { sync_mode: BUILD_SYNC_MODE },
+                ..Default::default()
+            });
+        }
+
+        let env = builder.open(path.as_ref())?;
 
         Ok(MdbxDatabase { inner: env })
     }
@@ -427,5 +467,35 @@ mod test {
         let temp_dir = tempdir().expect("failed to create temp dir");
         let db = open_db(temp_dir.path());
         db_simp_bench(db, "MDBX");
+    }
+
+    /// #917: under the `test`/`test-utils` cfg the env must open in `SafeNoSync` (no hot-path
+    /// `fsync`) -- specifically not `Durable` (would keep the fsync) and, critically, not
+    /// `UtterlyNoSync` (risks whole-DB corruption on an OS/power crash). The e2e restart suite
+    /// cannot tell the two no-sync modes apart, so this const-assert is the real guard.
+    #[test]
+    fn test_open_uses_safe_no_sync_under_test_cfg() {
+        use super::BUILD_SYNC_MODE;
+        use reth_libmdbx::{Mode, SyncMode};
+
+        // Pin the exact compiled mode. `SyncMode` is `PartialEq`, so this catches a drift to
+        // `Durable` (fsync back on) or to `UtterlyNoSync` (recovery broken) -- the latter is the
+        // one the readback below cannot see, so this assert is what guards the hard constraint.
+        assert_eq!(BUILD_SYNC_MODE, SyncMode::SafeNoSync);
+
+        // And prove the flag actually reached MDBX: the opened env reports a non-`Durable` mode,
+        // so the hot-path fsync is genuinely gone. NOTE: MDBX defines
+        // `MDBX_UTTERLY_NOSYNC = MDBX_SAFE_NOSYNC | <extra bit>`, and reth-libmdbx's `mode()`
+        // tests the UtterlyNoSync mask before the SafeNoSync one, so it reports `UtterlyNoSync`
+        // for a genuine `SafeNoSync` env. We therefore assert only "not Durable / not read-only"
+        // from the readback; the exact-mode guarantee is pinned by the const assert above. Do not
+        // "fix" this into `assert_eq!(mode, ..SafeNoSync)` -- it cannot pass by construction.
+        let temp_dir = tempdir().expect("failed to create temp dir");
+        let db = open_db(temp_dir.path());
+        let mode = db.inner.info().expect("read env info").mode();
+        assert!(
+            matches!(mode, Mode::ReadWrite { sync_mode } if sync_mode != SyncMode::Durable),
+            "test-build MDBX env must not be Durable (hot-path fsync must be off), got {mode:?}"
+        );
     }
 }


sstanfield (commented):
Not sure about this.  We might mask a timing issue caused by syncing that the e2e tests would fail to identify.

MavenRain:
That's fair enough.  Let me lay out exactly where a
durability commit does and doesn't sit on a timing-sensitive path, as tracing
it changed how I'd want to land this.

The bulk of consensus persistence isn't affected by the sync mode.  Every write goes
through `LayeredDatabase`, whose `commit()`/`insert()` only touch the in-memory
layer and hand the physical MDBX commit (the Durable-mode `fsync`) to the background
writer thread (`layered_db.rs`).  The writing task returns without ever blocking on
that `fsync`, so certificate-store writes, GC, batch sync, and consensus-output
persistence are all fire-and-forget with respect to Durable vs SafeNoSync.

The places that *do* matter are the three `persist_durable().await` barriers on the
externalization path: the certifier before it forwards/gossips its own certificate
(`certifier.rs:474`), the proposer before it broadcasts a header (`proposer.rs:346`),
and the vote handler before it replies with a vote (`handler.rs:933`).  Each awaits a
whole-DB durability barrier before releasing something onto the wire, and that
barrier's latency races the proposer's min/max header timers (1000/2500ms) and the
vote-collection window.  So you're right that a durability commit sits inside a
timing-sensitive section at these three sites, and SafeNoSync does shorten it.

Two things keep that safe, and one thing I'd like to add because of it:

1. The barrier is a FIFO drain of the background-writer queue, not an fsync-to-media
   wait.  `DurableBarrier` (`layered_db.rs:265`) just holds the ack until the covering
   txn `commit()`s;  it never calls `env.sync`.  So SafeNoSync preserves the *ordering*
   these barriers exist for: the certifier still gossips only after the cert is
   committed, the proposer still broadcasts only after `LastProposed` is committed,
   the voter still replies only after `Votes` is committed.  What SafeNoSync removes is
   the `fsync` *inside* that commit, not the barrier itself.  And because the data is
   in the write_map'd page cache, a process kill+restart re-maps it, which is the only
   recovery the suite exercises, so the equivocation/crash-consistency invariant those
   barriers protect stays intact.

2. On the disk the e2e nodes actually use (a throwaway `TempDir`), the Durable
   per-commit `fsync` is already small relative to those 1000/2500ms timers, so the
   barriers are close to their SafeNoSync latency today.  The regime where the barrier
   is genuinely slow (the "tens to hundreds of ms" the comment calls out) is
   catch-up / epoch-close and slow CI disks, which is exactly the latency this PR is
   trying to remove.

Here's your point restated in a way I agree with: the one setting that genuinely
exercises "externalize after a real durability delay, racing the leader timers" is
Durable on a slow disk, and this change removes it from the e2e suite.  Given the
suite is already timing-sensitive (the basefee e2e test is documented as flaky under
parallel load, I had raised the CI slow-timeout from 120s to 720s, and the
catch-up-vs-backfill races have been a recurring fix area), I don't want to delete
that timing regime on faith.

So rather than the hard compile-time switch this PR currently uses, I'd make the sync
mode a runtime override: SafeNoSync stays the default for the fast path, but
`TN_TEST_MDBX_SYNC=durable` forces Durable at `open()` with no rebuild.  Then we keep a
Durable e2e lane (nightly, or the heavy `#[ignore]`d restart/epoch tests run as a
separate step) that still *identifies* any latency-sensitive bug the way you want, on
a schedule where the wall-clock cost doesn't matter, while PR-time CI gets the
speedup.  It's a few lines and no call-site churn, and I'll keep the `BUILD_SYNC_MODE`
const-assert so the compiled default still can't drift to `UtterlyNoSync`.

Do you want me to fold that into this PR, or land this as-is and track the Durable lane
separately?

sstanfield (approved):

Issue #942, opened 2026-07-16
Reputation Schedule - Evaluate periodic reset


grantkee:
# Problem

With clarity around committee and epochs, we should evaluate how the reputation scores clear and ensure meaningful reputation impact for byzantine behavior.

Issue #435, opened 2025-09-30
Unfair authority selection in `retrieve_first_nodes` based on stake distribution rather than reputation scores


leopashov:
**Cantina finding:** #686  
**Severity:** Medium

### Relevant Context

Telcoin Network implements a reputation-based leader scheduling system as part of its Narwhal/Bullshark consensus mechanism. The `LeaderSwapTable` maintains sets of "good" and "bad" nodes based on their reputation scores, allowing the system to swap out poorly performing leaders during consensus rounds.

The `retrieve_first_nodes` function in `crates/consensus/primary/src/consensus/leader_schedule.rs` is responsible for selecting authorities up to a configurable stake threshold. This function is used to determine both good and bad node sets, where good nodes are selected from the highest reputation scores and bad nodes are selected from the lowest reputation scores.

### Vulnerability Description

The `retrieve_first_nodes` function exhibits unfair selection behavior that can result in authorities being classified as "bad nodes" despite having identical reputation scores to authorities classified as "good nodes". This occurs because the function prioritizes stake-based thresholds over reputation-based fairness when authorities have equal reputation scores.

The vulnerability is particularly severe when new reputation score sets are created and stakes do not meet the threshold to clear the bad nodes set, causing some validators to be immediately inserted into the `bad_nodes` set based on arbitrary tie-breaking rather than actual performance differences.

The issue stems from the deterministic tie-breaking mechanism in `ReputationScores::authorities_by_score_desc()` combined with the stake-based selection logic in `retrieve_first_nodes`:

---

#### Deterministic tie-breaking in reputation scores

The reputation scoring system in `crates/types/src/primary/reputation.rs` implements deterministic tie-breaking:

```rust
authorities.sort_by(|a1, a2| {
    match a2.1.cmp(&a1.1) {
        Ordering::Equal => {
            // we resolve the score equality deterministically by ordering in authority
            // identifier order descending.
            a2.0.cmp(&a1.0)
        }
        result => result,
    }
});
```

When multiple authorities have the same reputation score, they are ordered by `AuthorityIdentifier` in descending order.

---

#### Stake-based selection logic

The `retrieve_first_nodes` function in `crates/consensus/primary/src/consensus/leader_schedule.rs` selects authorities based on cumulative stakes thresholds:

```rust
fn retrieve_first_nodes(
    committee: &Committee,
    authorities: impl Iterator<Item = (AuthorityIdentifier, u64)>,
    voting_power_threshold: u64,
[...]

Issue #434, opened 2025-09-30
Bad nodes face economic exclusion through irrecoverable reputation score trap


leopashov:
**Cantina finding**: #681
**Severity**: Medium

---

## Relevant Context

Telcoin Network implements a reputation-based leader scheduling system as part of its Narwhal/Bullshark consensus mechanism. The LeaderSwapTable maintains sets of "good" and "bad" nodes based on their reputation scores, allowing the system to swap out poorly performing leaders during consensus rounds.

The reputation system works by tracking how often validators reference previous leaders' certificates as parents in their proposals. Validators earn reputation points when their certificates are included in committed sub-DAGs and correctly reference the previous leader. The system resets reputation scores every 300 committed sub-DAGs (num_sub_dags_per_schedule) and uses these final scores to create new leader swap tables that classify validators based on relative ranking.

When the bad_nodes_stake_threshold parameter is greater than 0, validators with relatively low reputation scores may classified as "bad nodes" and are systematically excluded from leader selection through the swap mechanism. However, this creates an irrecoverable reputation trap that permanently excludes validators from being selected as leaders.

---

## Vulnerability Description

The reputation system creates a structural impossibility for "bad" validators to recover their relative ranking, permanently trapping them in the "bad" status. Once classified as "bad," validators face asymmetric opportunities for earning reputation points that make recovery impossible, even when they perform identically to "good" validators.

The core issue lies in the intersection of three mechanisms:

### 1. Asymmetric inclusion in Sub-DAGs

The `order_dag` function in `crates/consensus/primary/src/consensus/utils.rs` determines which certificates are included in a `committed_sequence` when scoring reputation:

```rust
pub fn order_dag(leader: &Certificate, state: &ConsensusState) -> Vec<Certificate> {
    let mut buffer = vec![leader];  // Leader always included
    while let Some(x) = buffer.pop() {
        ordered.push(x.clone());
        // Add parent certificates that this cert references
        for parent in x.header().parents() {
            // Include certificates transitively referenced by the leader
        }
    }
}
```

Note that good nodes (who always become leaders) are guaranteed inclusion in every sub-DAG they lead.

### 2. Equal point distribution

The reputation scoring in `crates/consensus/primary/src/consensus/bullshark.rs` lines 94-106 awards points equally to all included certificates:

```rust
for certificate in committed_sequence {  // All validators in sub-DAG
    if certificate.header().parents().iter().any(|digest| 
[...]

PR #983, opened 2026-07-22
docs(consensus): evaluate reputation schedule reset and byzantine impact (#942)

MavenRain:
Addresses #942.

## Summary

Issue #942 asks us to evaluate how consensus reputation scores clear and whether byzantine behavior has a meaningful reputation impact, now that committee/epoch semantics are settled.  This PR delivers that evaluation as a reference document that lives next to the code it describes, `crates/consensus/primary/src/consensus/reputation.md`, following the pattern already set by `crates/network-libp2p/src/peers/penalty.md`.

The document is docs-only.  It changes no behavior.  Every mechanism it states is written as an assertion with a `file:line` citation, and every claim was verified against the code at the cited location.

## Why a document, and why now

Consensus reputation is committed state: the final scores are hashed into the `CommittedSubDag` digest (`crates/types/src/primary/output.rs:467`).  So the good/bad leader sets derived from them are a safety property, not a performance heuristic . . . every honest validator must compute byte-identical scores from the same committed DAG or the network forks.  This exact code is also the fix for two prior Cantina-medium findings (#434 "irrecoverable reputation trap", #435 "unfair stake-based selection").  Any actual change to scoring or reset cadence is therefore a coordinated hard fork and warrants a written evaluation and a maintainer decision before code.  That is what this PR provides.

## Key answers to the issue

- **How do scores clear?**  A hard reset to zero every `CONSENSUS_SCHEDULE_CHANGE_SUB_DAGS = 60` committed leaders (`crates/node/src/primary.rs:33`), with no decay and no carry-over, plus an effective per-epoch reset when scores reload against the current committee.  Because `sub_dag_index` is the leader nonce and `2^31 mod 60 = 8`, the intra-epoch window is deterministic but **not epoch-aligned**: the reset phase drifts 8 commits per epoch.
- **Is byzantine impact meaningful?**  **No.**  Scoring is a purely additive liveness count (`+1` for referencing the previous committed leader at exactly `leader_round + 1`);  byzantine faults are rejected before the commit path, so a committed-but-byzantine node scores as fully honest.  The only consequence of a low score is temporarily losing leader turns via the swap table, and the hard reset erases even that every 60 commits.  Reputation detects downtime, not malice.

## Findings

Seven findings, ranked by relevance to the issue, each scoped for the determinism constraint above.  Full statement, evidence, determinism/fork analysis, and recommendation for each are in section 5 of the document.

- [x] **F1** Byzantine behavior has no reputation impact (core ask).  Decide whether reputation should carry a byzantine signal at all;  if so, source it only from in-DAG committed evidence, never local/network observations.
- [x] **F2** The hard reset gives reputation no long-term memory (core ask).  Either keep the reset and document it as intentional, or add a bounded integer cross-window carry so repeat offenders accumulate.
- [x] **F3** The reset window is not epoch-aligned (`2^31 mod 60 = 8`), and naive cross-epoch carry is unsafe because the committee key set changes (would trip the `total_authorities == committee.size` assert and halt).
- [x] **F4** The two consensus-critical constants (`CONSENSUS_SCHEDULE_CHANGE_SUB_DAGS`, `DEFAULT_BAD_NODES_STAKE_THRESHOLD`) live in code, not genesis;  two builds with different values silently fork.  Moving them to genesis config is a safe prerequisite for tuning the reset.
- [x] **F5** `swap()` uses `StdRng`, whose algorithm is not guaranteed stable across `rand` major versions;  a dependency bump could fork validators.  Pin to a versioned PRNG (`rand_chacha::ChaCha20Rng`).
- [x] **F6** Classification math is integer and the one float (`sqrt`) is IEEE correctly-rounded and safe, but any future decay/penalty using a transcendental would not be bit-portable.  Keep the math integer-only.
- [x] **F7** No test can guard a #942 change today: no penalty path exists to test, `test_leader_schedule_from_store` is a failing TODO, and the serde roundtrip only exercises all-zero scores even though scores are in the digest.  Land guarding tests first.

## Threat model / determinism

The document's section 6 records the guardrails any future change MUST satisfy, all driven by the digest-inclusion fact: committed-DAG input only (no wall clock, no libp2p peer score, no local liveness);  integer arithmetic only except IEEE-correctly-rounded ops;  a reproducible version-pinned PRNG;  deterministic ordering over sorted collections;  coordinated rollout for any cadence/threshold/formula change;  and returning errors rather than asserting on legitimately-reachable states so a bad change cannot convert into a chain halt.

## Scope

- **In scope:** the evaluation and an accurate reference for the current mechanism.
- **Out of scope:** any behavior change.  F1 and F2 are design decisions for the maintainers to make first;  F4 and F7 are safe, independently-shippable prerequisites that can each be their own PR once a direction is chosen.  This PR intentionally ships none of them.

## Testing

Docs-only change: no Rust source is touched, so fmt, clippy, and the workspace build are unaffected relative to `origin/main`.  In place of a code test, every technical assertion in the document was verified against the code:

- Scoring, reset predicate, `final_of_schedule`, and `sub_dag_index = leader.nonce()` at `crates/consensus/primary/src/consensus/bullshark.rs:57-104,218`.
- Nonce layout `((epoch as u64) << 32) | round` at `crates/types/src/primary/header.rs:179`;  `2^31 mod 60 = 8` computed directly.
- Classification, std-dev gating, and `swap()` at `crates/consensus/primary/src/consensus/leader_schedule.rs:62-230`.
- Reputation hashed into the sub-dag digest at `crates/types/src/primary/output.rs:467`.
- Constants at `crates/node/src/primary.rs:33` and `crates/types/src/primary/mod.rs:32`.
- The F7 test-coverage claims: `reset_consensus_scores_on_every_schedule_change` (`bullshark_tests.rs:959`), `test_leader_swap_table` (`leader_schedule_tests.rs:13`), the failing `test_leader_schedule_from_store` with its `// TODO: this test is failing` marker (`leader_schedule_tests.rs:408,410`), and the all-zero-score serde roundtrip (`serde_tests.rs:62,73`).

### DIFF
--- a/crates/consensus/primary/src/consensus/reputation.md
+++ b/crates/consensus/primary/src/consensus/reputation.md
@@ -1,0 +1,418 @@
+# Consensus Reputation Schedule — Evaluation & Invariants Reference
+
+This document is the single source of truth for the telcoin-network **consensus**
+reputation system: how scores are earned, how they clear, and how they affect
+leader election. It also records the evaluation requested by
+[issue #942](https://github.com/Telcoin-Association/telcoin-network/issues/942)
+("Reputation Schedule - Evaluate periodic reset"): whether the periodic reset is
+appropriate given current committee/epoch semantics, and whether byzantine
+behavior has a meaningful reputation impact.
+
+Every mechanism below is stated as an assertion with a source citation. Treat
+each entry as something to verify against the code.
+
+> Scope note: this covers **consensus reputation** (`ReputationScores`), which is
+> committed on-chain and drives leader scheduling. It is a distinct system from
+> the libp2p **peer** reputation/penalty score
+> (`crates/network-libp2p/src/peers/`, documented in `peers/penalty.md`), which
+> is a local, per-node, decaying network-health score and never feeds consensus.
+> Section 2.5 explains why the two must stay separate.
+
+## 0. The one constraint that governs everything: determinism
+
+Consensus reputation is **committed state**. The final scores are hashed into the
+`CommittedSubDag` digest:
+
+```rust
+hasher.update(encode(&self.inner.reputation_scores).as_ref());
+```
+
+`crates/types/src/primary/output.rs:467` (field at `:301`, accessor at `:452-453`).
+
+Because the scores are in the digest, and because the swap table derived from
+them determines which node leads each round, **every honest validator must
+compute byte-identical scores and byte-identical good/bad sets from the same
+committed DAG, or the network forks.** This is not a soft performance property;
+it is a safety property. It is the hard boundary on every option discussed in
+Section 5:
+
+- Reputation must be a pure, deterministic function of the **committed DAG**
+  only. No wall clock, no local network observations, no libp2p peer score.
+- No non-bit-portable floating point. Integer arithmetic only, except where a
+  float operation is IEEE-mandated to be correctly rounded (see Section 6).
+- The tuning constants are compile-time and fork-requiring; changing them
+  requires a coordinated upgrade of all validators (see Section 2.2 and F4).
+
+## 1. What the score measures
+
+Consensus reputation is a **purely additive liveness/participation count**. There
+is no negative scoring, no decay, and no penalty anywhere in the consensus path.
+
+The one and only event that grants score, from
+`Bullshark::resolve_reputation_score`
+(`crates/consensus/primary/src/consensus/bullshark.rs:57-104`):
+
+> When a sub-dag is committed, the code looks back at the **previous** committed
+> leader (its digest and round). For every certificate in the committed sequence
+> that is at exactly `leader_round + 1` **and** lists that previous leader's
+> digest among its `header().parents()`, the certificate's author receives `+1`.
+
+```rust
+if let Some(last_committed_sub_dag) = state.last_committed_sub_dag.as_ref() {
+    let leader_digest = last_committed_sub_dag.leader().digest();
+    let leader_round  = last_committed_sub_dag.leader().round();
+    for certificate in committed_sequence.iter().filter(|c| c.round() == leader_round + 1) {
+        if certificate.header().parents().contains(&leader_digest) {
+            reputation_score.add_score(certificate.origin(), 1);
+        }
+    }
+}
+```
+
+`bullshark.rs:81-89`. `add_score` is the only mutator of a score and does only
+`*val += score` on a `u64` (`crates/types/src/primary/reputation.rs:30-36`).
+
+Consequences of this definition:
+
+| Property | Detail | Source |
+| --- | --- | --- |
+| Additive only | `u64` scores, monotonically non-decreasing within a window; no subtract/decay/penalty exists | `reputation.rs:30-36` |
+| Max +1 per commit | Exactly one previous leader per commit; only its round-`+1` supporters earn | `bullshark.rs:81-89` |
+| Exact-round only | Only certificates at `leader_round + 1` count; transitive/late support earns nothing | `bullshark.rs:84` |
+| No first-commit score | Reward loop is guarded on `last_committed_sub_dag.is_some()` | `bullshark.rs:81` |
+| Semantics | High score = consistently produced a timely certificate that referenced the committed leader (liveness/connectivity). Low score = down / poorly connected / non-participating | `leader_schedule.rs:74-100` |
+
+**What the score does NOT capture:** byzantine correctness. Equivocation, invalid
+headers, and signature faults are rejected by certificate/DAG validation *before*
+the commit path, so a certificate that reaches `resolve_reputation_score` is
+already treated as honest. The score cannot distinguish a correct validator from
+a byzantine one that still produces well-formed certificates. The only fault it
+reflects is non-participation, and only implicitly (a missing `+1`).
+
+## 2. How the score clears (the periodic reset)
+
+### 2.1 The window
+
+Scores are **hard-reset to zero** at the start of every schedule window. In
+`resolve_reputation_score`:
+
+```rust
+let mut reputation_score =
+    if (sub_dag_index / 2).is_multiple_of(self.num_sub_dags_per_schedule as u64) {
+        ReputationScores::new(&self.committee)        // hard reset to all-zero
+    } else if let Some(last) = state.last_committed_sub_dag.as_ref() {
+        last.reputation_scores().clone()              // carry forward within the window
+    } else {
+        ReputationScores::new(&self.committee)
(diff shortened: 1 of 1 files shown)


crates/consensus/primary/src/consensus/reputation.md
```diff
@@ -0,0 +1,418 @@
...
+must be integer arithmetic (F6), and cross-**epoch** carry is unsafe as written
+(F3) because the committee key set changes.
+
+### F3 — The reset window is not epoch-aligned
+
+**Statement.** Because `sub_dag_index` is the leader nonce and `2^31 mod 60 = 8`,
+the reset phase shifts 8 commit-indices per epoch (Section 2.2). Windows drift
+relative to epoch starts. Separately, naive cross-epoch carry of scores is unsafe:
+the first sub-dag of a new epoch generally does not satisfy the reset predicate,
+so if prior-epoch scores were cloned into a changed committee the
```
sstanfield, line 279: This should not happen and is a bug if it is.  Bullshark is intended to be reset at each epoch, i.e. each epoch is basically a new chain so nothing bull shark related should be carried over including reputation.  If it is then we need to do a review of epoch/bullshark code.

crates/consensus/primary/src/consensus/reputation.md
```diff
@@ -0,0 +1,418 @@
...
+
+**Evidence.** `primary.rs:28-33,88-93`, `types/primary/mod.rs:32`.
+
+**Recommendation.** Move both into genesis/committee config so their values are
+part of coordinated, auditable state rather than binary-dependent. This is a
+prerequisite for ever tuning the reset (F2/F3) safely. It is behavior-preserving
+if the migrated defaults equal the current constants, but it still touches a
+consensus-critical path and warrants its own PR and tests.
+
+### F5 — `swap()` PRNG is an implicit cross-version consensus dependency
```
sstanfield, line 310: This seems like a serious footgun we need to address.

crates/consensus/primary/src/consensus/reputation.md
```diff
@@ -0,0 +1,418 @@
...
+## 5. Findings and recommendations
+
+Ranked by relevance to issue #942. Each finding is scoped for the determinism
+constraint in Section 0. None of these are implemented by this document; it is an
+evaluation to inform a maintainer decision, because every option that changes
+scoring or reset cadence is a coordinated fork.
+
+### F1 — Byzantine behavior has no reputation impact (core of #942)
+
+**Statement.** Reputation measures liveness only. Byzantine faults are filtered
```
sstanfield, line 231: This is (I am almost certain) from the bull shark paper and is in fact just to measure liveness not malice.  Personally I think this is fine, we have other ways of dealing with byzantine validators and this needs to be simple and deterministic across BFTs.  In short it is just supposed to minimize the times a flaky committee member gets selected as the leader to improve overall performance.

sstanfield (approved):

grantkee (approved):

MavenRain:
Thanks for the careful read, Steve. Taking your three comments in order of consequence, epoch carry first since you flagged it as a possible bug.

**(B) Cross-epoch carry (F3, line 279): nothing carries over today. Verified, no bug.**

You wrote: "This should not happen and is a bug if it is." I traced every path and it is not happening; your intent ("each epoch is basically a new chain") holds in the code:

- `Bullshark` and `LeaderSchedule` are built fresh once per epoch from the current epoch's committee in `spawn_consensus` (node/src/primary.rs:80-93).
- `Consensus::spawn` rebuilds `ConsensusState` each epoch and scopes recovery to `consensus_config.epoch()`; the sub-dag lookup carries the literal comment `// ignore previous epochs` (consensus/state.rs:289-304).
- The store reads are hard-scoped to that epoch's own pack: `latest_consensus_header_from_pack`, `read_last_committed`, and `read_latest_commit_with_final_reputation_scores` return data only when `pack.epoch() == epoch` or from that exact epoch's static pack, else None/empty (storage/consensus.rs:1023-1068).
- `Bullshark` holds no score state; `resolve_reputation_score` derives scores from `state.last_committed_sub_dag`, which is `None` at epoch start, so the fresh `ReputationScores::new(&self.committee)` branch runs (bullshark.rs:70-77).

So no epoch/bullshark review is needed. The sentence you flagged is a subjunctive guardrail ("if prior-epoch scores *were* cloned ... *would* panic" at the `assert_eq!`, bullshark.rs:101) aimed at any future naive carry implementation, not a claim about current behavior; Section 4 of the doc (lines 216-218) states the per-epoch reset explicitly. Since the sentence misread easily enough to trip you, I have rewritten it to say outright, with the citations above, that the precondition is unreachable today. Rather than push that here and have the ruleset dismiss your approval, I am carrying it in the follow-up PR described below, so this one can merge exactly as you approved it at 6d3c2d54.

**(A) Liveness-only scoring (F1, line 231): agreed, and now recorded as by-design.**

The code matches your description exactly: the only score mutation is the additive +1 for round leader_round+1 certificates referencing the previous leader (bullshark.rs:81-89), `add_score` has no penalty or decay path (types/primary/reputation.rs:30-36), the sole behavioral consumer of scores is the leader swap (leader_schedule.rs:209-230), and equivocation is rejected at DAG insertion before scoring ever runs (consensus/state.rs:151-156). F1 was written as an open question needing maintainer sign-off, and your comment supplies it. I have written a short disposition note under F1 quoting your comment (qualifiers intact: "I am almost certain" on the Bullshark-paper provenance, "Personally I think this is fine") so F1 no longer reads as an open defect; it lands with the same follow-up PR.

**(C) StdRng footgun (F5, line 310): confirmed, follow-up PR incoming.**

Verified end to end: rand 0.9.2's own docs mark `StdRng` "Non-portable: any future library version may replace the algorithm" (rand-0.9.2/src/rngs/std.rs:16-22; it currently wraps ChaCha12). The seed is only `leader_round` (leader_schedule.rs:211-213), and the swap result selects the leader certificate that gets hashed into the committed `ConsensusDigest` (leader() at leader_schedule.rs:285-304; output.rs:377 asserts the leader is the last certificate, output.rs:460-471 hashes every header digest plus the reputation scores). So a silent algorithm change on a routine dep bump is an immediate digest fork, not a slow drift. Good news: pinning `rand_chacha::ChaCha12Rng::from_seed` matches what `StdRng` wraps today, so the pin is behavior-preserving with zero coordinated change, and `rand_chacha` is already a direct workspace dep (Cargo.toml:199-200). I will open that as a separate PR with a golden test (fixed seed maps to a fixed good-node index) and fold the two doc updates above into it, leaving this PR untouched.

PR #987, opened 2026-07-22
fix(execution): commit the finalized/safe markers atomically

grantkee:
- `RethEnv::finish_executing_output` now persists the finalized/safe block numbers in the same database transaction as the blocks (derived from the last saved block), closing the crash window where a restart could observe the persisted tip ahead of the persisted marker.
- `RethEnv::finalize_block` no longer writes the database — it only updates the in-memory finalized/safe watches and prunes in-memory canonical state, since the DB write now happens earlier alongside the blocks.
- `RethEnv::heal_finalized_to_persisted_tip` is retained as defense-in-depth for databases written by pre-fix versions, and now rewrites the persisted marker rows directly instead of relying on `finalize_block` for the DB half.
- Added `plant_finalized_marker` test helper to construct a pre-fix-style lagging database directly, since `finalize_block` no longer writes the DB and can't be (ab)used to fabricate that state.
- Added `finalized_marker_commits_atomically_with_blocks`, asserting the persisted marker equals the persisted tip immediately after commit and that the heal takes its no-op path.
- Updated comments/docs across `payload_builder.rs`, `exex/context.rs`, `exex/lib.rs`, `engine/inner.rs`, `manager/node.rs`, and `error.rs` to describe the atomic-commit behavior and reframe the heal as legacy-database defense-in-depth rather than an expected steady-state repair.

### DIFF
--- a/crates/engine/src/payload_builder.rs
+++ b/crates/engine/src/payload_builder.rs
@@ -192,7 +192,8 @@ pub fn execute_consensus_output(
         executed_blocks,
         Some((leader_round, consensus_num_hash, engine_update_tx)),
     )?;
-    // remove blocks from memory and stores them in the database
+    // update the in-memory finalized/safe watches and prune persisted blocks from memory
+    // (the database markers committed atomically with the blocks above)
     reth_env.finalize_block(canonical_header.clone())?;
 
     // return new canonical header for next engine task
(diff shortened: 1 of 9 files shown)


MavenRain (approved):
LGTM

sstanfield (approved):

PR #988, opened 2026-07-23
exex: preserve FinishedHeight reporting on the replay_and_subscribe path

### LINKED ISSUE #981: exex: an ExEx built on `replay_and_subscribe` can never report `FinishedHeight`, and one such ExEx pins the node-wide `min_finished_height` to `None` forever (silently disables future pruning)
**Class:** latent correctness/robustness gap in the ExEx pruning-coordination signal.  Confirmed mechanism, no consumer today, so nothing is broken on a running node.  The value is as a guard-rail to land *before* a pruner is wired to `min_finished_height`, not a defect affecting the node today.

## Summary

The ExEx manager aggregates a node-wide "minimum durably-finished height" across all ExExes, intended as the floor a future pruner would respect so it never removes block data an ExEx still needs to replay (`crates/exex/src/event.rs:16-19`).  Two design choices interact so that this floor collapses to `None` (pruning disabled) as soon as a single ExEx uses the recommended convenience path:

1. `replay_and_subscribe` consumes `self` and **drops the event sender** (`crates/exex/src/context.rs:178`), and `report_finished_height` is the only writer of `FinishedHeight` and sends on exactly that sender (`context.rs:113-115`).  So an ExEx built on `replay_and_subscribe` is *structurally* incapable of ever reporting a height;  its `ExExHandle::finished_height` stays `None` for the life of the node.

2. `recompute_min` folds the per-ExEx heights with `let height = h.finished_height?;` (`crates/exex/src/manager.rs:371-383`), so a single `None` short-circuits the whole fold and `send_replace(None)` publishes a node-wide minimum of `None`.

Combined: one ExEx on `replay_and_subscribe` pins the published `min_finished_height` to `None` for *every* ExEx, permanently.  The failure is fail-safe (a `None` floor means "prune nothing", so no data is ever lost) but fail-silent (a pruner keyed on this signal would simply never prune, node-wide, with no error and no log), and it is triggered by the exact path the docs recommend for stateful ExExes (`context.rs:150-157`).

- `min_finished_height()` has **zero callers** in the workspace: the only references to the receiver are its own accessor and the plumbing (`manager.rs:125,143,155,382,402,409-410`).  Nothing consumes the floor yet.
- The handle that carries the receiver is **intentionally dropped at its only construction site** (`crates/node/src/manager/node.rs:963-973`):

  ```rust
  // NOTE: `_handle` exposes the minimum finished height across ExExes for
  // future pruning coordination. TN currently runs in archive mode (no
  // pruning), so there is no consumer yet and the handle is intentionally
  // dropped. See `tn_exex::TnExExEvent::FinishedHeight`.
  let (manager, _handle) = tn_exex::TnExExManager::new( /* ... */ );
  ```

- `FinishedHeight` is documented as "currently informational only ... TN runs in archive mode (pruning disabled), so nothing consumes it yet" (`event.rs:16-19`).

So `min_finished_height` is not merely uncalled, it is currently unreachable at runtime.  The bug bites the day someone wires `_handle` into a pruner.  At that point the very first stateful ExEx that follows the recommended `replay_and_subscribe` pattern silently turns the new pruner into a no-op, and the symptom (disk never shrinks) points nowhere near the ExEx layer.  Landing the fix, or at minimum the documentation, before that consumer exists removes a trap from the future pruner's path.

## The defect

**Locations:** `crates/exex/src/context.rs:113-115` (`report_finished_height`, the only writer), `context.rs:174-178` (`replay_and_subscribe` drops the sender), `crates/exex/src/manager.rs:371-383` (`recompute_min` short-circuits to `None`).

**`report_finished_height` is the only writer, and it sends on `self.events`** (`context.rs:113-115`):

```rust
pub fn report_finished_height(&self, height: BlockNumber) {
    let _ = self.events.try_send(TnExExEvent::FinishedHeight(height));
}
```

**`replay_and_subscribe` consumes `self` and drops `events`** (`context.rs:174-178`):

```rust
pub fn replay_and_subscribe(
    self,
    start_block: BlockNumber,
) -> impl futures::Stream<Item = eyre::Result<TnExExNotification>> {
    let TnExExContext { notifications, reth_env, events: _, consensus_chain: _ } = self;
    // ...
```

The `events: _` bind drops the sender.  After this call the ExEx holds only the combined stream, so `report_finished_height` is unreachable for that ExEx.  This local consequence is documented at `context.rs:169-173`;  what is not documented is the *systemic* consequence below.

**`recompute_min` lets one `None` pin the global minimum** (`manager.rs:371-383`):

```rust
fn recompute_min(&mut self) {
    // Minimum finished height: None if any ExEx hasn't reported yet.
    let min_height = self
        .exexes
        .iter()
        .try_fold(None, |acc: Option<BlockNumber>, h| {
            let height = h.finished_height?;                       // one None -> whole fold is None
            Some(Some(acc.map_or(height, |a: BlockNumber| a.min(height))))
        })
        .flatten();

    self.min_finished_height_tx.send_replace(min_height);
}
```

`finished_height` is only ever set from a `FinishedHeight` event (`manager.rs:358-366`), and a `replay_and_subscribe` ExEx can never emit one, so its handle's `finished_height` is a permanent `None`, and every `recompute_min` folds the global minimum back to `None`.

The comment on the `try_fold` ("`None` until the first commit") describes the intended startup transient: the floor is `None` until *every* ExEx has reported at least once.  The unintended part is that a structurally-non-reporting ExEx makes that transient permanent and node-wide.

## Trigger conditions

1. A future pruner is wired to `min_finished_height` (that is, `_handle` at `node.rs:967` is retained and consulted).
2. At least one ExEx initializes via `replay_and_subscribe` (the documented convenience path for stateful ExExes, `context.rs:148-157`).

Given both, the pruner's floor is `None` for the life of the node and pruning never runs.  Note that the shipped example ExEx deliberately avoids `replay_and_subscribe` for exactly this reason (`examples/exex-indexer/src/lib.rs:27-34`), which is evidence that the sharp edge is already understood locally but is currently worked around by convention rather than removed.

## Root cause

`replay_and_subscribe` trades away the ability to report progress (it drops the sender to move the notification stream out of `self`), but `recompute_min` treats "this ExEx has not reported yet" and "this ExEx can never report" identically, as `None`.  A single ExEx that has permanently opted out of reporting therefore gates the floor for all ExExes.

## Suggested fix

The naive workarounds are both subtly unsafe for a *pruner* consumer, and the issue should call that out rather than paper over it:

- **"Exclude non-reporters from the minimum"** is unsafe for `replay_and_subscribe` specifically.  That path *does* consume history (it replays from `start_block`), so its true floor is its own last durably-indexed height.  Excluding it lets the pruner delete blocks below the floor set by the *other* ExExes, which can be *above* this ExEx's last durable point, so on restart the ExEx replays from `L+1` and finds the blocks it needs already pruned.  That is data loss, the exact thing the floor exists to prevent.
- **"Auto-report the delivered height from inside `replay_and_subscribe`"** is also unsafe, because delivered `!=` durably processed.  `FinishedHeight` means the ExEx has *durably persisted* its work up to `height`;  only the ExEx knows when that is true.  Reporting the delivered height would over-claim durability and let the pruner delete data the ExEx has not yet durably indexed, again a crash-gap.

The sound resolutions, in order of preference:

1. **Preserve the ExEx's ability to report on the convenience path.**  Do not drop the sender in `replay_and_subscribe`;  give the combined-stream API a way for the ExEx to still call `report_finished_height` after it durably persists (for example, return the sender alongside the stream, or expose a small reporter handle).  This keeps the correct semantics (the ExEx reports its own durable height) and leaves `recompute_min` correct as written, with no special-casing.

2. **Make "opted out of pruning coordination" an explicit, safe state**, distinct from "not yet reported".  If an ExEx genuinely never participates in pruning *and* never needs old data, model that as a first-class handle state so `recompute_min` can exclude it deliberately.  This must not be applied to `replay_and_subscribe` (which needs history), so it is only correct in combination with (1) or with a hard rule that `replay_and_subscribe` ExExes are never marked opted-out.

3. **Minimum acceptable (documentation only):**  state in the `replay_and_subscribe` doc comment that using it makes the node's `min_finished_height` permanently `None`, and that a pruner-integrated node must initialize stateful ExExes via `replay_from` (which retains the sender) rather than `replay_and_subscribe`.  Also update the `recompute_min` comment to say the `None` floor is permanent, not merely a startup transient, in the presence of any non-reporting ExEx.  This removes the surprise even if the code is left as-is.

Whichever route is chosen, the acceptance property is the same: a single ExEx must not be able to silently and permanently disable the node-wide pruning floor by choosing the recommended catch-up path.

## Suggested test design

- **Unit test on `recompute_min` semantics:**  construct a manager with two ExEx handles, drive `FinishedHeight` on one but never the other, and assert the published `min_finished_height` matches the intended contract for a non-reporting ExEx (under fix (1): a reporting ExEx's height is published once every *reporting* ExEx has reported;  under fix (3): assert and document that it stays `None`, locking in the known behavior).
- **Regression test for the convenience path (fix (1)):**  build an ExEx via `replay_and_subscribe`, have it call `report_finished_height` after processing, and assert the manager observes the height and the floor advances.  This test fails today because the sender is dropped, which is the regression the fix closes.
- **Confirm-by-mutation:**  the `recompute_min` test must be seen to fail if the `try_fold`'s `?` short-circuit is changed, so it genuinely pins the fold semantics rather than passing vacuously.

## Acceptance criteria

- A stateful ExEx initialized via the recommended convenience path can report `FinishedHeight` (or the docs state unambiguously that it cannot and that pruner-integrated nodes must use `replay_from`).
- One non-reporting ExEx cannot pin the node-wide `min_finished_height` to `None` for ExExes that do report, or the permanence of that behavior is documented at both the `replay_and_subscribe` and `recompute_min` sites.
- A test pins the `recompute_min` contract and is shown to fail under mutation of the short-circuit.

MavenRain:
Closes #981.

## Summary

The ExEx manager aggregates a node-wide "minimum durably-finished height" across all ExExes (`min_finished_height`), intended as the floor a future pruner will respect so it never removes block data an ExEx still needs to replay. Today that signal has no runtime consumer (TN runs in archive mode, and the manager handle is intentionally dropped at `crates/node/src/manager/node.rs`), so nothing is broken on a running node. This change lands a guard-rail *before* a pruner is wired to that floor: it removes a trap on the future pruner's path where a single ExEx could silently and permanently disable pruning node-wide.

The trap: an ExEx built on the recommended convenience path (`TnExExContext::replay_and_subscribe`) was *structurally* unable to report a finished height, and one such ExEx pins the aggregated floor to `None` (pruning disabled) for every ExEx, forever. The failure is fail-safe (a `None` floor means "prune nothing", so no data is ever lost) but fail-silent (a pruner keyed on this signal would simply never prune, node-wide, with no error and no log), and it is triggered by the exact path the docs recommend for stateful ExExes.

## Root cause

Two design choices interacted:

1. `replay_and_subscribe` consumes `self` and **dropped the event sender** (`events: _`). `report_finished_height` is the only writer of `FinishedHeight` and sends on exactly that sender, so an ExEx built on `replay_and_subscribe` could never report a height. Its `ExExHandle::finished_height` stayed `None` for the life of the node.
2. `Router::recompute_min` folds the per-ExEx heights with `let height = h.finished_height?;`, so a single `None` short-circuits the whole fold and publishes a node-wide minimum of `None`.

Combined: one ExEx on `replay_and_subscribe` had a permanently-`None` finished height, and `recompute_min` folded the global floor back to `None` on every recomputation.

## Fix

Preserve the ExEx's ability to report on the convenience path (the issue's preferred resolution), and document the residual fail-safe behavior at both sites.

- **`replay_and_subscribe` now returns `(FinishedHeightReporter, ReplaySubscribeStream)`** instead of dropping the event sender. The sender is carried out in a small `FinishedHeightReporter` handle whose `report_finished_height` has identical semantics to the context method (non-blocking `try_send`, latest-wins). A stateful ExEx catching up on this path can now report durable progress, so it is no longer structurally forced to `None`. Because the reporter can only be built from the moved-out `events` sender, "the sender survives the move" is now enforced by the type system rather than by a comment.
- **`recompute_min` is unchanged.** Its `?` short-circuit is the correct fail-safe: the manager cannot know how far a silent ExEx has durably processed, so any unreported ExEx must hold the floor at `None` ("prune nothing"). What was undocumented is that this is *permanent* while any ExEx never reports, not merely a startup transient. That permanence is now documented at both the `replay_and_subscribe` and `recompute_min` sites, plus on `TnExExManagerHandle::min_finished_height`, per the issue's acceptance criteria.

The two naive alternatives are called out in the code as unsafe, because they are:

- *Excluding non-reporters from the minimum* is unsafe for `replay_and_subscribe` specifically. That path consumes history (it replays from `start_block`), so its true floor is its own last durably-indexed height. Excluding it lets a pruner delete blocks below the floor set by the *other* ExExes, and on restart this ExEx replays into already-pruned data. That is the exact data loss the floor exists to prevent.
- *Auto-reporting the delivered height from inside `replay_and_subscribe`* is unsafe because delivered is not durably-processed. Only the ExEx knows when its work is durably persisted; reporting the delivered height would over-claim durability and let a pruner delete data the ExEx has not yet indexed.

So the sound fix is to keep every stateful ExEx *able* to report (it now can, on every catch-up path) rather than to guess a floor for one that stays silent.

## Changes

- [x] `crates/exex/src/context.rs`: add `FinishedHeightReporter` (public, `Clone` + `Debug`) and the `ReplaySubscribeStream` boxed-stream alias; `replay_and_subscribe` returns `(FinishedHeightReporter, ReplaySubscribeStream)` and carries the event sender out instead of dropping it. Boxed because `impl Trait` cannot be nested in a tuple return (E0562); the one allocation happens once, at ExEx startup.
- [x] `crates/exex/src/context.rs`: rewrite the `replay_and_subscribe` doc (a `# Progress reporting` section), the `events` field doc, and the runnable doc example to show the reporter, and to state that an ExEx which never reports pins the node-wide floor to `None`.
- [x] `crates/exex/src/manager.rs`: document that the `None` floor is permanent (not a startup transient) while any ExEx never reports, on both `recompute_min` and `TnExExManagerHandle::min_finished_height`. No behavior change.
- [x] `crates/exex/src/lib.rs`: export `FinishedHeightReporter` and `ReplaySubscribeStream`.
- [x] `examples/exex-indexer/src/lib.rs`: correct the "Why not `replay_and_subscribe`?" note (it now *can* report progress via the returned reporter; the remaining reason to drive `replay_from` directly is re-replay-to-reconcile-lag control).
- [x] Tests: `recompute_min` contract (one non-reporting ExEx pins the floor to `None`; the floor tracks the minimum once all report) and a `FinishedHeightReporter` delivery test.

## Testing

Pinned toolchain (1.94), OOM-safe `-j 2`:

- `cargo +nightly fmt -- --check` . . . clean.
- `cargo +1.94 clippy -p tn-exex -p exex-indexer --all-targets -- -D warnings` . . . clean.
- `cargo +1.94 test -p tn-exex` . . . 10 passed, 0 failed (the doc example is `ignore`, as before).
- `CARGO_TARGET_DIR=<isolated> cargo +1.94 test --no-run --workspace -j 2` . . . workspace test-build (attest proxy in an isolated target dir, per the shared-target false-green class).

**Confirm-by-mutation.** The new `recompute_min` tests were seen to fail when the fold's `?` short-circuit is removed (i.e. when non-reporters are silently excluded, the unsafe alternative): with the short-circuit gone, the "one non-reporting ExEx pins the floor to `None`" assertion publishes a non-`None` floor and the tests fail. Restored, all pass. The tests therefore pin the fold contract rather than passing vacuously.

## Scope and notes

- Latent by design: `min_finished_height` has zero runtime consumers today, so this changes no running-node behavior. It removes the sharp edge before a pruner is wired to the floor.
- Out of scope: wiring an actual pruner to `min_finished_height` (the handle is still intentionally dropped at the construction site).
- `replay_and_subscribe` had zero in-repo callers, so the return-type change breaks nothing in the workspace.


sstanfield (dismissed):

sstanfield (dismissed):

grantkee (approved):
Good stuff

PR #990, opened 2026-07-23
test(e2e): speed up the three heaviest ignored tests (#978)

### LINKED ISSUE #978: Speed up three ignored e2e tests (epoch_sync + two delayed restarts) by >=25% each
## Summary

Three of the heaviest `#[ignore]` e2e tests can each be cut by at least 25% of their solo wall-clock without losing coverage:

| Test | Solo baseline (TN_BIN_PATH set) | Proposed cut | Resulting |
| --- | --- | --- | --- |
| `epochs::test_epoch_sync` | ~105s | ~35s (33%) | ~70s |
| `restarts::test_restarts_delayed` | ~132s | ~39s (30%) | ~93s |
| `restarts::test_restarts_lagged_delayed` | ~112s | ~39s (35%) | ~73s |

Baselines are solo runs with the prebuilt binary passed via `TN_BIN_PATH` (i.e.  `make test-restarts` / `make test-e2e`), so the escargot build tax is excluded.  All line numbers are against current `main` (`0c59c15b`);  re-confirm before landing.

Runtime here is protocol wall-clock, not contention or build tax.  The dominant terms are: for `test_epoch_sync`, walking 15 real epoch boundaries (~78s, 73%);  for both restart tests, a deliberate fixed 60s downtime floor that keeps the killed node outside the GC window (~45% and ~54% respectively).

## Already landed (do not redo)

These earlier levers are already in `main` and should not be re-proposed:

- e2e group serialized at `nextest max-threads = 1` (`.config/nextest.toml:22-23`), so cross-test contention is gone.
- `EPOCH_DURATION = 5` (`epochs.rs:41`) with the deadline floors already applied, e.g.  `(EPOCH_DURATION * 3).max(30)` (`epochs.rs:150`) and `(EPOCH_DURATION * 6).max(60)` (`epochs.rs:404`).
- Prebuilt binary reused via `TN_BIN_PATH` in `make test-restarts` / `make test-e2e` (`Makefile:110-125`) and `etc/test-and-attest.sh:103-106`.
- Dead-node balance probes at `retries = 0` (`restarts.rs:82,150,341-344`).
- Lagged post-send settle is event-driven via `get_balance_above_with_retry` (`restarts.rs:161-162`), not a fixed 5s sleep.
- BLS keygen KDF floored to 1 round (`INSECURE_TEST_KDF_ROUNDS`, `crates/e2e-tests/src/lib.rs:54`).

### 1.  `test_epoch_sync`: cut the epoch-walk iteration counts 5 -> 3 (saves ~30s)

`test_epoch_sync_inner` walks epochs three times: `loop_epochs(0,5)`, `loop_epochs(5,5)`, `loop_epochs(10,5)` (`epochs.rs:359,370,378`).  Each iteration blocks one full 5s epoch boundary via `wait_until` on `getCurrentEpochInfo` (`epochs.rs:303-328`), so 15 boundaries x ~5s ~= 78s, the dominant term.  The `start` argument is cosmetic;  each call drives exactly `iterations` boundaries.

Change each count 5 -> 3 (`loop_epochs(0,3)`, `loop_epochs(3,3)`, `loop_epochs(6,3)`).  This keeps: at least one pre-kill certified epoch, three missed epochs while validator-3 is down (so the state-sync back-fill path is genuinely exercised), and three post-restart epochs proving the 4-of-5 committee kept advancing and the node rejoined.  Saves 6 epochs x ~5s ~= 30s.

Coverage: the verification loop (`epochs.rs:388`) still asserts an epoch pack file plus a cert-verified `EpochRecord` for every epoch `0..=latest` on all endpoints including the resynced validator-3.  `latest` shrinks with the reduced count, so verification breadth drops from ~15 to ~9 epochs;  the sync property itself is fully preserved.  If breadth matters, a conservative variant cuts only the first two loops (`:359`, `:370`) and leaves `:378` at 5 (saves ~20s), since `:378`'s return value defines the verification range.

### 2.  `test_epoch_sync`: trim the first-epoch settle sleep 6s -> 1s (saves ~4.5s)

`epochs.rs:356` sleeps a fixed `EPOCH_DURATION + 1 = 6s` pad after the RPC-up gate before the assertion loop.  `loop_epochs` immediately after takes a fresh `getCurrentEpochInfo` baseline (`:309`) and waits for the next epoch change with its own 20s deadline (`:314`), so it self-synchronizes regardless of where in epoch 0 execution starts.

Reduce to `sleep(1)` (keep a small genesis-settle guard;  do not bare-remove, so `getCurrentEpochInfo` does not race an unsettled epoch-0 registry).  Saves ~5s when the first epoch has already settled.

### 3.  Both restart tests: shrink the deliberate downtime via a restart-test gc_depth (saves ~35s each) [key restart lever]

Both delayed restart tests call `do_restarts(60,...)` (`restarts.rs:461,470`), which floors the killed-node downtime at a fixed 60s inside `wait_for_downtime` (`restarts.rs:249-276`, floor at `:260`, cap at `:262`).  The 60s exists so the killed CVV falls outside the GC window and demotes to `NodeMode::CvvInactive` (the follow / state-sync recovery path the tests exist to exercise, per docstrings at `restarts.rs:455-456,464-465`).

The demotion threshold is `parameters().gc_depth - 10` DAG rounds (`handler.rs:171,175`): the node demotes when `effective_exec_round + (gc_depth - 10) < round`.  `gc_depth` defaults to `MAX_GC_DEPTH = 50` (`crates/types/src/primary/mod.rs:45`;  `node.rs:303-305`), so the threshold is ~40 rounds.  The e2e testnet runs at `--max-header-delay-ms 1000` (`crates/e2e-tests/src/lib.rs:193-196`), i.e.  ~1 round/s idle, so ~40 rounds ~= 40s and the 60s floor carries ~20s of slack.

The downtime is set by the GC threshold, which is set by `gc_depth`.  Lower `gc_depth` for the restart tests and the required downtime drops proportionally while the exact same follow-path is exercised:

- Set a restart-test `gc_depth ~= 25` (effective `25 - 10 = 15` rounds ~= 15s, plus ~10s margin).
- Lower the `wait_for_downtime` floor from 60 to ~25s to match (`do_restarts(25,...)` at `restarts.rs:461,470`;  the cap at `:262` shrinks with it).

`Parameters::validate` (`node.rs:384-398`) enforces only ceilings (`gc_depth <= MAX_GC_DEPTH`), no minimum, so `gc_depth = 25` is legal.  Lowering `gc_depth` strictly reduces the `gc_depth x max_header_num_of_batches` batch-reconstruction bound (`node.rs:377-397`), so packs stay reconstructible;  the only effect is a narrower catch-up window, which is exactly the demotion path under test.  Keep `gc_depth >= ~15` (the demotion math uses `saturating_sub(10)` and `.max(6)`, so below ~15 the effective window collapses toward 0).

Required plumbing: there is no `--gc-depth` seam today (`GenesisArgs` in `crates/telcoin-network-cli/src/genesis/mod.rs` exposes `--max-header-delay-ms`, `--min-header-delay-ms`, `--epoch-duration-in-secs`, but not gc_depth).  Add a `--gc-depth` arg mirroring `--max-header-delay-ms` (`mod.rs:105-108` and the override at `:299-304`), thread it through `crates/e2e-tests/src/lib.rs:193-197`, and pass it from the restart tests.  This is a small, well-scoped production add.  (A test-only alternative is to rewrite the generated parameters file after the genesis ceremony and before node start, avoiding a production change at the cost of a hack.)

Coverage: `wait_for_downtime` still blocks until a live peer has advanced past the kill point AND the floor elapses, so the killed node still misses more than `gc_depth - 10` rounds and still takes the CvvInactive follow path.  No existing assertion is weakened.  See finding 6 for making that path assertion explicit.

Saving: floor 60 -> ~25s ~= 35s per test.  That is ~26% of `test_restarts_delayed` and ~31% of `test_restarts_lagged_delayed` from this lever alone.

### 4.  Both restart tests: delete the fixed 2s "advancing" pauses (saves ~2-4s each)

`restarts.rs:191` (in `run_restart_tests2`, on both restart paths) is a fixed `sleep(2)` fully subsumed by the `wait_for_block` synchronization loop that immediately follows (`restarts.rs:197-205`), which polls every node up to the max EL block before `test_blocks_same` runs.  Deleting it is zero-risk.  The phase-1 pauses at `:39` (`run_restart_tests1`) and `:131` (`run_restart_tests_lagged1`) can also go for ~2s more each at low risk, since `network_advancing` already blocks until all RPCs answer and `get_block` retries a lagging node 30x1s (`common.rs:216-221`).  Do not convert these to event-waits: under skip-empty-execution there is no idle EL-block event to await, and a consensus-round wait floors at ~1 round/s, costing the same ~2s plus an extra RPC.

Note: `restarts.rs:384` is in `run_observer_tests` (only `test_restarts_observer`, not a target) and earns no credit;  the large `sleep(5)`/`sleep(15)` at `:615,:625` belong to `test_observer_reconnect_after_pause` and are unreachable from the three target tests.

### 5.  Both restart tests: make `wait_for_downtime` adaptive via round-delta (optional, pairs with 3)

Replace the fixed floor in `wait_for_downtime` (`restarts.rs:249-276`) with a wait until an observed leader-round delta reaches `gc_depth - 10 + margin`.  The DAG round is available from `tn_latestConsensusHeader` (`ConsensusHeader.sub_dag`, `crates/types/src/primary/block.rs:13-32`;  leader round via `CommittedSubDag::leader_round`, `output.rs:427-429`), or the coarser `number` field that the existing helper already reads (`common.rs:457-478`) can serve as a proxy.  This is test-only.  On its own it saves little at the default threshold;  combined with finding 3's lower `gc_depth` it returns right at threshold plus margin instead of over-waiting, and it is robust to cadence variation.  Retain a fail-safe time cap.

### 6.  Both restart tests: surface `NodeMode` and assert `CvvInactive` (deferred hardening)

No RPC exposes `NodeMode` today (`RpcNodeInfo` in `crates/execution/tn-rpc/src/lib.rs:16-41` carries only chain_id/version/name/keys/addresses), so neither restart test asserts which recovery path ran;  they assert only balance/block behavior, which passes on either path.  Add a `node_mode` field to `RpcNodeInfo` (plus the `EngineToPrimary`/handler wiring) and assert the killed node was observed `CvvInactive` after restart.  This is a production change and is the prerequisite that makes any downtime reduction provably safe rather than silently path-dependent.  Recommend landing findings 1-5 first, then this as follow-up hardening.

### 7.  Minor poll-cadence trims (saves ~1-3s total)

`test_epoch_sync`'s hand-rolled startup RPC-ready poll sleeps 1s (`epochs.rs:347`);  tighten to ~250ms (or route through `wait_until` at 10ms).  The restart tests already detect RPC-up via `network_advancing` -> `wait_until_blocking` at 10ms (`common.rs:301`), so this residual is epoch_sync-only.  Low payoff, credited honestly at well under 1s each;  include only alongside the larger findings.

## Acceptance criteria

- All three tests pass their existing assertions unchanged (epoch pack + cert-verified `EpochRecord` for every epoch `0..=latest` on all endpoints;  restart balance/block sync;  killed node still takes the follow/catch-up path).
- The killed node still misses more than `gc_depth - 10` rounds while down (so the CvvInactive path is still exercised);  after finding 6 lands, assert this explicitly.
- No epoch-count or missed-epoch coverage is silently trivialized (the down-window still spans at least one certified epoch boundary;  validator-3 still back-fills a real gap).
- Demonstrate the per-test cuts above on a serial `make test-restarts` / `make test-e2e` run (escargot tax excluded).

## Notes and ordering

- Apply order for the restart lever: add the `--gc-depth` seam (finding 3 plumbing) before lowering the `do_restarts` floor, so the floor is never cut below the demotion threshold.
- The 60 -> 25s floor cut is only valid together with `gc_depth ~= 25`;  do not lower the floor at the default `gc_depth = 50` (that would risk leaving the node inside the GC window on a slow-cadence run and silently flip it to the live-rejoin path).
- Cadence is pinned near 1 round/s by the serialized e2e group (`max-threads = 1`) plus `--max-header-delay-ms 1000`, which is what keeps the margin math honest.
- Findings 1, 2, 4, 7 are verified coverage-safe from source-cadence analysis.  Finding 3 is verified sound (no `validate` floor, reconstruction bound only shrinks) with the one caveat that it needs the small `--gc-depth` add.  Findings 5 and 6 are optional/deferred.

MavenRain:
Closes #978.

## Summary

Cuts the solo wall-clock of the three heaviest `#[ignore]` e2e tests without weakening any assertion, and adds the `NodeMode`-over-RPC hardening the issue asked for so the delayed restart tests now *prove* the killed node took the follow / catch-up path instead of only inferring it from balances.

Baselines are solo runs with the prebuilt binary (`TN_BIN_PATH` set), so the escargot build tax is excluded. Per-test targets (from the issue):

| Test | Before | Target |
| --- | --- | --- |
| `epochs::test_epoch_sync` | ~105s | ~70s |
| `restarts::test_restarts_delayed` | ~132s | ~93s |
| `restarts::test_restarts_lagged_delayed` | ~112s | ~73s |

All sites were re-confirmed against `origin/main` and the branch is rebased onto `b6b3189e` (the issue was written against `0c59c15b`, so several line numbers had drifted and finding 2 had already landed via #951). The rebase pulls in #952 (epoch-entry reads) and #963 / #964 (anti-equivocation durability); the tree recompiles clean on the new base.

## Changes

Numbered to the issue's findings.

1. **`crates/e2e-tests/tests/it/epochs.rs` (finding 1)** . . . cut the three `loop_epochs` walks from 5 to 3 boundaries each (`(0,3)` / `(3,3)` / `(6,3)`), saving ~30s. Each call drives exactly `iterations` epoch boundaries; the `start` argument only labels a log line. The verification loop still asserts an epoch pack file plus a cert-verified `EpochRecord` plus the executed final block for every epoch `0..=latest` on all endpoints including the resynced validator-3, and `latest` derives from the on-chain `epochId`, so the range self-adjusts to the shorter walk and cannot collapse.
2. **Finding 2 . . . already in `main`.** #951 removed the fixed `EPOCH_DURATION + 1` settle sleep; nothing to redo.
3. **`crates/telcoin-network-cli/src/genesis/mod.rs`, `crates/e2e-tests/src/lib.rs`, `restarts.rs` (finding 3, the key restart lever)** . . . added a `--gc-depth` genesis arg mirroring the existing `--max-batch-delay-ms` seam, which overrides `Parameters.gc_depth` before `Parameters::validate` (validate enforces only `gc_depth <= MAX_GC_DEPTH`, no minimum). Threaded it through `config_local_testnet_inner` into a new `config_local_testnet_with_gc_depth`, so the delayed restart tests run at `RESTART_TEST_GC_DEPTH = 25` (demotion threshold `gc_depth - 10` = 15 DAG rounds) with the deliberate-downtime floor lowered 60s to 25s. Saves ~35s each. Leaving `--gc-depth` unset preserves the exact prior `MAX_GC_DEPTH` default for every non-test caller.
4. **`restarts.rs` (finding 4)** . . . deleted three fixed `sleep(2)` "advancing" pauses, each subsumed by the synchronization that immediately follows it (`network_advancing` already blocks until all RPCs answer; `run_restart_tests2`'s `max_block` / `wait_for_block` loop; `get_block`'s 30x1s retry). The observer-test sleeps were left untouched.
5. **Finding 5 . . . folded into the `wait_for_downtime` hardening below** rather than shipped as a standalone speedup (the helper was already adaptive-with-floor by design).
6. **`crates/types/src/primary/node_mode.rs`, `consensus_bus.rs`, `tn-rpc`, `crates/node`, `restarts.rs` (finding 6, hardening)** . . . surfaced the consensus `NodeMode` over RPC as `tn_nodeMode`. Added a serializable `tn_types::NodeMode` plus an exhaustive `From<primary::NodeMode>` at the boundary (tn-rpc depends only on tn-types, so the wire type lives there and the consensus enum is left untouched). The RPC reads the mode live via `consensus_bus.current_node_mode()`, not a snapshot. The two delayed restart tests now assert the restarted node (index 2) was observed `CvvInactive` during catch-up, so a silent flip to the live-rejoin path fails the test instead of passing on balances alone.
7. **`epochs.rs` (finding 7)** . . . tightened the `test_epoch_sync` startup RPC-ready poll from 1s to 250ms.

**`restarts.rs` `wait_for_downtime` (findings 5 + 6, robustness).** The finding-6 assertion would otherwise flake on a badly overloaded runner: demotion fires on a DAG-round gap `> gc_depth - 10` = 15, but a fixed 25s downtime only guarantees that at the nominal ~500ms/round cadence (~50 rounds); past ~1.67s/round it covers fewer than 15 rounds and the node never demotes. So `wait_for_downtime` now returns only once BOTH the wall-clock floor has elapsed AND the live peers' consensus-header number has climbed by `>= (gc_depth - 10) + 3` = 18 (each header commits one sub-dag whose leader round strictly exceeds its predecessor's, so a header delta of D guarantees `>= D` rounds past the kill point). This is finding 5's round-delta idea applied for robustness, keeping the wall-clock floor as the fast-path gate at normal cadence and the `min_secs * 2 + 30` cap as the wedged-network fail-safe. The short rejoin test (`min_secs = 2`, gap `1`) is unchanged and intentionally stays inside the GC window.

## Threat model and correctness

- **No coverage trivialized.** The down-window still spans multiple certified epoch boundaries, validator-3 still back-fills a real gap, and no assertion was removed. The `CvvInactive` assertion is non-vacuous: it returns `Err` when the follow path never ran (it does not default-pass), and it is gated to `delay_secs >= RESTART_TEST_DOWNTIME_SECS` so only the two delayed tests carry it (the short rejoin test never demotes and is excluded).
- **Demotion is guaranteed, not assumed.** The round-gap gate derives the required peer advance from the real parameter (`RESTART_TEST_GC_DEPTH - 10`), not a guessed wall-clock number, so the killed node is provably outside the GC window before the assertion runs regardless of CI cadence. No attacker is involved; this is a determinism fix for a serial `#[ignore]` suite.
- **The seam changes no production default.** `--gc-depth` is opt-in; unset yields the pre-existing `MAX_GC_DEPTH` behavior, and the override runs before `validate`, so an out-of-range value is rejected.
- One correction to the issue's rationale: lowering `gc_depth` does **not** shrink the pack-reconstruction bound. That bound (`crates/storage/src/consensus_pack.rs`) is computed from the `MAX_GC_DEPTH` constant, not the configured value, so packs stay reconstructible; the only effect is the intended narrower catch-up window.

## Acceptance criteria

- [x] All three tests keep their existing assertions (epoch pack + cert-verified `EpochRecord` + executed final block for every epoch `0..=latest`; restart balance / block sync; killed node still takes the follow path).
- [x] The killed node still misses more than `gc_depth - 10` rounds while down, now asserted explicitly via `tn_nodeMode == CvvInactive`.
- [x] No epoch-count or missed-epoch coverage silently trivialized.
- [ ] Per-test cuts demonstrated on a serial `make test-restarts` / `make test-e2e` run (pre-merge, see Verification).

## Verification

Static gates, all on the pinned toolchain:

- `cargo +1.94 check -p telcoin-network-cli -p e2e-tests --tests --all-features` . . . clean, 0 warnings.
- `cargo +1.94 check -p tn-rpc -p tn-node --all-features` . . . clean.
- `cargo +1.94 clippy --all-targets -p tn-types -p tn-primary -p tn-rpc -p tn-node -p telcoin-network-cli -p e2e-tests --all-features` . . . no warnings from any changed file.

Remaining pre-merge (heavy, run serially): `make attest` (adds the full `test --no-run --workspace` build on 1.94) and the timed `make test-restarts` / `make test-e2e` runs to record the actual per-test before / after and to confirm-by-mutation that the new `CvvInactive` assertion fails when the follow path is suppressed.


sstanfield (approved):
Looks good as long as your sure it still asserts the same stuff.

Issue #954, opened 2026-07-19
Consensus params `gc_depth` / `max_header_num_of_batches` are node-local with only an upper-ceiling check: add lower/cross-field bounds and promote the operative values to protocol constants (follow-up to #902)


MavenRain:
## Summary

#902 made the consensus-output batch-count bound symmetric across writer and reader and added `Parameters::validate()` to reject a node configured *above* the protocol ceilings (`gc_depth <= MAX_GC_DEPTH`, `max_header_num_of_batches <= MAX_HEADER_NUM_OF_BATCHES`).  In review of that PR, @grantkee raised two follow-ups at `crates/config/src/node.rs:384` ("this could also validate min `gc_depth > 0` ... we should probably move these to protocol constants as well, since deviation between committee validators would cause issues").  This issue substantiates both against the PR-head code and records two refinements the review comment did not have in front of it.

1. **The validation is one-sided.**  `Parameters::validate()` bounds only the maximum.  A `gc_depth` of `0` (and any small value) and a `max_header_num_of_batches` of `0` both pass validation today and both break the node, so a lower bound is warranted.  Refinement: `> 0` is not a sufficient floor for `gc_depth`.  The consensus code contains a hardcoded `gc_depth.saturating_sub(10)` activity buffer, so any `gc_depth <= 10` zeroes that window and drives the node inactive during ordinary pipelining.  The real floor is `> 10`, not `> 0`.

2. **Both values are purely node-local, and divergence across a committee is a latent safety fault, not merely a nuisance.**  Neither value is carried in the committee, genesis, or on-chain state;  each node reads its own `parameters.yaml`, and #902's `validate()` compares each node's values only against the static ceilings, never against its peers'.  Two validators in one committee can therefore silently run different values.  Refinement: the two parameters are *not* symmetric in how badly divergence hurts.  A divergent `gc_depth` is a genuine (conditional) state-root fork;  a divergent `max_header_num_of_batches` within `[0, 10]` is already benign *because* #902 made `Header::validate` check the protocol constant rather than the peer's configured value.  So the safety-critical one to protocolize is `gc_depth`.

All line numbers below are against the #902 head (branch `fix/896-batch-count-symmetry`, merge `23a9663a`).

## Background: what #902 added and where the nit lands

`Parameters::validate()` (`crates/config/src/node.rs:384-398`) is upper-bound only:

```rust
pub fn validate(&self) -> eyre::Result<()> {
    eyre::ensure!(
[...]

PR #994, opened 2026-07-23
fix(config): reject consensus params below their operational floors (#954)

MavenRain:
Addresses #954. Follow-up to #902.

## Summary

Consensus parameters `gc_depth` and `max_header_num_of_batches` were validated against their protocol ceilings only (added in #902), so a node could still be configured *below* the values at which consensus keeps working. This PR adds the missing lower floors and the cross-field coupling to `Parameters`, promotes the `gc_depth` activity buffer to a named `tn-types` constant so the floor and the buffer can never drift apart, and enforces the new floors at node startup while leaving DAG test fixtures free to use small `gc_depth` values.

This is the interim, independently shippable step called for in #954. Eliminating per-node tunability of `gc_depth` entirely (the "identical across validators by construction" acceptance criterion) is a larger consensus-wiring change and is left as a follow-up. see the scope note below.

## Problem

`Parameters::validate` enforced only upper ceilings (`gc_depth <= MAX_GC_DEPTH`, `max_header_num_of_batches <= MAX_HEADER_NUM_OF_BATCHES`). Three lower/cross-field footguns remained, none of which require an attacker: a single misconfigured `parameters.yaml` is enough.

1. **`gc_depth` has no floor, and `> 0` is not enough.** The consensus network handler computes a node's activity window as `gc_depth - 10` (`crates/consensus/primary/src/network/handler.rs:187`, `gc_depth.saturating_sub(10)`). Any configured `gc_depth <= 10` saturates that window to `0`, so `outside_gc_window` (`handler.rs:191`) fires as soon as the node is a single round behind, and the node steps itself down to an inactive CVV during normal operation. The true floor is `> 10`, coupled to that buffer, not `> 0`.

2. **`max_header_num_of_batches` has no floor.** With `0`, the proposer's per-header drain `self.digests.len().min(self.max_header_num_of_batches)` is `0` (`crates/consensus/primary/src/proposer.rs:598`), so `self.digests.drain(..0)` removes nothing while workers keep appending (`proposer.rs:706`). Headers keep being sealed on the delay timers but carry empty payloads: a permanent transaction stall plus unbounded digest-backlog growth. The floor is `>= 1`.

3. **`header_num_of_batches_threshold` is uncoupled from `max_header_num_of_batches`.** The proposer seals a header once it holds `threshold` digests (`proposer.rs:760`) but includes at most `max_header_num_of_batches` of them. A `0` threshold seals empty headers on the fast path, and a threshold above the max is a configuration that can never be satisfied as intended. The invariant is `1 <= header_num_of_batches_threshold <= max_header_num_of_batches`.

## Changes

- [x] `crates/types/src/primary/mod.rs`: add `pub const GC_ACTIVITY_BUFFER: Round = 10`, documenting that the handler's activity window is `gc_depth - GC_ACTIVITY_BUFFER` and that `Parameters::validate_operational_floors` keeps `gc_depth` above it.
- [x] `network/handler.rs`: replace the hardcoded `gc_depth.saturating_sub(10)` with `gc_depth.saturating_sub(tn_types::GC_ACTIVITY_BUFFER)` so the buffer the handler subtracts and the floor the config validates are one shared constant.
- [x] `crates/config/src/node.rs`: split validation into `validate` (the existing protocol **ceilings**, still run for every constructor) and a new `validate_operational_floors` (the **lower floors** and the `threshold <= max` cross-field invariant).
- [x] `crates/config/src/consensus.rs`: enforce `validate_operational_floors` at the production entry points `ConsensusConfig::new` and `new_for_epoch`; the shared test-facing `new_with_committee` is untouched, so it still enforces the ceilings for everyone while the DAG-test fixture constructor keeps its small `gc_depth` (the "retain test-only overrides" the issue asks for).
- [x] Unit tests for every new floor and the cross-field invariant, plus accept-tests for the minimal valid configuration and the default parameters.

## Threat model and design notes

- **No attacker required.** All three failures are reachable from a single validator's own `parameters.yaml`; the point of the change is to reject them loudly at startup instead of degrading silently once running.
- **The floor is derived, not guessed.** `gc_depth`'s floor is `> GC_ACTIVITY_BUFFER`, the exact constant the handler subtracts, so the two are one value and cannot drift on a future edit to the buffer.
- **Producer/reader symmetry from #902 is preserved.** The ceiling checks (`gc_depth <= MAX_GC_DEPTH`, `max_header_num_of_batches <= MAX_HEADER_NUM_OF_BATCHES`) still run for every constructor, and the reader-side `Header::validate` cap against `MAX_HEADER_NUM_OF_BATCHES` (`crates/types/src/primary/header.rs:121`) is unchanged. `max_header_num_of_batches` divergence between validators is therefore already benign (the reader checks the constant, not the peer's config); `gc_depth` is the divergence-sensitive one.
- **Scope note (why the interim).** A validated `[11, 50]` range bounds `gc_depth` but does not make it identical across validators, so it does not by itself close the divergence vector in #954's finding 3: `order_dag` (`crates/consensus/primary/src/consensus/utils.rs:13,22`) descends only to `leader.round() - gc_depth`, so two validators running different in-range `gc_depth` values can still commit different sub-DAGs once uncommitted ancestry outruns the smaller depth. Closing that "by construction" means making `gc_depth` a non-tunable protocol constant in the operative path (retaining test-only overrides), which is a larger consensus-wiring change. That divergence is latent (it needs sustained asynchrony or leader-stall, and the default `gc_depth = 50` masks it), so this PR ships the startup-validation hardening first and leaves de-tunabilization as a tracked follow-up.

## Testing

Pinned toolchain (`rust-toolchain.toml` = 1.94), OOM-capped at `-j 2`.

- `cargo +nightly fmt -- --check`: clean.
- `cargo +1.94 clippy -p tn-types -p tn-config -p tn-primary --all-targets`: clean (workspace `missing_docs = warn` covers `tn-types` / `tn-config`; every new item is documented).
- `cargo +1.94 test -p tn-config --lib node::test`: 10 passed. Each new reject-test was **confirmed by mutation**: removing the four floor `ensure!`s makes exactly the four floor/cross-field reject-tests fail while the accept-tests and ceiling-tests stay green; the floors were then restored.
- `cargo +1.94 test -p tn-primary --lib test_gc_round_update_during_fetch test_gc_pending_certs`: 2 passed. These two fixtures deliberately use `gc_depth = 5`; they build via the test-only constructor, which skips the operational floors, so they keep working unchanged.
- Attest-proxy build in an isolated target dir (the `--workspace` shape `make attest` runs, not a `-p` subset): `CARGO_TARGET_DIR=.tn-954-iso cargo +1.94 test --no-run --workspace -j 2`: green, every workspace test binary compiled on the pinned toolchain.

### DIFF
--- a/crates/config/src/consensus.rs
+++ b/crates/config/src/consensus.rs
@@ -57,6 +57,12 @@ where
         network_config: NetworkConfig,
         next_committee_keys: Vec<BlsPublicKey>,
     ) -> eyre::Result<Self> {
+        // Production entry point: enforce the operational floors that the shared, test-facing
+        // `new_with_committee` deliberately skips so DAG test fixtures may use small `gc_depth`
+        // values. The protocol ceilings are still validated for every constructor inside
+        // `new_with_committee`.
+        config.parameters.validate_operational_floors()?;
+
         // load committee from file
         let committee: Committee =
             Config::load_from_path_or_default(tn_datadir.committee_path(), ConfigFmt::YAML)?;
@@ -105,6 +111,11 @@ where
         network_config: NetworkConfig,
         next_committee_keys: Vec<BlsPublicKey>,
     ) -> eyre::Result<Self> {
+        // Production entry point: enforce the operational floors (see
+        // [`Parameters::validate_operational_floors`]); the shared test-facing constructor skips
+        // them so DAG fixtures may use small `gc_depth` values.
+        config.parameters.validate_operational_floors()?;
+
         Self::new_with_committee(
             config,
             node_storage,
--- a/crates/config/src/node.rs
+++ b/crates/config/src/node.rs
@@ -371,16 +371,20 @@ impl Default for Parameters {
 }
 
 impl Parameters {
-    /// Validate protocol-critical parameters against the ceilings the rest of the protocol depends
-    /// on.
+    /// Validate the protocol **ceilings** every node (production or test fixture) must honor.
     ///
     /// `gc_depth` and `max_header_num_of_batches` together bound how many unique batches a single
     /// committed [`ConsensusOutput`](tn_types::ConsensusOutput) can reference. The consensus-pack
     /// reader derives its reconstruction bound from [`tn_types::MAX_GC_DEPTH`] and
     /// [`tn_types::MAX_HEADER_NUM_OF_BATCHES`], so a node configured above those ceilings could
     /// commit an output that no node (including itself, on restart) can later reconstruct.
-    /// Rejecting such a configuration at startup keeps the producing and reconstructing halves
-    /// in agreement.
+    /// Rejecting such a configuration keeps the producing and reconstructing halves in
+    /// agreement.
+    ///
+    /// These are safe to enforce everywhere, so this runs for every constructor. The lower
+    /// **operational floors** are enforced separately by
+    /// [`Parameters::validate_operational_floors`] at the production entry points only, so DAG
+    /// test fixtures may still use small `gc_depth` values.
     pub fn validate(&self) -> eyre::Result<()> {
         eyre::ensure!(
             self.gc_depth <= tn_types::MAX_GC_DEPTH,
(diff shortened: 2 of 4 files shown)


grantkee (approved):
LGTM - these are network-wide parameters than can cause issues if validators aren't consistent (ie - different max batches per header). We might consider reading this from a smart contract like `WorkerConfig`

sstanfield (approved):

MavenRain:
> LGTM - these are network-wide parameters than can cause issues if validators aren't consistent (ie - different max batches per header). We might consider reading this from a smart contract like `WorkerConfig`

Posted a comment on #954 raising this

PR #996, opened 2026-07-23
Guard ExEx manager against empty canonical chains (#991)

### LINKED ISSUE #991: exex: an empty canonical Chain would panic the ExEx manager in handle_canon_commit / dedup_chain_executed;  unreachable today, but non-emptiness is now only caller-implied in the payload builder, with no guard at the panic sites
**Class:** latent panic-safety / robustness gap in the ExEx canonical-commit path.  Confirmed mechanism, no reachable trigger today, so nothing is broken on a running node.  The value is as a defense-in-depth guard to land while the non-emptiness invariant is still enforced only by the shape of the payload builder's control flow, not as a defect affecting the node today.

## Summary

The ExEx manager consumes reth `CanonStateNotification`s and, on every canonical commit, reads the committed chain's block range to detect gaps before fanning the executed chain out to every registered ExEx.  Two consuming sites call reth `Chain` accessors that **panic on an empty chain**, with no emptiness guard at either site:

1. `handle_canon_commit` calls `new.range()` (which composes `Chain::first()` and `Chain::tip()`), then dereferences `*range.start()` / `*range.end()` (`crates/exex/src/manager.rs:342-355`).  The `Reorg` arm degrades into this same function (`manager.rs:327-335`).
2. `dedup_chain_executed`, the dedup wrapper on the replay-plus-live subscribe stream, reads `new.tip().number()` (`crates/exex/src/context.rs:230`).

reth `Chain::tip()` and `Chain::first()` both `.expect("Chain should have at least one block")` on an empty chain, and `Chain::range()` composes them, so it panics on empty as well (reth `v1.11.3`, `crates/evm/execution-types/src/chain.rs:255-256,246,269-271`).  An empty `Arc<Chain>` reaching either site therefore panics the ExEx manager task.

An empty commit cannot reach those sites today: the sole shipping producer of a canonical commit always broadcasts at least one block.  But that guarantee is now enforced **only** by an early-return in the payload builder's control flow, with no explicit `is_empty` / `.expect` guard at the producer or at either consuming site.  This report documents the panic surface and recommends a small, local guard so the invariant does not rest entirely on that control-flow shape.

The panic is unreachable on current code (see "Reachability" below): the only production caller that broadcasts a canonical commit always passes a non-empty block set.  So no empty `Chain` is delivered to `handle_canon_commit` or `dedup_chain_executed`, and neither panics in practice.

It is still worth landing a guard because the non-emptiness guarantee is fragile and undocumented at the sites that depend on it:

- There is **no explicit guard** anywhere on the path.  The producer `RethEnv::finish_executing_output` (`crates/tn-reth/src/lib.rs:933`) does not check `blocks.is_empty()`;  non-emptiness is established only by the caller's early-return at `crates/engine/src/payload_builder.rs:98`.
- The two consuming sites have **zero defense**: an empty chain arriving at `manager.rs:343` or `context.rs:230` panics immediately, with no log and no degrade.
- A future refactor that adds a new production caller of `finish_executing_output`, or that lets an empty `executed_blocks` slip past the early-return, would silently make the panic reachable.  The symptom (a manager task that vanished, or a whole node that shut down under `exex_critical`) would point nowhere near the payload builder that lost the invariant.

Landing a cheap, local guard at the consuming sites removes that trap before it can be reintroduced upstream, and it is consistent with the crate's existing degrade-not-panic posture at `manager.rs:317-321`, where the manager comment notes it "is spawned once and never respawned", so a panic there would kill the shared fan-out for every ExEx.

## The defect

**Locations:** `crates/exex/src/manager.rs:342-355` (`handle_canon_commit`, `range()` / `*start()` / `*end()`), `crates/exex/src/manager.rs:327-335` (the `Reorg` arm degrading into it), `crates/exex/src/context.rs:230` (`dedup_chain_executed`, `tip()`).

**`handle_canon_commit` reads the range with no emptiness guard** (`manager.rs:342-355`):

```rust
fn handle_canon_commit(&mut self, new: Arc<Chain>) {
    let range = new.range();
    let first = *range.start();
    canon_gap(&mut self.last_canon_tip, first, *range.end()).into_iter().for_each(|missed| {
        warn!(
            target: "exex::manager",
            missed,
            first,
            "canonical ChainExecuted gap detected; surfacing Lagged",
        );
        self.fan_out(&TnExExNotification::Lagged { missed });
    });
    self.fan_out(&TnExExNotification::ChainExecuted { new });
}
```

`new.range()` calls both `Chain::first()` and `Chain::tip()`, each of which panics on an empty chain, so the panic occurs on the first line of the function for an empty `new`.

**The `Reorg` arm funnels the new side into that same function** (`manager.rs:327-335`):

```rust
CanonStateNotification::Reorg { old, new } => {
    error!(
        target: "exex::manager",
        old_blocks = old.len(),
        new_blocks = new.len(),
        "unexpected Reorg for finalized state; degrading to commit of the new side",
    );
    self.handle_canon_commit(new);
}
```

(`old.len()` / `new.len()` are safe: reth `Chain::len()` returns `0` on an empty chain.  Only the subsequent `handle_canon_commit(new)` is exposed.)

**`dedup_chain_executed` reads the tip with no emptiness guard** (`context.rs:226-232`):

```rust
stream
    .scan(None::<BlockNumber>, |last_height, item| {
        // Borrow `item` only long enough to read the execution height.
        let height = match &item {
            Ok(TnExExNotification::ChainExecuted { new }) => Some(new.tip().number()),
            _ => None,
        };
```

`new` is `Arc<Chain>` (`crates/exex/src/notification.rs`, `ChainExecuted { new: Arc<Chain> }`), and `Chain::tip()` panics on empty.

**The reth accessors that panic** (`v1.11.3`, `crates/evm/execution-types/src/chain.rs`):

```rust
    /// Return true if chain is empty and has no blocks.
    pub fn is_empty(&self) -> bool {          // line 129 — the guard that exists but is never consulted
        self.blocks.is_empty()
    }

    /// Get the tip of the chain.
    ///
    /// # Panics
    ///
    /// If chain doesn't have any blocks.
    #[track_caller]
    pub fn tip(&self) -> &RecoveredBlock<N::Block> {          // line 255
        self.blocks.last_key_value().expect("Chain should have at least one block").1
    }

    /// Returns the range of block numbers in the chain.
    ///
    /// # Panics
    ///
    /// If chain doesn't have any blocks.
    pub fn range(&self) -> RangeInclusive<BlockNumber> {          // line 269
        self.first().header().number()..=self.tip().header().number()
    }
```

`Chain::is_empty()` exists (line 129) but is not consulted inside `tip()` / `first()` / `range()`.

## Reachability (why no empty chain arrives today)

The manager subscribes to `blockchain_provider.canonical_state_stream()` (`crates/tn-reth/src/lib.rs:709`).  On current code the only shipping producer of a `CanonStateNotification::Commit` is `RethEnv::finish_executing_output`, whose sole broadcast is `notify_canon_state` at `crates/tn-reth/src/lib.rs:987`:

```rust
        let chain_update = NewCanonicalChain::Commit { new: blocks };
        let canonical_head = chain_update.tip();          // lib.rs:960
        // ...
        let notification = chain_update.to_chain_notification();
        self.canonical_in_memory_state().notify_canon_state(notification);          // lib.rs:987
```

Two facts keep an empty commit from ever being delivered:

1. **The producer's only production caller always passes at least one block.**  `finish_executing_output` is called from `crates/engine/src/payload_builder.rs:191`.  The one path that would yield zero blocks (no batches and the epoch is not closing) returns early, before that call, and broadcasts nothing (`payload_builder.rs:84-101`):

   ```rust
   if batches.is_empty() {
       if !output.close_epoch() {
           // Skip execution entirely — no batches and epoch is not closing.
           // Leader count was already incremented above for rewards tracking.
           info!(target: "engine", "skipping execution for empty non-epoch-closing output");
           crate::metrics::ENGINE_METRICS.empty_outputs_skipped_total.increment(1);
           span.record("executed_blocks", "0");
           // Notify consensus that this round was processed (no block produced)
           engine_update_tx.blocking_send((leader_round, consensus_num_hash, None)).map_err(/* ... */)?;
           return Ok(canonical_header);
       }
       // Execute single empty block to close the epoch.  -> executed_blocks gets exactly one block
   ```

   The epoch-closing branch executes exactly one block, and the non-empty-batches path pushes at least one, so `executed_blocks` is always non-empty by the time `finish_executing_output` is reached.  All other callers are test-only (integration-test targets under `tests/`, a `#[cfg(test)] mod tests`, and the `#[cfg(any(feature = "test-utils", test))]`-gated `test_utils`), each passing a single-block vec.

2. **reth itself emits no independent commit in TN, and even an empty producer call would fault first.**  reth's own `notify_canon_state` paths (the engine tree) are dead code in this node: `EngineApiTreeHandler` is only a type alias and is never instantiated;  the reth blockchain-provider `notify_canon_state` calls are `#[tokio::test]`-only.  And even if an empty `blocks` reached `finish_executing_output`, `chain_update.tip()` at `lib.rs:960` operates on the same empty set and would panic in the producer, one line before `notify_canon_state`, so an empty commit is never broadcast.

The `Reorg` arm (`manager.rs:327`) is documented-unreachable under TN's immediate-finality consensus (reth should only ever emit `Commit`), and it degrades to `handle_canon_commit` rather than panicking on its own.

## Root cause

The non-emptiness of a committed `Chain` is a real invariant of the payload builder, but it is expressed only as control flow (an early-return skip path), never as a checked precondition.  The two consumers, `handle_canon_commit` and `dedup_chain_executed`, and the producer `finish_executing_output`, all rely on that invariant implicitly and call panic-on-empty reth accessors without a guard.  Nothing at the point of use, or in the producer, records or enforces "this chain is non-empty", so the property is one refactor away from silently becoming false with no local defense.

## Blast radius if it were reached

The ExEx manager runs for the lifetime of the node, spawned once and never respawned (`manager.rs:317-321`).  The spawn policy (`crates/node/src/manager/node.rs:975-988`) has two modes:

- **Default (`exex_critical = false`):** the manager runs as a non-critical task wrapped in `run_isolated_exex_future`, whose `catch_unwind` absorbs the panic and always resolves `Ok(())` (`crates/node/src/manager/exex.rs:15-36`).  A panic in `handle_canon_commit` therefore **leaves the node running but permanently ends fan-out to every ExEx**, silently, with no self-heal.
- **Operator opt-in (`exex_critical = true`):** the manager runs via `run_critical_exex_future`, which does not `catch_unwind` (`exex.rs:49-62`), so the same panic **shuts down the whole node**.

Either way, a single panic on the canonical-commit path is a global outcome for the ExEx subsystem (default) or the node (opt-in), triggered by one malformed input the code assumes cannot occur.

## Suggested fix

Add a local emptiness guard at the point of use, consistent with the crate's existing degrade-not-panic posture.  In `handle_canon_commit`, treat an empty commit as a no-op with a warning rather than reading its range:

```rust
fn handle_canon_commit(&mut self, new: Arc<Chain>) {
    // reth's Chain::range()/tip() panic on an empty chain; TN never commits one,
    // but the manager is spawned once and never respawned, so a stray empty commit
    // must not be allowed to kill fan-out (or, under exex_critical, the node).
    if new.is_empty() {
        warn!(target: "exex::manager", "received empty canonical commit; ignoring");
        return;
    }
    // ... existing body ...
}
```

Because the manager is the only live producer of `ChainExecuted`, guarding `handle_canon_commit` also protects the `dedup_chain_executed` `tip()` call for the live path;  the replay path builds each `Chain` from a single real block, so it is non-empty by construction.  If a guard is preferred directly at the dedup site as well, `keep_height` can be skipped for an empty chain (treat it as a pass-through with no height).

Alternatively, or in addition, document the non-emptiness precondition at `finish_executing_output` and assert it there (`debug_assert!(!blocks.is_empty(), ...)` plus an early return on release), so the invariant is stated where it is established rather than only where it is consumed.

## Suggested test design

- **Unit test on `handle_canon_commit` with an empty `Chain`:** construct a manager, drive a `CanonStateNotification::Commit { new }` whose `Chain` has no blocks, and assert the manager does not panic and fans out nothing (or a single benign marker), rather than aborting the task.
- **Confirm-by-mutation:** the test must be seen to fail if the `is_empty()` guard is removed (the manager task then panics), so it genuinely pins the guard rather than passing vacuously.
- **Optional producer-side assertion test:** if the `finish_executing_output` `debug_assert!` route is taken, add a debug-mode test that an empty `blocks` argument is rejected there.

## Acceptance criteria

- An empty canonical `Chain` delivered to the ExEx manager does not panic the manager task (default) or shut down the node (`exex_critical`);  it is ignored or degraded with a log.
- The non-emptiness precondition is either enforced at the consuming sites (`handle_canon_commit`, and by extension the live `dedup_chain_executed` path) or asserted and documented at the producer `finish_executing_output`.
- A test pins the empty-commit behavior and is shown to fail under mutation of the guard.

MavenRain:
Closes #991.

## Problem

The ExEx manager reaches reth `Chain` accessors that panic on an empty chain.  reth's `Chain::tip()`, `Chain::first()`, and `Chain::range()` each `.expect("Chain should have at least one block")`, so an empty `Chain` panics at two consumer sites:

- `crates/exex/src/manager.rs` `handle_canon_commit` calls `new.range()` (then dereferences `range.start()` / `range.end()`).
- `crates/exex/src/context.rs` `dedup_chain_executed` calls `new.tip().number()`.

This is latent, not a live crash today.  The sole live producer, `RethEnv::finish_executing_output`, never commits an empty chain because it early-returns before building one (`crates/engine/src/payload_builder.rs`), and the replay producer (`ReplayStream`) always yields single-block chains.  But that guarantee lives only in the engine's control flow, with no explicit check at either consumer.

No attacker is required: any future refactor that lets an empty commit through would panic the manager task.  Because the manager is spawned once and never respawned, that panic permanently disables the shared ExEx fan-out for every ExEx for the node's lifetime in the default mode, or crashes the whole node when an ExEx is registered with `exex_critical = true`.  `handle_canon_commit` is also the target of the degraded `Reorg` arm in `handle_canon`, so the guard covers that path as well.

## Fix

Guard both consumer sites independently (defense in depth), so neither relies on the producer's control-flow invariant:

- **`handle_canon_commit`** (`crates/exex/src/manager.rs`): ignore an empty commit (`if new.is_empty()` logs a `warn` and skips) before touching `new.range()`.  A non-empty commit takes the unchanged original path.
- **`dedup_chain_executed`** (`crates/exex/src/context.rs`): read the execution height through the non-panicking `blocks().last_key_value()` map instead of `Chain::tip()`, and drop an empty `ChainExecuted` (it carries no height and has nothing to deliver) rather than panic.  Every non-empty item keeps its previous keep / drop decision, and non-execution items (errors, `Lagged` markers, certificate / output signals) still pass through unchanged.
- Removed the now-unused `tn_types::BlockHeader` import (the trait supplied the `.number()` call that is gone).

The fix is confined to the two ExEx consumer sites and adds no cross-crate API surface.  The producer is left as is: its early-return already guarantees non-emptiness, and reth's own `Chain::new` carries a `debug_assert!(!blocks.is_empty())`, so the value of this change is at the consumers, where the panic actually lives.

## Testing

- `cargo +1.94 test -p tn-exex` . . . 9 tests pass.
- `cargo +1.94 clippy -p tn-exex --all-targets -- -D warnings` . . . clean.
- `cargo +nightly-2026-03-20 fmt -p tn-exex -- --check` . . . clean.

Two deterministic regression tests, each confirmed by mutation (reverting the guard makes the test panic with reth's `Chain should have at least one block`, observed to fail, then restored):

- `manager::tests::empty_canonical_commit_is_ignored_not_panicked`: drives an empty `Chain` through `handle_canon_commit`;  asserts nothing is fanned out and the gap tracker (`last_canon_tip`) does not advance.
- `context::tests::dedup_drops_empty_chain_without_panicking`: pushes an empty `ChainExecuted` followed by a `Lagged` marker through `dedup_chain_executed`;  asserts the empty item is dropped and the trailing marker still passes through.

The de-duplication decision itself (`keep_height`) keeps its existing in-isolation coverage (`keep_height_is_monotonic_and_dedups`), and the only new wiring is a single `blocks().last_key_value()` read that equals the pre-change `tip().number()` for any non-empty chain.  A dedup-level test over two real-height `ChainExecuted` items is deliberately left out: this crate carries no dev-dependencies and its tests avoid constructing execution types, so such a test would add `reth-ethereum-primitives` / `alloy-consensus` dev-deps to exercise one BTreeMap access.  Happy to add it if preferred.

### DIFF
--- a/crates/exex/src/context.rs
+++ b/crates/exex/src/context.rs
@@ -5,7 +5,7 @@ use futures::{stream, StreamExt};
 use std::pin::Pin;
 use tn_reth::RethEnv;
 use tn_storage::consensus::ConsensusChain;
-use tn_types::{BlockHeader as _, BlockNumber};
+use tn_types::BlockNumber;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 
@@ -225,16 +225,21 @@ fn dedup_chain_executed(
 
     stream
         .scan(None::<BlockNumber>, |last_height, item| {
-            // Borrow `item` only long enough to read the execution height.
-            let height = match &item {
-                Ok(TnExExNotification::ChainExecuted { new }) => Some(new.tip().number()),
-                _ => None,
+            // A `ChainExecuted` is de-duplicated by execution height, read through
+            // the non-panicking `blocks()` map: `Chain::tip()` panics on an empty
+            // chain ("Chain should have at least one block"). An empty chain (never
+            // produced today, but guarded defensively) carries no height and is
+            // dropped; it advances nothing and has nothing to deliver. Every other
+            // item (errors, `Lagged` markers, certificate/output signals) carries
+            // no height and always passes through.
+            let keep = match &item {
+                Ok(TnExExNotification::ChainExecuted { new }) => new
+                    .blocks()
+                    .last_key_value()
+                    .is_some_and(|(&height, _)| keep_height(last_height, height)),
+                _ => true,
             };
-            let keep = match height {
-                Some(h) => keep_height(last_height, h).then_some(item),
-                None => Some(item),
-            };
-            futures::future::ready(Some(keep))
+            futures::future::ready(Some(keep.then_some(item)))
         })
         .filter_map(futures::future::ready)
 }
@@ -271,5 +276,22 @@ mod tests {
         assert!(matches!(out[0], Ok(TnExExNotification::Lagged { missed: 1 })));
         assert!(out[1].is_err());
         assert!(matches!(out[2], Ok(TnExExNotification::Lagged { missed: 2 })));
+    }
+
+    #[tokio::test]
+    async fn dedup_drops_empty_chain_without_panicking() {
+        // A (defensively) empty `ChainExecuted` carries no height, and
+        // `Chain::tip()` panics on it. dedup must drop the empty item and keep
+        // draining the stream: the trailing `Lagged` marker still passes through.
+        // Reverting the guard to `new.tip().number()` makes this test panic.
+        use crate::Chain;
+        use std::sync::Arc;
+        let items: Vec<eyre::Result<TnExExNotification>> = vec![
+            Ok(TnExExNotification::ChainExecuted { new: Arc::new(Chain::default()) }),
+            Ok(TnExExNotification::Lagged { missed: 7 }),
+        ];
+        let out: Vec<_> = dedup_chain_executed(stream::iter(items)).collect().await;
+        assert_eq!(out.len(), 1);
+        assert!(matches!(out[0], Ok(TnExExNotification::Lagged { missed: 7 })));
     }
 }
--- a/crates/exex/src/manager.rs
+++ b/crates/exex/src/manager.rs
@@ -337,21 +337,35 @@ impl Router {
     }
 
     /// Handle a canonical commit: detect any block-number gap (reth drops canonical-stream lag
-    /// silently), surface it as `Lagged`, then deliver the executed chain. Shared by the `Commit`
-    /// and degraded `Reorg` arms.
+    /// silently), surface it as `Lagged`, then deliver the executed chain. An empty commit is
+    /// ignored (no gap check, no fan-out), since reth's `Chain` accessors panic on it. Shared by
+    /// the `Commit` and degraded `Reorg` arms.
     fn handle_canon_commit(&mut self, new: Arc<Chain>) {
-        let range = new.range();
-        let first = *range.start();
-        canon_gap(&mut self.last_canon_tip, first, *range.end()).into_iter().for_each(|missed| {
-            warn!(
-                target: "exex::manager",
-                missed,
-                first,
-                "canonical ChainExecuted gap detected; surfacing Lagged",
+        // Defense-in-depth: reth's `Chain::range()` (and `Chain::tip()`) panic on
+        // an empty chain ("Chain should have at least one block"). The sole live
+        // producer, `RethEnv::finish_executing_output`, never commits an empty
+        // chain: it early-returns before building one. That guarantee lives only
+        // in the engine's control flow, not in a check here, so guard it. A panic
+        // on this task would permanently disable fan-out for every ExEx (or, with
+        // `exex_critical` set, crash the node); an empty commit is ignored instead.
+        if new.is_empty() {
+            warn!(target: "exex::manager", "ignoring empty canonical commit");
+        } else {
+            let range = new.range();
+            let first = *range.start();
+            canon_gap(&mut self.last_canon_tip, first, *range.end()).into_iter().for_each(
+                |missed| {
+                    warn!(
+                        target: "exex::manager",
+                        missed,
+                        first,
+                        "canonical ChainExecuted gap detected; surfacing Lagged",
+                    );
+                    self.fan_out(&TnExExNotification::Lagged { missed });
+                },
             );
-            self.fan_out(&TnExExNotification::Lagged { missed });
-        });
-        self.fan_out(&TnExExNotification::ChainExecuted { new });
+            self.fan_out(&TnExExNotification::ChainExecuted { new });
+        }
     }
 
     /// Record an ExEx progress event, then recompute the minimum finished height.
@@ -505,6 +519,28 @@ mod tests {
         // Exactly one notification, no spurious Lagged marker.
         assert!(matches!(rx.try_recv(), Ok(TnExExNotification::Lagged { missed: FILLER })));
         assert!(rx.try_recv().is_err());
+    }
+
+    #[tokio::test]
+    async fn empty_canonical_commit_is_ignored_not_panicked() {
+        // reth's `Chain::range()`/`tip()` panic on an empty chain. An empty
+        // canonical commit must be dropped (no fan-out, no gap-tracker advance)
+        // rather than panic the manager task and disable ExEx fan-out for the
+        // node's lifetime. Reverting the `is_empty` guard makes this test panic
+        // with "Chain should have at least one block".
+        let (tx, mut rx) = mpsc::channel(4);
+        let (min_finished_height_tx, _min_rx) = watch::channel(None);
+        let mut router = Router {
+            exexes: vec![ExExHandle::new("test".to_string(), tx)],
+            min_finished_height_tx,
+            last_canon_tip: None,
+        };
+
+        router.handle_canon_commit(Arc::new(Chain::default()));
+
+        // Nothing fanned out, and the discontinuity tracker did not advance.
+        assert!(rx.try_recv().is_err());
+        assert_eq!(router.last_canon_tip, None);
     }
 
     #[test]


grantkee (approved):
This should be impossible, but this is a better production-ready solution

PR #997, opened 2026-07-23
exex: bound ExEx manager stream-merge starvation with a round-robin schedule

### LINKED ISSUE #992: exex: the manager's select_with_strategy merge has a constant PollNext::Left bias and no fairness primitive, so a sustained certificate/consensus feed can defer the canonical-state and ExEx-event streams on a consensus-following node
**Class:** fairness / scheduling gap in the ExEx manager's stream merge.  Confirmed mechanism, actively exercised on follower / inactive-CVV nodes, but bounded to no observable harm today by downstream factors (canonical-state gap recovery, an inert event consumer under archive mode, and non-critical ExEx spawning). The value is a small, local fairness primitive that makes the "no stream is starved" property explicit and hold under the changes already in flight (`#981` / `#988` build out the `FinishedHeight` reporting a future pruner would consume;  that consumer is not itself in flight), rather than a defect degrading the node today.

## Summary

The ExEx manager merges four input sources into one stream with a strict, fixed priority (`certificates > consensus output > canonical state > events`) built from three nested `futures::stream::select_with_strategy` calls, each biased by the same constant closure `prefer_left`, whose state type is `()` and which unconditionally returns `PollNext::Left` (`crates/exex/src/manager.rs:196-207,225-227`).

Because the strategy closure carries a `()` state it structurally cannot alternate: it returns `PollNext::Left` on every poll, forever. `select_with_strategy` still polls the lower-priority side whenever the chosen (higher-priority) side is `Pending`, so this is not a deadlock.  But whenever a higher-priority stream is *continuously ready*, the merge drains it in preference and defers the lower streams for as long as that readiness lasts.  There is no fairness primitive (no round-robin, no starvation bound) anywhere on the path.

The two high-priority streams (certificates, consensus output) are populated only on consensus-following nodes (Observer and inactive CVV), which is exactly where the ExEx manager runs, so the scheduling mechanism is genuinely exercised in production.  Under a sustained certificate or consensus burst, the canonical-state stream and the per-ExEx event stream can be deferred indefinitely relative to that burst.

Nothing is broken today because each deferrable stream is independently and deliberately protected (see below), not because the merge bounds starvation.  This report documents the scheduling gap and recommends swapping the constant bias for a round-robin strategy, which preserves the priority *ordering* as bounded throughput shares while guaranteeing no source is deferred forever.

## Why this is Low today (and why a fairness primitive is still worth landing)

The constant bias can genuinely defer the two lower-priority streams under sustained higher-priority load, but neither deferral is observable on current code:

- **Canonical state (`Canon`) is separately recovered.** Even if canonical-state notifications are deferred (or dropped by reth's silent broadcast lag), `handle_canon_commit` detects any block-number discontinuity via `canon_gap` and fans out a `TnExExNotification::Lagged { missed }` marker before the next `ChainExecuted`, so a stateful ExEx reconciles rather than carrying a silent hole (`crates/exex/src/manager.rs:342-354,444-456`). Late delivery, not lost delivery.
- **The lowest-priority `events` stream has no live consumer.** The per-ExEx `TnExExEvent` receivers carry `FinishedHeight` progress reports, aggregated by `recompute_min` into `min_finished_height`. That value is published on a watch channel whose only receiver (`TnExExManagerHandle::min_finished_height`) is intentionally dropped at the wiring site, because TN runs in archive mode with pruning disabled (`crates/node/src/manager/node.rs:963-968`; `crates/exex/src/event.rs:16-20`: "currently informational only ... nothing consumes it yet"). Deferring these events changes nothing an observer can see.
- **The ExEx manager is spawned non-critical by default.** `exex_critical` defaults to `false` (`crates/config/src/node.rs:46`), so even a fully stalled event stream cannot affect node liveness or consensus.

It is still worth landing a fairness primitive because the safety of this design rests entirely on those deliberate downstream properties, not on the merge itself:

- The design's own documentation states the strict-priority intent (`manager.rs:196-198`: "drains the higher-priority stream fully before the next"), but a constant `PollNext::Left` provides *no upper bound* on how long a lower stream is deferred.  Strict priority and bounded starvation are not the same property, and only the former is implemented.
- The one stream where deferral is currently harmless (`events`) is on a deliberate path toward becoming a live, consumed dependency: `#981` and `#988` are actively building out `FinishedHeight` reporting and the `min_finished_height` pruning floor.  The moment a pruner reads that floor and archive mode is disabled, the lowest-priority stream sits at the bottom of an unbounded-bias priority ladder (see "The forward-looking trap").
- A round-robin swap is a two-line change that keeps the intended priority ordering (as throughput shares) while removing the unbounded-deferral trap before it can matter.

## The mechanism

**Location:** `crates/exex/src/manager.rs:196-207` (the nested merge), `crates/exex/src/manager.rs:223-227` (the constant strategy).

The four sources are merged with three nested `select_with_strategy` calls, each using `prefer_left`:

```rust
// Biased priority certs > consensus > canon > events via a constant `PollNext::Left`, which
// drains the higher-priority stream fully before the next (matching the original sequential
// poll order, where each stream was drained to `Pending` before the next was polled).
let merged = select_with_strategy(
    certs,
    select_with_strategy(
        consensus,
        select_with_strategy(canon, events, prefer_left),
        prefer_left,
    ),
    prefer_left,
);
```

```rust
/// Always drains the higher-priority (left) stream first, giving the fixed source order
/// certs > consensus output > canonical state > events.
fn prefer_left(_: &mut ()) -> PollNext {
    PollNext::Left
}
```

The strategy's state type is `()`. `select_with_strategy` threads `&mut State` to the closure on every poll so a strategy *can* base its choice on prior choices (that is the whole point of the `State` parameter), but `prefer_left` ignores it and returns the same value unconditionally, so the choice never changes.

Per the `futures` `0.3.32` semantics (`select_with_strategy.rs`): the closure only decides which side to poll *first*;  if that side is `Pending`, the other side is still polled, and once either input completes the other is polled exclusively.  So a constant `PollNext::Left` does not block the lower streams outright.  What it removes is any guarantee that the lower streams make progress *while the higher one keeps yielding items*: under a continuously-ready left stream, the right stream is only ever reached when the left goes `Pending`.

The `events` source is itself a `select_all` over every ExEx's `TnExExEvent` receiver (`manager.rs:192-194`), and unlike the three consensus/canon sources it carries no `Closed` sentinel, so it can never terminate the fold;  it can only be deferred.

## Where the mechanism is actually exercised

The two high-priority streams are gated off the active-validator hot path, so they flow precisely on the nodes where the ExEx manager is live:

- **Certificates:** emitted only when `!is_active_cvv()` (`crates/consensus/primary/src/network/handler.rs:322-324`). An active validator (`is_active_cvv() == true`) never populates this feed.
- **Consensus output:** emitted by `notify_exex_consensus_output` inside `handle_sync_output` (`crates/consensus/executor/src/subscriber.rs:180`), whose only callers are the follower / inactive-CVV rejoin paths (`catch_up_rejoin_consensus`, `follow_consensus`). The active-validator path (`handle_consensus_output`) deliberately omits the emit.

The corollary is a nuance worth stating plainly: the dismissal "no starvation on an active validator" is true but not the reason this is safe, because the ExEx manager does not run its delivery path on an active validator anyway.  On the nodes where it *does* run (Observer, inactive CVV), both high-priority feeds are populated and the constant-bias merge is fully in effect.  The safety therefore comes entirely from the three downstream mitigations above, not from the gating.

## The forward-looking trap

The lowest-priority `events` stream is harmless to defer today only because its aggregate (`min_finished_height`) has no consumer.  That is changing by design:

- `#981` describes how a non-reporting ExEx pins `min_finished_height` to `None`, which matters only once a pruner reads it, i.e. the pruning floor is an intended future consumer of this exact stream.
- `#988` restores `FinishedHeight` reporting on the `replay_and_subscribe` path, feeding that same aggregate.

Once a pruner consumes `min_finished_height` and archive mode is disabled, deferring the `events` stream directly stalls the advance of the pruning floor.  That failure stays fail-safe in direction (a stale, lower floor prunes *less*, never deleting history an ExEx still needs), so it is not itself an escalation.  But it converts today's "harmless to defer" stream into a live, consumed dependency sitting at the bottom of a priority ladder with no starvation bound.  Landing the fairness primitive now, while the consumer is still inert, is strictly cheaper than retrofitting it after the dependency goes live.

## A correction to the buffering assumption

Any assessment that leans on a large buffer absorbing the deferral should note the actual capacities: the ExEx broadcast feeds are **not** `10_000`. They are bounded at `1_000` (certificates, `EXEX_CERTIFICATES_CHANNEL_CAPACITY`) and `100` (consensus output, `EXEX_CONSENSUS_OUTPUT_CHANNEL_CAPACITY`) at `crates/consensus/primary/src/consensus_bus.rs:44,50,330-331`. The `10_000` constant (`CHANNEL_CAPACITY`, `crates/types/src/sync.rs`) governs the *internal mpsc bus*, not these broadcast feeds.  Smaller buffers make bursts drain faster (which cuts the other way and reduces harm), but they also mean there is materially less headroom than a `10_000` figure would suggest, so buffer size is not a load-bearing part of the safety argument.  The manager's own fan-out to ExExes is non-blocking `try_send` with a drop-and-count on `Full` (`crates/exex/src/manager.rs:65,87`; `crates/exex/src/notification.rs`), and sends into the broadcast feeds are gated on `receiver_count() > 0` (`consensus_bus.rs:615-618`);  both remain correct and are unaffected by this report.

## Root cause

The manager expresses "strict priority" with a stateless constant (`fn(&mut ()) -> PollNext` returning `PollNext::Left`). Strict priority as intended ("drain the higher stream first") is reasonable, but the constant form also encodes "and defer the lower streams without any bound," which is a stronger and unintended property.  Because `select_with_strategy` gives the strategy a mutable `State` precisely so it can bound its own bias, the fix is to use that state.

## Suggested fix

Replace the constant strategy with a round-robin strategy, using `PollNext` itself as the state.  This is the exact round-robin idiom documented for `select_with_strategy` (`PollNext::toggle(&mut self)` returns the old value and flips the stored state):

```rust
/// Round-robins the two sides: returns the current choice and flips the stored
/// state, so each side is polled first on alternating polls. Nested across the
/// three merges this preserves the certs > consensus > canon > events ordering as
/// bounded throughput shares (1/2 > 1/4 > 1/8 = 1/8 when all are ready) while
/// guaranteeing no source is deferred indefinitely.
fn round_robin(last: &mut PollNext) -> PollNext {
    last.toggle()
}
```

with the state seeded to `PollNext::Left` at each merge so certificates still win the first poll:

```rust
let merged = select_with_strategy(
    certs,
    select_with_strategy(
        consensus,
        select_with_strategy(canon, events, round_robin),
        round_robin,
    ),
    round_robin,
);
```

Applied at all three merge points, this keeps the priority *ordering* (higher-priority sources get a share at least as large as lower-priority ones: strictly larger for certs and consensus, with the two lowest, canon and events, tying at 1/8) but caps the deferral of any single source to one poll of its sibling, so a continuously-ready certificate/consensus feed can no longer defer the canonical-state or event streams without bound.  The `State` type on each `select_with_strategy` changes from `()` to `PollNext`.

If preserving the current "drain the higher stream fully in bursts" behaviour is desired while still bounding starvation, an alternative is a small credit-based strategy whose state is a counter: return `PollNext::Left` while credit remains (decrementing), then yield one poll to the right and reset.  That keeps a stronger, tunable bias than round-robin at the cost of a slightly larger state type.  Round-robin is the simpler default and is sufficient to remove the unbounded-deferral property;  the choice between them is a tuning decision, not a correctness one.

Whichever is chosen, the doc comment at `manager.rs:196-198` should be updated to describe the bounded-fairness behaviour rather than "drains the higher-priority stream fully before the next."

## Suggested test design

- **Starvation-bound unit test:** drive the merge with a higher-priority source that is *always ready* (e.g. `stream::repeat` of a cert marker) alongside a lower-priority source that yields a single, distinguishable item, and assert the lower item is observed within a bounded number of polls.  Under the current constant bias this never completes (the lower item is never reached while the higher stream stays ready);  under `round_robin` it is observed promptly.
- **Confirm-by-mutation:** the test must be seen to fail (hang / time out) if the strategy is reverted to `prefer_left`, so it genuinely pins the fairness property rather than passing vacuously.
- **Ordering-preserved test:** with all four sources continuously ready, assert the observed throughput ordering still favours certs > consensus > canon and does not regress the intended priority (the shares should be monotone in priority), so the fix does not silently flatten priority into pure round-robin across all four.

## Acceptance criteria

- The manager's stream merge bounds how long any single source can be deferred;  a continuously-ready certificate or consensus feed can no longer defer the canonical-state or ExEx-event streams indefinitely.
- The intended priority ordering (`certs > consensus > canon > events`) is preserved as a bounded bias, not removed.
- A test pins the starvation bound and is shown to fail (hang) under mutation back to the constant strategy.
- The merge's doc comment reflects the bounded-fairness behaviour.

MavenRain:
Closes #992.

## Summary

The ExEx manager merges its four input sources (verified certificates, consensus outputs, canonical-state notifications, and the per-ExEx event streams) into one stream using three nested `futures::stream::select_with_strategy` calls. Every layer was biased by the same constant strategy `prefer_left` (state type `()`, unconditionally returning `PollNext::Left`), expressing strict priority `certs > consensus > canon > events`.

A constant `PollNext::Left` does more than express priority: it places no upper bound on how long a lower-priority source is deferred. `select_with_strategy` still polls the lower side whenever the higher side is `Pending`, so this is not a deadlock. But whenever a higher-priority stream is *continuously* ready, the merge drains it in preference and defers the lower streams for as long as that readiness lasts. The two high-priority feeds (certificates, consensus output) are populated precisely on consensus-following nodes (Observer, inactive CVV), which is exactly where the ExEx manager runs, so under a sustained certificate or consensus burst the canonical-state and ExEx-event streams can be deferred indefinitely relative to that burst.

This is Low severity on current code: each deferrable stream is independently and deliberately protected (canonical-state gaps are reconciled via a `Lagged` marker driven by `canon_gap`, the events aggregate `min_finished_height` has no live consumer under archive mode, and the manager is spawned non-critical by default). So this PR is a fairness hardening, not a fix for an observable defect today. It makes the "no source is starved" property hold in the merge itself rather than resting entirely on those downstream mitigations, which matters as the `FinishedHeight` reporting in flight (#981 / #988) moves the lowest-priority stream toward a live pruner consumer. See #992 for the full analysis.

## Root cause

Strict priority ("drain the higher stream first") and unbounded starvation ("and defer the lower streams without any bound") are distinct properties, but a stateless constant strategy encodes both at once. `select_with_strategy` threads a mutable `State` to the strategy on every poll precisely so the strategy can bound its own bias; `prefer_left` carries a `()` state and ignores it, so its choice can never change and the deferral it permits has no ceiling.

## Fix

`crates/exex/src/manager.rs`:

- Replace the constant `prefer_left` with a round-robin strategy that uses `PollNext` itself as the state:
  ```rust
  fn round_robin(last: &mut PollNext) -> PollNext {
      last.toggle()
  }
  ```
  `PollNext::toggle` returns the current choice and flips the stored state, and the state is seeded from `PollNext::default()` (`Left`), so certificates still win the very first poll. Applied at all three merge points, round-robin preserves the priority *ordering* as a bounded bias: when all four sources are continuously ready the merge yields them in the fixed proportion `certs : consensus : canon : events == 4 : 2 : 1 : 1` (that is, 1/2 > 1/4 > 1/8 = 1/8, monotone in priority), while capping the deferral of any single source to one poll of its sibling. A continuously-ready certificate or consensus feed can no longer defer the canonical-state or event streams without bound.
- Extract the three-level merge into a small generic helper `fn merge_sources<T>(certs, consensus, canon, events) -> impl Stream<Item = T>`, so the exact merge topology and strategy the node runs are the same construction the tests exercise (with a light tagging enum as the item type). This closes the gap where a strategy could be correct in isolation but mis-wired at one of the three call sites.
- Update the merge's doc comment to describe the bounded-fairness behaviour rather than "drains the higher-priority stream fully before the next."

Threat model: no attacker is required. The high-priority feeds are ordinary consensus traffic on a following node; a sustained but entirely honest certificate or consensus burst is enough to exercise the deferral. The change is attacker-independent and local to the scheduling of an already-in-process stream merge . . . it adds no new external surface, and the non-blocking `try_send` fan-out (drop-and-count on `Full`), the `receiver_count() > 0` send gating, and the `Lagged` / `canon_gap` reconciliation contract are all unchanged. The only observable effect is a bounded, priority-preserving interleaving in place of an unbounded one.

## Testing

Verified on the pinned toolchain: `cargo +1.94 test -p tn-exex` (all 9 tests pass), `cargo +1.94 clippy -p tn-exex --all-targets` and `cargo +nightly fmt -- --check` clean, and the full `make attest` proxy (`cargo +1.94 test --no-run --workspace -j 2`) green in an isolated target dir.

Three tests were added to `manager.rs`, exercising the exact `merge_sources` construction:

- `round_robin_seeds_left_then_alternates` pins that the strategy is seeded `Left` (certificates win the first poll) and then alternates, so priority is not lost and neither side is polled first twice running.
- `merge_bounds_low_priority_starvation` drives the merge with a continuously-ready, infinite highest-priority source alongside a single distinguishable lowest-priority item, and asserts the item is observed within a bounded number of polls . . . the starvation-bound test the issue asks for.
- `merge_preserves_priority_ordering_without_flattening` drives all four sources continuously-ready and asserts the observed throughput is the monotone `4 : 2 : 1 : 1` bias (certs > consensus > canon, canon == events), so the fix does not silently flatten priority into a uniform round-robin.

Confirm-by-mutation: reverting `round_robin` to a constant `PollNext::Left` was run and both merge tests were seen to fail (`merge_bounds_low_priority_starvation` observes `[Cert; 8]` with the event starved; `merge_preserves_priority_ordering_without_flattening` fails `consensus 0 !> canon 0`), then the fix was restored. The tests genuinely pin the fairness property rather than passing vacuously.

### DIFF
--- a/crates/exex/src/manager.rs
+++ b/crates/exex/src/manager.rs
@@ -17,7 +17,9 @@
 //! ExEx can detect the gap and reconcile via [`replay`](crate::replay).
 
 use crate::{Chain, TnExExEvent, TnExExNotification};
-use futures::stream::{self, select_all, select_with_strategy, PollNext, StreamExt, TryStreamExt};
+use futures::stream::{
+    self, select_all, select_with_strategy, PollNext, Stream, StreamExt, TryStreamExt,
+};
 use reth_provider::CanonStateNotification;
 use std::sync::Arc;
 use tn_reth::CanonStateNotificationStream;
@@ -193,18 +195,11 @@ impl TnExExManager {
             ReceiverStream::new(rx).map(move |ev| RouterEvent::Event { idx, ev })
         }));
 
-        // Biased priority certs > consensus > canon > events via a constant `PollNext::Left`, which
-        // drains the higher-priority stream fully before the next (matching the original sequential
-        // poll order, where each stream was drained to `Pending` before the next was polled).
-        let merged = select_with_strategy(
-            certs,
-            select_with_strategy(
-                consensus,
-                select_with_strategy(canon, events, prefer_left),
-                prefer_left,
-            ),
-            prefer_left,
-        );
+        // Merge the four sources under a bounded-fairness round-robin schedule (see
+        // `merge_sources`): the priority ordering certs > consensus > canon > events is kept as
+        // throughput shares, but no source can be deferred indefinitely by a continuously-ready
+        // higher-priority feed.
+        let merged = merge_sources(certs, consensus, canon, events);
 
         // The fold ends only by observing a closed source stream (signalled as `StreamEnded`); it
         // never yields a real error, so the result is discarded and the manager reports `Ok(())`.
@@ -220,10 +215,40 @@ impl TnExExManager {
     }
 }
 
-/// Always drains the higher-priority (left) stream first, giving the fixed source order
+/// Round-robins the two sides of a merge: returns the current choice and flips the stored
+/// `PollNext` state, so each side is polled first on alternating polls. This is the round-robin
+/// idiom documented for [`select_with_strategy`]. The state is seeded from `PollNext::default()`
+/// (`Left`), so the left (higher-priority) side still wins the very first poll.
+fn round_robin(last: &mut PollNext) -> PollNext {
+    last.toggle()
+}
+
+/// Merges the manager's four input streams into one, in priority order
 /// certs > consensus output > canonical state > events.
-fn prefer_left(_: &mut ()) -> PollNext {
-    PollNext::Left
+///
+/// Each of the three nested [`select_with_strategy`] layers uses [`round_robin`] rather than a
+/// constant `PollNext::Left` bias. Round-robin at every layer keeps the priority *ordering* as a
+/// bounded bias: when all four sources are continuously ready the merge yields them in the fixed
+/// proportion `certs : consensus : canon : events == 4 : 2 : 1 : 1` (i.e. 1/2 > 1/4 > 1/8 = 1/8,
+/// monotone in priority), while capping the deferral of any single source to one poll of its
+/// sibling. A continuously-ready certificate or consensus feed therefore can no longer starve the
+/// canonical-state or ExEx-event streams, which a constant `PollNext::Left` bias permitted without
+/// any upper bound.
+fn merge_sources<T>(
+    certs: impl Stream<Item = T>,
+    consensus: impl Stream<Item = T>,
+    canon: impl Stream<Item = T>,
+    events: impl Stream<Item = T>,
+) -> impl Stream<Item = T> {
+    select_with_strategy(
+        certs,
+        select_with_strategy(
+            consensus,
+            select_with_strategy(canon, events, round_robin),
+            round_robin,
+        ),
+        round_robin,
+    )
 }
 
 /// A single merged input to the [`TnExExManager`] event loop.
@@ -548,5 +573,80 @@ mod tests {
         // Out-of-order/duplicate commit (older range) → no gap, tip never rewinds.
         assert_eq!(canon_gap(&mut tip, 5, 8), None);
         assert_eq!(tip, Some(13));
+    }
+
+    // Tags which source a merged item came from, so the merge tests can assert both
+    // fairness (no source is starved) and the preserved priority ordering.
+    #[derive(Clone, Copy, PartialEq, Eq, Debug)]
+    enum Source {
+        Cert,
+        Consensus,
+        Canon,
+        Event,
+    }
+
+    #[test]
+    fn round_robin_seeds_left_then_alternates() {
+        // Seeded from the default (`Left`), certificates win the first poll; the state then
+        // flips on every call, so neither side is ever polled first twice in a row.
+        let mut state = PollNext::default();
+        assert!(matches!(round_robin(&mut state), PollNext::Left));
+        assert!(matches!(round_robin(&mut state), PollNext::Right));
+        assert!(matches!(round_robin(&mut state), PollNext::Left));
+        assert!(matches!(round_robin(&mut state), PollNext::Right));
+    }
+
+    #[tokio::test]
+    async fn merge_bounds_low_priority_starvation() {
+        // The highest-priority source is continuously ready and infinite; the lowest-priority
+        // source carries a single, distinguishable item. Under the round-robin schedule that lone
+        // event must surface within a small, bounded number of polls rather than be deferred
+        // forever behind the ready certificate feed.
+        //
+        // Confirm-by-mutation: reverting `round_robin` to a constant `PollNext::Left` (the original
+        // `prefer_left`) leaves the `Event` unreachable while the certificate stream stays ready,
+        // so `saw_event` stays false and this assertion fails.
+        let certs = stream::repeat(Source::Cert);
+        let consensus = stream::empty::<Source>();
+        let canon = stream::empty::<Source>();
+        let events = stream::once(async { Source::Event });
+
+        let observed: Vec<Source> =
+            merge_sources(certs, consensus, canon, events).take(8).collect().await;
+
+        let saw_event = observed.contains(&Source::Event);
+        assert!(saw_event, "low-priority event starved under a ready cert feed: {observed:?}");
+    }
+
+    #[tokio::test]
+    async fn merge_preserves_priority_ordering_without_flattening() {
+        // With all four sources continuously ready, the round-robin nest yields the fixed
+        // proportion certs : consensus : canon : events == 4 : 2 : 1 : 1 over each 8-item cycle.
+        // Assert the priority ordering survives as a strict, monotone bias (certs > consensus >
+        // canon) with the two lowest tying (canon == events), i.e. the fix does not silently
+        // flatten priority into a uniform round-robin across all four sources.
+        //
+        // Confirm-by-mutation: under a constant `PollNext::Left` every item is a certificate, so
+        // `consensus_n > canon_n` (0 > 0) fails.
+        const N: usize = 8 * 100;
+        let certs = stream::repeat(Source::Cert);
+        let consensus = stream::repeat(Source::Consensus);
+        let canon = stream::repeat(Source::Canon);
+        let events = stream::repeat(Source::Event);
+
+        let observed: Vec<Source> =
+            merge_sources(certs, consensus, canon, events).take(N).collect().await;
+
+        let count = |want: Source| observed.iter().filter(|s| **s == want).count();
+        let certs_n = count(Source::Cert);
+        let consensus_n = count(Source::Consensus);
+        let canon_n = count(Source::Canon);
+        let events_n = count(Source::Event);
+
+        assert!(certs_n > consensus_n, "certs {certs_n} !> consensus {consensus_n}");
+        assert!(consensus_n > canon_n, "consensus {consensus_n} !> canon {canon_n}");
+        assert_eq!(canon_n, events_n, "canon {canon_n} != events {events_n} (lowest two must tie)");
+        // The exact 4:2:1:1 shares document the schedule the ordering rests on.
+        assert_eq!((certs_n, consensus_n, canon_n, events_n), (N / 2, N / 4, N / 8, N / 8));
     }
 }


sstanfield (approved):
Seems reasonable to me.

PR #1012, opened 2026-07-23
refactor(engine): split the epoch boundary into three system calls

Huwonk:
## Summary

Client side of the epoch-boundary rework. Pairs with tn-contracts#162, which splits `applyIncentives` and `applySlashes` back out of `concludeEpoch`.

The closing block now makes three system calls in order - `applyIncentives(RewardInfo[])`, then `applySlashes(Slash[])`, then `concludeEpoch(address[])` - instead of one unified `concludeEpoch(address[], RewardInfo[], Slash[])`.

## Why the split

A review of the unified design found two latent consensus-safety gaps once slashing ships:

- The unified `concludeEpoch` validated `newCommittee.length` against post-slash `nextCommitteeSize` *inside* the call, but the client assembles the committee *before* the call. A slash-to-zero ejection would make the length check revert - and a revert in the epoch close is a deterministic fleet-wide halt.
- `_ejectFromCommittees` cannot touch the caller's calldata committee array, so an ejected validator could still be seated in a future committee.

Sequencing `applySlashes` before the client reads `nextCommitteeSize` and runs the committee shuffle makes both reads post-slash by construction, so both findings dissolve.

## Changes

- `system_calls.rs`: the `ConsensusRegistry` binding's unified three-argument `concludeEpoch` splits into `applyIncentives`, `applySlashes`, and `concludeEpoch(address[])`.
- `evm/block.rs`: `apply_closing_epoch_contract_call` issues the three calls in order, running `shuffle_new_committee` only after `applySlashes` commits so the eligible-pool read reflects ejections. Slashing is not live, so `applySlashes` always carries an empty array; the call is issued anyway so the sequence is correct by construction when slashing ships. The legacy fork-replay path is unchanged - pre-fork chains never executed `applySlashes`, so it stays a two-call replay, and the restored post-fork `applyIncentives`/`concludeEpoch(address[])` selectors are byte-identical to it, so the two paths differ only by the interposed `applySlashes`.
- `evm/mod.rs`: points the last stale `30,000,000` doc comment at the `SYSTEM_CALL_GAS_LIMIT` constant. The budget raise to 100M itself landed earlier on this branch (needed because the split isolates the version-queue settlement wave into `concludeEpoch`, whose worst case measures ~33.9M at a 1000-entry wave - comfortable under 100M, over the old 30M).

## Lockstep

Draft until the `tn-contracts` submodule pin advances to a master containing tn-contracts#162 (and its regenerated artifacts), since the execution tests deploy the vendored registry bytecode. `cargo check` passes now; the submodule bump and e2e follow the merge. Folds into the same pin as #909 (delegation deadline) and #1070 (WorkerConfigs).


grantkee (approved):
LGTM! Great work @Huwonk

sstanfield (approved):

PR #1021, opened 2026-07-23
fix(epoch): write and persist the epoch record on all three epoch-close arms

### LINKED ISSUE #1015: Recovery-arm epoch closes never write the epoch record; an epoch 0 closed this way bricks the node on its next restart
## Problem

The epoch manager closes an epoch on three arms: the live boundary arm, a replay-and-close recovery arm (the node restarted mid-boundary and replays missed consensus output to finish the close), and a leftover-drain recovery arm (the boundary output was already executed when the epoch's tasks exited, but the close never completed). Only the live arm writes the finalized epoch record. Both recovery arms close the epoch, clear the per-epoch consensus tables, and advance to the next epoch without ever writing a record.

The sharpest consequence is epoch 0. Epoch 0 starts from an unsigned dummy record that the record store holds in a memory-only slot: persist never writes it, and a reopen resets it. An epoch 0 closed through either recovery arm therefore leaves no durable record 0, and the node's next restart trips the entry guard that insists epoch 0 exist in the record store once the node is past epoch 0. Every subsequent start fails the same way. Peers cannot heal it: validating a downloaded record 0 anchors on the locally stored epoch-0 committee keys, that lookup returns nothing without a durable record 0, so state-sync sees the record as unanchored and its step-back-one retry saturates at epoch 0 without ever making progress. The node is permanently bricked.

The same gap breaks the epochs that follow within the same process. A recovery-closed epoch 0 leaves the store's append-only index empty, so when epoch 1 later closes on the live arm the store rejects its record as out of order. And because the record write is also what publishes the record for signing, a recovery-closed epoch never reaches the record vote collector: no votes, no certificate, no quorum.

A durability gap rides along even on the live arm. The record write only saved the record; the store was persisted to disk later by the asynchronous certificate-quorum path, so a crash after the close but before quorum lost the record. Epochs past 0 can be backfilled from peers; epoch 0 cannot, for the same no-anchor reason, so the crash window ends in the same brick.

## Solution

All three close arms should write the epoch record through one shared step, invoked immediately after the close and before the per-epoch tables are cleared. That step should take the epoch number directly rather than reading it from the live primary, because the replay arm closes before the epoch's consensus components are ever created and has no primary to ask. Routing every close through the same step also publishes recovery-closed records for signing, so they reach the vote collector like live closes do.

The record write needs the boundary consensus header, and both recovery paths already hold it at the moment they detect the boundary output: the replay path and the leftover drain should stash that header for the close-and-write sequence, and a missing header at write time should surface as an error rather than a panic now that the write has three callers.

The write should also persist the record store to disk (fsync) immediately after saving, instead of leaving durability to the asynchronous certificate-quorum path. A storage test should pin the epoch-0 failure shape: the dummy record must vanish on reopen even after an explicit persist, while a real record 0, saved and persisted, must survive reopen with its digest and committee anchor intact.

grantkee:
- The replay-and-close and leftover-drain recovery arms now call `write_epoch_record` immediately after `close_epoch`, before clearing epoch-scoped tables; previously only the live boundary arm wrote the record, so a recovery close finished the epoch with no record at all.
- The stakes are highest for epoch 0: its dummy record lives in a memory-only slot that reopen resets, so a recovery-closed epoch 0 left no durable record 0 and every later restart died on the epoch-0 entry guard; peers cannot backfill it because `get_committee_keys(0)` returns nothing, downloaded record-0 validation reports `NoAnchor`, and state-sync's step-back retry saturates at 0.
- `write_epoch_record` takes the `Epoch` directly instead of `&PrimaryNode`: the replay arm closes before the epoch's primary exists, so reading the committee from a live primary was never callable there.
- Both replay paths stash the boundary consensus header into `last_consensus_header` when they detect the boundary output (`replay_missed_consensus`, plus both drain phases of `send_leftover_consensus_output_to_engine`); the record write needs that header, and its missing-header `expect` becomes an `eyre` error now that three arms call it.
- `write_epoch_record` calls `epochs().persist()` right after `save_record`: durability previously arrived only via the asynchronous cert-quorum path, so a crash between a live close and quorum lost the record, unrecoverable for epoch 0 (no peer-fetch anchor).
- Two more shapes this closes: a recovery-closed epoch 0 poisoned epoch 1's later live close with `EpochOutOfOrder` (the append index never saw record 0), and recovery-closed epochs never reached the record vote collector because nothing was published on `epoch_record_watch`.
- New storage test `test_dummy_epoch0_lost_on_reopen_but_persisted_record_survives`: the dummy epoch-0 record dies on reopen even after an explicit `persist`, while a real record 0, saved and persisted, survives with its digest and committee anchor intact.

closes #1015

### DIFF
--- a/crates/node/src/manager/node/close_epoch.rs
+++ b/crates/node/src/manager/node/close_epoch.rs
@@ -14,7 +14,7 @@
 //! next epoch starts from a clean slate. Historic data survives in the
 //! `ConsensusChain` store; only the per-epoch working tables are cleared.
 
-use crate::{engine::ExecutionNode, manager::EpochManager, primary::PrimaryNode};
+use crate::{engine::ExecutionNode, manager::EpochManager};
 use eyre::eyre;
 use std::collections::BTreeSet;
 use tn_config::TelcoinDirs;
@@ -112,7 +112,10 @@ where
     ///
     /// Returns the [`ConsensusHeaderDigest`] of the first output committed at or
     /// past the epoch boundary, signalling that an epoch-boundary output was
-    /// reached; `None` if no such output was found.
+    /// reached; `None` if no such output was found. That boundary output's header
+    /// is also stashed in `last_consensus_header` so the caller's close-and-write
+    /// sequence (`close_epoch`, then `write_epoch_record`) can commit the epoch's
+    /// record.
     pub(super) async fn send_leftover_consensus_output_to_engine(
         &mut self,
         consensus_output: &mut impl TnReceiver<ConsensusOutput>,
@@ -121,6 +124,8 @@ where
         // Phase 1: Drain broadcast channel (existing behavior)
         while let Ok(output) = consensus_output.try_recv() {
             let result = if output.committed_at() >= self.epoch_boundary {
+                // stash the boundary header for the caller's close-and-write sequence
+                self.last_consensus_header = Some(output.clone().into());
                 Some(output.consensus_header_hash())
             } else {
                 None
@@ -143,6 +148,8 @@ where
                 match self.consensus_chain.get_consensus_output_current(number).await {
                     Ok(output) => {
                         let result = if output.committed_at() >= self.epoch_boundary {
+                            // stash the boundary header for the caller's close-and-write sequence
+                            self.last_consensus_header = Some(output.clone().into());
                             Some(output.consensus_header_hash())
                         } else {
                             None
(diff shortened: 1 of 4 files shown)


crates/node/src/manager/node/close_epoch.rs
```diff
@@ -248,6 +256,10 @@ where
         )?;
 
         self.consensus_chain.epochs().save_record(epoch_rec.clone()).await?;
+        // the cert-quorum path persists asynchronously; a crash before that flush would lose
+        // the record just saved — fatal for epoch 0, which has no peer-fetch anchor — so
+        // flush it durably here before publishing
+        self.consensus_chain.epochs().persist().await?;
```
grantkee, line 262: #977 may affect this call @MavenRain @sstanfield

MavenRain: Noted

MavenRain (approved):
Should be fine

sstanfield (approved):

PR #1022, opened 2026-07-23
fix(engine): Hard-error on undecodable BLS keys in the pinned committee read

### LINKED ISSUE #1016: Pinned committee read silently drops undecodable on-chain BLS keys
## Problem

The engine's pinned committee read, `validators_for_epoch_at_block`, fetches a committee's BLS public key bytes from on-chain registry state and decodes each key with `filter_map(... .ok())`. A key whose bytes fail to decode is dropped: the read succeeds, returns a committee missing that validator, and logs nothing.

A decode failure means the on-chain bytes are not a valid BLS key — corrupted state or an unexpected registry write, something a healthy chain never produces. There is no correct way to keep operating with those bytes, yet this path treats the failure as if the validator did not exist. The epoch-entry committee read decodes the same on-chain bytes and hard-errors on failure, so the two paths disagree on the severity of the same fault.

The worst-placed caller is epoch close, where this read supplies the committee and next-committee key sets written into the epoch record. Record validation deliberately tolerates a shrunken committee because mid-epoch governance burns eject validators legitimately, so a decode-failure shrink is indistinguishable from a real ejection: the node commits a wrong committee to its permanent epoch-record chain and diverges from every peer that decoded the key, with no signal that anything went wrong. Epoch start uses the same read for the outgoing committee's peer handling and the next committee's config keys, so a dropped key there misclassifies a live validator's network identity.

## Solution

The pinned committee read should reject the whole read when any key fails to decode, mirroring the epoch-entry path: return an error naming the epoch and the pinned block instead of dropping the key. Every caller already propagates errors, so no signatures change and epoch close halts before writing a wrong record — the correct failure mode when the alternative is permanent divergence.

The epoch+2 committee prefetch at epoch start is a best-effort network warm-up and should keep its warn-and-skip handling, so a decode failure there stays a logged skip rather than an aborted epoch start.

grantkee:
- `validators_for_epoch_at_block` decoded on-chain BLS key bytes with `filter_map(... .ok())`, so an undecodable key was silently dropped and the returned committee shrank; the decode is now a hard error naming the epoch and pinned block, matching the epoch-entry committee read, which already hard-errors on this decode.
- The silent shrink was worst at epoch close: this read supplies the committee sets written into the epoch record, and record validation tolerates shrunken committees (mid-epoch burns are legitimate ejections), so a corrupt key meant committing a permanently wrong record instead of halting.
- No signature changes: the function already returned `eyre::Result` and every caller propagates with `?`. The epoch+2 prefetch keeps its own warn-and-skip, so the best-effort warm-up still cannot abort epoch start.
- The read's doc comment now states the decode-failure contract.

closes #1016

### DIFF
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -383,16 +383,24 @@ impl ExecutionNodeInner {
     }
 
     /// Read committee validator keys for epoch, pinned to the block identified by `block_hash`.
+    /// On-chain BLS key bytes are decoded here; a decode failure is a hard error, mirroring the
+    /// epoch-entry committee read.
     pub(super) fn validators_for_epoch_at_block(
         &self,
         epoch: u32,
         block_hash: B256,
     ) -> eyre::Result<Vec<BlsPublicKey>> {
-        Ok(self
-            .reth_env
+        self.reth_env
             .bls_pubkeys_for_epoch_at_block(epoch, block_hash)?
             .iter()
-            .filter_map(|bls| BlsPublicKey::from_literal_bytes(bls.as_ref()).ok())
-            .collect())
+            .map(|bls| {
+                BlsPublicKey::from_literal_bytes(bls.as_ref()).map_err(|err| {
+                    eyre::eyre!(
+                        "failed to create bls key from on-chain bytes for epoch {epoch} at block \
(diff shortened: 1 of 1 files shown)


sstanfield (approved):

MavenRain (approved):
Looks clean

PR #1024, opened 2026-07-23
Epoch read docs and consolidation

### LINKED ISSUE #1017: Epoch entry performs five EVM constructions and five header lookups to resolve one pinned view
## Problem

Entering an epoch resolves all epoch-scoped state against one pin: the previous epoch's closing block, held as the epoch-start header. The epoch-state read seats the committee, then three more committee reads follow: the previous committee (so peers from the outgoing committee are not banned during handover), the next committee (folded into the consensus config), and the epoch+2 committee (a best-effort network prefetch). Every read constructs its own EVM, each committee read re-fetches the epoch-start header by hash even though the entry path already holds it, and the epoch-state read re-fetches the closing block by number for its timestamp even though on this path the pin IS the closing block. Counting the tip-side bootstrap that identifies the epoch, entry totals five EVM constructions and five header lookups, four of the lookups resolving a header already in hand.

The four pinned reads target the same state root, so building a fresh EVM and re-looking-up the header for each is pure overhead, paid on every entry shape: fresh boundary crossing, crash-restart replay, and mode-change re-entry. The scattering is the deeper problem. The next-committee read hides inside consensus configuration and the previous-committee read inside consensus creation, so the pin discipline the entry-read invariant depends on is enforced by convention at three separate call sites, each one refactor away from quietly reading at a different header — and after a mid-epoch governance burn swap-and-pops the stored committee arrays, a different header means a different committee.

Two smaller defects sit in the same path. The entry comment credits the concurrent-re-entry safety to a shrink guard inside the gas accumulator's worker-count setter that does not exist (the setter truncates unconditionally; the actual guard is that the pinned re-read yields identical values), and the engine documents its pinned committee accessor as the only committee read it exposes while its tip-state read also returns committee arrays. Separately, the registry layer still exposes unpinned per-epoch committee readers with zero production callers; their results shift under a mid-epoch burn, a standing invitation to reintroduce exactly the read class the pin discipline exists to remove.

## Solution

Consolidate the entry reads onto the header the path already holds. Resolve the previous and next committees in one batched registry read against a single pinned EVM at the epoch-start header, hoisted into the epoch-entry sequence beside the epoch-state read, and thread the two key sets into consensus configuration and consensus creation as parameters; consensus configuration should issue no chain read of its own, so the config cannot derive from a different pin than the rest of the entry. The epoch-state read should reuse the held header for the closing timestamp when the pin is the closing block, and the epoch+2 prefetch should read at the held header instead of by hash, keeping its documented fail-open (log and skip) behavior. Entry then costs four EVM constructions and one header lookup, with every read value-identical to before: same pin, same state, neighbor reads still hard errors. Decode semantics should not change here; a separate branch hardens decode.

Alongside, correct the two comments to name the real guards, and delete the unpinned per-epoch committee readers, migrating their test callers to the pinned at-block variants so every committee read in the tree names its pin explicitly.

grantkee:
- Docs: the ENTRY-READ INVARIANT's concurrent-execution argument credited `GasAccumulator::set_num_workers` with refusing to shrink below an in-flight worker id — it truncates unconditionally. The comment now names the real guard: the pinned re-read yields the identical worker count so the resize no-ops, and a genuine shrink would trip `inc_block`'s production panic rather than pass silently.

- Docs: `validators_for_epoch_at_block` claimed to be "the ONLY committee read the engine exposes", but `epoch_state_from_canonical_tip` also returns committee fields, read at the tip. The claim is now the true one (every committee-keys read the engine exposes is pinned), and the tip reader documents that a mid-epoch governance burn mutates the committee arrays it returns, so it is for point-in-time queries only.

- Consolidation: entry goes from five EVM constructions and five header lookups to four and one. The previous/next committee resolution moves out of `create_consensus`/`configure_consensus` into `run_epoch`'s entry sequence as one batched read (`validators_for_epochs_at_header`) against a single pinned EVM at the held epoch-start header, with the key sets threaded in as parameters; `configure_consensus` drops its engine parameter and performs no chain read, so the config cannot derive from a different pin than the rest of the entry.

- Consolidation: `RethEnv::bls_pubkeys_for_epochs_at_header` batches one `getCommitteeBlsPubkeys` call per epoch through the existing `read_consensus_registry_batch_at_header`, results ordered to match the input; a unit test asserts the batch equals the single pinned reads. `bls_pubkeys_for_epoch_at_block` now delegates to the single-epoch `_at_header` wrapper, and `epoch_state_at_header` reuses the held header for the closing timestamp when the pin IS the closing block (the epoch-start path) instead of re-fetching it by number.

- Consolidation: no read changes value — same pin, same state. Neighbor reads stay hard errors (each already halted entry on failure), the epoch+2 prefetch keeps its documented fail-open (warn and skip) and only switches to the held header, and decode (`from_literal_bytes(...).ok()` filtering) is deliberately untouched; a separate branch hardens decode.

- Reader deletion: `RethEnv::validators_for_epoch` and `bls_pubkeys_for_epoch` read the canonical tip unpinned, had zero production callers, and their results shift under a mid-epoch governance burn — an open invitation for new unpinned epoch-scoped reads. Deleted; test callers (tn-reth unit tests and the epoch-record IT) migrate to the `_at_block` variants pinned to an explicit canonical-tip hash, value-identical where they ran.

closes #1017

### DIFF
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -18,8 +18,8 @@ use tn_reth::{
 use tn_rpc::{EngineToPrimary, TelcoinNetworkRpcExt, TelcoinNetworkRpcExtApiServer};
 use tn_types::{
     gas_accumulator::GasAccumulator, Address, BatchSender, BatchValidation, BlockHeader,
-    BlsPublicKey, ConsensusHeaderDigest, ConsensusOutput, EngineUpdate, Epoch, ExecHeader, Noticer,
-    SealedHeader, TaskSpawner, WorkerId, B256,
+    BlsPublicKey, Bytes, ConsensusHeaderDigest, ConsensusOutput, EngineUpdate, Epoch, ExecHeader,
+    Noticer, SealedHeader, TaskSpawner, WorkerId, B256,
 };
 use tn_worker::WorkerNetworkHandle;
 use tokio::sync::mpsc;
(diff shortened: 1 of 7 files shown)


sstanfield (dismissed):

crates/node/src/engine/inner.rs
```diff
@@ -395,4 +395,37 @@ impl ExecutionNodeInner {
...
+        Ok(self
+            .reth_env
+            .bls_pubkeys_for_epochs_at_header(epochs, header)?
+            .iter()
+            .map(|keys| {
+                keys.iter()
+                    .filter_map(|bls| BlsPublicKey::from_literal_bytes(bls.as_ref()).ok())
+                    .collect()
+            })
+            .collect())
```
MavenRain: Might be useful to extract a private `decode_bls_keys` for 

```
.filter_map(|bls| BlsPublicKey::from_literal_bytes(bls.as_ref()).ok()).collect()
```

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -244,9 +249,37 @@ where
...
+                (HashSet::new(), next)
+            } else {
+                let [previous, next] = engine
+                    .validators_for_epochs_at_header(
+                        &[entered - 1, entered + 1],
+                        &epoch_start_header,
+                    )
+                    .await?
+                    .try_into()
+                    .map_err(|_| eyre::eyre!("neighbor-committee batch arity mismatch"))?;
```
MavenRain: Nit: Might be worth extracting a method for the repeated error-handling bit:

```
.try_into().map_err(|_| eyre::eyre!("neighbor-committee batch arity mismatch"))
```

MavenRain (dismissed):
Minor suggestions

crates/node/src/engine/inner.rs
grantkee: Good idea - added

crates/node/src/manager/node/run_epoch.rs
grantkee: I made the errors more descriptive

MavenRain (approved):

PR #1025, opened 2026-07-23
fix(epoch): Classify and retry provider faults on epoch-record and epoch-entry committee reads

### LINKED ISSUE #1018: Epoch-boundary committee reads halt on the first transient provider fault
## Problem

The epoch boundary already has a classified-retry contract for node-local faults: `StateReadError` splits `Provider` faults (transient node-local I/O, where a fresh read may succeed) from `ChainGlobal` faults (deterministic products of the pinned chain state, identical on every node), and the close-time worker fee-config reads retry `Provider` faults a bounded number of times with backoff before halting. Only those close-time reads use it.

The committee reads at the same boundary do not. The epoch-record reads at close (the current and next committee, pinned to the epoch-closing block and written into the signed epoch record) and every epoch-entry read (the epoch-state read plus the previous-, next-, and epoch+2-committee reads at the epoch-start pin) return unclassified errors, so one transient MDBX/provider hiccup at exactly the wrong moment halts the node with no retry. These are the two moments where a halt is most expensive: writing the epoch record at close and entering the next epoch.

A retry is meaningful here, not decorative. Every one of these reads constructs a fresh state provider (a fresh read transaction) per call, so a transient fault genuinely can clear on re-read, and the retry-recovers contract is already encoded by a unit test on the worker-config path. The asymmetry is arbitrary: the same I/O blip is survivable during the fee-config reads and fatal moments later on the committee reads.

The record path carries a second, smaller cost: its two committee reads run serially, each resolving the same pin header and building its own state view, where one snapshot could serve both.

## Solution

Classify the underlying reth-layer reads the same way the worker-config reads already are: the epoch-state read (at a given header, and its epoch-start variant) and the pinned committee-pubkey reads should return `StateReadError`, with state-provider construction, database faults, and header lookups as `Provider`, and EVM-environment construction plus everything downstream of a successfully executing EVM as `ChainGlobal`. Add a batched multi-epoch pubkey read so the record's two committees resolve from one pinned snapshot with one header resolution.

Generalize the existing retry helper to async reads and wrap the record-write committee reads and all four entry reads in it. The policy itself should not change: `Provider` faults retry up to the existing attempt budget and then halt; `ChainGlobal` failures propagate immediately, including the registry-convention tripwire, which must never be retried because re-reading cannot un-break a chain-state convention. No hard read may fail open — a wrong committee in a signed epoch record is permanent divergence every syncing node inherits, while a halt is a single-node liveness failure. The epoch+2 prefetch should keep its warn-and-skip fail-open, applied after the retries.

The RPC-facing batch read path should stay unclassified: the RPC layer distinguishes contract reverts through it.

grantkee:
- `epoch_state_at_header`, `epoch_state_at_epoch_start`, and the pinned committee-pubkey reads now return `StateReadResult`: state-provider construction, database faults during registry reads, and header lookups classify as `Provider` (node-local, may clear on re-read); EVM-env construction and everything downstream of a successfully executing EVM classify as `ChainGlobal` (identical on every node). The internal registry helpers route through the existing `classified_registry_read`.

- New `bls_pubkeys_for_epochs_at_block` reads several epochs' committees from one pinned EVM, results ordered as requested; the single-epoch read is now a wrapper over it.

- `retry_provider_faults` generalized to async reads and shared across the boundary modules; the sync close-time callers adapt with `ready`. Attempt budget and backoff unchanged.

- Record write at close: the two serial committee reads (two header resolutions, two state views) collapse into one retried batched read at the closing-block pin, halting on exhausted `Provider` retries or any `ChainGlobal` failure. Deliberately no fail-open — a wrong committee in a signed epoch record is permanent divergence every syncing node inherits, while a halt is a single-node liveness failure.

- Epoch entry: the epoch-state read and the previous/next committee reads retry provider faults, then halt. The epoch+2 prefetch retries first and keeps its warn-and-skip fail-open, so a transient blip no longer silently drops the warm-up.

- The registry-convention tripwire classifies `ChainGlobal` and is therefore never retried: re-reading cannot un-break a chain-state convention.

- The RPC-facing `read_consensus_registry_batch` stays on unclassified `EvmReadResult` on purpose: the RPC layer distinguishes contract reverts through it.

- Tests: classification asserts for the epoch-state and pubkey reads (unresolvable header or block hash is `Provider`, absent registry at a resolvable block is `ChainGlobal`), the batched read matches the single reads at the pin in epoch order, and the retry-helper tests carry over to the async signature.

closes #1018

### DIFF
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -11,6 +11,7 @@ use tn_batch_validator::BatchValidator;
 use tn_config::Config;
 use tn_engine::ExecutorEngine;
 use tn_reth::{
+    error::{StateReadError, StateReadResult},
     system_calls::EpochState,
     worker::{WorkerComponents, WorkerNetwork},
     RethEnv, RpcServerHandle, WorkerTxPool,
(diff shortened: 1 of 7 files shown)


sstanfield (dismissed):

crates/node/src/engine/inner.rs
```diff
@@ -378,21 +379,51 @@ impl ExecutionNodeInner {
...
+        epochs: Vec<Epoch>,
+        block_hash: B256,
+    ) -> StateReadResult<Vec<Vec<BlsPublicKey>>> {
+        Ok(self
+            .reth_env
+            .bls_pubkeys_for_epochs_at_block(&epochs, block_hash)?
+            .iter()
+            .map(|keys| {
+                keys.iter()
+                    .filter_map(|bls| BlsPublicKey::from_literal_bytes(bls.as_ref()).ok())
```
MavenRain: If I am understanding this correctly, a malformed key can silently shrink the committee instead of halting.  Consider, alternatively, propagating a typed `StateReadError` to callers who can decide to retry or halt instead of proceeding with the reduced committee.

MavenRain (dismissed):
Just one (latent?) observation

crates/node/src/engine/inner.rs
grantkee: This is checked at the contract level, but I agree returning an error here is more correct

MavenRain (approved):

PR #1026, opened 2026-07-23
Drop the debug_assert shadowing the close-time base-fee identity guard

### LINKED ISSUE #1019: debug_assert shadows the close-time base-fee identity guard
## Problem

When an epoch closes, the epoch manager recomputes each worker's next-epoch base fee from on-chain worker configs read at the canonical tip, which at that point in the close sequence is the epoch's closing block. Before touching fees it checks the close-time identity: the newly entered epoch's on-chain `blockHeight`, stamped by `concludeEpoch` in the ConsensusRegistry, must equal canonical tip + 1. On a violation the function returns a descriptive error and refuses to price base fees off a non-closing block.

That guard is preceded by a `debug_assert_eq!` of the exact same condition. In any build with debug assertions, a violation aborts the process on the assert before the recoverable `Err` path can run, so the guard's error branch is dead code in every debug build.

This makes the guard untestable: `cargo test` compiles with debug assertions, so a unit test driving a violating tip through the function aborts instead of observing the error. The only profile where the guard is live is release, exactly the profile no test exercises, so production relies on a consensus-adjacent safety check that nothing has ever executed. The assert also adds no protection of its own: in release it compiles out and the hard guard already halts the close, and in debug it only converts a recoverable, logged failure into a process abort.

## Solution

Drop the debug_assert and keep the hard error guard as the single enforcement of the identity in every build profile. At the detection site the guard should log an error with the tip number, entered epoch, and on-chain block height before returning, matching the tripwire shape already used by the epoch-start identity guard in the reth wrapper.

Add a unit test that drives a non-closing tip through the base-fee adjustment and proves the guard trips: it must return the identity-violation error and leave the per-worker fees untouched. With the assert gone the path is reachable in debug builds, so the test runs in the default profile.

grantkee:
- The identity guard in `adjust_base_fees` (canonical tip + 1 must equal the entered epoch's on-chain `blockHeight`) was preceded by a `debug_assert_eq!` of the same condition, so every debug build aborted on the assert before the recoverable `Err` could run: the guard was dead code in debug builds and its release-only live path had no test coverage.
- The assert is gone; the hard `Err` guard is now the single enforcement in all build profiles.
- The detection site now logs `error!` with `tip_number`, `entered_epoch`, and `block_height` before returning, the same tripwire shape as the epoch-start identity guard in tn-reth.
- New test `adjust_base_fees_errors_when_tip_is_not_a_closing_block`: registry deployed but no closing block executed, so the identity read succeeds against a genesis tip with epoch 0's `blockHeight` of 0 and the guard must trip. Asserts the identity-violation error and that worker fees stay untouched, since the guard fires before the config read.

closes #1019

### DIFF
--- a/crates/node/src/manager/node/run_epoch.rs
+++ b/crates/node/src/manager/node/run_epoch.rs
@@ -754,14 +754,14 @@ async fn adjust_base_fees(
     };
 
     let block_height = epoch_info.blockHeight;
-    debug_assert_eq!(
-        tip.number + 1,
-        block_height,
-        "close-time base-fee identity: canonical tip {} + 1 != entered-epoch {entered_epoch} \
-         blockHeight {block_height}",
-        tip.number,
-    );
     if tip.number + 1 != block_height {
+        error!(
+            target: "epoch-manager",
+            tip_number = tip.number,
+            entered_epoch,
+            block_height,
+            "close-time base-fee identity: canonical tip + 1 != entered-epoch blockHeight"
+        );
         return Err(eyre::eyre!(
             "close-time base-fee identity violated: canonical tip {} + 1 != entered-epoch \
              {entered_epoch} blockHeight {block_height} at tip {:?} - refusing to price base fees \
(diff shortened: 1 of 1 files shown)


MavenRain (approved):
LGTM

sstanfield (approved):

PR #1027, opened 2026-07-23
Rename engine committee read to `committee_keys_for_epoch_at_block`

### LINKED ISSUE #1020: Engine's `validators_for_epoch_at_block` returns BLS keys and collides with a reth method of the same name
## Problem

The engine deliberately exposes a single committee read: a pinned method named `validators_for_epoch_at_block`. Despite the name, it does not return validators — it reads the committee's BLS pubkeys from the ConsensusRegistry at the pinned block and returns `Vec<BlsPublicKey>`. Since it is the engine's only committee read, this name fronts every committee read the epoch manager makes.

One layer down, the reth environment has a method with the exact same name that does return validator records: its `validators_for_epoch_at_block` reads the registry's committee-validators call and returns `Vec<ValidatorInfo>`. Same name, same arguments, different return types, one layer apart. The engine method is not even a wrapper over its reth namesake; it delegates to the reth layer's BLS-pubkeys read, so the shared name suggests a delegation relationship that does not exist.

This is a standing hazard for anyone reading or refactoring across the engine/reth boundary: a search for the name lands in both layers, and confusing the two changes the result from bare keys to full validator records. The existing call sites already contradict the name — the epoch close and entry paths bind the result to `committee_keys`, `next_committee_keys`, and `previous_committee_keys`, so every use has to correct the name at the binding.

## Solution

Rename the engine-layer method to `committee_keys_for_epoch_at_block`: the name should say what the caller receives — the committee's BLS keys, pinned at a block. The rename should cover the public wrapper, the internal definition, and the five call sites in the epoch close and entry paths, and the doc comments should say "BLS keys" rather than "validator keys".

The reth-layer `validators_for_epoch_at_block` should keep its name: it returns validator records, so name and return type already agree. This is a rename only — zero behavior change.

grantkee:
- The engine's pinned committee read was named `validators_for_epoch_at_block` but returns `Vec<BlsPublicKey>`; reth has a different method with the same name that returns `Vec<ValidatorInfo>`. Same name, different return types, one layer apart — a trap for anyone reading or grepping across the boundary.
- The old name also implied a delegation that never existed: the engine method wraps reth's `bls_pubkeys_for_epoch_at_block`, not its `validators_for_epoch_at_block` namesake.
- Rename covers the engine definition, its public wrapper, and the five call sites in the epoch close/entry paths; doc comments now say "BLS keys" instead of "validator keys".
- Call sites already bound the result to `committee_keys`, `next_committee_keys`, and `previous_committee_keys`; the method name now matches its bindings.
- Reth's `validators_for_epoch_at_block` keeps its name: its return type matches it. Zero behavior change.

closes #1020

### DIFF
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -382,8 +382,8 @@ impl ExecutionNodeInner {
         self.reth_env.epoch_state_at_epoch_start()
     }
 
-    /// Read committee validator keys for epoch, pinned to the block identified by `block_hash`.
-    pub(super) fn validators_for_epoch_at_block(
+    /// Read the committee's BLS keys for `epoch`, pinned to the block identified by `block_hash`.
+    pub(super) fn committee_keys_for_epoch_at_block(
         &self,
         epoch: u32,
         block_hash: B256,
--- a/crates/node/src/engine/mod.rs
+++ b/crates/node/src/engine/mod.rs
@@ -305,18 +305,18 @@ impl ExecutionNode {
         guard.epoch_state_at_epoch_start()
     }
 
-    /// Read committee validator keys for epoch, pinned to the block identified by `block_hash`.
+    /// Read the committee's BLS keys for `epoch`, pinned to the block identified by `block_hash`.
     ///
     /// This is deliberately the ONLY committee read the engine exposes: an unpinned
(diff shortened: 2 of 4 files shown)


sstanfield (approved):

MavenRain (approved):
LGTM

Issue #738, opened 2026-06-11
Nodes expose no operational metrics; `--metrics` is an accepted no-op flag


grantkee:
## Problem

A telcoin-network node runs a Narwhal/Bullshark consensus stack alongside a reth EVM
execution engine, but emits no machine-readable metrics. Operators have no way to watch
round progression, certificate formation, batch sealing, epoch boundaries, consensus sync
distance, or basic process health. Diagnosing a stalled validator, a flapping node, or
sync lag relies on scraping logs by hand.

The CLI already advertises a `--metrics <SOCKET>` flag, which makes the gap worse: the
flag is parsed and threaded through node construction, but nothing consumes it. Enabling
it does nothing and returns no error, so an operator who sets it reasonably assumes metrics
are being served when they are not. A leftover `PrometheusMetricsParameters` config type
and an unused `axum` "consensus metrics" dependency are remnants of an earlier attempt that
was never finished.

reth's own instrumentation compounds the problem. reth ships rich built-in metrics for the
database, transaction pool, and provider, but they are invisible here because no global
`metrics` recorder is ever installed — every metric handle binds to the default noop
recorder and is silently discarded.

The result is a validator that cannot be monitored or alerted on in production. There is no
dashboard story, no way to define SLOs, and no signal short of log archaeology when a node
misbehaves.

## Solution

Introduce a `tn-metrics` crate that owns the two pieces every node needs: a global
Prometheus recorder and a minimal scrape endpoint. The recorder should apply a *selective*
prefix — names starting with `tn` pass through untouched, everything else (reth's
instrumentation and process stats) gets a `reth_` prefix — so reth metrics render
identically to a stock reth node and upstream dashboards work unchanged, while telcoin
metrics keep clean `tn_*` names.

The recorder must be installed at the very top of the `node` command, before any reth
component is constructed (in particular before the reth database). Derive-style metric
handles bind to whichever recorder is live at construction time; anything registered
before installation is lost to the noop recorder. Nodes started without `--metrics` should
skip installation entirely and keep zero overhead.

Each consensus and execution crate should be instrumented with `reth_metrics::Metrics`
derives under a `tn_<scope>` namespace (worker, batch-builder, primary, executor, engine,
network, state-sync, epoch, node), following a strict low-cardinality rule: no peer ids,
digests, or epochs as label values; round/height/epoch are gauge *values*. Consensus state
that already lives in watch channels (round, committed round, heights, node mode) should be
[...]

PR #1035, opened 2026-07-25
feat(worker): surface a stalled batch fetch with a rate-limited warning

### LINKED ISSUE #865: worker: `BatchFetcher::fetch_for_primary` spins in a zero-backoff loop (DB reads + empty write-txn commits + network round-trips) whenever it has batches to fetch but no connected worker peers
## Summary

When a worker needs to fetch batches referenced by committed consensus output and has no connected worker peers, `BatchFetcher`'s fetch loop spins with no backoff.  The outer loop in `fetch_for_primary_inner` has no delay of its own, and it relies on `request_batches` for backoff, but `request_batches` returns `Ok(empty)` immediately when the connected-peer set is empty, short-circuiting the 500ms retry sleep it would otherwise apply.  The loop therefore repeats as fast as the runtime will schedule it: each iteration does a local-DB multi-get, opens and commits an empty RocksDB write transaction, and issues a `connected_peers` command to the network manager, then loops again because the digests are still missing.

The trigger is an ordinary, non-adversarial condition: a node that has certified output to hydrate but whose worker network has not yet connected to (or has been partitioned from) its peers.  While it holds, the loop consumes CPU and I/O continuously with no rate limit, hammering the batch store with empty commits and the network manager with `connected_peers` requests, per pending fetch.  An attacker who can transiently drop a target's worker-peer connections can convert a brief outage into sustained CPU and I/O amplification for its duration.  It self-resolves once a worker peer connects, so severity is a reliability / resource-exhaustion concern rather than a consensus-safety one.

## The defect

**Locations:** `crates/consensus/worker/src/batch_fetcher.rs:154-210` (the outer loop) and `crates/consensus/worker/src/network/handle.rs:169-174` (the empty-peer short-circuit).

**Outer loop has no backoff** (`batch_fetcher.rs:154-210`):

```rust
// loop until `missing_digests` empty
loop {
    if missing_digests.is_empty() {
        return Ok(fetched_batches);
    }

    // 1) fetch from local storage
    self.fetch_local(&mut missing_digests, &mut fetched_batches).await?;
    if missing_digests.is_empty() {
        return Ok(fetched_batches);
    }

    // 2) fetch from peers over network
    // NOTE: `request_batches` performs bounded retries (3 attempts with 500ms backoff)
    // internally, so RPCError here means all retries were exhausted. ...
    if let Ok(mut new_batches) = self.request_batches_from_peers(&mut missing_digests).await
    {
        /* ... open write txn, insert new_batches (possibly none), commit ... */
        let mut txn = self.batch_store.write_txn()?;
        for (digest, batch) in new_batches.iter_mut() { /* ... */ }
        txn.commit()?;                                   // committed even when new_batches is empty
        /* ... */
    }
    // loops again with no sleep
}
```

The loop's only source of pacing is the comment's assumption that `request_batches` applies "3 attempts with 500ms backoff" internally.  That assumption does not hold when there are no connected peers.

**`request_batches` returns immediately on an empty peer set** (`handle.rs:169-174`), before it ever reaches the retry sleep at `handle.rs:243`:

```rust
for attempt in 0..MAX_BATCH_REQUEST_RETRIES {          // handle.rs:169
    // re-fetch peers each attempt to pick up newly connected peers
    let peers = self.handle.connected_peers().await?;
    if peers.is_empty() {
        return Ok(all_batches);                        // handle.rs:172-173: returns Ok(empty), no sleep
    }
    /* ... try peers, then ... */
    if attempt + 1 < MAX_BATCH_REQUEST_RETRIES {
        tokio::time::sleep(BATCH_REQUEST_RETRY_DELAY).await;   // handle.rs:243 (never reached when peers is empty)
    }
}
```

`request_batches_from_peers` forwards that empty `Ok` verbatim (`batch_fetcher.rs:253-256`):

```rust
let recovered_batches =
    self.network.request_batches(missing_digests).await?.into_iter().collect();
Ok(recovered_batches)
```

So with zero connected worker peers and a non-empty `missing_digests`, `request_batches` returns `Ok(empty)` on the first attempt without sleeping, `request_batches_from_peers` returns `Ok(empty)`, the `if let Ok(...)` arm at `batch_fetcher.rs:177` runs with an empty map (opening and committing an empty write transaction), `missing_digests` is still non-empty, and the outer loop repeats immediately.

**Effect per iteration, with no delay between iterations:** a `fetch_local` multi-get against the batch store, a `write_txn` + empty `commit` against RocksDB (`batch_fetcher.rs:182-200`), and a `connected_peers` command round-trip to the network manager (`handle.rs:171`).  The loop yields at these await points (so it does not starve the executor), but it runs at the maximum rate the runtime allows, with no upper bound, for as long as the no-peer condition persists, and once per pending `fetch_for_primary`.

## Trigger conditions

`fetch_for_primary` hydrates the batches a committed sub-DAG references but the node does not have locally.  It enters the spin whenever `missing_digests` is non-empty and `connected_peers()` is empty, which happens when:

- a node has just started or restarted and has pending certified output to hydrate before its worker network has connected to peers;
- the worker is network-partitioned from its peers while consensus output is still arriving; or
- an attacker transiently drops (or eclipses) the target's worker-peer connections while it has output to hydrate.

Worker peers are a distinct set from primary peers, so a node can be making primary-side progress while its worker peer set is empty.

## Root cause

`fetch_for_primary_inner` delegates all pacing to `request_batches`, but `request_batches`'s no-peer path is an early `Ok(empty)` return that intentionally skips the retry sleep.  The two together leave the outer loop with no backoff in exactly the case (no peers) where retrying cannot make progress.

## Suggested fix

- Add an explicit backoff in `fetch_for_primary_inner` before re-looping when an attempt made no progress (no new batches fetched), so a persistent no-peer or all-fail condition cannot spin.  A short bounded delay (or exponential backoff capped at a small maximum) is enough.
- Alternatively (or additionally), have `request_batches` distinguish "no peers available" from "peers tried and failed" to the caller (for example a dedicated `Ok`-with-status or a typed error), so the fetcher can wait for a peer-connected signal instead of re-polling.  Skip the empty write-transaction commit when there is nothing to insert.
- Consider gating the fetch loop on a "worker peers available" notification rather than re-polling `connected_peers` unbounded.

## Suggested test design

Add a test in `crates/consensus/worker` that drives `fetch_for_primary` (or `fetch_for_primary_inner`) with a non-empty `missing_digests` and a network handle whose `connected_peers()` returns empty, and asserts that the loop does not iterate more than a bounded number of times within a fixed interval (for example by counting `connected_peers` calls or `write_txn` opens over a short window), rather than spinning at full rate.

## Acceptance criteria

- With batches to fetch and no connected worker peers, the fetcher backs off between attempts instead of looping at the maximum rate.
- No empty write transaction is committed on an attempt that fetched no batches.
- A regression test covers the no-peer case and bounds the loop rate.

---

Note: this defect surfaced during an AI-assisted review of the worker batch-fetch path, with the cited locations verified against current `main`.  It is distinct from the batch-stream reader hardening in the worker network codec (a different function and trigger).

MavenRain:
## Summary

When the worker `BatchFetcher` is asked for certified batches it cannot resolve locally or from any peer (for example under a network partition or eclipse), `fetch_for_primary_inner` retries forever by design.  That retry is the correct liveness-preserving choice . . . a certified batch is persisted by at least `f + 1` honest peers, so the fetch will succeed once connectivity returns, and erroring out instead would shut the node down.  The gap is observability: today the stall is completely silent.  #944 (closes #865) added backoff pacing so the loop no longer spins, and #889 added a `record_batch_fetch_duration` metric, but that duration is recorded only after a fetch completes, so an in-flight stall produces no signal at all.

This adds a rate-limited `warn!` that fires while a fetch is making no progress, so an operator can see a stuck node instead of a node that looks healthy (it keeps voting and certifying, since the executor is downstream of consensus) while never finalizing.

## Changes

`crates/consensus/worker/src/batch_fetcher.rs`:

- [x] Count consecutive no-progress passes in `fetch_for_primary_inner` (`consecutive_stalls`, reset to `0` whenever a pass resolves at least one digest, incremented with `saturating_add` otherwise).
- [x] Emit a rate-limited `warn!` (target `batch_fetcher`, fields `remaining` and `consecutive_stalls`) once at `FETCH_STALL_WARN_AFTER` (10 passes, roughly 10s once the backoff has saturated at the 1s cap) and then every `FETCH_STALL_WARN_INTERVAL` (30) passes thereafter, so a prolonged stall keeps signalling without flooding the log.
- [x] Factor the cadence into a pure `should_warn_stall` function so it is unit-testable without a tracing subscriber.

## Threat model / design notes

- **Non-fatal by construction.**  The stall is only logged;  the fetch still never returns `Err` on a stall.  A timeout or attempt cap was deliberately not added: a fetch `Err` propagates to a fatal node shutdown (`executor/src/subscriber.rs` returns the error out of the critical `subscriber consensus` task), so a hard deadline would convert a self-recovering stall into a guaranteed crash.
- **Why retrying forever is safe.**  A primary only votes on a header after `sync_header_batches` (itself bounded by a 30s `RetryConfig` plus a committed-round abort) has made every referenced batch locally available, so a 2f+1 certificate implies at least `f + 1` honest peers hold every committed digest.  The missing batch is therefore recoverable once the partition heals;  the warning just makes the wait visible.
- **No attacker required.**  The condition this surfaces is an honest liveness stall (unreachable peers), not an exploit.

## Testing

`cargo +1.94 test -p tn-worker --lib stall_warn_fires_at_threshold_then_every_interval` (pinned toolchain, `-j 2`).

The new `stall_warn_fires_at_threshold_then_every_interval` test pins the cadence: silent for every count below the threshold (including a reset counter at `0`, which satisfies `0 % INTERVAL == 0` and would spuriously warn without the `> FETCH_STALL_WARN_AFTER` guard), a single warning exactly at the threshold, silence between the threshold and the interval multiples above it, and a warning on each interval multiple beyond it.  Confirmed by mutation: raising `FETCH_STALL_WARN_AFTER` (and, separately, deleting the interval clause of `should_warn_stall`) makes the test fail;  reverting restores green.

## Out of scope

An in-flight `batch_fetch_stalled` gauge belongs with the metrics work in #738 (`--metrics` is currently an accepted no-op), so it is intentionally left out here;  a `warn!` is chosen because a log line is useful today without depending on that infrastructure landing first.

### DIFF
--- a/crates/consensus/worker/src/batch_fetcher.rs
+++ b/crates/consensus/worker/src/batch_fetcher.rs
@@ -14,7 +14,7 @@ use std::{
 };
 use tn_storage::{consensus::ConsensusChain, tables::NodeBatchesCache};
 use tn_types::{now, Batch, BlockHash, Database, DbTxMut, Epoch, B256};
-use tracing::{debug, error, instrument};
+use tracing::{debug, error, instrument, warn};
 
 /// Minimum delay before retrying a [`BatchFetcher::fetch_for_primary`] pass that
 /// recovered no batches (locally or from a peer).
@@ -30,6 +30,34 @@ const FETCH_RETRY_MIN_BACKOFF: Duration = Duration::from_millis(50);
 /// Caps how long a persistent no-peer condition waits before re-checking, so a
 /// worker peer that (re)connects is still picked up promptly.
 const FETCH_RETRY_MAX_BACKOFF: Duration = Duration::from_secs(1);
+
+/// Number of consecutive no-progress fetch passes in [`BatchFetcher::fetch_for_primary_inner`]
+/// before the first stall warning is emitted.
+///
+/// At [`FETCH_RETRY_MAX_BACKOFF`] (1s) per saturated pass this is roughly ten seconds of a fetch
+/// making no progress, which on a healthy network only happens when the worker cannot reach any
+/// peer holding the batches (a self-recovering eclipse). The warning is purely observational: the
+/// fetch never fails on a stall, because erroring out would shut the node down. A certified batch
+/// is persisted by at least `f + 1` honest peers, so retrying until connectivity returns is the
+/// correct liveness-preserving choice; this log only makes an otherwise silent stall visible.
+const FETCH_STALL_WARN_AFTER: u32 = 10;
+
+/// Once a stall has exceeded [`FETCH_STALL_WARN_AFTER`], re-emit the warning every this many
+/// further consecutive no-progress passes, so a prolonged stall keeps signalling without flooding
+/// the log on every pass.
+const FETCH_STALL_WARN_INTERVAL: u32 = 30;
+
+/// Whether [`BatchFetcher::fetch_for_primary_inner`] should emit a stall warning after
+/// `consecutive_stalls` consecutive no-progress passes: once exactly at [`FETCH_STALL_WARN_AFTER`],
+/// then on every [`FETCH_STALL_WARN_INTERVAL`]th pass beyond it. Kept as a pure function so the
+/// cadence is unit-testable without installing a tracing subscriber. The `> FETCH_STALL_WARN_AFTER`
+/// guard is what stops a reset counter (`0`, which also satisfies `0 % INTERVAL == 0`) from
+/// warning.
+fn should_warn_stall(consecutive_stalls: u32) -> bool {
+    consecutive_stalls == FETCH_STALL_WARN_AFTER
+        || (consecutive_stalls > FETCH_STALL_WARN_AFTER
+            && consecutive_stalls.is_multiple_of(FETCH_STALL_WARN_INTERVAL))
+}
 
 #[derive(Debug)]
 /// The type to fetch batches for the primary.
@@ -170,6 +198,10 @@ impl<DB: Database> BatchFetcher<DB> {
         // all-peers-failed condition backs off instead of spinning (see issue #865)
         let mut backoff = FETCH_RETRY_MIN_BACKOFF;
 
+        // count consecutive passes that recovered nothing, so a prolonged stall (peers holding
+        // these batches unreachable) is surfaced to operators via a rate-limited warning below.
+        let mut consecutive_stalls: u32 = 0;
+
         // loop until `missing_digests` empty
         loop {
             debug!(target: "batch_fetcher", "loop start");
@@ -243,7 +275,18 @@ impl<DB: Database> BatchFetcher<DB> {
             // without delaying a pass that is still fetching batches (see issue #865).
             if missing_digests.len() < missing_before {
                 backoff = FETCH_RETRY_MIN_BACKOFF;
+                consecutive_stalls = 0;
             } else {
+                consecutive_stalls = consecutive_stalls.saturating_add(1);
+                if should_warn_stall(consecutive_stalls) {
+                    warn!(
+                        target: "batch_fetcher",
+                        remaining = missing_digests.len(),
+                        consecutive_stalls,
+                        "batch fetch making no progress; worker peers holding these batches may be \
+                         unreachable (node stays up and keeps retrying)"
+                    );
+                }
                 tokio::time::sleep(backoff).await;
                 backoff = backoff.saturating_mul(2).min(FETCH_RETRY_MAX_BACKOFF);
             }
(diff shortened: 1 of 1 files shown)


grantkee (approved):
Great thinking

sstanfield (approved):

PR #1038, opened 2026-07-25
consensus: add a canonical-DB fallback to wait_for_execution's ring-miss arm (#1036)

### LINKED ISSUE #1036: consensus: wait_for_execution can misclassify a ring-evicted-but-persisted execution tip as a fork; add a canonical-DB fallback on the miss arm
## Summary

`ConsensusBusApp::wait_for_execution` decides "have we executed this block, or have we forked?" purely from the in-memory `RecentBlocks` ring. On the miss arm it returns `Err`, and two callers treat that `Err` as a confirmed execution fork and bring the node down. The ring only holds the last `gc_depth` (default 50) consensus-output tips, so the miss arm silently assumes every queried target is still resident in that window. That assumption is upheld today by caller-side invariants, not by `wait_for_execution` itself. This proposes making `wait_for_execution` self-sufficient: on a ring miss where `target_number <= local canonical tip`, consult the canonical DB before declaring a fork.

This is defense-in-depth. It is **not** a live bug on current `main` (see "Why this is not exploitable today"). The goal is to remove a brittle latent edge, not to fix an observed failure.

## Current behavior (HEAD `addaf849`)

`crates/consensus/primary/src/consensus_bus.rs:665-689`:

```rust
pub async fn wait_for_execution(&self, block: BlockNumHash) -> Result<(), WaitForExecutionElapsed> {
    let mut watch_execution_result = self.recent_blocks().subscribe();
    let target_number = block.number;
    while self.recent_blocks().borrow().is_empty() { watch_execution_result.changed().await?; }
    let mut current_number = self.latest_execution_block_num_hash().number;
    while current_number < target_number {
        watch_execution_result.changed().await?;
        current_number = self.latest_execution_block_num_hash().number;
    }
    if self.recent_blocks().borrow().contains_execution_hash(block.hash) {
        Ok(())
    } else {
        Err(WaitForExecutionElapsed()) // <- treated as "we forked"
    }
}
```

`contains_execution_hash` (`crates/consensus/primary/src/recent_blocks.rs:82-89`) scans only the in-memory `executed_blocks: VecDeque<SealedHeader>` ring, which is capacity-bounded to `num_blocks = gc_depth` (wired at `crates/node/src/manager/node.rs:763`; `default_gc_depth() == MAX_GC_DEPTH == 50`). There is no DB or `ConsensusChain` fallback on the miss.

Fatal consumers of the `Err`:

- `crates/state-sync/src/lib.rs:415-427`: logs `"wait_for_execution failed - execution fork detected"` and returns `Err(eyre!(...))`, halting the follow loop (node dies).
- `crates/consensus/primary/src/consensus/state.rs:421-422`: same pattern on the commit path.

Non-fatal consumer (fine as is): `crates/consensus/primary/src/network/handler.rs:714` uses the `Err` only to decline voting on a peer header.

## Why this is not exploitable today

The miss-then-die outcome is currently unreachable for a genuinely-executed (non-forked) block, because the query target is always a recent output tip:

- One `EngineUpdate` per consensus output. The per-batch loop in `execute_consensus_output` (`crates/engine/src/payload_builder.rs:151-206`) builds N EVM blocks but sends nothing; `finish_executing_output` (`crates/tn-reth/src/lib.rs:988-999`) then sends exactly one `EngineUpdate` carrying `Some(chain_update.tip())`, the single final tip. The empty-skip path sends one `None` (`payload_builder.rs:97`). So one output pushes at most one `executed_blocks` entry (`node.rs:1327`), regardless of sub-dag size.
- Every fatal caller queries `leader().latest_execution_block()` (an output tip), never an intermediate EVM block.
- `wait_for_execution` first blocks until `current_number >= target_number` with no yield before the hash check, and the follow loop skips already-processed outputs (`state-sync/src/lib.rs:402`) before ever reaching the fork check.

Net: the target is always at or near the ring tip, so it is present when checked. (This is the basis for classifying the previously-raised "sub-dag of >50 EVM blocks evicts a real block" concern as a false positive.)

## The latent risk this hardening removes

The safety above lives entirely in the callers. `wait_for_execution`'s own contract, "a hash I cannot find in the last 50 tips means we forked", is stronger than the truth: a block can be absent from the ring yet fully present and canonical in the DB. It becomes a live false-positive fork (and a node halt) the moment any of the guarding invariants shifts, for example:

- a new or changed caller that queries a tip older than 50 outputs back (a lagging follower, a longer catch-up window, a batched/asynchronous verify path);
- a change to the restart seed (`try_restore_state` currently reseeds the last 50 output tips) or to how far behind the engine may drift from the caller;
- lowering `gc_depth` (see #954, which is adding lower/cross-field bounds to `gc_depth`) shrinks the ring window directly.

Any of these turns a benign ring eviction into "execution fork detected" and a self-inflicted node death, with the correct block sitting in the local DB the whole time.

## Proposed change

On the miss arm, before returning the fork error, fall back to the canonical DB: if `target_number <= local canonical tip number` and the DB's canonical block at `target_number` hashes to `block.hash`, return `Ok(())`. Only a genuine hash mismatch (or `target_number` above the canonical tip) is a real fork.

Two placements, either acceptable:

1. Inside `wait_for_execution`, by threading a by-number canonical reader into `ConsensusBusApp`. `RethEnv` already exposes header lookups (for example `sealed_header_by_hash`), so a canonical-hash-by-number read is available or a small addition. This fixes every current and future caller in one place (preferred).
2. At the two fatal call sites only (`state-sync/src/lib.rs`, `consensus/state.rs`), which already hold chain/engine handles, gating the "die" decision on a DB confirmation. Smaller blast radius, but leaves the weak contract in `wait_for_execution` for the next caller.

A rounds-based ring capacity was considered as an alternative but only widens the window; it does not make `wait_for_execution` authoritative against the DB, so the DB fallback is preferred.

## Scope / non-goals

- No behavior change on the happy path or on a genuine fork (hash mismatch still fails hard).
- Not urgent: this guards a currently-unreachable edge. Reasonable to defer or to fold into the #954 `gc_depth` bounds work, since that change is the most likely near-term trigger.
- Tests: a `wait_for_execution` unit test where the target is evicted from the ring but present in the canonical DB should return `Ok`; a target with a mismatching DB hash should still `Err`.

MavenRain:
Closes #1036

## Summary

`ConsensusBusApp::wait_for_execution` decides "have we executed this block, or have we forked?" purely from the in-memory `recent_blocks` ring. On a ring miss it returns `Err`, and two callers treat that error as a confirmed execution fork and bring the node down. The ring only holds the last `gc_depth` (default `MAX_GC_DEPTH == 50`) executed tips, so the miss arm silently assumes every queried target is still resident in that window. That assumption is upheld today by caller-side invariants, not by `wait_for_execution` itself.

This makes `wait_for_execution` self-sufficient: on a ring miss it consults the canonical execution DB before declaring a fork. A block that is absent from the ring but persisted as canonical at its height is a stale ring, not a fork.

This is defense-in-depth. It is **not** a live bug on current `main` (the query target is always a recent output tip, so it is present when checked). The goal is to remove a brittle latent edge, not to fix an observed failure. Closes #1036.

## Problem

`wait_for_execution`'s own contract, "a hash I cannot find in the last 50 tips means we forked", is stronger than the truth: a block can be absent from the ring yet fully present and canonical in the DB. The two fatal consumers turn that error into a self-inflicted node halt:

- `crates/state-sync/src/lib.rs`: logs `"wait_for_execution failed - execution fork detected"` and returns `Err`, halting the follow loop.
- `crates/consensus/primary/src/consensus/state.rs`: same pattern on the commit path (goes `CvvInactive` and notifies shutdown).

(The third, non-fatal consumer in `network/handler.rs` only declines to vote on a peer header and is left unchanged.)

The safety that makes this unreachable today lives entirely in the callers (one `EngineUpdate` per output, callers always query the leader's `latest_execution_block()` tip, the follow loop blocks until `current_number >= target_number` before the hash check). It becomes a live false-positive fork the moment any guarding invariant shifts, for example a lagging follower querying a tip older than 50 outputs back, a change to the restart reseed window, or lowering `gc_depth` (see #954, which is adding bounds to `gc_depth` and is the most likely near-term trigger).

## Fix

A narrow read-only capability, installed into the bus once the engine exists:

- **`tn_types::CanonicalExecutionReader`** (new): `fn canonical_execution_hash(&self, number) -> Option<BlockHash>`. `None` means "above the persisted tip or unreadable" (not confirmed); a returned hash is always the canonical one at that height. Kept in `tn-types` so the consensus crates depend on the capability alone, not on the execution environment (`tn-reth` is not a runtime dependency of `tn-primary`).
- **`impl CanonicalExecutionReader for RethEnv`** (in `tn-reth`) via the existing `sealed_header_by_number(n)` read; a transient read error maps to `None` (conservative: the caller keeps its fork handling).
- **`ConsensusBusApp`** gains a late-bound `CanonicalReaderSlot` (a `parking_lot::Mutex<Option<Arc<dyn CanonicalExecutionReader>>>`). `wait_for_execution`'s miss arm now returns `Ok` only when the DB confirms the exact hash at the target height; a hash mismatch or an absent/above-tip block still fails hard.
- The node manager installs the reader (`set_canonical_reader`) in `create_engine`, right after `RethEnv::new`.

**Why this placement.** Issue #1036 floated two options: (1) thread a reader into `ConsensusBusApp`, or (2) fix the two fatal call sites. Investigating the current tree showed both literal forms need adjustment: the bus is constructed **before** `RethEnv` exists (`node.rs` builds the bus from `gc_depth`, then builds `RethEnv` later in `create_engine`), and **neither** fatal caller can reach a `RethEnv` (state-sync has no `tn-reth` dependency; `tn-primary` has `tn-reth` only as a test-utils dep). The bus is the one conduit already threaded to all three call sites, so a late-injected reader on the bus is the true central fix (option 1's intent), and it uses last-writer-wins so a per-epoch engine rebuild just refreshes the handle. `Arc<dyn ...>` matches house style (trait objects are used across the workspace); the trait carries no `Debug` bound because `RethEnv`'s reader handle need not be `Debug`, and the slot supplies a manual `Debug` so `ConsensusBusAppInner` still derives it.

## Testing

Three deterministic `#[tokio::test]`s in `consensus_bus_tests.rs`, using an in-memory `FakeCanonicalReader` (no MDBX): the ring holds a later tip so `wait_for_execution` reaches the hash check, and the target hash is absent from the ring.

- `..._confirms_ring_evicted_block_via_canonical_db`: DB confirms the evicted tip -> `Ok`. **Reverting the new `else if` branch makes this test fail** (non-vacuous).
- `..._reports_fork_when_canonical_hash_differs`: DB has a different hash at the height (genuine fork) -> `Err`.
- `..._without_reader_keeps_ring_only_fork_check`: no reader installed -> original ring-only `Err` (regression guard).

Happy path and genuine-fork behavior are unchanged.

## Threat model

A dishonest peer cannot use this to smuggle a forked block past the fork check: confirmation requires the queried hash to equal the *canonical* hash the local node itself persisted at that height, read from the local DB. A real fork (different hash) or a not-yet-persisted block (`None`) still fails hard. The change only ever converts a false-positive halt into correct progress on the node's own canonical data.

## Scope / non-goals

- No behavior change on the happy path or on a genuine fork.
- Widening the ring capacity was rejected as an alternative: it only moves the window, it does not make `wait_for_execution` authoritative against the DB.

## Pre-push gate (house checklist)

- [ ] `make attest` on the pinned `+1.94` toolchain (fmt + clippy + test-build). **Deferred: not run in-session (disk floor).** Run before pushing. `cargo fmt` may reorder the new `use tn_types::CanonicalExecutionReader;` line in `tn-reth`.
- [x] Doc comments on every new item (trait, method, newtype, fields).
- [x] House combinators (no `match` on `Option`/`Result`; `.inspect_err().ok().flatten().map(...)`, `.and_then(...).is_some_and(...)`).
- [x] Deterministic tests (fixed hashes, no timing/sleeps).
- [x] Branch off `origin`, not the fork.


sstanfield (dismissed):

grantkee (dismissed):
LGTM - merge conflicts

grantkee (approved):

PR #1049, opened 2026-07-26
consensus: harden always-active assert/expect panics on the commit hot path and startup recovery (defense-in-depth)

### LINKED ISSUE #1043: consensus: harden always-active assert!/expect panics on the commit hot path and startup recovery (defense-in-depth)
## Summary

The consensus commit path and startup recovery contain several always-active (not `debug_assert`) `assert_eq!` / `.expect` panics.  Each encodes a genuine internal invariant, and under the standard `< f` Byzantine fault assumption I could not construct an input that makes any of them fire on current `main` (`746d59cb`).  This is therefore not a vulnerability report.  It proposes converting these panics into structured, diagnosable handling as defense in depth, for three reasons:

1. They all run on the single consensus task, which is spawned via `spawn_critical_task`.  Any resolution of a critical task takes the whole node down (see next section), so a panic there is a node-level event.
2. A future refactor that breaks one of these construction invariants would turn a diagnosable condition into a hard panic and unwind with poor operator diagnostics.
3. The codebase already handles an analogous "this should not happen" condition gracefully (the `wait_for_execution` bogus-sub-dag arm at `crates/consensus/primary/src/consensus/state.rs:422-434`), so these panics are the inconsistent outliers.

## Why a panic buys nothing over a clean shutdown here

`spawn_critical_task` (`crates/types/src/task_manager.rs:170-179`) tracks the task as critical;  its doc says "When the task resolves, other tasks will shutdown."  The supervisor (`task_manager.rs:431-488`) breaks its loop and calls `shutdown.notify()` for a critical task on **every** resolution kind: a returned `Err` (`CriticalExitError`), a panic (`CriticalJoinError`), and even a returned `Ok` (`CriticalExitOk`).  The consensus task is spawned at `state.rs:347-349` as `Ok(s.run(rx_new_certificates).await?)`.

Consequence: for these sites a panic provides no availability benefit over a clean, structured shutdown.  It only costs an unwind and a low-context backtrace (`assert_eq!(leader.round() % 2, 0)` prints two integers and nothing else).  The one thing that changes the node's post-shutdown classification is the `NodeMode::CvvInactive` flip that the existing graceful arm performs before `shutdown().notify()`.

## Sites (current `main`, `746d59cb`)

Hot loop, consensus task:

- `crates/consensus/primary/src/consensus/bullshark.rs:101`  `assert_eq!(reputation_score.total_authorities() as usize, self.committee.size());`  Invariant: the reputation map always carries exactly the committee.  Holds today because every committed certificate's `origin()` is committee-checked upstream (`Certificate::validate_and_verify` -> `Header::validate`, `crates/types/src/primary/header.rs:133-136`) and `Committee::new` asserts distinct authority ids.
- `bullshark.rs:196-197`  `state.dag.get(&(leader_round + 1)).expect("We should have the whole history by now")`.  Invariant: the child round of a committable leader is present in the DAG.  Holds today by GC math (`gc_round = commit_round.saturating_sub(gc_depth)`;  on a schedule-change retry `committed_round < leader_round`).
- `bullshark.rs:289`  `assert_eq!(leader.round() % 2, 0);`  Invariant: leaders are elected only on even rounds.
- `state.rs:470`  `assert_eq!(self.state.last_round.committed_round, leader_commit_round);`  Invariant: both equal the max round of the sub-dag(s) just committed in this call.

Startup / recovery, `Consensus::spawn`:

- `state.rs:308-314`  `assert_eq!(sub_dag.leader_round(), last_committed_round, ...)`.  This is an internal self-consistency tripwire, not a durability check, and despite how it reads it does not compare two independently-persisted stores.  Both operands derive from a single append-only store, the current epoch's `ConsensusPack` (`consensus_pos_idx` plus `data`): `read_last_committed` has no dedicated table, it reconstructs the per-authority max by scanning the same `consensus_pos_idx` (`crates/storage/src/consensus_pack.rs:1211-1228`) that `latest_consensus_header_from_pack` reads for its latest entry (`consensus_pack.rs:1202-1208`), and both resolve to the same `current_pack()` (`crates/storage/src/consensus.rs:1052`, `:1068`).  A crash cannot tear the two apart: a torn tail is reconciled in lockstep at pack open by `trunc_and_heal` (`consensus_pack.rs:808`, `:851`, healing at `:686-718`), so both reads observe the identical healed boundary, and the equality itself (max round == leader round) is a Bullshark construction property (the leader is the highest-round header in its sub-dag, hard-asserted at `crates/types/src/primary/output.rs:377`), not a persistence property.  A concurrent-writer race is also ruled out on restart, because the current-epoch pack writer starts after these reads (`crates/node/src/manager/node/run_epoch.rs:301-315`).  So this assert can only fire on a logic-level regression in sub-dag construction, never an honest single-node crash.

Listed for completeness, likely intentionally fatal and out of scope: `state.rs:74` `.expect("error when recovering DAG from store")` and `state.rs:98` `.expect("database available")` are genuine storage failures.  Converting them would improve only the message, not add recoverability.

## Proposed change

- Startup (`state.rs:308`): optionally return a typed error from `Consensus::spawn` (already `-> Result<(), ConsensusError>`) carrying both rounds, so a construction-invariant regression degrades to a clear, recoverable startup error rather than aborting recovery with an `assert_eq!` backtrace.  This is the weakest case for change: it is foreclosed on any honest crash (single store plus lockstep heal, per the site note above), so it fires only on an internal logic regression, and a hard fail-fast on a broken invariant is a reasonable thing to keep.  Listed for completeness, not as a priority.
- Hot loop (`bullshark.rs:101`, `:197`, `:289`, `state.rs:470`): replace each `assert!` / `.expect` with a structured `tracing::error!` that records the relevant rounds, digests, and committee size, followed by the same graceful path the bogus-sub-dag arm already uses (`node_mode().send_modify(|v| *v = NodeMode::CvvInactive)` then `shutdown().notify()`), or a new `ConsensusError` variant propagated through the existing `?` chain.  `ConsensusError` (`crates/consensus/primary/src/consensus/mod.rs:24-39`) currently has no `Invariant` / `Internal` variant, so a new one would be added.  To be explicit, and to avoid overselling: this does not keep the node running (see the critical-task semantics above).  The benefit is a clean shutdown, a defined `CvvInactive` mode, and a diagnosable log instead of a panic and unwind.

## Testing

There are no `#[should_panic]` tests for these sites (`rg should_panic crates/consensus/primary/src` returns nothing), and the recoverable `CertificateEquivocation` case is already asserted as a returned `Err` in tests (`crates/consensus/primary/src/consensus/tests/consensus_tests.rs:430`, `bullshark_tests.rs:949`).  So converting these sites to error-returns or graceful shutdown keeps the existing suite green.  A targeted regression test can construct a violated-invariant input (for example a recovered sub-dag whose leader round differs from the last committed round) and assert the new typed error or `CvvInactive` flip instead of a panic.

## Non-goals

This is not a security or exploitability claim;  under the `< f` assumption none of these panics is known to be reachable on current `main`.  There is no change to happy-path consensus behavior.  The two DB-failure expects (`state.rs:74`, `:98`) are out of scope beyond message quality.

MavenRain:
Closes #1043.

## Summary

The consensus commit path and startup recovery contain several always-active (not `debug_assert`) `assert_eq!` / `.expect` panics. Each encodes a genuine internal construction invariant, and under the standard `< f` Byzantine fault assumption none is reachable on current `main`. This PR converts them into structured, diagnosable handling as defense in depth: a would-be node panic (with a low-context backtrace) becomes a typed error and a clean, defined shutdown. There is **no security claim and no change to happy-path consensus behavior**.

## Problem

All of these sites run on the single consensus task, spawned via `spawn_critical_task`. Any resolution of a critical task takes the whole node down, so a panic there is already a node-level event — a panic buys nothing over a clean shutdown, it only costs an unwind and a backtrace that prints two integers and nothing else. A future refactor that breaks one of these construction invariants would turn a diagnosable condition into a hard panic with poor operator diagnostics. The codebase already handles the analogous "this should not happen" case gracefully (the bogus-sub-dag arm in `new_certificate`), so these panics are the inconsistent outliers.

Sites addressed (all on `main` @ `746d59cb`):

- `bullshark.rs` reputation map — `assert_eq!(reputation_score.total_authorities(), committee.size())`
- `bullshark.rs` leader support tally — `.expect("We should have the whole history by now")` on the child round
- `bullshark.rs` `order_leaders` — `assert_eq!(leader.round() % 2, 0)`
- `state.rs` commit accounting — `assert_eq!(last_round.committed_round, leader_commit_round)`
- `state.rs` startup recovery in `Consensus::spawn` — `assert_eq!(sub_dag.leader_round(), last_committed_round, ...)`

The two genuine DB-failure expects (`state.rs` DAG-recovery and database-available) are intentionally fatal and left out of scope per the issue.

## Fix

Two mechanisms, matched to each site's access:

- **`Bullshark` protocol sites** (no node-infra access) return a new typed `ConsensusError::Invariant(ConsensusInvariant::…)` and propagate through the existing `?` chain: `process_certificate → new_certificate → run` returns `Err`, and the `spawn_critical_task` supervisor performs the clean shutdown. The former `assert!`/`.expect` become combinator guards: `.then_some(_).ok_or_else(_)`, `.ok_or(_)?`, `.is_multiple_of(2).then_some(()).ok_or(_)?`.
- **`state.rs` commit-accounting site** sits inside `new_certificate` beside the existing bogus-sub-dag arm, so it mirrors that idiom exactly: on mismatch, `tracing::error!` with the operands, flip `NodeMode::CvvInactive`, `shutdown().notify()`, then fall through to the function's trailing `Ok(())` (the notified shutdown ends the run loop on the next `select`). The happy-path `committed_round_updates().send_replace(…)` moves into the match branch, unchanged.
- **`state.rs` startup site** (`spawn` already returns `Result`) becomes a typed `ConsensusInvariant::RecoveredLeaderRoundMismatch` returned via `?`. This is the weakest case for change (foreclosed on any honest crash — both operands derive from one append-only `ConsensusPack` reconciled in lockstep at pack open), included for completeness so a construction regression degrades to a clear startup error instead of an `assert_eq!` backtrace.

New error type (`mod.rs`), consistent with the existing `thiserror`-based `ConsensusError`:

```rust
pub enum ConsensusInvariant {
    ReputationAuthorityCount { actual: usize, expected: usize },
    MissingLeaderChildRound(Round),
    LeaderRoundNotEven(Round),
    RecoveredLeaderRoundMismatch { leader_round: Round, last_committed_round: Round },
}
```

wrapped by `ConsensusError::Invariant(#[from] ConsensusInvariant)`. Each variant records the disagreeing operands so a regression is diagnosable from the error alone. Doc comments on every new item; no new imperative loops or early `return`s (combinators + `?`); exhaustive matches preserved. `variantscan` confirms the additive variant breaks no existing `match` on `ConsensusError` workspace-wide.

To be explicit and avoid overselling: this does not keep the node running. The benefit is a clean, defined shutdown (`CvvInactive` for the commit site), a diagnosable log, and a typed error instead of a panic and unwind.

## Testing

- Added `order_leaders_rejects_odd_round_leader` (`bullshark_tests.rs`): feeds an odd-round certificate — the exact input the former `assert_eq!(leader.round() % 2, 0)` tripped on — and asserts `order_leaders` returns `ConsensusError::Invariant(ConsensusInvariant::LeaderRoundNotEven(_))`. Non-vacuous by construction: under the old code this input panics inside `order_leaders` (a failing test); under the new code it returns the typed error.
- The existing suite stays green: converting these sites to error-returns / graceful shutdown does not touch any happy-path assertion, and the recoverable `CertificateEquivocation` case is still asserted as a returned `Err`.
- `cargo +1.94 check -p tn-primary --tests` compiles clean (lib + test target); `cargo nextest run -p tn-primary order_leaders` → 2 passed (`order_leaders`, `order_leaders_rejects_odd_round_leader`), 0 failed.

The remaining converted sites (reputation count, missing child round, startup mismatch, and the commit-accounting `CvvInactive` flip) are left without dedicated regression tests: their invariants cannot be violated through the public API without contrived internal-state manipulation, and the sibling bogus-sub-dag graceful arm they mirror is itself untested — so this matches the codebase's existing posture rather than introducing brittle white-box tests.

## Non-goals

Not a security or exploitability claim; under `< f`, none of these panics is known to be reachable on current `main`. No change to happy-path consensus behavior. The two DB-failure expects are out of scope beyond message quality.

### DIFF
--- a/crates/consensus/primary/src/consensus/bullshark.rs
+++ b/crates/consensus/primary/src/consensus/bullshark.rs
@@ -1,8 +1,8 @@
 //! Bullshark
 
 use crate::consensus::{
-    utils, ConsensusError, ConsensusInvariant, ConsensusState, Dag, LeaderSchedule, LeaderSwapTable,
-    Outcome,
+    utils, ConsensusError, ConsensusInvariant, ConsensusState, Dag, LeaderSchedule,
+    LeaderSwapTable, Outcome,
 };
 use std::collections::VecDeque;
 use tn_types::{
@@ -244,7 +244,8 @@ impl Bullshark {
             debug!(min_round, "Subdag has {} certificates", num_certificates);
 
             // We resolve the reputation score that should be stored alongside with this sub dag.
-            let reputation_score = self.resolve_reputation_score(state, &sequence, sub_dag_index)?;
+            let reputation_score =
+                self.resolve_reputation_score(state, &sequence, sub_dag_index)?;
 
             let sub_dag: CommittedSubDag = CommittedSubDag::new(
                 sequence,
--- a/crates/consensus/primary/src/consensus/bullshark.rs
+++ b/crates/consensus/primary/src/consensus/bullshark.rs
@@ -1,7 +1,8 @@
 //! Bullshark
 
 use crate::consensus::{
-    utils, ConsensusError, ConsensusState, Dag, LeaderSchedule, LeaderSwapTable, Outcome,
+    utils, ConsensusError, ConsensusInvariant, ConsensusState, Dag, LeaderSchedule, LeaderSwapTable,
+    Outcome,
 };
 use std::collections::VecDeque;
 use tn_types::{
@@ -62,7 +63,7 @@ impl Bullshark {
         state: &mut ConsensusState,
         committed_sequence: &[Certificate],
         sub_dag_index: u64,
-    ) -> ReputationScores {
+    ) -> Result<ReputationScores, ConsensusError> {
         // we reset the scores for every schedule change window, or initialise when it's the first
         // sub dag we are going to create.
         // sub_dag_index is based on epoch and round so / 2 so this check works on commit rounds
@@ -97,10 +98,18 @@ impl Bullshark {
             ((sub_dag_index / 2) + 1).is_multiple_of(self.num_sub_dags_per_schedule as u64);
 
         // Always ensure that all the authorities are present in the reputation scores - even
-        // when score is zero.
-        assert_eq!(reputation_score.total_authorities() as usize, self.committee.size());
-
-        reputation_score
+        // when score is zero. This is a construction invariant: every committed certificate's
+        // origin is committee-checked upstream, so it holds under the `< f` fault assumption.
+        // If a future regression ever breaks it, surface a typed error and shut down cleanly
+        // rather than panicking on the single consensus task.
+        let authorities = reputation_score.total_authorities() as usize;
+        (authorities == self.committee.size()).then_some(reputation_score).ok_or_else(|| {
+            ConsensusInvariant::ReputationAuthorityCount {
+                actual: authorities,
+                expected: self.committee.size(),
+            }
+            .into()
+        })
     }
 
     #[instrument(level = "debug", skip_all, fields(round = certificate.round(), origin = ?certificate.origin()))]
@@ -194,7 +203,7 @@ impl Bullshark {
         let voting_power: VotingPower = state
             .dag
             .get(&(leader_round + 1))
-            .expect("We should have the whole history by now")
+            .ok_or(ConsensusInvariant::MissingLeaderChildRound(leader_round + 1))?
             .values()
             .filter(|(_, x)| x.header().parents().contains(&leader_digest))
             .map(|(_, x)| self.committee.voting_power_by_id(x.origin()))
@@ -212,7 +221,7 @@ impl Bullshark {
         // the last committed leader, and commit all preceding leaders in the right order.
         // Committing a leader block means committing all its dependencies.
         let mut committed_sub_dags = Vec::new();
-        let mut leaders_to_commit = self.order_leaders(leader, state);
+        let mut leaders_to_commit = self.order_leaders(leader, state)?;
 
         while let Some(leader) = leaders_to_commit.pop_front() {
             let sub_dag_index = leader.nonce();
@@ -235,7 +244,7 @@ impl Bullshark {
             debug!(min_round, "Subdag has {} certificates", num_certificates);
 
             // We resolve the reputation score that should be stored alongside with this sub dag.
-            let reputation_score = self.resolve_reputation_score(state, &sequence, sub_dag_index);
+            let reputation_score = self.resolve_reputation_score(state, &sequence, sub_dag_index)?;
 
             let sub_dag: CommittedSubDag = CommittedSubDag::new(
                 sequence,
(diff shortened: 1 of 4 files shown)


grantkee (approved):
Thank you - great to have this for production - good work!

sstanfield (approved):

PR #1051, opened 2026-07-26
tn-reth: meter blsVerify by message length and cap it

### LINKED ISSUE #1044: `tn-reth`: BLS verify precompile (`0xb151`) charges flat gas for a variable-length `message`
`tn-reth`: BLS verify precompile (`0xb151`) charges flat gas for a variable-length `message`

**Class:** gas-metering hardening / forward-looking robustness in the native BLS precompile.  Confirmed on `origin/main` (`746d59cb`).  **This is not a live bug and needs no action today.**  It is an already-documented design decision (the in-code comment prescribes the exact remediation below);  the purpose of this issue is to *track* that prescribed follow-up so it is not lost when the precompile gains a caller that supplies a variable-length message.  Severity: **Informational**.

## Summary

`handle_bls_verify` charges a flat `BLS_VERIFY_GAS_COST = 150_000` for every `blsVerify(bytes,bytes,bytes)` call, independent of `message.len()` (`crates/tn-reth/src/evm/bls_precompile/mod.rs:77`, `:114-130`).  The underlying verification hashes the message to the curve, and hash-to-curve cost (the `expand_message_xmd` SHA-256 expansion inside `bls_verify_secure`) grows with message length, so the charge is flat while the work it meters is not.

This is acceptable and correct for the **only** on-chain caller today, and the code says so explicitly.  The doc comment on the constant already states the calibration and prescribes the fix (`mod.rs:63-76`):

> This is a flat charge calibrated for the registry's fixed-size proof-of-possession message (119 bytes).  As a generic primitive, `blsVerify` accepts an arbitrary `message` whose hash-to-curve cost scales with its length;  in practice that length is bounded by the calldata gas the caller already pays to supply it.  Before any non-proof-of-possession caller relies on `blsVerify` with large messages, add a per-word component to this charge (mirroring EIP-2537 hash-to-curve pricing) or cap `message.len()`.

This issue exists so that "before any non-proof-of-possession caller relies on it" has a tracked home, because the precompile is a *generic* primitive (its own module docs, `mod.rs:10-13`: "any contract can verify any BLS-signed message"), so a future caller can appear far from this file.

## Current behavior

The charge is flat and gated before decode (`mod.rs:114-130`):

```rust
    if gas_limit < BLS_VERIFY_GAS_COST {
        return Err(PrecompileError::OutOfGas);
    }

    Ok(PrecompileOutput::new(BLS_VERIFY_GAS_COST, Bytes::from(verified.abi_encode())))
}
```

There is no per-word term and no upper bound on `decoded.message.len()`.  A caller that supplies an N-byte message pays the same `150_000` for any N.

## Why it is not exploitable today

Two facts keep this Informational rather than a live denial-of-service:

1. **The sole caller supplies a fixed-size message.**  On `main` the `tn-contracts` submodule is pinned at `ac22e265` (the "generic BLS verify precompile + validator-set migration" commit, #126).  There, the only on-chain staticcall to the precompile address is `ConsensusRegistry._verifyProofOfPossession` (`tn-contracts/src/consensus/ConsensusRegistry.sol:731-732`):

   ```solidity
   (bool ok, bytes memory result) = BLS_G1_ADDRESS.staticcall(
       abi.encodeWithSelector(IBlsG1.blsVerify.selector, g1Pop.signature, blsPubkey, popMessage)
   );
   ```

   `popMessage` is built by `proofOfPossessionMessage` (`ConsensusRegistry.sol:679-692`) as `abi.encodePacked(POP_INTENT_PREFIX, pubkey, validatorAddress)`, and the pubkey length is gated to exactly 96 bytes in **two** places (`ConsensusRegistry.sol:690` and the caller `:707`, both reverting `InvalidBLSPubkey`).  The message is therefore a fixed size (a constant intent prefix plus the 96-byte compressed pubkey plus the 20-byte address, 119 bytes total) and is not attacker-inflatable through the registry.

2. **Direct calls are bounded by calldata gas.**  The precompile is generic and callable directly by any contract or EOA, so a large `message` *can* be sent to `0xb151` today outside the registry.  But supplying N bytes of calldata already costs the caller on the order of `16` gas per non-zero byte, which grows faster than the marginal, natively fast (`blst`, microseconds) hash-to-curve compute the flat charge under-prices.  The mispricing is a metering-hygiene gap, not a cheap-DoS primitive.

## Suggested fix

Either of the two options the in-code comment already names.  Both preserve today's behavior for the fixed 119-byte proof-of-possession caller.

**Option A (minimal, if `blsVerify` is intended to stay proof-of-possession-only): cap `message.len()`.**  A one-line bound forecloses the unbounded case while leaving the flat charge intact for the current caller:

```rust
/// Upper bound on the message the precompile will hash. The sole on-chain caller
/// (ConsensusRegistry proof-of-possession) supplies a fixed 119-byte message; this
/// cap keeps the flat `BLS_VERIFY_GAS_COST` honest by foreclosing an arbitrarily long
/// hash-to-curve charged at the flat rate.
const MAX_BLS_VERIFY_MESSAGE_LEN: usize = 128;

    if gas_limit < BLS_VERIFY_GAS_COST {
        return Err(PrecompileError::OutOfGas);
    }

    if decoded.message.len() > MAX_BLS_VERIFY_MESSAGE_LEN {
        return Err(PrecompileError::Other("blsVerify: message exceeds maximum length".into()));
    }

    Ok(PrecompileOutput::new(BLS_VERIFY_GAS_COST, Bytes::from(verified.abi_encode())))
}
```

**Option B (general, if variable-length messages are a legitimate future use): add a per-word component.**  Price the hashed message per 32-byte word, mirroring EIP-2537 hash-to-curve pricing, using checked arithmetic and `try_from` in place of a naked `as` cast:

```rust
/// Per-32-byte-word surcharge over the hashed `message`, mirroring EIP-2537
/// hash-to-curve pricing. Benchmark-derived; placeholder until calibrated.
const BLS_VERIFY_PER_WORD_GAS_COST: u64 = 100;

    // Flat base plus a per-word charge over the message, so a longer message pays
    // for the extra hash-to-curve work it induces.
    let cost = u64::try_from(decoded.message.len().div_ceil(32))
        .ok()
        .and_then(|words| words.checked_mul(BLS_VERIFY_PER_WORD_GAS_COST))
        .and_then(|word_gas| BLS_VERIFY_GAS_COST.checked_add(word_gas))
        .ok_or(PrecompileError::OutOfGas)?;

    if gas_limit < cost {
        return Err(PrecompileError::OutOfGas);
    }

}
```

Note that Option B reorders decode ahead of the gas gate so the message length is known before pricing;  the decode is cheap and the previous early gate was only a coarse floor.

## Suggested test design

- **Metering test (Option B):** call `dispatch` with two well-formed inputs that differ only in `message.len()` (one 119 bytes, one several words longer) and assert `gas_used` is strictly larger for the longer message.  This must be *seen to fail* on today's flat-charge code (both return `150_000`) and pass after the per-word change.
- **Bound test (Option A):** call with `message.len() == MAX_BLS_VERIFY_MESSAGE_LEN` (accepted) and `+ 1` (rejected), asserting the reject path returns an error rather than charging the flat cost and succeeding.
- **Regression guard for the current caller:** the existing fixed-size proof-of-possession round-trip must still succeed and still cost `BLS_VERIFY_GAS_COST` under either option (the 119-byte message is under any sane cap and adds a fixed, small per-word delta).
- **Confirm-by-mutation:** revert the fix and confirm the metering/bound test fails, so it pins the new pricing rather than passing vacuously.

## Acceptance criteria

- `blsVerify` gains a per-word gas component over `message` (Option B), with a test that pins the length-dependent charge, or
- `blsVerify` caps `message.len()` (Option A), with a test that pins the bound, or
- the maintainers ratify the flat charge as intended and the "generic primitive, proof-of-possession-only in practice" invariant is documented at the caller boundary (it is already largely captured in `mod.rs:63-76`), and this issue is closed as `wontfix` / tracked-in-code.

In all cases the fixed-size proof-of-possession verification through `ConsensusRegistry` continues to succeed at its current cost.

## Provenance

Forward-looking item surfaced during a review of the EVM + precompiles area (R1 #4), re-verified against `origin/main` `746d59cb` (2026-07-25): the flat charge, the absence of a per-word term or `message.len()` cap, and the single fixed-size proof-of-possession caller all hold, and no open PR or issue changes the pricing or adds a variable-length-message caller.

MavenRain:
# tn-reth: meter `blsVerify` by message length and cap it

## Summary

The native BLS verify precompile (`0x…b151`) charged a flat `BLS_VERIFY_GAS_COST = 150_000` for every `blsVerify(bytes,bytes,bytes)` call, independent of `message.len()`, while the work it meters grows with the message.  This adds a per-word surcharge over the message *and* a generous upper bound, so the charge tracks the hash-to-curve work while the precompile stays a general primitive.  Closes #1044.

## Problem

`handle_bls_verify` gated on the flat cost and returned it verbatim as `gas_used`.  But `bls_verify_secure` hashes the message to the curve, and the message-dependent part of that (the `expand_message_xmd` SHA-256 expansion) scales with `message.len()`.  The charge was flat while the metered work was not.

This is **not a live bug** and needed no urgent action:

- The **only** on-chain caller today is `ConsensusRegistry`'s proof-of-possession `staticcall`, whose message is a fixed 119 bytes (`intentPrefix(3) || compressedPubkey(96) || address(20)`) and is not attacker-inflatable through the registry.
- Direct callers can send a large `message` to `0x…b151`, but supplying N calldata bytes already costs ~16 gas/byte, which grows faster than the marginal (microsecond, native `blst`) hash-to-curve compute the flat charge under-prices.

It is, however, a forward-looking metering-hygiene gap: the module documents `blsVerify` as a **generic** primitive that "hard-codes nothing about proof-of-possession", so a future caller supplying a variable-length message would ride the flat rate.  The in-code doc already prescribed the remediation (add a per-word component *or* cap `message.len()`);  this PR does both.

## Fix

`crates/tn-reth/src/evm/bls_precompile/mod.rs`:

1. **Per-word surcharge.**  `BLS_VERIFY_PER_WORD_GAS_COST = 12` gas per 32-byte word of the message, mirroring the `SHA256` precompile's 12-gas/word rate for the same variable-length hashing.  A new `bls_verify_gas_cost(len) = BLS_VERIFY_GAS_COST + 12 * ceil(len / 32)` computes the total with `u64::try_from` + `checked_mul`/`checked_add` (no naked `as`, total on overflow rather than panicking).
2. **Generous cap.**  `MAX_BLS_VERIFY_MESSAGE_LEN = 4096` bytes — ~34× the 119-byte PoP message and far above any realistic signed message — forecloses an unbounded hash-to-curve request while leaving `blsVerify` a general primitive.  Over-cap input is rejected as `PrecompileError::Other`;  it also bounds the cost arithmetic so it cannot overflow for any accepted message.
3. **Decode before the gas gate.**  The message length (which now determines the charge) must be known before pricing.  The decode is cheap and the previous early gate was only a coarse floor.  The gate is now expressed as a `Result` combinator chain (cap → cost → gas), keeping the handler `return`-free per house style.

This implements *both* remediation options the issue offered (Option A cap + Option B per-word), for defense in depth.

## Why it is safe

- **Monotonic pricing.**  The per-word term can only *raise* the cost of a longer message;  nothing gets cheaper.  Distinct error classes are preserved: over-cap → `Other`, under-funded → `OutOfGas`.
- **Deterministic across validators.**  `div_ceil`, `try_from`, and the checked ops are platform-independent on the 64-bit targets;  the cap bounds the arithmetic so no accepted message can overflow.
- **Existing caller unaffected in practice.**  The fixed 119-byte PoP message rounds to 4 words, so the registry path now pays `150_000 + 48 = 150_048` — a fixed, small delta, far inside its 1M gas budget.  The `STATICCALL` relay integration test (the exact `ConsensusRegistry` path) still verifies.

## Testing

- **Unit** (`cargo test -p tn-reth --lib bls_precompile`): added `bls_verify_gas_cost_charges_per_word` (formula, word rounding), `dispatch_meters_by_message_length` (two calls differing only in message length;  strict cost inequality — this *fails* on the old flat charge), and `dispatch_rejects_oversized_message` (accept at cap, reject at cap+1 as `Other` even with ample gas).  Updated the fixed-cost regressions to pin `base + 4·per-word` for the 119-byte message.  14/14 pass.
- **Integration** (`bls_precompile_props`, incl. the `STATICCALL` relay): 11/11 pass.
- **Confirmed by mutation.**  Flattening `bls_verify_gas_cost` and neutering the cap makes exactly the four pricing/bound tests fail;  reverting restores green.  The pins are not vacuous.
- `cargo fmt --check` and `cargo clippy -p tn-reth` clean.

Closes #1044.

### DIFF
--- a/crates/tn-reth/src/evm/bls_precompile/mod.rs
+++ b/crates/tn-reth/src/evm/bls_precompile/mod.rs
@@ -60,21 +60,49 @@ sol! {
     ) external view returns (bool);
 }
 
-/// Gas charged for a BLS signature verification.
+/// Base gas charged for a BLS signature verification, independent of message length.
 ///
 /// Priced to reflect the equivalent EIP-2537 work the verification represents: a 2-pairing check
-/// (`37_700 + 2 * 32_600 = 102_900`) plus hash-to-curve and point decompression, rounded up. This
-/// keeps the on-chain cost proportional to the cryptography while remaining well within a normal
-/// transaction's gas budget (`stake` / `delegateStake` run with a 1M default limit). The native
-/// implementation completes in microseconds; the charge exists for metering, not compute time.
+/// (`37_700 + 2 * 32_600 = 102_900`) plus point decompression and the fixed portion of
+/// hash-to-curve, rounded up. This keeps the on-chain cost proportional to the cryptography while
+/// remaining well within a normal transaction's gas budget (`stake` / `delegateStake` run with a 1M
+/// default limit). The native implementation completes in microseconds; the charge exists for
+/// metering, not compute time.
 ///
-/// This is a flat charge calibrated for the registry's fixed-size proof-of-possession message (119
-/// bytes). As a generic primitive, `blsVerify` accepts an arbitrary `message` whose hash-to-curve
-/// cost scales with its length; in practice that length is bounded by the calldata gas the caller
-/// already pays to supply it. Before any non-proof-of-possession caller relies on `blsVerify` with
-/// large messages, add a per-word component to this charge (mirroring EIP-2537 hash-to-curve
-/// pricing) or cap `message.len()`.
+/// The length-dependent portion of hash-to-curve is charged separately by
+/// [`BLS_VERIFY_PER_WORD_GAS_COST`]; the total for a given message is [`bls_verify_gas_cost`].
 const BLS_VERIFY_GAS_COST: u64 = 150_000;
+
+/// Per-32-byte-word surcharge over the hashed `message`.
+///
+/// `blsVerify` is a generic primitive, so it accepts a message of any length up to
+/// [`MAX_BLS_VERIFY_MESSAGE_LEN`]. The message-dependent cost of [`bls_verify_secure`] is the
+/// `expand_message_xmd` SHA-256 expansion, which hashes the message once; this rate mirrors the
+/// `SHA256` precompile's 12-gas-per-word charge for the same variable-length hashing, so a longer
+/// message pays for the extra work it induces instead of riding the flat base charge.
+const BLS_VERIFY_PER_WORD_GAS_COST: u64 = 12;
+
+/// Upper bound on the `message` the precompile will hash.
+///
+/// A generous ceiling - far above any realistic signed message, and orders of magnitude above the
+/// sole on-chain caller's fixed 119-byte proof-of-possession message - that keeps `blsVerify` a
+/// general primitive while foreclosing an unbounded hash-to-curve request charged through the
+/// per-word rate. It also bounds [`bls_verify_gas_cost`]'s arithmetic so the charge cannot overflow
+/// for any accepted message.
+const MAX_BLS_VERIFY_MESSAGE_LEN: usize = 4096;
+
+/// Total gas for verifying a `message_len`-byte message: the flat [`BLS_VERIFY_GAS_COST`] base plus
+/// [`BLS_VERIFY_PER_WORD_GAS_COST`] for each 32-byte word of the message (rounded up).
+///
+/// Returns `None` on arithmetic overflow. That is unreachable for any `message_len` at or under
+/// [`MAX_BLS_VERIFY_MESSAGE_LEN`] (the caller gates on the cap first), but the checked arithmetic
+/// keeps the function total rather than panicking if a future caller ever skips the gate.
+fn bls_verify_gas_cost(message_len: usize) -> Option<u64> {
+    u64::try_from(message_len.div_ceil(32))
+        .ok()
+        .and_then(|words| words.checked_mul(BLS_VERIFY_PER_WORD_GAS_COST))
+        .and_then(|word_gas| BLS_VERIFY_GAS_COST.checked_add(word_gas))
+}
 
 /// Registers the BLS precompile at [`BLS_G1_PRECOMPILE_ADDRESS`] in the given map.
 ///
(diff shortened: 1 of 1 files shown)


sstanfield (approved):

Huwonk (approved):
@grantkee The fork implications of this change (difference in gas if this was called at any point in the past) may need to be addressed, but overall LGTM.

PR #1053, opened 2026-07-26
tn-reth: seed reth's txpool defaults so an explicit --txpool.max-account-slots 16 is honored

### LINKED ISSUE #1050: `tn-reth`: `--txpool.max-account-slots 16` is silently promoted to 256, so the reth-standard value is unreachable
`tn-reth`: `--txpool.max-account-slots 16` is silently promoted to 256, so the reth-standard value is unreachable

**Class:** config handling / silent override of explicit operator intent.  Confirmed on `origin/main` (`746d59cb`) against the pinned reth `v1.11.3` (`d6324d63`).  `RethConfig::new` decides "the operator did not set this flag" by comparing the parsed value against reth's default constant, but that constant is exactly what clap injects when the flag is absent, so an operator who explicitly types the reth-standard `16` is indistinguishable from one who typed nothing and is silently given `256`.  Severity is **Low**: this is operator-facing, not remotely triggerable, and it is **not** a memory-safety or unbounded-growth issue.  The global subpool caps (10,000 txs or 20 MiB per subpool, reth `crates/transaction-pool/src/config.rs:15`, `:18`) bound the pool regardless and are independent of `max_account_slots`.  This issue also does not contest the `256` default itself, which is a deliberate design decision;  it contests only that the flag documented to override it cannot express one of its values.

## Summary

`crates/tn-reth/src/lib.rs:334-338` raises reth's per-sender transaction-pool slot limit from 16 to 256 unless the operator set it themselves:

```rust
// TN default: 256 slots per sender (Reth default is 16).
// Only apply if user hasn't explicitly set a custom value via --txpool.max-account-slots.
if txpool.max_account_slots == TXPOOL_MAX_ACCOUNT_SLOTS_PER_SENDER {
    txpool.max_account_slots = 256;
}
```

The guard is sentinel equality against `TXPOOL_MAX_ACCOUNT_SLOTS_PER_SENDER` (= 16, reth `crates/transaction-pool/src/config.rs:12`), and reth's clap default for the flag resolves to that same constant:

```rust
#[arg(long = "txpool.max-account-slots", alias = "txpool.max_account_slots",
      default_value_t = DefaultTxPoolValues::get_global().max_account_slots)]
pub max_account_slots: usize,
```

(reth `crates/node/core/src/args/txpool.rs:330-331`, with `DefaultTxPoolValues::default()` setting `max_account_slots: TXPOOL_MAX_ACCOUNT_SLOTS_PER_SENDER` at `:265`.)

## The defect

| CLI input | value reaching `RethConfig::new` | guard at `lib.rs:336` fires | effective limit |
| --- | --- | --- | --- |
| flag omitted | 16 | yes | 256 (intended) |
| `--txpool.max-account-slots 16` | 16 | yes | **256 (bug: the operator asked for 16)** |
| `--txpool.max-account-slots 32` | 32 | no | 32 (intended) |

The alias form `--txpool.max_account_slots 16` is swallowed identically.  The sharp statement of the defect is not that 16 is mis-detected but that **16 is unreachable**: no CLI input yields an effective per-sender limit of 16, while 15 and 17 both work.  The operator receives no warning, no log line, and no way to observe the substitution.

## Why the stated intent cannot be satisfied as written

`TxPoolArgs::max_account_slots` is a bare `usize` with a clap default (reth `crates/node/core/src/args/txpool.rs:331`), not an `Option`, so the parsed struct carries no "was this set" bit. `RethConfig::new` receives only the already-parsed struct (`crates/tn-reth/src/lib.rs:332`, `let RethCommand { mut rpc, mut txpool, db } = reth_config;`) and `Cli::parse_args` discards the `ArgMatches` (`crates/telcoin-network-cli/src/cli.rs:77-79`, `Self::parse()`). The predicate the comment at `lib.rs:335` promises is therefore not recoverable at that point by any local change.

This is the only sentinel of its kind in the function.  Every other TN override of a reth default in `RethConfig::new` is an unconditional struct literal over a field that `RethCommand` does not expose to operators at all (`lib.rs:263-275` exposes only `rpc`, `txpool`, `db`), so there is no wider class of silently-swallowed flags to fix here.  The adjacent RPC-module filtering at least warns (`lib.rs:311`, `:314`);  the txpool path is silent.

## Why it is worth fixing

`max_account_slots` is not a cosmetic knob.  In reth it is an admission-time per-sender anti-spam gate, enforced at exactly one site: `AllTransactions::ensure_valid` (reth `crates/transaction-pool/src/pool/txpool.rs:1815-1836`), doc-headed *"Spam protection: reject new non-local transaction from a sender that exhausted its slot capacity"*, rejecting the incoming transaction with `InsertErr::ExceededSenderTransactionsCapacity` once a sender already holds that many slots pool-wide.  It exempts local-origin transactions and any transaction at the sender's next executable nonce, but public `eth_sendRawTransaction` submits with `TransactionOrigin::External` (reth `crates/rpc/rpc-eth-api/src/helpers/transaction.rs:85`), so the gate does bind on real ingress.

The delta an operator silently loses is 16x: roughly 40 senders can saturate the 10,000-transaction pending subpool at 256 slots, against roughly 625 at 16.

Two facts make the trap worse:

- The one place the setting is documented steers operators directly onto it.  `crates/telcoin-network-cli/README.md:386` reads:

  ```
  | `--txpool.max-account-slots` | `256` | Max pending transactions per sender (Reth default is 16; Telcoin Network overrides to 256) |
  ```

  An operator who reads that row and wants reth-standard tightness will type `16`, which is precisely the value that gets swallowed.
- The result is unobservable in production.  The txpool RPC namespace is rejected (`crates/tn-reth/src/lib.rs:313-315`) and reth's txpool metrics are not exported today (#738), so nothing surfaces the effective value.

## Suggested fix

**Option A (recommended): seed reth's global pool defaults before parsing, then delete the sentinel.** reth exposes a one-shot global for exactly this purpose (`crates/node/core/src/args/txpool.rs:21`, `:62-64`, `:132-135`), and `DefaultTxPoolValues` is re-exported through the `reth::args` path `tn-reth` already imports from (`crates/tn-reth/src/lib.rs:47-52`).

```rust
/// Seed reth's global transaction-pool defaults with Telcoin Network's per-sender slot limit.
///
/// Must run before any `TxPoolArgs` is parsed or default-constructed: reth resolves the
/// `--txpool.max-account-slots` clap default from this `OnceLock`, and `get_global` locks
/// it in on first read. Once seeded, an explicit `--txpool.max-account-slots 16` is honored
/// because 16 is no longer the default.
pub fn init_txpool_defaults() {
    if DefaultTxPoolValues::default().with_max_account_slots(256).try_init().is_err() {
        warn!(
            target: "tn::reth",
            "transaction pool defaults already initialized; TN per-sender slot default not applied"
        );
    }
}
```

This is a change set, not a one-liner, and the following three parts are all load-bearing:

1. Add the helper above.
2. Call it from every entry point that can reach `DefaultTxPoolValues::get_global()`. Two exist: `Cli::parse_args` before `Self::parse()` (`crates/telcoin-network-cli/src/cli.rs:77-79`), and the test harness before its own parse at `crates/test-utils/src/execution.rs:68` (`NodeCommand::<CliExt>::try_parse_from(cli_args)`). Note the hazard is wider than clap: `impl Default for TxPoolArgs` also reads the global (reth `crates/node/core/src/args/txpool.rs:434-467`), so any `NodeConfig::default()` in-process locks it first.
3. Delete the block at `crates/tn-reth/src/lib.rs:334-338`. Without this deletion the fix is inert, since the sentinel would keep rewriting an explicit 16.

`try_init` consumes `self` and returns `Result<(), Self>`, so the failure path is logged rather than unwrapped, per the repo's no-`unwrap`/`expect` convention.

Side benefit worth confirming during review: `RethEnv::new_for_temp_chain` constructs `RethConfig(node_config)` directly (`crates/tn-reth/src/lib.rs:1466`) and never runs the promotion block, so that path currently uses reth's 16 while the CLI path uses 256. Seeding the global makes the two agree.

**Option B (runner-up, more faithful, more plumbing): thread `ArgMatches::value_source`.** Replace `Self::parse()` with `Cli::command().get_matches()` plus `from_arg_matches`, carry `ValueSource::CommandLine` for this argument through `NodeCommand` into `RethConfig::new`, and repeat at the test-utils parse.  This expresses "the user explicitly set it" exactly, at the cost of multi-crate plumbing and a second parse path to keep in sync.  Option A gets the same operator-visible behavior with a far smaller diff.

Either way, update the README row at `crates/telcoin-network-cli/README.md:386` so the documented default matches the mechanism.

## Suggested test design

In `crates/telcoin-network-cli` (or wherever the parse is exercised), pin the resolved value for all three rows of the table above:

- flag omitted resolves to 256,
- `--txpool.max-account-slots 16` resolves to **16**, which fails on current `main`,
- `--txpool.max-account-slots 32` resolves to 32,
- the alias form `--txpool.max_account_slots 16` resolves to 16.

Because the seeding is process-global via `OnceLock`, the assertions must live in a test that controls initialization order, or the resolved value must be read back through a helper that seeds first.  A test that merely calls `TxPoolArgs::default()` after some other test locked the global would pass vacuously.

**Confirm-by-mutation:** restore the deleted sentinel block (or remove the seeding call) and confirm the explicit-16 test fails, so the test pins the new behavior rather than passing regardless.

## Acceptance criteria

- `--txpool.max-account-slots 16` yields an effective per-sender limit of 16, the default with no flag stays 256, and a test pins both and has been confirmed by mutation.
- Or the maintainers ratify 256 as a hard protocol-level floor rather than a default, in which case the flag should reject or explicitly warn on values below the floor instead of silently rewriting one specific value, and `lib.rs:334-335` plus `README.md:386` should be corrected to say so.
- Or this is closed as wontfix with a one-line note that operator-supplied 16 is intentionally indistinguishable from unset, so the next reader of `lib.rs:335` is not misled by a comment the code cannot honor.

In all cases `make attest` stays green on the pinned 1.94 toolchain.

## Provenance

Found while re-verifying a prior review note about the 16 to 256 override against `origin/main` `746d59cb` on 2026-07-26.  That note's own conclusion held (the override is bounded and not a memory-safety defect);  this config-handling defect is a separate, previously unreported consequence of how the override is applied.

Not a duplicate: no open PR touches `crates/tn-reth/src/lib.rs:334-338` or `crates/tn-reth/src/txn_pool.rs`, and no issue in any state mentions `max_account_slots`, `SubPoolLimit`, or `pool_config`. Related to #738, which is why the effective value is currently unobservable.

MavenRain:
Closes #1050.

## Problem

`--txpool.max-account-slots 16` was silently promoted to 256, so reth's own per-sender slot value was unreachable from the CLI.  No input produced an effective limit of 16, while 15 and 17 both worked, and the operator got no warning and no way to observe the substitution.

`max_account_slots` is not cosmetic.  It is reth's admission-time per-sender spam gate, enforced in `AllTransactions::ensure_valid`, which rejects an incoming external transaction once its sender already holds that many slots pool-wide.  The delta an operator silently lost is 16x: roughly 40 senders saturate the 10,000 transaction pending subpool at 256 slots, against roughly 625 at 16.  This is operator-facing only.  It is not remotely triggerable and it is not an unbounded-growth issue, since the global subpool caps bound the pool regardless.

## Root cause

`RethConfig::new` decided "the operator did not set this flag" by comparing the parsed value against reth's default constant:

```rust
if txpool.max_account_slots == TXPOOL_MAX_ACCOUNT_SLOTS_PER_SENDER {
    txpool.max_account_slots = 256;
}
```

That constant is exactly what clap injects when the flag is absent (`default_value_t = DefaultTxPoolValues::get_global().max_account_slots`, reth `crates/node/core/src/args/txpool.rs:330`).  `TxPoolArgs::max_account_slots` is a bare `usize`, not an `Option`, so the parsed struct carries no "was this set" bit, and `RethConfig::new` receives only the already-parsed struct.  The predicate the comment promised was not recoverable at that point.

## Fix

Move the override from after the parse to before it, so the default reth injects is already TN's and an explicit value never has to be distinguished from an absent one.

- [x] `crates/tn-reth/src/lib.rs`: add `TN_TXPOOL_MAX_ACCOUNT_SLOTS_PER_SENDER` (256) and `init_txpool_defaults`, which seeds reth's process-wide `OnceLock` via `DefaultTxPoolValues::try_init`.  `try_init` consumes `self` and returns `Result<(), Self>`, so the late-init path is logged rather than unwrapped, per the no-`unwrap`/`expect` convention.  The warning fires only on an actual mismatch, so repeat calls are silent.
- [x] `crates/tn-reth/src/lib.rs`: delete the sentinel block in `RethConfig::new`.  Without this the seeding is inert, because the sentinel would then rewrite an explicit 16 and nothing else.
- [x] `crates/tn-reth/src/lib.rs`: seed in `RethEnv::new_for_temp_chain`.  Its `..NodeConfig::default()` reaches the same `OnceLock` through `TxPoolArgs::default`, so a temp chain built first (the genesis ceremony does exactly this) would otherwise lock the default at 16 for the rest of the process and make every later seeding call a no-op.  This is the one part of the change that is load-bearing but not obvious from the flag's name.
- [x] `crates/telcoin-network-cli/src/cli.rs`: seed in `parse_args` and `try_parse_args_from`, and move both onto the generic `impl<Ext> Cli<Ext>`.  On `impl Cli` they covered only `Cli<NoArgs>`, which would have left the crate's own documented ExEx extension path (`Cli::<MyArgs>::parse()`) resolving to 16 once the sentinel was gone.  The doc example now uses `parse_args`.  In-tree callers gain a `::<NoArgs>` turbofish.
- [x] `crates/test-utils/src/execution.rs`, `crates/e2e-tests/src/lib.rs`: seed before each harness parses a `NodeCommand`, so tests configure nodes the way the binary does.  `tn-reth` moves from dev-dependencies to dependencies in `e2e-tests` because `src/lib.rs` now needs it.
- [x] `crates/telcoin-network-cli/README.md`: the documented row now says an explicit value is honored, including 16.

Side effect worth naming in review: `RethEnv::new_for_temp_chain` previously ran with reth's 16 while the CLI path ran with 256, because it never reached the sentinel.  Seeding makes the two agree at 256.

Not addressed here, and deliberately so: this does not contest the 256 default itself, which is a design decision.  It contests only that the flag documented to override it could not express one of its values.

## Testing

Worktree `1050-txpool-seed-reth-pool-defaults` off `origin/main` 746d59cb.

- `cargo +nightly-2026-03-20 clippy -p telcoin-network -p telcoin-network-cli -p tn-reth -p tn-test-utils -p e2e-tests -- -D warnings`, default features and `--all-features` . . . 0 errors, 0 warnings.  This matches attest's clippy scope (no `--all-targets`).
- `cargo test -p telcoin-network-cli` . . . 53 passed, 0 failed.  `cargo test -p tn-reth --lib` . . . 65 passed, 0 failed.  `cargo test -p telcoin-network-cli --doc` . . . 1 passed, which compile-checks the edited `Cli::run` example.

New tests:

- `crates/telcoin-network-cli/tests/txpool_defaults.rs` pins all four rows of the table in the issue: absent flag resolves to 256, explicit `16` resolves to 16, the `--txpool.max_account_slots` alias form resolves to 16, and `32` resolves to 32.  It gets its own integration test binary on purpose.  The seeding is a process-wide `OnceLock`, and a sibling in the library test binary (`Cli::command()` in `test_parse_help_all_subcommands`) would take the lock first and make the default-value assertion vacuous.  Every test in this binary seeds the same value, so any interleaving yields the same lock contents under both `cargo test` and `nextest`.
- `crates/tn-reth/src/lib.rs` pins that `RethConfig::new` passes an explicit value through untouched.  This one asserts only the explicit case, which is independent of who reached the `OnceLock` first, so it is order-safe in a shared test binary.

Each was confirmed by mutation, seen to fail before being restored:

1. Restoring the sentinel block in `RethConfig::new` failed `explicit_max_account_slots_survives_reth_config_new` with `left: 256, right: 16`.
2. Removing `init_txpool_defaults()` from `Cli::try_parse_args_from` failed `absent_flag_resolves_to_tn_default` with `left: Some(16), right: Some(256)`.
3. Removing `#![allow(unused_crate_dependencies)]` from the new test root raised 29 `unused_crate_dependencies` errors under `-D warnings`, confirming the attribute (which every other test root in the repo carries) is required rather than decorative.

`make attest` was not run end to end here, since it requires a clean tree and submits an on-chain transaction.  Its local gates are green as listed above.

### DIFF
--- a/bin/telcoin-network/src/main.rs
+++ b/bin/telcoin-network/src/main.rs
@@ -1,13 +1,15 @@
 //! Main binary for TN CLI
 
-use telcoin_network_cli::{cli::Cli, passphrase::get_bls_passphrase_from_env};
+use telcoin_network_cli::{cli::Cli, passphrase::get_bls_passphrase_from_env, NoArgs};
 use tn_node::launch_node;
 
 fn main() {
     // Must be the first statement of main: reads and clears TN_BLS_PASSPHRASE
     // before any threads exist (see `get_bls_passphrase_from_env`).
     let preloaded = get_bls_passphrase_from_env();
-    let cli = Cli::parse_args(); // = Cli::<NoArgs>::parse(), cli.rs:77-79
+    // Seeds reth's transaction pool defaults, then parses (see `Cli::parse_args`). The seeding
+    // has to precede the parse: clap resolves `--txpool.max-account-slots` against those defaults.
+    let cli = Cli::<NoArgs>::parse_args();
 
     let passphrase = cli.resolve_bls_passphrase(preloaded).unwrap_or_else(|err| {
         eprintln!("{err}");
--- a/crates/e2e-tests/Cargo.toml
+++ b/crates/e2e-tests/Cargo.toml
@@ -16,6 +16,8 @@ clap = { workspace = true, features = ["derive", "env"] }
 eyre = { workspace = true }
 tn-config = { workspace = true }
 tn-node = { workspace = true }
+# not dev-only: `src/lib.rs` seeds reth's transaction pool defaults before it parses a node command
+tn-reth = { workspace = true, features = ["test-utils"] }
 tn-types = { workspace = true, features = ["test-utils"] }
 tracing = { workspace = true }
 jsonrpsee = { workspace = true }
@@ -31,7 +33,6 @@ alloy = { workspace = true }
 
 [dev-dependencies]
 futures = { workspace = true }
-tn-reth = { workspace = true, features = ["test-utils"] }
 tn-test-utils = { workspace = true }
 tempfile = { workspace = true }
 serde_json = { workspace = true }
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -14,6 +14,7 @@ use std::{
 use telcoin_network_cli::{genesis::GenesisArgs, keytool::KeyArgs, node::NodeCommand, NoArgs};
 use tn_config::{Config, ConfigFmt, ConfigTrait, KeyConfig};
 use tn_node::launch_node;
+use tn_reth::init_txpool_defaults;
 use tn_types::{test_utils::CommandParser, Address, Genesis, GenesisAccount};
 use tracing::{error, info};
 // unused deps warnings
@@ -294,6 +295,8 @@ pub fn spawn_local_testnet(
             ipc_path: ipc_path_str.clone(),
         });
 
+        // seed reth's pool defaults before the parse resolves `--txpool.max-account-slots`
+        init_txpool_defaults();
         let command = NodeCommand::parse_from([
             "tn",
             "--http",
--- a/crates/telcoin-network-cli/README.md
+++ b/crates/telcoin-network-cli/README.md
@@ -383,7 +383,7 @@ The `admin` and `txpool` modules are not available at this time.
 
 | Flag                         | Default | Description                                                                                |
 | ---------------------------- | ------- | ------------------------------------------------------------------------------------------ |
-| `--txpool.max-account-slots` | `256`   | Max pending transactions per sender (Reth default is 16; Telcoin Network overrides to 256) |
+| `--txpool.max-account-slots` | `256`   | Max pending transactions per sender. Telcoin Network raises Reth's default of 16 to 256; any explicit value is honored, including 16 |
 
 ## Networking
 
--- a/crates/telcoin-network-cli/src/cli.rs
+++ b/crates/telcoin-network-cli/src/cli.rs
@@ -9,7 +9,7 @@ use clap::{Parser, Subcommand};
 use std::{ffi::OsString, fmt, path::PathBuf, str::FromStr};
 use tn_config::KeyConfig;
 use tn_node::engine::TnBuilder;
-use tn_reth::{dirs::DEFAULT_ROOT_DIR, LogArgs};
+use tn_reth::{dirs::DEFAULT_ROOT_DIR, init_txpool_defaults, LogArgs};
 use tokio::{runtime::Builder, task::JoinHandle};
 use tracing::info_span;
 
@@ -72,23 +72,30 @@ pub struct Cli<Ext: clap::Args + fmt::Debug = NoArgs> {
     pub logs: LogArgs,
 }
 
-impl Cli {
+impl<Ext: clap::Args + fmt::Debug> Cli<Ext> {
     /// Parsers only the default CLI arguments
+    ///
+    /// Seeds reth's transaction pool defaults first: the clap default for
+    /// `--txpool.max-account-slots` is resolved while the command is built, so seeding after the
+    /// parse would be too late. Prefer this over [`clap::Parser::parse`] for that reason, and note
+    /// that it is generic over `Ext` so an extended binary gets the same defaults as `tn` itself.
     pub fn parse_args() -> Self {
+        init_txpool_defaults();
         Self::parse()
     }
 
     /// Parsers only the default CLI arguments from the given iterator
+    ///
+    /// Seeds reth's transaction pool defaults first, for the same reason as [`Self::parse_args`].
     pub fn try_parse_args_from<I, T>(itr: I) -> Result<Self, clap::error::Error>
     where
         I: IntoIterator<Item = T>,
         T: Into<OsString> + Clone,
     {
-        Cli::try_parse_from(itr)
+        init_txpool_defaults();
+        Self::try_parse_from(itr)
     }
-}
 
-impl<Ext: clap::Args + fmt::Debug> Cli<Ext> {
     /// Execute the configured cli command.
     ///
     /// This accepts a closure that is used to launch the node via the
@@ -100,7 +107,6 @@ impl<Ext: clap::Args + fmt::Debug> Cli<Ext> {
     /// Parse additional CLI arguments for the node command and use it to configure the node.
     ///
     /// ```no_run
-    /// use clap::Parser;
     /// use telcoin_network_cli::{cli::Cli, passphrase::get_bls_passphrase_from_env};
     /// use tn_node::launch_node;
     ///
@@ -114,7 +120,9 @@ impl<Ext: clap::Args + fmt::Debug> Cli<Ext> {
     ///
     /// // 1. Read (and clear) TN_BLS_PASSPHRASE before any threads exist.
     /// let preloaded = get_bls_passphrase_from_env();
-    /// let cli = Cli::<MyArgs>::parse();
+    /// // `parse_args` rather than `clap::Parser::parse`: it seeds reth's transaction pool
+    /// // defaults, which must happen before the clap command resolves its own defaults.
+    /// let cli = Cli::<MyArgs>::parse_args();
     /// // 2. Resolve the passphrase for the parsed subcommand.
     /// let passphrase = cli.resolve_bls_passphrase(preloaded)?;
     /// // 3. Run, installing ExExes on the builder before launching the node.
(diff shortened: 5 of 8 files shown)


grantkee (approved):
Good improvement

sstanfield (approved):

Issue #1031, opened 2026-07-24
metric: Add metric for `concludeEpoch` call at epoch boundary


grantkee:
# Problem

System call at the epoch boundary concludes the epoch. This call changes a lot of state. The execution is currently set at 30M gas but this may not be enough as the network grows.

## Solution

- metric to capture gas usage for every epoch closing call (gauge)
- prometheus dashboard query `max_over_time(my_event_value_gauge[30d])`
- benchmarks in `tn-contracts` for scenarios, num validators, etc. that max out 30M gas

More information is needed, but the protocol can safely set this value to a theoretical maximum gas for executing the `concludeEpoch` system call based on:
- max number of validators that can stake
- max number of state changes possible (reward balances, slashes, stake version upgrades, etc.)

related to #1012

Huwonk:
Measured this against the unified call on tn-contracts#140 with a new benchmark (ConcludeEpochGasBench in that PR). One sharpening of the problem statement first: the 30M in transact_system_call is per call, and the boundary previously made two system calls (applyIncentives, then concludeEpoch) - the unified call gets a single 30M envelope for rewards + slashes + version-queue settlement + rotation, so headroom effectively halved at the same time the call gained a stage.

Worst-case measurements (full committee rotation + rewards for every validator + partial slashes + a full settlement wave of queued stake decreases):

| Scenario | Gas | % of 30M |
|---|---|---|
| N=50 | 3.85M | 13% |
| N=100 | 7.66M | 26% |
| N=150 | 11.5M | 38% |

Roughly linear at ~76k per validator all-in; extrapolated ceiling around N~390 in-service validators, all hitting the same boundary. These numbers are after tn-contracts#140 converted boundary refunds to pull-based credits - before that, each settling refund also pushed to its recipient with a 100k gas stipend, which let hostile recipients add ~100k x N of attacker-controlled gas and put the adversarial cliff near N~200. With the pull conversion the boundary makes no recipient-facing external calls, so the cost is storage-bounded and deterministic.

On the remaining items here:
- The gauge on res.result.gas_used() in apply_closing_epoch_contract_call is a two-line addition; natural home is the consolidated #1012.
[...]

PR #1056, opened 2026-07-26
tn-reth: count transactions dropped during block building

### LINKED ISSUE #959: count transactions dropped during block building in `build_block_from_batch_payload`
## Background

#938 changed the signer-recovery phase of `build_block_from_batch_payload` from a fatal `map(...).collect::<Result>()?` to a non-fatal `par_iter().filter_map(...).ok()`.  A transaction whose signer cannot be recovered is now dropped and execution continues, instead of deterministically halting (and crash-looping on restart-replay) the whole network on a single undecodable transaction inside a certified sub-DAG.  This is a defense-in-depth bound on the blast radius;  the primary fix (validating batches before they can be certified) means a recovery failure should never actually happen in production.

Because it should never happen, an operator has no cheap signal that it did.  Today the only evidence is a per-transaction `error!` log:

```rust
// crates/tn-reth/src/lib.rs, build_block_from_batch_payload (post-#938)
let recovered_txs = transactions
    .par_iter()
    .filter_map(|tx_bytes| {
        reth_recover_raw_transaction::<TransactionSigned>(tx_bytes)
            .inspect_err(|e| {
                error!(
                    target: "engine",
                    batch=?batch_digest,
                    ?tx_bytes,
                    "failed to recover signer, dropping transaction: {e}"
                )
            })
            .ok()
    })
    .collect::<Vec<_>>();
```

Confirming whether the certified-but-invalid path ever fires means grepping logs across nodes.  A counter makes it alertable: any nonzero value on this metric indicates a certified batch carried undecodable bytes, i.e. that batch validation was bypassed somewhere, which is exactly the condition #933 / #938 exists to catch.

## Two drop sites (scope note)

`build_block_from_batch_payload` has two silent drop/skip points, with opposite expectedness:

1. **Recovery-phase drop** (the one @grantkee flagged, `crates/tn-reth/src/lib.rs` ~line 776-790, `filter_map(...).ok()`, `error!` log).  Should never happen;  a nonzero count is an alert.
2. **Execute-phase skip** (`crates/tn-reth/src/lib.rs` ~line 793-810, `continue` on `BlockExecutionError::Validation(BlockValidationError::InvalidTx { .. })`, `warn!` log).  This is *expected* in normal operation (another worker's batch legitimately included the same transaction), so a counter here is informational at best and noisy.

The required deliverable is a counter for **(1)**.  A separate counter for **(2)** is optional and, if added, must be named/documented so it is not mistaken for an error signal.

## Proposed change

`crates/tn-reth` currently has **no** metrics infrastructure of its own (no `metrics` / `reth-metrics` dependency, no metrics module), and `build_block_from_batch_payload`'s `self` (`&RethEnv`, inner `Arc<RethEnvInner>`) holds no metrics handle.  The existing worker counter `batch_validation_failures_total` lives in a different crate (`crates/consensus/worker/src/metrics.rs`) and is not reachable here.

The clean home is a new `crates/tn-reth/src/metrics.rs` that mirrors the existing execution-layer precedent, `EngineMetrics` in `crates/engine/src/metrics.rs`:

```rust
// crates/tn-reth/src/metrics.rs  (new)
use std::sync::LazyLock;
use reth_metrics::Metrics;
use metrics::Counter;

/// Global metrics for the reth execution environment.
///
/// First use must happen after the global recorder is installed. `tn-metrics`
/// installs the recorder before `RethEnv::new_database`, so any use inside a
/// `RethEnv` method binds to the real recorder.
pub(crate) static RETH_METRICS: LazyLock<RethEnvMetrics> = LazyLock::new(RethEnvMetrics::default);

/// Prometheus metrics for `RethEnv` block building.
#[derive(Metrics)]
#[metrics(scope = "tn_reth")]
pub(crate) struct RethEnvMetrics {
    /// Transactions dropped during block building because their signer could
    /// not be recovered. Should always be zero; a nonzero value means a
    /// certified batch carried undecodable transaction bytes (batch
    /// validation was bypassed, see #933 / #938).
    pub(crate) unrecoverable_txs_dropped_total: Counter,
}
```

Increment **once** by the count difference after the `collect`, rather than per-item inside the rayon closure (avoids per-transaction atomic contention and mirrors the count-guarded idiom of `WorkerMetrics::record_batches_fetched`):

```rust
// after `.collect::<Vec<_>>();`
let dropped = transactions.len() - recovered_txs.len();
if dropped > 0 {
    RETH_METRICS.unrecoverable_txs_dropped_total.increment(dropped as u64);
}
```

Resulting series name: `tn_reth_unrecoverable_txs_dropped_total` (matches the `noun_verbed_total` style of `tn_engine_empty_outputs_skipped_total` and `tn_worker_batch_validation_failures_total`).

## Implementation notes / gotchas

- **Do not** put the counter in `EngineMetrics`: `crates/engine` depends on `crates/tn-reth` (`crates/engine/Cargo.toml`), so `tn-reth` cannot reach `ENGINE_METRICS` without a dependency cycle, and callers cannot distinguish recovery drops from execute-phase skips by diffing block tx counts.  The counter must live in `tn-reth`.
- **Deps to add** to `crates/tn-reth/Cargo.toml`, mirroring `crates/engine/Cargo.toml`: `metrics = { workspace = true }` and `reth-metrics = { workspace = true }`, plus `metrics-util = { workspace = true }` as a dev-dependency for the registration test.
- **Version pin:** use the workspace `metrics` dep only.  The root `Cargo.toml` pins `metrics = "0.24"` with an explicit warning that a version mismatch with reth's metrics "silently splits the global recorder."
- **Recorder ordering** is already satisfied: `crates/tn-metrics/src/recorder.rs` installs the recorder before `RethEnv::new_database`, so a `LazyLock` static first used inside a `RethEnv` method binds to the real recorder.  Prefer the static over threading a handle field through `RethEnvInner`, which would touch every construction site and caller signature (`crates/engine/src/payload_builder.rs`, `crates/tn-reth/src/lib.rs`, `crates/node/src/manager/node/run_epoch.rs`, plus tests).
- **Test:** port the `DebuggingRecorder` registration test from `crates/engine/src/metrics.rs` (construct under `metrics::with_local_recorder` so the handle binds to the local recorder, assert the counter registers/increments).
- **Doc comments** on the new module, struct, and field are required per the repo's documentation bar for all items.

## Acceptance criteria

- [ ] A counter increments by the number of transactions dropped in the recovery phase of `build_block_from_batch_payload`, so a nonzero value is visible in Prometheus without reading logs.
- [ ] The counter lives in `crates/tn-reth` (no dependency cycle introduced) and follows the repo's `#[derive(Metrics)]` + `LazyLock` static pattern with `scope = "tn_reth"`.
- [ ] The existing per-transaction `error!` log is retained.
- [ ] A registration/increment test covers the new metric.
- [ ] All new items carry doc comments.
- [ ] (Optional) A separately named counter for the expected execute-phase `InvalidTx` skip, documented so it is not read as an error signal.

MavenRain:
Closes #959.

## Summary

`build_block_from_batch_payload` drops a transaction whose signer cannot be recovered instead of halting on it.  That is deliberate (#933 / #938): a certified sub-DAG is fixed and identical on every honest node, so returning an error there would deterministically halt the network, and crash-loop it on restart-replay, over a single undecodable transaction.  Dropping keeps the block deterministic because every node drops the same bytes.

Because it is a path that should never fire, an operator had no cheap signal that it did.  The only evidence was a per-transaction `error!` log, so confirming whether the certified-but-invalid path ever ran meant grepping logs across nodes.

This adds the counter that makes it alertable.  Any nonzero value on `tn_reth_unrecoverable_txs_dropped_total` means a certified batch carried undecodable transaction bytes, i.e. batch validation was bypassed somewhere upstream, which is exactly the condition #933 / #938 exists to bound.

## Changes

- [x] `crates/tn-reth/src/metrics.rs` (new): `RethEnvMetrics` behind a `RETH_METRICS` `LazyLock`, `#[derive(Metrics)]` with `#[metrics(scope = "tn_reth")]`, mirroring `EngineMetrics` in `crates/engine/src/metrics.rs`.  Two counters, documented per field, plus an `init()` that forces the lock so both series register at zero.
- [x] `crates/tn-reth/src/lib.rs`: `RethEnv::new` calls `metrics::init()`, so the counters exist from process start rather than from the first drop.  Increment `unrecoverable_txs_dropped_total` once by the count difference after the rayon `collect`, not per item inside the closure.  Increment `invalid_txs_skipped_total` in the existing execute-phase `InvalidTx` arm.  The per-transaction `error!` log is retained unchanged.
- [x] `crates/tn-reth/Cargo.toml`: `metrics` and `reth-metrics` from the workspace, `metrics-util` as a dev-dependency for the registration test.  No new package and no version move enters `Cargo.lock` (all three were already in the graph via `crates/engine`), so nothing re-resolves and no feature unification changes for other crates.
- [x] `etc/grafana/telcoin-node-metrics.json`: the new counter as a third target on the existing "Validation Failures & Protocol Violations /s" panel in the Security row, alongside `tn_worker_batch_validation_failures_total`.  Same signal class, so it reads next to its sibling rather than as a new panel.

Rendered series name is `tn_reth_unrecoverable_txs_dropped_total`, matching the `noun_verbed_total` style of `tn_engine_empty_outputs_skipped_total` and `tn_worker_batch_validation_failures_total`.  The `tn_` prefix means `TnPrefixRecorder` passes it through untouched rather than prepending `reth.`.

## Why the counter lives in `tn-reth`

`crates/engine` depends on `crates/tn-reth`, so `tn-reth` cannot reach `ENGINE_METRICS` without a dependency cycle.  Diffing block transaction counts from the caller side is not a substitute either, since that cannot separate recovery-phase drops from execute-phase skips.

The static is preferred over threading a handle through `RethEnvInner`, which would touch every construction site and caller signature (`crates/engine/src/payload_builder.rs`, `crates/node/src/manager/node/run_epoch.rs`, plus tests) for two counters.  The recorder-ordering requirement that a `LazyLock` implies is already an invariant of the codebase: `tn_metrics::install_recorder` runs before `RethEnv::new_database`, so first use inside a `RethEnv` method binds to the real recorder, not the noop one.

## Counter semantics and threat model

The two drop sites in this function have opposite expectedness, so they are counted separately rather than summed:

1. **Recovery-phase drop** (`filter_map(...).ok()`, `error!` log).  Should never happen.  Alert on any nonzero value.  This is the attacker-visible one: getting undecodable bytes into a certified batch is precisely the bypass being watched for, and the counter is the detector, so it firing is the intended outcome rather than a weakness.
2. **Execute-phase skip** (`continue` on `BlockValidationError::InvalidTx`, `warn!` log).  Expected in normal operation, because workers batch independently and the same transaction can legitimately reach execution twice inside one consensus output.  Counted as `invalid_txs_skipped_total` and documented in the field doc as not an error signal and not alertable, so a steady nonzero rate is not read as a bypass.  This is the optional deliverable in the issue;  it is a self-contained six lines if reviewers would rather drop it.

The recovery-phase count is taken once per block from `transactions.len().saturating_sub(recovered_txs.len())` rather than incremented inside the rayon closure.  Per-item increments would put every transaction in a batch on one atomic in the parallel phase, which is a contention lever whose size an attacker controls through batch width.  The count is exact because `filter_map` is the only thing that removes elements between the two lengths, and it reports rather than gates, so nothing about admission or execution now depends on it.  No control flow changed: the same transactions are dropped, skipped, and executed as before this PR.

## Why the series is registered at startup

Worth calling out because the obvious version of this change is wrong.  A `LazyLock` registers on first deref, and the only derefs are on drop paths a healthy node never takes, so a counter written the natural way (`if dropped > 0 { increment(dropped) }`) is absent from `/metrics` entirely until the day something breaks.  That defeats the counter twice over: `absent()` cannot tell "no drops" apart from a mistyped metric name or a misrouted scrape job, so the negative case of an "alert on any nonzero" rule is unverifiable for exactly as long as it matters;  and `rate()` over a series whose first ever sample is already nonzero renders no step, so a one-shot burst on a quiet node never shows on the panel.

Two changes avoid it.  `RethEnv::new` forces the lock once, after `install_recorder` has run, and the recovery-phase increment is unconditional rather than guarded on a nonzero count.  Both counters therefore baseline at zero from process start, the same way `WorkerMetrics` and the executor metrics they sit beside on that dashboard panel already do.  The cost is one relaxed atomic add per block, which is free next to executing the block.

## Testing

Pinned toolchain (`rust-toolchain.toml` 1.94, nightly-2026-03-20 for fmt):

- `cargo test -p tn-reth --lib metrics::` . . . green, two tests.  The first is the registration test ported from `crates/engine/src/metrics.rs`: it builds the handles under `metrics::with_local_recorder` against a `DebuggingRecorder` so they bind to the local recorder, then asserts both series register under their exact rendered names and carry the values written.  The second pins the startup baseline, asserting both series register at zero with no drop having happened, which is the property `init()` exists for.
- Confirmed by mutation, not just by passing.  Renaming the scope to `tn_rethz` fails with `metric tn_reth.unrecoverable_txs_dropped_total not registered`;  changing an increment from 2 to 3 fails the counter assertion;  expecting `Counter(1)` instead of `Counter(0)` fails the baseline test.  All three reverted, green again.  The tests pin the rendered name and the value, which is what a dashboard or alert rule depends on.
- `cargo +1.94 test --no-run --workspace -j 2` . . . compiles.
- `etc/grafana/telcoin-node-metrics.json` re-parsed after the edit, panel 40 targets are `A`/`B`/`C` with no `refId` collision.

Scope note on test depth: this covers registration and increment on the metrics struct, matching how `tn-engine`, `tn-state-sync` and `tn-batch-builder` test their metrics.  It does not assert either counter from inside `build_block_from_batch_payload`, nor the `init()` call from inside `RethEnv::new`, since a process-global static plus a certified batch of deliberately undecodable bytes does not reduce to a deterministic unit test.  Happy to add an e2e pin if you want the call sites themselves covered.


grantkee (approved):
Good work. This is also a required precursor for #1031

sstanfield (approved):

Issue #912, opened 2026-07-14
Restrict consensus-output and epoch-vote gossip topics to committee publishers


MavenRain:
## Problem

Gossip propagation is gated before forwarding by `verify_gossip`
(`crates/network-libp2p/src/consensus.rs:1421`), whose result feeds
`report_message_validation_result` (the call that decides whether gossipsub relays the
message).  For a topic whose `authorized_publishers` entry is `Some(set)`, `verify_gossip`
rejects (and therefore does not propagate) any message whose authenticated `source` is not
in `set`.  For a topic whose entry is `None` (open), it accepts any signed source.

Today the primary's three gossip topics are gated inconsistently:

- `primary_topic` (certificates) is **restricted** to the current committee.  It is
  (re)subscribed every epoch via `subscribe_with_publishers(primary_topic, committee_keys)`
  (`crates/node/src/manager/node/start_epoch.rs:507-514`), where `committee_keys` is the
  current committee's protocol keys (`start_epoch.rs:471`).
- `epoch_vote_topic` and `consensus_output_topic` are **open**.  They are subscribed once at
  node startup with the plain `subscribe(...)` (publishers `None`):
  `crates/node/src/manager/node.rs:457` and `:461`, inside the process-lifetime `run()`
  path (`node.rs:401`), not the per-epoch routine.

So a non-committee node can publish an `EpochVote` or a `ConsensusResult` on those two
topics and have the mesh propagate it after only gossipsub's message-signature check.  Both
payload kinds are produced only by committee members in practice
(`PrimaryNetworkHandle::publish_consensus` is called from the active-CVV subscriber,
`crates/consensus/executor/src/subscriber.rs`;  `publish_epoch_vote` from the epoch-vote
collector, `crates/node/src/manager/epoch_votes.rs`), so the open subscription is broader
than the legitimate publisher set.

## Why it matters

This is the propagation-side analogue of the attribution fix in #889/#871 (which itself
addresses #801/#785/#819).  #889 stops an honest *relayer* from being penalized for a
content fault it could not screen, but it does not stop a non-committee node from getting
its consensus-output / epoch-vote gossip *propagated* in the first place.  Restricting the
publisher set closes that off at the front door for exactly the topics whose faults #889
now re-attributes (`UnknownConsensusHeaderCert` from `ConsensusResult`, `InvalidHeader`
from `EpochVote`), matching the treatment the certificate topic already gets, and keeping
[...]

PR #1057, opened 2026-07-26
network: gate worker batch-digest gossip subscription on node mode

### LINKED ISSUE #960: stop non-CVV nodes subscribing to batch-digest gossip, and correct stale sync comments
## Background

Workers gossip accepted batch digests on the `worker_batch_topic`.  A subscriber that hears a digest it does not already have runs a best-effort prefetch (`process_gossip` -> `request_batches` -> `validate_and_cache_prefetched_batch`), warming its `NodeBatchesCache` for the current epoch.

Under the new sync, nodes that only follow consensus no longer need this.  `crates/state-sync/src/lib.rs:71` routes both `NodeMode::Observer` and `NodeMode::CvvInactive` to "follow latest consensus from peers," and `crates/state-sync/src/consensus.rs` pulls a verified consensus output (header + batches) plus epoch pack files.  So the batches an observer needs arrive inside the consensus output / packs it already downloads;  the digest-gossip prefetch just fetches the same bytes a second time.

The gossip prefetch therefore only benefits nodes that consume individual current-epoch batches: active committee members and workers warming the cache for a committee they are about to join.

## Problem 1: the subscription is unconditional

Every node subscribes to `worker_batch_topic` every epoch, regardless of role.  In `crates/node/src/manager/node/start_epoch.rs` (`spawn_worker_network_for_epoch`), lines 697-705:

```rust
// Get gossip from committee members about batches every epoch.
// Useful for non-CVVs to prefetch and harmless for CVVs.
network_handle
    .inner_handle()
    .subscribe_with_publishers(
        tn_config::LibP2pConfig::worker_batch_topic(consensus_config.chain_id()),
        committee_keys.into_iter().collect(),
    )
    .await?;
```

`committee_keys` restricts who may *publish*, but every node still *subscribes* and receives every digest.  For an `Observer`, each received digest triggers a prefetch of a batch it will never use on the vote path, because it is not on the vote path.  The code already admits this in `process_gossip` (`crates/consensus/worker/src/network/handler.rs`, post-#938 lines 191-193):

```rust
// Note: retrieving this batch for no reason is wasteful, it should
// only effect nodes catching up old epochs though...
```

The waste is harmless to correctness (validated before caching by #938, and rate-limited by `try_admit_prefetch`), but it is real bandwidth on every follower for no benefit, which is exactly the case @sstanfield flagged.

### Proposed fix

Skip the `worker_batch_topic` subscription for nodes that only follow consensus.  Node mode is reachable at the subscription site through the consensus bus (`self.consensus_bus`, already used for other subscriptions in `start_epoch.rs`), via `node_mode()` / the `is_cvv()` / `is_observer()` helpers on `NodeMode` (`crates/consensus/primary/src/consensus_bus.rs:195-215`).

Two design points to settle in the issue, not assume:

- **Which modes skip.**  `Observer` clearly gains nothing and should skip.  `CvvInactive` also follows via the pack / consensus-output path while catching up (`state-sync/src/lib.rs:71` groups it with `Observer`), so the minimal-waste choice is to gate on `is_active_cvv()` (only `CvvActive` subscribes).  If we prefer inactive CVVs to warm their cache before rejoining, gate on `is_cvv()` instead and document why.  Recommend `is_active_cvv()` unless there is a measured cache-warming benefit for the inactive-to-active transition.
- **Mode is dynamic;  subscription is per-epoch.**  `node_mode` is a watch channel that can change mid-epoch (an inactive CVV catches up and becomes active), but the subscription is decided once in `spawn_worker_network_for_epoch`.  Decide whether a node promoted mid-epoch waits until the next epoch to start prefetching (simplest, probably fine given epoch cadence) or whether the subscription should react to a mode transition.  State the chosen behavior explicitly.

## Problem 2: stale comments about the sync model

Several comments still describe the old model in which non-committee / following peers obtain batches by requesting them off this gossip.  Under the new sync they obtain batches via packs / full consensus output, so these are now inaccurate (a docs-only pass, no behavior change).  Note only `handler.rs` is touched by #938;  the other files are unchanged by the PR, so this can be a separate small docs PR.

- `crates/consensus/worker/src/network/handler.rs:138-141` (post-#938) — `process_gossip` doc: "Workers gossip the Batch Digests once accepted so that non-committee peers can request the Batch."  Non-committee followers no longer request batches this way.
- `crates/consensus/worker/src/network/handler.rs:186-188` — "This allows non-CVVs to pre fetch batches they will soon need."  Non-CVV followers no longer consume batches via this path.
- `crates/consensus/worker/src/network/handler.rs:158-160` — "it will most likely be removed before we can use it and will be fetched later when needed."  For a node that is behind, the batch now arrives inside the downloaded consensus output / pack, not via a later per-batch fetch.
- `crates/consensus/worker/src/network/handler.rs:192-193` — "it should only effect nodes catching up old epochs though..."  Old-epoch catch-up is via epoch packs, and the epoch `ensure!` at line 161 already rejects non-current-epoch gossip, so this framing is doubly stale.
- `crates/consensus/worker/src/network/message.rs:13` — `WorkerGossip::Batch` variant doc still says the digest is published so non-committee peers can fetch it.
- `crates/consensus/worker/src/worker.rs:316-317` — publish-site comment says the digest is published for "any nodes listening to this gossip (non-committee members)."
- `crates/consensus/worker/src/network/mod.rs:67-72` — `MAX_CONCURRENT_GOSSIP_PREFETCHES` doc: keep the Byzantine-fan-out cap rationale (still accurate), but fix the "fetched on demand later" fallback to name the pack / consensus-output path as the authoritative follower sync.
- `crates/node/src/manager/node/start_epoch.rs:625-627` and `:697-698` — the subscription doc and inline comment ("Useful for non-CVVs to prefetch and harmless for CVVs") should reflect that observers/inactive CVVs no longer rely on this prefetch to follow.  If Problem 1 is fixed, this comment gets rewritten anyway.

Each should be reworded to: the digest gossip lets peers that still consume individual current-epoch batches (active committee members, and workers warming the cache for a committee they are about to join) prefetch best-effort, while observers and inactive CVVs sync via epoch packs / full consensus output (`crates/state-sync`).

## Acceptance criteria

- [ ] A node that only follows consensus (`Observer`, and per the decision above possibly `CvvInactive`) no longer subscribes to `worker_batch_topic`, so it receives no batch-digest gossip and runs no prefetches.
- [ ] The chosen gating predicate (`is_active_cvv()` vs `is_cvv()`) and the mid-epoch mode-transition behavior are documented in the code and PR.
- [ ] Active committee members still subscribe and prefetch exactly as before (no regression to the CVV vote path or cache warming).
- [ ] The comments listed in Problem 2 are updated to describe the pack / consensus-output follower sync;  no comment still claims non-committee peers fetch batches via this gossip.
- [ ] Doc comments are present on any item touched, per the repo's documentation bar.

MavenRain:
Closes #960.

## Problem

Every node subscribes to `worker_batch_topic` every epoch, regardless of role.  `spawn_worker_network_for_epoch` restricted who may *publish* to the committee, but every node still *subscribed* and received every accepted batch digest.  For a node that only follows consensus, each received digest triggered a best-effort prefetch (`process_gossip` -> `request_batches` -> `validate_and_cache_prefetched_batch`) of a batch body it was already downloading: `crates/state-sync` delivers observers a verified consensus output (header plus batches) and epoch pack files, so the digest gossip fetched the same bytes a second time.  Harmless to correctness, real bandwidth on every follower.

## Root cause

The subscription decision was never gated on node mode, and the gate could not simply be added, because two properties of the current design make a one-sided "skip the subscribe call" fix both ineffective and actively harmful:

1. **The worker swarm is process-lifetime.**  It is built once in `spawn_node_networks` (`node.rs`), outside the `run_epochs` loop, and `spawn_worker_network_for_epoch` runs "over the shared swarm" every epoch.  Gossipsub subscription state therefore survives epoch boundaries and mode-change re-entries.  There was no unsubscribe anywhere in the repo: `NetworkCommand` had a `Subscribe` variant and no counterpart, and `consensus.rs` only ever called `gossipsub.subscribe`.  A validator that subscribed in epoch N and rotated out to `Observer` in N+1 would keep receiving digest gossip forever.

2. **The subscribe call is the only writer of the topic's publisher allowlist.**  `NetworkCommand::Subscribe` is the sole mutation of `authorized_publishers` (an `insert`, with no `remove` / `clear` / `retain` anywhere).  Skipping it would freeze the entry on whichever committee was current when the node last subscribed.  `verify_gossip` documents the entry semantics as absent => not subscribed => reject, so a stale entry does not fail open, it fails *closed against the wrong set*: the current committee's honest batch gossip is rejected as `UnauthorizedAuthor`, is no longer re-propagated to downstream peers, and feeds `RejectPenalty::FatalAuthor` into `process_penalty` against honest validators.  Today that harm is contained only by the peer manager's prev/current/next committee exemption, so a skip-only gate would have silently depended on an unrelated safety net.

## Fix

- [x] **`network-libp2p/src/types.rs`** . . . new `NetworkCommand::Unsubscribe { topic, reply }` and `NetworkHandle::unsubscribe(topic) -> NetworkResult<bool>`, reporting whether the node was subscribed.  The reply is a bare `bool` because `libp2p-gossipsub` 0.49.4 (the pinned version) returns `bool` from `unsubscribe`, so the only error paths are channel failures and unsubscribing a never-subscribed topic is a benign no-op, not an error.  That matters: the epoch-start gate calls it on every observer's first epoch and propagates with `?`.
- [x] **`network-libp2p/src/consensus.rs`** . . . the mirror arm in `process_command`'s exhaustive match calls `gossipsub.unsubscribe(&sub)` **and** `authorized_publishers.remove(&topic)`.  The removal is required for the invariant above, not hygiene, and the comment says so.
- [x] **`node/src/manager/node/start_epoch.rs`** . . . the batch-topic decision is now two-sided and re-made every epoch: committee validators subscribe (which also refreshes the publisher allowlist for the new committee), everyone else unsubscribes.  Never a bare skip, for reason 1 above.
- [x] **`should_subscribe_batch_topic(mode)`** . . . the predicate is extracted as a pure function with an exhaustive match over `NodeMode`, so a new variant fails to compile until the decision is made for it, and so all three modes are unit-testable without an `EpochManager` harness (this file previously had no test module).
- [x] **Comment pass** . . . `worker/src/network/handler.rs` (the `process_gossip` doc, the "fetched later when needed" and "fetched on demand later" fallbacks, the inverted "this allows non-CVVs to pre fetch" claim, and the stale "only effect nodes catching up old epochs" note, now deleted), `worker/src/network/message.rs`, `worker/src/worker.rs`, `worker/src/network/mod.rs` (the Byzantine fan-out cap rationale is kept, only the fallback framing is corrected), and the subscription-policy comment hub in `node/src/manager/node.rs` that already enumerates the #898 / #912 topic decisions.

### The two design points the issue asked to settle

**Which modes skip: gate on `is_cvv()`, not `is_active_cvv()`.**  The issue recommended `is_active_cvv()` unless there is a measured cache-warming benefit for the inactive-to-active transition.  There is one, and it is structural rather than measured.  A `ModeChange` re-entry does not clear `NodeBatchesCache`: `clear_tables_for_next_epoch` is set only on `run_epoch`'s boundary arm and its leftover-drain arm, never on the mode-change arm.  The prefetch admits current-epoch digests.  So a `CvvInactive` node catching up warms a cache that *survives its own promotion*, and the vote path (`PrimaryReceiverHandler::synchronize`) reads that cache local-first.  Gossipsub does not backfill, so under `is_active_cvv()` a promoted node would re-subscribe and hold a subscription with a cold cache at exactly the moment it resumes voting at the live tip.  `is_cvv()` also matches the existing precedent in this same file, where the primary network's dial decision already gates on `self.consensus_bus.is_cvv()`.

**Mode is dynamic, subscription is per-epoch: no reactive subscription is needed, and not merely because the epoch cadence is short.**  Under `is_cvv()` the gate value provably cannot change mid-epoch.  Every mid-epoch mutation site that notifies shutdown (`primary/src/network/handler.rs`, `executor/src/subscriber.rs`, `primary/src/consensus/state.rs`) moves only between `CvvActive` and `CvvInactive`, which `is_cvv()` scores identically.  The only mutation to `Observer` is the observer-config branch at `EpochManager` construction, which runs once at process start.  The transitions that *do* flip `is_cvv()` are committee-membership changes, and those are recomputed at epoch entry by `identify_node_mode` (`!in_committee || tn_config.observer`), which `create_consensus` awaits strictly before it reaches the subscription site.  A mode transition also tears the epoch down and re-enters it as `RunEpochMode::ModeChange`, re-running the decision.  So the per-epoch decision is never stale.

### Threat model

The gate is attacker-independent: it is a local read of this node's own mode, taken once per epoch before any peer input, so it cannot be influenced by a peer and adds no work to the message path.  It only narrows what this node accepts, so it cannot widen any trust boundary.  Rotation correctness is the load-bearing part and is explicit above: the subscribe branch re-authorizes the entering committee, and the unsubscribe branch removes the entry entirely rather than leaving the outgoing committee's allowlist installed, which is what would have turned a bandwidth optimization into a penalty against honest current-committee workers.  Removing the prefetch cannot affect liveness or safety in any mode: it is best-effort by construction (the existing admission gate already drops digests silently when the concurrency cap is reached), a genuinely needed batch is still fetched by the vote path itself, and observers receive bodies inside the consensus output they follow.

## Testing

Run on the pinned toolchain (`rust-toolchain.toml` = 1.94), with the `-j 2` OOM cap and an isolated target dir so a sibling worktree's artifacts cannot produce a false green:

```
cargo +nightly-2026-03-20 fmt -- --check
cargo +1.94 clippy --all-targets
CARGO_TARGET_DIR=/Users/oobi/Documents/.tn-960-iso cargo +1.94 test --no-run --workspace -j 2
cargo +1.94 nextest run -p tn-network-libp2p -p tn-node
```

New tests:

- `network-libp2p/src/tests/network_tests.rs::test_unsubscribe_stops_gossip_delivery` . . . asserts unsubscribing a never-subscribed topic returns `false` rather than erroring (the observer first-epoch path), establishes a delivery baseline while subscribed, then unsubscribes and asserts the drop is reported, that the publisher stops observing this peer on the topic, and that publishing then fails for want of a subscriber.  Convergence uses `wait_until`, never `sleep`.
- `node/src/manager/node/start_epoch.rs::tests` . . . `should_subscribe_batch_topic` over all three `NodeMode` variants.

Scope note on what the tests do and do not cover: `authorized_publishers.remove` in the unsubscribe arm is not observable through the network API in a black-box test, because once gossipsub has unsubscribed, no message on that topic reaches `verify_gossip` for the map to be consulted.  It is asserted by construction (the two-sided gate means the entry is either removed or overwritten with the entering committee's set) and by the comment at the call site, not by the new test.  Calling it out rather than implying coverage.

### DIFF
--- a/crates/node/src/manager/node/start_epoch.rs
+++ b/crates/node/src/manager/node/start_epoch.rs
@@ -708,9 +708,13 @@ where
     /// committee peers — the peer manager drops dials to peers already connected — then waits
     /// for peers before spawning the network on the epoch-scoped spawner.
     ///
-    /// The batch topic is subscribed, restricted to committee publishers so non-CVVs can
-    /// prefetch batches (harmless for CVVs). Non-CVVs push the transactions they accept to
-    /// the committee over RPC rather than gossiping them (issue #804).
+    /// The batch topic is subscribed only by committee validators, restricted to committee
+    /// publishers, so they can prefetch batch bodies into `NodeBatchesCache` ahead of the vote
+    /// path. Observers skip it and unsubscribe: they receive batch bodies inside the verified
+    /// consensus output and epoch packs they already download (`crates/state-sync`), so the
+    /// digest gossip would only fetch the same bytes a second time (issue #960). Non-CVVs push
+    /// the transactions they accept to the committee over RPC rather than gossiping them
+    /// (issue #804).
     #[allow(clippy::too_many_arguments)]
     async fn spawn_worker_network_for_epoch(
         &mut self,
@@ -780,15 +784,28 @@ where
 
         Self::wait_for_network_peers(network_handle.inner_handle(), "worker network").await?;
 
-        // Get gossip from committee members about batches every epoch.
-        // Useful for non-CVVs to prefetch and harmless for CVVs.
-        network_handle
-            .inner_handle()
-            .subscribe_with_publishers(
-                tn_config::LibP2pConfig::worker_batch_topic(consensus_config.chain_id()),
-                committee_keys.into_iter().collect(),
-            )
-            .await?;
+        // Decide the batch-digest gossip subscription for this epoch (issue #960). Committee
+        // validators subscribe to warm the vote path's batch cache; observers unsubscribe,
+        // because state-sync already delivers batch bodies with the consensus output they
+        // follow, so the prefetch would refetch bytes they are downloading anyway.
+        //
+        // The decision is two-sided rather than a bare skip because the worker swarm is
+        // process-lifetime: a validator that subscribed in one epoch stays subscribed into every
+        // later epoch unless the subscription is explicitly dropped. Skipping alone would also
+        // skip the only refresh of this topic's authorized-publisher allowlist, freezing it on
+        // the committee that was current when the node last subscribed.
+        let batch_topic = tn_config::LibP2pConfig::worker_batch_topic(consensus_config.chain_id());
+        let mode = self.consensus_bus.current_node_mode();
+        if should_subscribe_batch_topic(mode) {
+            debug!(target: "epoch-manager", ?mode, "subscribing to worker batch topic");
+            network_handle
+                .inner_handle()
+                .subscribe_with_publishers(batch_topic, committee_keys.into_iter().collect())
+                .await?;
+        } else {
+            debug!(target: "epoch-manager", ?mode, "skipping worker batch topic - follows consensus output");
+            network_handle.inner_handle().unsubscribe(batch_topic).await?;
+        }
 
         // spawn worker network
         WorkerNetwork::new(
@@ -960,5 +977,48 @@ impl ReplayResult {
     /// Take `Self::last_replayed_hash` if it exists.
     pub(super) fn take_last_replayed_hash(&mut self) -> Option<ConsensusHeaderDigest> {
         self.last_replayed_hash.take()
+    }
+}
+
+/// Whether a node in this [`NodeMode`] should subscribe to the worker batch-digest gossip topic.
+///
+/// Only nodes that consume individual current-epoch batches benefit: a committee validator
+/// prefetches the body into `NodeBatchesCache` so the vote path finds it locally instead of
+/// fetching it on demand. That includes `CvvInactive` — a validator catching up to rejoin warms
+/// the cache it will vote against, and a mode-change re-entry does not clear that cache, so the
+/// warm-up survives its promotion. An `Observer` never votes and receives batch bodies inside the
+/// verified consensus output and epoch packs it already downloads (`crates/state-sync`), so for it
+/// the prefetch is pure duplicate bandwidth.
+///
+/// Equivalent to [`ConsensusBus::is_cvv`], written as an exhaustive match so a new [`NodeMode`]
+/// variant fails to compile until this decision is made for it.
+fn should_subscribe_batch_topic(mode: NodeMode) -> bool {
+    match mode {
+        NodeMode::CvvActive | NodeMode::CvvInactive => true,
+        NodeMode::Observer => false,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{should_subscribe_batch_topic, NodeMode};
+
+    /// An active committee validator prefetches batches for the vote path.
+    #[test]
+    fn active_cvv_subscribes_to_batch_topic() {
+        assert!(should_subscribe_batch_topic(NodeMode::CvvActive));
+    }
+
+    /// A validator catching up to rejoin warms the cache it is about to vote against; the cache
+    /// survives the mode-change re-entry that promotes it.
+    #[test]
+    fn inactive_cvv_subscribes_to_batch_topic() {
+        assert!(should_subscribe_batch_topic(NodeMode::CvvInactive));
+    }
+
+    /// An observer follows consensus output and would only refetch bytes it already downloads.
+    #[test]
+    fn observer_does_not_subscribe_to_batch_topic() {
+        assert!(!should_subscribe_batch_topic(NodeMode::Observer));
     }
 }
--- a/crates/consensus/worker/src/network/handler.rs
+++ b/crates/consensus/worker/src/network/handler.rs
@@ -137,8 +137,11 @@ where
 
     /// Process gossip from the committee.
     ///
-    /// Workers gossip the Batch Digests once accepted so that non-committee peers can request the
-    /// Batch.
+    /// Workers gossip a batch digest once it is accepted so that peers which consume individual
+    /// current-epoch batches — committee validators, including one catching up to rejoin — can
+    /// prefetch the body into `NodeBatchesCache` before the vote path needs it. Observers do not
+    /// subscribe to this topic (issue #960): they receive batch bodies inside the verified
+    /// consensus output and epoch packs they already download (`crates/state-sync`).
     pub(super) async fn process_gossip(&self, msg: &GossipMessage) -> WorkerNetworkResult<()> {
         // deconstruct message
         let GossipMessage { data, source: _, sequence_number: _, topic } = msg;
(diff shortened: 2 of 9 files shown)


crates/node/src/manager/node/start_epoch.rs
```diff
@@ -962,3 +979,46 @@ impl ReplayResult {
...
+/// the cache it will vote against, and a mode-change re-entry does not clear that cache, so the
+/// warm-up survives its promotion. An `Observer` never votes and receives batch bodies inside the
+/// verified consensus output and epoch packs it already downloads (`crates/state-sync`), so for it
+/// the prefetch is pure duplicate bandwidth.
+///
+/// Equivalent to [`ConsensusBus::is_cvv`], written as an exhaustive match so a new [`NodeMode`]
+/// variant fails to compile until this decision is made for it.
+fn should_subscribe_batch_topic(mode: NodeMode) -> bool {
+    match mode {
+        NodeMode::CvvActive | NodeMode::CvvInactive => true,
```
sstanfield, line 997: Note that currently an inactive CVV is just an observer for the epoch.  So it probably should not subscribe either.

sstanfield (approved):
LGTM and overdue.  Thanks for taking this.

grantkee (approved):
Well done

crates/node/src/manager/node/start_epoch.rs
grantkee: CvvInactive can also be syncing within the same epoch if it falls outside the GC window. Would this help it sync more quickly?

PR #1058, opened 2026-07-26
fix(tn-rpc): run the proofOfPossessionMessage BLS decompress off the shared runtime

### LINKED ISSUE #976: tn-rpc: unauthenticated `proofOfPossessionMessage` runs inline BLS G2 decompress on the shared consensus runtime (liveness/availability)
## Summary

Impact class: availability / liveness (consensus scheduling degradation), reachable by an anonymous caller.

The `tn` JSON-RPC method `proofOfPossessionMessage` performs a BLS12-381 G2 point decompress (plus an on-curve check) INLINE on the async runtime, inside a yield-free `async fn` body with no `spawn_blocking`, no semaphore permit, and no `.await` yield point. That method is served from the public, unauthenticated `tn` namespace (merged into the general HTTP/WS/IPC transport, not reth's JWT-authenticated engine server), and the RPC server shares ONE tokio multi-thread runtime with the consensus primary/worker tasks, built with no `worker_threads` cap. A flood of unauthenticated `tn_proofOfPossessionMessage` calls therefore pins async worker threads with CPU-bound crypto and can starve consensus task scheduling, not just the RPC surface.

Severity note: this is a liveness/availability degradation of consensus driven by unauthenticated RPC. Per-call cost is sub-millisecond (order tens of microseconds), so the amplification asymmetry is weaker than a classic amplification DoS, but the path is anonymously reachable, yield-free, and concurrency-uncapped (the existing 64-permit guard does not cover it).

## Defect

The handler runs the BLS parse and message construction inline, with no offload and no yield point.

`crates/execution/tn-rpc/src/rpc_ext.rs:374-387`
```rust
    async fn proof_of_possession_message(
        &self,
        bls_pubkey: Bytes,
        validator_address: Address,
    ) -> TelcoinNetworkRpcResult<Bytes> {
        // As of the compressed-bytes pivot the registry's `proofOfPossessionMessage` is an internal
        // helper (not externally callable), so build the message natively from the same `tn_types`
        // routine the validator signs and the registry reproduces on-chain byte-for-byte:
        // `intent || compressed pubkey || address`.
        let pubkey = BlsPublicKey::from_literal_bytes(&bls_pubkey).map_err(|e| {
            TNRpcError::InvalidParams(format!("invalid 96-byte compressed BLS pubkey: {e:?}"))
        })?;
        Ok(construct_proof_of_possession_message(&pubkey, &validator_address).into())
    }
```

The parse delegates straight to `blst`:

`crates/types/src/crypto/bls_public_key.rs:99-101`
```rust
    pub fn from_literal_bytes(bytes: &[u8]) -> Result<Self, BLST_ERROR> {
        CorePublicKey::from_bytes(bytes).map(|key| key.into())
    }
```

`crates/types/src/crypto/bls_public_key.rs:3`
```rust
use blst::{min_sig::PublicKey as CorePublicKey, BLST_ERROR};
```

`CorePublicKey` is `blst::min_sig::PublicKey`, i.e. a BLS12-381 G2 point (96-byte compressed in the min-signature scheme). `from_bytes` performs the compressed-point deserialize (decompress + on-curve check), whose dominant cost is an Fp2 modular square root to recover y:

`blst-0.3.16/src/lib.rs:869-886`
```rust
            pub fn deserialize(pk_in: &[u8]) -> Result<Self, BLST_ERROR> {
                if (pk_in.len() == $pk_ser_size && (pk_in[0] & 0x80) == 0)
                    || (pk_in.len() == $pk_comp_size && (pk_in[0] & 0x80) != 0)
                {
                    let mut pk = <$pk_aff>::default();
                    let err = unsafe { $pk_deser(&mut pk, pk_in.as_ptr()) };
                    if err != BLST_ERROR::BLST_SUCCESS {
                        return Err(err);
                    }
                    Ok(Self { point: pk })
                } else {
                    Err(BLST_ERROR::BLST_BAD_ENCODING)
                }
            }

            pub fn from_bytes(pk_in: &[u8]) -> Result<Self, BLST_ERROR> {
                PublicKey::deserialize(pk_in)
            }
```

Precision note (do not overstate): this path uses `from_bytes` / `deserialize`, which does NOT call `validate()`, so it does NOT perform the KeyValidate subgroup-membership or infinity check. Only `key_validate` (below) would add those, and it is not on this path. The op is a G2 decompress plus on-curve check, still an unbounded pairing-curve point operation on the async thread:

`blst-0.3.16/src/lib.rs:812-828`
```rust
            pub fn validate(&self) -> Result<(), BLST_ERROR> {
                unsafe {
                    if $pk_is_inf(&self.point) {
                        return Err(BLST_ERROR::BLST_PK_IS_INFINITY);
                    }
                    if !$pk_in_group(&self.point) {
                        return Err(BLST_ERROR::BLST_POINT_NOT_IN_GROUP);
                    }
                }
                Ok(())
            }

            pub fn key_validate(key: &[u8]) -> Result<Self, BLST_ERROR> {
                let pk = PublicKey::from_bytes(key)?;
                pk.validate()?;
                Ok(pk)
            }
```

The method is exposed under the public `tn` namespace with no auth guard:

`crates/execution/tn-rpc/src/rpc_ext.rs:39-40`
```rust
#[rpc(server, namespace = "tn")]
pub trait TelcoinNetworkRpcExtApi {
```

`crates/execution/tn-rpc/src/rpc_ext.rs:146-151`
```rust
    #[method(name = "proofOfPossessionMessage")]
    async fn proof_of_possession_message(
        &self,
        bls_pubkey: Bytes,
        validator_address: Address,
    ) -> TelcoinNetworkRpcResult<Bytes>;
```

The tn methods are merged into the general reth transport module set (HTTP/WS/IPC), not the JWT-authenticated engine/auth server:

`crates/tn-reth/src/lib.rs:1212-1217`
```rust
        let mut server = rpc_builder.build(modules_config, eth_api, engine_events);
        if let Err(e) = server.merge_configured(other) {
            tracing::error!(target: "tn::execution", "Error merging TN rpc module: {e:?}");
        }

        server
```

The only concurrency guard in this crate (`blocking_io_guard`, `MAX_CONCURRENT_REGISTRY_READS = 64`) wraps ONLY the EVM registry-read path via `spawn_registry_read`, so it does not cover `proof_of_possession_message` at all:

`crates/execution/tn-rpc/src/rpc_ext.rs:175`
```rust
const MAX_CONCURRENT_REGISTRY_READS: usize = 64;
```

`crates/execution/tn-rpc/src/rpc_ext.rs:444-466`
```rust
    async fn spawn_registry_read<R, F>(&self, read: F) -> TelcoinNetworkRpcResult<R>
    where
        R: Send + 'static,
        F: FnOnce(RethEnv) -> EvmReadResult<R> + Send + 'static,
    {
        // bound concurrent blocking reads; queues (does not reject) at capacity, mirroring
        // reth's eth_call guard. acquire only fails if the semaphore is closed, which never
        // happens because it lives as long as the RPC server.
        let permit = self.blocking_io_guard.clone().acquire_owned().await.map_err(|e| {
            tracing::warn!(target: "tn::rpc", error = %e, "registry read semaphore closed");
            TNRpcError::Internal
        })?;

        let evm = self.evm_state.clone();
        self.evm_state.get_task_spawner().spawn_blocking_task("tn-rpc-registry-read", move || {
            // hold the permit until the blocking read completes
            let _permit = permit;
            if tx.send(read(evm)).is_err() {
                tracing::debug!(target: "tn::rpc", "registry read receiver dropped before result");
            }
            Ok(())
        });
```

The registry-backed `tn` methods (`get_validators`, `get_bls_pubkey`, and the other `ConsensusRegistry::*Call` reads) dispatch through the semaphore-guarded `registry_read` / `spawn_registry_read` EVM path; the remaining methods (`info`, `latest_header`, `genesis`, `epoch_record*`) route through `inner_node_network` / `chainspec`. None of them perform inline BLS work, and the 64-permit guard covers none of them. `proof_of_possession_message` is the sole handler performing heavy crypto directly on the async worker thread.

## Reproduction

1. Build and start a node via the `node` subcommand with the default general HTTP/WS RPC enabled. This merges the `tn` namespace into the public transport (`merge_configured`, `crates/tn-reth/src/lib.rs:1212-1217`), with no JWT/auth guard on that namespace.
2. Confirm anonymous reachability: send a plain JSON-RPC request to `tn_proofOfPossessionMessage` with a valid 96-byte compressed BLS pubkey and a 20-byte address over unauthenticated HTTP/WS. A valid input forces the decompress to run to completion (rather than erroring early on encoding).
3. Launch N concurrent clients issuing `tn_proofOfPossessionMessage` in a tight loop with valid 96-byte inputs, scaling N up to and beyond the host core count (tokio's default `worker_threads`, which is uncapped here).
4. Sample tokio worker-thread occupancy alongside consensus round latency / task scheduling metrics (primary/worker task turnaround).
5. Observe that worker threads are pinned by the yield-free BLS G2 decompress (no `.await`, no `spawn_blocking`), and that consensus task scheduling latency grows as request concurrency approaches and exceeds the worker-thread count, because the RPC handler and consensus tasks contend for the same runtime with no isolation and no concurrency bound on this method.

## Root cause

The handler performs CPU-bound, yield-free crypto directly on the async runtime worker thread with no offload (`spawn_blocking`), no semaphore, and no auth (`crates/execution/tn-rpc/src/rpc_ext.rs:374-387`), while the node runs the RPC server and consensus on ONE shared tokio multi-thread runtime with no `worker_threads` cap.

The single runtime is built at `crates/telcoin-network-cli/src/cli.rs:153-159` (note: `enable_io()` + `enable_time()`, not a literal `enable_all()`, and no `.worker_threads(...)`, so tokio defaults to core count, uncapped):
```rust
                let runtime = Builder::new_multi_thread()
                    .thread_name("telcoin-network")
                    .enable_io()
                    .enable_time()
                    .build()?;

                let res = runtime.block_on(async move {
```

The RPC server is started as an `.await` on that same ambient runtime (`crates/node/src/engine/inner.rs:165-166`):
```rust
        // start the RPC server
        let rpc_handle = self.reth_env.start_rpc(&server).await?;
```

`RethEnv::start_rpc` constructs no runtime of its own; it starts the jsonrpsee server on the current runtime (`crates/tn-reth/src/lib.rs:1221-1224`):
```rust
    pub async fn start_rpc(&self, server: &RpcServer) -> TnRethResult<RpcServerHandle> {
        let server_config = self.inner.node_config.rpc.rpc_server_config();
        Ok(server_config.start(server).await?)
    }
```

Consensus tasks are spawned onto that same ambient runtime through the task manager's `tokio::spawn` (`crates/types/src/task_manager.rs:181-198`):
```rust
    /// The main function to spawn tasks on the `tokio` runtime. These tasks are tracked by the
    /// manager.
    fn create_task<F, S: ToString>(&self, name: S, future: F, critical: bool)
    where
    {
        let name = name.to_string();
        let rx_shutdown = self.local_shutdown.subscribe();
        let handle = tokio::spawn(async move {
            tokio::select! {
                _ = rx_shutdown => {
                    Ok(())
                }
                res = future => {
                    res
                }
            }
        });
```

Because RPC handlers and consensus tasks share this uncapped runtime, unauthenticated inline crypto on an RPC worker thread directly competes with consensus scheduling.

## Why not already covered

This is a distinct subsystem from the durability cluster (issues #961 to #964); it is an RPC-runtime scheduling/liveness defect, not a storage-durability one.

Coverage search result: an `rg` over `crates/execution/tn-rpc` and `crates/telcoin-network-cli` for `spawn_blocking|starv|dos|denial|#759|worker_threads` produced no matches describing this path. The only `spawn_blocking_task` hit is at `rpc_ext.rs:459`, which is the registry-read path, not the PoP endpoint. No CHANGELOG file exists in the repo.

The nearest in-tree guard was added by #736 (`Epoch Info RPC`, commit `0bd36414`): the `blocking_io_guard` semaphore and `MAX_CONCURRENT_REGISTRY_READS = 64` (`rpc_ext.rs:163-175`). That guard bounds a DIFFERENT DoS (unbounded tokio blocking-pool growth from concurrent `registry_read` -> `spawn_blocking_task` calls) and, by construction, does not cover `proof_of_possession_message`, which never routes through `registry_read` / `spawn_blocking`.

Lineage of the PoP handler (no regression claim): `git log -S 'spawn_blocking'` and `-S 'Semaphore'` over `crates/execution/tn-rpc` each return only commit `0bd36414` (#736). Before #759, `proof_of_possession_message` called `self.registry_read(calldata).await` against an on-chain `ConsensusRegistry::proofOfPossessionMessageCall`, so it routed through the semaphore-guarded path, but the guarded work there was an EVM/DB read, not an in-node BLS crypto op. Commit `42f74b33` (#759) replaced that on-chain call with the inline native `BlsPublicKey::from_literal_bytes(...)` + `construct_proof_of_possession_message(...)` construction, with no permit and no `spawn_blocking`. There is no git evidence that any commit ever wrapped the BLS crypto op itself in a `spawn_blocking` + semaphore guard; the guard (from #736) only ever protected the on-chain registry read. Accordingly this is filed as "runs inline, unguarded", not as a regression of a guard around the BLS op. Issues #731, #732, and #759 concern the on-chain EVM BLS precompile in `crates/tn-reth/src/evm/bls_precompile/`, not RPC runtime starvation.

Conclusion: no existing issue or PR covers this specific defect (unauthenticated inline BLS G2 decompress in `proofOfPossessionMessage` starving the shared consensus async runtime).

## Suggested fix

1. Offload the BLS parse/decompress off the async worker threads: wrap the `BlsPublicKey::from_literal_bytes` call (and any follow-on crypto) in `spawn_blocking` (or the existing task-spawner blocking pool), mirroring the `spawn_registry_read` pattern at `rpc_ext.rs:444-466`.
2. Bound concurrency for this path with a semaphore permit held for the duration of the blocking crypto (reuse `blocking_io_guard` or add a dedicated guard sized for BLS work), so a request flood cannot occupy unbounded worker/blocking-pool capacity.
3. Isolate RPC compute from consensus: consider a dedicated blocking pool for RPC crypto and/or a `worker_threads` floor on the shared runtime (`crates/telcoin-network-cli/src/cli.rs:153-159`) so that RPC compute cannot starve consensus task scheduling.
4. Rate-limit and/or authenticate `tn_proofOfPossessionMessage` (and reconsider exposing the compute-bearing `tn` method on the unauthenticated public transport), so anonymous callers cannot drive uncapped crypto.

## Test design

1. Confirm-by-mutation liveness test: stand up a node with consensus running, saturate `tn_proofOfPossessionMessage` with concurrent valid-input requests, and assert that consensus task scheduling latency (or round latency) stays within a bounded threshold. To confirm the test is non-vacuous, apply a mutation that removes the offload/semaphore (i.e. restore the current inline path) and verify the test FAILS (latency blows past the bound), then restore the fix and verify it passes.
2. Off-worker-thread assertion (unit/integration): assert that the BLS decompress for `proof_of_possession_message` executes off the async runtime's worker threads (for example by asserting it runs on a blocking-pool thread / distinct thread-name, or by observing that the async worker is free to make progress on a cooperating future while the crypto is in flight), so a regression that reinlines the crypto is caught.
3. Concurrency-bound assertion: drive more concurrent PoP requests than the configured permit count and assert that in-flight blocking crypto never exceeds the semaphore bound (queueing, not unbounded fan-out).

## Acceptance criteria

- The BLS decompress in `proof_of_possession_message` no longer runs inline on the async runtime worker threads; it is offloaded via `spawn_blocking` (or the blocking task pool).
- A bounded semaphore caps concurrent PoP crypto work, with the permit held for the full duration of the blocking op.
- RPC compute is isolated from consensus (dedicated blocking pool and/or a `worker_threads` floor on the shared runtime) so RPC load cannot starve consensus scheduling.
- `tn_proofOfPossessionMessage` is rate-limited and/or authenticated (or otherwise removed from the unauthenticated public transport).
- A regression test saturates the method and asserts consensus task scheduling latency stays bounded, and is proven non-vacuous by confirm-by-mutation.
- A test asserts the BLS op runs off the shared runtime worker threads.

## Assumptions to verify

- jsonrpsee handler execution model: `RpcServerConfig::start` is external (reth/jsonrpsee) and its internals are not quoted here. The consensus-starvation (as opposed to RPC-surface-only) conclusion assumes jsonrpsee drives the async `proof_of_possession_message` handler on the shared multi-thread runtime's worker threads rather than a dedicated pool. If jsonrpsee isolates request handling on its own threads, the impact narrows to RPC-surface degradation. Confirm against the jsonrpsee/reth server configuration in use.
- Public-interface exposure: the absence of an auth guard on the `tn` namespace is confirmed in code, but whether the shipped node config binds the general HTTP/WS/IPC transport to a public interface (rather than localhost or a firewalled interface) is deployment-dependent and not evidenced here.
- Amplification magnitude: per-call cost is an order-of-magnitude estimate (sub-millisecond, tens of microseconds); the concrete request rate and concurrency needed to measurably degrade consensus round latency on a representative validator was not benchmarked.
- blst variant (resolved): this issue is written to the source-quoted `blst-0.3.16` reading, in which `from_bytes` / `deserialize` performs a G2 decompress plus on-curve check but NOT the `key_validate` subgroup or infinity check. Confirm the compiled blst version matches.

MavenRain:
Closes #976.

## Problem

`tn_proofOfPossessionMessage` is served from the unauthenticated `tn` namespace and decompresses a BLS12-381 G2 point inline in the handler body.  That body contains no `.await`, so once a runtime worker thread picks it up, nothing else scheduled on that thread runs until the crypto finishes . . . including consensus tasks, which share the same runtime.

Measured on this branch (blst 0.3.16, release, Apple M2 Pro performance core):

| operation | cost per call |
| --- | --- |
| decompress a 96-byte compressed key | 24.8 us |
| decompress 96 arbitrary bytes with the compression bit set | 24.8 us |
| decompress a 192-byte uncompressed key | 0.51 us |
| re-compress a parsed key | 0.10 us |
| reject a wrong-length input | 0.0012 us |

Two things follow from those numbers.  An attacker needs no key material at all: 96 bytes that are not on the curve cost the server exactly as much as a real validator key, because the square root runs before the on-curve check can fail.  And a wrong-length input is essentially free, because blst screens the length before it touches the curve, so the expensive case is specifically a correctly sized one.

The per-request cost understates the exposure, because a caller does not have to send one call per request.  jsonrpsee runs the entries of a JSON-RPC batch sequentially inside a single task (`jsonrpsee-server-0.26.0/src/middleware/rpc.rs:157-160`), and batching is unlimited: jsonrpsee defaults to `BatchRequestConfig::Unlimited` (`server.rs:366`) and reth never calls `set_batch_request_config`, so the only ceiling is the default 15 MB request body (`RPC_DEFAULT_MAX_REQUEST_SIZE_MB`).  At roughly 340 bytes per batch entry that is about 44,000 calls in one HTTP request, roughly 1.1 s of a single worker thread with no yield point anywhere inside it.  The default connection limit is 500.

Threat model: no privilege, no stake, no key material, no prior handshake.  Any anonymous client that can reach the general HTTP/WS/IPC transport.

## Root cause

This is the only handler in the namespace that does its work inline.  Every registry-backed `tn` method already goes through `spawn_registry_read`, which takes a permit and runs on the blocking pool specifically so that, in the words of its own doc comment, "consensus-critical reads are never throttled by RPC traffic".

`proof_of_possession_message` used to be one of those: it called the on-chain `ConsensusRegistry::proofOfPossessionMessageCall` through the guarded path.  #759 replaced that with native in-process construction, which is the right call for correctness and cost, but it also took the handler off the guarded path and nothing replaced the guard.  So this is not a regression of a check that once wrapped the BLS work.  That work has never had one;  the #736 guard only ever protected the EVM read.

The shared-runtime premise is confirmed rather than assumed.  The node builds exactly one `Builder::new_multi_thread()` (`telcoin-network-cli/src/cli.rs:153`) with no `worker_threads` override.  `RethEnv::start_rpc` never attaches a dedicated runtime, so jsonrpsee's `ServerConfig.tokio_runtime` stays at its `None` default and both the accept loop and every per-connection task are plain `tokio::spawn` on that runtime.  Consensus tasks reach the same runtime through the same `tokio::spawn` (`types/src/task_manager.rs:189`).  An async trait method is dispatched inline by `register_async_method`, with no offload, which is what lets a yield-free body hold a worker.

## Changes

- [x] `tn-rpc/src/rpc_ext.rs`: new `spawn_bounded_blocking` . . . acquire a permit, dispatch to the task spawner's blocking pool, and move the permit into the task so it is released when the work finishes rather than when the caller stops waiting.  `spawn_registry_read` is now expressed on top of it, so the two endpoints share one implementation of the permit-and-dispatch dance instead of keeping two copies in step.
- [x] `tn-rpc/src/rpc_ext.rs`: `proof_of_possession_message` dispatches the decompress through it.  Per-call latency goes from 24.8 us to 33.7 us;  the worker thread is free for all of it, and in the batch case the await between entries is what lets consensus interleave.
- [x] `tn-rpc/src/rpc_ext.rs`: cheap-first screen.  An impossible encoding length is rejected inline, before a permit or a pool thread is committed, so a flood of malformed keys cannot displace real work: 0.0012 us to reject against a 5.8 us round-trip to dispatch it.
- [x] `tn-rpc/src/rpc_ext.rs`: `MAX_CONCURRENT_REGISTRY_READS` becomes `MAX_CONCURRENT_BLOCKING_RPC_WORK`, one bound over both kinds of blocking work.  A second guard sized independently would let the two worst cases add up and quietly double the pool occupancy the existing 64 was chosen to permit.
- [x] `tn-types/src/crypto/bls_public_key.rs`: `COMPRESSED_BYTES`, `UNCOMPRESSED_BYTES` and `is_plausible_encoding`, so the screen and the parser agree by construction instead of the RPC layer hard-coding 96 and 192 next to a `blst` call that owns those numbers.
- [x] Doc fix: `proofOfPossessionMessage` also accepts the 192-byte uncompressed encoding and normalizes it to the compressed form.  The trait doc claimed 96 only.

Behaviour for a successful call is unchanged: both encodings still produce the same message bytes, and a bad point still returns `InvalidParams`.  Two error paths do change, both deliberately:

- A wrong-length key now returns an `InvalidParams` naming the expected lengths, instead of a bare blst code.
- The endpoint gains an `Internal` (-32603) failure class it did not have when it ran inline, reachable if the blocking dispatch cannot deliver (realistically only in a shutdown window, where the runtime drops a queued closure).  Nothing in the repo or in `crates/e2e-tests` asserts on either message.

`TNRpcError::Internal`'s text moves from "consensus registry read failed" to "internal error" for that reason: the variant is now returned by an endpoint that performs no registry read, and its whole design is to say nothing and log the cause server-side.

## On sharing one guard

This is the part most worth arguing with, so both directions, not just the flattering one.

The direction this PR creates is registry-read to PoP.  An EVM read holds its permit for milliseconds;  a PoP call holds one for tens of microseconds.  Since the semaphore is FIFO, a PoP call arriving behind a saturated queue of registry reads now waits where before it waited for nothing at all, and the wait queue itself has no depth cap.

The reverse direction, PoP to registry read, is the benign one: in-flight PoP calls are capped by the 500-connection limit because batch entries run sequentially per connection, permits turn over every ~30 us, and a registry read arriving behind a full queue waits on the order of a millisecond.

What makes the trade worth taking is that the status quo is not "registry reads keep their own 64 permits".  It is a flood that holds runtime worker threads, which stalls the registry endpoints, consensus, and the rest of the RPC surface at once.  Bounded queueing against a fair semaphore is a better failure mode than that, and a second independently sized guard would restore the unbounded-occupancy problem the first one exists to prevent.  If reviewers would rather keep the endpoints isolated, the alternative is two guards summing to 64 rather than two guards of 64.

## Testing

```
cargo +nightly-2026-03-20 fmt -- --check
cargo +nightly-2026-03-20 clippy --workspace -- -D warnings
cargo +nightly-2026-03-20 clippy --workspace --all-features -- -D warnings
cargo +1.94 test --no-run --workspace
cargo +1.94 test -p tn-rpc -p tn-types --lib
```

All green.  `tn-rpc` 10 passed, `tn-types` 91 passed.

Every new test was confirmed by mutation, meaning each was seen to fail against a deliberately broken build before being trusted:

| mutation applied | test that failed |
| --- | --- |
| length screen removed | `malformed_key_is_rejected_before_taking_a_permit` |
| permit acquired from a throwaway semaphore | `only_the_expensive_path_consults_the_guard`, `permit_is_held_until_the_work_completes` |
| work run inline instead of dispatched | `blocking_work_runs_off_the_async_thread` |
| permit released at dispatch instead of held | `permit_is_held_until_the_work_completes` |
| `acquire_owned().await` swapped for `try_acquire_owned()` | `at_capacity_a_well_formed_key_queues_rather_than_failing` |
| screen accepts every length | `screen_rejects_only_what_the_parser_would_reject` |
| screen drops the uncompressed length | `encoding_length_constants_match_blst`, `screen_rejects_only_what_the_parser_would_reject` |

The queue-not-reject case earns its place: under the `try_acquire_owned` mutation every other test stays green, because they all run with either a free permit or a closed semaphore, and a closed semaphore fails both forms identically.  Only an exhausted-but-open guard tells them apart.

No test asserts on wall-clock time.  The offload is pinned by comparing thread ids across the dispatch.  The two request paths are separated by closing the semaphore, which fails any acquisition immediately, so a well-formed key surfaces `Internal` while a malformed one is still answered by the screen;  that distinguishes them without waiting on anything.  `screen_rejects_only_what_the_parser_would_reject` asserts which lengths survive before asserting the implication over the rest, because a screen that accepted everything would otherwise empty the set and pass vacuously.

## Out of scope

The issue also lists rate limiting or authentication for the method, and a `worker_threads` floor.  Those are deliberately not here:

- Authenticating `tn_*` or removing it from the public transport is a decision about who may call the namespace, not a defect fix.
- A batch-size limit is the most valuable remaining hardening, since batching is what turns one request into 44,000 calls.  It is set on reth's `ServerConfigBuilder` and would apply to the whole node RPC surface, including `eth_*` where clients batch legitimately, so it deserves its own issue and its own default.  Happy to open one.
- A `worker_threads` floor changes process-wide scheduling for every deployment.

What this PR removes is the endpoint's ability to hold worker threads at all, which is the part that is unambiguously a defect.  The aggregate CPU a flood can consume is a rate-limiting concern and is unchanged.

One measurement noted for the record and deliberately not acted on: `construct_proof_of_possession_message` re-compresses the key through `to_bytes()` rather than reading the already-cached bytes, so the handler performs one decompress and two compresses.  At 0.10 us against 24.8 us that is 0.4% of the call and not the lever here, and the function is on consensus paths, so it does not belong in a DoS fix.

### DIFF
--- a/crates/execution/tn-rpc/Cargo.toml
+++ b/crates/execution/tn-rpc/Cargo.toml
@@ -21,6 +21,7 @@ tracing = { workspace = true }
 
 [dev-dependencies]
 alloy = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "sync"] }
 
 [lints]
 workspace = true
--- a/crates/execution/tn-rpc/src/error.rs
+++ b/crates/execution/tn-rpc/src/error.rs
@@ -31,8 +31,12 @@ pub enum TNRpcError {
         /// Raw ABI-encoded revert bytes, hex-encoded into the error `data` field.
         output: Bytes,
     },
-    /// Internal failure reading the registry. Detail logged server-side only.
-    #[error("consensus registry read failed")]
+    /// Internal failure serving the request. Detail logged server-side only.
+    ///
+    /// Deliberately says nothing about which endpoint or stage failed: it is returned for registry
+    /// reads and for the blocking-dispatch failures behind `proofOfPossessionMessage` alike, and
+    /// the cause is only ever logged server-side.
+    #[error("internal error")]
     Internal,
     /// Caller supplied an invalid parameter (e.g. a malformed BLS public key).
     #[error("{0}")]
@@ -101,7 +105,7 @@ mod tests {
     fn internal_maps_to_internal_error_without_detail() {
         let err: ErrorObject<'static> = TNRpcError::Internal.into();
         assert_eq!(err.code(), -32603);
-        assert_eq!(err.message(), "consensus registry read failed");
+        assert_eq!(err.message(), "internal error");
         assert!(err.data().is_none());
     }
 
--- a/crates/execution/tn-rpc/src/rpc_ext.rs
+++ b/crates/execution/tn-rpc/src/rpc_ext.rs
@@ -16,7 +16,7 @@ use tn_reth::{
 use tn_types::{
     construct_proof_of_possession_message, Address, BlsPublicKey, Bytes, ConsensusHeader, Epoch,
     EpochCertificate, EpochDigest, EpochRecord, Genesis, NodeMode, SolCall, SolType, SolValue,
-    B256, U256,
+    TaskSpawner, B256, U256,
 };
 use tokio::sync::{oneshot, Semaphore};
 
@@ -149,7 +149,8 @@ pub trait TelcoinNetworkRpcExtApi {
     /// Return the BLS12-381 proof-of-possession message a validator signs:
     /// `intentPrefix(3) || compressedBlsPubkey(96) || validatorAddress(20)`.
     ///
-    /// Expects the 96-byte compressed BLS public key.
+    /// Expects the 96-byte compressed BLS public key. The 192-byte uncompressed encoding is also
+    /// accepted and normalized to the compressed form, so both yield the same message.
     #[method(name = "proofOfPossessionMessage")]
     async fn proof_of_possession_message(
         &self,
@@ -167,19 +168,116 @@ pub trait TelcoinNetworkRpcExtApi {
     async fn undistributed_issuance(&self) -> TelcoinNetworkRpcResult<U256>;
 }
 
-/// Maximum number of concurrent blocking [`ConsensusRegistry`] reads served by the `tn`
-/// namespace.
+/// Maximum number of concurrent blocking tasks the `tn` namespace may have in flight.
 ///
-/// Each public read builds a fresh EVM and state snapshot at the canonical tip - synchronous
-/// database/CPU work on a blocking thread. A semaphore permit is held for the full lifetime of
-/// the blocking work, so this bounds true blocking-pool occupancy, not just in-flight requests.
+/// Covers both kinds of synchronous work this namespace dispatches off the async runtime:
+/// [`ConsensusRegistry`] reads, which build a fresh EVM and state snapshot at the canonical tip
+/// (database plus CPU), and the BLS12-381 G2 decompress behind `proofOfPossessionMessage` (pure
+/// CPU, tens of microseconds per call). A semaphore permit is held for the full lifetime of the
+/// blocking work, so this bounds true blocking-pool occupancy, not just in-flight requests.
+///
+/// One bound covers both deliberately. The budget below is a share of a pool the whole process
+/// contends for, so splitting it into a guard per endpoint would let their worst cases add up and
+/// silently double the occupancy this number is chosen to permit.
 ///
 /// Mirrors reth's `eth_call` guard (`DEFAULT_MAX_BLOCKING_IO_REQUEST` = 256): tokio's blocking
 /// pool defaults to 512 threads and grows unbounded under request load without a bound. Set
 /// below reth's 256 because reth's own eth namespace shares the same pool; 64 keeps headroom
 /// for BLS signing and engine blocking tasks. Lives at the RPC layer only, so consensus-critical
 /// reads (`epoch_state_from_canonical_tip`) are never throttled by RPC traffic.
-const MAX_CONCURRENT_REGISTRY_READS: usize = 64;
+const MAX_CONCURRENT_BLOCKING_RPC_WORK: usize = 64;
+
+/// Run `work` on `spawner`'s blocking pool under a `guard` permit, and await its value.
+///
+/// This is how every synchronous endpoint in this namespace runs its work. Handler bodies are
+/// polled on the shared runtime's worker threads, the same threads consensus tasks are scheduled
+/// on, so synchronous work that does not yield holds a worker for its entire duration and nothing
+/// else can be scheduled there until it returns. Dispatching it here keeps the worker free and puts
+/// the work on a pool whose occupancy `guard` bounds.
+///
+/// The permit is moved into the blocking task rather than held across the await, so it is released
+/// when the work actually finishes rather than when the caller stops waiting. A client that
+/// disconnects mid-call therefore cannot release capacity that is still in use.
+///
+/// At capacity this queues rather than rejecting, so a burst of legitimate callers is served late
+/// rather than failed.
+///
+/// `name` labels the task for the task manager's bookkeeping.
+async fn spawn_bounded_blocking<R, F>(
+    guard: &Arc<Semaphore>,
+    spawner: &TaskSpawner,
+    name: &'static str,
+    work: F,
+) -> TelcoinNetworkRpcResult<R>
+where
+    R: Send + 'static,
+    F: FnOnce() -> R + Send + 'static,
+{
+    // bound concurrent blocking work; queues (does not reject) at capacity, mirroring
+    // reth's eth_call guard. acquire only fails if the semaphore is closed, which never
+    // happens because it lives as long as the RPC server.
+    let permit = guard.clone().acquire_owned().await.map_err(|e| {
+        tracing::warn!(target: "tn::rpc", error = %e, task = name, "rpc blocking semaphore closed");
+        TNRpcError::Internal
+    })?;
+
+    let (tx, rx) = oneshot::channel();
+    spawner.spawn_blocking_task(name, move || {
+        // hold the permit across the work itself and release it before handing the result back, so
+        // capacity frees the moment the expensive part is done rather than after the handoff. The
+        // ordering also means a caller woken by the send always observes the permit as returned.
+        let value = {
+            let _permit = permit;
+            work()
+        };
+        if tx.send(value).is_err() {
+            tracing::debug!(target: "tn::rpc", task = name, "blocking rpc receiver dropped before result");
+        }
+        Ok(())
+    });
+
+    rx.await.map_err(|e| {
+        tracing::warn!(target: "tn::rpc", error = %e, task = name, "blocking rpc result channel closed");
+        TNRpcError::Internal
+    })
+}
+
+/// Build the proof-of-possession message `validator_address` signs with `bls_pubkey`.
+///
+/// Screening the encoding length happens here, on the async thread, because it is the one check
+/// that costs nothing and it settles every input that could never parse before a permit or a pool
+/// thread is committed to it. A flood of malformed keys therefore cannot occupy capacity that real
+/// work needs. Whatever survives the screen still has to decompress a BLS12-381 G2 point, which
+/// takes tens of microseconds and must not run where consensus tasks are scheduled, so it goes to
+/// the blocking pool under a permit.
+async fn proof_of_possession_message_bounded(
+    guard: &Arc<Semaphore>,
+    spawner: &TaskSpawner,
+    bls_pubkey: Bytes,
+    validator_address: Address,
+) -> TelcoinNetworkRpcResult<Bytes> {
+    BlsPublicKey::is_plausible_encoding(&bls_pubkey).then_some(()).ok_or_else(|| {
+        TNRpcError::InvalidParams(format!(
+            "invalid BLS pubkey: expected {} bytes (compressed) or {} bytes (uncompressed), got {}",
+            BlsPublicKey::COMPRESSED_BYTES,
+            BlsPublicKey::UNCOMPRESSED_BYTES,
+            bls_pubkey.len(),
+        ))
+    })?;
+
+    spawn_bounded_blocking(guard, spawner, "tn-rpc-proof-of-possession", move || {
+        // As of the compressed-bytes pivot the registry's `proofOfPossessionMessage` is an internal
+        // helper (not externally callable), so build the message natively from the same `tn_types`
+        // routine the validator signs and the registry reproduces on-chain byte-for-byte:
+        // `intent || compressed pubkey || address`.
+        BlsPublicKey::from_literal_bytes(&bls_pubkey)
+            .map(|pubkey| construct_proof_of_possession_message(&pubkey, &validator_address).into())
+            .map_err(|e| {
+                TNRpcError::InvalidParams(format!("invalid 96-byte compressed BLS pubkey: {e:?}"))
+            })
+    })
+    .await?
+}
 
 /// The type that implements `tn` namespace trait.
 #[derive(Debug)]
@@ -190,8 +288,8 @@ pub struct TelcoinNetworkRpcExt<N: EngineToPrimary> {
     ///
     /// The interface that handles primary <-> engine network communication.
     inner_node_network: N,
-    /// Bounds concurrent blocking [`ConsensusRegistry`] reads (see
-    /// [`MAX_CONCURRENT_REGISTRY_READS`]). Acquired before spawning the blocking read and held
+    /// Bounds the blocking work this namespace dispatches (see
+    /// [`MAX_CONCURRENT_BLOCKING_RPC_WORK`]). Acquired before spawning the blocking task and held
     /// until it completes, capping blocking-pool occupancy from RPC load.
     blocking_io_guard: Arc<Semaphore>,
 }
@@ -387,14 +485,13 @@ where
         bls_pubkey: Bytes,
         validator_address: Address,
     ) -> TelcoinNetworkRpcResult<Bytes> {
-        // As of the compressed-bytes pivot the registry's `proofOfPossessionMessage` is an internal
-        // helper (not externally callable), so build the message natively from the same `tn_types`
-        // routine the validator signs and the registry reproduces on-chain byte-for-byte:
-        // `intent || compressed pubkey || address`.
-        let pubkey = BlsPublicKey::from_literal_bytes(&bls_pubkey).map_err(|e| {
-            TNRpcError::InvalidParams(format!("invalid 96-byte compressed BLS pubkey: {e:?}"))
-        })?;
-        Ok(construct_proof_of_possession_message(&pubkey, &validator_address).into())
+        proof_of_possession_message_bounded(
+            &self.blocking_io_guard,
+            self.evm_state.get_task_spawner(),
+            bls_pubkey,
+            validator_address,
+        )
+        .await
     }
 
     async fn get_next_committee_size(&self) -> TelcoinNetworkRpcResult<u16> {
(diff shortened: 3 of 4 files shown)


sstanfield (approved):

grantkee (approved):
Good catch

PR #1059, opened 2026-07-27
exex: fuse ReplayStream on the first replay error

### LINKED ISSUE #984: exex: ReplayStream treats a replay error as a per-item event, not a stream terminator, so a log-and-continue consumer silently skips the failed block and every gap the delivery contract promises to surface
**Class:** latent correctness/robustness gap in the ExEx historical-replay path.  Confirmed mechanism, no in-tree consumer triggers it today (every shipped consumer aborts on the first `Err`), so nothing is broken on a running node.  The value is as a guard-rail to land *before* a consumer that logs-and-continues exists, so the "abort on error" contract is enforced by the stream itself rather than by consumer discipline alone.  Companion to #981 (same `crates/exex` replay surface, distinct mechanism).

## Summary

`ReplayStream` is the `futures::Stream` that replays finalized blocks from the DB during ExEx catch-up (`crates/exex/src/replay.rs:39-61`).  Its item type is `eyre::Result<TnExExNotification>`, and on any non-happy outcome it yields `Poll::Ready(Some(Err(..)))` and then **keeps going**: the next poll re-enters at the following block number and resumes yielding.  An error is modelled as a per-item event, not as a terminator.

`poll_next` advances the cursor *before* the match and sets no terminal state on the error arms (`replay.rs:49-59`):

```rust
let block_number = this.current_block;
this.current_block += 1;

match this.reth_env.replay_block_as_chain(block_number) {
    Ok(Some(chain)) => {
        Poll::Ready(Some(Ok(TnExExNotification::ChainExecuted { new: chain })))
    }
    Ok(None) => Poll::Ready(Some(Err(eyre::eyre!(
        "block {block_number} not found in database during replay"
    )))),
    Err(e) => Poll::Ready(Some(Err(e.into()))),
}
```

The only stop condition is the range guard at the top (`replay.rs:45-47`): `if this.current_block > this.end_block { return Poll::Ready(None) }`.  Because the cursor was already bumped past `block_number`, a poll following an `Err` computes `block_number + 1` and produces the next item.  There is no fused/terminal flag, so the stream happily emits blocks *after* an error.

The delivery contract documented for the replay path promises `ChainExecuted` "at-least-once and in **monotonic height order**" with gaps surfaced explicitly, not silently (`crates/exex/src/context.rs:159-167`).  A consumer that logs-and-continues past an `Err` receives `block_number + 1` next with **no gap marker**, so the block that errored is silently dropped from the delivered sequence.  That is a silent hole: the sequence stays monotonic and looks complete, but a block is missing, which is exactly the failure the contract says will not happen silently.

Every consumer that reads the `Result`-bearing stream aborts on the first `Err`, so none of them log-and-continue:

- The documented pattern uses `let notification = result?;` inside the loop (`context.rs:152-156`).
- The shipped example consumer uses `if let TnExExNotification::ChainExecuted { new } = result?` (`examples/exex-indexer/src/lib.rs:189-191`).
- `replay_and_subscribe`'s combined stream has **no in-tree consumer at all** (only doc references), so there is no second live-path consumer to worry about today.
- `dedup_chain_executed` passes error items straight through without masking them (`context.rs:221-236`: a non-`ChainExecuted` item has no height, so it is kept), but it also does not terminate the stream, so it does not help here either.

So the hole is unreachable in-tree right now, held shut purely by the convention that every consumer writes `result?`.  It bites the day someone writes the natural alternative: a consumer that treats the stream's per-item `Result` as recoverable and does "log the bad block, keep indexing".  That is a reasonable thing to write against a `Stream<Item = Result<..>>`, and it is *invited* by the one documented error, `"block {n} not found in database during replay"`, which reads like a block-specific, skippable condition rather than a fatal one.  Landing the fuse now removes the trap before that consumer exists, and turns "abort on error" from an unwritten rule into a property of the type.

## The defect

**Location:** `crates/exex/src/replay.rs:42-61` (`ReplayStream::poll_next`), specifically the cursor advance at `:49-50` and the two error arms at `:56-59`.

**Reachable via both public entry points**, since both funnel through the same `poll_next`:

- `TnExExContext::replay_from` returns the raw `ReplayStream` (`context.rs:143-146`) — this is the path the shipped example actually uses (`exex-indexer/src/lib.rs:189`).
- `TnExExContext::replay_and_subscribe` wraps it in `dedup_chain_executed(replay.chain(live))` (`context.rs:174-200`).

**The two ways an `Err` arises mid-range:**

1. `Ok(None)` — a block in `[start_block, end_block]` is absent from the DB (`replay.rs:56-58`).
2. `Err(e)` — `replay_block_as_chain` itself fails (`replay.rs:59`).

In both cases the cursor has already been advanced (`replay.rs:50`), no terminal state is set, and the next poll resumes at `block_number + 1`.

`size_hint` compounds the wrong mental model: after an error it still reports `end_block - current_block + 1` remaining (`replay.rs:63-70`), advertising blocks that, in a log-and-continue consumer, will be delivered on top of a hole.

## Trigger conditions

1. A consumer of `ReplayStream` (via `replay_from` or `replay_and_subscribe`) that does **not** abort on the first `Err` — it logs and keeps polling.
2. A replay that actually produces an `Err` mid-range (either `Ok(None)` for a missing block, or a `replay_block_as_chain` failure).

Given both, the consumer silently skips the failed block: the next delivered `ChainExecuted` is `block_number + 1` with no `Lagged`/gap marker.  Note that the `Lagged` gap-signal mechanism lives on the **live** channel (`context.rs:165-166`), not on the replay range, so nothing in the replay path marks the skipped block.  A stateful ExEx (indexer, MEV/order-flow accounting, anything that folds over every block) then commits a state that is missing a block, monotonic and gap-free on its face.

## Root cause

A `futures::Stream` does not, by contract, stop after yielding `Err`; "an error ends the replay" is a semantic that lives entirely in consumer discipline (`result?`).  `ReplayStream` never encodes that invariant in the stream itself: `poll_next` advances `current_block` before the match and returns the error arms without setting any terminal state, so the only stop condition (`current_block > end_block`) is not reached until the whole range is walked.  The stream therefore treats "this block failed" and "keep producing subsequent blocks" as compatible, which they are not for a replay whose entire purpose is a gap-free monotonic sequence.

## Suggested fix

Fuse the stream on the first non-`Ok(Some(..))` outcome, so the "abort on error" contract is enforced by the stream, not just by every consumer.  Advance the cursor past `end_block` on the terminal arms so the next poll hits the range guard and yields `None`:

```rust
match this.reth_env.replay_block_as_chain(block_number) {
    Ok(Some(chain)) => {
        Poll::Ready(Some(Ok(TnExExNotification::ChainExecuted { new: chain })))
    }
    Ok(None) => {
        this.current_block = this.end_block.saturating_add(1); // fuse: next poll -> None
        Poll::Ready(Some(Err(eyre::eyre!(
            "block {block_number} not found in database during replay"
        ))))
    }
    Err(e) => {
        this.current_block = this.end_block.saturating_add(1); // fuse
        Poll::Ready(Some(Err(e.into())))
    }
}
```

After this, a log-and-continue consumer sees the `Err` and then end-of-stream, never a silently-skipped block, and `size_hint` naturally reports `0` remaining once fused (`replay.rs:63-70`).  A consumer that still writes `result?` is unaffected.

Two notes on alternatives:

- **`futures::StreamExt::fuse()` does not fix this.**  `Fuse` only guarantees that once a stream returns `Poll::Ready(None)` it keeps returning `None`;  it does nothing on `Some(Err(..))`.  Because `ReplayStream` keeps returning `Some(..)` after an error and never reaches `None` on its own, wrapping it in `.fuse()` changes nothing.  The terminal state has to be set inside `poll_next`.
- **Cursor-overflow shape (related, separately filed as Finding 4 in the same review):** the unchecked `this.current_block += 1` at `replay.rs:50` is the sibling concern.  It is far more remote (it requires `current_block == u64::MAX`, unreachable for real block heights), but if this fix is taken, moving the happy-path advance into the `Ok(Some)` arm as `saturating_add(1)` closes both shapes in one place and leaves no naked `+= 1` on any path.

## Suggested test design

- **Fuse-on-`Ok(None)`:** drive a `ReplayStream` over a range in which `replay_block_as_chain` returns `Ok(None)` for a mid-range block, poll past that point, and assert the sequence is `Ok(_)*`, then exactly one `Err`, then `Poll::Ready(None)` — never another item after the `Err`.
- **Fuse-on-`Err`:** same, with `replay_block_as_chain` returning `Err` for the mid-range block;  assert identical fused shape.
- **Happy path unchanged:** a full clean range still yields every block in order and then `None`.
- **Confirm-by-mutation:** removing the `this.current_block = this.end_block.saturating_add(1)` assignment on the terminal arms must make the fuse tests *fail* (the stream yields a further item after the `Err`), so the tests genuinely pin the terminal semantics rather than passing vacuously.

If `RethEnv::replay_block_as_chain` is awkward to drive to `Ok(None)`/`Err` in a unit test, the same property can be pinned one level down by extracting the "block-number to `Result<Option<Chain>>`" step behind a small trait/closure the test can control, without changing the public API.

## Acceptance criteria

- After `ReplayStream` yields its first non-`Ok(Some(..))` item (any `Err`), the very next poll returns `Poll::Ready(None)` and the stream stays terminated.
- A consumer that logs-and-continues past an `Err` observes end-of-stream, not a silently-skipped block, so the delivered `ChainExecuted` sequence can never contain a silent hole regardless of consumer discipline.
- A test pins the fused terminal semantics for both `Ok(None)` and `Err`, and is shown to fail under mutation of the fuse assignment.

MavenRain:
Closes #984.

## Problem

`ReplayStream::poll_next` advanced its cursor before matching on `replay_block_as_chain`, and the two error arms set no terminal state.  An error was modelled as a per-item event: the poll after an `Err` resumed at the next block number and kept yielding.  Every in-tree consumer writes `let notification = result?;`, so nothing is broken on a running node, but the "abort on error" contract was enforced only by consumer discipline.  A consumer that treats the per-item `Result` as recoverable (log the bad block, keep indexing, the natural reading of a `Stream<Item = Result<..>>`, and one invited by the block-specific "not found in database" error text) would silently skip the failed block: the delivered `ChainExecuted` sequence stays monotonic and looks complete, which is exactly the silent hole the delivery contract in `context.rs` promises will not happen.  No attacker is required;  a single missing or unreadable block in the replay range is enough.

## Fix

- [x] `crates/exex/src/replay.rs`: the cursor is now a sum type (`ReplayCursor::Active(next)` / `ReplayCursor::Done`), and the stream is a `ReplayCore<S>` state machine generic over a private `ReplaySource` trait implemented by `RethEnv` (static dispatch;  `ReplayStream`'s public API is unchanged).  Both error arms (`Ok(None)` missing block, `Err` failed read) transition to `Done`, so the first error is yielded as the stream's final item and every later poll returns `None`.  `futures::StreamExt::fuse()` could not provide this: `Fuse` only latches after `Ready(None)`, and the old stream never reached `None` on its own after an error.
- [x] The failed-read error is wrapped with the failing height (`replay failed at block N`), symmetric with the missing-block message.  Because the error is now the stream's final item, the height doubles as the resume point for a consumer that re-replays the fused gap.
- [x] The happy-path advance closes the range exactly at `end_block` (`Done` on the last block, `saturating_add(1)` otherwise), so the sibling cursor-overflow shape (the old unchecked `current_block += 1`) is gone as well;  no naked increment remains on any path, and the cursor arithmetic is total even at `BlockNumber::MAX`.
- [x] `size_hint` reports `(0, Some(0))` once fused, and a lower bound of `1` while active: with a fuse in the semantics, only the next item is guaranteed, so the old exact lower bound would overpromise.
- [x] `crates/exex/src/context.rs`: the `replay_from` and `replay_and_subscribe` docs now state the fused contract.  On the combined path the first replay `Err` ends the historical phase and later items come from the live channel at tip heights, so the docs state explicitly that the `Err` is the signal that the rest of the historical range was not delivered, that a consumer MUST abort on it, and that this path (which consumed the context) cannot re-`replay_from` the fused gap;  an ExEx that must resume after a replay failure should drive `replay_from` and the live channel directly.

## Testing

- `cargo +1.94 test -p tn-exex -j 2`: 20 passed, 0 failed.  Five new deterministic unit tests drive `ReplayCore` through a test `ReplaySource` (no database): fuse on missing block, fuse on failed read (message carries the height), clean range yields every block then `None`, empty range immediately terminated, range ending at `u64::MAX` terminates cleanly.  The failing test source succeeds again after the failing block precisely so an unfused stream would visibly keep yielding `Ok` items past the `Err`.
- Confirmed by mutation: reverting the two `Done` transitions to the old cursor advance makes `missing_block_fuses_the_stream` and `read_error_fuses_the_stream` fail (an item is yielded after the `Err`);  restoring the fuse returns them to green.
- `cargo +nightly-2026-03-20 fmt -- --check` green.
- `cargo +nightly-2026-03-20 clippy --workspace -- -D warnings` green.
- `cargo +1.94 test --no-run --workspace -j 2` green (shared target;  full `make attest` re-runs the workspace suite on the attested commit).

### DIFF
--- a/crates/exex/src/context.rs
+++ b/crates/exex/src/context.rs
@@ -143,6 +143,11 @@ impl TnExExContext {
     /// # Errors
     ///
     /// Returns an error if the current chain tip cannot be read from the DB.
+    ///
+    /// The returned stream is **fused on failure**: a block missing from the
+    /// database or a failed read is yielded as a single `Err` item, after
+    /// which the stream terminates; an error is never followed by a
+    /// silently-skipped block. See [`ReplayStream`].
     pub fn replay_from(&self, start_block: BlockNumber) -> eyre::Result<ReplayStream> {
         let tip = self.reth_env.last_block_number()?;
         Ok(ReplayStream::new(self.reth_env.clone(), start_block, tip))
@@ -168,8 +173,14 @@ impl TnExExContext {
     /// channel while replay runs) is de-duplicated by height so a block is never
     /// processed twice. If the live channel overflowed during a slow replay, the
     /// resulting gap surfaces as a [`TnExExNotification::Lagged`] marker rather
-    /// than silently — re-`replay_from` the gap to reconcile. See
-    /// [`replay_from`](Self::replay_from) for replay state-diff fidelity.
+    /// than silently; re-`replay_from` the gap to reconcile. A replay error is
+    /// likewise terminal for the historical phase: the first `Err` fuses the
+    /// replay stream (see [`ReplayStream`]), so no further historical block is
+    /// delivered after it and any later items come from the live channel at
+    /// tip heights. The `Err` is the explicit signal that the rest of the
+    /// historical range was not delivered; a consumer MUST abort on it (the
+    /// documented `result?` pattern) rather than continue into the live tail.
+    /// See [`replay_from`](Self::replay_from) for replay state-diff fidelity.
     ///
     /// # Progress reporting
     ///
@@ -187,9 +198,10 @@ impl TnExExContext {
     /// [`TnExExEvent::FinishedHeight`].
     ///
     /// This path returns one combined stream, so it cannot itself re-`replay_from`
-    /// to reconcile a [`Lagged`](TnExExNotification::Lagged) gap; an ExEx that must
-    /// reconcile lag should drive [`replay_from`](Self::replay_from) and the live
-    /// channel directly instead.
+    /// to reconcile a [`Lagged`](TnExExNotification::Lagged) gap or a fused replay
+    /// error; an ExEx that must reconcile lag or resume after a replay failure
+    /// should drive [`replay_from`](Self::replay_from) and the live channel
+    /// directly instead.
     pub fn replay_and_subscribe(
         self,
         start_block: BlockNumber,
--- a/crates/exex/src/replay.rs
+++ b/crates/exex/src/replay.rs
@@ -3,69 +3,297 @@
 //! When an ExEx starts (or restarts after a crash), it may need to catch up
 //! to the current chain tip. Since TN has no reorgs, this is simply reading
 //! finalized blocks from the database in ascending order.
+//!
+//! Replay errors are terminal: the stream fuses on the first error, so an
+//! `Err` item is always the last item; a consumer that keeps polling sees
+//! end-of-stream, never a silently-skipped block.
 
-use crate::TnExExNotification;
+use crate::{Chain, TnExExNotification};
+use futures::Stream;
 use std::{
     pin::Pin,
+    sync::Arc,
     task::{Context, Poll},
 };
 use tn_reth::RethEnv;
 use tn_types::BlockNumber;
 
-/// Stream that replays historical blocks from the database as ExEx notifications.
+/// Source a replay reads from: maps a block number to the chain segment
+/// stored for it.
 ///
-/// Reads finalized blocks from `current_block` to `end_block` (inclusive) and yields
-/// `TnExExNotification::ChainExecuted` for each block.
-///
-/// If `current_block > end_block`, the stream is immediately empty.
-///
-/// Replayed `ChainExecuted` notifications carry an **empty `BundleState`** (state
-/// diffs are read from the DB by block number, not re-derived here) — see
-/// [`RethEnv::replay_block_as_chain`](tn_reth::RethEnv::replay_block_as_chain).
-#[derive(Debug)]
-pub struct ReplayStream {
-    reth_env: RethEnv,
-    current_block: BlockNumber,
-    end_block: BlockNumber,
+/// The one production implementation is [`RethEnv`]; tests substitute a
+/// source that fails on demand to pin the stream's fused error semantics.
+trait ReplaySource {
+    /// Read block `block_number` back as a single-block [`Chain`].
+    ///
+    /// Returns `Ok(None)` if the block is not in the database.
+    fn replay(&self, block_number: BlockNumber) -> eyre::Result<Option<Arc<Chain>>>;
 }
 
-impl ReplayStream {
-    /// Create a new replay stream from `start_block` to `end_block` (inclusive).
-    pub fn new(reth_env: RethEnv, start_block: BlockNumber, end_block: BlockNumber) -> Self {
-        Self { reth_env, current_block: start_block, end_block }
+impl ReplaySource for RethEnv {
+    fn replay(&self, block_number: BlockNumber) -> eyre::Result<Option<Arc<Chain>>> {
+        self.replay_block_as_chain(block_number).map_err(Into::into)
     }
 }
 
-impl futures::Stream for ReplayStream {
+/// Cursor for a replay range: the next block to replay, or the terminal state.
+///
+/// `Done` is entered when the range is exhausted **or on the first error**, so
+/// an error is a stream terminator, not a per-item event. Once an `Err` has
+/// been yielded every later poll returns `None`, making "abort on error" a
+/// property of the stream itself rather than of consumer discipline.
+#[derive(Clone, Copy, Debug)]
+enum ReplayCursor {
+    /// The next block number to replay.
+    Active(BlockNumber),
+    /// Terminal: the range is exhausted or an error was yielded; every
+    /// subsequent poll returns `None`.
+    Done,
+}
+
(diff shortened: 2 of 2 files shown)


sstanfield (approved):

grantkee (approved):
LGTM!

PR #1060, opened 2026-07-27
fix(tn-reth): enforce the archive-mode assumption behind pinned registry reads

### LINKED ISSUE #1028: Snapshot:nothing enforces the archive-mode assumption that every pinned registry read depends on
## Problem

Every pinned registry read — the batched registry read at a header (`read_consensus_registry_batch_at_header`), the epoch-state read at a header (`epoch_state_at_header`), the current-epoch-info read at a header (`get_current_epoch_info_at_header`), and the general pinned contract read (`read_contract_inner`) — resolves historical state through `state_by_block_hash`, and the correctness of that resolution rests entirely on a comment: the ARCHIVE-MODE ASSUMPTION note in the reth wrapper (crates/tn-reth/src/lib.rs:1966-1972) records that the node runs with pruning disabled, so historical state for any past block is always materialized.

If pruning were ever enabled, reth's `HistoricalStateProvider` can classify a queried block as `MaybeInPlainState` and silently serve state from the canonical tip instead of the pinned block. A "pinned" read then returns tip state with no error and no log — silently unpinning exactly the reads whose entire correctness argument is the pin: committee membership, epoch records, and epoch state. Under a mid-epoch governance burn, two nodes issuing the same pinned read would decode different committees, which is consensus divergence with no diagnostic trail.

Nothing enforces the assumption at runtime: there is no startup assertion, no feature gate, and no CI check that pruning stays disabled for the providers backing pinned reads. The assumption is also documented unevenly — `epoch_state_at_header`, the read that feeds the epoch-entry state, does not carry the comment at all, so a future maintainer touching provider configuration has no local signal that these reads depend on it.

## Solution

Introduce a startup assertion that pruning is disabled for the providers backing pinned reads, so a node misconfigured with pruning refuses to start instead of silently serving unpinned state. The check should live where the node constructs its reth environment, and its failure message should name the pinned-read contract it protects.

Carry the ARCHIVE-MODE note onto `epoch_state_at_header` and the other pinned-read constructors, so every function whose correctness depends on the assumption states it locally rather than by proximity to a sibling.

Consider a CI grep-guard that fails if `PrunerBuilder` (or an equivalent pruning entry point) is introduced into the node's provider construction, turning a silent semantic change into a visible review decision.

@sstanfield - we need to ensure snapshot support for validators covers at least the previous epoch boundary for startup consensus info reads

MavenRain:
Closes #1028.

## Problem

Every pinned registry read resolves historical state through `state_by_block_hash`: the batched registry read at a header (`read_consensus_registry_batch_at_header`), the epoch-state read (`epoch_state_at_header`), the current-epoch-info read (`get_current_epoch_info_at_header`), the general pinned contract read (`read_contract_inner`), the worker fee-config read (`worker_fee_configs_inner`), and the epoch-record read (`get_epoch_info_at_block`).

The correctness of that resolution rested entirely on a comment.  `crates/tn-reth/src/env/epoch.rs` carries an ARCHIVE-MODE ASSUMPTION note recording that the node runs with pruning disabled, so historical state for any past block is always materialized.  Nothing enforced it.  There was no startup assertion, no feature gate, and no CI check that pruning stays disabled for the providers backing pinned reads.

**One correction to the issue as filed.**  The issue (and the original comment) describes the hazard as reth silently serving canonical-tip state for a pinned block.  That is not what reth v1.11.3 does on the pruning path, and the difference matters for what this PR can honestly claim.  `HistoricalStateProviderRef` compares the pinned block against the account- and storage-history watermarks it derives from the `PruneCheckpoints` table and returns `ProviderError::StateAtBlockPruned` (`crates/storage/provider/src/providers/state/historical.rs:135-136, 158-159, 253-256`).  Pruning therefore breaks pinned reads *loudly*.

That is still a real problem, just a different one.  The failure lands in the middle of a consensus read (committee construction, epoch entry, base-fee seeding), where the reachable outcomes are a stalled epoch transition or a halted node, and it lands on whichever node happens to be pruning rather than at a moment an operator can act on.  Refusing to start is the same fact, delivered when it is still actionable.

The genuinely silent variant does exist, and it is narrower than the issue states: those watermarks come from the prune checkpoints, so history missing *without* one (an externally truncated datadir, or a snapshot-restored node with no pre-snapshot history) leaves them unset, `HistoryInfo::NotYetWritten` classifies the lookup, and the read returns `Ok(None)`: every account and storage slot reads as never written, with no error at all.  That is a property of the datadir, not of the configuration, so a startup config check cannot see it.  Since this PR was first reviewed, #1131 closed the snapshot-restore half of that gap with a TN-owned floor; see the maintainer decisions and scope notes.

## Fix

Three layers, so the assumption is enforced at each level where it can break rather than assumed at all of them.

**Startup assertion.**  `RethConfig::ensure_archive_mode` fails when `NodeConfig::prune_config()` returns `Some`, and `RethEnv::new` calls it before any database work happens.  `RethEnv::new` is the single chokepoint: every `RethEnv` in the process (production node, genesis ceremony temp env, snapshot restorer, every test env) funnels through it, and `BlockchainProvider` is constructed nowhere else.

The check is deliberately conservative.  `prune_config()` returns `None` exactly when no pruning was requested, so any `Some` fails here: a per-segment mode, `--full`, `--minimal`, or a non-default block interval.  A configuration that happens to leave account and storage history intact still fails, because deciding which segments a pinned read can survive is a review decision, not a startup one.

**Local documentation.**  The existing long-form note on `pinned_state_and_env` is rewritten: it justified the assumption on the grounds that "no `PrunerBuilder` exists in the repo", which was true but unenforced, and it described the silent-tip-state mechanism corrected above.  The rewrite separates the three ways "history is missing" can present, only one of which is loud.

Five of the six pinned reads funnel through `pinned_state_and_env` and inherit that note.  `read_contract_inner` does not: it builds its state from `read_only_state_db` directly, so it carries its own note pointing at the normative one.  The three places that name a chokepoint now name `read_only_state_db`, which is the helper all six actually share.

**Source guard.**  `etc/archive-mode-guard.sh` fails the build if a pruner entry point (`PrunerBuilder`, `Pruner::new`, `with_prune_modes`, `set_prune_modes`, `PruneModes::`) or a reth.toml loader (`reth_config::Config`) enters `crates/`, `bin/`, or `examples/`.  This covers the two things a configuration check cannot see: a pruner driven straight against the provider, bypassing `NodeConfig` entirely, and a reth.toml `[prune]` section, which reth's own launcher merges *behind* `NodeConfig::prune_config` (TN hand-rolls its init path and never loads one today).  The guard resolves nothing and compiles nothing.

It is genuinely fail-closed.  `grep` exits 0 on a match, 1 on no match, and 2 or higher on an error, and an `if` condition would fold that error into the same branch as "no match" (bash suspends `errexit` for `if` conditions), so an unreadable tree or a missing `grep` would report success.  The status is captured explicitly and anything above 1 fails the run.

The guard is dual-wired on purpose.  `.github/workflows/pr.yaml` skips fmt / clippy / test / rocksdb-gate for maintainers, so a CI-only job would never run on a maintainer's own PR, which is the population most likely to be changing provider configuration.  It therefore runs from `etc/test-and-attest.sh` as well, ahead of the compile gates.

## Maintainer decisions (2026-08-10)

The three questions that sent this PR back to draft were answered by @sstanfield (https://github.com/Telcoin-Association/telcoin-network/pull/1060#issuecomment-5246347812):

1. **The startup assertion stays.**  It cannot fire on today's CLI (`PruningArgs` is a hardcoded all-off literal), and that is recorded in its doc as deliberate: it exists for the reth-upgrade case, where a default change or a newly flattened arg would make pruning reachable without any change to this repo.  Steve: "While not strictly needed the runtime check seems harmless and won't slow us down (one time check at startup) so seems fine to keep to me."
2. **Snapshot-restored validators are supported in principle.**  Steve: "technically I think the answer is yes we should support snapshotted validators but we may not want to for other reasons."  That fixes the severity of the restored-state floor #1131 landed: a pin below the floor is a hard `StateReadError::Provider` refusal, and the epoch-close path halts on it rather than continuing on silently-wrong state.  No softening.
3. **A non-genesis start always comes from a snapshot** of an epoch's closing state.  Every supported non-genesis datadir therefore carries the floor marker the restorer persists, so the partial-sync/truncation composition case has an owner: the floor covers it.  A datadir truncated by anything other than the restorer is manual surgery and stays unsupported.

## Changes

- [x] `crates/tn-reth/src/cli.rs`: add `RethConfig::ensure_archive_mode` (`pub(crate)`, since the call site now lives in another module)
- [x] `crates/tn-reth/src/env/mod.rs`: call it as the first statement of `RethEnv::new`
- [x] `crates/tn-reth/src/env/epoch.rs`: corrected rewrite of the normative ARCHIVE-MODE note on `pinned_state_and_env`, plus the module-header summary, both reconciled with the RESTORED-DATADIR FLOOR paragraph #1131 added to the same doc block
- [x] `crates/tn-reth/src/env/helpers.rs`: ARCHIVE-MODE note on `read_contract_inner`, the one pinned read that does not route through `pinned_state_and_env`
- [x] `etc/archive-mode-guard.sh`: new fail-closed source guard over `crates/`, `bin/`, and `examples/`
- [x] `etc/test-and-attest.sh`: run the guard before fmt
- [x] `.github/workflows/pr.yaml`: `archive-mode-gate` job modeled on `rocksdb-gate`, wired into the `ci-success` needs list and its failure check (now alongside main's `adiri-test`)

## Testing

Six new tests.  Five predicate tests live in `crates/tn-reth/src/cli.rs` next to `ensure_archive_mode`, and the wiring test lives in `crates/tn-reth/src/env/genesis.rs`, where the crate's `RethEnv`-construction tests are.  Both modules are plain `#[cfg(test)]`, so these run in the default-feature workspace pass rather than only the adiri-gated one.

- `test_archive_mode_accepts_production_config` . . . the config `RethConfig::new` builds passes
- `test_archive_mode_accepts_default_node_config` . . . the `NodeConfig::default()` path `new_for_temp_chain` uses passes (every test env in the workspace is built through it)
- `test_archive_mode_rejects_account_history_pruning` . . . the knob that directly gates `state_by_block_hash`
- `test_archive_mode_rejects_storage_history_pruning` . . . storage history carries the `ConsensusRegistry` slots the committee is decoded from
- `test_archive_mode_rejects_full_node_pruning` . . . `--full` derives a whole prune-mode set from the chain spec, exercising the composite path
- `test_reth_env_new_rejects_pruned_config` . . . proves the guard is wired into startup, not merely callable in isolation

```
cargo +1.94 nextest run -p tn-reth --lib
    Summary [22.113s] 133 tests run: 133 passed, 0 skipped
```

Each new test was confirmed by mutation:

- Neutering `ensure_archive_mode` so it always returns `Ok(())`: the four rejection and wiring tests FAIL, the two acceptance tests still pass.
- Removing the `ensure_archive_mode()?` call from `RethEnv::new` but leaving the function intact: only `test_reth_env_new_rejects_pruned_config` FAILS, confirming it tests the wiring rather than the function.

The shell guard was exercised on all three of its exit paths:  clean tree exits 0;  a file containing `PrunerBuilder::new()` under `bin/telcoin-network/src/` exits 1 naming the path (this case is what motivated widening the scan beyond `crates/`);  an unreadable directory under a scan root exits 1 with "could not scan", rather than passing.

## Scope notes

This branch is merged with `main` past #1069, which moved every symbol it touched out of `crates/tn-reth/src/lib.rs` into `cli.rs` and `env/*.rs` and consolidated the epoch reads behind `pinned_state_and_env`, and past #1131, which landed the restored-state floor.  The merge is why the paths above differ from the ones in the original commit.

The issue's closing note to @sstanfield (snapshot support covering at least the previous epoch boundary for startup consensus info reads) was split out as #1105, and #1131 has since landed it: `SnapshotRestorer` persists the restored window's lower bound, and `pinned_state_and_env` refuses any pin below it before touching the provider.  The merged tree therefore enforces both halves: this PR's configuration and source gates against pruning, and the floor against restored-datadir truncation.  One latent edge sits outside both and is filed separately rather than folded in here: `read_contract_at_block` builds its state from `read_only_state_db` directly rather than through `pinned_state_and_env`, so it is not floor-guarded; it has no production caller today (test helpers only).

Two residual limits, both stated in the guard's doc comment rather than left implicit:

- `ensure_archive_mode` reads the configuration *this* node was built with, so it cannot detect a datadir some other binary already pruned.  The source guard is what keeps a pruner out of this repo.
- `--prune.block-interval 5`, exactly reth's default, is indistinguishable from passing no flag at all.  This is harmless: the interval alone prunes no segment.


sstanfield (dismissed):
Approving but I think we need to dig into this deeper.  For instance once my partial sync lands will that be a problem?  It is not technically pruning and should have everything needed to start the next epoch so I don't think so.  But certainly history is missing.  I think we need a better solution to this, for instance if you try to use a pinned block that is not in your history then why not let the node die a noisy death?  The sync case should have all that is needed going forward (or it is broken) and a prune config that was so aggressive it removes blocks you still need is not very useful.  I may be missing something though.  But a quick Look at how it is used and I don't know why we would not be using it on very recent history.

MavenRain:
Thanks Steve, this was worth digging into and you were right to push. I went through reth v1.11.3's historical provider, the sync/snapshot paths, and every caller of the pinned reads. Short version:
you are right on three of your four points.  I also found that my own PR description had the mechanism backwards, so I've corrected it.

## "once my partial sync lands will that be a problem?"

No, and your reasoning holds.  Two independent reasons:

The sync path never touches pruning. The forward drain is contiguous (`crates/state-sync/src/lib.rs` iterates `last_consensus_height + 1..=max_consensus_height` with no skipping), and there is no prune reference anywhere in `crates/state-sync/`. You end up with the same contiguous history a header-by-header sync gives you, from wherever you started.

More to the point, and this is the part I did not expect: **`ensure_archive_mode` cannot reject any TN node today, from any path.**  It fires only when `NodeConfig::prune_config()` is `Some`, and that is purely a function of `PruningArgs`.  `RethCommand` flattens only rpc/txpool/db, and `PruningArgs` is a hardcoded literal with every field off, under the existing `// Ignore Reth pruning for now.` comment in `cli.rs`.  There is no reth.toml loader. So the check is currently unfalsifiable.  It waves through every node, partial-sync ones included.

I've rewritten the PR description to say that plainly rather than implying it guards a reachable
misconfiguration.  Its remaining value is narrow but real: it catches a reth-upgrade default change or a newly flattened CLI arg, which the source grep would not see.

## "why not let the node die a noisy death?"

Partly agree, and your instinct matches what already happens, but the useful version of this is
narrower than it looks.

The loud path already exists end to end.  reth returns `ProviderError::StateAtBlockPruned`, that
flows through `read_only_state_db` into `pinned_state_and_env`'s `map_err(StateReadError::Provider)`, and from there the call sites already implement a deliberate, documented spread from hard halt (`close_epoch.rs` halts on both error classes) through documented fail-open for `ChainGlobal` only (`run_epoch.rs`).  So no new death-plumbing and no new error variant is needed. `Provider` is already the right arm.

The gap is detection, not handling.  The two outcomes that actually lose us determinism raise no
error at all, so no call-site policy can see them, and this is where I had it backwards in the PR
description:

- **`MaybeInPlainState`** silently serves tip state for a pinned block. It is reached only when a
  prune checkpoint *exists* (`historical.rs`: the `else if lowest_available_block_number.is_some()`
  arm). So pruning is the only thing that substitutes tip state, and it does so *above* the
  watermark, not below it.
- **`NotYetWritten`** returns `Ok(None)`, so the account or slot reads as never written. This is
  the checkpoint-less case, and it is what a snapshot-scaffolded node hits below its window.

A read-site check against reth's prune checkpoints would not help with either: reth already
performs exactly that comparison and already errors, and `is_account_history_available` defaults to *available* when no checkpoint exists, so an app-level version reads the same absent row and is blind precisely in the snapshot case. The finer-grained check is also not reachable through the handle we hold, since `read_only_state_db` returns a `StateProviderBox` and `StateProvider` does not carry `account_history_lookup`.

Where your suggestion does hold genuinely new coverage is a **TN-owned** floor rather than a reth
one. We write the snapshot scaffold, so we know `window_start`;  persisting it and having
`pinned_state_and_env` reject `header.number < floor` with `StateReadError::Provider` is
checkpoint-independent, slots into the existing error arm, and is the only guard that covers the
truncation that exists on main today.  I've filed that as a follow-up rather than folding it in here,
since it needs a decision about whether a snapshot-restored node is a supported validator
configuration.

## "I don't know why we would not be using it on very recent history"

This is ight for five of the six pinned-read families.  `epoch_state_at_header`, `get_current_epoch_info_at_header` and the batched registry read are all bounded to the canonical or finalized tip, or exactly one epoch boundary back.

The exception is `derive_idle_worker_fee_at` (`crates/node/src/manager/node.rs`), which loops
`k = epoch, epoch-1, ...` doing a pinned state read at each prior epoch's closing block until it
finds a Static-strategy anchor, a genuine produced block, a freshly created slot, or epoch 0.  Its own doc says the walk "terminates in at most `epoch` steps".  It is not an exotic path either: it runs at every epoch entry via `derive_base_fees_for_entered_epoch`. So the worst case for a chronically idle non-Static worker is a read at epoch 0, the chain's full lifetime.

One reassuring bound: this is chain-driven, not attacker-driven. `tn_getEpochInfo` takes an `Epoch` and does a tip read against the registry ring buffer, so there is no RPC surface that lets a caller pin an arbitrary old block.

## "a prune config that was so aggressive it removes blocks you still need is not very useful"

I agree with the sentiment.  Where I'd push back is on which number is safe: because of the idle-worker walk above, "blocks you still need" is not a fixed depth, it is a function of worker idleness, and nothing in the code bounds it. We can't pick a retention floor today without first bounding that walk.  I've filed that as a follow-up too, and referenced it from the PR description as the precondition for ever relaxing this.

However, I want to be straight that this does not rescue the artifact you were reviewing.  Your criticism is still valid, just for a different reason than "too strict": the check is strict about something unreachable while the live truncation path (snapshot restore / `db load-state`) walks straight past it.  That is fair.

## What changed in this PR

- Rebased onto the #1069 reorg. `ensure_archive_mode` now lives in `cli.rs` next to the
  `PruningArgs` defaults it asserts;  the call site is `RethEnv::new` in `env/mod.rs`.
- The six per-read ARCHIVE-MODE notes collapsed to one, since #1069 already funnelled every pinned read through `pinned_state_and_env`. That note and the module summary are corrected: they claimed pruning causes a silent fallback to tip state, which is only half right and had the checkpoint condition inverted.
- Description rewritten to drop the overstated claims, including the vacuity above.

## Open questions I'd rather you called

1. Given the runtime check cannot fire today, do you want it at all, or is the source grep enough?  I've kept it for the reth-upgrade case, but that's a judgement call and yours to overrule.
2. Should a snapshot-restored node be a supported validator configuration? That decides whether the window-floor guard is a hard halt or something softer.
3. Is your partial sync's starting point ever non-genesis *without* a snapshot import?  If it can be, the two features compose into a truncation case neither one owns.

MavenRain:
Returning to draft mode until the open questions are answered by @sstanfield and/or @grantkee

sstanfield:
> ## Open questions I'd rather you called
> 1. Given the runtime check cannot fire today, do you want it at all, or is the source grep enough?  I've kept it for the reth-upgrade case, but that's a judgement call and yours to overrule.
> 2. Should a snapshot-restored node be a supported validator configuration? That decides whether the window-floor guard is a hard halt or something softer.
> 3. Is your partial sync's starting point ever non-genesis _without_ a snapshot import?  If it can be, the two features compose into a truncation case neither one owns.

My thoughts on the questions:
1. While not strictly needed the runtime check seems harmless and won't slow us down (one time check at startup) so seems fine to keep to me.
2. I think so but this gets into non-technical issues like making sure we have archive nodes.  So technically I think the answer is yes we should support snapshotted validators but we may not want to for other reasons.
3. Not sure I follow this one.  If we are starting from non-genesis we by definition have to have a snapshot (and it needs to be the closing state of an epoch) or we can't start.  Am I missing something?

MavenRain:
Thanks Steve, all three calls are what I needed. Taking them in order:

**1. Runtime check stays.** Agreed on the cost: it is one `prune_config()` inspection at startup, before any database work. I've recorded the decision in the description so the next person who notices it cannot fire today knows that is by design, not an oversight: it is there for the reth-upgrade case, where a default change or a newly flattened CLI arg would make it reachable without anyone touching this repo's code.

**2. Snapshotted validators supported, hard halt confirmed.** That answer settles the severity question, and the code already agrees with you: #1131 landed the restored-state floor with exactly the hard disposition. A pin below the floor surfaces as `StateReadError::Provider`, and the epoch-close path halts on it rather than continuing on silently-wrong state. Your operational caveat (needing archive nodes to serve the snapshots) is real but lives outside this PR; nothing in the code forecloses it either way.

**3. You are not missing anything; that is the answer I was hoping for.** If a non-genesis start always comes from a snapshot of an epoch's closing state, then every supported non-genesis datadir carries the floor marker the restorer persists, and the case I was worried about (a partial sync starting from history the floor knows nothing about) cannot arise in a supported configuration. The only remaining silent case is a datadir truncated by something other than the restorer, i.e. manual surgery, and treating that as unsupported is the right call. I've written that invariant into the scope notes so it survives as documentation rather than as a comment thread.

With all three questions closed I've merged main (past #1131 and the new `adiri-test` CI job), updated the description, and taken the PR out of draft.

grantkee (approved):
LGTM - we plan to support pruning in the future, but best to enforce full history initially

PR #1062, opened 2026-07-27
fix(consensus): round-bound canonical epoch-close seed chain (#1032)

### LINKED ISSUE #1032: Epoch-close committee-shuffle randomness — redesign
## Problem

The end-of-epoch committee shuffle is seeded from the leader's **aggregate** BLS signature: `randomness = keccak256(leader.aggregated_signature().to_bytes())` (`crates/types/src/primary/output.rs:378-382`), surfaced via `keccak_leader_sigs()` (`:263-265`), attached to the closing block at `crates/tn-reth/src/payload.rs:87-91`, and consumed by `shuffle_new_committee` (`crates/tn-reth/src/evm/block.rs:429-483`) to build the `concludeEpoch` calldata that writes the next committee on-chain.

The certificate digest is **header-only** (`crates/types/src/primary/certificate.rs:473-479`) and excludes both the aggregate signature and the `signed_authorities` bitmap. The certificate store is idempotent by digest (`crates/consensus/primary/src/state_sync/cert_validator.rs:93-98`, first-write-wins). With any committee of n ≥ 4 (f ≥ 1), multiple distinct quorum-meeting signer subsets exist. A Byzantine leader can therefore aggregate two different 2f+1 subsets for the **same** header → two valid certificates with the same digest but different aggregate bytes → ship different variants to different honest validators → each validator derives a **different seed → different committee → different `concludeEpoch` calldata → divergent EVM state root at epoch close**. `randomness` also folds into the consensus-header digest (`output.rs:469`), so the consensus chain itself forks. This is a BFT safety fork, mitigated operationally only by the permissioned validator set.

## Why the seed was the aggregate signature (the property to preserve)

The aggregate signature only materialises after 2f+1 **other** validators sign the leader's header. The leader cannot compute it at proposal time, so it cannot *grind**: it cannot try many candidate headers, compute the resulting randomness for each offline, and pick the header that yields the most favorable committee shuffle. This anti-grinding / anti-anticipation property is the reason the aggregate signature was chosen. The flaw is purely that the aggregate is **not canonical** (subset choice → different bytes), which is what enables the fork.

## Attacker surface (why grinding is a real, not theoretical, concern)

- Leader election is deterministic: round-robin `(round/2 - 1) % n` plus a reputation swap table that only changes every 60 committed sub-dags (`crates/consensus/primary/src/consensus/leader_schedule.rs:285-304`). A validator knows its leader rounds well in advance; the proposer already computes this (`crates/consensus/primary/src/proposer.rs:358-390`).
- The epoch boundary is `epoch_start + epochDuration`, a fixed on-chain value known at epoch start (`crates/node/src/manager/node/run_epoch.rs:138-143`). The closing sub-dag is the first committed leader with `commit_timestamp = max(prev, leader.created_at)` ≥ boundary (`output.rs:369`). `created_at` is only bounded to `[max_parent_ts, now+1s]` (`crates/consensus/primary/src/network/handler.rs:780-834`), so a leader near the boundary can nudge ±1s to **claim or defer** the closing slot.
- Net: an adversarial validator can predict, and partly steer, whether it is the epoch-closing leader — a repeated, steerable opportunity every epoch. The payoff (biasing the next committee's composition via the shuffle) compounds across epochs.

## Options

Seed from the **leader's individual BLS signature over a FIXED canonical message**, not over its own header and not from the aggregate. `seed = keccak256(leader_sig)`.

- BLS signatures are deterministic (`s = H(m)^sk`, no nonce) → exactly one valid signature per (key, message) → **canonical → the fork is fixed** (a Byzantine leader cannot produce two valid signatures).
- The message is **fixed before the leader proposes** and **not leader-controllable**, so the leader cannot grind by reshaping batches/parents/timestamp. The only residual lever is the coarse `created_at` positional choice (claim vs defer the closing slot), which is far weaker than unlimited offline hashing.
- Known residual: the leader can compute its own signature early (it holds the key), so it knows its own potential seed in advance — but it cannot change it. Fully removing early knowledge requires a threshold VRF / randomness beacon (unique threshold signature over a fixed message), which needs distributed key generation (DKG) infrastructure the network does not have today. 

Pick the concrete canonical message the leader signs. It MUST be:

1. **Canonical** — every honest node computes byte-identical inputs from committed / on-chain state.
2. **Not leader-controllable** — no component the leader can vary to change the seed (rules out its own header digest, its own batch set, its own `created_at`).
3. **Fixed before the closing header is proposed** — so there is nothing to grind against at proposal time.
4. **Reachable at the derivation point** (see plumbing constraints) OR the derivation point moves to where the inputs are reachable.

Candidate messages (evaluate and choose):

- **`keccak(epoch_number)`** — trivially canonical and fixed. Weakness: the leader can precompute its signature for all future epochs indefinitely, and the outcome carries no fresh cross-epoch entropy (pure function of which validator closes + epoch#).
- **`keccak(epoch_number ‖ prior_epoch_final_consensus_digest)`** *(recommended starting point)* — adds cross-epoch entropy the leader cannot touch. `EpochRecord` N-1 is available in-DB at epoch N (`crates/node/src/manager/node/run_epoch.rs:510-523`) but is **not currently plumbed** into the consensus/executor derivation point — verify reachability (see below).
- Any other canonical, leader-independent, pre-proposal value — justify against the four constraints.

## Plumbing constraints (the real implementation cost)

- Randomness is derived today in `CommittedSubDag::new` (`crates/types/src/primary/output.rs:355-391`), from `leader: Certificate`. The certificate carries only the **aggregate** signature and the `signed_authorities` bitmap — it does **not** carry the leader's individual signature over any fixed message. So a new signature must be **produced by the leader, carried through proposal → certificate/output → `CommittedSubDag::new`, and verified** by other nodes before it can seed the shuffle.
- Decide WHERE the leader's fixed-message signature lives and how it is validated:
  - A new field in `HeaderInner` — but this changes the header digest and thus certificate/DAG identity; assess the blast radius on parents, dedup, and store keys.
  - A field carried alongside the certificate/output, verified against the fixed message leader public key at vote time (`vote_inner`, `handler.rs`) and/or at certificate validation (`certificate.rs:225-251`), before it can be committed.
- Reconsider the current silent fallback `BlsSignature::default()` when the signature is missing (`output.rs:378-381`). Under a self-sig scheme a missing/invalid leader signature should be handled **deterministically** — prefer skipping the leader (as an invalid closing signature) over silently defaulting, which is itself a grind / DoS vector.
- Preserve the invariant that `randomness` folds deterministically into the consensus-header digest (`output.rs:469`).
- If the chosen message needs prior-epoch or on-chain data not reachable in `CommittedSubDag::new`, decide whether to (a) plumb that data down to the consensus layer, or (b) move seed derivation to the payload/executor layer (`crates/tn-reth/src/payload.rs`, `evm/block.rs`) where epoch/registry context is available — and keep the consensus-header digest deterministic either way.

## Required tests (design + list here; implement in PR 1)

- **Fork determinism**: two differing-but-valid quorum subsets of the same leader header produce the **same** seed (the original report's requested pin).
- **Grinding resistance**: the leader cannot change the seed by reshaping its header (vary batch set / parents / `created_at` → seed unchanged).
- **Cross-node agreement**: all honest nodes derive the identical seed for the same committed closing sub-dag.
- **Missing/invalid leader signature**: deterministic handling (leader skipped, no silent default).

## Alternative

Threshold VRF / randomness beacon for full anti-anticipation (no single party can compute the seed early). Requires DKG.

MavenRain:
Closes #1032.

## Problem

Two defects share one derivation, and this PR closes both.

**1. The seed is not canonical, so the epoch boundary forks.**  The end-of-epoch committee shuffle is seeded from the leader's **aggregate** BLS signature: `randomness = keccak256(leader.aggregated_signature().to_bytes())` (`crates/types/src/primary/output.rs`).  The certificate digest is header-only and the certificate store is idempotent by digest (first-write-wins), so a Byzantine leader can aggregate two different 2f+1 signer subsets over the **same** header, producing two valid certificates with the same digest but different aggregate bytes.  Shipping the two variants to different honest validators makes each derive a different seed, hence a different shuffled committee, a different `concludeEpoch` calldata, and a divergent EVM state root at epoch close.  Because `randomness` also enters the sub-dag digest and from there the consensus-header digest, the consensus chain itself forks.  It is a BFT safety fork held off only by the permissioned validator set.

The aggregate was chosen for a real reason: it only materialises after 2f+1 *other* validators sign the leader's header, so the leader cannot grind candidate headers offline to bias the shuffle.  The defect is not that property, it is that the aggregate is **not canonical** (the signer-subset choice changes the bytes).

**2. A canonical signature over a per-epoch message publishes the entire epoch's seed schedule.**  Replacing the aggregate with a deterministic signature over a message that is *fixed for the epoch* removes the fork but opens a grinding lever.  min_sig BLS is nonce-free, so a signature over a fixed message is a publicly derivable constant: from round 1 of an 8 hour epoch, every authority's future contribution is already computable by everyone.  The closing leader then steers **which** of the n pre-known values lands on the wall-clock boundary, and that seed selects committee membership two epochs out.  Mixing in more signers makes this worse rather than better, because the only remaining freedom is which contributions are included, and that is exactly what a leader controls through its parent set.

## Root cause

For (1), the seed input is a value the leader does not uniquely determine: any quorum-meeting subset satisfies certification, and nothing pins *which* subset's aggregate seeds the shuffle, so the same committed header admits multiple valid seeds.

For (2), the load-bearing property is not how many signers contribute, it is whether the signed message contains a component that is **late-revealed**.  A message fixed for the epoch has none.

## Fix

Seed from the leader's **individual, deterministic** BLS signature over a canonical message that **binds the header's round**, and make the epoch's randomness a **hash chain over its committed prefix** rather than a value read off a single signature.

- **`crates/types/src/primary/epoch.rs`** adds the canonical message `EpochSeedMessage { epoch, round, prior_epoch_record }`, where `prior_epoch_record = EpochRecord(epoch - 1).digest()` (`EpochDigest::default()` for epoch 0, matching the repo's epoch-0 filler convention).  It is canonical (every honest node derives byte-identical inputs), it is not leader-controllable (no header content enters it), it carries cross-epoch entropy the leader cannot touch, and the round binding means a signature for round `r` cannot exist before its author proposes at `r`.  `intent_message` is private so signing and verification can never diverge on the encoded bytes.
- **Domain separation.**  The message is signed under a new `IntentScope::EpochCloseSeed = 4` (`crates/types/src/crypto/intent.rs`), byte-prefix-disjoint from proof-of-possession (0), epoch-boundary (1), consensus-digest (2) and system-message (3) signatures, so a seed signature can never be replayed as a vote or a consensus-digest signature, or the reverse.
- **Deterministic BLS.**  min_sig signing has no nonce (`s = H(m)^sk`), so exactly one valid signature exists per `(key, message)` and therefore per `(authority, epoch, round)`.  Canonical message plus deterministic signature means a Byzantine leader cannot produce a second valid seed for a committed header, which closes the fork.
- **`crates/types/src/primary/header.rs`** carries the author's signature as a `seed_signature: Option<BlsSignature>` field on `HeaderInner`, present on the wire exactly when `seed_signature_active(header.epoch)` holds (from `SEED_SIGNATURE_FORK_EPOCH` onward under the `adiri` feature, from genesis in every other build;  see Rollout below).  When present it lives inside the header, so it is covered by the header digest, and therefore by every vote, the certificate aggregate, digest-keyed dedup, and the committed sub-dag, with no change to the certificate envelope.
- **`crates/types/src/primary/seed_chain.rs`** (new) holds the derivation as one type, `EpochSeedChainValue`, with domain-separated root and fold:

      root(epoch)      = keccak256("TN_EPOCH_SEED_ROOT_V1" || epoch_le)
      randomness(C_i)  = keccak256("TN_EPOCH_SEED_FOLD_V1" || prev || leader_round_le || seed_sig)

  It has **no `Default`**, deliberately, so no code path can silently re-root a chain: the only ways to obtain a value are `epoch_root(epoch)`, `from_committed(randomness)` (recovering the value carried by a commit this node already has), and the pinned `genesis_placeholder()` (`B256::ZERO`), which is reserved for `ConsensusHeader::default` (state sync's pre-genesis anchor) and test fixtures.
- **The chain root depends on the epoch number alone.**  Folding `prior_epoch_record` into the root was considered and rejected: that digest is read from an uncertified, first-write-wins local `EpochRecord`, and the chain value reaches `randomness`, then the sub-dag digest, then the consensus-header digest, then the executed block's `parent_beacon_block_root`.  A node with a divergent record would therefore fork **execution** permanently on the epoch's *first* commit.  The cross-epoch anchor stays in the signed message, where voters check it, so a divergent node merely fails to get its own headers certified while its fold over everyone else's commits still matches the network exactly.
- **`crates/consensus/primary/src/proposer.rs`** signs the message in `propose_header`, immediately before `Header::new`, rather than once in `Proposer::new`: the message binds the round, so a per-epoch signature is not expressible.  Signing is gated on `seed_signature_active` for the header's own epoch, so dormant deploys propose byte-identical legacy headers.  The cost is one BLS sign per proposed header once active.  `KeyConfig` and the epoch's `prior_epoch_record` travel with the new `HeaderIdentity` type (round, epoch, author, prior record, signer), so the spawned header-building task takes one self-describing argument instead of a longer positional list.
- **`crates/consensus/primary/src/network/handler.rs`** verifies the seed signature in `vote_inner` before any state mutation, after the cheap structural checks and before the expensive parent sync, and refuses to vote with a new `HeaderError::InvalidSeedSignature`.  The check gates on the header's own epoch, mirroring the signer.  Because every honest vote is produced through this path, a certificate carrying a 2f+1 aggregate over the header digest proves an honest validator verified the leader's seed signature, which closes the gossip and fetch paths (`VerifiedIndirectly` certs) with no re-check needed at those entry points.
- **`InvalidSeedSignature` is `Penalty::Severe`, not `Penalty::Fatal`** (`crates/consensus/primary/src/error/network.rs`).  It has a reachable honest cause: the anchor is read from an uncertified, first-write-wins `EpochRecord` store, so a node whose record diverged signs an anchor no peer accepts *and* rejects every honest peer's header.  A fatal penalty would turn one node's storage divergence into a mutual, non-self-healing ban.  (The certified-anchor hardening below shrinks that honest window further.)
- **`crates/types/src/primary/output.rs`** folds in `CommittedSubDag::new`.  A post-fork leader header carries the signature, and the fold consumes it;  a pre-fork leader reproduces main's legacy expression **verbatim** (`keccak256` of the leader certificate's aggregate signature, with the same `error!` + default fallback), so every pre-fork commit derives the exact historical randomness.  The silent-default fallback for a *post-fork* missing signature remains unrepresentable: the field is digest-covered, so a post-fork header without it never certifies.
- **`crates/consensus/primary/src/consensus/{state.rs, bullshark.rs}`** thread the chain through consensus: `ConsensusState` holds the current value, Bullshark folds each commit and advances it in commit order, and `resolve_seed_chain_anchor` recovers the anchor at startup.  That function is **fallible and exhaustive** over the four `(latest_sub_dag, last_committed_round > 0)` combinations, and is split out of `Consensus::spawn` so all four are testable without standing up a node.  Only two combinations are valid (continue from the commit the cursor points at, or root a genuinely fresh epoch);  the other two, plus a leader-round mismatch, return `EpochSeedChainError` rather than falling back, because the only available fallback (start from the epoch root) re-roots the chain mid-epoch, and a fold divergence that reaches `parent_beacon_block_root` never re-converges.  The previous `unwrap_or_default` turned a pack read error into `None` while the guarding `assert_eq` sat *inside* the `if let`, so the check was skipped on exactly the failure path.
- **`crates/storage/src/{consensus.rs, consensus_pack.rs}`** make the read path fail closed.  `ConsensusPack::latest_consensus_header` returns `Result<Option<ConsensusHeader>, PackError>` (the pack actor's oneshot now carries the `Result`, and a dropped channel is `PackError::SendFailed` rather than "no header"), and `ConsensusChain::{latest_consensus_header_from_pack, read_last_committed}` propagate a failed pack open instead of returning `None` / an empty map.  Startup recovery reads both cursors from that same pack, so failing open in both reproduced the silent re-root one layer down.  `None` now means only "this epoch has committed nothing yet", and the API leaves callers that legitimately probe an epoch this node may not hold (state sync's partial-pack catch-up) to degrade the error at their own call site, where "not held" is the intended reading.
- **`crates/config/src/consensus.rs`** plumbs `prior_epoch_record: EpochDigest` as a single field on `ConsensusConfig`, with a `prior_epoch_record()` accessor and a test constructor that seeds a specific digest.  The epoch manager (`crates/node/src/manager/node/{run_epoch,start_epoch}.rs`) supplies the real prior-record digest, which is guaranteed present before the epoch's primary spawns.
- **`crates/node/src/manager/node/run_epoch.rs`** epoch-open gate corrected from `previous_epoch == 0` to `current_epoch == 0`.  `previous_epoch = current_epoch.saturating_sub(1)` collapses both epoch 0 and epoch 1 onto `previous_epoch == 0`, so the pre-existing filler-record branch fired for epoch 1 as well as genuine genesis.  A node opening epoch 1 before it has synced the real epoch-0 record would fabricate a filler whose digest differs from the real record, diverging `prior_epoch_record` across nodes and breaking canonicality at the first boundary.  Gating on `current_epoch == 0` sends epoch 1 down the same wait-or-error path as epoch >= 2 while preserving the genesis case.
- **`crates/tn-reth/src/lib.rs`** anchors the `consensus_output_for_tests` fixture on `epoch_root(epoch)` instead of a single `genesis_placeholder()` for every sub-dag, matching production, where epoch `e`'s first commit roots at `epoch_root(e)`.  Pre-chain this was invisible (the fixture leader is `Certificate::default()`, so the seed was already constant);  under the fold, one anchor for all epochs froze the shuffle seed across every epoch instead of varying it.
- Stale "aggregate signature" comments updated in `output.rs` and `crates/tn-reth/src/{payload.rs, evm/config.rs, evm/block.rs}`, and the `CommittedSubDag::keccak_leader_sigs` accessor renamed to `committee_shuffle_seed` to name what it now returns.  Remaining `aggregated_signature()` references are the certificate's own aggregate, which is unchanged and still used for quorum verification.

## Rollout: epoch-gated in-band serde, no flag day (#1086)

An earlier draft of this PR shipped `seed_signature` as a mandatory field, a deliberate flag-day wire break.  That is gone.  The field is now epoch-gated in-band, per the plan approved in #1086 (amendments 1 and 3 kept, amendment 2 reduced to the startup log):

- **`crates/types/src/forks.rs`** adds `SEED_SIGNATURE_FORK_EPOCH` (`adiri`-only, deliberately `u32::MAX` so this PR ships **dormant**) and `seed_signature_active(epoch)`.  A dedicated epoch-setting PR flips the constant to a concrete epoch (at least current + 8, about 48h of lead time at 6h epochs) once every validator and observer runs a gate-capable build.  Non-adiri builds (mainnet) are active from genesis and never carry the legacy layout;  the dormant period exists only where a live chain must keep replaying its history.
- **The gate is always keyed on the epoch inside the value being encoded or decoded** (`HeaderInner::epoch`, `leader.epoch()`), never `Committee::epoch()` or other node-local state, so mixed-epoch containers (certificate vectors, sub-dags, pack records) decode correctly at any nesting depth and historical digests are preserved end to end.  BCS makes this sound: it emits no struct or tuple framing and enforces no visitor arity below the top level, so a hand-written `visit_seq` can read the seven legacy fields, branch on the just-decoded epoch, and conditionally read the eighth;  only the top-level `from_bytes` checks full consumption.
- **Single source of truth for bytes and digests:** the serde derives on `HeaderInner` are removed and replaced by a borrowing `HeaderRef` used by **both** `Serialize` and `Hash`, so the digest preimage equals the wire bytes in both layouts by construction and the wrong preimage is unrepresentable.  The IndexMap `serde_seq` payload shape is preserved through delegate types, keeping payload bytes untouched.
- **Pre-fork behavior is main-verbatim:** pre-fork headers serialize the seven legacy fields byte-identically to `origin/main` (frozen golden byte and digest pins below), and pre-fork commits derive the exact historical leader-aggregate randomness.
- **Straggler story:** past the fork epoch, a node still on an old build fails to decode post-fork headers **loudly** and drops out rather than silently diverging: the new field is covered by the header digest, and decode failures charge `Penalty::Fatal` to the author while committee authors stay exempt from bans.
- **Observability, deliberately minimal:** the node logs its full fork schedule at startup (`crates/telcoin-network-cli/src/node.rs`) so operators can diff it across the fleet;  a compile-time constant that differs between binaries has no other in-protocol detection.  Per #1086 this is the only observability shipped (amendment 2 relaxed).
- **Blocking the epoch-setting PR (amendment 1, Tier 5):** a mixed-fleet rehearsal (old binary and dormant gated binary exchanging pre-fork traffic) and a third golden vector cut on a validation host against a live adiri archive.

## Hardening: certified epoch anchor and fail-loud startup (amendment 3)

The seed message's cross-epoch anchor and the startup resume path previously trusted whatever the local store returned.  Both now fail loudly:

- **`crates/storage/src/epoch_records.rs`** adds `certified_record_by_epoch`, which releases a record only after verifying it against its stored `EpochCertificate`, with the typed `CertifiedRecordError` (`MissingRecord` / `MissingCertificate` / `InvalidCertificate` / `Storage`, retryability encoded per variant), plus a polling `certified_record_by_epoch_with_timeout` for the epoch-open path.  The read path does not collapse storage failures into "missing": genuine absence is classified by the same `fetch_error_is_absent` predicate the pack read path uses, while CRC / IO / deserialize failures become the non-retryable `Storage` variant (logged with the underlying error), so the poll fails fast on real corruption instead of spinning out the timeout and reporting a misleading miss.
- **`crates/node/src/manager/node/run_epoch.rs`** captures the anchor exhaustively three ways: epoch 0 keeps the documented filler;  fork-active epochs require the certificate-verified record, polling up to a bounded budget;  pre-fork epochs keep the legacy read, so dormant deploys behave exactly as today.  The budget is **derived, not guessed**: `epoch_votes.rs` names its own timing constants (per-iteration vote timeout, max timeout count, peer-recovery attempts and their request timeout), and `CERTIFIED_ANCHOR_WAIT` is a const expression over them ... the local-quorum worst case (65s) plus the full peer-recovery worst case (150s) plus one spare recovery attempt of margin (30s), 245s total ... so the anchor wait strictly exceeds every latency the vote protocol itself treats as legitimate, and an honest slow epoch boundary cannot be hard-killed by a shorter local clock.  The wait stays finite, so fail-loud abort semantics are preserved.
- **`crates/storage/src/consensus.rs`**: `save_consensus_output` silently returned `Ok(0)` on a non-increasing consensus number, which can bury a real divergence;  it now returns a typed `NonMonotonicConsensusNumber` error.  The primary network layer (`crates/consensus/primary/src/network/mod.rs`) maps it to a **no-penalty** arm: duplicate delivery at a CVV/observer mode switchover is a benign, reachable cause, so the write is refused and logged without banning the peer.
- **`crates/consensus/primary/src/consensus_bus.rs`** last-executed-block getters are Result-typed, and **`crates/state-sync/src/lib.rs`** propagates errors with context instead of collapsing them via `.ok()?` / `Ok(None)` / `unwrap_or_default()`, so a storage fault at resume is a startup error, not a silent re-root.

## Threat model

The vote gate is cheap-first and attacker-independent: membership and structural checks, then the BLS verify, all before parent sync, so a bogus-seed header mutates no state and costs no fetch.  The seed no longer depends on any signer-subset choice, so the epoch-close shuffle and the consensus-header digest are identical for every honest node regardless of which quorum certified the closing header.  On grinding, the honest statement is a bound, not an elimination: a signature for round `r` cannot exist before its author proposes at `r`, so the prediction horizon drops from the whole epoch to roughly two rounds.  A lone Byzantine closer keeps a **veto** (withhold the commit and forfeit its slot) but can no longer **select** among candidate seeds;  f colluders with a run of `c` consecutive leader slots choose among at most 2^c values, and `c` is 1 in the common case.  The residual is the irreducible RANDAO last-actor bias: the committing leader can evaluate its own fold before broadcasting.  Removing that needs a per-committee DKG or threshold VRF, and the tree has no such primitive, so it is out of scope.  Scale note: shipped genesis committees are 4 (mainnet) and 5 (testnet), where the eligible pool equals the committee size and the shuffle has no membership power at all;  every grinding number above is conditional on onboarding growing the pool past the committee, while the fork fix is unconditional.

The gate adds one new surface: a mixed fleet where binaries disagree on the fork schedule.  That is exactly the failure the release gates below pin against (pre-fork bytes frozen against `origin/main`), and past the boundary a disagreeing node fails loudly (digest-covered field, `Penalty::Fatal` decode) instead of diverging silently.

## Testing

Every regression test below was confirmed by mutation: the fixed arm was reverted, the test was watched failing, then restored.

`crates/types/tests/it/randomness_tests.rs` (7 tests, expectations rebuilt from an independent `independent_fold` / `independent_root` implementation rather than the production helper, so a bug in the helper cannot make its own test pass;  fixtures build at `seed_active_epoch()`, a doc-commented helper returning the first fork-active epoch per cfg lane, so **both** the dormant-adiri and active-from-genesis lanes exercise the V1 seed-chain path):

- `test_randomness_identical_across_certificate_vote_subsets` ... the issue's requested pin.  Committee size 7 (f = 2, quorum 5), one header certified by two genuinely distinct 5-vote subsets.  Asserts the two certificates share a digest, carry different aggregate signatures, and still fold to an **identical** seed.  Fails under the pre-fix aggregate-seed derivation.
- `test_randomness_immune_to_header_content_grinding` ... same authority, epoch and round, headers differing in payload / parents / `created_at`, produce an identical seed signature and identical randomness.  The signature-equality assert `.expect()`s both sides, so a vacuous `None == None` can never satisfy it.
- `test_epoch_seed_message_binds_round` ... a signature for one round does not verify at another, the property the whole redesign rests on.
- `test_fold_binds_previous_commit` ... the fold consumes the previous chain value, so a commit's seed cannot be computed before its predecessor is published.  Its leader-signature asserts also `.expect()` both sides.
- `test_epoch_root_depends_only_on_the_epoch` ... roots are per-epoch constants derived from no node-local state, and two epochs never share a root.
- `test_genesis_placeholder_is_the_pinned_zero_anchor` ... pins `B256::ZERO`, since `ConsensusHeader::default` is reached on the non-test state-sync path and a node-local derivation there would fork the chain at its first link.
- `test_randomness_from_default_seed_signature_is_total` ... folding a default (infinity) seed signature does not panic, catching reintroduction of the removed silent fallback.

`crates/types/src/primary/epoch.rs` (unit tests on the message itself):

- `test_epoch_seed_signature_deterministic` ... the same `(authority, epoch, round, prior record)` always signs to byte-identical output, which is what makes the seed canonical.
- `test_epoch_seed_domain_separation` ... pins the encoded intent prefixes directly: a seed message starts with scope byte 4 and a vote with scope byte 2, and the two encodings are equal-length and unequal, so neither signature can be replayed as the other.

`crates/consensus/primary/src/tests/network_tests.rs` (vote gate):

- `test_vote_rejects_invalid_seed_signature`, `test_vote_rejects_default_seed_signature` ... a wrong-message signature and a default (infinity) signature are each refused with `InvalidSeedSignature` (the blst subgroup check rejects infinity).
- `test_vote_rejects_seed_signature_for_wrong_round` ... a signature valid at a different round is refused, so round binding is enforced at the verifier and not only at the signer.
- `test_vote_seed_signature_uses_configured_prior_epoch_record` ... cross-epoch entropy end to end: an authority with a real non-default prior record earns a vote from a voter holding the same digest and is refused by a voter holding a different one.  Catches a mutation that defaults `config.prior_epoch_record()` at either the sign or the verify site.

`crates/consensus/primary/src/tests/proposer_tests.rs`:

- `test_proposed_header_seed_signature_binds_its_own_epoch_and_round` ... takes a header from the **real** proposer path and asserts its signature verifies at its own round and not at the round before it.  This one is load-bearing and was initially vacuous: the fixture proposes at round 1, so a proposer signing a hard-coded `1` passed all 210 tests until the test was extended to drive a second round.  It skips to the round-2 header with `stream::unfold` plus `skip_while` (the proposer re-proposes on `max_header_delay`, so taking one header is racy).

`crates/consensus/primary/src/consensus/tests/consensus_tests.rs`:

- `test_consensus_recovery_with_bullshark` (extended) ... a new `assert_seed_chain_advances` helper pins the **production** commit loop's chain wiring over sub-dags taken off the real `sequence` channel in commit order, which nothing else exercises: the epoch's first commit folds the epoch root, every later commit folds its predecessor's value, and no commit re-roots.  It catches both ways the loop can collapse the chain to a per-commit constant, deleting `state.set_seed_chain(sub_dag.seed_chain_value())` and substituting a fixed value for `state.seed_chain()`;  both were seen to fail and then restored.
- `test_seed_chain_survives_restart` ... persist the epoch's first commit, drop the handle (the crash), then reopen with `ConsensusChain::new`, the call a restarting node actually makes, and assert the chain value round-trips through the pack unchanged and that `resolve_seed_chain_anchor` returns exactly that commit's value.
- `test_seed_chain_recovery_fails_closed_on_inconsistent_cursors` ... each invalid `(latest_sub_dag, last_committed_round)` combination returns its `EpochSeedChainError` instead of silently re-rooting.

### Release gates for the fork machinery (amendment 1)

Tier 0, byte pins (`crates/types/src/primary/header.rs`, in-crate, run in default-feature CI):

- `test_header_serde_v1_layout` / `test_header_serde_legacy_layout` ... both layouts round-trip through independent `HeaderReprV1` / `HeaderReprLegacy` mirror structs, so the hand-written serde is checked against derived encodings it does not share code with.
- `test_digest_is_wire_hash_v1_layout` / `test_digest_is_wire_hash_legacy_layout` ... the header digest equals the hash of the exact wire bytes in **both** layouts: the single-source `HeaderRef` property stated as a test.
- `test_golden_v1_wire_bytes_pinned`, `test_golden_v1_digest_pinned`, `test_golden_legacy_wire_bytes_pinned`, `test_golden_legacy_digest_pinned`, `test_golden_legacy_header_wire_bytes_pinned`, `test_golden_legacy_header_digest_pinned` ... frozen hex golden vectors for wire bytes and digests in both layouts.  A layout regression fails against pinned **bytes**, not against code that could drift with it.  The legacy pins freeze pre-fork bytes to `origin/main`'s encoding, which is the mixed-fleet safety property.

Tier 1, nesting proof (`crates/types/tests/it/nesting_tests.rs`, `header_sweep_tests.rs`):

- `test_mixed_epoch_certificate_vector_round_trip` ... a certificate vector mixing pre-fork and post-fork epochs round-trips with each element keeping its own layout, proving the epoch gate works below the top level where BCS enforces no arity.
- `test_epoch_zero_certificate_vector_round_trip` ... the all-legacy container case.
- `test_mixed_epoch_vector_epoch_corruption_fails_loudly` ... corrupting a nested element's epoch makes the decode fail loudly rather than silently mis-framing the following bytes.
- `test_header_layout_parameter_sweep` ... sweeps header layouts across epochs against the gate's prediction, with a byte-level oracle per grid point: the encoded length must equal an independently-encoded seven-field legacy mirror's length plus the pinned signature wire bytes exactly when the gate is active.  Confirmed by mutation: removing the epoch check symmetrically from **both** the serialize and deserialize gates (the escape that an accessor-only assertion cannot see) fails the length oracle at the first pre-fork grid point.

`crates/types/src/primary/block.rs`:

- `test_consensus_header_default_digest_pinned` ... pins the pre-genesis state-sync anchor digest per cfg lane (the value is build-dependent by design and the doc comment now says so).

### Hardening tests (amendment 3)

`crates/storage/src/epoch_records.rs` (all five confirmed by mutation in this cycle: five semantic mutants ... inverted `verify_with_cert` check, wrong error variant at the certificate lookup, verification skipped entirely, wrong variant at the record lookup, `MissingCertificate` made non-retryable ... each killed by an assertion failure and restored to 5/5 green):

- `test_certified_record_by_epoch_accepts_certified_record`, `test_certified_record_by_epoch_refuses_uncertified_record`, `test_certified_record_by_epoch_refuses_invalid_certificate`, `test_certified_record_by_epoch_missing_record`, `test_certified_record_with_timeout_heals_when_cert_arrives`.
- `test_certified_record_by_epoch_reports_storage_corruption` ... writes two certified records, flips one payload byte on disk to break the CRC, and asserts the read classifies it as the non-retryable `Storage` variant with the timeout variant returning immediately instead of spinning out its budget;  a positive control shows the uncorrupted epoch still resolves.  Killed by the mutant that classifies the corruption as `MissingRecord` (the pre-fix behavior).

`crates/node/src/manager/node/run_epoch.rs`:

- `certified_anchor_wait_covers_the_vote_protocol_worst_case` ... pins the anchor-budget derivation against the same named constants `epoch_votes.rs` runs on: the wait strictly exceeds the local-quorum worst case and the full vote-plus-recovery ceiling.  Killed by the mutant that reverts the budget to a bare 30s literal.

`crates/storage/src/consensus.rs`:

- `test_save_consensus_output_non_monotonic_rejected` ... the silent `Ok(0)` is gone: a non-increasing consensus number is a typed error, and the store is unchanged afterwards.

`crates/consensus/primary/tests/it/{recovery_tests,storage_tests}.rs` were renumbered deliberately: consensus outputs numbered from 0 were silently never persisted before this branch, so fixtures now number from 1 and the tests would catch the old silent drop.

Remaining changes are mechanical: `CommittedSubDag::new` call sites take the new `seed_chain` argument, and `HeaderBuilder` / test-utils gain a `seed_signature` setter.  `test_header_serde` and its `HeaderData` mirror cover the new field.

## Deployment

This is **no longer a flag-day wire change**.  The gate ships dormant (`SEED_SIGNATURE_FORK_EPOCH = u32::MAX` under `adiri`): every pre-fork epoch is wire- and digest-identical to `origin/main`, pinned by the Tier 0 golden vectors, so this PR deploys gradually across a mixed fleet with no protocol bump, no migration, and no coordination window.  Activation is a later, dedicated epoch-setting PR (current + 8 minimum), itself gated on the Tier 5 mixed-fleet rehearsal and a validation-host golden vector against a live adiri archive.  Non-adiri builds take the field from genesis and have no dormant period.

Gates: `cargo +nightly-2026-03-20 fmt` (the pinned nightly from `rust-nightly`;  stable `fmt` silently skips `wrap_comments` / `comment_width` / `imports_granularity`);  `cargo +1.94 nextest run --workspace` at **1203 passed, 18 skipped**;  `cargo +1.94 nextest run -p tn-types --features adiri` fully green at 128/128 (the dormant-gate lane) and `-p tn-node --features adiri` fully green at 64/64;  `cargo +1.94 nextest run -p tn-storage --features adiri` green except exactly the two pre-existing `ADIRI_DUP_BATCH_EPOCH` replay failures that predate this branch (`test_consensus_pack_dup_batch_across_certs`, `test_v0_shared_batch_served_as_v1_bytes`);  `cargo +nightly-2026-03-20 clippy --workspace --all-features -- -D warnings` clean.  The full `make attest` (workspace nightly clippy, the `#[ignore]`-gated e2e restart pack, and the on-chain attestation submit) has not been run end to end and should be exercised by CI.


crates/storage/src/consensus_pack.rs
```diff
@@ -575,12 +575,15 @@ impl ConsensusPack {
...
     /// slot files (LatestConsensus) and is always consistent with read_last_committed.
-    pub async fn latest_consensus_header(&self) -> Option<ConsensusHeader> {
+    ///
+    /// Fails closed: a read or channel failure is returned, never reported as "no header". See the
+    /// note on the inner reader.
+    pub async fn latest_consensus_header(&self) -> Result<Option<ConsensusHeader>, PackError> {
         let (tx, rx) = oneshot::channel();
         if self.tx.send(PackMessage::LatestConsensusHeader(tx)).await.is_ok() {
-            rx.await.unwrap_or(None)
+            rx.await.unwrap_or(Err(PackError::SendFailed))
```
sstanfield, line 582: Shouldn't this be a RecievedFailed?

sstanfield (dismissed):
This looks pretty spiffy to me.  As noted it doesn't cover that final withhold issue but should make that a lot harder to catch and keep up.  Also might be a good scaffolding if/when we close that.

grantkee (dismissed):
LGTM!

grantkee (approved):

PR #1070, opened 2026-07-29
feat(engine): bind the WorkerConfigs runtime-update ABI

Huwonk:
## Summary

Companion to Telcoin-Association/tn-contracts#161. Updates the `WorkerConfigs` `sol!` bindings for the widened `uint184` data fields and declares the contract's new runtime-update surface.

## Changes

- `system_calls.rs`: `uint128` data widens to `uint184` across the constructor, `getWorkerConfig`, `getAllWorkerConfigs`, and `setWorkerConfig`; new declarations for the system-call batch mutators `setWorkerConfigsData` / `setWorkerConfigsValue` (strategy and sibling fields preserved, only previously configured workers accepted), the owner-only strictly-increasing `setMaxStrategy`, and the `MAX_STRATEGY` view (now a storage-backed getter on the contract side).
- `lib.rs`: the pre-genesis deploy zero-fills `datas` with `U184::ZERO`, and the governance worker-config test encodes its `setWorkerConfig` data with the widened type. The per-epoch fee read is behaviorally unaffected: only `strategies` and `values` are consumed today, and the `datas` arity check is type-agnostic.

No caller emits the new system calls yet; that wiring lands with whichever protocol feature drives runtime fee updates. This PR keeps the bindings in lockstep so the ABI surface is ready.

## Lockstep

Rust tests that deploy `WorkerConfigs` pre-genesis encode constructor args against the vendored artifact bytecode, so this stays draft until the `tn-contracts` submodule pin and artifacts advance past tn-contracts#161 (artifacts already regenerated there). Folds into the same pin sequence as #1012/#909.

`cargo check -p tn-reth` passes.

### DIFF
--- a/crates/tn-reth/src/lib.rs
+++ b/crates/tn-reth/src/lib.rs
@@ -1647,7 +1647,7 @@ impl RethEnv {
             );
 
             let (strategies, values): (Vec<u8>, Vec<u64>) = worker_configs.iter().copied().unzip();
-            let datas = vec![0u128; strategies.len()];
+            let datas = vec![alloy::primitives::aliases::U184::ZERO; strategies.len()];
             let constructor_args =
                 WorkerConfigs::constructorCall { strategies, values, datas, owner_: owner_address }
                     .abi_encode();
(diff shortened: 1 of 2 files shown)


Huwonk:
Merged into #1012 (feat/unified-conclude-epoch) so the WorkerConfigs uint184 runtime-update bindings ride with the tn-contracts dc248c0 pin they depend on.

PR #1072, opened 2026-07-30
tn-reth epoch-boundary hardening sweep

### LINKED ISSUE #1073: Epoch-boundary execution: silent failure modes and untested consensus invariants
## Problem

The epoch-closing block is the node's most consensus-sensitive execution path. Every validator must re-derive the identical boundary block or the fleet forks, and a fatal error in a boundary system call halts every node deterministically. Several failure modes along this path are silent, misleading, or undiagnosable:

- A failed boundary system call reports only "failed {description}". The revert output is never decoded, so the on-chain cause of a fleet halt (a revert selector like `InvalidCommitteeSize`, or a halt reason) cannot be recovered from the error, and reverts are indistinguishable from halts.
- `get_epoch_info_at_block` assumes the registry reverts for any epoch outside its ring buffer. It doesn't: a future epoch inside the serving window returns a synthesized projection with `blockHeight == 0`. The epoch manager uses the returned `blockHeight` as a block-scan floor when deriving next-epoch base fees, so a synthesized or mislabeled record silently corrupts the scan range instead of erroring.
- `send_leftover_consensus_output_to_engine` discards forwarding errors. If the engine channel closes mid-drain, the method still reports the boundary hash and the caller blocks forever in `wait_for_consensus_execution`, waiting on execution of output the engine never received.
- The pre-genesis ceremony maps a missing or empty constructor-storage harvest to `with_storage(None)`, which ships a deployed-but-uninitialized ConsensusRegistry: unownable, with the first epoch close reverting on chain.
- The executor merges transitions inside `finish()` even though reth's block-builder and executor wrappers both merge after `finish()` returns, so every epoch-closing block's bundle carries a phantom empty reverts entry.
- A zero `getNextCommitteeSize()` passes the committee-assembly length check (`0 == 0`) and submits `concludeEpoch([])`, an opaque on-chain revert.

The core invariants of the close are also under-tested. Boundary-block determinism was pinned only inside two adiri-gated tests that CI never compiles, so a default-build regression (a nondeterministic iteration order feeding `applyIncentives`, an RNG-draw reorder in the committee shuffle) had no tripwire. The executor's three-call ordering (`applyIncentives`, then `applySlashes`, then committee reads and `concludeEpoch`) was pinned only by hand-rolled call sequences on a raw EVM, never through the production close path, and reward accounting through the close was exercised only by the engine e2e suite.

## Solution

Every failure surface on the boundary path should fail loud with an error that names the cause. System-call failures should decode the revert reason and selector and keep reverts distinguishable from halts, so a fleet halt is diagnosable from the error alone. `get_epoch_info_at_block` should reject records whose `epochId` doesn't match the request and reject `blockHeight == 0` for any epoch other than 0, the only epoch whose first block is genesis. The leftover drain should stop at the first forwarding failure and report no boundary, letting the dead channel resurface as a hard error on re-entry. The genesis ceremony should refuse an empty storage harvest, and committee assembly should refuse a zero target size client-side with a distinct message. The redundant in-`finish()` transition merge should be dropped so the closing bundle carries exactly the one reverts entry the wrapper's merge produces.

The untested invariants should be pinned at default features through the production close path. A `cfg(test)`-only slash-injection seam on `TNPayload` lets a test drive a non-empty slash list through the real boundary sequence without touching production behavior, so the slash-before-committee-reads ordering can be asserted end to end, with the victim's recorded `exitEpoch` discriminating a slash issued before `concludeEpoch` from one issued after. Determinism should be pinned by executing a byte-identical closing payload in two independent environments and comparing state roots and installed next-epoch state, and reward accounting by seeding leader counts and asserting the exact on-chain split after the close.

grantkee:
Hardening pass over the epoch-close execution path: fail-loud error surfaces, a bundle-hygiene fix, and default-feature tests for the boundary's consensus invariants.

- Boundary system-call failures now decode the revert reason and selector and distinguish reverts from halts, so a deterministic fleet halt is diagnosable from the error alone.
- `get_epoch_info_at_block` rejects synthesized future-epoch records (`blockHeight == 0` outside epoch 0) and epoch-id mismatches instead of handing callers a sentinel block-scan floor.
- Leftover consensus-output draining stops on a closed engine channel and reports no boundary instead of returning a hash the caller would wait on forever in `wait_for_consensus_execution`.
- The pre-genesis ceremony errors when the constructor storage harvest is empty rather than shipping a deployed-but-uninitialized registry.
- Dropped the in-`finish()` `merge_transitions`: both reth wrappers merge after `finish()` returns, so the old call gave every epoch-closing bundle a phantom empty reverts entry.
- Committee assembly refuses a zero `getNextCommitteeSize()` client-side instead of submitting `concludeEpoch([])`.
- New `cfg(test)` slash-injection seam on `TNPayload` feeds the executor's `epoch_boundary_slashes` hook; production builds keep the always-empty body.
- Tests: slash-to-zero ejection through the production close (with an `exitEpoch` ordering discriminator), boundary-block determinism across independent environments at default features (previously only adiri-gated), reward flow from `RewardsCounter` through `applyIncentives` to on-chain balances, bundle hygiene (no `SYSTEM_ADDRESS`, exactly one reverts entry), and `RewardsCounter::clear` semantics.
- Named the pre-genesis create gas ceiling (`PRE_GENESIS_CREATE_GAS_LIMIT`) and documented the lockstep fleet-upgrade requirement on `SYSTEM_CALL_GAS_LIMIT`.
- Docs: the epoch-boundary READMEs now describe the shipped three-call design, and the top-level README build command requires the `adiri` feature.

closes #1073

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,9 @@ Check out the repo and update the submodules:
 
 ### Run an observer against testnet
 
-Build a release version of the node software:
-`cargo build -p telcoin-network --bin telcoin-network --release`
+Build a release version of the node software with the `adiri` feature (required to join the
+adiri testnet — the node refuses the `--chain adiri` flag at startup without it):
+`cargo build -p telcoin-network --bin telcoin-network --release --features adiri`
 
 Generate a config and keys for your observer node:
 `target/release/telcoin-network keytool generate observer --datadir DATADIR --address 0x4444444444444444444444444444444444444444 --bls-passphrase-source ask`
--- a/crates/tn-reth/src/evm/mod.rs
+++ b/crates/tn-reth/src/evm/mod.rs
@@ -67,9 +67,22 @@ pub use utils::calculate_gas_penalty;
 /// Gas budget for a single protocol system call.
 ///
 /// System calls run at `gas_price: 0` and do not count toward the block gas limit, so this
-/// bound exists only to stop runaway execution. It is sized well above the block gas limit to
-/// give the epoch-closing `concludeEpoch` call headroom: that call applies rewards, slashes,
-/// and queued stake settlement across the full validator set in one transaction.
+/// bound exists only to stop runaway execution. The 100M headroom (vs the historical 30M
+/// budget) is for the one-shot `migrateValidatorSets()` walk at the ConsensusRegistry fork
+/// boundary (see `tn-types` `forks.rs`), which touches every validator in a single call; the
+/// regular boundary calls fit far below it.
+///
+/// LOCKSTEP FLEET-UPGRADE REQUIREMENT: this value participates in consensus.
+/// `transact_system_call` swaps `block.gas_limit` to this value for the call's duration, and a
+/// binary built with the old 30M diverges from a 100M binary on any system call needing more
+/// than 30M — exactly the calls the headroom exists for. Every validator must run a binary
+/// with the same value before any close can depend on it. (Relatedly, the epoch boundary is
+/// now three calls — `applyIncentives`, `applySlashes`, `concludeEpoch` — vs two on `main`;
+/// `applySlashes` is present in the deployed testnet and mainnet genesis bytecode and an
+/// empty-array call is a state-neutral no-op, so existing history replays unchanged.)
+///
+/// Read-path note: contract reads that share the system-call plumbing (including RPC reads via
+/// `read_contract_at_block`) inherit this ceiling — an accepted consequence of the raise.
 pub(crate) const SYSTEM_CALL_GAS_LIMIT: u64 = 100_000_000;
 
 /// Gas budget for a single pre-genesis constructor CREATE
--- a/crates/tn-reth/src/evm/mod.rs
+++ b/crates/tn-reth/src/evm/mod.rs
@@ -72,6 +72,16 @@ pub use utils::calculate_gas_penalty;
 /// and queued stake settlement across the full validator set in one transaction.
 pub(crate) const SYSTEM_CALL_GAS_LIMIT: u64 = 100_000_000;
 
+/// Gas budget for a single pre-genesis constructor CREATE
+/// (`TNEvm::transact_pre_genesis_create`).
+///
+/// Deliberately 30M — the chain's block gas limit — as a deploy-size ceiling for the genesis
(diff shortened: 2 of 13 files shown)


sstanfield (dismissed):

README.md
```diff
@@ -16,8 +16,9 @@ Check out the repo and update the submodules:
 
 ### Run an observer against testnet
 
-Build a release version of the node software:
-`cargo build -p telcoin-network --bin telcoin-network --release`
+Build a release version of the node software with the `adiri` feature (required to join the
+adiri testnet — the node refuses the `--chain adiri` flag at startup without it):
```
MavenRain, line 20: I think we probably also want this change in `crates/telcoin-network-cli/README.md`

crates/tn-reth/src/evm/mod.rs
```diff
@@ -67,11 +67,34 @@ pub use utils::calculate_gas_penalty;
...
+/// budget) is for the one-shot `migrateValidatorSets()` walk at the ConsensusRegistry fork
+/// boundary (see `tn-types` `forks.rs`), which touches every validator in a single call; the
+/// regular boundary calls fit far below it.
+///
+/// LOCKSTEP FLEET-UPGRADE REQUIREMENT: this value participates in consensus.
+/// `transact_system_call` swaps `block.gas_limit` to this value for the call's duration, and a
+/// binary built with the old 30M diverges from a 100M binary on any system call needing more
+/// than 30M — exactly the calls the headroom exists for. Every validator must run a binary
+/// with the same value before any close can depend on it. (Relatedly, the epoch boundary is
+/// now three calls — `applyIncentives`, `applySlashes`, `concludeEpoch` — vs two on `main`;
```
MavenRain: Nit: I think it is already these three calls on main.

MavenRain (dismissed):
Just a couple of doc items

README.md
grantkee: Added

crates/tn-reth/src/evm/mod.rs
grantkee: Good catch - updated

MavenRain (approved):
LGTM

PR #1081, opened 2026-08-04
tn-node: offload state-export fs work to tokio's blocking pool

### LINKED ISSUE #1076: tn-node: state-export completion task runs an unbounded `std::fs::copy` (plus sync dir removals) on tokio runtime worker threads
## Summary

Impact class: availability / async-runtime scheduling hygiene.  Opt-in feature (`--enable-state-export`, default `false`), so this is a low-severity hygiene issue, not a reachable liveness bug on a default node.

The epoch state-export completion task is spawned with `TaskSpawner::spawn_task`, which lands on plain `tokio::spawn`: a runtime worker thread on the shared multi-thread runtime, not the blocking pool.  Inside that future, `std::fs::copy(&src_consensus, tmp_dir.join("consensus_data"))` synchronously copies the closed epoch's entire consensus pack.  That file's size is unbounded (it holds one full epoch of consensus data), so at real epoch lengths the copy pins a worker thread for the duration of a large sequential file copy, at exactly the epoch boundary when the runtime is busiest.  The failure-path cleanups (`remove_tmp_export`, 8 call sites, each a synchronous `std::fs::remove_dir_all` of a possibly partially-written export) run on the same worker thread, and one more synchronous `remove_dir_all` sits on the epoch-close critical path itself, inside `export_epoch_state`, which `run_epoch` awaits.

The same completion task already demonstrates the intended pattern 33 lines below the copy: the records/certs bundle write is offloaded via `tokio::task::spawn_blocking` inside `export_bounded_bundle`, with the comment "These are blocking calls that may take some time, so don't jamb up an async thread."  The copy and the dir removals are the outliers, not a deliberate design choice.

Line numbers as of `e95d73b8` on `main` (the export path was last touched by #985, `b6178779`).

## Defect

The completion task is spawned onto the runtime, not the blocking pool:

`crates/node/src/manager/node/close_epoch.rs:421-426`
```rust
        match exporter.trigger_export(reth_env.clone(), block, tmp_dir.clone()) {
            Ok(rx) => {
                // fire-and-forget: log the result without blocking the epoch transition.
                // The reth env task spawner should be the node/app scoped spawner NOT the epoch
                // scoped spawner (about to die).
                reth_env.get_task_spawner().spawn_task(format!("exporting execution state for epoch {epoch}"), async move {
```

`TaskSpawner::spawn_task` routes to `create_task`, which is `tokio::spawn`:

`crates/types/src/task_manager.rs:163-168` and `:189`
```rust
    pub fn spawn_task<F, S: ToString>(&self, name: S, future: F)
    where
    {
        self.create_task(name, future, false);
    }
```
```rust
        let handle = tokio::spawn(async move {
```

Inside that future, the consensus pack copy is synchronous:

`crates/node/src/manager/node/close_epoch.rs:432-440`
```rust
                        Ok(Ok(Some(outcome))) => {
                            // Safe: the closed epoch's pack is sealed at export time (see the
                            // persist note above) — no writer appends to a concluded epoch's pack,
                            // so this reads a complete, immutable file.
                            if let Err(e) = std::fs::copy(&src_consensus, tmp_dir.join("consensus_data")) {
                                error!(target: "tn::snapshot", epoch, error = %e, "failed to copy consensus pack into export");
                                remove_tmp_export(&tmp_dir, epoch);
                                return Ok(());
                            }
```

`src_consensus` is the closed epoch's whole consensus pack data file (`epochs/epoch-{N}/data`, `consensus_pack::DATA_NAME`), built at `close_epoch.rs:393-394`.  Its size scales with the epoch's consensus throughput and has no bound.

The cleanup helper is also synchronous and is called from the completion task at 8 sites (`close_epoch.rs:438, 453, 463, 479, 494, 501, 506, 510`):

`crates/node/src/manager/node/close_epoch.rs:79-87`
```rust
fn remove_tmp_export(tmp_dir: &Path, epoch: Epoch) {
    match std::fs::remove_dir_all(tmp_dir) {
        Ok(()) => {}
        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
        Err(e) => {
            warn!(target: "tn::snapshot", epoch, path = ?tmp_dir, error = %e, "failed to remove temp export dir");
        }
    }
}
```

A removal of a partially-written export can itself be arbitrarily large (it may contain most of a finished state walk plus the copied pack).

Two more sites outside the completion task, listed for completeness:

- `close_epoch.rs:412`: `std::fs::remove_dir_all(&tmp_dir)` runs inline in `export_epoch_state` BEFORE the completion task is spawned.  `run_epoch.rs:378` awaits `export_epoch_state` during the epoch transition, so this one blocks the epoch-close critical path (the stale temp it clears can be a large partial export from a crashed run).
- `close_epoch.rs:55-74` (`sweep_stale_tmp_exports`, sync `read_dir` + `remove_dir_all` per stale entry) is called at startup from `node.rs:772`.  Startup, so the least urgent of the set.

(`std::fs::rename` at `close_epoch.rs:483` is also sync, but it is a same-directory rename: a constant-time metadata op, fine to leave or to fold into the tail of the same blocking closure.)

## Why this is clearly unintended

The blocking-IO discipline is already established in this exact code path, in two places:

1. The very same completion task, 33 lines after the copy, calls `export_bounded_bundle` (`close_epoch.rs:469-476`), which offloads its pack writes:

   `crates/storage/src/epoch_records.rs:572-579`
   ```rust
           // These are blocking calls that may take some time, so don't jamb up an async thread.
           tokio::task::spawn_blocking(move || -> Result<(), EpochDbError> {
               write_bounded_pack(&records_path, Inner::PACK_EPOCH, &records)?;
               write_bounded_pack(&certs_path, Inner::CERT_PACK_EPOCH, &certs)?;
               Ok(())
           })
           .await
           .map_err(|_| EpochDbError::JoinError)??;
   ```

2. `TaskSpawner::spawn_blocking_task` exists precisely for this (`task_manager.rs:210-219`) and is used elsewhere, e.g. `crates/execution/tn-rpc/src/rpc_ext.rs:473`.

## Impact and severity

The runtime worker threads are shared with the consensus primary/worker tasks (same runtime topology as #976).  A pinned worker degrades scheduling latency for consensus tasks for the duration of the copy, and the copy fires at the epoch boundary, which is already the busiest moment for that runtime.

Deliberately downgraded to Low:

- the entire path is gated behind `--enable-state-export`, default `false` (`crates/telcoin-network-cli/src/node.rs:61-64`), and the exporter is only spawned when enabled (`node.rs:764-766`);
- the runtime is multi-threaded, so one pinned worker is degradation, not a stall;
- it fires once per epoch, not per request.

The structural point stands at any size: an unbounded file copy on a runtime worker thread.  Multi-GB packs are an extrapolation to long mainnet epochs, not a measured figure.

## Suggested fix

Mirror the `export_bounded_bundle` pattern for every sync filesystem op on the export path.

The copy, inside the completion task:

```rust
let copy_res = tokio::task::spawn_blocking({
    let dst = tmp_dir.join("consensus_data");
    move || std::fs::copy(&src_consensus, &dst)
})
.await;
```

with a `JoinError` folded into the existing error arm (both outcomes already funnel into "log, `remove_tmp_export`, return").

Make `remove_tmp_export` async and offload the removal, keeping today's NotFound-is-silent logging:

```rust
async fn remove_tmp_export(tmp_dir: &Path, epoch: Epoch) {
    let dir = tmp_dir.to_path_buf();
    let outcome = tokio::task::spawn_blocking(move || std::fs::remove_dir_all(&dir)).await;
    // flatten the JoinError layer, keep the existing NotFound-is-ok / warn-on-error logging
}
```

The 8 call sites become `remove_tmp_export(&tmp_dir, epoch).await`.  The critical-path removal at `close_epoch.rs:412` gets the same `spawn_blocking(...).await` treatment (`export_epoch_state` is already async, so it drops in), and `sweep_stale_tmp_exports` can either stay sync at its startup call site or be offloaded the same way for uniformity.

`TaskSpawner::spawn_blocking_task` is the alternative offload, but it is fire-and-forget (no join handle returned to the caller), and the completion task needs to sequence work after the copy, so `tokio::task::spawn_blocking(..).await` inside the existing task is the smaller change.

## Acceptance criteria

- The completion task performs the consensus-pack copy and all temp-dir removals via the blocking pool;  no `std::fs` call that can scale with epoch size executes on a runtime worker thread in the export path.
- `export_epoch_state` performs no synchronous directory removal before spawning the completion task.
- Existing export tests still pass (the `sweep_stale_tmp_exports` tests at `close_epoch.rs:801` and `:809` are unaffected by signature changes to the completion-task helpers).

## Assumptions to verify

- The `reth_env.get_task_spawner()` spawner forwards to the shared multi-thread runtime described in #976.  If the exporter's completion future were ever moved to a dedicated runtime, this issue shrinks to the critical-path removal at `close_epoch.rs:412`.
- Consensus pack size at production epoch lengths.  The severity call (Low) does not depend on it, but the priority of fixing does.

MavenRain:
Closes #1076.

## Problem

`export_epoch_state` spawns the state-export completion task with `TaskSpawner::spawn_task`, which lands on plain `tokio::spawn`:  a worker thread of the shared multi-thread runtime, not the blocking pool.  Inside that task, `std::fs::copy` duplicated the closed epoch's entire consensus pack, and every failure-path or skip-path cleanup ran a synchronous `std::fs::remove_dir_all` over a possibly partially-written export.  Both scale with epoch size and have no bound, so at real epoch lengths they pin a runtime worker for the duration of a large sequential file walk, at exactly the epoch boundary when the runtime is busiest.  One more synchronous `remove_dir_all` sat on the epoch-close critical path itself:  `run_epoch` awaits `export_epoch_state`, which cleared any stale `epoch-{N}.tmp` inline before spawning the completion task.

No attacker is required, and the export path is opt-in (`--enable-state-export`, default off), so default nodes are unaffected:  this is availability / async-scheduling hygiene.  The completion task already demonstrates the intended pattern for its records/certs bundle write, which is offloaded via `spawn_blocking` inside `export_bounded_bundle`;  the copy and the directory removals were the outliers.

## Fix

- [x] `crates/node/src/manager/node/close_epoch.rs`:  `remove_tmp_export` is now async and runs its `remove_dir_all` on tokio's blocking pool via `tokio::task::spawn_blocking`, keeping the NotFound-is-ok / warn-on-error semantics.  The `JoinError` layer is flattened into the same warn path (`map_err(io::Error::other)` then `flatten`), so a cancelled or panicked blocking task stays observable rather than silently dropped.  All eight completion-task call sites gain `.await`.
- [x] The consensus-pack copy runs on the blocking pool with the same `JoinError` flattening;  its error handling and cleanup ordering are unchanged.
- [x] The pre-spawn stale-temp clear in `export_epoch_state` gets the same offload, so the epoch-close critical path no longer performs a potentially unbounded synchronous removal.
- [x] `sweep_stale_tmp_exports` stays synchronous by design:  it runs once at node startup, where no export is in flight and nothing latency-sensitive contends for the runtime.

## Testing

- `cargo +1.94 test -p tn-node --lib close_epoch`:  9 passed, 0 failed, including the new `remove_tmp_export_removes_dir_and_tolerates_missing` (`#[tokio::test]`), which pins recursive removal plus the silent NotFound fast path.
- Mutation check on the new test:  a mutant that no-ops the offloaded removal fails it (the test was seen to fail, then the pre-mutation bytes were restored).
- `cargo +nightly-2026-03-20 fmt -- --check` clean;  `cargo +nightly-2026-03-20 clippy --workspace --all-features` clean.
- `cargo +1.94 test --no-run --workspace -j 2` in an isolated per-branch target dir (attest's exact test-build command):  clean, exit 0.

### DIFF
--- a/crates/node/src/manager/node/close_epoch.rs
+++ b/crates/node/src/manager/node/close_epoch.rs
@@ -76,8 +76,18 @@ pub(super) fn sweep_stale_tmp_exports(export_root: &Path) {
 /// Best-effort removal of an export temp dir, logging at warn on failure so a leftover temp is
 /// observable (the next boundary's [`sweep_stale_tmp_exports`] will retry it). Replaces the silent
 /// `let _ = remove_dir_all(..)` cleanups on the export completion task's failure/skip paths.
-fn remove_tmp_export(tmp_dir: &Path, epoch: Epoch) {
-    match std::fs::remove_dir_all(tmp_dir) {
+///
+/// Runs the removal on tokio's blocking pool: a partially-written export can be arbitrarily large
+/// (most of a finished state walk plus the copied consensus pack), so an inline `remove_dir_all`
+/// would pin a runtime worker thread. A cancelled/panicked blocking task surfaces through the same
+/// warn path as an IO failure.
+async fn remove_tmp_export(tmp_dir: &Path, epoch: Epoch) {
+    let dir = tmp_dir.to_path_buf();
+    match tokio::task::spawn_blocking(move || std::fs::remove_dir_all(&dir))
+        .await
+        .map_err(std::io::Error::other)
+        .flatten()
+    {
         Ok(()) => {}
         Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
         Err(e) => {
@@ -337,6 +347,11 @@ where
     /// also waits a bounded time for epoch N's own certificate (only aggregated at the next epoch's
     /// start) and fails the export without it, so an importer never has to store the tip record
     /// unverified.
+    ///
+    /// Filesystem work that scales with epoch size (the consensus-pack copy and every temp-dir
+    /// removal) runs on tokio's blocking pool ([`tokio::task::spawn_blocking`]), never on a
+    /// runtime worker thread. The one pre-spawn removal below goes through the same offload,
+    /// since `run_epoch` awaits this method on the epoch-close critical path.
     pub(super) async fn export_epoch_state(
         &self,
         primary: &PrimaryNode<DB>,
@@ -409,7 +424,15 @@ where
         // If it cannot be removed, writing into it would append onto stale bytes and publish a
         // doubled bundle, so skip this export — NOT fatal, the next epoch boundary retries.
         let tmp_dir = export_root.join(format!("epoch-{epoch}.tmp"));
-        match std::fs::remove_dir_all(&tmp_dir) {
+        // The stale temp can itself be a large partial export from a crashed run, so its removal
+        // is offloaded to the blocking pool too: `run_epoch` awaits this method, so an inline
+        // removal here would block the epoch-close critical path.
+        let stale_tmp = tmp_dir.clone();
+        match tokio::task::spawn_blocking(move || std::fs::remove_dir_all(&stale_tmp))
+            .await
+            .map_err(std::io::Error::other)
+            .flatten()
+        {
             Ok(()) => {}
             Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
             Err(e) => {
@@ -432,10 +455,20 @@ where
                         Ok(Ok(Some(outcome))) => {
                             // Safe: the closed epoch's pack is sealed at export time (see the
                             // persist note above) — no writer appends to a concluded epoch's pack,
-                            // so this reads a complete, immutable file.
-                            if let Err(e) = std::fs::copy(&src_consensus, tmp_dir.join("consensus_data")) {
+                            // so this reads a complete, immutable file. The pack scales with the
+                            // epoch's consensus throughput, so copy it on the blocking pool
+                            // instead of pinning a runtime worker for the duration; a
+                            // cancelled/panicked blocking task surfaces as the same copy error.
+                            let copy_dst = tmp_dir.join("consensus_data");
+                            let copied = tokio::task::spawn_blocking(move || {
+                                std::fs::copy(&src_consensus, &copy_dst)
+                            })
+                            .await
+                            .map_err(std::io::Error::other)
+                            .flatten();
+                            if let Err(e) = copied {
                                 error!(target: "tn::snapshot", epoch, error = %e, "failed to copy consensus pack into export");
-                                remove_tmp_export(&tmp_dir, epoch);
+                                remove_tmp_export(&tmp_dir, epoch).await;
                                 return Ok(());
                             }
 
@@ -450,7 +483,7 @@ where
                             // to store the tip record unverified, so fail the export instead.
                             let Some(record_n) = consensus_chain.epochs().record_by_epoch(epoch).await else {
                                 warn!(target: "tn::snapshot", epoch, "epoch record missing at export time; skipping export");
-                                remove_tmp_export(&tmp_dir, epoch);
+                                remove_tmp_export(&tmp_dir, epoch).await;
                                 return Ok(());
                             };
                             if consensus_chain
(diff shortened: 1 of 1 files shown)


grantkee (approved):
LGTM

PR #1083, opened 2026-08-04
fix(storage): surface consensus-store read failures instead of collapsing them to "not found"

### LINKED ISSUE #1075: Storage read failures reach the startup restore guard as "no record": a corrupt or unreadable static consensus pack triggers the delete-chain-data message
## Summary

`check_restore_consistency` (`crates/node/src/manager/node.rs:1366`, added in #985) decides "datadir is an incomplete state restore" from a single bit: `producing_header == None` while `tip.number > 0`.  But `None` conflates two very different situations: the consensus record is genuinely absent, or the consensus store **failed to read it** (corrupt-at-rest static pack, transient I/O failure at open time).  Every error on the lookup path is silently downgraded to "not found" before it reaches the guard, so a node whose data is intact but momentarily or partially unreadable is told to remove `db`, `static_files`, and `consensus-db`.  Following that instruction on a transient read failure destroys recoverable chain data and forces a full re-import or a genesis resync.

All line references are at `origin/main` `e95d73b8` ("test(storage): open test-build MDBX envs in SafeNoSync ... (#982)").

## The operator-facing symptom

`crates/node/src/manager/node.rs:1366-1381`:

```rust
fn check_restore_consistency(
    tip: &SealedHeader,
    producing_header: Option<&ConsensusHeader>,
) -> eyre::Result<()> {
    if tip.number > 0 && producing_header.is_none() {
        let (tip_epoch, _) = deconstruct_nonce(tip.nonce.into());
        return Err(eyre!(
            "datadir is an incomplete state restore: execution is at block {} (epoch {tip_epoch}) \
             but the consensus store has no record of the consensus that produced it. Remove the \
             chain-data directories (`db`, `static_files`, and `consensus-db`) under the datadir \
             and re-run `db load-state` with a COMPLETE export bundle — do NOT delete the datadir \
             itself, which holds your node keys. Or start from an empty datadir to sync from \
             genesis.",
            tip.number
        ));
    }
    Ok(())
}
```

Called from `try_restore_state` at `node.rs:1311-1313` with the `Option` produced by `state_sync::last_executed_consensus_block`.

## Where read errors become "not found"

Layer by layer, from the guard down:

1. `crates/state-sync/src/lib.rs:126-133`: pass-through wrapper, forwards the `Option` unchanged.

2. `crates/consensus/primary/src/consensus_bus.rs:713-729`: the lookup tail is

   ```rust
   consensus_chain
       .consensus_header_by_digest(epoch, consensus_hash.into())
       .await
       .unwrap_or_default()
   ```

   The `.unwrap_or_default()` (line 725) types away the `Result`.  Today this is only a latent collapse: as written, the callee below can never return `Err`, because it swallows failures one layer down.  It becomes a live error-discard the moment anyone fixes the lower layer without also fixing this line.

3. `crates/storage/src/consensus.rs:783-818`, `ConsensusChain::consensus_header_by_digest`.  This is the operative swallow point.  The body contains no `?` and every path returns `Ok(..)`;  in particular, line 806:

   ```rust
   if let Ok(pack) = self.get_static(epoch).await {
       Ok(pack.consensus_header_by_digest(digest).await)
   } else if let Some(staging) = self.staging() {
       // only matches the currently-forming staging epoch
   } else {
       // Don't have this epoch data.
       Ok(None)
   }
   ```

   A real `PackError` from `get_static` (`consensus.rs:1188`, returns `Result<ConsensusPack, PackError>`) falls through to `Ok(None)`.  The staging fallback (lines 808-810) only matches the not-yet-sealed next epoch, so it cannot rescue a lookup against a sealed historical epoch whose pack failed to open.  The same `if let Ok(pack) = self.get_static(epoch).await` discard appears at 12 sites in this file (491, 590, 806, 844, 883, 913, 945, 1001, 1088, 1104, 1120, 1157);  this issue is about the one on the restore path, but the pattern is file-wide.

4. `crates/storage/src/consensus_pack.rs`, what that discarded `Err` can actually carry.  `get_static` resolves through `ConsensusPack::open_static` (line 273) to `Inner::open_static` (line 859), which can fail with:

   - `PackError::EpochLoad` (line 871): the epoch metadata fetch failed;
   - `OpenError::IndexFileOpen` (lines 881, 889, wrapped into `PackError::Open` via the `From` impl at line 1954): either hash-index file failed to open;
   - `PackError::CorruptPack` (line 898): `files_consistent` detected a data/index mismatch, with the source's own comment "Corrupt static file is bad (damaged at rest?), produce an error."

   Crucially, the static open path has **no repair**: `trunc_and_heal` (line 662) is called only from `Inner::open_append` (line 806) and `Inner::open_append_exists` (line 849).  A static pack that is corrupt at rest fails to open on every retry, so the operator sees the delete-data message deterministically, exactly the case where the advice is most destructive if the corruption is limited to one epoch's pack (or to one of its two rebuildable index files).

5. The current-pack branch discards too.  If the node restarts mid-epoch, the lookup takes the `epoch == pack.epoch()` branch instead, which bottoms out in `Inner::consensus_header_by_digest` (`consensus_pack.rs:1101`): index load `.ok()?` at line 1105, `if pos >= self.data.file_len() { return None; }` at line 1110 (source comment: "this could happen if a file is damaged and repaired"), fetch/decode `.ok()?` at line 1113, plus the channel wrapper's `rx.await.unwrap_or(None)` at line 513.  So both restart timings (mid-epoch via the current pack, at an epoch boundary via `get_static` on the just-sealed epoch) funnel read failures into the same indistinguishable `None`.

## The guard's documented invariant does not hold under read errors

`node.rs:1353-1359` justifies the guard:

> Invariant relied on: "execution follows consensus" — `save_consensus_output` persists a block's producing consensus HEADER to the consensus store BEFORE that block executes.  So any healthy node that has executed block `B > 0` always resolves `B`'s producing header, ... (the closing epoch's pack is persisted static before the next opens, so it stays readable).  The only way the header is absent with a populated reth tip is a state-only / partial / crashed-mid-import restore, which this rejects.

"So it stays readable" is doing unstated work: persistence does not guarantee readability.  The invariant is about what was written;  the guard's input reports what could be read.  Those differ exactly when the pack is damaged at rest or the open hits transient I/O errors, and in that gap the guard misdiagnoses a healthy datadir as a partial restore.

## Impact

- An operator hitting a transient I/O failure or a single corrupt/rebuildable pack file at startup is instructed to delete `db`, `static_files`, and `consensus-db`, i.e. all chain data, when nothing (or only one epoch's static pack) is actually damaged.
- Because the static path never heals, the message repeats on every restart, which reads as confirmation that the datadir is unrecoverable.
- The pre-existing consumers of the same lookup are milder but consistent: `prime_consensus` (`state-sync/src/lib.rs:40`) and `last_consensus_parent` (`lib.rs:138-143`) `unwrap_or_default()` the `None` into a default/genesis header, so a read failure there degrades into silently-defaulted state-sync behavior rather than a surfaced storage error.

## Proposed fix

1. `consensus.rs`: stop conflating "absent" and "unreadable" at the `get_static` boundary.  Bare `?` on line 806 is not enough: a genuinely absent epoch also surfaces as `Err(PackError)` from `open_static` today (missing files fail the opens), which is presumably why the call sites use `if let Ok`.  The fix needs a not-found classification, e.g. map the underlying io `NotFound` (missing epoch directory/files) to `Ok(None)` and propagate everything else (`CorruptPack`, `EpochLoad`, index-open failures on present files) as `Err`.
2. Change `ConsensusBus::last_executed_consensus_block` (`consensus_bus.rs:713`) to return `Result<Option<ConsensusHeader>, _>`, replacing the `.unwrap_or_default()` at line 725 with `?`.  Signature blast radius in `crates/state-sync/src/lib.rs`: the pass-through `last_executed_consensus_block` (126-133), `prime_consensus` (32-36), and `last_consensus_parent` (138-141);  `get_missing_consensus` (155-162) already returns `eyre::Result` and only needs its body's `.unwrap_or_default()` swapped for `?`.
3. Branch on the distinction in `try_restore_state` (`node.rs:1311-1313`): on `Err`, fail startup with a distinct message along the lines of "failed to READ the consensus store (epoch N pack): this is a storage error, not a missing record;  do NOT delete chain data", instead of routing through `check_restore_consistency`.
4. Reword the invariant comment at `node.rs:1353-1359` so it no longer claims readability, and document that the guard's `None` now means "confirmed absent".
5. Add a regression test beside the existing `restore_guard_*` tests (`node.rs:1607-1634`): corrupt one byte of a sealed epoch's static pack (or truncate one of its `.hdx` index files), then assert the startup error is the read-failure message and does not contain the delete-data instruction.  The three existing tests all pass a hand-built `Option` literal, so none can catch this distinction.

## Notes

- Found by code review;  I have not reproduced a corrupt-pack startup end to end.  The static analysis chain above (guard input, then `unwrap_or_default`, then the `if let Ok(pack)` discard, then the `Inner::open_static` error sites, and the absence of any `trunc_and_heal` call on the static path) was verified line by line at `e95d73b8`.
- The guard itself is a good addition (a real partial restore should be rejected);  this issue is only about the error channel feeding it.

MavenRain:
Fixes #1075.

## Problem

`check_restore_consistency` refuses startup when execution is past genesis but the tip's producing consensus header resolves to `None`, and its message instructs the operator to remove `db`, `static_files`, and `consensus-db` and re-import.  But `None` conflated two very different situations: the record is genuinely absent (a real partial restore), or the consensus store FAILED TO READ it (corrupt-at-rest static pack, damaged index file, transient I/O at open).  Every error on the lookup path was downgraded to "not found" before it reached the guard, so a node whose data is intact but momentarily or partially unreadable was told to delete all chain data.

No attacker is required.  A single damaged epoch pack (or one of its two rebuildable index files) reproduces the delete-data message deterministically on every restart, because the static open path never heals;  the repeated message reads as confirmation that the datadir is unrecoverable, exactly when following it is most destructive.

## Root cause

Layer by layer (all at `e95d73b8`): `ConsensusChain::consensus_header_by_digest` swallowed the `get_static` error with `if let Ok(pack)`, answering `Ok(None)` for a pack that failed to OPEN;  `ConsensusBus::last_executed_consensus_block` typed the remaining `Result` away with `.unwrap_or_default()`;  the state-sync wrappers passed the collapsed `Option` to `try_restore_state`, which handed it to the guard.  The guard's documented invariant ("the closing epoch's pack is persisted static before the next opens, so it stays readable") let persistence stand in for readability;  those differ exactly when the pack is damaged at rest.

## Fix

- [x] `crates/storage/src/consensus_pack.rs`: new `PackError::is_missing_static_files` classifies a static open failure.  io `NotFound` from the data-file or an index-file open means the epoch's files are absent on disk (`Inner::open_static` opens the data file first, so a missing `epoch-{N}` directory always lands there;  the index-file arm keeps the staging fallback alive while `stream_import`'s cleanup removes an incomplete epoch directory entry by entry);  everything else (`CorruptPack`, `EpochLoad`, header or index damage on present files, non-`NotFound` I/O) is a storage read error.
- [x] `crates/storage/src/consensus.rs`: sealed-epoch lookups in `consensus_header_by_digest` route through the new `get_static_if_present`: a genuinely absent epoch still resolves `Ok(None)` (staging fallback preserved), an unreadable pack now propagates `Err`.  The current-epoch branch still collapses pack-internal read failures (the in-memory channel wrapper types them away below this layer);  that is documented at the branch and left for a follow-up, as are the 12 other `get_static` swallow sites in the file (11 verbatim `if let Ok(pack)` plus the `if let Ok(_pack)` in `contains_decode_epoch`).
- [x] `crates/consensus/primary/src/consensus_bus.rs`: `last_executed_consensus_block` returns `eyre::Result<Option<ConsensusHeader>>`;  `Ok(None)` now means confirmed absent, never "could not read".
- [x] `crates/state-sync/src/lib.rs`: `last_executed_consensus_block`, `prime_consensus`, and `last_consensus_parent` return `eyre::Result`;  `get_missing_consensus` swaps its lookup's `.unwrap_or_default()` for `?`, and `last_consensus_parent` stops discarding `consensus_header_latest` read errors the same way.  An absent record still primes from a default header exactly as before;  only read failures propagate.
- [x] `crates/node/src/manager/node.rs`: `try_restore_state` fails startup on a read error with a distinct message ("this is a storage error, not a missing record - do NOT delete the chain-data directories"), at both the recent-blocks priming loop and the guard's lookup, so only a confirmed absence reaches `check_restore_consistency`.  The `run()` priming of the last forwarded consensus number carries the same wrap.  The guard's invariant comment now states that persistence does not guarantee readability and that detectable read failures are surfaced before the guard runs, and the guard's own message now tells the operator to verify the consensus-db epoch pack files are readable BEFORE deleting anything.
- [x] `crates/node/src/manager/node/start_epoch.rs`, `crates/consensus/executor/src/{subscriber,errors}.rs`: remaining callers thread the `Result`;  the new `SubscriberError::ConsensusChainRead` keeps the executor's restore path typed instead of conflating unreadable with absent.

Scope note: this PR closes the OPEN-time error channel, the class the issue reproduces.  Two collapse channels remain and are documented in code for a follow-up: record-level damage inside a pack that still opens (the open-time consistency check compares file lengths, not per-record CRCs) and the current-epoch pack's own reads (the mid-epoch restart path);  both still resolve as `None`.  The reworded guard message compensates by directing the operator to verify pack readability before deleting.

## Testing

- New regression tests in `crates/storage/src/consensus.rs` (`cargo +1.94 test -p tn-storage --lib`;  full crate suite 179 passed, 0 failed on the pinned toolchain):
  - `test_is_missing_static_files_classifies_absent_vs_unreadable`: a never-created epoch classifies as missing (io `NotFound` on the data file);  a present-but-garbage data file does not;  direct-construction rows pin the index-file arm (io `NotFound` on an index file classifies as missing, `PermissionDenied` on it does not).
  - `test_header_by_digest_distinguishes_unreadable_pack_from_absent`: with epoch 0 current and two saved outputs, a really-present header resolves (positive control, so the negative rows cannot pass vacuously);  epoch 99 (never created) resolves `Ok(None)`;  a garbage `epoch-1` data file surfaces `Err`.
  - Mutation-confirmed, both fixes: reverting the fixed branch back to `if let Ok(pack) = self.get_static(epoch).await` makes the `Err` assertion fail;  removing the `IndexFileOpen` arm from `is_missing_static_files` makes the index-file `NotFound` row fail on exactly that assertion.  Restored, everything passes.  The three pre-existing `restore_guard_*` tests stay green (the guard predicate itself is unchanged).
- Attest's gates reproduced on the staged diff: `cargo +nightly fmt -- --check` clean;  `cargo +nightly clippy --workspace -- -D warnings` and `cargo +nightly clippy --workspace --all-features -- -D warnings` clean;  attest's test build reproduced in an isolated target dir (`cargo +1.94 test --no-run --workspace`) green.


sstanfield (approved):
Good catch.  This might indicate we need to rethink the line we use eyre errors on but that's out of scope and a bigger lift.

MavenRain:
> Good catch. This might indicate we need to rethink the line we use eyre errors on but that's out of scope and a bigger lift.

I definitely hear this, and am always open to reviewing how we handle errors.  Perhaps we could discuss in a standup session.

PR #1085, opened 2026-08-04
fix(cli): verify the epoch-record certificate chain before the state import

### LINKED ISSUE #1079: db load-state: the multi-GB state import runs before the epoch-record certificate chain is verified, so a bad bundle is rejected only after the expensive import
## Summary

`db load-state` orders its work so that all cryptographic verification of the bundle's epoch records happens after the reth state import.  The import (`restore_pack`, `crates/telcoin-network-cli/src/db.rs:227`) writes and root-checks the full exec state, which the code's own comments describe as a multi-GB operation;  the first BLS certificate verification does not run until `verify_and_save_epoch_records` (`db.rs:489`), inside `restore_consensus_and_records` after the import has committed.  The only pre-import binding between the state pack and the record chain is `check_state_pack_matches_record` (`db.rs:204`), which compares the pack's final state to a tip record that has not been signature-checked at that point.

The trust chain does close: a bundle with invalid records or certificates is always rejected before the command reports success, and the error path removes what the import created.  The cost is purely operator-facing.  An operator restoring a large snapshot waits for the whole import, potentially hours, to receive a rejection that the certificate-chain walk could have delivered in well under a second (estimate;  one super-quorum BLS verification per epoch record).  This is an ordering / operator-experience issue, not a security hole.

All line references are at `origin/main` `47bd71f8` ("Epoch read docs and consolidation (#1024)").

## The ordering, step by step

`DbLoadStateArgs::execute` (`db.rs:181`) runs, in order:

1. `db.rs:192-198`: `reject_non_resumable_bundle` reads the bundle's epoch records (`db.rs:331`, `EpochRecordDb::read_records_from_pack`) and returns the last one.  This is a structural read only;  no certificate or signature is checked.  Its own doc comment says so (`db.rs:322-323`):

   > The epoch check is a lightweight up-front read of the records (no index sidecar needed);
   > `restore_consensus_and_records` re-reads and fully verifies them after the reth restore.

2. `db.rs:204`: `check_state_pack_matches_record` binds the exec pack's meta (final block number/hash) to that record's `final_state`.  The function's doc comment (`db.rs:364-366`) is explicit that the cryptography comes later: the record "is cryptographically verified against its certificate later, in `verify_and_save_epoch_records`", and the early pin "makes the super-quorum certificate over the record transitively cover the exec pack's identity".

3. `db.rs:227`: `restore_pack` builds the fresh reth DB, imports the chain scaffold, streams in the full state, and recomputes the state root (`db.rs:411-421`).  This is the multi-GB step;  the comment at `db.rs:201-203` frames the whole point of the pre-flight checks as refusing bad bundles "instead of after a multi-GB import has already committed".

4. `db.rs:247-253`: `restore_consensus_and_records` re-reads the records (`db.rs:455`) and certs (`db.rs:464`) and calls `verify_and_save_epoch_records` (`db.rs:489`).  That is the first cryptographic verification anywhere in the command: per record, `validate_downloaded_record` (`db.rs:609`, defined at `crates/storage/src/epoch_records.rs:609`) checks the epoch slot, the parent-hash link, committee anchoring, and finally the super-quorum certificate signature via `record.verify_with_cert(cert)` (`epoch_records.rs:628`).

So a bundle whose `epoch_records`/`epoch_certs` are garbage in any way the structural checks cannot see (a forged or mismatched certificate, a committee chain that does not anchor to genesis, a broken parent-hash link, a missing cert for some epoch) sails through steps 1-2, pays for the entire step 3, and dies at step 4.

## Impact

- Operator time: the rejection arrives after the import instead of before it.  For a mainnet-sized snapshot that is likely hours of wasted wall-clock and disk churn per attempt (estimate, not benchmarked;  see Notes), to catch a defect that a pre-import certificate-chain walk should reject in well under a second (also an estimate;  one super-quorum BLS verification per epoch record).
- No security impact claimed.  `db load-state` is a local clap subcommand taking a filesystem path;  there is no network listener and no remote trigger, and every bad bundle is still rejected before the node would start.  The normal error path also cleans up correctly: `db.rs:257-270` removes exactly the chain-data dirs the import created (`clean_created_chain_data`), leaving keys and config intact.

## Related: crash-mid-import recovery is undocumented

The cleanup at `db.rs:257-270` only runs when the closure returns `Err`.  A SIGKILL or power loss during the multi-hour step 3 leaves half-written chain-data dirs behind, and on the next attempt `SnapshotRestorer::open` refuses a datadir that already holds chain data (`db.rs:218-219`, `db.rs:410`).  The SYNC.md caveats (`SYNC.md:158-165`) document the failed-import cleanup (`SYNC.md:164`) but not this case, so the operator's first hint is an open-refusal error with no stated remedy.  The remedy is the same surgical one the error path performs: remove `db`, `static_files`, and `consensus-db` under the datadir (never the datadir itself, which holds keys) and re-run.

## Proposed fix

1. Preferred: verify the record chain in memory before touching the datadir.  Between `db.rs:204` and the `RethConfig` setup, hoist the genesis-committee load (currently inside the closure at `db.rs:238-246`;  its yaml is already materialized by the `Config::load_*` step at `db.rs:184-190`), read the certs with `EpochRecordDb::read_certs_from_pack` exactly as `db.rs:464` does, and walk the chain with no DB and no writes:

   ```rust
   fn verify_record_chain_in_memory(
       genesis_keys: &BTreeSet<BlsPublicKey>,
       records: &[EpochRecord],
       cert_by_hash: &HashMap<EpochDigest, EpochCertificate>,
   ) -> eyre::Result<()>
   ```

   `validate_downloaded_record` cannot be reused directly because it anchors epoch k to the already-saved record k-1 in the DB (`epoch_records.rs:615-622`), which is exactly why verification currently interleaves with saving.  The in-memory walk is a `try_fold` over `records` carrying the previous record as the accumulator: anchor epoch 0 to `genesis_keys`, epoch k to the accumulator's `next_committee` and digest, and apply the same four checks as `epoch_records.rs:625-628`.  `restore_consensus_and_records` then re-verifies while saving, which is harmless belt-and-braces at negligible cost next to the import.

2. Cheaper alternative: move `restore_consensus_and_records` (`db.rs:247-253`) above `restore_pack` (`db.rs:227`).  It has no data dependency on the import's outputs, and the failure cleanup already covers both orderings (`preexisting` is snapshotted before the closure at `db.rs:223-224`).  One thing to confirm before choosing this option: that `SnapshotRestorer::open`'s freshness check does not treat the just-written `consensus-db` as pre-existing chain data (its refusal is described as being about reth chain data, `db.rs:218-219`, but I have not traced that check).

3. Either way, add a test in the existing import-verification section (`db.rs:1091` onward): a bundle with one bad certificate must be rejected with the datadir left untouched (no `db/`, `static_files/`, or `consensus-db/` created).  The current tests exercise `verify_and_save_epoch_records` directly and cannot see the ordering.

4. Docs: add a SYNC.md caveat next to the failed-import one (`SYNC.md:164`) covering a killed import: what is left behind, that `SnapshotRestorer::open` will refuse it, and the surgical remove-and-re-run remedy.

## Notes

- Found by code review at `47bd71f8`;  the ordering chain above (steps 1-4 and the absence of any earlier `verify_with_cert` reachable from `execute`) was verified line by line.  I have not timed a mainnet-sized restore;  "hours" is an estimate scaled from the code's own "multi-GB" description, and the sub-second figure for the chain walk is an estimate (one aggregate BLS verification per epoch).
- The pre-flight checks that do exist are well built (`reject_non_resumable_bundle` and `check_state_pack_matches_record` both refuse before any datadir mutation, and their comments show the before-mutation ordering is a deliberate design goal).  This issue is only about the certificate chain not being part of that pre-flight set.

MavenRain:
Closes #1079.

## Problem

`db load-state` did all of its cryptographic verification *after* the reth state import.  The order in `DbLoadStateArgs::execute` was:

1. `reject_non_resumable_bundle` reads the bundle's epoch records structurally.  No signature is checked, and its own doc comment said so.
2. `check_state_pack_matches_record` pins the exec pack to a tip record that had not been signature checked at that point.
3. `restore_pack` writes and root checks the full exec state, the multi-GB step the code's own comments describe.
4. `restore_consensus_and_records` then runs `verify_and_save_epoch_records`, the first `verify_with_cert` anywhere reachable from `execute`.

The trust chain still closed.  A bundle carrying a forged certificate, a broken parent link, a committee that does not anchor to the local genesis, or a missing certificate was always rejected before the command reported success, and the failure path removed what the import created.  So there is no security hole here and none is claimed: `db load-state` is a local clap subcommand over a filesystem path, with no network listener, no remote trigger, and no attacker in the picture.

The cost was operator time.  A rejection that one super-quorum BLS verification per epoch record can deliver in well under a second arrived only after an import that can run for hours on a mainnet-sized snapshot.

## Fix

Verify the whole record chain in memory, before any chain data is written.

One precision, since the wording matters for an operator: "before any chain data" means before any of `db/`, `static_files/`, or `consensus-db/` exists.  It does not mean the datadir is bit-for-bit untouched, because `--chain` resolves through `Config::load_adiri` / `load_mainnet`, which materializes `genesis/committee.yaml` if it is absent.  That file is the trust root this very check reads, it is config rather than chain data, and the cleanup path already excludes it deliberately.

- [x] `crates/storage/src/epoch_records.rs`: extract the four acceptance checks (`epoch_matches`, `parents_match`, `committee_valid`, `cert_valid`) out of `EpochRecordDb::validate_downloaded_record` into a new `validate_record_against_anchor`, which takes the trusted anchor as an argument instead of reading it from a database.  `validate_downloaded_record` now looks the anchor up locally and delegates.  There is exactly one place the acceptance rules live, so the pre-import path and the persisting path cannot drift apart . . . the same reason the file already gives for sharing one validation routine between state-sync ingest and failed-quorum recovery.
- [x] `crates/telcoin-network-cli/src/db.rs`: new `verify_record_chain_in_memory`, a `try_fold` over the bundle's records carrying the previous record as the accumulator (epoch 0 anchors to the local genesis committee and the null parent digest, epoch k to the record just verified), calling the shared predicate.  It opens no database and takes no datadir, which is what lets it run ahead of the import.  `validate_downloaded_record` could not be reused directly: it anchors epoch k to record k-1 *as already stored in a database*, which is exactly the interleaving of verifying and saving that forced verification to wait for the import.
- [x] Each record is anchored at its **position** in the bundle rather than its self-declared `record.epoch`, so a record for the wrong slot cannot satisfy the walk.
- [x] `execute` now runs: structural classification, genesis-committee load (hoisted out of the import closure, so a datadir missing its committee also fails before the import), certificate-chain walk, exec-pack binding, then `restore_pack`.  `check_state_pack_matches_record` now pins the exec pack to a record that has already been verified, which is what its doc comment always claimed the super-quorum certificate transitively covered.
- [x] `restore_consensus_and_records` still re-verifies while persisting.  That keeps the guarantee attached to the code that actually writes the records, at a cost that is noise next to the import.
- [x] Two shared helpers (`read_bundle_certs`, `reject_non_contiguous_records`) so the pre-import path and the persisting path read certificates and check contiguity the same way instead of each carrying a copy.
- [x] `SYNC.md`: the operator sequence now states that the chain is verified before any chain data is written, and a new caveat covers the killed-import case the issue raised (a SIGKILL, OOM, or power loss during the import leaves partial `db` / `static_files` / `consensus-db` behind, the next attempt refuses the datadir as non-empty, and the remedy is the same surgical removal the failure path performs, never the datadir itself, which holds the keys).

**Threat model.**  This is a cheap-first ordering, the same shape as a cheap membership check ahead of an expensive BLS verify.  The work moved earlier is bounded by the number of epoch records (one aggregate verification each) and is attacker independent: a bad bundle is refused before any chain data is created, and a good bundle pays the same total verification cost it always did.  No check was removed or weakened and the rejection set is unchanged . . . only the point at which the rejection arrives moves.  The committee handoff is the one place the walk could silently diverge from the database path, so it is anchored on the predecessor's `next_committee` and tested with a genuinely rotating committee (see below), not just with fixtures where a record's own committee and the one it hands forward are the same set.

## Testing

`cargo +1.94 test -p telcoin-network-cli -p tn-storage --lib` . . . 70 and 177 passed, 0 failed.  Of those, 27 are the `db::` tests, 10 of them new.

New tests, all in the `db::tests` module:

- `verify_chain_accepts_complete_bundle` (positive control, so the refusals below are specific rather than a blanket reject)
- `verify_chain_accepts_a_rotating_committee`
- `verify_chain_rejects_a_successor_outside_the_handed_forward_committee`
- `verify_chain_rejects_forged_tip_certificate`
- `verify_chain_rejects_wrong_genesis_committee`
- `verify_chain_rejects_missing_certificate`
- `verify_chain_rejects_broken_parent_link`
- `verify_chain_rejects_non_contiguous_records`
- `preflight_refuses_exactly_what_the_persisting_path_refuses` (the hoisted check is not a weaker check)
- `load_state_refuses_a_bad_chain_before_creating_chain_data` (end to end through `DbLoadStateArgs::execute`)

The rotation pair exists because every other fixture in this module sets a record's `committee` and its `next_committee` to the same value, which makes the epoch-to-epoch handoff unobservable.  `rotating_chain` gives each epoch a distinct committee and hands the successor's forward, so a walk that threaded a record's own committee instead of the one it handed on is caught.

The end-to-end test builds a real exec-state pack plus a records/certs bundle signed by a committee that is not adiri's, so the bundle clears every structural check and only the certificate chain can refuse it.  Its discriminating assertion is the **absence** of the import's cleanup wrapper: every failure raised inside the import closure is wrapped with `state import failed; removed the chain data it created`, so a bare verification error proves `restore_pack` never ran.  Asserting only that the chain-data directories are absent would not tell the two orderings apart, because the failure path deletes them either way.

Both regression tests were confirmed by mutation:

1. Reverting `execute` to take the tip record without verifying it (the old ordering) failed exactly `load_state_refuses_a_bad_chain_before_creating_chain_data`, and failed with `state import failed; removed the chain data it created ... recomputed state root ... does not match the pack's declared root`, which is the import having run before any certificate check.  24 passed, 1 failed.
2. Weakening the shared predicate inside `validate_record_against_anchor` failed 6 tests, including the pre-existing `verify_and_save_rejects_wrong_genesis_committee`, confirming the extracted predicate is load bearing on both the pre-import and the persisting path.  19 passed, 6 failed.
3. Threading `prev.committee` instead of `prev.next_committee` through the walk's fold failed exactly the two rotation tests and nothing else.  25 passed, 2 failed.  The "nothing else" is the point: it is direct evidence that the other 25 tests could not see that class of bug, which is why the rotation pair was added.

## Scope notes

- The issue also listed a cheaper alternative (move `restore_consensus_and_records` above `restore_pack`).  That was not taken: it writes `consensus-db` before the exec pack is validated, so a refusal would no longer leave the chain-data directories empty.  For the record, and answering the open question the issue raised, `SnapshotRestorer::open`'s freshness check is on the reth canonical tip rather than on `consensus-db`, so that alternative would have worked . . . it is just weaker on the property the existing pre-flight checks were built for.
- No behaviour change for a good bundle beyond one extra in-memory verification pass, and no change to what is stored.


grantkee (dismissed):

grantkee (approved):
Good work @MavenRain !

PR #1087, opened 2026-08-04
perf(storage): bound per-boundary state-export work with a certified-prefix watermark and incremental bundles

### LINKED ISSUE #1078: Epoch-boundary state export rescans every epoch from 0: `first_missing_historical_cert` and `export_bounded_bundle` are O(N) per boundary, O(N^2) over chain life, with the cert scan awaited on the epoch-transition path
**Severity:** Low.  Performance / unbounded growth on the opt-in state-export path (`--enable-state-export`, default off), so default nodes never run either loop.  This is a growth-curve report, not a near-term operational problem;  the sharpest edge is that one of the two loops sits on the epoch-transition path rather than in the background export task.

## Summary

Every epoch boundary on an export-enabled node re-derives, from epoch 0, state that was already derived at the previous boundary:

1. `first_missing_historical_cert` (`crates/storage/src/epoch_records.rs:516`) scans `0..tip_epoch`, awaiting `get_epoch_by_number` per epoch, and restarts at 0 every boundary.  It is awaited inline in `export_epoch_state` (`crates/node/src/manager/node/close_epoch.rs:368`).
2. `export_bounded_bundle` (`epoch_records.rs:547`) loops `0..=through_epoch` with the same per-epoch call, then writes a fresh, self-contained records pack and certs pack covering the full range into the new `epoch-{N}.tmp` bundle dir (`close_epoch.rs:471-483`).  Every historical record and cert is refetched, re-serialized, and ZStd-recompressed at every boundary.

`get_epoch_by_number` (`epoch_records.rs:497`) is two awaited actor round-trips (`RecordByEpoch`, then `CertByDigest`) to the single dedicated DB thread (`epoch_records.rs:217`), each served by a sidecar-index load (`epoch_records.rs:886`, `:897`) plus a file seek, read, CRC32 check, ZStd stream decode, and deserialize (`crates/storage/src/archive/pack.rs:346-383`).  Nothing is cached, incremental, or bounded: no watermark, no decoded-record cache, and both loops restart at 0 each boundary.  Per-boundary work is therefore O(N) in the epoch number, and cumulative work over the chain's life is O(N^2).

## Mechanism and call-site split

Both call sites live in `export_epoch_state` (`close_epoch.rs:340`), which `run_epoch.rs:418` awaits serially in the epoch-boundary arm, after `close_epoch` and `write_epoch_record` and before the next epoch starts.  The two loops land on opposite sides of a blocking split:

- The cert-scan precheck (`close_epoch.rs:368`) is awaited inline, so its 2N serialized round-trips sit on the path to next-epoch startup.  The call site comment calls it a "Cheap pre-check before the expensive plain-state walk" (`close_epoch.rs:360`), and per epoch that is true;  the cost is that N grows without bound and the scan always restarts at 0.
- `export_bounded_bundle` (`close_epoch.rs:471`) runs inside the fire-and-forget completion task (`close_epoch.rs:423-426`), sharing it with the full plain-state walk and an up-to-90s cert wait, so its O(N) work does not block the transition;  it does contend on the same single DB actor thread that serves all other epoch-record reads and writes.

The function's own doc (`epoch_records.rs:506-508`) describes the scan as "Cheap: record + cert actor lookups, no state I/O".  That framing is per-epoch-relative (versus the plain-state walk) and stays accurate;  this issue is about the asymptotic direction, not any single boundary today.

Scale, honestly framed: at 8-hour epochs the chain adds roughly 1,100 epochs per year.  After a decade N is about 11,000, so the inline precheck alone is about 22,000 serialized actor round-trips per boundary, and the export refetches and recompresses about 11,000 records plus certs each boundary.  Payloads are kilobyte-scale (a record is dominated by two committee key lists at 96 bytes per BLS key).  Thousands of round-trips spread over a boundary that already does multi-second background work is why this is Low, but the curve only bends one way.

## Suggested fix

Two independent, individually shippable changes, both local to the storage crate plus one call site:

1. **Contiguous-certified-prefix watermark for the scan.**  Track the highest epoch E such that every epoch in `0..=E` has both a record and a stored cert, and have `first_missing_historical_cert` resume from it.  Two soundness constraints matter:
   - It must be a contiguous-prefix watermark that only advances while `epoch == watermark` has record plus cert, never a max-certified-epoch.  Certs for older epochs legitimately arrive after newer epochs are certified: failed-quorum recovery downloads and persists a cert from a peer after the local close missed quorum (`crates/node/src/manager/epoch_votes.rs:191-251`) and state-sync backfills any epoch missing a cert (`crates/state-sync/src/epoch.rs:30-72`).  A contiguous watermark waits at the hole and self-heals on a later scan;  epochs below it are immutable (record insertion is idempotent and gap-rejecting, `epoch_records.rs:813-838`;  certs are append-only, `:863-877`;  no delete message exists, `:40-65`).
   - Recompute it per process rather than persisting it: `open_append`'s heal step can truncate trailing records or certs after a crash (`epoch_records.rs:677-725`), which would invalidate a persisted value.  `open()` already does a full pass over all records to build `final_numbers` (`:212-215`), so seeding it there is natural.

   The natural home is the `Inner` actor via one new `EpochDbMessage` variant, which also collapses the remaining scan to a single round-trip instead of two per epoch.

2. **Incremental export.**  At boundary N, copy the previously published `state_exports/epoch-{N-1}/epoch_records` and `epoch_certs` files into the fresh `epoch-{N}.tmp` dir, validate the copy (its last record must be epoch N-1), reopen with `Pack::open` in append mode, and append only epoch N's record and cert.  The bundle format supports this: `write_bounded_pack`'s doc notes the pack opens read-write in append mode (`epoch_records.rs:1009-1011`), cert matching is order-independent (`:252-254`), and published `epoch-{K}` dirs are retained at stable paths and never swept (`close_epoch.rs:50`, `:403-419`).  The full rebuild must remain as an unconditional fallback (the previous bundle can be legitimately absent: skipped exports, cert-wait timeouts, operator pruning), and the published `epoch-{N-1}` dir must never be appended to in place, preserving the atomic-publish contract (`close_epoch.rs:483`) and the documented property that live-pack appends cannot race into a bundle (`epoch_records.rs:535-542`).  This is a constant-factor win at publish time (the byte copy is still O(N)), but it eliminates the per-epoch actor round-trips, deserialization, and ZStd recompression;  a self-contained `0..=N` bundle is inherently proportional to its size, so O(1) publishing would need a per-epoch delta layout and matching importer, which this issue does not propose.

Fix 1 is the smaller and lower-risk change and removes the only O(N) work on the blocking path;  with it in place the precheck stays necessary and becomes O(1) amortized.

## Acceptance criteria

- On an export-enabled node, per-boundary `RecordByEpoch`/`CertByDigest` traffic from `first_missing_historical_cert` is bounded by the number of epochs newly certified since the previous boundary, not by the chain's age, and the watermark never advances past a missing cert (late backfill still gets requested).
- `export_bounded_bundle` no longer refetches and recompresses every historical record and cert per boundary in the common case, falls back to the full rebuild whenever the previous bundle is missing or fails validation, and the produced bundle still round-trips through `read_records_from_pack` / `read_certs_from_pack` as a contiguous `0..=N` set.
- Published-bundle atomicity (write to `.tmp`, rename) and the no-race-into-bundle property are unchanged.

## Related, not duplicate

- #1076 covers the same completion task but a different defect (sync `std::fs::copy` and dir removals on runtime worker threads);  it explicitly notes the bundle write itself is already `spawn_blocking`-offloaded.  Both functions here were introduced in #985.

## References

- `crates/storage/src/epoch_records.rs:516-523` (scan loop), `:547-581` (export loop and pack writes), `:497-504` (`get_epoch_by_number`), `:506-508` (the "Cheap" doc), `:217` (single DB thread), `:115-123` (actor serving both lookups), `:813-838` / `:863-877` / `:40-65` (immutability of the certified prefix), `:677-725` (heal truncation at open), `:212-215` (existing full startup pass), `:1009-1011` (append-mode pack evidence), `:535-542` (no-race design intent).
- `crates/node/src/manager/node/close_epoch.rs:340` (`export_epoch_state`), `:345-347` (exporter opt-in gate), `:360` / `:368` (inline precheck), `:423-426` / `:471-483` (fire-and-forget export, per-epoch bundle paths, atomic rename), `:50` / `:403-419` (published bundles never swept).
- `crates/node/src/manager/node/run_epoch.rs:406-448` (boundary sequencing: the precheck sits before next-epoch startup).
- `crates/node/src/manager/epoch_votes.rs:191-251`, `crates/state-sync/src/epoch.rs:30-72` (out-of-order older-cert arrival paths that constrain the watermark design).

All line numbers at `origin/main` `47bd71f8`.

MavenRain:
## Problem

Every epoch boundary on an export-enabled node (`--enable-state-export`, default off) re-derives, from epoch 0, state that was already derived at the previous boundary:

1.  `first_missing_historical_cert` scanned `0..tip_epoch` with two awaited actor round-trips per epoch (`RecordByEpoch`, then `CertByDigest`), restarting at 0 every boundary.  It is awaited inline in `export_epoch_state`, so its `2N` serialized round-trips to the single DB thread sit on the path to next-epoch startup.
2.  `export_bounded_bundle` refetched, re-serialized, and ZStd-recompressed every historical record and cert into a fresh self-contained `epoch-{N}.tmp` bundle each boundary (in the fire-and-forget completion task, contending on the same DB actor).

Per-boundary work was O(N) in the epoch number and cumulative work O(N^2) over the chain's life.  Low severity (opt-in path, kilobyte payloads), but one of the loops sits on the epoch-transition path and the curve only bends one way.

## Root cause

Nothing was cached, incremental, or bounded: no watermark for the cert scan and no reuse of the previous boundary's published bundle, so both loops restarted at epoch 0 each boundary.

## Fix

Two independent changes, both local to the storage crate plus the one call site, per the analysis in #1078.

**1.  Contiguous-certified-prefix watermark for the cert scan.** `Inner` (the DB actor) now tracks the highest epoch E such that every epoch in `0..E` has both a record and a stored certificate, and `first_missing_historical_cert` resumes the scan from it.  A new `EpochDbMessage::FirstMissingHistoricalCert` runs the whole scan inside the actor, collapsing per-boundary traffic from two round-trips per epoch to one round-trip total.

Soundness properties the implementation holds:

- The watermark is a contiguous prefix, never a max-certified epoch.  Certs for older epochs legitimately arrive after newer epochs are certified (failed-quorum recovery, state-sync backfill), so the watermark waits at the hole, keeps reporting it (late backfill still gets requested), and self-heals on the first scan after the backfill lands.  Epochs below the watermark are immutable within a process: record insertion is idempotent and gap-rejecting, certs are append-only, and no delete message exists.
- It is recomputed per process and never persisted: `open_append`'s heal step can truncate trailing records or certs after a crash, which would invalidate any persisted value.  Seeding happens in `open_append` after the heal step, bounded by the stored epochs.
- The public API fails closed: if the DB actor is unreachable, the caller sees epoch 0 as unconfirmed (for any nonzero tip), so the export path skips rather than proceeding blind.

**2.  Incremental bundle export.** At boundary N, `export_incremental_bundle` copies the previously published `state_exports/epoch-{N-1}/epoch_records` and `epoch_certs` packs into the fresh `epoch-{N}.tmp` dir, validates the copies, and appends only epoch N's record and cert (fetched in the single remaining actor round-trip).  Any failure on the incremental path falls back unconditionally to the untouched `export_bounded_bundle` full rebuild, which remains the correctness baseline: the previous bundle can be legitimately absent (first exported epoch, skipped or failed prior export, operator pruning) or fail validation.

Validation anchors both copied packs to the new record before anything is appended, so a stale, padded, truncated, reordered, or wrong-chain previous bundle fails closed instead of leaking into the export:

- each pack must hold exactly N entries (epochs `0..=N-1`), walked via the record-size chain and required to land exactly on the file length (a torn pack can never validate);
- the final copied record must be epoch N-1 with the digest the new record's `parent_hash` names;
- the final copied cert must certify that same digest.

Honest cost accounting: per-boundary work remains linear in the chain length, because the file copy and the CRC-checking chain walk both touch every historical byte.  What the incremental path removes is the expensive part the issue measured: the per-epoch round-trips into the single DB thread, per-record deserialization, and ZStd recompression (all replaced by one actor round-trip plus sequential local file I/O on a blocking thread).  The walk's per-entry CRC check is also what keeps a bit-rotted or truncated previous bundle from propagating into new exports;  it fails closed into the full rebuild.

The published `epoch-{N-1}` dir is only ever read (`std::fs::copy`), never appended in place, so published bundles stay immutable and the caller's atomic tmp-then-rename publish contract and the no-race-into-bundle property are unchanged.  The produced bundle still round-trips through `read_records_from_pack` / `read_certs_from_pack` as a contiguous `0..=N` set.

## Changes

- [x] `EpochDbMessage::FirstMissingHistoricalCert` variant + actor arm;  public `first_missing_historical_cert` is now one round-trip and fails closed on a dead actor
- [x] `Inner::certified_watermark` with `epoch_certified` / `first_missing_historical_cert` / `seed_certified_watermark` (seeded in `open_append` after the heal step)
- [x] `EpochRecordDb::export_incremental_bundle` / `try_append_previous_bundle`;  file-scope helpers `pack_entry_chain`, `append_to_copied_pack`, `append_bundle_increment`;  `EpochDbError::BundleValidation`
- [x] `export_epoch_state` (`close_epoch.rs`) derives the previous published bundle's pack paths and calls `export_incremental_bundle`;  tmp-dir handling, cert wait, and atomic rename untouched
- [x] Tests: watermark resume + hole-wait + backfill self-heal, watermark seeding at reopen, incremental append with sentinel positive control, fallback when the previous bundle is absent, and a four-row rejection table (stale / padded / badcert / wronglast), each row pinned to one validation clause

## Testing

- `cargo +1.94 test -p tn-storage`: 182 tests + 1 doc-test green, including the 5 new test fns
- Every new test confirmed by mutation (seen to fail as a real assertion failure, then the original bytes restored): 8 single-line mutants, each killed by exactly its targeted test or table row
  - M1 watermark jumps holes -> `certified_watermark_resumes_and_waits_at_hole`
  - M2 seeding no-op -> `certified_watermark_seeded_at_open`
  - M3 incremental path never taken -> sentinel positive control in `export_incremental_bundle_appends_to_previous_bundle`
  - M4 fallback removed -> `export_incremental_bundle_falls_back_without_previous_bundle`
  - M5 entry-count clause dropped -> `padded` row
  - M6 cert-predicate dropped -> `badcert` row
  - M7 record-predicate dropped -> `wronglast` row
  - M8 whole validation gate dropped -> `stale` row
- Attest gates all green, run twice (before and after a final doc-comment pass): nightly `fmt -- --check`, nightly `clippy --workspace -- -D warnings`, nightly `clippy --workspace --all-features -- -D warnings`, and isolated-target `cargo +1.94 test --no-run --workspace`, each exit 0;  `cargo +1.94 test -p tn-node` also green (47 + 20 tests)

Closes #1078.

### DIFF
--- a/crates/node/src/manager/node/close_epoch.rs
+++ b/crates/node/src/manager/node/close_epoch.rs
@@ -332,8 +332,10 @@ where
     /// `final_state`.
     ///
     /// Unlike a plain copy of the live shared `epochs.pack` / `epoch_certs.pack`, the records/certs
-    /// bundle is a bounded `0..=N` selection built through the records-DB actor in the completion
-    /// task, so a later epoch appending to the live packs cannot race into the export. The task
+    /// bundle is a bounded `0..=N` set assembled in the completion task, so a later epoch appending
+    /// to the live packs cannot race into the export. In the common case it is built by copying the
+    /// previous boundary's published bundle and appending only epoch N's record and cert; when no
+    /// previous bundle validates, it is rebuilt in full through the records-DB actor. The task
     /// also waits a bounded time for epoch N's own certificate (only aggregated at the next epoch's
     /// start) and fails the export without it, so an importer never has to store the tip record
     /// unverified.
@@ -400,6 +402,16 @@ where
         let consensus_chain = self.consensus_chain.clone();
         let consensus_bus = self.consensus_bus.clone();
 
+        // The previous boundary's published bundle, if any: lets the completion task build the
+        // records/certs bundle by copy + single append instead of a full 0..=N rebuild. It can be
+        // legitimately absent (first epoch, a skipped or failed prior export, operator pruning);
+        // the storage layer then falls back to the full rebuild. Published bundles are only ever
+        // read, never appended to in place.
+        let prev_bundle = epoch.checked_sub(1).map(|prev_epoch| {
+            let prev_dir = export_root.join(format!("epoch-{prev_epoch}"));
+            (prev_dir.join("epoch_records"), prev_dir.join("epoch_certs"))
+        });
+
         // Export into a temp sibling and atomically rename it into place on success, so external
         // tooling only ever observes a complete `epoch-{N}` directory. Clear any leftover temp for
         // THIS epoch first: a successful export renames its temp to the final dir (and the
@@ -464,12 +476,16 @@ where
                                 return Ok(());
                             }
 
-                            // write the bounded 0..=N records+certs from the actor, so a later epoch
-                            // appending to the live packs cannot race into this bundle.
+                            // write the bounded 0..=N records+certs bundle: extend the previous
+                            // boundary's published bundle with epoch N's entries when it
+                            // validates, rebuilding in full through the actor otherwise. either
+                            // way the set is bounded, so a later epoch appending to the live
+                            // packs cannot race into this bundle.
                             if let Err(e) = consensus_chain
                                 .epochs()
-                                .export_bounded_bundle(
+                                .export_incremental_bundle(
                                     epoch,
+                                    prev_bundle,
                                     &tmp_dir.join("epoch_records"),
                                     &tmp_dir.join("epoch_certs"),
                                 )
--- a/crates/storage/src/epoch_records.rs
+++ b/crates/storage/src/epoch_records.rs
@@ -22,7 +22,7 @@ use tokio::sync::{
     mpsc::{self, Receiver, Sender},
     oneshot, watch,
 };
-use tracing::error;
+use tracing::{debug, error};
 
 use crate::archive::{
     data_file::create_dir_synced,
@@ -59,6 +59,9 @@ enum EpochDbMessage {
     ContainsRecordDigest(EpochDigest, oneshot::Sender<bool>),
     /// Return the latest (highest epoch) [`EpochRecord`] stored, if any.
     LatestRecord(oneshot::Sender<Option<EpochRecord>>),
+    /// Return the first epoch in `0..tip` whose record or certificate is not yet stored,
+    /// resuming from (and advancing) the actor's contiguous-certified-prefix watermark.
+    FirstMissingHistoricalCert(Epoch, oneshot::Sender<Option<Epoch>>),
     /// Flush all pending writes to disk.
     Persist(oneshot::Sender<Result<(), EpochDbError>>),
     Shutdown,
@@ -129,6 +132,9 @@ fn run_db_loop(
             }
             EpochDbMessage::LatestRecord(tx) => {
                 let _ = tx.send(inner.latest_record());
+            }
+            EpochDbMessage::FirstMissingHistoricalCert(tip_epoch, tx) => {
+                let _ = tx.send(inner.first_missing_historical_cert(tip_epoch));
             }
             EpochDbMessage::Persist(tx) => {
                 let _ = tx.send(inner.persist());
@@ -504,22 +510,25 @@ impl EpochRecordDb {
     }
 
     /// Scan the historical epochs `0..tip_epoch` and return the first whose certificate (or record)
-    /// is not yet stored, or `None` if every one has a cert. Cheap: record + cert actor lookups, no
-    /// state I/O — the same per-epoch queries
-    /// [`export_bounded_bundle`](Self::export_bounded_bundle) already does, run before the
-    /// export's plain-state walk so it can skip early when a required historical cert is
-    /// permanently missing (e.g. a network-wide failed-quorum epoch no peer can supply) instead
-    /// of walking the whole state and then failing.
+    /// is not yet stored, or `None` if every one has a cert. Cheap: a single actor round-trip that
+    /// resumes from the actor's contiguous-certified-prefix watermark, so per-boundary work is
+    /// bounded by the epochs newly certified since the previous scan rather than by the chain's
+    /// age. Run before the export's plain-state walk so it can skip early when a required
+    /// historical cert is permanently missing (e.g. a network-wide failed-quorum epoch no peer can
+    /// supply) instead of walking the whole state and then failing.
     ///
     /// `tip_epoch` itself is EXCLUDED: the exported tip's own cert is only aggregated at the next
     /// epoch's start, so it is normally still pending at export time and is waited for separately.
     pub async fn first_missing_historical_cert(&self, tip_epoch: Epoch) -> Option<Epoch> {
-        for epoch in 0..tip_epoch {
-            if !matches!(self.get_epoch_by_number(epoch).await, Some((_, Some(_)))) {
-                return Some(epoch);
-            }
+        let (tx, rx) = oneshot::channel();
+        // On a dead or dying actor, report epoch 0 as unconfirmed (whenever any historical epoch
+        // exists) so the caller skips the export rather than proceeding blind; the per-epoch
+        // lookups this scan replaced degraded the same way.
+        if self.tx.send(EpochDbMessage::FirstMissingHistoricalCert(tip_epoch, tx)).await.is_ok() {
+            rx.await.unwrap_or_else(|_| (tip_epoch > 0).then_some(0))
+        } else {
+            (tip_epoch > 0).then_some(0)
         }
-        None
     }
 
     /// Retrieve the epoch record and certificate (if available) by record digest.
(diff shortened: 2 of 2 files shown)


grantkee (approved):
LGTM

PR #1088, opened 2026-08-05
tn-storage: propagate non-NotFound remove_file failures in  write_bounded_pack

### LINKED ISSUE #1080: write_bounded_pack ignores non-NotFound remove_file failures and appends onto a stale pack
## Summary

`write_bounded_pack` in `crates/storage/src/epoch_records.rs` promises a fresh pack file (its doc comment: "a leftover file ... would be appended to, not replaced"), but it enforces that contract with a discarded-result removal:

https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/storage/src/epoch_records.rs#L1019-L1023

```rust
    // Enforce the "fresh" contract: a leftover file (e.g. from a prior failed export attempt) would
    // be appended to, not replaced. NotFound is the normal case and is ignored.
    let _ = std::fs::remove_file(path);
    let mut pack =
        Pack::<V>::open(path, uid_idx, false, PackCompression::ZStd, EPOCH_PACK_VERSION)?;
```

`NotFound` is the normal case and is rightly ignored, but every other failure (for example `PermissionDenied` on the export directory) is swallowed too.  When removal fails and the stale file survives, the subsequent `Pack::open(..., false, ...)` appends onto it rather than replacing it, producing a doubled pack.

Severity: Low.  The failure path is not reachable through current in-tree callers (details below);  the value of the fix is making the function's own invariant self-contained instead of relying on a caller-side guard.

## Problem

The append mechanism, step by step (all links at `main`, commit `47bd71f8`):

1. [`epoch_records.rs#L1021`](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/storage/src/epoch_records.rs#L1021): `let _ = std::fs::remove_file(path);` discards any error, including non-`NotFound` ones.
2. `Pack::open` with `ro: false` reaches [`DataFile::open`](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/storage/src/archive/data_file.rs#L64-L77), which opens the file with `.append(!read_only)`, i.e. append mode for writers.
3. [`PackInner::open_data_file`](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/storage/src/archive/pack.rs#L314-L341) sees a non-empty file and loads the existing header instead of writing a fresh one.  Validation ([`load_header_from_buffer`](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/storage/src/archive/pack.rs#L530-L549) plus the version/appnum checks in `open_data_file`) covers magic, CRC32, uid, version, and appnum only.  A stale pack from a prior export attempt has the same `uid_idx`, so it passes every check, and the new records are appended after the old ones.

The result is a records (or certs) pack containing two full copies of the epoch run, silently, from a code path whose comment and doc comment both state the file must be fresh.

## Why this is currently unreachable (hence Low)

The only external caller chain is the epoch snapshot export in `close_epoch.rs`, which already applies exactly the discipline this issue asks for, one level up: it clears the target `epoch-{N}.tmp` with `remove_dir_all`, tolerates `NotFound`, and on any other error logs and skips the export entirely ([`close_epoch.rs#L403-L419`](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/node/src/manager/node/close_epoch.rs#L403-L419)) before [`export_bounded_bundle`](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/node/src/manager/node/close_epoch.rs#L471) ever runs.  So today, `remove_file` inside `write_bounded_pack` can only fail with `NotFound` in practice.

On the consumer side, the CLI importer would also reject a doubled records pack up front: it requires the bundle's records to be contiguous from epoch 0 ([`telcoin-network-cli/src/db.rs#L469-L478`](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/telcoin-network-cli/src/db.rs#L469-L478)), and a doubled pack restarts at epoch 0 halfway through.

Both are caller-side and consumer-side guards for what is a local invariant of `write_bounded_pack`.  A future caller that does not pre-clear its target directory inherits the doubled-pack hazard silently.

## Proposed fix

Propagate non-`NotFound` removal failures instead of discarding them:

```rust
    // Enforce the "fresh" contract: a leftover file (e.g. from a prior failed export attempt) would
    // be appended to, not replaced. NotFound is the normal case; any other removal failure means the
    // stale file may survive, so surface it instead of appending a doubled pack.
    std::fs::remove_file(path)
        .or_else(|e| (e.kind() == std::io::ErrorKind::NotFound).then_some(()).ok_or(e))?;
```

The `?` lands on the existing [`From<std::io::Error>` conversion](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/storage/src/epoch_records.rs#L1002) into `EpochDbError::IO`, so no new error plumbing is needed.  If the explicit three-way shape is preferred, the tmp-dir guard at `close_epoch.rs#L412-L419` is the in-repo precedent to mirror.

## Testing

The existing regression test [`write_bounded_pack_overwrites_stale_file`](https://github.com/Telcoin-Association/telcoin-network/blob/47bd71f84d42cf12cff2e36b39d5d48250e35fd1/crates/storage/src/epoch_records.rs#L1184) covers the successful-removal path (stale file present, removal succeeds, fresh pack written) and stays green.  The new error branch can be exercised on Unix by making the pack's parent directory read-only (`set_permissions` with mode `0o555`) so `remove_file` fails with `PermissionDenied`, then asserting `Err(EpochDbError::IO(_))` and that the stale file's bytes are untouched;  such a test needs a non-root, non-Windows guard, so it may be reasonable to land the fix with the existing test coverage alone.

MavenRain:
Closes #1080.

## Problem

`write_bounded_pack` in `crates/storage/src/epoch_records.rs` promises a fresh pack file (its doc comment: a leftover file "would be appended to, not replaced"), but enforces that contract with a discarded-result removal:

```rust
let _ = std::fs::remove_file(path);
```

`NotFound` is the normal case and is rightly ignored, but every other failure (for example `PermissionDenied` on the export directory) is swallowed too. When removal fails and the stale file survives, the subsequent `Pack::open(.., ro: false, ..)` opens in append mode, sees a non-empty file, and loads the existing header instead of writing a fresh one. A stale pack from a prior export attempt carries the same `uid_idx`, so it passes every header check (magic, CRC32, uid, version, appnum), and the new records land after the old ones: a silently doubled pack, from a code path whose comment and doc comment both state the file must be fresh.

No attacker required: the trigger is operator-side filesystem state (a permissions problem on the export directory), and the blast radius is a doubled export bundle, not consensus state. The path is also not reachable through current in-tree callers . . . the epoch snapshot export in `close_epoch.rs` pre-clears its `epoch-{N}.tmp` target (tolerating only `NotFound`) before `export_bounded_bundle` runs, and the CLI importer rejects a doubled records pack up front (records must be contiguous from epoch 0). But both of those are caller-side and consumer-side guards for what is a local invariant of `write_bounded_pack`; a future caller that does not pre-clear its target directory inherits the doubled-pack hazard silently. Hence Low severity, and the value of the fix is making the function's own invariant self-contained.

## Fix

- [x] `crates/storage/src/epoch_records.rs`: propagate non-`NotFound` removal failures instead of discarding them:

  ```rust
  std::fs::remove_file(path)
      .or_else(|e| (e.kind() == io::ErrorKind::NotFound).then_some(()).ok_or(e))?;
  ```

  The `?` lands on the existing `From<io::Error>` conversion into `EpochDbError::IO`, so no new error plumbing is needed.
- [x] The doc comment on `write_bounded_pack` now states the error contract: a removal failure other than `NotFound` surfaces as `EpochDbError::IO` rather than being ignored.

## Testing

- New regression test `write_bounded_pack_surfaces_remove_file_failure` (`#[cfg(unix)]`): writes a stale pack, sets the parent directory to mode `0o555` so `remove_file` fails with `PermissionDenied`, and asserts both `Err(EpochDbError::IO(_))` and that the stale pack's bytes are untouched. A probe file created before the chmod detects root (directory permissions are bypassed there, so the error branch is unreachable); in that case the test prints a loud SKIPPED marker instead of passing vacuously. On this repo's CI (hosted `ubuntu-latest` runners, non-root) the error branch is exercised for real.
- Mutation-confirmed: reverting the fix to `let _ = std::fs::remove_file(path);` makes the new test fail on `removal failure must surface, not append`; the restored fix passes. `Pack::open` failures map to the distinct `EpochDbError::Open` variant, so the `IO` assertion cannot pass vacuously off an unrelated open error.
- Existing regression test `write_bounded_pack_overwrites_stale_file` (the successful-removal path) stays green.
- `cargo +1.94 test -p tn-storage --lib`: full crate suite green (one unrelated `composite_db::test::test_compositedb_clear` flake under a parallel full-suite run re-ran green in isolation; that module is untouched here).
- `cargo +nightly-2026-03-20 fmt -- --check` and `clippy -p tn-storage --all-features` green; attest-shape test-build `cargo +1.94 test --no-run --workspace -j 2` in an isolated target dir green.

### DIFF
--- a/crates/storage/src/epoch_records.rs
+++ b/crates/storage/src/epoch_records.rs
@@ -1008,17 +1008,21 @@ impl From<io::Error> for EpochDbError {
 ///
 /// Removes any pre-existing file at `path` first, so a re-run overwrites rather than appends —
 /// `Pack::open` opens read-write in append mode, so writing over a leftover file would otherwise
-/// prepend prior-attempt records. Then opens the pack (which creates the file and writes its
-/// header), appends every value in order, and commits so the file is complete on disk. Keeps the
-/// sentinel tags and codec inside this crate so the export bundle stays readable by
-/// `read_records_from_pack` / `read_certs_from_pack`.
+/// prepend prior-attempt records. A removal failure other than `NotFound` is surfaced as
+/// [`EpochDbError::IO`] rather than ignored, since the stale file would otherwise survive and be
+/// appended onto. Then opens the pack (which creates the file and writes its header), appends
+/// every value in order, and commits so the file is complete on disk. Keeps the sentinel tags and
+/// codec inside this crate so the export bundle stays readable by `read_records_from_pack` /
+/// `read_certs_from_pack`.
 fn write_bounded_pack<V>(path: &Path, uid_idx: u64, values: &[V]) -> Result<(), EpochDbError>
 where
     V: std::fmt::Debug + serde::Serialize + serde::de::DeserializeOwned,
 {
     // Enforce the "fresh" contract: a leftover file (e.g. from a prior failed export attempt) would
-    // be appended to, not replaced. NotFound is the normal case and is ignored.
-    let _ = std::fs::remove_file(path);
+    // be appended to, not replaced. NotFound is the normal case; any other removal failure means
+    // the stale file may survive, so surface it instead of appending a doubled pack.
+    std::fs::remove_file(path)
+        .or_else(|e| (e.kind() == io::ErrorKind::NotFound).then_some(()).ok_or(e))?;
     let mut pack =
         Pack::<V>::open(path, uid_idx, false, PackCompression::ZStd, EPOCH_PACK_VERSION)?;
     for value in values {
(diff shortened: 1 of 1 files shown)


grantkee (approved):
LGTM!

PR #1093, opened 2026-08-05
tn 942 prng pin

MavenRain:
- **docs(consensus): address review feedback on reputation evaluation (#983)**
- **fix(consensus): pin leader swap PRNG to ChaCha12Rng (#942)**


grantkee (approved):
LGTM - great idea

Issue #1065, opened 2026-07-29
`tn-storage`: `EpochRecordDb::persist()` can return `Ok(())` after a failed background `save_record`, and the error can be consumed by an unrelated caller


MavenRain:
`tn-storage`: `EpochRecordDb::persist()` can return `Ok(())` after a failed background `save_record`, and the error can be consumed by an unrelated caller

## Problem

`EpochRecordDb` is a handle in front of a background thread.  Writes are fire-and-forget: `save_record` enqueues a message and returns immediately, so a write that fails on disk cannot be reported to the caller that made it.  The documented compensating mechanism is stated at `crates/storage/src/epoch_records.rs:70`:

```
/// Errors from background writes are surfaced on the next call via [`get_error`].
```

`persist()` is the call that is supposed to make a write durable and therefore the call that most needs that guarantee.  It does not provide it.  There are two independent defects in the mechanism.

### 1.  `persist()` samples the error before enqueueing, not after completing

```rust
// crates/storage/src/epoch_records.rs:387
pub async fn persist(&self) -> Result<(), EpochDbError> {
    self.get_error()?;
    let (tx, rx) = oneshot::channel();
    let _ = self.tx.send(EpochDbMessage::Persist(tx)).await;
    rx.await.map_err(|_| match &*self.error.borrow() {
        Some(e) => e.clone(),
        None => EpochDbError::ReceiveFailed,
    })?
}
```

The `get_error()?` on entry observes only failures the background thread had already recorded at the moment `persist` was called.  The messages are processed FIFO, so this sequence is entirely ordinary:

1.  A caller runs `save_record(record)`.  It checks `get_error()`, finds nothing, enqueues `SaveRecord`, and returns `Ok(())`.
2.  The caller runs `persist()`.  If the background thread has not yet drained `SaveRecord`, the entry `get_error()?` still finds nothing, and `Persist` is enqueued behind `SaveRecord`.
3.  The background thread processes `SaveRecord`, `inner.save_record` fails, and the error goes into the watch channel (`:97-101`).
4.  The background thread then processes `Persist` and runs `inner.persist()` (`:133-135`).

`Inner::persist` (`:781-790`) only commits and syncs the underlying tables:

```rust
fn persist(&mut self) -> Result<(), EpochDbError> {
    if !self.records.read_only() {
        self.records.commit().map_err(|e| EpochDbError::PersistError(e.to_string()))?;
[...]

PR #1094, opened 2026-08-05
fix(storage): report a queued write failure from EpochRecordDb::persist (#1065)

MavenRain:
## Summary

`EpochRecordDb::persist()` could return `Ok(())` after a background write had failed, so the epoch-close path could complete and report success for an epoch record that never reached disk.  This makes the flush report that failure.  Fixes the first defect in #1065.

## Problem

`EpochRecordDb` is a handle in front of an actor thread.  Writes are fire-and-forget: `save_record` (`epoch_records.rs:323-330`) enqueues and returns, so a write that fails on disk cannot be reported to the caller that made it.  The compensating mechanism is documented on the handle: "Errors from background writes are surfaced on the next call via `get_error`."

`persist()` is the call that is supposed to make a write durable, and it did not provide that.  It samples the error slot with `get_error()?` *before* enqueueing, and the `Persist` arm forwarded `Inner::persist()` (`:929-937`), which only commits and syncs the tables and never consults the error watch.  So:

1.  A caller runs `save_record(record)`.  It finds no error, enqueues `SaveRecord`, returns `Ok(())`.
2.  The caller runs `persist()`.  The actor has not drained `SaveRecord` yet, so the entry `get_error()?` finds nothing and `Persist` queues behind it.
3.  The actor processes `SaveRecord`, `inner.save_record` fails, the error goes into the watch.
4.  The actor processes `Persist`, commits tables the record was never appended to, and replies `Ok(())`.

The `self.error.borrow()` in `persist` does not cover this: it sits inside the `map_err` closure, so it runs only if the oneshot receive itself fails, meaning the actor died.  On the ordinary path the watch is never read.

This is not schedule-dependent.  The actor drains a single FIFO `mpsc::channel(1000)` (`:208`, `:89`), so a save queued ahead of a flush is *guaranteed* to be processed first.

The epoch-close path issues exactly that pair back to back, and documents that it depends on the flush:

```rust
// close_epoch.rs:343-347
self.consensus_chain.epochs().save_record(epoch_rec.clone()).await?;
// the cert-quorum path persists asynchronously; a crash before that flush would lose
// the record just saved — fatal for epoch 0, which has no peer-fetch anchor — so
// flush it durably here before publishing
self.consensus_chain.epochs().persist().await?;
```

Its doc comment (`:228-233`) states the guarantee directly: "The record is flushed durably to disk before returning ... fatal for epoch 0, which has no peer-fetch anchor to heal from."

The always-on epoch-record collector (`state-sync/epoch.rs:150`, 5s cadence, rescans from epoch 0) will usually back-fill a lost record from peers, so this is a silently violated durability contract and a peer-dependent recovery window rather than guaranteed permanent loss.  Epoch 0 is the case with no peer anchor.

## Fix

The `Persist` arm takes the pending error and folds it into its reply:

```rust
EpochDbMessage::Persist(tx) => {
    let pending = tx_error.send_replace(None);
    let flushed = inner.persist();
    let _ = tx.send(pending.map_or(flushed, Err));
}
```

The flush still runs, so the only behavioural change is that a previously silent failure reaches the caller that asked for durability.  Every caller already propagates the error with `?` or logs it.

Taking the error here also gives it an owner: the flush is the call that wanted the guarantee, so it consumes the failure rather than leaving it to be misattributed to an arbitrary later caller.

I also documented the clear-on-read semantics of `get_error` on the handle itself, and pointed at `peek_error` (added in #985) for the non-consuming check.

## Scope

This addresses defect 1 of #1065 only.  Defect 2 — `get_error` clearing on read, so a failure can be consumed by an unrelated write-side caller on the shared handle — is left alone deliberately.  Fixing it properly means either giving the save messages their own `oneshot<Result>` replies, as `ConsensusPack::save_consensus_output` (`consensus_pack.rs:351`) already does, or a sticky latch like the #975 fix in `layered_db.rs`.  Both change observable behaviour for existing callers, so that looks like a maintainer call rather than something to fold into this PR.

Worth noting separately: `CertificatePack::persist` (`certificate_pack.rs:188-196`) has the identical sample-before-enqueue shape and would need the same treatment.

## Testing

New regression test `persist_reports_a_write_that_failed_while_the_flush_was_queued`.  It queues a save the actor rejects (epoch 5 on an empty db is out of order) and a flush behind it, both sent straight to the channel so the handle-side guards do not reject the record before it reaches the actor, then asserts the flush reports the earlier failure.  Single-consumer FIFO ordering makes this deterministic rather than timing-dependent.

`cargo test -p tn-storage --lib` is green: 178 passed, 0 failed.

Confirmed by mutation: reverting the arm to `let _ = tx.send(inner.persist());` gives 177 passed, 1 failed, and the single failure is the new test.  So it is bound to this change rather than passing vacuously.

### DIFF
--- a/crates/storage/src/epoch_records.rs
+++ b/crates/storage/src/epoch_records.rs
@@ -67,7 +67,10 @@ enum EpochDbMessage {
 /// Handle to the epoch records database.
 ///
 /// Operations are dispatched to a background thread that owns the file handles.
-/// Errors from background writes are surfaced on the next call via [`get_error`].
+/// Errors from background writes are surfaced on the next call via [`get_error`], which clears
+/// the slot as it reads, so exactly one subsequent caller observes a given failure. Use
+/// [`peek_error`] to check without consuming. [`persist`] is the durability barrier: it reports
+/// any earlier write failure even if that write was still queued when the flush was requested.
 #[derive(Debug, Clone)]
 pub struct EpochRecordDb {
     /// Channel to send commands to the background thread.
@@ -131,7 +134,13 @@ fn run_db_loop(
                 let _ = tx.send(inner.latest_record());
             }
             EpochDbMessage::Persist(tx) => {
-                let _ = tx.send(inner.persist());
+                // Fold a write that failed while this persist was queued into the reply.
+                // `persist()` samples the error slot before enqueueing, and writes are
+                // fire-and-forget, so a save that fails after that sample but before this arm
+                // would otherwise be acknowledged as a successful flush.
+                let pending = tx_error.send_replace(None);
+                let flushed = inner.persist();
+                let _ = tx.send(pending.map_or(flushed, Err));
             }
             EpochDbMessage::Shutdown => {
                 let _ = inner.persist();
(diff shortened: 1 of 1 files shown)


grantkee (approved):
LGTM!

PR #1095, opened 2026-08-05
cli: refuse a faucet-compiled binary on canonical Telcoin mainnet

### LINKED ISSUE #1066: `telcoin-network-cli`: the `faucet` cargo feature has no chain-id guard, so a faucet-enabled build can join a non-testnet chain with a divergent TEL precompile dispatch table
`telcoin-network-cli`: the `faucet` cargo feature has no chain-id guard, so a faucet-enabled build can join a non-testnet chain with a divergent TEL precompile dispatch table

## Problem

The node has a runtime guard that refuses to start when the compiled feature set does not match the chain being joined, at `crates/telcoin-network-cli/src/node.rs:158-173`.  It checks exactly one feature, `adiri`, against exactly one chain id, 2017:

```rust
if tn_config.genesis().config.chain_id == 2017 {
    // If we are trying to start an Adiri node without the adiri feature flag then error
    return Err(eyre::eyre!(
        "Must compile with adiri feature flag in order to connect to adiri (testnet)!"
    ));
}
#[cfg(feature = "adiri")]
if tn_config.genesis().config.chain_id != 2017 {
    return Err(eyre::eyre!(
        "Must NOT compile with adiri feature flag when connecting to non-adiri (testnet) networks!"
    ));
}
```

`faucet` is a separate, independently selectable feature.  It is not implied by anything and it implies nothing:

```toml
# crates/tn-reth/Cargo.toml:84
faucet = []
# crates/tn-reth/Cargo.toml:86
adiri = ["faucet", "tn-types/adiri"]
```

`adiri` turns `faucet` on, but not the reverse.  `crates/node/Cargo.toml:59` re-exports it as `faucet = ["tn-reth/faucet"]`.  So `cargo build --features faucet`, without `adiri`, produces a binary that passes the guard above for every chain id other than 2017, including mainnet's 487.  There is no other check: outside `Cargo.toml` and the precompile module itself, the string `faucet` does not appear anywhere in the CLI or node startup path.

### Why that matters: the feature changes consensus-relevant execution

`faucet` is not additive.  It swaps the dispatch table of the TEL precompile at its canonical address, so the same transaction has different results on a faucet build and a default build.

The `mint` entry point is replaced rather than extended.  The re-export switches on the feature (`crates/tn-reth/src/evm/tel_precompile/mod.rs:64-75`):

```rust
/// mainnet `mint(uint256)` call type (timelocked, governance-only).
#[cfg(not(feature = "faucet"))]
pub use burnable::mintCall;
/// Faucet `mint(address, uint256)` call type (instant, role-gated).
#[cfg(feature = "faucet")]
pub use faucet::mintCall;
```

and so does the dispatch arm (`:130-133`):

```rust
#[cfg(not(feature = "faucet"))]
burnable::mintCall::SELECTOR => {
    burnable::handle_mint(&mut input.internals, calldata, input.caller, input.gas)
}
```

The two `mint` signatures have different selectors.  The role-management arms are likewise feature-gated (`:145-148`).  The consequence is a hard state divergence between two nodes on the same chain:

- calldata for `mint(uint256)` executes the timelocked governance mint on a default build and hits an unknown selector on a faucet build
- calldata for `mint(address,uint256)` mints instantly on a faucet build and hits an unknown selector on a default build

One of the two nodes reverts where the other mutates TEL supply.  They compute different state roots for the same block, which is a chain split rather than a local misconfiguration.  A faucet build joining a live non-testnet chain also exposes an instant mint to any address holding the mint role, where the default build would require the timelock at `burnable::TIMELOCK_DURATION`.

The access control inside the faucet module is not the issue and appears correct.  `handle_mint_faucet` gates on `has_mint_role(caller)` (`crates/tn-reth/src/evm/tel_precompile/faucet.rs:72`), and `handle_grant_mint_role` and `handle_revoke_mint_role` both gate on `caller != GOVERNANCE_SAFE_ADDRESS` (`:142`, `:168`).  The problem is that the feature is reachable on the wrong chain at all, not that it is unguarded once reached.

### Why this is easy to hit by accident

Cargo unifies features across a build graph.  Any dependency edge that enables `tn-reth/faucet`, including the existing `faucet = ["tn-reth/faucet"]` in `crates/node/Cargo.toml`, turns it on for every crate in that build.  An operator building the node as part of a larger workspace, or with a tool that enables features transitively, can end up with a faucet-enabled binary without ever typing the flag, and nothing at startup tells them.  The `adiri` guard exists because this class of mistake was already considered worth catching;  `faucet` is the same hazard with no guard.

## Impact

I am filing this as a footgun rather than a live vulnerability.  Reaching it requires an operator to build with a feature the repository documents as testnet-only in three places, so this is not an attack an outsider can mount.  What makes it worth closing is that the failure is silent, the blast radius is a consensus split rather than a crashed node, and feature unification means the operator may not have chosen the feature deliberately.

## Solution

Mirror the existing `adiri` guard for `faucet` in the same block at `crates/telcoin-network-cli/src/node.rs:158-173`, so a faucet-enabled binary refuses to start against any chain where the faucet precompile is not intended:

```rust
#[cfg(feature = "faucet")]
if tn_config.genesis().config.chain_id != 2017 {
    return Err(eyre::eyre!(
        "Must NOT compile with faucet feature flag when connecting to non-testnet networks!"
    ));
}
```

If devnet chain ids other than 2017 are meant to run the faucet, the condition should be an allow-list of those ids rather than a single equality.

Two alternatives, in case you prefer them:

1.  Make the relationship explicit in the manifest by removing standalone `faucet` and folding its contents into `adiri`, so the existing guard covers it by construction.  This is the smallest change but it removes the ability to build a faucet devnet that is not adiri.
2.  Add a `compile_error!` that fires when `faucet` is enabled without `adiri`.  This catches it earlier than startup, though it cannot see the chain id, so it is strictly weaker than the runtime guard for the mainnet case.

Option 1 or the runtime guard both look preferable to the status quo;  I have not sent a patch since the right answer depends on whether a non-adiri faucet devnet is a configuration you intend to support.

## Notes

Found during a read-only review of `origin/main` at `5dbb764e`.  I have not run a faucet-enabled binary against a non-testnet chain to observe the divergence;  the claim rests on the feature-gated dispatch arms and re-exports cited above, which show the two builds cannot agree on the `mint` selector.

MavenRain:
Closes #1066.

## Problem

The node refuses to start when the compiled feature set does not match the chain being joined, but the check at `crates/telcoin-network-cli/src/node.rs` covers exactly one feature, `adiri`, against exactly one chain id, 2017.  The `faucet` feature is independently selectable (`tn-reth` defines `faucet = []`, and `bin/telcoin-network` propagates it through `tn-node/faucet` without touching the CLI crate), and it is not additive:  it swaps the TEL precompile's `mint` dispatch, replacing the timelocked, governance-only `mint(uint256)` with the instant, role-gated `mint(address,uint256)`.  The two signatures have different selectors, so a faucet build and a default build cannot agree on the result of any TEL `mint` transaction.  A faucet-enabled binary joining mainnet therefore computes different state roots than the rest of the network for the same block . . . a consensus split, not a local misconfiguration.  No attacker is required:  cargo feature unification can switch the feature on transitively without the operator ever typing it, and nothing at startup says so.

## Fix

- [x] `crates/tn-reth/src/lib.rs`:  export `pub const FAUCET_ENABLED: bool = cfg!(feature = "faucet")`.  The constant lives in the crate that owns the feature, so cargo feature unification makes it the ground truth for the whole binary no matter which crate in the build graph enabled `tn-reth/faucet`.  A `#[cfg]` in the CLI crate could not see this (the CLI crate has no `faucet` feature, and `bin/telcoin-network` routes the flag through `tn-node` only).
- [x] `crates/telcoin-network-cli/src/node.rs`:  add `validate_faucet_build(FAUCET_ENABLED, tn_config.genesis())` immediately after the existing `adiri` guard.  It errors iff the build is faucet-enabled and the loaded genesis computes the same genesis block hash as the canonical embedded `MAINNET_GENESIS`.
- [x] `crates/telcoin-network-cli/src/db.rs`:  run the same guard in `db load-state` before any datadir mutation;  it is the other CLI path that resolves a `--chain mainnet` config, and it should not prepare mainnet chain data under a faucet build either.

The check keys on the genesis block hash (the sealed genesis header:  header fields plus the state root computed from `alloc`, built by the same chain-spec construction the node boots with), which is the identity peers actually agree on.  The two plausible alternatives are both wrong:

- Chain id:  the guard the issue literally proposed (refuse `chain_id != 2017` in faucet builds) would reject both supported faucet flows.  The docker-devnet scripts generate their genesis with chain id 487 (`etc/genesis.sh`), which is also mainnet's id, and the e2e faucet test runs the default 911329;  keying on mainnet's own 487 instead would still reject the devnet flow.
- Struct equality on `Genesis`:  the fork-config fields (`config`) sit outside the genesis header preimage, so a genesis differing from canonical mainnet only in such a field still computes mainnet's exact genesis block hash (and could join mainnet consensus) while comparing unequal.  Hash identity refuses it;  a config-only-divergence regression test pins this.

The comparison is cheap-first and attacker-independent:  it runs once at startup, and the embedded-genesis parse and hash only happen in faucet builds.  Adiri/testnet (2017) needs no new handling:  it is faucet-intended, and the existing `adiri` guard already rejects any non-adiri build on 2017 in both directions.

Out of scope, noted for follow-up:  the pre-existing `adiri` guard is still a compile-time `#[cfg]` with the same transitive-unification exposure this PR closes for `faucet`, and the guard runs at the CLI entry points rather than inside the `tn_node::launch_node` library boundary (both of that function's current callers route through `NodeCommand::execute`).

## Testing

- [x] 6 deterministic unit tests in `node.rs`:  a faucet build rejects the canonical mainnet genesis;  a default build accepts it;  a faucet build accepts `adiri_genesis()`;  a faucet build accepts a devnet-style genesis that reuses chain id 487 but differs in content (the docker-devnet shape a bare chain-id check would have broken);  a faucet build rejects a genesis diverging from mainnet only in fork config (same genesis hash yet struct-unequal, with the hash equality asserted as a positive control . . . the case struct equality would have missed);  and the compiled feature set gates the canonical genesis exactly (`FAUCET_ENABLED` wired as at the call site).
- [x] Every regression test confirmed by mutation, each killed by a test failure (not a compile error) with the tree restored byte-identical:  disabling the comparison fails the reject-canonical test;  ignoring the `faucet_enabled` flag fails the default-accepts test;  reverting hash identity to struct equality fails the config-divergence test;  substituting a chain-id comparison fails the devnet test.
- [x] `cargo +nightly fmt -- --check`, `cargo +1.94 clippy --workspace --all-targets`, and `cargo +1.94 test --no-run --workspace` green on the pinned toolchain in an isolated target dir.
- [x] `cargo +1.94 test -p telcoin-network-cli --lib node::tests`:  6 passed.

### DIFF
--- a/crates/telcoin-network-cli/src/db.rs
+++ b/crates/telcoin-network-cli/src/db.rs
@@ -5,7 +5,10 @@
 //! `MissingBatches` check and classifying each missing batch as Absent (a real data gap) vs
 //! Misordered (present, but in the wrong consensus-header group).
 
-use crate::{node::NamedChain, version::SHORT_VERSION};
+use crate::{
+    node::{validate_faucet_build, NamedChain},
+    version::SHORT_VERSION,
+};
 use clap::{Args, Parser, Subcommand};
 use comfy_table::{Cell, Row, Table as ComfyTable};
 use eyre::{bail, eyre};
@@ -188,6 +191,10 @@ impl DbLoadStateArgs {
             Some(NamedChain::MainNet) => Config::load_mainnet(&datadir, false, SHORT_VERSION)?,
             None => Config::load(&datadir, false, SHORT_VERSION)?,
         };
+
+        // A faucet-compiled binary must not prepare mainnet chain data either: same guard as
+        // the node command, refused before any datadir mutation.
+        validate_faucet_build(tn_reth::FAUCET_ENABLED, tn_config.genesis())?;
 
         // Reject a non-resumable bundle BEFORE `restore_pack` mutates the datadir, so a refusal
         // leaves the target untouched.
--- a/crates/telcoin-network-cli/src/node.rs
+++ b/crates/telcoin-network-cli/src/node.rs
@@ -9,7 +9,8 @@ use rayon::ThreadPoolBuilder;
 use std::{net::SocketAddr, path::PathBuf, sync::Arc, thread::available_parallelism};
 use tn_config::{Config, KeyConfig, TelcoinDirs as _};
 use tn_node::engine::TnBuilder;
-use tn_reth::{parse_socket_address, RethCommand, RethConfig};
+use tn_reth::{parse_socket_address, RethChainSpec, RethCommand, RethConfig, FAUCET_ENABLED};
+use tn_types::{Genesis, B256, MAINNET_GENESIS};
 use tokio::task::JoinHandle;
 use tracing::{debug, error, info};
 
@@ -171,6 +172,9 @@ impl<Ext: clap::Args + fmt::Debug> NodeCommand<Ext> {
                 "Must NOT compile with adiri feature flag when connecting to non-adiri (testnet) networks!"
             ));
         }
+        // Runtime value, not a #[cfg], so the check holds no matter which crate in the
+        // build graph enabled the faucet feature.
+        validate_faucet_build(FAUCET_ENABLED, tn_config.genesis())?;
         debug!(target: "cli", validator = ?tn_config.node_info.name, "tn datadir for node command: {tn_datadir:?}");
         info!(target: "cli", validator = ?tn_config.node_info.name, "config loaded");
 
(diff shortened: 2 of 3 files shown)


grantkee (approved):
Good idea

PR #1096, opened 2026-08-05
fix(worker-gateway): key per-IP rate limit buckets on a network prefix (#956)

### LINKED ISSUE #956: Per-IP rate limiter keyed on full client IP is bypassed by source-IP rotation (trivial over IPv6)
### Summary
The per-IP token-bucket limiter keys buckets on the full `IpAddr`.  Every first-seen IP receives a fresh, full bucket and is admitted, and once the entry table is full further new IPs are admitted untracked without spending a token.  A client that rotates its source address therefore never accumulates *spent* per-IP budget: each new address gets its own fresh full bucket, and any bucket it abandons refills and is reclaimed by the GC sweep, so no bucket it uses ever reaches its limit.  That client is admitted on essentially every request, leaving the global limiter as the only remaining cap.

### Location
- `bin/worker-gateway/src/ratelimit.rs:150` (`HashMap<IpAddr, Bucket>`)
- `bin/worker-gateway/src/ratelimit.rs:155-192` (`admit` / `admit_new_ip`)
- `bin/worker-gateway/src/ratelimit.rs:44` (`DEFAULT_MAX_PER_IP_ENTRIES = 100_000`)
- `bin/worker-gateway/src/ratelimit.rs:309` (peer keyed as full `IpAddr` via `info.0.ip()`)
- default rate: `bin/worker-gateway/src/cli.rs:150`

Line numbers re-verified against `origin/main` @ `48e00b4b` (2026-08-04).  The `ratelimit.rs` citations are stable, since that file's only commit is `88ccf6e8` (#908).  The `cli.rs` and `server.rs` numbers shifted after this issue was filed (#972, #1054, #1055) and are re-pinned here.

### Evidence
1. **First-seen IP admitted with a fresh full bucket.**  `ratelimit.rs:183-187`: when `buckets.len() < max_entries`, `Bucket::full(now, capacity)` sets `tokens = capacity` (`:105-107`), `try_admit` spends one and returns `true` (burst default 200), then inserts.  The first request from any IP is always admitted.
2. **Table full leads to untracked admit, no token spent.**  `ratelimit.rs:188-191` returns `true` with no insert and no token spend.  Pinned by `full_table_admits_untracked_new_ip` (`:438-446`) and self-documented at `:168-174`.
3. **Key is the full `IpAddr`, no prefix aggregation.**  `ratelimit.rs:150`;  the key flows unmasked from `info.0.ip()` (`:309`) through `check` to `admit` to `get_mut`/`insert`.  No `/64`, `/24`, or masking anywhere in the crate.
4. **Global evaluated first, `&&` short-circuits.**  `ratelimit.rs:245-253`: `global_ok` first, then `global_ok && ... per_ip.admit(...)`.  The single `GlobalLimiter` is the only aggregate limiter;  the `Semaphore` in `server.rs:212` (inside `accept_loop`, `server.rs:197`) is a connection-concurrency cap (`--max-connections`, default 500 at `cli.rs:99`), not a rate or subnet backstop.

Defaults (`cli.rs:150/155/160/165`, `ratelimit.rs:44`, test `edge_protection_defaults` `cli.rs:526-537`): per-IP 100/s (burst 200), global 3000/s (burst 6000), 100k max entries.

### Impact
A source with a fresh IP per request drains no per-IP bucket, so the per-IP limiter rejects nothing for that client;  the sustained ceiling becomes the global 3000/s.  The bypass does not require the table to be full: even against an empty map, rotation is admitted every time via points 1 and 3 above (point 2 only affects whether memory is spent).  Because global/per-IP is 3000/100 = 30, only about 30 distinct source IPs are needed to consume the whole global budget while staying under every per-IP limit;  a single IPv6 /64 (2^64 addresses) makes this trivial and single-use-per-IP, evading any per-IP detection.  Legitimate clients then receive `429` from the global limiter while their own per-IP buckets are healthy, defeating the module's stated goal ("stops any single source monopolizing that budget", `ratelimit.rs:4-6`) for any normal IPv6 client.

### Why Medium, not High
The global cap protects the worker (never more than 3000/s regardless of rotation), so this is fairness/availability among clients, not worker overload.  The trade-off is explicitly documented (`ratelimit.rs:168-174`) and memory is bounded (100k cap + 60s GC), so there is no unbounded-memory DoS.  IPv4 rotation needs genuinely routed addresses (stateful TCP cannot be source-spoofed), so a /24 yields at most 256 IPs;  IPv6 /64 rotation is fully practical.

### Proposed fix
Key the per-IP bucket on a network prefix so rotation within an allocation collapses to one bucket.  Mask the IP before keying (at `ratelimit.rs:309` or inside `admit`);  the `Bucket` logic, `check` short-circuit, and `gc` are unchanged, and this also improves the memory profile (one attacker maps to one bucket, making the untracked-admit path far less likely to trigger).

```rust
fn masked_key(ip: IpAddr, v4_prefix: u8, v6_prefix: u8) -> IpAddr {
    match ip {
        IpAddr::V4(v4) => {
            let bits = u32::from(v4);
            let mask = if v4_prefix == 0 { 0 } else { u32::MAX << (32 - v4_prefix) };
            IpAddr::V4(Ipv4Addr::from(bits & mask))
        }
        IpAddr::V6(v6) => {
            let bits = u128::from(v6);
            let mask = if v6_prefix == 0 { 0 } else { u128::MAX << (128 - v6_prefix) };
            IpAddr::V6(Ipv6Addr::from(bits & mask))
        }
    }
}
```

Recommended defaults: IPv6 `/64`, IPv4 `/32`, both configurable via new CLI flags mirroring the existing rate flags.

**Requires a human decision on IPv4 policy:** `/24` groups rotation within a subnet but can also group unrelated customers behind carrier NAT, so `/32` is the safe default and `/24` an opt-in.

Secondary (optional, lower priority): when the table is at capacity, charge first-seen IPs against a shared "overflow" bucket instead of admitting untracked;  with prefix aggregation this path rarely triggers.  Document the residual limitation for operators who terminate client identity upstream.

MavenRain:
Closes #956.

## Problem

The per-IP token bucket keyed on the full client address, so rotating the source address defeated it entirely.

`PerIpLimiter` stored `HashMap<IpAddr, Bucket>` and took its key straight from the connection's peer address.  A first-seen address takes the `admit_new_ip` path, which builds `Bucket::full(..)` and spends one of its `capacity` tokens, so the very first request from any address is always admitted.  A client that never reuses an address therefore never accumulates *spent* per-IP budget: each address gets its own fresh full bucket, and any bucket it abandons refills and is reclaimed by the GC sweep, so no bucket it touches ever reaches its limit.

The per-IP limiter was the only thing standing between one source and the whole gateway budget, so the sustained ceiling collapsed to the global limit.  At the shipped defaults (per-IP 100/s, global 3000/s) roughly 30 distinct addresses consume the entire global budget while every per-IP bucket stays healthy.  A single IPv6 /64 holds 2^64 addresses, which makes single-use-per-address rotation trivial and invisible to any per-address detection.  Legitimate clients then take the `429` from the global limiter while the module's stated goal, "stops any single source monopolizing that budget", is not met for any normal IPv6 client.

No attacker sophistication is required and nothing needs to be malformed: this is ordinary traffic from many addresses in one allocation.

## Root cause

The guard existed and was structurally correct.  Only its key was wrong.  Nothing masked the address between the peer lookup and the map key, so an allocation the operator thinks of as one client presented as up to 2^64 independent clients.

## Fix

Mask the client address to a network prefix before it becomes the bucket key, so rotation inside one allocation collapses onto a single bucket.

- `PrefixLen` is a validated newtype whose only constructors (`PrefixLen::v4` / `PrefixLen::v6`) return `Result`, so an out-of-range prefix cannot be built.
- `PrefixPolicy` holds the v4 and v6 lengths and exposes `key(IpAddr) -> IpAddr`.  It matches `IpAddr` exhaustively with no wildcard arm.
- `mask_v4` / `mask_v6` clear the host bits via `checked_shl(host_bits).unwrap_or(0)`.  The `checked_shl` is load-bearing rather than defensive: a prefix of 0 shifts by the full type width, which is an arithmetic overflow that panics in debug builds.  Conversions go through `u32::from` / `u128::from` / `Ipv4Addr::from` / `Ipv6Addr::from`, so there is no naked `as`.
- `Bucket`, `try_admit`, the `check` short-circuit and `gc` are logically unchanged.  The only behavioural change is which key the bucket hangs on.

Defaults are IPv6 `/64` and IPv4 `/32`, both settable with `--rate-limit-per-ip-v6-prefix` / `--rate-limit-per-ip-v4-prefix` (and the matching `WORKER_GATEWAY_*` env vars).

The IPv4 default of `/32` is deliberate and conservative.  It preserves today's exact IPv4 behaviour, so this change cannot newly group unrelated customers sitting behind one carrier NAT.  Operators who want `/24` aggregation opt into it.  IPv6 is where the bypass is actually cheap, and `/64` is the smallest block normally routed to a single customer, so that is where the default does the work.

An IPv4-mapped peer (`::ffff:a.b.c.d`) is unmapped and keyed under the IPv4 prefix.  Masking it as IPv6 would have been actively harmful: all mapped addresses share their top 96 bits, so under the `/64` default a dual-stack listener would have collapsed every IPv4 client onto the single key `::`, which is a self-inflicted denial of service strictly worse than the bug being fixed.

Residual limitation, stated in the README rather than left implicit: an attacker holding many distinct allocations can still rotate across them.  Prefix keying raises the cost from "one /64" to "one allocation per bucket", it does not make rotation free of cost or impossible.

## Changes

- [x] `bin/worker-gateway/src/ratelimit.rs` . . . `PrefixLen`, `PrefixLenError`, `PrefixPolicy`, `mask_v4` / `mask_v6`;  `PerIpLimiter` gains a `prefix` field and masks in `admit`;  module doc corrected to describe prefix keying instead of per-address keying
- [x] `bin/worker-gateway/src/cli.rs` . . . the two prefix flags, `Settings.rate_limit_prefix`, and `resolve_prefix_policy`, which mirrors the existing `resolve_rate_limit` shape and surfaces an out-of-range prefix as a startup error
- [x] `bin/worker-gateway/src/app.rs` . . . thread the policy through to `RateLimiters::new`
- [x] `bin/worker-gateway/src/server.rs` . . . two test call sites updated for the new argument
- [x] `bin/worker-gateway/README.md` . . . document both flags and the residual limitation

## Testing

`cargo +1.94 check -p tn-worker-gateway --all-targets` is clean with zero warnings, and `cargo +nightly fmt -- --check` passes.

Nine new tests in `ratelimit.rs` and three in `cli.rs`, all using the existing `ManualClock` and helper shapes with no wall-clock assumptions:

- `rotation_inside_one_v6_prefix_shares_a_bucket` (the regression test)
- `rotation_across_v6_prefixes_uses_separate_buckets` (pins the residual limit honestly)
- `ipv4_default_prefix_preserves_per_address_buckets`
- `ipv4_mapped_peers_key_as_ipv4`
- `prefix_zero_collapses_every_address_to_one_bucket`, `prefix_zero_masks_to_the_unspecified_address`
- `max_prefix_masking_is_identity` (/32 and /128), `masking_clears_only_the_host_bits` (/24 and /64)
- `out_of_range_prefix_is_rejected`
- `per_ip_prefix_defaults_are_v4_32_and_v6_64`, `per_ip_prefixes_are_configurable`, `out_of_range_per_ip_prefix_is_rejected`

**The regression test was confirmed by mutation, not assumed.**  Reverting the one line in `PerIpLimiter::admit` back to the unmasked key:

```diff
-        let key = self.prefix.key(ip);
+        let key = ip;
```

takes the suite from `17 passed; 0 failed` to `13 passed; 4 failed`, and `rotation_inside_one_v6_prefix_shares_a_bucket` fails on its own assertion ("address rotation inside one /64 must not mint a fresh bucket"), not on a build error or a timeout.  Restoring the line returns the suite to `17 passed; 0 failed`.

### DIFF
--- a/bin/worker-gateway/src/ratelimit.rs
+++ b/bin/worker-gateway/src/ratelimit.rs
@@ -2,10 +2,21 @@
 //!
 //! Two token-bucket limiters shed load before a request reaches the proxy or
 //! has its body buffered: a gateway-wide bucket caps aggregate throughput to
-//! roughly what the upstream workers can absorb, and a per-client-IP bucket
-//! stops any single source monopolizing that budget. An over-limit request
-//! receives the gateway's JSON-RPC "rate limit exceeded" envelope (HTTP `429`),
-//! never a bare connection reset.
+//! roughly what the upstream workers can absorb, and a per-client bucket keyed
+//! on the client's *network prefix* caps what one source can take of that
+//! budget. An over-limit request receives the gateway's JSON-RPC "rate limit
+//! exceeded" envelope (HTTP `429`), never a bare connection reset.
+//!
+//! The per-client key is the peer address masked to a configurable prefix
+//! ([`PrefixPolicy`]), not the bare address. Keyed on the bare address, a client
+//! that rotates its source address gets a fresh, full bucket per address and so
+//! never accumulates spent budget, which a single IPv6 `/64` makes trivial;
+//! masking collapses one allocation onto one bucket. The defaults are `/64` for
+//! IPv6 and `/32` for IPv4, so IPv4 behaviour is unchanged and unrelated
+//! customers behind one carrier NAT are never grouped. This bounds rotation
+//! *within* an allocation, not across them: a client holding many distinct
+//! allocations still earns a bucket per allocation, and only the gateway-wide
+//! bucket caps their total.
 //!
 //! The client identity is the immediate TCP peer address (`ConnectInfo`,
 //! injected per connection by the accept loop). The gateway is meant to run
@@ -16,7 +27,8 @@
 
 use std::{
     collections::HashMap,
-    net::{IpAddr, SocketAddr},
+    fmt,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     num::NonZeroU32,
     sync::{Arc, Mutex, MutexGuard, PoisonError},
     time::{Duration, Instant},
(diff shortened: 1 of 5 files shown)


bin/worker-gateway/src/ratelimit.rs
```diff
@@ -354,19 +536,45 @@ mod tests {
         IpAddr::from([10, 0, 0, last])
     }
 
+    /// An address in `2001:db8::/32`: `subnet` picks the `/64`, `host` the
+    /// address within it.
+    fn ip6(subnet: u16, host: u16) -> IpAddr {
+        IpAddr::from([0x2001, 0x0db8, 0, subnet, 0, 0, 0, host])
+    }
+
+    fn prefixes(v4: u8, v6: u8) -> PrefixPolicy {
```
grantkee, line 546: nit: missing doc comment

grantkee (dismissed):
LGTM! ONe nit about a missing doc comment, but not a blocker

grantkee (approved):

PR #1097, opened 2026-08-05
chain-configs: regenerate mainnet genesis with post-fork registry runtime

### LINKED ISSUE #1063: Committed mainnet genesis carries the pre-fork ConsensusRegistry runtime; the default build cannot close its first epoch
## Summary

`chain-configs/mainnet/genesis.yaml` seeds the ConsensusRegistry account with the pre-fork runtime bytecode, but the default (non-`adiri`) build has no pre-fork code path: it issues the post-fork registry reads unconditionally.  As committed, a mainnet chain launched from this genesis runs normally until its first epoch boundary and then cannot produce the epoch-closing block.

## Details

The close-epoch arm reads the committee pool with the post-fork three-call `getValidatorsInfo(status)` ABI.  The legacy routing that would detect pre-fork registry code and fall back to `getValidators` exists only under the `adiri` feature:

- `crates/tn-reth/src/evm/block.rs:502-503`: the `registry_code_is_pre_fork()` check is `#[cfg(feature = "adiri")]`.
- `crates/tn-reth/src/evm/block.rs:538-552`: `registry_code_is_pre_fork` itself is `#[cfg(feature = "adiri")]`.
- `crates/tn-reth/src/evm/block.rs:815`: the in-protocol registry fork (which would swap in the post-fork code) is also `adiri`-gated.
- `crates/tn-reth/Cargo.toml:76,79`: `default = []`; `adiri` is opt-in, wired to the docker-adiri image via `bin/telcoin-network/Cargo.toml:18-22`.

So the default binary requires the genesis registry code to already speak the post-fork ABI.  The committed mainnet genesis does not:

```
rg -c '2870259d|aaffd270' chain-configs/mainnet/genesis.yaml   # no matches:
                                                               # getValidatorsInfo and
                                                               # migrateValidatorSets selectors absent
rg -c '8cc05eda' chain-configs/mainnet/genesis.yaml            # 1: legacy getValidators present
```

Inspecting the registry code bytes directly shows the mainnet alloc is an older pre-fork variant than the one testnet carries: 23680 bytes, and it also lacks `eb8535c2` (`getNextCommitteeSize`) and `7b55a300`, both of which are present in `chain-configs/testnet/genesis.yaml`'s pre-fork registry.  So the committed mainnet registry does not even match the pre-fork surface the `adiri` legacy routing was written against, let alone the post-fork ABI the default build requires.

At the first epoch close, the pool read reverts (selector unknown to the deployed code), `generate_conclude_epoch_calldata` fails, and `apply_closing_epoch_contract_call` returns an error that aborts block execution (`crates/tn-reth/src/evm/block.rs:794-831`).  Every attempt to build the closing block fails the same way.

I cannot see from the repository whether the live mainnet was actually launched from this exact yaml, so this is filed as a repo-internal inconsistency of the committed artifact.  A secondary consequence either way: there is no committed mainnet fixture carrying the post-fork registry, which any downstream tooling modeling the current ABI would want.

## Suggested fix

Regenerate the registry alloc in `chain-configs/mainnet/genesis.yaml` from the current tn-contracts artifact (the submodule gitlink `ac22e265`), so genesis carries the post-fork runtime directly.  That would also be a natural moment to record the post-fork deployed-code hash pin that `crates/types/src/forks.rs:71-75` leaves as a pre-deploy TODO.  Alternatively, if mainnet is intended to replay the registry fork the way adiri does, the legacy routing and fork machinery would need to ride a feature that mainnet builds actually enable.

MavenRain:
Closes #1063.

## Problem

A mainnet node built with default features cannot close its first epoch. No attacker is required: the committed `chain-configs/mainnet/genesis.yaml` allocates a pre-fork ConsensusRegistry runtime (23,680 bytes) that predates the registry read API, so the dispatcher has no code for the selectors the node calls at every epoch boundary:

- `getValidatorsInfo(uint8)` (`0x2870259d`)
- `getNextCommitteeSize()` (`0xeb8535c2`)
- `getCommitteeBlsPubkeys(uint32)` (`0x6ae9942c`)

Each call reverts with empty output (`Revert { output: 0x }`), which surfaces as a liveness failure at the first epoch close. Independently, the committed `committee.yaml` still uses the old flat schema without the required `bootstrap_servers` field, so it fails deserialization before the node even boots.

## Root cause

The mainnet placeholder yamls were committed before the ConsensusRegistry fork and the committee schema change, and nothing tied them to the current code: no test loaded the committed files through the real node paths, and no test compared the genesis registry alloc against the contract artifact the genesis ceremony actually splices in.

## Fix

Regenerate all three `chain-configs/mainnet` files with the repo's own in-process genesis ceremony (`GenesisArgs::execute` plus the keytool through `CommandParser`, the same path the e2e tests use), and add regression gates that would have caught both defects.

Preserved from the committed files: chain id 487, the owner and dev-funded account `0x3DCc9a6f3A71F0A6C8C659c65558321c374E917a`, the 86400s epoch duration (decoded from the committed registry storage; the only non-default stake config value), and the four placeholder execution addresses (`0x1111...`, `0x2222...`, `0x3333...`, `0x4444...`). Fresh placeholder BLS identities are minted because the old proofs of possession are not recoverable from the committed data; these are placeholders either way and must be replaced by the real ceremony before launch.

Expected regeneration drift, now documented in the runner: fresh genesis timestamp, `cancunTime`/`pragueTime: 0` now present, and the 17 stale precompile-config accounts replaced by the current 14-account set.

## Changes

- [x] `chain-configs/mainnet/genesis.yaml`: regenerated; the ConsensusRegistry alloc is now byte-identical to the current contract artifact's deployed bytecode
- [x] `chain-configs/mainnet/committee.yaml`: regenerated on the current schema (including `bootstrap_servers`), four authorities
- [x] `chain-configs/mainnet/parameters.yaml`: regenerated; gains only the previously missing `parallel_fetch_request_delay_interval`
- [x] `crates/config/src/node.rs`: Gate A. Loads all three committed files through the real node deserialization paths (including `Config::load_from_path::<Committee>`), and asserts the genesis registry code equals the `deployedBytecode` of the bundled contract artifact, the same artifact the ceremony splices
- [x] `crates/tn-reth/src/env/epoch.rs`: Gate B. Boots `RethEnv` over the committed mainnet genesis in-process, issues the three post-fork reads first (this is the #1063 symptom), then cross-checks the on-chain committees for epochs 0 through 2 against `committee.yaml` by exact BLS pubkey set equality, and pins the next committee size to the four genesis authorities
- [x] `crates/telcoin-network-cli/src/genesis/mod.rs`: an `#[ignore]`d `regenerate_mainnet_chain_configs` runner so future refreshes use the identical ceremony; it reads all ceremony outputs before writing any committed file, so a failure cannot leave `chain-configs/mainnet` half-written
- [x] `crates/tn-reth/Cargo.toml` and `Cargo.lock`: `serde_yaml` as a dev-dependency for Gate B's committee cross-check

Both gates run under default features in the normal test suite, so a future stale-artifact commit fails CI instead of failing mainnet.

## Testing

- Red first: both gates were written before the regeneration and observed failing against the committed files (Gate B fails at the first read with `Revert { output: 0x }`; Gate A fails on both the schema load and the bytecode comparison), then green after
- Mutation confirmation: the set-equality assertion was seen to fail on a single flipped pubkey byte and pass again on restore; the committee-schema gate was confirmed the same way
- `cargo +nightly fmt --all -- --check` clean
- `cargo +1.94 clippy --all-targets` clean (workspace)
- `cargo +1.94 test --no-run --workspace` clean in an isolated target dir (attest's test-build)
- Suites re-run green on the pinned toolchain: Gate A (2 passed), the tn-reth `env::` suite (18 passed), the telcoin-network-cli suite (60 passed, 1 ignored)


sstanfield (dismissed):
This is cool, these will obviously be replaced again when main net launches.

grantkee (approved):

PR #1098, opened 2026-08-05
storage: end the epoch-record waits with a lookup at the deadline

### LINKED ISSUE #1082: Epoch-record poll loops cannot observe a value arriving during their final sleep: `cert_by_digest_with_timeout` and `record_by_epoch_with_timeout` sleep, then test the deadline, and return `None` without a last check
**Severity:** Informational.  Polling-loop ordering on two bounded waits.  The 200 ms blind window at the deadline is real at both call sites, but the outcomes match what each caller already accepts for a genuinely late value: the export path skips one bundle (the epoch's record and certificate still reach the next boundary's cumulative bundle;  its own snapshot bundle is never produced), and the record path ends the epoch loop with an error that a process restart clears, the record being durable by then.  This is a loop-shape report with a test gap, not an operational problem.

## Summary

Both timeout helpers in the epoch-records DB poll with the same ordering (`crates/storage/src/epoch_records.rs:383-391`, `:426-434`):

```rust
loop {
    if let Some(cert) = self.cert_by_digest(digest).await {
        return Some(cert);
    }
    tokio::time::sleep(Duration::from_millis(200)).await;
    if tokio::time::Instant::now() >= deadline {
        return None;
    }
}
```

Check, then sleep, then deadline test.  A value persisted during the final sleep is never observed: after that sleep the deadline test returns `None`, one actor round-trip short of success.

The blind window sits exactly where these helpers earn their keep.  Both exist to bridge a known arrival race: a just-closed epoch's certificate is aggregated only at the next epoch's start (`epoch_records.rs:414-419`), and a node restarting mid-catch-up waits for the epoch record collector to fetch the previous record from peers (`crates/node/src/manager/node/run_epoch.rs:531-538`).  A wait that runs long is precisely the case where the value lands near the deadline, and the last 200 ms before it are unobservable.

## Call-site consequences

Two call sites, both with bounded fallout, which is why this is Informational rather than Low:

1. Export path: the completion task waits `CERT_WAIT` (90 s, `crates/node/src/manager/node/close_epoch.rs:45`) for epoch N's own certificate and abandons the export on `None` (`close_epoch.rs:456-465`).  By that point the expensive plain-state walk has already finished (the task body runs only after `rx.await` resolves, `close_epoch.rs:427`), so a certificate landing in the final sleep discards a completed export.  The damage stops there: the idempotency check keys on epoch N's own directory (`close_epoch.rs:353-358`), so the next boundary is not suppressed;  its historical-cert precheck scans `0..N+1` (`epoch_records.rs:516-523`), which passes because the certificate did arrive, and the `0..=N+1` bundle then carries epoch N's record and certificate (`epoch_records.rs:547-560`).  What is permanently lost is epoch N's own per-epoch bundle: `export_once` refuses to export a block that is no longer the persisted tip (`crates/node/src/manager/export_exec.rs:194-213`), so per-epoch bundles are now-or-never by design, and this outcome is identical to the certificate genuinely missing the window.

2. Record path: `open_epoch_pack` (`run_epoch.rs:501`) waits 30 s for the previous epoch record during restart catch-up and turns `None` into an error (`run_epoch.rs:562-573`).  That error propagates out of epoch startup (`run_epoch.rs:230`) into `run_epochs`, where any epoch error aborts the epoch loop (`crates/node/src/manager/node.rs:1202-1236`) and thereby ends the node (`node.rs:1081-1096`).  A record persisted during the final sleep therefore turns a would-be success into a node-ending error;  the mitigating detail is that the record is durable by then, so the next process start finds it on the immediate lookup (`run_epoch.rs:508-510`) and proceeds without waiting.

## Suggested fix

Test the deadline before sleeping, and clamp the sleep so the loop always ends with a check (`epoch_records.rs:426-434`):

```rust
loop {
    if let Some(cert) = self.cert_by_digest(digest).await {
        return Some(cert);
    }
    let now = tokio::time::Instant::now();
    if now >= deadline {
        return None;
    }
    tokio::time::sleep_until(now + Duration::from_millis(200).min(deadline - now)).await;
}
```

Same reorder in `record_by_epoch_with_timeout` (`epoch_records.rs:383-391`).  The final check now starts at or after the deadline, and the DB actor serves reads and writes in FIFO order on one thread, so any save enqueued strictly before the deadline is observed.  The one residual race is the very first check: a `timeout` smaller than that check's own actor round-trip can expire while the read is in flight, missing a save enqueued behind it;  the real call sites' 30 s and 90 s timeouts are nowhere near that regime.  The wait still ends within one actor round-trip of the deadline.  A side benefit: a `timeout` below the 200 ms poll interval stops over-sleeping (today a 50 ms timeout still sleeps the full 200 ms, and a 300 ms timeout returns only after about 400 ms).  Both existing tests keep passing under the reorder;  the absent-cert test asserts only a 200 ms floor on the elapsed time (`epoch_records.rs:1852-1853`).

The two existing tests cover a certificate already present (`epoch_records.rs:1817-1832`) and one that never arrives (`:1835-1854`);  neither drives a save inside the final poll interval.  A test that saves the certificate inside that interval would pin the corrected behavior: with the reorder, any save strictly before the deadline must be returned.

## Acceptance criteria

- A record or certificate whose save reaches the DB actor strictly before the deadline is returned, including during what is today the final sleep (the final check runs at or after the deadline).
- The helpers still return `None` within one actor round-trip after the deadline when the value never arrives.
- A `timeout` shorter than the 200 ms poll interval no longer sleeps past its own deadline.
- Both helpers share the corrected shape, and a test exercises a save landing inside the final poll interval.

## Related, not duplicate

- #573 (closed) asked for an event-driven wait (a `notify_read`-style structure) to replace these sleep-polls, and `record_by_epoch_with_timeout` still carries its marker (`// TODO issue 573, clean this up.`, `epoch_records.rs:382`).  That refactor would subsume this fix by deleting the loop.  Until it lands, the reorder above is the minimal correctness change to the loop as written.
- No open PR currently touches `epoch_records.rs`.

## References

- `crates/storage/src/epoch_records.rs:376-392` (`record_by_epoch_with_timeout`;  loop `:383-391`;  the issue-573 TODO at `:382`), `:420-435` (`cert_by_digest_with_timeout`;  loop `:426-434`), `:414-419` (why the certificate wait exists), `:516-523` (`first_missing_historical_cert` scans `0..tip_epoch`), `:547-560` (`export_bounded_bundle`;  the `0..=through_epoch` loop at `:560`), `:1817-1832` / `:1835-1854` (existing tests).
- `crates/node/src/manager/node/close_epoch.rs:45` (`CERT_WAIT` is 90 s), `:353-358` (per-epoch idempotency key), `:360-376` (historical-cert precheck retries next epoch), `:421-465` (completion task ordering: state walk first, certificate wait second, abandon on `None`).
- `crates/node/src/manager/node/run_epoch.rs:501` (`open_epoch_pack`), `:508-510` (the immediate lookup that succeeds after a restart), `:531-538` (restart-catch-up scenario), `:562-573` (30 s wait;  `None` becomes an error), `:230` (the error propagates out of epoch startup).
- `crates/node/src/manager/node.rs:1202-1236` (`run_epochs`: any epoch error aborts the epoch loop), `:1081-1096` (the aborted loop's error ends the node).
- `crates/node/src/manager/export_exec.rs:194-213` (`export_once` requires the exported block to still be the persisted tip;  per-epoch bundles are now-or-never).

All line numbers at `origin/main` `c6d3698a`.

MavenRain:
Closes #1082.

## Problem

Both bounded waits in the epoch-records database polled in an order that cannot observe a value arriving in their final interval (`crates/storage/src/epoch_records.rs`, `record_by_epoch_with_timeout` and `cert_by_digest_with_timeout`):

```rust
loop {
    if let Some(cert) = self.cert_by_digest(digest).await {
        return Some(cert);
    }
    tokio::time::sleep(Duration::from_millis(200)).await;
    if tokio::time::Instant::now() >= deadline {
        return None;
    }
}
```

Check, then sleep, then test the deadline.  A value persisted during that last sleep is never looked for: the loop wakes, sees the deadline has passed, and gives up one round trip short of success.

The blind window sits exactly where these two helpers earn their keep.  Both exist to bridge a known arrival race, so a wait that runs long is precisely a wait whose value is likely to land near the deadline, and the last 200 ms before it were unobservable.

This is a loop-shape defect with a test gap, not an operational problem, which is why the issue is filed Informational.  Both call sites already accept the same outcome for a genuinely late value:

- The export path waits `CERT_WAIT` (90 s, `crates/node/src/manager/node/close_epoch.rs:45`) for the closing epoch's own certificate, and abandons the export when the wait returns `None` (`close_epoch.rs:456-465`).  The plain-state walk has already finished by then, because the completion task body only runs after `rx.await` resolves (`close_epoch.rs:427`, fed by the blocking `export_once` on the exporter thread, `crates/node/src/manager/export_exec.rs:175`), so a certificate landing in the final sleep discards a completed export.  What is lost is epoch N's own per-epoch bundle, permanently: `export_once` refuses to export a block that is no longer the persisted tip (`export_exec.rs:204`), and `export_epoch_state` runs once per boundary (`run_epoch.rs:418`).  The damage stops there.  The idempotency check keys on epoch N's own directory (`close_epoch.rs:355`), so the next boundary is not suppressed, and its historical-cert precheck scans `0..tip_epoch` (`epoch_records.rs:516`), which passes because the certificate did arrive.
- The record path waits 30 s for the previous epoch record during restart catch-up and turns `None` into an error (`crates/node/src/manager/node/run_epoch.rs:565-573`).  That error leaves epoch startup (`run_epoch.rs:230`) and reaches `run_epochs`, where any epoch error aborts the epoch loop (`crates/node/src/manager/node.rs:1234`) and ends the node.  The mitigation is that the record is durable by then, so the next process start finds it on the immediate lookup (`run_epoch.rs:508-509`) and proceeds without waiting.

A second, smaller consequence of the same ordering: a `timeout` below the poll interval was rounded up to a full interval.  A 50 ms wait slept 200 ms;  a 300 ms wait returned after about 400 ms.

## Fix

Test the deadline before sleeping, and clamp each sleep to it, so the wait always ends with a lookup rather than with a sleep.

- [x] `crates/storage/src/epoch_records.rs`: new private `EpochRecordDb::poll_until_deadline`, which polls a caller-supplied lookup every `POLL_INTERVAL` until it yields a value or the timeout expires.  The 200 ms interval, duplicated as a literal in both loops, is now the single `POLL_INTERVAL` constant.
- [x] Both `record_by_epoch_with_timeout` and `cert_by_digest_with_timeout` delegate to it, so the two waits cannot drift apart, and the `// TODO issue 573` refactor that replaces polling with an event-driven wait has one loop to delete instead of two.
- [x] `crates/storage/Cargo.toml`: `test-util` added to the dev-dependency `tokio` features.  The new tests drive a paused clock, and without this the crate compiles them only because another workspace member happens to enable the feature for the whole build.

The corrected shape:

```rust
let deadline = tokio::time::Instant::now() + timeout;
loop {
    if let Some(found) = lookup().await {
        return Some(found);
    }
    let now = tokio::time::Instant::now();
    if now >= deadline {
        return None;
    }
    tokio::time::sleep_until(deadline.min(now + POLL_INTERVAL)).await;
}
```

Why the final lookup is enough: `EpochRecordDb` dispatches every message over one mpsc channel to a single background thread that drains it with `blocking_recv` (`epoch_records.rs:84-142`), so reads and writes are served in enqueue order.  A save enqueued strictly before the deadline is therefore visible to a lookup enqueued at the deadline.

One residual race is left, and it is stated in the helper's doc comment rather than papered over: the very first lookup can be in flight when a short deadline passes, so a save enqueued behind it is missed.  That needs a `timeout` shorter than one round trip to the database thread;  the real call sites pass 30 s and 90 s.  The wait still ends within one round trip of the deadline.

Threat model: none of this is adversary-facing.  `cert_by_digest_with_timeout` and `record_by_epoch_with_timeout` are internal database reads with node-chosen timeouts, and the values they wait for are certificates and records that are verified elsewhere before they are stored.  The change alters when a wait gives up, not what it accepts.

## Testing

Three new tests in `crates/storage/src/epoch_records.rs`, all on a paused clock so the ordering they assert is exact rather than probable.  The database thread is a real thread, but a paused clock only advances to the next timer while the runtime is idle, so the delayed save lands between the last pre-deadline lookup and the deadline no matter how slow the machine is.  That is what lets these tests pin a 200 ms window without being timing races.

- [x] `cert_by_digest_with_timeout_sees_cert_saved_in_final_poll_interval`: the certificate is saved 900 ms into a 1 s wait, after the last lookup a sleep-then-test loop would make (800 ms) and before the deadline.
- [x] `record_by_epoch_with_timeout_sees_record_saved_in_final_poll_interval`: the same property on the wait restart catch-up depends on.
- [x] `cert_by_digest_with_timeout_ends_at_a_deadline_inside_one_poll_interval`: a 50 ms timeout must not spend a full 200 ms poll interval.

The two existing tests are unchanged and still pass, including the one that asserts a 200 ms floor on an absent-certificate wait (a 300 ms timeout still crosses one full interval before the clamped final sleep).

`cargo +1.94 test -p tn-storage --lib epoch_records`: 32 passed, 0 failed.

Both halves of the fix were then confirmed by mutation, and each mutation killed exactly the tests that pin the property it broke:

1. Putting the sleep back before the deadline test, the pre-fix shape, failed exactly the three new tests and nothing else: 29 passed, 3 failed.  All three failures are assertion panics rather than compile errors, and the two pre-existing timeout tests are among the passes, so the reordering is not something they could ever have caught.
2. Keeping the deadline test first but dropping the clamp on the sleep failed exactly one test, `cert_by_digest_with_timeout_ends_at_a_deadline_inside_one_poll_interval`: 31 passed, 1 failed.  The two save-in-the-final-interval tests still passed under it, which is the point . . . they pin the reordering and only the third pins the clamp, so neither property rides on the other's test.

## Scope notes

- The `// TODO issue 573, clean this up.` marker is kept, and moves onto the helper where the poll loop now lives, since the extraction would otherwise leave it sitting over a one-line delegation with nothing to clean up.  #573 asked for an event-driven wait to replace these sleep-polls;  that refactor subsumes this fix by deleting the loop, and until it lands this is the minimal correctness change to the loop as written.
- The poll interval stays 200 ms.  Nothing here changes how long a caller waits, only where the wait ends.
- Three other loops in the tree end on a sleep the same way, and are deliberately left alone.  `export_once` sleeps on its last pin attempt and then falls out of the `for` bound without re-testing `verify_tip` (`crates/node/src/manager/export_exec.rs:199-256`);  `consensus_output_bytes` refreshes `my_number` after its last sleep but discards it on the attempt count before the `while` head can re-test it (`crates/consensus/primary/src/network/handler.rs:1088-1096`);  `dial_peer_bls` gives up after a backoff without re-dialing (`crates/node/src/manager/node/start_epoch.rs:672-682`).  Each has a different caller and a different consequence, so each belongs in its own change rather than widening this one.  The house shape is the corrected one: `wait_until` in `crates/test-utils/src/wait.rs`, `wait_for_network_peers`, `retry_provider_faults`, and the state-sync fetch retry all re-check before giving up.

### DIFF
--- a/crates/storage/Cargo.toml
+++ b/crates/storage/Cargo.toml
@@ -46,8 +46,10 @@ tn-reth = { workspace = true }
 tn-test-utils = { workspace = true }
 rand = { workspace = true }
 roaring = { workspace = true }
-# Async barrier tests for the durable `persist` need a runtime + timers.
-tokio = { workspace = true, features = [ "macros", "rt", "time" ] }
+# Async barrier tests for the durable `persist` need a runtime + timers. `test-util` is for the
+# paused-clock tests of the bounded waits in `epoch_records`, which would otherwise depend on
+# another workspace member happening to enable it for the whole build.
+tokio = { workspace = true, features = [ "macros", "rt", "time", "test-util" ] }
 
 [features]
 redb = []
--- a/crates/storage/src/epoch_records.rs
+++ b/crates/storage/src/epoch_records.rs
@@ -8,6 +8,7 @@
 use std::{
     error::Error,
     fmt::Display,
+    future::Future,
     hash::BuildHasherDefault,
     io,
     path::{Path, PathBuf},
@@ -36,6 +37,10 @@ use crate::archive::{
 
 /// Current version of the epoch pack file.
 const EPOCH_PACK_VERSION: u16 = 0;
+
+/// Interval between lookups in the bounded waits [`EpochRecordDb::record_by_epoch_with_timeout`]
+/// and [`EpochRecordDb::cert_by_digest_with_timeout`].
+const POLL_INTERVAL: Duration = Duration::from_millis(200);
 
 enum EpochDbMessage {
     /// Save a "dummy" epoch 0 [`EpochRecord`] without a certificate.
@@ -371,24 +376,51 @@ impl EpochRecordDb {
         }
     }
 
+    /// Poll `lookup` every [`POLL_INTERVAL`] until it yields a value or `timeout` elapses,
+    /// whichever happens first.
+    ///
+    /// The wait always ends with a lookup at the deadline rather than with a sleep: the deadline
+    /// is tested before sleeping, and each sleep is clamped to it. A value stored during the last
+    /// poll interval is therefore still observed, which is the case both callers exist to cover.
+    /// They bridge a known arrival race, so a wait that runs long is precisely a wait whose value
+    /// is likely to land near the deadline. Clamping also keeps a `timeout` shorter than
+    /// [`POLL_INTERVAL`] from being rounded up to a full interval.
+    ///
+    /// Every message is served on one background thread in enqueue order, so a save enqueued
+    /// strictly before the deadline is visible to that final lookup. The residual race is the
+    /// first lookup only: a `timeout` shorter than one round trip to that thread can expire while
+    /// the lookup is in flight, missing a save enqueued behind it. The wait still ends within one
+    /// round trip of the deadline.
+    async fn poll_until_deadline<T, F, Fut>(timeout: Duration, lookup: F) -> Option<T>
+    where
+        F: Fn() -> Fut,
+        Fut: Future<Output = Option<T>>,
+    {
+        let deadline = tokio::time::Instant::now() + timeout;
+        // TODO issue 573, clean this up: an event-driven wait would replace this poll entirely.
+        loop {
+            if let Some(found) = lookup().await {
+                return Some(found);
+            }
+            let now = tokio::time::Instant::now();
+            if now >= deadline {
+                return None;
+            }
+            tokio::time::sleep_until(deadline.min(now + POLL_INTERVAL)).await;
+        }
+    }
+
     /// Retrieve an [`EpochRecord`] by epoch number.
     /// This version will wait up to timeout time for the record to show up if not available.
+    ///
+    /// The wait ends with a lookup at the deadline, so a record saved during the last poll
+    /// interval is still returned (see `poll_until_deadline`).
     pub async fn record_by_epoch_with_timeout(
         &self,
         epoch: Epoch,
         timeout: Duration,
     ) -> Option<EpochRecord> {
-        let deadline = tokio::time::Instant::now() + timeout;
-        // TODO issue 573, clean this up.
-        loop {
-            if let Some(rec) = self.record_by_epoch(epoch).await {
-                return Some(rec);
-            }
-            tokio::time::sleep(Duration::from_millis(200)).await;
-            if tokio::time::Instant::now() >= deadline {
-                return None;
-            }
-        }
+        Self::poll_until_deadline(timeout, || self.record_by_epoch(epoch)).await
     }
 
     /// Retrieve an [`EpochRecord`] by its digest.
@@ -417,21 +449,15 @@ impl EpochRecordDb {
     /// A just-closed epoch's certificate is only aggregated at the next epoch's start, so a caller
     /// that needs epoch N's cert immediately after N closes (e.g. the state exporter) must give the
     /// collector a bounded window to produce it.
+    ///
+    /// The wait ends with a lookup at the deadline, so a certificate saved during the last poll
+    /// interval is still returned (see `poll_until_deadline`).
     pub async fn cert_by_digest_with_timeout(
         &self,
         digest: EpochDigest,
         timeout: Duration,
     ) -> Option<EpochCertificate> {
-        let deadline = tokio::time::Instant::now() + timeout;
-        loop {
-            if let Some(cert) = self.cert_by_digest(digest).await {
-                return Some(cert);
-            }
-            tokio::time::sleep(Duration::from_millis(200)).await;
-            if tokio::time::Instant::now() >= deadline {
-                return None;
-            }
-        }
+        Self::poll_until_deadline(timeout, || self.cert_by_digest(digest)).await
     }
 
     /// True if the database contains a record for the given epoch number.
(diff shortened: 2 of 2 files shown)


grantkee (approved):
LGTM!

PR #1101, opened 2026-08-05
Write next-epoch worker base fees to WorkerConfigs at epoch close

### LINKED ISSUE #1100: Record next-epoch worker base fees on-chain at epoch close
## Problem

Telcoin Network prices each worker's transactions with a per-worker base fee that can follow an EIP-1559-style adjustment strategy. Today, the fee a worker should charge for the *next* epoch exists only implicitly: it has to be derived by walking the previous epoch's block headers and re-running the adjustment formula against their recorded gas usage. Nothing in on-chain state records the fee directly.

This has two costs. First, every consumer of a worker's fee — the node entering a new epoch, any snapshot/replay path, external tooling — has to re-derive it by re-executing the same header-scan logic, rather than reading one value. That logic has to stay in lockstep with whatever accumulates gas during the epoch, and any divergence between the scan and the live accumulator is a silent consensus risk rather than something a single state read would catch. Second, the `WorkerConfigs` contract as currently deployed on the live adiri testnet predates this need: it stores each worker's fee strategy but exposes no way to write a computed value back into contract storage, and its `data` field is typed too narrowly (`uint128`) for a governance-agnostic reserved word.

Fixing the derivation problem without fixing the storage gap just moves the recomputation elsewhere. The live `WorkerConfigs` deployment has no setter selectors and an incompatible field width, so any on-chain write path requires upgrading the deployed bytecode — and that upgrade has to happen at a well-defined, fail-closed boundary, since it changes what a subsequent read of the same address means for every node re-executing history.

## Solution

At the moment an epoch closes, the block executor should compute each EIP-1559 worker's next-epoch base fee from that epoch's final gas accumulator state and write it into the corresponding `WorkerConfigs` row as an additional system call, alongside the existing `applyIncentives` / `applySlashes` / `concludeEpoch` sequence. `Static`-strategy workers, whose fee is a governance-set constant, should be left untouched. The write should run last in the closing sequence, after the registry-side calls, since it depends on none of their state and none of theirs depend on it.

Because the deployed `WorkerConfigs` contract lacks the setter this requires, the runtime code should be swapped for an updated artifact at the same fork boundary already used to upgrade the validator registry — the two upgrades are coupled: the registry swap is what flips the closing sequence onto the path that needs the new selector, so a build that applies one swap without the other would fail the very next closing block. The swap should be gated the same way the registry swap already is: fail closed if the on-chain code doesn't hash to the known pre-fork bytecode, so an unexpected deployment aborts rather than silently upgrading over an incompatible storage layout. Pre-fork closes should keep replaying the legacy call sequence unchanged, so historical blocks stay re-executable.

The fee formula itself should be consolidated into a single shared function so that every seam that prices a worker's next-epoch fee — the epoch-close write introduced here, and any other derivation path — computes from identical inputs and can't drift apart. This is the direct predecessor to also reading the fee back from that same `WorkerConfigs` row at epoch entry, replacing header-scanning entirely; that read path is deliberately out of scope for this change and left for a follow-up once the write path is proven.

grantkee:
- Adds a fourth epoch-closing system call, `setWorkerConfigsData`, run after `applyIncentives` / `applySlashes` / `concludeEpoch` since it reads and writes no registry state. For every `Eip1559`-strategy worker it prices the next epoch from that epoch's final gas accumulator totals and stores the result in the row's `data` word; `Static` rows are left alone.
- Swaps the `WorkerConfigs` runtime bytecode at the same `CONSENSUS_REGISTRY_FORK_EPOCH` boundary as the existing registry swap — the live deployment predates the setter selectors and has an incompatible `uint128` `data` field. Gated fail-closed on `WORKER_CONFIGS_PRE_FORK_CODE_HASH`, mirroring the registry's pre-fork pin. Pre-fork closes keep replaying the legacy `applyIncentives` + `concludeEpoch` pair unchanged.
- Consolidates the fee-strategy formula into `next_base_fee_for_config` in `gas_accumulator.rs` so the new write path and the existing `adjust_base_fees` / `fold_next_epoch_base_fees` seams all derive from one function instead of duplicated match arms.
- Threads `GasAccumulator` through `RethEnv` into the block-execution context (replacing the narrower `RewardsCounter`) so the executor can reach gas totals and current fees, not just leader counts, at close.
- Distinguishes `Revert` (decoded reason + selector) from `Halt` in the shared system-call error path, so a reverted closing call is diagnosable from the error/log alone instead of a generic failure message.
- Test coverage: 3-way cross-layer equality pinning the write against `adjust_base_fees` and the entry-derivation formula, mutant-verified scoping of the own-gas fold to the closing worker, a pinned decode fallback for unknown strategy ids, and e2e coverage across basefee, epoch-boundary, and restart scenarios.

closes #1100 
blocked by #1072

### DIFF
--- a/crates/engine/tests/it/main.rs
+++ b/crates/engine/tests/it/main.rs
@@ -126,7 +126,7 @@ async fn test_empty_output_skips_execution() -> eyre::Result<()> {
         Some(chain.clone()),
         None,
         tmp_dir.path(),
-        Some(gas_accumulator.rewards_counter()),
+        Some(gas_accumulator.clone()),
     )?;
     // update rewards counter so execution address is visible
     let committee =
@@ -231,7 +231,7 @@ async fn test_queued_outputs_bounded_with_backpressure() -> eyre::Result<()> {
         Some(chain.clone()),
         None,
         tmp_dir.path(),
-        Some(gas_accumulator.rewards_counter()),
+        Some(gas_accumulator.clone()),
     )?;
     let committee =
         create_committee_from_state(execution_node.epoch_state_from_canonical_tip().await?).await?;
@@ -341,7 +341,7 @@ async fn test_empty_output_with_close_epoch_still_executes() -> eyre::Result<()>
         Some(chain.clone()),
         None,
         tmp_dir.path(),
-        Some(gas_accumulator.rewards_counter()),
+        Some(gas_accumulator.clone()),
     )?;
     // update rewards counter so execution address is visible
     let committee =
@@ -753,7 +753,7 @@ async fn test_empty_output_increments_leader_count() -> eyre::Result<()> {
         Some(chain.clone()),
         None,
         tmp_dir.path(),
-        Some(gas_accumulator.rewards_counter()),
+        Some(gas_accumulator.clone()),
     )?;
     // update rewards counter so execution address is visible
     let committee =
@@ -933,7 +933,7 @@ async fn test_happy_path_full_execution_even_after_sending_channel_closed() -> e
         Some(chain.clone()),
         None,
         &tmp_dir.path().join("exc-node"),
-        Some(gas_accumulator.rewards_counter()),
+        Some(gas_accumulator.clone()),
     )?;
 
     // create committee from genesis state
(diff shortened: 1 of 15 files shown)


MavenRain (approved):
Good stuff

PR #1103, opened 2026-08-05
 Read entered-epoch base fees from chain instead of re-deriving them

### LINKED ISSUE #1102: Epoch entry should read next-epoch base fees instead of deriving them from a header scan
## Problem

Every epoch entry — a live boundary crossing, a restart, a mid-epoch re-entry, or a sync — currently derives each worker's base fee by scanning the entire previous epoch's block range, filtering out the genesis block and the synthetic empty-close block, extracting the last held fee and gas total per worker, then folding those through the worker's fee strategy. Workers with no genuine block in the scanned range (idle or newly configured workers) need a second mechanism: a backward walk through earlier epoch boundaries to find a chain-observable anchor, folding forward through every boundary crossed.

This derivation is O(epoch length) on every entry, and the idle-worker walk-back is unbounded in the worst case — a worker configured but never producing a block could force a walk back through many epoch boundaries before finding an anchor. The logic also carries several correctness-sensitive edge cases: identifying the synthetic empty-close block by its zero ommers hash, encoding worker id in the low 16 bits of a header's difficulty field, and keeping the fold formula identical to the one the live producer applies at close time so a restarted node reconstructs the exact value the committee holds in memory.

Base fees are exact-match consensus values. Any divergence between what a node derives on entry and what the rest of the fleet is running with is a safety problem, not just a performance one, which makes the complexity of this path more consequential than ordinary derived state.

## Solution

The closing block should record each worker's next-epoch base fee directly into on-chain state as part of the epoch-close system calls, so epoch entry can read that value with a single state lookup instead of re-deriving it from headers.

Concretely: the epoch-closing block's system call sequence should gain a step that writes each EIP-1559 worker's freshly computed next-epoch fee into the `WorkerConfigs` contract's per-worker `data` word (a field the contract already reserves per row). Epoch entry then resolves to one read of that word, decoded through the same strategy-mapping function the write path uses, rather than a scan over the whole prior epoch plus a conditional walk-back for idle workers. A worker whose row has never been written (no epoch has closed since it was configured) should read as the protocol minimum, which is value-identical to what the header-scan derivation already produced for every pre-existing epoch, so the cutover doesn't change any historical value.

Because the deployed `WorkerConfigs` contract predates this write path (its `data` field is too narrow and it exposes no setter for it), the contract's runtime code needs a one-time swap at the same fork epoch already planned for the `ConsensusRegistry` upgrade, gated fail-closed on a pinned hash of the live pre-fork deployment, mirroring the existing registry fork gate. Until that fork lands, governance must not touch any `WorkerConfigs` row on the live network — a write under the old contract's layout using the entry-read build's assumptions about that state would split the fleet.

The read itself should fail hard on any error: a node that cannot verify the exact fee for an entered epoch must halt rather than produce blocks with an unverifiable value, since keeping the wrong fee is a chain-halting risk while stopping one node is only a liveness cost. This is a stricter posture than the close-time update, which can fail open because keeping the current fee is a value every committee member computes identically.

grantkee:
- The epoch-closing block now records each `Eip1559` worker's next-epoch base fee directly into its `WorkerConfigs.data` word (`record_next_epoch_base_fees` in `tn-reth::evm::block`), using the same `next_base_fee_for_config` formula the live producer's `adjust_base_fees` already applied at close.
- Epoch entry (`run_epoch`) now reads that recorded value back with one pinned state read (`read_base_fees_for_entered_epoch` → `entry_fee_for_worker`) instead of scanning the whole prior epoch's headers and walking back through earlier boundaries to recover idle workers' fees. This replaces `derive_base_fees_for_entered_epoch`, `derive_idle_worker_fee`/`derive_idle_worker_fee_at`, `fold_next_epoch_base_fees`, `latest_base_fee_per_worker`, and `gas_used_per_worker` entirely — O(1) in epoch length instead of O(epoch length).
- `WorkerConfigEntry` (decoded from `getAllWorkerConfigs()`) now carries the raw `data` word alongside the strategy instead of dropping it; `get_worker_fee_configs_at_block` returns `Vec<WorkerConfigEntry>` rather than `Vec<WorkerFeeConfig>`.
- The entry read is a hard consensus-safety gate: any `StateReadError` (provider or chain-global) halts the node rather than entering an epoch on an unverifiable fee. Node-local provider faults are retried briefly first (`retry_provider_faults`, new in `run_epoch.rs`) since peers reading the same pinned block may succeed.
- A pin guard validates that the header being read is genuinely the entered epoch's closing block (registry reports `entered` beginning at `header.number + 1`) — the same boundary predicate now shared by `adjust_base_fees`, the entry read, and the snapshot restore precondition, so the three seams can't drift on what counts as a closing block.
- Snapshot export/import follows suit: `check_fee_precondition`/`check_fees_resumable` (whole-window scan) are replaced by `check_entry_readiness`/`entry_readiness_precondition` (single anchor-block check), since a restored node's first epoch entry now needs only the snapshot's final block to be a valid closing block with readable `WorkerConfigs` rows.
- `is_worker_batch_block` and the synthetic-empty-close-block filtering it existed for are removed — no longer needed once fees come from a direct state read instead of a header scan.
- `sync_num_workers_from_chain` and the epoch-0 arm of `derive_idle_worker_fee` are unchanged in behavior; epoch 0 still has no prior closing block and keeps `MIN_PROTOCOL_BASE_FEE` defaults.
- `forks.rs` documents a new adiri rollout constraint: until the entry-read fork epoch passes on a fleet running this build, governance must not write any `WorkerConfigs` row (no `setWorkerConfig`, no strategy flips, no `data` writes) — a stray non-zero or out-of-`u64` word before the write path activates would split the fleet on base fee or halt entry-read nodes outright.
- The close-time `adjust_base_fees` computation is kept for now (it still carries the close-time boundary-identity check and keeps the accumulator correct between close and the next entry) but is noted as a `FOLLOW-UP` for removal once the on-chain record has soaked in production.
- e2e tests updated: `basefee.rs` and `state_export_import.rs` now exercise the on-chain fee record end-to-end (full round trip, boundary-kill restart recovery) rather than the removed header-derivation path.

closes #1102


MavenRain (approved):
Claude flagged the following three potential test issues:

  state_export_import.rs never asserts a recovered base fee. Both tests use
  config_local_testnet_with_epoch_duration, never the worker-fee variant, so every
  worker sits at the genesis default and never moves off MIN. No assertion in the
  file reads base_fee_per_gas. check_entry_readiness only confirms the word is a
  valid u64, not that it is the correct value. This is the one suite covering the
  snapshot import boundary, which is where a decode bug would first appear, and it
  currently cannot catch one.

  test_boundary_added_worker_data_written_at_creation_close is mathematically
  vacuous. Its oracle is next_base_fee_for_config(new_cfg, MIN_PROTOCOL_BASE_FEE, 0).
  With current_base_fee already at MIN and gas_used == 0, the EIP-1559 formula can
  only hold or decrease, and is then floored at MIN. The expected value is provably
  MIN for every config, so the test cannot distinguish correct fresh-slot pricing
  from pricing off the wrong worker's config or from a hardcoded MIN. This is the
  only test covering the governance-added-worker write path.

  EpochBaseFees::apply has a tested growth path and an untested shrink path.
  set_num_workers truncates unconditionally and its own doc makes the safety bound a
  caller obligation, with inc_block panicking on a truncated id by design. The diff
  adds a thorough growth test but no shrink counterpart anywhere in the tree; all
  three setNumWorkersCall sites use numWorkers_: 2.

PR #1107, opened 2026-08-05
fix(tn-reth): reject state-mutating TEL precompile selectors under STATICCALL

### LINKED ISSUE #1089: `tn-reth`: the TEL precompile (`0x7e1`) mutates state under `STATICCALL`; `is_static` is never checked in `telcoin_precompile` or any handler
`tn-reth`: the TEL precompile (`0x7e1`) mutates state under `STATICCALL`; `is_static` is never checked in `telcoin_precompile` or any handler

**Severity:** Low.  The EVM's static-call write-protection invariant is broken for one address, but every mutating selector is still gated on authorization, so this is not a path to unauthorized issuance, and every node running the same build takes the same dispatcher branch, so a static-frame mutation is deterministic rather than a source of divergence between honest nodes.  Confirmed by static reading on `origin/main` (`48e00b4b`).

**Attacker model:** not an unauthorized-caller attack.  The mutating selectors are reachable only by `GOVERNANCE_SAFE_ADDRESS` (`burnable.rs:230`, `:327`) or a mint-role holder (`burnable.rs:164`), and an unauthorized caller inside a `STATICCALL` frame still gets `"unauthorized"` exactly as it would under `CALL`.

The precondition is narrow and worth stating before anything else, because it rules out most of the obvious shapes.  The handlers authorize on `input.caller`, and unlike `DELEGATECALL`, a `STATICCALL` does *not* preserve `msg.sender`: the precompile sees the immediate caller.  So the `STATICCALL` has to issue from an authorized address's own execution context, meaning the governance Safe itself, code `DELEGATECALL`ed into it, or (faucet builds) a mint-role-holding contract.  A generic wrapper or simulation relay that staticcalls `0x7e1` on governance's behalf is itself the caller and gets `"unauthorized"`, so that is not a path.  Neither is a plain `eth_call`: the top-level frame of a simulated transaction is not static, so `eth_call` only reaches this if some contract along the simulated path performs the `STATICCALL` itself.

What is left is still a real shape: a read-only probe of a governance action from inside the Safe's own context, which is the ordinary "simulate this first" pattern in Safe module and guard code.  Such a caller can get the full mint/claim/burn/role-change side effects where the EVM promises a revert.  To be explicit about what is and is not being claimed: I have not identified a deployed contract that does this, and no off-chain integrators were surveyed.  The claim is that the precompile does not prevent it and no test says it should not happen, not that something in production does it today.

## Summary

`telcoin_precompile` (`crates/tn-reth/src/evm/tel_precompile/mod.rs:120`) is the dispatcher for the native TEL precompile at `TELCOIN_PRECOMPILE_ADDRESS` (`0x7e1`).  It dispatches purely on the 4-byte selector extracted at `mod.rs:125` and never reads `input.is_static`.  Neither do the handlers: `rg -n "is_static" crates/tn-reth/src/evm/` returns zero matches across the entire `evm/` tree at `48e00b4b`.

The module's own doc comment states the exposure (`mod.rs:10-11`): "Any `CALL`/`STATICCALL` targeting that address is intercepted and routed to [`telcoin_precompile`] instead of executing bytecode."  Because the precompile short-circuits bytecode execution, the interpreter's usual write-protection check never runs on the writes it performs.  That check lives in the instruction path only: `revm-interpreter` 32.0.0 implements it as the `require_non_staticcall!` macro (`src/instructions/macros.rs:3-12`, the `is_static` test at `:8`) that `SSTORE` and `LOG` expand, and there is no second line of defence below it, since `rg -n "is_static" revm-context-13.0.0/src/` returns zero hits, so the journal the precompile writes through has no notion of static context at all.  Enforcement for a precompile is therefore the precompile's own responsibility, and it is not being done.

`PrecompileInput` carries the flag the dispatcher needs.  In the resolved `alloy-evm` 0.27.3 (the workspace requests `alloy-evm = { version = "0.27.2" }` at `Cargo.toml:163` and the lockfile resolves 0.27.3 at `Cargo.lock:331-334`, a plain crates.io registry dependency: there is no `[patch]` section in the workspace and the crate is not vendored here), `src/precompiles.rs:598-599` declares `pub is_static: bool` with the doc comment "Whether this call is in a STATICCALL context", `src/precompiles.rs:473` populates it from revm's `CallInputs` (`is_static: inputs.is_static,`), and `src/precompiles.rs:643-645` exposes it as `is_static_call()`.  `alloy-evm` passes the flag through and offers the accessor; it does not itself enforce write protection for custom stateful precompiles registered via `DynPrecompile::new_stateful` (`mod.rs:111`).  The flag is delivered to this repo and then dropped on the floor.

## Problem

The dispatcher takes the whole `PrecompileInput`, including `is_static`, and branches on the selector alone (`crates/tn-reth/src/evm/tel_precompile/mod.rs:120-165`, the first twenty lines shown):

```rust
fn telcoin_precompile(mut input: PrecompileInput<'_>) -> PrecompileResult {
    if input.data.len() < 4 {
        return Err(PrecompileError::Other("Invalid input: too short".into()));
    }

    let selector: [u8; 4] = input.data[0..4].try_into().unwrap();
    let calldata = &input.data[4..];

    match selector {
        // State-mutating functions
        #[cfg(not(feature = "faucet"))]
        burnable::mintCall::SELECTOR => {
            burnable::handle_mint(&mut input.internals, calldata, input.caller, input.gas)
        }
        burnable::claimCall::SELECTOR => {
            burnable::handle_claim(&mut input.internals, calldata, input.caller, input.gas)
        }
        burnable::burnCall::SELECTOR => {
            burnable::handle_burn(&mut input.internals, calldata, input.caller, input.gas)
        }
```

Note that only `input.internals`, `input.caller` and `input.gas` are forwarded.  `input.is_static` is in scope at every one of these arms and is read by none of them.

Each mutating handler checks gas, then authorization, then calldata length, and then mutates unconditionally.  `handle_claim` (`burnable.rs:220-309`) is the widest: one `balance_incr`, two storage clears, a `totalSupply` read-modify-write, and two log emissions, none of them guarded:

```rust
    // Credit recipient
    balance_incr(internals, recipient, amount)?;

    // Clear storage
    internals
        .sstore(TELCOIN_PRECOMPILE_ADDRESS, amt_slot, U256::ZERO)
        .map_err(|e| PrecompileError::Other(format!("sstore failed: {e:?}").into()))?;
    internals
        .sstore(TELCOIN_PRECOMPILE_ADDRESS, ts_slot, U256::ZERO)
        .map_err(|e| PrecompileError::Other(format!("sstore failed: {e:?}").into()))?;
```

The same shape holds for the rest of the mutating surface:

1. `handle_mint` (`burnable.rs:154-206`): gas at `:161`, `has_mint_role` at `:164`, calldata length at `:167`, then two unguarded `sstore` calls at `:181-189` plus a `Mint` log.
2. `handle_claim` (`burnable.rs:220-309`): gas at `:227`, `has_governance_role` at `:230`, calldata length at `:233`, then `balance_incr` at `:264`, two `sstore` clears at `:267-272`, the `totalSupply` increment at `:279-287`, and `Claim` plus `Transfer` logs at `:297` and `:306`.
3. `handle_burn` (`burnable.rs:317-377`): gas at `:324`, `has_governance_role` at `:327`, calldata length at `:330`, then `balance_decr` at `:337`, the `totalSupply` decrement at `:349-351`, and `Burn` plus `Transfer` logs at `:361` and `:374`.
4. Under the `faucet` feature, `handle_mint_faucet` (`faucet.rs:62`), `handle_grant_mint_role` (`faucet.rs:132`) and `handle_revoke_mint_role` (`faucet.rs:158`) follow the identical gas-then-authorization-then-length-then-mutate pattern, again with no `is_static` reference anywhere in the file.

The read-only selectors `totalSupply()` (`handle_total_supply`, `burnable.rs:83`) and, under the `faucet` feature, `hasMintRole` (`handle_has_mint_role`, `faucet.rs:184`) are legitimately callable under `STATICCALL` and must stay that way; only the mutating arms need rejecting.

## Why the outcome is contained today

1. Authorization gates every mutating selector before any write.  Reaching a mutation under `STATICCALL` still requires being `GOVERNANCE_SAFE_ADDRESS` or holding the mint role, so this is not an unauthorized-issuance bug.
2. There is one client and one dispatcher.  Every node running the same build takes the same branch on the same transaction and performs the same writes, so a `STATICCALL`-context mutation is deterministic and does not fork the chain.  (The one way that premise bends is a build-flag difference in which selectors exist at all, which is #1066's subject, not this one.)  The damage is to the semantics integrators compose against, not to consensus.
3. `DELEGATECALL` into `0x7e1` is pinned by an in-tree regression test (`crates/tn-reth/tests/it/tel_precompile_props.rs:389-415`), so the call-context question has been considered for one of the two non-`CALL` entry paths.  `STATICCALL` is the one that was not.

What this rests on is that no in-tree or integrator code path currently reaches a mutating selector through a static frame.  That is a property of today's callers, not an invariant the precompile enforces, and it is not asserted by any test.

## Suggested fix

Reject static-context mutation once, in the dispatcher, rather than per handler.  In `telcoin_precompile` (`mod.rs:120`), after the length check and selector extraction and before the `match`, or as a guard on the mutating arms, return `PrecompileError::Other("static call: state mutation not permitted")` when `input.is_static` is set.  Leave the `totalSupply()` and `hasMintRole` arms reachable, since those are exactly the selectors a static frame should be able to call.

Hoisting the check into the dispatcher is preferable to duplicating it in each of the three (or, with the `faucet` feature, six) handlers, for three reasons.  It keeps the guard on the same surface that already owns selector classification, so a future selector cannot be added to the mutating group and silently miss the check as long as the mutating arms are grouped as they are today.  It keeps the handlers' signatures unchanged: they currently receive only `internals`, `calldata`, `caller` and `gas`, and threading a fourth flag through six functions is a larger and more invasive diff for no additional safety.  And it puts the rejection before any gas accounting or storage read, which matches how the length check at `mod.rs:121-123` already behaves.

The alternative, a per-handler check, is the more defensive placement if the mutating and read-only selectors are ever interleaved or if handlers grow additional entry points, but it is a larger change and duplicates one condition six ways.  The dispatcher-level guard is the smaller edit and is what this issue proposes.

One detail to settle when implementing: whether the correct signal is `PrecompileError::Other(...)` or a dedicated variant.  The `Other` route matches every other rejection this precompile emits (`mod.rs:122`, `mod.rs:163`, `burnable.rs:231`) and is the lowest-friction choice, but a distinct variant would let callers distinguish "write protection" from "bad calldata".  Either satisfies the acceptance criteria below.

Note that no code in this repository was built or executed while producing this issue; every claim above is from static reading of the tree at `48e00b4b`, and the suggested fix has not been compiled.

## Suggested test design

The closest existing coverage is `test_delegatecall_writes_target_precompile_storage` (`crates/tn-reth/tests/it/tel_precompile_props.rs:389-415`).  It is the right neighbourhood and the wrong opcode: it deploys a hand-assembled relay at `RELAY_ADDR` (`tel_precompile_props.rs:349`) whose bytecode (`:364-370`) forwards calldata to `0x7e1` with `DELEGATECALL` (`0xf4`), sends `mint(1234)` from `GOVERNANCE_SAFE_ADDRESS` through it, and asserts the write lands in the precompile's storage and not the relay's.  It pins *where* the write goes and would pass unchanged whether or not the write is permitted at all, so it cannot catch this.

The unit tests in `mod.rs` do not cover it either: the module has exactly two tests (`mod.rs:168-187`, `test_input_too_short` at `:172` and `test_unknown_selector` at `:179`), both of which only assert a non-success result for malformed input.  The sixteen tests in `burnable.rs:380-589` do reach the dispatcher, but only through `TestEnv::exec` / `exec_default` / `mint` (`crates/tn-reth/src/evm/precompile_test_utils.rs:205-211`, `:241-244`), which builds a `TxEnv` and runs a top-level transaction against `0x7e1` via `MainnetHandler` (`:213-239`).  A top-level transaction frame is never static and `TxEnv` carries no `is_static` to set, so no test in either module can express a `STATICCALL` frame without routing through a deployed relay contract.  There is also no coverage upstream: `alloy-evm` 0.27.3 exercises `is_static` only in fixtures for its own `EthPrecompiles` and generic dispatch tests (`src/precompiles.rs:882`, `:917`, `:953`, `:1028`, `:1063`, `:1093`, all `is_static: false`), never for a stateful custom precompile reached under `STATICCALL`.

A covering test should reuse the relay pattern already proven twice in this repo, at `tel_precompile_props.rs:364-370` for `DELEGATECALL` into `0x7e1` and at `crates/tn-reth/tests/it/bls_precompile_props.rs:283-289` for `STATICCALL` into `0xb151`.  The bytecode needed is the TEL relay with a single opcode changed, `0xf4` to `0xfa`, and the operand kept at `0x07e1`; `STATICCALL` and `DELEGATECALL` take the same six stack operands, so nothing else in the relay moves.

One trap is worth calling out, because it turns the whole test suite vacuous if it is missed.  `STATICCALL` does not preserve `msg.sender`, so a relay deployed at a fresh `RELAY_ADDR` is itself the caller the precompile authorizes against, and every mutating selector would be rejected with `"unauthorized"` whether or not an `is_static` guard exists.  The relay bytecode therefore has to sit at an address the precompile already trusts.  For mainnet builds that means deploying it at `GOVERNANCE_SAFE_ADDRESS` and driving the outer transaction from a different sender such as `USER`, since revm rejects a transaction whose sender has code (EIP-3607).  For faucet builds the simpler option is to grant the mint role to `RELAY_ADDR` first.  Note also that `TestEnv::deploy_code` (`precompile_test_utils.rs:185-203`) writes a fresh `AccountInfo` with zero balance, so anything the assertions depend on at that address has to be re-established after the deploy.

- For each mutating selector (`mint`, `claim`, `burn`, plus the three `faucet`-feature selectors under that feature), invoke it through the static relay such that the precompile's `input.caller` is the authorized address, and assert the call does not succeed.
- Assert the absence of the side effect, not just the error: after the failed `mint`, the pending-amount slot under `TELCOIN_PRECOMPILE_ADDRESS` is still zero; after the failed `claim`, the recipient's balance and `TOTAL_SUPPLY_SLOT` are unchanged and no `Claim` or `Transfer` log was emitted.  An error-only assertion would pass against a fix that reverts after writing.
- Positive control in the same test module: `totalSupply()` invoked through the same static relay still succeeds and returns the expected value, so the fix is not proven by a guard that rejects everything.
- Confirm-by-mutation: revert the `is_static` guard and confirm each new mutating-selector test fails, then restore it.  Without that step the tests are not evidence, because a relay that fails to reach the precompile at all, or that reaches it with an unauthorized `msg.sender`, produces exactly the same non-success result.

## Acceptance criteria

- A `STATICCALL` into `0x7e1` carrying a state-mutating selector fails, from any caller including `GOVERNANCE_SAFE_ADDRESS` and mint-role holders, and leaves the precompile's storage, the affected balances, `TOTAL_SUPPLY_SLOT` and the log set byte-identical to their pre-call values.
- A `STATICCALL` into `0x7e1` carrying `totalSupply()` (and `hasMintRole` under the `faucet` feature) still succeeds and returns the same value it returns under `CALL`.
- Behavior under `CALL` and under `DELEGATECALL` is unchanged for every selector, and `test_delegatecall_writes_target_precompile_storage` still passes unmodified.
- `rg -n "is_static" crates/tn-reth/src/evm/` returns at least one hit in `tel_precompile/`, and reverting that hit makes at least one test fail.

## Related, not duplicate

- **#666** (closed, 2026-05-07), "`0x7e1` Stateful Precompile - DELEGATECALL", raised the risk of the stateful precompile under `DELEGATECALL`; its stated solution was to "Remove erc20 methods and use WETH for wrapped native TEL".  It addressed the `DELEGATECALL` half of the call-context question, and `test_delegatecall_writes_target_precompile_storage` is the in-tree test that now pins that behavior (I have not traced whether that test came out of #666 specifically).  `STATICCALL` write protection is the sibling question and was not addressed there.
- **#1066** (open), the `faucet` cargo feature having no chain-id guard, overlaps only in that it also concerns the `faucet`-gated dispatch arms.  It is about which selectors exist in a given build, not about whether a mutating selector may run in a static frame.
- No open PR currently touches `crates/tn-reth/src/evm/tel_precompile/` (all 22 open PRs checked by changed-file list at the time of writing).

## Provenance and confidence

Surfaced by a hardening sweep over `crates/engine` and `crates/tn-reth` at `origin/main` `48e00b4b` (2026-08-04).  Every line number and quote cited above was checked against the tree at that commit, including the `mod.rs:168-187` test-module boundary (the file is exactly 187 lines), the test harness in `precompile_test_utils.rs`, and the `alloy-evm` 0.27.3 field and accessor.  The `rg` zero-hit result was re-run verbatim.

Three limits worth stating.  Nothing was built or executed for this issue, so the claims are static-analysis claims and the suggested guard has not been compiled or type-checked against `PrecompileInput`'s borrow structure at `mod.rs:120` (the dispatcher takes `mut input` and then reborrows `input.internals` per arm, so exact placement of the read of `input.is_static` may need care).  The "no integrator currently reaches a mutating selector through a static frame" statement in the containment section is scoped to this repository; it is not a claim about deployed Safe modules or off-chain tooling, which were not surveyed.  And the suggested test has not been written or run, so the caller-identity constraint described above is reasoned from the opcode semantics and the harness source, not from a passing test.

## References

- `crates/tn-reth/src/evm/tel_precompile/mod.rs:10-11` (module doc: `CALL`/`STATICCALL` both routed to the dispatcher); `:109-115` (`add_telcoin_precompile`, `DynPrecompile::new_stateful` registration); `:120-165` (`telcoin_precompile`, selector-only dispatch, no `is_static` read); `:121-123` (length check, the existing early rejection); `:125` (selector extraction); `:128-139` (mutating arms); `:141-143`, `:153-156` (read-only arms); `:168-187` (`mod tests`, exactly two tests)
- `crates/tn-reth/src/evm/tel_precompile/burnable.rs:83` (`handle_total_supply`); `:107` (`has_mint_role`); `:136` (`has_governance_role`); `:154-206` (`handle_mint`; gas `:161`, auth `:164`, length `:167`, `sstore` `:181-189`); `:220-309` (`handle_claim`; auth `:230`, `balance_incr` `:264`, `sstore` clears `:267-272`, `totalSupply` `:279-287`, logs `:297`, `:306`); `:317-377` (`handle_burn`; auth `:327`, `balance_decr` `:337`, `totalSupply` `:349-351`, logs `:361`, `:374`); `:380-589` (`mod tests`, 16 tests, all driven as top-level non-static transactions)
- `crates/tn-reth/src/evm/tel_precompile/faucet.rs:62` (`handle_mint_faucet`); `:132` (`handle_grant_mint_role`); `:158` (`handle_revoke_mint_role`); `:184` (`handle_has_mint_role`, read-only)
- `crates/tn-reth/src/evm/precompile_test_utils.rs:185-203` (`TestEnv::deploy_code`, zero-balance account write); `:205-211` (`TestEnv::exec`, targets `TELCOIN_PRECOMPILE_ADDRESS`); `:213-239` (`TestEnv::exec_to`, builds a `TxEnv` and runs `MainnetHandler`); `:241-244` (`TestEnv::exec_default`, 100,000 gas)
- `crates/tn-reth/tests/it/tel_precompile_props.rs:349` (`RELAY_ADDR`); `:364-370` (`DELEGATECALL` relay bytecode, `0xf4`); `:389-415` (`test_delegatecall_writes_target_precompile_storage`, the nearest existing coverage)
- `crates/tn-reth/tests/it/bls_precompile_props.rs:268-289` (`STATICCALL` relay address and bytecode, `0xfa`, the pattern to reuse); `:293-316` (`test_staticcall_verify_pop`, an existing `STATICCALL`-through-a-relay test)
- `alloy-evm` 0.27.3 `src/precompiles.rs:473` (`is_static: inputs.is_static,`, population from revm's `CallInputs`); `:598-599` (`pub is_static: bool`, "Whether this call is in a STATICCALL context"); `:643-645` (`is_static_call()`); `:882`, `:917`, `:953`, `:1028`, `:1063`, `:1093` (its own `is_static: false` fixtures, generic dispatch only) (`Cargo.lock:331-334`, crates.io registry dependency)
- `revm-interpreter` 32.0.0 `src/instructions/macros.rs:3-12` (`require_non_staticcall!`, the interpreter-only write-protection check, `is_static` test at `:8`); `revm-context` 13.0.0 `src/` (zero `is_static` hits: the journal has no static-context notion) (`revm` 34.0.0, `Cargo.lock:10014-10016`)

All line numbers hold on `origin/main` `1f33e42c`.  The sweep read the tree at `48e00b4b`; `main` has advanced two commits since (`f8bdb648`, then `1f33e42c` = the merge of #1025), and none of the files cited above changed in either, so every anchor above is byte-identical at both commits.

MavenRain:
Closes #1089.

## Problem

`0x7e1` is registered as a stateful precompile, so any `CALL` or `STATICCALL` targeting it is intercepted and routed to `telcoin_precompile` instead of executing bytecode.  Short-circuiting bytecode execution also short-circuits the EVM's static-call write protection: `require_non_staticcall!` lives in the interpreter's instruction path (the macro `SSTORE` and `LOG` expand through), and the journal beneath it carries no static-context flag.  For a precompile there is no second line of defence, so enforcement is the precompile's own responsibility.

It was not being done.  The dispatcher branched on the 4-byte selector alone and never read `input.is_static`, and neither did any handler . . . `rg -n "is_static" crates/tn-reth/src/evm/` returned zero matches across the whole tree.  `PrecompileInput` carries the flag (`alloy-evm` 0.27.3 populates `is_static` from revm's `CallInputs` and exposes `is_static_call()`), so it was delivered and dropped.  Every mutating selector (`mint`, `claim`, `burn`, and with the `faucet` feature `mint`, `grantMintRole`, `revokeMintRole`) would perform its full set of `sstore`, balance, and log side effects inside a frame where the EVM promises a revert.

**Threat model.** This is not an unauthorized-issuance path, and no attacker is required for it to matter.  Authorization still gates every mutating selector before any write, and `STATICCALL` (unlike `DELEGATECALL`) does not preserve `msg.sender`, so the precompile authorizes against the immediate caller.  A wrapper that staticcalls `0x7e1` on governance's behalf is itself the caller and still gets `"unauthorized"`.  The reachable shape is narrower and entirely honest: a read-only probe issued from an authorized address's own execution context, which is the ordinary "simulate the action first" pattern in Safe module and guard code.  Such a caller gets the real mint/claim/burn side effects where it expected a revert.  There is also no consensus-divergence exposure here . . . one client, one dispatcher, so every node takes the same branch and the mutation is deterministic.  The damage is to the semantics integrators compose against.

The gap was in coverage, not just in code.  `DELEGATECALL` into `0x7e1` is already pinned by a regression test, so the call-context question had been asked for one of the two non-`CALL` entry paths.  `STATICCALL` was the one that was not, and no test could express it: `TestEnv` drives top-level transactions, a top-level frame is never static, and `TxEnv` carries no `is_static` to set.  Reaching a static frame at all requires routing through a deployed relay contract.

## Fix

`crates/tn-reth/src/evm/tel_precompile/mod.rs` . . . classify the selector, then refuse mutation in a static frame before dispatch:

- [x] Read-only selectors (`totalSupply()`, plus `hasMintRole` under the `faucet` feature) are named explicitly and stay callable inside a `STATICCALL`.  Both were confirmed write-free: each only `sload`s.
- [x] Everything else the dispatcher recognises is refused when `input.is_static` is set, via a single `?` guard placed before the existing `match`.  The guard runs ahead of any gas accounting or storage read, matching how the existing length check already behaves.
- [x] The classification is by **read-only**, not by **mutating**, so the failure direction is safe by construction: a selector added to the dispatcher later is guarded unless it is explicitly named read-only.  Forgetting to classify a new mutating selector cannot produce an unguarded write.
- [x] Rejection uses `PrecompileError::Other`, consistent with every other rejection this precompile emits.  Documented consequence: an unrecognised selector inside a static frame now reports the static-call message rather than `"Unknown function selector"`.  Both are `PrecompileError::Other` and both revert the frame, so only the text differs.
- [x] Module and dispatcher docs record the invariant and why the check cannot live in the interpreter.

The guard is hoisted rather than duplicated across the three (or, with `faucet`, six) handlers: it keeps handler signatures unchanged, avoids threading a fourth argument through six functions, and keeps one condition in one place.

No caller is broken by this.  The only in-tree Solidity path that reaches a mutating selector is `StablecoinManager::_drip`, which uses a low-level `.call(...)` (`tn-contracts/src/testnet/StablecoinManager.sol:211`), not a staticcall.

## Testing

`crates/tn-reth/tests/it/tel_precompile_props.rs` . . . two tests built on the relay pattern already proven twice in this repo (the `DELEGATECALL` relay for `0x7e1`, and the `STATICCALL` relay for `0xb151`).

- [x] `test_staticcall_cannot_mutate_precompile_state` . . . `mint` through a `STATICCALL` relay is refused, and the pending-mint slot stays zero.
- [x] `test_staticcall_still_serves_read_only_selectors` . . . `totalSupply()` through the same relay still succeeds, so the guard is not over-broad.

Three details decide whether these tests mean anything, and the obvious version of the test gets all three wrong:

1. **The relay is hosted at `GOVERNANCE_SAFE_ADDRESS`.** `STATICCALL` does not preserve `msg.sender`, so a relay at a fresh address is itself the caller and every mutating selector would be rejected with `"unauthorized"` whether or not the guard exists . . . the test would pass against unfixed code. The outer transaction is driven by `USER` instead, since revm rejects a transaction whose sender has code (EIP-3607).
2. **The relay returns the inner call's success flag** rather than `POP`ping it. The existing `DELEGATECALL` relay discards that flag and forwards the precompile's returndata, which cannot express this distinction: a successful `mint` and a rejected one both return empty bytes, and the outer transaction succeeds either way because the relay always `RETURN`s. Asserting on outer transaction success would have been vacuous.
3. **Each relay gets its own `TestEnv`.** Within one env the journal caches account code across transactions while `deploy_code` writes to the database underneath it, so swapping the relay bytecode mid-test does not take effect and the second transaction silently re-runs the first relay. The first draft of this test did exactly that and reported the `STATICCALL` as succeeding; the failure was the harness, not the guard.

The mutating test carries its own positive control: the identical relay under `CALL` (differing only in the zero `value` operand and `0xf1` against `0xfa`) mints successfully and writes `1234`, so authorization and the write path are proven live before the `STATICCALL` case asserts they do not fire. Its assertion message says so explicitly, so a future regression that breaks the control reports itself rather than passing quietly.

Commands run, `cargo fmt` on the pinned nightly and the rest on the repo-pinned 1.94:

```
cargo +nightly-2026-03-20 fmt -- --check                          # clean
cargo +1.94 clippy -p tn-reth --all-targets                       # clean, 0 warnings
cargo +1.94 clippy -p tn-reth --features faucet --all-targets     # clean, 0 warnings
cargo +1.94 test -p tn-reth --test it                             # 49 passed, 0 failed
cargo +1.94 test -p tn-reth --features faucet --test it           # 37 passed, 0 failed
```

Confirmed by mutation, not just by passing: with the guard reverted to `(read_only || true)`, `test_staticcall_cannot_mutate_precompile_state` fails on its "must be refused by the precompile" assertion while the read-only test and the pre-existing `DELEGATECALL` test keep passing. This was then confirmed a second time by accident, which is the stronger evidence: a stale `libtn_reth` built *without* the guard got linked into the test binary, and the test failed with exactly that assertion and no other. It detects a missing guard.

This also settles empirically what the issue could only reason about statically: revm does propagate `is_static` through `alloy-evm` into `PrecompileInput` for a stateful custom precompile, so the flag was genuinely available and unread.

Not run here: the full `make attest` workspace gate. A concurrent build held the shared target-dir lock throughout, so the numbers above required forcing a rebuild of `tn-reth` from this worktree to avoid linking a sibling branch's artifacts. That is exactly why `-p <crate>` against the shared target is not a valid attest proxy, and the isolated `--workspace` build is still owed before pushing.

### DIFF
--- a/crates/tn-reth/src/evm/tel_precompile/mod.rs
+++ b/crates/tn-reth/src/evm/tel_precompile/mod.rs
@@ -8,7 +8,10 @@
 //!
 //! The precompile is registered as a [`DynPrecompile`] inside a [`PrecompilesMap`] at
 //! [`TELCOIN_PRECOMPILE_ADDRESS`] (`0x7e1`). Any `CALL`/`STATICCALL` targeting that address is
-//! intercepted and routed to [`telcoin_precompile`] instead of executing bytecode.
+//! intercepted and routed to [`telcoin_precompile`] instead of executing bytecode. Bypassing
+//! bytecode execution also bypasses the interpreter's static-call write protection, so
+//! [`telcoin_precompile`] enforces that itself: inside a `STATICCALL` frame it serves the
+//! read-only selectors and refuses every state-mutating one.
 //!
 //! # Module structure
 //!
@@ -37,6 +40,8 @@
 //!   `claim`, `burn`, and (with faucet) `grantMintRole`/`revokeMintRole`.
 //! - **Mint-role holders** (faucet only): can call the faucet `mint(address, uint256)`.
 //! - **Any account**: can call `totalSupply()` (read-only).
+//! - **Static frames**: regardless of caller, only the read-only selectors (`totalSupply()`, plus
+//!   `hasMintRole` with the faucet feature) are served inside a `STATICCALL`.
 //!
 //! # Feature flags
 //!
@@ -114,9 +119,30 @@ pub fn add_telcoin_precompile(map: &mut PrecompilesMap) {
     });
 }
 
+/// Rejection emitted when a state-mutating selector is reached inside a `STATICCALL` frame.
+const STATIC_CALL_MUTATION: &str = "static call: state mutation not permitted";
+
 /// Top-level dispatcher for the Telcoin precompile.
 ///
 /// Extracts the 4-byte selector from calldata and routes to the matching handler.
+///
+/// # Static-call write protection
+///
+/// [`totalSupplyCall`] and (with the `faucet` feature) [`hasMintRoleCall`] only read state, so they
+/// stay callable inside a `STATICCALL` frame. Every other selector dispatched below writes storage,
+/// native balances, or logs, and is refused when [`PrecompileInput::is_static`] is set.
+///
+/// Enforcing this here is not redundant with the interpreter. Registering this address as a
+/// precompile short-circuits bytecode execution, so the `require_non_staticcall!` check that
+/// `SSTORE` and `LOG` expand through never runs for calls routed to [`telcoin_precompile`], and the
+/// journal beneath it carries no static-context flag to catch the writes either. Write protection
+/// for a precompile is the precompile's own responsibility.
+///
+/// Classifying by *read-only* rather than by *mutating* selector is deliberate: a selector added to
+/// the dispatcher later is guarded unless it is explicitly named read-only here, so the failure
+/// direction is a refused read rather than an unguarded write. One consequence is that an
+/// unrecognised selector inside a static frame reports [`STATIC_CALL_MUTATION`] rather than
+/// `"Unknown function selector"`; both are [`PrecompileError::Other`] and both revert the frame.
 fn telcoin_precompile(mut input: PrecompileInput<'_>) -> PrecompileResult {
     if input.data.len() < 4 {
         return Err(PrecompileError::Other("Invalid input: too short".into()));
@@ -124,6 +150,15 @@ fn telcoin_precompile(mut input: PrecompileInput<'_>) -> PrecompileResult {
 
     let selector: [u8; 4] = input.data[0..4].try_into().unwrap();
     let calldata = &input.data[4..];
+
+    // Selectors that touch no state. Everything else the `match` below recognises mutates.
+    // `hasMintRole` is dispatched only under the `faucet` feature, so it counts as read-only only
+    // there: this list mirrors the read-only arms of that `match` exactly, under both feature sets.
+    let read_only = matches!(selector, burnable::totalSupplyCall::SELECTOR)
+        || (cfg!(feature = "faucet") && matches!(selector, hasMintRoleCall::SELECTOR));
+    (read_only || !input.is_static)
+        .then_some(())
+        .ok_or_else(|| PrecompileError::Other(STATIC_CALL_MUTATION.into()))?;
 
     match selector {
         // State-mutating functions
--- a/crates/tn-reth/tests/it/main.rs
+++ b/crates/tn-reth/tests/it/main.rs
@@ -5,6 +5,7 @@
 mod bls_precompile_props;
 mod economics_props;
 mod pipeline_helpers;
+mod precompile_relays;
 mod recover_drop_props;
 
 // testnet
--- a/crates/tn-reth/tests/it/precompile_relays.rs
+++ b/crates/tn-reth/tests/it/precompile_relays.rs
@@ -1,0 +1,65 @@
+//! Hand-assembled relay contracts for reaching the TEL precompile from a nested call frame.
+//!
+//! A top-level transaction frame is never static and `TxEnv` carries no `is_static` to set, so a
+//! `STATICCALL` into `0x7e1` can only be expressed by routing through deployed bytecode. These
+//! relays are shared by the mainnet and `faucet` precompile test modules so the two builds exercise
+//! byte-identical call paths.
+//!
+//! Both relays end by storing the inner call's success flag at memory word 0 and returning it, so a
+//! test can tell "the precompile accepted this" from "the precompile refused it". The
+//! `DELEGATECALL` relay in `tel_precompile_props.rs` instead `POP`s that flag and forwards the
+//! precompile's returndata, which cannot express this distinction: an accepted mutation and a
+//! refused one both return empty bytes, and the outer transaction succeeds either way because the
+//! relay always `RETURN`s.
+//!
+//! # Hosting address
+//!
+//! `STATICCALL` does not preserve `msg.sender`, so the precompile authorizes against the relay's
+//! own address. A relay at a fresh address is rejected with `"unauthorized"` whether or not any
+//! static guard exists, which would make a write-protection test pass vacuously. Deploy these at an
+//! address the precompile already trusts and drive the outer transaction from a different sender,
+//! since revm rejects a transaction whose sender has code (EIP-3607).
+
+/// Runtime bytecode forwarding calldata to `0x7e1` with `CALL`, returning the inner success flag.
+///
+/// The positive control for the `STATICCALL` write-protection tests: it differs from
+/// [`STATICCALL_RELAY_BYTECODE`] only in pushing a zero `value` operand (`CALL` takes seven stack
+/// operands, `STATICCALL` six) and in the call opcode itself, `0xf1` against `0xfa`.
+///
+/// Disassembly:
+/// ```text
+///   CALLDATASIZE; PUSH1 0; PUSH1 0; CALLDATACOPY   // mem[0..csize] = calldata
+///   PUSH1 0; PUSH1 0; CALLDATASIZE; PUSH1 0;       // retSize, retOffset, argsSize, argsOffset
+///   PUSH1 0; PUSH2 0x07e1; GAS; CALL               // value, address, gas
+///   PUSH1 0; MSTORE                                // mem[0..32] = success flag
+///   PUSH1 32; PUSH1 0; RETURN                      // return that word
+/// ```
+pub(crate) const CALL_RELAY_BYTECODE: &[u8] = &[
+    0x36, 0x60, 0x00, 0x60, 0x00, 0x37, // CALLDATACOPY(0, 0, CALLDATASIZE)
+    0x60, 0x00, 0x60, 0x00, 0x36, 0x60, 0x00, // retSize, retOffset, argsSize, argsOffset
+    0x60, 0x00, // value = 0
+    0x61, 0x07, 0xe1, 0x5a, 0xf1, // PUSH2 0x07e1; GAS; CALL
+    0x60, 0x00, 0x52, // MSTORE(0, success)
+    0x60, 0x20, 0x60, 0x00, 0xf3, // RETURN(0, 32)
+];
+
+/// Runtime bytecode forwarding calldata to `0x7e1` with `STATICCALL`, returning the success flag.
+///
+/// Identical to [`CALL_RELAY_BYTECODE`] but for the dropped `value` operand and `0xfa` in place of
+/// `0xf1`.
+///
+/// Disassembly:
+/// ```text
+///   CALLDATASIZE; PUSH1 0; PUSH1 0; CALLDATACOPY   // mem[0..csize] = calldata
+///   PUSH1 0; PUSH1 0; CALLDATASIZE; PUSH1 0;       // retSize, retOffset, argsSize, argsOffset
+///   PUSH2 0x07e1; GAS; STATICCALL                  // address, gas
+///   PUSH1 0; MSTORE                                // mem[0..32] = success flag
+///   PUSH1 32; PUSH1 0; RETURN                      // return that word
+/// ```
+pub(crate) const STATICCALL_RELAY_BYTECODE: &[u8] = &[
+    0x36, 0x60, 0x00, 0x60, 0x00, 0x37, // CALLDATACOPY(0, 0, CALLDATASIZE)
+    0x60, 0x00, 0x60, 0x00, 0x36, 0x60, 0x00, // retSize, retOffset, argsSize, argsOffset
+    0x61, 0x07, 0xe1, 0x5a, 0xfa, // PUSH2 0x07e1; GAS; STATICCALL
+    0x60, 0x00, 0x52, // MSTORE(0, success)
+    0x60, 0x20, 0x60, 0x00, 0xf3, // RETURN(0, 32)
+];
--- a/crates/tn-reth/tests/it/tel_precompile_faucet_props.rs
+++ b/crates/tn-reth/tests/it/tel_precompile_faucet_props.rs
@@ -5,17 +5,20 @@
 //! - Mint role grant/revoke semantics
 //! - Supply accounting with faucet mints
 
+use crate::precompile_relays::{CALL_RELAY_BYTECODE, STATICCALL_RELAY_BYTECODE};
 use alloy::sol_types::SolCall;
 use proptest::prelude::*;
 use reth_revm::primitives::{address, Address};
 use tn_config::GOVERNANCE_SAFE_ADDRESS as GOVERNANCE;
 use tn_reth::{
-    claimCall, grantMintRoleCall, mintCall, revokeMintRoleCall,
+    claimCall, faucet_mint_role_slot, grantMintRoleCall, hasMintRoleCall, mintCall,
+    revokeMintRoleCall,
     test_utils::precompile_test_utils::{
-        assert_not_success, assert_success, TestEnv, RECIPIENT, USER,
+        assert_not_success, assert_success, decode_bool, TestEnv, RECIPIENT, USER,
     },
+    TELCOIN_PRECOMPILE_ADDRESS,
 };
-use tn_types::U256;
+use tn_types::{Bytes, U256};
 
 const FAUCET: Address = address!("0000000000000000000000000000000000000F00");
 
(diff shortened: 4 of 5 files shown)


grantkee (approved):
LGTM - good catch

PR #1109, opened 2026-08-06
Docs and tests for the engine/epoch code, plus two bugs the effort surfaced

grantkee:
- Removed an un-gated `debug_assert_eq!` in `execute_consensus_output` that shadowed the fail-closed uneven-batch check and panicked in every debug-assertions build (including the e2e profile and `adiri` feature) for epochs ≤ `ADIRI_DUP_BATCH_EPOCH`; added tests covering both the rejected and adiri-tolerated arms.
- Removed three never-constructed `TnEngineError` variants (`SealBlockWithSenders`, `MissingSigner`, `MissingFinalBlock`) whose names implied checks the engine crate doesn't perform — `MissingFinalBlock` implied fork detection that doesn't exist. Fixed a broken intra-doc link and an unescaped generic parsed as HTML in the same pass; crate docs now build clean under `--document-private-items`. `batch-builder` carried an identically dead twin of `SealBlockWithSenders` — same never-constructed status, same phantom length check implied by its message — removed for the same reason.
- Documented `crates/engine`'s trust boundary at the crate root and README: which `ConsensusOutput` inputs are re-validated versus trusted from Bullshark, and where each check actually lives.
- Made `retry_provider_faults`'s pin an explicit generic parameter (`&'pin P`) instead of relying on callers to resolve it outside the closure by convention (previously 4 prose comments, zero enforcement); threaded it through all 8 production call sites in `node.rs`, `run_epoch.rs`, `close_epoch.rs`, `start_epoch.rs`. Added tests for pin identity across retry attempts, a real provider-fault-then-halt path, and retry metric emission.
- Documented `worker_id_from_header`'s one production consumer (`catchup_accumulator`) and its second implementation to stay in sync with (`BlockCtx::worker_id`), and expanded the premises under which the `Eip1559` base-fee read is value-identical to the whole-epoch derivation it replaced.
- Rewrote `export_once`'s rustdoc for its entry-readiness precondition, corrected the snapshot-restore boundary rule, and added an e2e assertion that the recorded entry fee survives export/import.
- Added a test for a worker-count shrink through `EpochBaseFees::apply` and pinned the live entry-read's oversized-data halt with its context; a follow-up pass caught that the shrink test incremented the wrong worker's gas counters, so its "untouched by apply" assertion held regardless of what `apply` did — fixed to increment the surviving worker and mutation-tested against a `set_num_workers` that clears instead of truncates.
- Brought the TEL precompile README's STATICCALL section in line with the static-call guard added earlier in the branch (#1107): the intro still described unconditional dispatch, the access-control table had no static-frame row, and Security considerations covered the analogous DELEGATECALL hazard but not the more consequential STATICCALL one that the guard actually closes.
- Two review passes over the branch's own documentation (`README`s and rustdoc written in the commits above) turned up seventeen accumulated inaccuracies — stale line citations, an enumeration that dropped a real call site or recovery path, a warm/cold mislabel, a check described as holding for a reader variant it doesn't — and corrected each in place. The production changes and test assertions themselves were unaffected; only the prose was wrong.


crates/e2e-tests/tests/it/state_export_import.rs
```diff
@@ -684,3 +703,415 @@ async fn test_state_export_import_idle_epoch_inner() -> eyre::Result<()> {
...
+    );
+
+    // ---- 5) Control: the live network prices the import epoch's successor at `expected`.
+    // `concludeEpoch` records the entered epoch's first block as `closing block + 1`, so
+    // `import_block + 1` is the first block of epoch `import_epoch + 1`. Asserting it here pins the
+    // block the recorded word priced *from the chain*, so the observer comparison below is against
+    // a block whose expected fee is established rather than assumed.
+    let first_new_block = import_block + 1;
+    let network_fee = read_base_fee(&client_urls[0], first_new_block)?;
+    assert_eq!(
```
MavenRain, line 1041: This assertion succeeds only when there is one idle boundary between gas and import epoch.  If there are two (FEE_WARMUP_FLOOR seems to budget for this), then this could flaky-fail.

crates/e2e-tests/tests/it/state_export_import.rs
```diff
@@ -684,3 +703,415 @@ async fn test_state_export_import_idle_epoch_inner() -> eyre::Result<()> {
...
+
+    // ---- 7) THE HEADLINE ASSERTION. The observer's only source for epoch `import_epoch + 1`'s fee
+    // is the word recorded in the imported snapshot block: it holds no pre-`B` history to walk and
+    // nothing carried in memory. A failed entry read would have left it on MIN
+    // (`MIN_PROTOCOL_BASE_FEE`, ruled out by the guard) or on the previous boundary's word
+    // (`stale`, also ruled out) — and either would put a different `base_fee_per_gas` in its
+    // executed block than the one the network committed at the same height.
+    let obs_fee = read_base_fee(&obs_url, first_new_block)?;
+    assert_eq!(
+        obs_fee,
```
MavenRain, line 1101: I don't think that this can distinguish between a legitimate restore and a stale or MIN fee because the value is inherited from the imported header (as opposed to captured from worker config, for example).

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -1070,6 +1087,179 @@ mod tests {
...
+    ///
+    /// The retry evidence here is real elapsed time — 2 x [`CLOSE_READ_RETRY_BACKOFF`] — so this
+    /// test deliberately runs on a real clock. `#[tokio::test(start_paused = true)]` is not usable:
+    /// a paused clock auto-advances past both backoffs (erasing the evidence), and its auto-advance
+    /// only fires when the runtime has nothing else to poll, which a live `RethEnv` does not
+    /// guarantee.
+    ///
+    /// The entry read's OTHER retried site ("epoch-entry base fees") is deliberately not covered
+    /// by a fault test: it pins the very header the guard above already resolved, so a pin that
+    /// faults there faults the guard first — there is no way to fail the second read without
```
MavenRain, line 1316: Claude says the second read (`get_worker_fee_configs_at_block`) does a header index lookup via `pinned_header_by_hash` that the guard path doesn't do, so that a transient fault lands on the second read alone.

MavenRain (dismissed):
A couple of potential issues, though no technical blockers

PR #1111, opened 2026-08-06
fix(types): clamp current_base_fee before alloy's unchecked fee-increase add (#1108)

### LINKED ISSUE #1108: `compute_next_base_fee_eip1559` can overflow `u64`, silently collapsing a worker's next-epoch base fee in release builds
## Summary

`compute_next_base_fee_eip1559` (`crates/types/src/gas_accumulator.rs:455`) bounds its inputs in every direction except one: it never clamps `current_base_fee` from above.  It then delegates to alloy's `calc_next_block_base_fee`, whose fee-increase arm performs a plain unchecked `u64` addition.  When `current_base_fee` exceeds roughly `u64::MAX * 8 / 9` and the epoch ran over its gas target, that addition overflows.

The two build profiles disagree about what happens next, which is the part that matters:

- `dev` / `test` / `e2e` have `overflow-checks = true`, so the node panics ("attempt to add with overflow").
- `release` leaves `overflow-checks` at the Cargo default of `false`, so the addition wraps silently and the function returns a base fee far below the one it was asked to increase.

The wrapped value is `sum - 2^64`, so the size of the collapse depends on how far past the threshold the input sits.  This is a consensus-affecting value, and because every honest node computes the same wrapped result, the collapse validates cleanly across the network rather than forking it.

The function's own doc comment at line 454 states the guarantee this breaks:

> The result is clamped to `[MIN_PROTOCOL_BASE_FEE, u64::MAX]`.

The lower half of that clamp exists (`.max(MIN_PROTOCOL_BASE_FEE)` at line 464).  The upper half does not.

## Environment

- Verified on `origin/main` at commit `e7a23bcb`.
- `alloy-eips` 1.8.3 (the version resolved in `Cargo.lock`).
- Profile settings from `.cargo/config.toml`: `dev`, `test` and `e2e` set `overflow-checks = true`;  there is no `[profile.release]` override, so release keeps the Cargo default of `false`.

## The code

`crates/types/src/gas_accumulator.rs:455-465`:

```rust
pub fn compute_next_base_fee_eip1559(current_base_fee: u64, gas_used: u64, target_gas: u64) -> u64 {
    if target_gas == 0 {
        return current_base_fee.max(MIN_PROTOCOL_BASE_FEE);
    }

    let params = BaseFeeParams::ethereum();
    let gas_limit = target_gas.saturating_mul(params.elasticity_multiplier as u64);
    let gas_used = gas_used.min(gas_limit);
    let new_base_fee = calc_next_block_base_fee(gas_used, gas_limit, current_base_fee, params);
    new_base_fee.max(MIN_PROTOCOL_BASE_FEE)
}
```

The defense against alloy's unguarded arithmetic is already half-built here, and the doc comment at lines 449-452 says so explicitly: `gas_used` is clamped to `gas_limit` to "avoid the unbounded delta arithmetic alloy would otherwise produce when callers pass `gas_used > gas_limit`".  The `current_base_fee` axis into that same arithmetic was left unguarded.

The overflow itself is in `alloy-eips` 1.8.3, `src/eip1559/helpers.rs:109`, in the `Ordering::Greater` arm:

```rust
core::cmp::Ordering::Greater => {
    base_fee
        + (core::cmp::max(
            1,
            base_fee as u128 * (gas_used - gas_target) as u128
                / (gas_target as u128 * base_fee_params.max_change_denominator),
        ) as u64)
}
```

Note the `max(1, ..)` floor.  It means the delta is never zero on an over-target epoch, so even `current_base_fee == u64::MAX` with `gas_used == target_gas + 1` overflows.  (The `max(1, ..)` floor itself is canonical EIP-1559 and matches go-ethereum's `math.BigMax(num, common.Big1)`.  It is not a defect, but it does remove the "the increment rounds down to zero at the ceiling" escape hatch one might otherwise expect.)

## Reproduction

Extracting both functions verbatim and compiling the same source twice:

```
$ rustc -C opt-level=2 -C overflow-checks=off overflow_demo.rs   # release semantics
$ rustc -C opt-level=2 -C overflow-checks=on  overflow_demo.rs   # dev/test/e2e semantics
```

Measured results, with `target_gas = 15_000_000` and `gas_used` at the elasticity bound (the maximal +12.5% move):

| input `current_base_fee` | `overflow-checks=on` | `overflow-checks=off` |
| --- | --- | --- |
| `16_397_105_843_297_379_214` (threshold minus 1) | `18_446_744_073_709_551_615` | same |
| `16_397_105_843_297_379_215` (threshold) | panic | `7` |
| `u64::MAX`, `gas_used = target_gas + 1` | panic | wraps |
| `u64::MAX`, maximal move | panic | `2_305_843_009_213_693_950` |

The threshold is `16_397_105_843_297_379_215`, which is `u64::MAX * 8 / 9` as expected from a `1/8` maximum increase.

At the exact threshold the sum equals `2^64` and wraps to `0`, so `.max(MIN_PROTOCOL_BASE_FEE)` returns `7`.  Further past the threshold the wrap lands higher: at `u64::MAX` the true sum is `20_752_587_082_923_245_566` and the wrapped result is `2_305_843_009_213_693_950`, an approximately 87.5% collapse that the floor does not catch.  Neither case is a panic in a release build, and neither is signalled to the operator.

The demo program is attached at the end of this issue.

## Reachability

Here are the three paths I checked, in decreasing order of plausibility:

**1. Governance configuration (most plausible).** `decode_worker_fee_configs` in `crates/tn-reth/src/system_calls.rs:409-411` reads the strategy and value straight out of the `WorkerConfigs` contract and applies no bound to `value` in either arm:

```rust
let config = match strategy {
    0 => WorkerFeeConfig::Eip1559 { target_gas: value },
    1 => WorkerFeeConfig::Static { fee: value },
```

So a worker configured `Static { fee }` with a very large `fee`, then switched by governance to `Eip1559`, carries that value into `current_base_fee` and overflows on the next over-target epoch.  That is two ordinary owner transactions with no congestion required.  I could not audit the contract-side validation because the Solidity sources are not in this repository, so I cannot say whether `setWorkerConfig` bounds the value it stores.  If it does, this path closes and the finding drops to defense in depth.  **This is the question I would most like a maintainer to answer.**

**2. Organic growth (implausible on its own).** Climbing from `MIN_PROTOCOL_BASE_FEE` at the maximum +12.5% per epoch takes **363 consecutive fully congested epochs** to reach the threshold.  At the 24 hour epoch duration in `crates/tn-reth/src/env/epoch.rs`, that is about a year of uninterrupted maximum congestion.  Not a realistic attack on its own, but it does mean the ceiling is not unreachable in principle.

**3. Downstream absorption (none).** Nothing re-checks the value after the fact.  `BatchValidator::validate_basefee` (`crates/batch-validator/src/validator.rs:188-195`) compares for equality only:

```rust
fn validate_basefee(&self, base_fee: u64) -> BatchValidationResult<()> {
    let expected_base_fee = self.base_fee;
    if base_fee != expected_base_fee {
        Err(BatchValidationError::InvalidBaseFee { expected_base_fee, base_fee })
    } else {
        Ok(())
    }
}
```

Since every node recomputes the same wrapped value, the collapsed fee is what they all expect, and validation passes.

## Related: `Static` bypasses `MIN_PROTOCOL_BASE_FEE`

While tracing the above I noticed the same value-bound gap in the other direction, at two sites.  Both `Static` arms return the configured fee unmodified, with no protocol floor:

- `next_base_fee_for_config`, `crates/types/src/gas_accumulator.rs:488`: `WorkerFeeConfig::Static { fee } => fee,`
- `entry_fee_for_worker`, `crates/types/src/gas_accumulator.rs:514`: `WorkerFeeConfig::Static { fee } => Ok(fee),`

The sibling `Eip1559` arms both apply `.max(MIN_PROTOCOL_BASE_FEE)` (lines 464 and 523), and the decode path above applies no bound either, so `Static { fee: 0 }` produces zero-priced blocks that validate cleanly on every node for the same equality-only reason.

This may well be intentional: the doc at line 470 says `Static` "pins the fee to the governance-set value", and `static_config_pins_to_configured_fee` asserts exactly that.  But the values that test pins are `12_345` and `500`, both above the floor, so the below-floor case is neither tested nor documented as deliberate.  The invariant stated at line 110, "a recorded fee is never below `MIN_PROTOCOL_BASE_FEE`", is scoped to written rows and so is not violated by this, but it does show the floor is treated as a protocol-wide property elsewhere in the same file.

Worth an explicit decision either way: apply the floor, or document that `Static` is deliberately exempt from it.

## Minor: the saturation path drifts from its documented contract

For `target_gas > u64::MAX / 2`, `target_gas.saturating_mul(2)` saturates and alloy recovers `u64::MAX / 2` rather than the configured target, falsifying the doc's "so alloy recovers the same `gas_target`" at line 449. Measured consequences at `target_gas = u64::MAX`: usage exactly equal to the target raises the fee by 12.5% (the doc says usage at target leaves it unchanged), and usage at 54% of target still raises it.

I am flagging this as latent rather than live.  It requires a single epoch's `gas_used` to approach `9.2e18` to matter, which the per-block gas limit makes unreachable today.  It is a doc accuracy issue now, and a correctness issue only if the gas ceiling ever moves.

## Suggested fix

Clamping `current_base_fee` from above before the alloy call closes the overflow:

```rust
/// The largest base fee for which alloy's unchecked `base_fee + delta` cannot overflow:
/// the maximum delta is `base_fee / 8`, so `base_fee + base_fee / 8 <= u64::MAX` requires
/// `base_fee <= u64::MAX * 8 / 9`.
const SAFE_MAX_BASE_FEE: u64 = u64::MAX - u64::MAX / 8;

let current_base_fee = current_base_fee.min(SAFE_MAX_BASE_FEE);
```

That restores the upper half of the documented `[MIN_PROTOCOL_BASE_FEE, u64::MAX]` clamp and is the direct analogue of the `gas_used.min(gas_limit)` line already sitting above it.

Clamping only at the point of use leaves the out-of-range value stored, though, so the more robust shape is to reject or bound it where it enters: in `decode_worker_fee_configs`, and in `BaseFeeContainer::set_base_fee`.  The same pass could apply the `MIN_PROTOCOL_BASE_FEE` floor to the `Static` arms if the maintainers decide that gap should close.

This is also the shape the repository already uses for this exact hazard.  `crates/tn-reth/src/evm/handler.rs` guards a comparable multiplication with `saturating_mul` and pins it with a mutation-guard test whose doc reads: "with the multiplication reverted to a raw `*` it fails (overflow-checks panic in the test profile, a wrapped and therefore wrong credit in a release build)."

## Suggested tests

The existing tests never place `current_base_fee` near the ceiling.  Three cases would pin the behavior:

1. `compute_next_base_fee_eip1559(u64::MAX, target * 2, target)` returns a value rather than panicking, and that value is `>= current_base_fee` (an increase request must never decrease the fee).
2. The same at the exact threshold `16_397_105_843_297_379_215`, which is where the wrap currently lands hardest.
3. `target_gas >= 2^63` behaves per the documented contract, or the doc is corrected to describe the saturating behavior.

Each should be confirmed by mutation: with the proposed `.min(SAFE_MAX_BASE_FEE)` removed, cases 1 and 2 must fail.  Note that under the test profile they will fail by panicking rather than by asserting, since `overflow-checks` is on there.  That is worth being deliberate about, because it is precisely the difference between the test profile and the release profile that hides this defect.

## Appendix: reproduction program

```rust
use std::cmp::Ordering;

const MIN_PROTOCOL_BASE_FEE: u64 = 7;
const ELASTICITY: u64 = 2;
const MAX_CHANGE_DENOM: u128 = 8;

/// Verbatim alloy 1.8.3 arithmetic, including the unchecked `base_fee + (..)`.
fn calc_next_block_base_fee(gas_used: u64, gas_limit: u64, base_fee: u64) -> u64 {
    let gas_target = gas_limit / ELASTICITY;
    match gas_used.cmp(&gas_target) {
        Ordering::Equal => base_fee,
        Ordering::Greater => {
            base_fee
                + (std::cmp::max(
                    1,
                    base_fee as u128 * (gas_used - gas_target) as u128
                        / (gas_target as u128 * MAX_CHANGE_DENOM),
                ) as u64)
        }
        Ordering::Less => base_fee.saturating_sub(
            (base_fee as u128 * (gas_target - gas_used) as u128
                / (gas_target as u128 * MAX_CHANGE_DENOM)) as u64,
        ),
    }
}

/// Verbatim telcoin wrapper.
fn compute_next_base_fee_eip1559(current_base_fee: u64, gas_used: u64, target_gas: u64) -> u64 {
    if target_gas == 0 {
        current_base_fee.max(MIN_PROTOCOL_BASE_FEE)
    } else {
        let gas_limit = target_gas.saturating_mul(ELASTICITY);
        let gas_used = gas_used.min(gas_limit);
        calc_next_block_base_fee(gas_used, gas_limit, current_base_fee).max(MIN_PROTOCOL_BASE_FEE)
    }
}

fn main() {
    let target: u64 = 15_000_000;
    let full = target * 2;
    let threshold: u64 = 16_397_105_843_297_379_215;

    println!("overflow-checks on in this build? {}", cfg!(overflow_checks));
    println!("below threshold : {}", compute_next_base_fee_eip1559(threshold - 1, full, target));
    println!("at threshold    : {}", compute_next_base_fee_eip1559(threshold, full, target));
    println!("at u64::MAX     : {}", compute_next_base_fee_eip1559(u64::MAX, full, target));
}
```

Compile twice to see the split:

```
rustc -C opt-level=2 -C overflow-checks=off demo.rs -o demo_release && ./demo_release
rustc -C opt-level=2 -C overflow-checks=on  demo.rs -o demo_test    && ./demo_test
```

MavenRain:
## Summary

`compute_next_base_fee_eip1559` clamps `gas_used` from above and clamps its result from below, but never clamps `current_base_fee` from above.  It then hands that value to alloy's `calc_next_block_base_fee`, whose fee-increase arm performs a plain unchecked `u64` addition.  This adds the missing upper clamp, closing the overflow and restoring the `[MIN_PROTOCOL_BASE_FEE, u64::MAX]` guarantee the function's own doc comment already promised.

Fixes #1108.

## Problem

alloy's `calc_next_block_base_fee` (`alloy-eips` 1.8.3, `src/eip1559/helpers.rs`) increases a fee with `base_fee + delta`, unchecked:

```rust
core::cmp::Ordering::Greater => {
    base_fee
        + (core::cmp::max(
            1,
            base_fee as u128 * (gas_used - gas_target) as u128
                / (gas_target as u128 * base_fee_params.max_change_denominator),
        ) as u64)
}
```

The wrapper clamped one of the two inputs into that arithmetic and not the other.  `gas_used.min(gas_limit)` was already there, and the doc comment explains it exists to "avoid the unbounded delta arithmetic alloy would otherwise produce", but `current_base_fee` went in unbounded.  Once `current_base_fee` exceeds `u64::MAX * 8 / 9` and the epoch ran over its gas target, the addition overflows.  alloy's `max(1, ..)` delta floor means there is no escape hatch at the ceiling: even `u64::MAX` with `gas_used == target_gas + 1` overflows.

The two build profiles then disagree, which is the part that matters:

- `dev` / `test` / `e2e` set `overflow-checks = true`, so the node panics.
- `release` has no `[profile.release]` override, so it keeps Cargo's default of `false`, the addition wraps, and the function returns a base fee far *below* the one it was asked to increase.

Because the wrap is deterministic, every honest node computes the same collapsed fee, and `BatchValidator::validate_basefee` compares for equality only.  The bad value validates cleanly network-wide instead of forking: a silent consensus-value collapse with nothing signalled to the operator.

**Reachability.** The `WorkerConfigs` owner surface is `setWorkerConfig(uint16 workerId, uint8 strategy, uint64 value, uint184 data)`, which sets the strategy and the raw `data` word in a single call.  On epoch entry, `entry_fee_for_worker`'s `Eip1559` arm reads that word back as the worker's base fee, rejecting only words wider than `u64`:

```rust
let recorded = u64::try_from(entry.data).map_err(|_| { .. })?;
Ok(recorded.max(MIN_PROTOCOL_BASE_FEE))
```

Any word in `(u64::MAX * 8 / 9, u64::MAX]` passes that guard and becomes `current_base_fee`.  The repo's own `entry_fee_eip1559_accepts_u64_max_word` test pins exactly this: a `data` word of `u64::MAX` returns `Ok(u64::MAX)`.  So one owner transaction is enough to arm the overflow, and the next over-target epoch triggers it.  No congestion required.

(Issue #1108 proposed a different route: a `Static { fee }` value carried into `Eip1559` by a strategy switch.  That mechanism does not hold, because the `Eip1559` arm reads the recorded `data` word rather than the config's `value`.  The `data` argument above reaches the same place more directly, in one transaction rather than two.)

## Fix

One clamp, in the one shared fee formula, mirroring the `gas_used.min(gas_limit)` line directly above it:

```rust
let current_base_fee = current_base_fee.min(SAFE_MAX_BASE_FEE);
```

`SAFE_MAX_BASE_FEE` is `u64::MAX - u64::MAX / 8`, doc-commented with the bound's proof: with `gas_used` already clamped to `gas_limit`, `gas_limit - gas_target` never exceeds `gas_target`, so alloy's delta never exceeds `base_fee / 8`, and `base_fee + base_fee / 8` stays inside `u64`.  The bound holds in the saturating-`target_gas` regime too, where `gas_target` becomes `u64::MAX / 2` rather than the configured target.

The clamp lives at the point of use rather than at the entry points because both consensus seams that price a worker's next-epoch fee, `record_next_epoch_base_fees` (which writes the on-chain word inside the closing block) and `adjust_base_fees` (the live accumulator at close), route through this single function.  Clamping here fixes both symmetrically and keeps them byte-for-byte identical by construction.

**Safety.** For every `current_base_fee <= SAFE_MAX_BASE_FEE`, `.min(SAFE_MAX_BASE_FEE)` is a literal no-op, so the change cannot alter any value the network has ever computed or would compute for any in-range input.  It changes behaviour only on inputs that previously panicked or wrapped.

The doc comment's claim that the synthetic `gas_limit` makes "alloy recover the same `gas_target`" is also corrected in passing: for `target_gas > u64::MAX / 2` the multiply saturates and alloy recovers `u64::MAX / 2` instead.  That is a documentation accuracy fix, not a behaviour change.

## Testing

Three tests added to `crates/types/src/gas_accumulator.rs`, all pinned by mutation.  With the `.min(SAFE_MAX_BASE_FEE)` line removed they fail, and under the test profile's `overflow-checks` they fail by panicking rather than by asserting, which is precisely the profile difference that hid the defect:

- `ceiling_base_fee_never_overflows` is the mutation guard.  It covers the exact threshold (`16_397_105_843_297_379_215`, where the unclamped sum is `2^64` and wraps to `0`), `u64::MAX` at the maximal move, and the smallest over-target move that still takes alloy's floored delta.  It asserts that a fee increase never collapses the fee.
- `clamp_does_not_change_in_range_fees` pins the no-op argument, including `SAFE_MAX_BASE_FEE` itself as the largest unaffected input, and one below it still moving the full 12.5%.
- `ceiling_base_fee_never_overflows_with_saturating_target` covers the other axis into alloy's arithmetic, where `target_gas` saturates the synthetic `gas_limit`.

The bound was additionally verified out-of-tree against a verbatim copy of alloy 1.8.3's arithmetic with `checked_add` substituted for the unchecked `+`, over a boundary sweep (1,760 cases across both `target_gas` regimes) and a 3M-case random sweep: zero overflows with the clamp, 165 overflows without it, zero divergence below the threshold, and no case where an over-target epoch decreased the fee.

## Out of scope

Issue #1108 raises a second gap: the `Static` arms of `next_base_fee_for_config` and `entry_fee_for_worker` return the governance fee without the `MIN_PROTOCOL_BASE_FEE` floor their `Eip1559` siblings apply.  That is documented behaviour ("`Static` pins the fee to the governance-set value"), is unrelated to the overflow's causal chain, and changing it would alter a governance-visible value, so it is left for a separate decision rather than folded in here.

### DIFF
--- a/crates/types/src/gas_accumulator.rs
+++ b/crates/types/src/gas_accumulator.rs
@@ -437,6 +437,23 @@ impl GasAccumulator {
     }
 }
 
+/// The largest `current_base_fee` [`compute_next_base_fee_eip1559`] may hand to alloy.
+///
+/// alloy's [`calc_next_block_base_fee`] adds its delta to `base_fee` with a plain `+`. That
+/// addition is unchecked, so it panics under `overflow-checks` (the `dev`, `test` and `e2e`
+/// profiles) and wraps silently under `release`, where a wrapped fee is a consensus value every
+/// node agrees on and therefore validates cleanly instead of forking.
+///
+/// With `gas_used` already clamped to `gas_limit`, the delta alloy computes is at most
+/// `base_fee * (gas_limit - gas_target) / (gas_target * 8)`, and `gas_limit - gas_target` never
+/// exceeds `gas_target`, so the delta never exceeds `base_fee / 8`. Bounding `base_fee` by
+/// `u64::MAX - u64::MAX / 8` therefore keeps `base_fee + base_fee / 8` inside `u64`.
+///
+/// This bound holds in the saturating regime too. For a `target_gas > u64::MAX / 2` the synthetic
+/// `gas_limit` saturates at `u64::MAX`, alloy recovers `gas_target = u64::MAX / 2`, and
+/// `gas_limit - gas_target` still does not exceed `gas_target`, so the `/ 8` ceiling is unchanged.
+const SAFE_MAX_BASE_FEE: u64 = u64::MAX - u64::MAX / 8;
+
 /// EIP-1559-style base fee adjustment computed once per epoch.
 ///
 /// Compares `gas_used` (total gas consumed by a single worker during the epoch)
@@ -446,10 +463,13 @@ impl GasAccumulator {
 /// Delegates the formula to alloy's [`calc_next_block_base_fee`] using
 /// [`BaseFeeParams::ethereum`] (`elasticity_multiplier = 2`,
 /// `max_change_denominator = 8`). The synthetic `gas_limit` passed to alloy is
-/// `target_gas * 2` so alloy recovers the same `gas_target`. `gas_used` is
-/// clamped to `gas_limit` to enforce the EIP-1559 elasticity bound and avoid
-/// the unbounded delta arithmetic alloy would otherwise produce when callers
-/// pass `gas_used > gas_limit`.
+/// `target_gas * 2` so alloy recovers the same `gas_target`, except for a
+/// `target_gas > u64::MAX / 2`, where the multiply saturates and alloy instead
+/// recovers `u64::MAX / 2`. Both inputs into alloy's delta arithmetic are
+/// clamped so that arithmetic cannot leave `u64`: `gas_used` to `gas_limit`,
+/// enforcing the EIP-1559 elasticity bound, and `current_base_fee` to
+/// [`SAFE_MAX_BASE_FEE`], bounding the unchecked addition alloy performs on the
+/// fee-increase arm.
 ///
 /// The result is clamped to `[MIN_PROTOCOL_BASE_FEE, u64::MAX]`.
 pub fn compute_next_base_fee_eip1559(current_base_fee: u64, gas_used: u64, target_gas: u64) -> u64 {
@@ -460,6 +480,7 @@ pub fn compute_next_base_fee_eip1559(current_base_fee: u64, gas_used: u64, targe
     let params = BaseFeeParams::ethereum();
     let gas_limit = target_gas.saturating_mul(params.elasticity_multiplier as u64);
     let gas_used = gas_used.min(gas_limit);
+    let current_base_fee = current_base_fee.min(SAFE_MAX_BASE_FEE);
     let new_base_fee = calc_next_block_base_fee(gas_used, gas_limit, current_base_fee, params);
     new_base_fee.max(MIN_PROTOCOL_BASE_FEE)
 }
@@ -590,6 +611,56 @@ mod tests {
         let base = 1_000_000_000_000_000u64;
         let result = compute_next_base_fee_eip1559(base, u64::MAX, 1);
         assert_eq!(result, base + base / 8);
+    }
+
+    /// Mutation guard for the `.min(SAFE_MAX_BASE_FEE)` clamp: with the clamp removed, alloy's
+    /// unchecked `base_fee + delta` overflows on every case here: a panic under the test
+    /// profile's `overflow-checks`, and a wrapped (far lower) fee in a release build.
+    #[test]
+    fn ceiling_base_fee_never_overflows() {
+        let target = 15_000_000u64;
+
+        // the exact threshold: the unclamped sum is 2^64, which wraps to 0 and would surface as
+        // MIN_PROTOCOL_BASE_FEE after the floor
+        let threshold = 16_397_105_843_297_379_215u64;
+        let at_threshold = compute_next_base_fee_eip1559(threshold, target * 2, target);
+        assert!(
+            at_threshold >= SAFE_MAX_BASE_FEE,
+            "a fee increase must never collapse the fee: {at_threshold}"
+        );
+
+        // the widest word `entry_fee_for_worker` accepts, at the maximal +12.5% move
+        let at_max = compute_next_base_fee_eip1559(u64::MAX, target * 2, target);
+        assert_eq!(at_max, SAFE_MAX_BASE_FEE + SAFE_MAX_BASE_FEE / 8);
+
+        // the smallest over-target move still takes alloy's `max(1, ..)` floored delta
+        let barely_over = compute_next_base_fee_eip1559(u64::MAX, target + 1, target);
+        assert!(barely_over >= SAFE_MAX_BASE_FEE, "must not wrap: {barely_over}");
+    }
+
+    /// The clamp is a no-op for every fee the protocol can actually hold, so no epoch that did not
+    /// already overflow changes value. `SAFE_MAX_BASE_FEE` itself is the largest unaffected input.
+    #[test]
+    fn clamp_does_not_change_in_range_fees() {
+        let target = 15_000_000u64;
+        assert_eq!(compute_next_base_fee_eip1559(1000, target * 2, target), 1125);
+        assert_eq!(
+            compute_next_base_fee_eip1559(SAFE_MAX_BASE_FEE, target * 2, target),
+            SAFE_MAX_BASE_FEE + SAFE_MAX_BASE_FEE / 8
+        );
+        // one below the clamp still moves by the full 12.5%
+        let below = SAFE_MAX_BASE_FEE - 1;
+        assert_eq!(compute_next_base_fee_eip1559(below, target * 2, target), below + below / 8);
+    }
+
+    /// A saturating `target_gas` is the other axis into alloy's delta arithmetic; the clamp has to
+    /// hold there too, where `gas_target` is `u64::MAX / 2` rather than the configured target.
+    #[test]
+    fn ceiling_base_fee_never_overflows_with_saturating_target() {
+        [u64::MAX / 2 + 1, u64::MAX - 1, u64::MAX].iter().for_each(|&target| {
+            let out = compute_next_base_fee_eip1559(u64::MAX, u64::MAX, target);
+            assert!(out >= SAFE_MAX_BASE_FEE, "target {target} wrapped to {out}");
+        });
     }
 
     #[test]


grantkee (approved):
Nice, good catch

PR #1112, opened 2026-08-06
fix(tn-reth): bound state-pack export memory with streamed storage chunks

### LINKED ISSUE #1091: `tn-reth`: `export_state_pack` holds a whole-account storage `BTreeMap` plus a never-evicted `code_cache`, so export-side peak memory scales with chain state
`tn-reth`: `export_state_pack` holds a whole-account storage `BTreeMap` plus a never-evicted `code_cache`, so export-side peak memory scales with chain state

**Severity:** Low.  Availability only, on an opt-in path: the export runs only when the operator boots with `--enable-state-export` (a `clap` flag defaulting to `false`, `crates/telcoin-network-cli/src/node.rs:61-64`).  No state divergence, no effect on the produced pack's contents, and no effect at all on a node that leaves the flag off.  The failure mode is a resident-memory spike during the plain-state walk on a node that has opted in; if the host is memory-constrained the OOM killer takes the whole process, since the exporter runs on a thread inside the node process (`crates/node/src/manager/node/close_epoch.rs:324-326`).  To be exact about the blast radius of that: nothing here changes what consensus *decides*, but an opted-in node that gets OOM-killed is down, and if that node is a validator its liveness is the thing lost until it restarts.

**Attacker model:** the two allocations are sized by chain state, and chain state is writable by anyone willing to pay gas: deploying distinct contracts grows one sink, writing slots to an attacker-deployed contract with an unbounded mapping grows the other.  Nothing caps slots per account in the EVM or in `tn-reth`.  What is *not* open: reaching the code at all requires operator opt-in, there is no remote or RPC trigger today (the module doc names an on-demand trigger source only as a future possibility, `crates/node/src/manager/export_exec.rs:17-19`), and the growth is slow and expensive to drive, so this is a hardening item rather than a cheap remote DoS.

**Confidence:** this is a static finding.  The unbounded growth is structural and readable from the source, but no OOM was reproduced and no peak memory was measured; the byte figures below are arithmetic on the container's element size, not measurements.  Nothing here was compiled.

## Summary

`PinnedStateView::export_state_pack` (`crates/tn-reth/src/snapshot.rs:157-248`) walks `PlainAccountState` and `PlainStorageState` as a merge-join and emits one pack record stream.  Inside that single loop (`:182-243`) there are two allocations whose size is a function of chain state and which have no bound, no budget, and no eviction:

1. `storage: BTreeMap<B256, B256>` (`:185`), rebuilt per account, into which *every* non-zero storage slot for that address is drained (`:186-202`) before a single `writer.append_account(&account)` call (`:238`).  Peak is set by the largest single account's non-zero slot count.
2. `code_cache: HashMap<B256, Bytes>` (`:170`), declared once before the loop and populated at `:219`, read at `:206`.  It is never cleared or evicted (`rg code_cache crates/tn-reth/src/snapshot.rs` returns exactly those three lines).  Peak is set by the number of distinct bytecodes in the entire account table, and every entry stays live until `export_state_pack` returns.

Both contradict the discipline the same file documents for the opposite direction.  `import_state`'s `# Memory profile` block (`:581-587`) states that on restore, "At most one account header plus one `STORAGE_CHUNK_SLOTS`-bounded storage chunk is resident at a time", and that "a whale account's storage is never fully materialized".  The writer API needed to hold the export to the same standard is already public: `ExecStatePackWriter::append_account_header` (`crates/storage/src/exec_state_pack.rs:303`) plus `append_storage_chunk` (`:326`), whose own doc says "the caller never needs to hold an account's full storage in memory" (`:299-302`).  `export_state_pack` calls neither; it calls the `append_account` convenience wrapper (`:272`), whose signature requires an already-materialized `ExecStateAccount` and therefore cannot bound the caller's working set.

To be precise about what is *not* broken: the pack's **on-disk** layout is already bounded.  `append_account` splits the storage map it is handed into `STORAGE_CHUNK_SLOTS`-sized records before writing (`exec_state_pack.rs:279-295`, constant at `:65` is `64 * 1024`).  The defect is confined to the exporter's **in-RAM working set**, which is a separate axis from the record layout.

## Problem

The memo, declared once for the whole export (`crates/tn-reth/src/snapshot.rs:170`; its explanatory comment at `:167-169` is elided here because it wraps):

```rust
let mut code_cache: HashMap<B256, Bytes> = HashMap::new();
```

The loop body, abridged (`crates/tn-reth/src/snapshot.rs:182-243`; bytecode resolution at `:204-226` and the `ExecStateAccount` construction at `:228-237` are summarized, comments elided):

```rust
while let Some((address, account)) = pending_account {
    let mut storage: BTreeMap<B256, B256> = BTreeMap::new();
    while let Some((storage_address, entry)) = pending_storage {
        match storage_address.cmp(&address) {
            Ordering::Less => {
                pending_storage = storage_cursor.next().map_err(ProviderError::from)?;
            }
            Ordering::Equal => {
                if !entry.value.is_zero() {
                    storage.insert(entry.key, B256::from(entry.value.to_be_bytes::<32>()));
                }
                pending_storage = storage_cursor.next().map_err(ProviderError::from)?;
            }
            Ordering::Greater => break,
        }
    }

    // ...:204-226 resolve `code` through `code_cache`, inserting on miss...
    // ...:228-237 build `ExecStateAccount` with `storage: (!storage.is_empty()).then_some(storage)`...

    writer.append_account(&account).map_err(|e| { /* ... */ })?;

    pending_account = account_cursor.next().map_err(ProviderError::from)?;
}
```

The memo's miss path, which is what makes the retention unbounded (`crates/tn-reth/src/snapshot.rs:206-222`, abridged):

```rust
let bytes = match code_cache.get(&hash) {
    Some(bytes) => bytes.clone(),
    None => {
        let bytecode = tx.get::<Bytecodes>(hash).map_err(ProviderError::from)?
            .ok_or_else(|| { /* ...:212-217 missing-bytecode error... */ })?;
        let bytes = bytecode.original_bytes();
        code_cache.insert(hash, bytes.clone());
        bytes
    }
};
```

Step by step:

1. `code_cache` is allocated at `:170`, before the loop, and lives until the function returns at `:245-247`.  Nothing in the function removes an entry.
2. For each account, `storage` is allocated fresh at `:185` and the dup cursor is drained into it at `:186-202` until an address greater than the current one is seen (`:200`).  Every non-zero slot for that account is resident simultaneously at the end of that inner drain.
3. Bytecode is resolved at `:204-226`.  On a miss, the deployed bytes are read from `Bytecodes` (`:209-217`) and inserted into the memo at `:219`.  Whether `Bytes` clones share a refcounted buffer or copy it, the memo is what keeps every distinct bytecode buffer alive to the end of the walk.
4. The full map is moved into `account.account.storage` at `:234` and the account is handed to a single `writer.append_account` call at `:238`.  Peak residency for that account is reached *before* the writer sees anything, so the writer's internal chunking (`exec_state_pack.rs:279-295`) cannot reduce it.
5. The loop advances (`:242`) and `storage` drops, but `code_cache` does not.
6. Wall-clock exposure is long by design: `pin_state_view` calls `disable_long_read_transaction_safety()` precisely because a full-state export can outrun MDBX's roughly five-minute reader timeout (`snapshot.rs:11-16`).  The memo is live for that entire window.

Trigger path, confirmed live rather than from the module doc alone: `run_epoch` calls `self.export_epoch_state(&primary, &reth_env).await?` unconditionally on the epoch-boundary success path (`crates/node/src/manager/node/run_epoch.rs:417-418`), and `export_epoch_state` short-circuits only when no exporter was spawned (`crates/node/src/manager/node/close_epoch.rs:345-347`), which happens only when the flag is unset (`crates/node/src/manager/node.rs:766`).  There is no second opt-in gate: once `--enable-state-export` is on, every epoch boundary runs the walk.

## Blast radius: the two sinks scale on different axes

They are worth treating as two items, not one, because their bounds differ:

- **`storage` (`:185`) is per-account.** It is rebuilt each iteration, so its peak is the single largest account's non-zero slot count, independent of how many accounts exist.  Raw key plus value bytes alone are 64 per slot, so ten million slots on one contract is roughly 640 MB before any `BTreeMap` node overhead, and a hundred million is roughly 6.4 GB.  A single popular token or registry contract is the shape that reaches those counts first; whether any account on this chain is near that size today is not something this finding measured, so read those figures as what the structure permits, not as a present-day reading.
- **`code_cache` (`:170`) is whole-export.** Its peak is the count of distinct code hashes across all of plain state, times the deployed size of each.  Per-entry size is capped by the EVM's deployed-code limit; the entry *count* is not capped by anything, and it is the axis an attacker can inflate most directly, since deploying a distinct contract is a single transaction.  That cap is also the honest counterweight: because each entry is small, driving this sink to gigabyte scale means a correspondingly large number of separate deployments, so the axis is unbounded without being cheap.  Its comment (`:167-169`) justifies the memo as a deduplication win for proxies sharing bytecode, which it is; what is missing is a ceiling.

Consequently a fix for one does not fix the other, and either alone leaves an unbounded axis.

## Suggested fix

Two independent changes; they can land separately.

**Option A (structural, for the per-account map): drive the export through the streaming writer pair that already exists.** Replace the collect-then-`append_account` shape with one `append_account_header` call per account (`exec_state_pack.rs:303`) followed by `append_storage_chunk` calls (`:326`) issued as the dup cursor is drained, flushing every `STORAGE_CHUNK_SLOTS` slots and once more at the end of the account.  This is the path the writer's own doc advertises for exactly this purpose (`:299-302`, and `append_account`'s doc points at it explicitly at `:269-271`).  Two things to settle before landing it.  First, chunk ordering is part of the pack's determinism: `append_account` gets ascending key order for free from `BTreeMap` iteration (`:280-282`), so the replacement must establish that the dup-sorted `PlainStorageState` cursor yields slots in the same ascending subkey order per address, and assert it rather than assume it.  Second, `STORAGE_CHUNK_SLOTS` is private to `tn-storage` (`:65`), so either export the constant or, cleaner, add a writer-side entry point that accepts a slot iterator and owns the chunking internally, keeping the constant private and keeping the chunk-boundary invariant in one place.  The zero-value filter currently at `snapshot.rs:194` applies per slot and carries over unchanged either way.

**Option B (small, for the memo): give `code_cache` a ceiling, or delete it.** A byte-budgeted LRU keeps the proxy-dedup benefit the comment at `:167-169` describes while stating a maximum.  Deleting the memo outright is the smaller diff: the cost is one `Bytecodes` point read per contract account (`:209-211`), which is a B-tree lookup on an already-open, already-pinned read transaction that is walking both full tables anyway.  That trade is unmeasured, so if the memo turns out to be load-bearing for export throughput on a proxy-heavy state, the LRU variant keeps the dedup and still states a maximum.  Deleting also removes the need to pick a budget number, which is the only real design question in this option.

Whichever lands, add a `# Memory profile` doc block to `export_state_pack` stating the resulting bound, mirroring the one `import_state` already carries at `:581-587`, so the invariant is written down on both sides and a future change has something to violate.

## Suggested test design

The neighbourhood is already covered, which is plausibly why this slipped through.

`import_streams_whale_storage_in_chunks` (`crates/tn-reth/src/snapshot.rs:1716-1717`) builds a whale account with `MULTI_CHUNK_SLOTS` slots, defined as `4 * 64 * 1024 + 7` at `:1696`, that is 262,151 slots, and it *does* drive the real exporter: it calls the `export_pack` helper at `:1823`, which calls `PinnedStateView::export_state_pack` (`:1259-1265`).  So `export_state_pack` is exercised today with an account whose storage spans five `STORAGE_CHUNK_SLOTS` records, and it passes.

It would not catch this for two reasons.  Its only export-side assertion is `big_chunks > 1` (`:1838-1841`), computed by reopening the finished pack and counting `Storage` records.  That is an assertion about the on-disk layout, which `append_account` guarantees unconditionally (`exec_state_pack.rs:279-295`) no matter how much the caller held in RAM, so it is satisfied by the current code and would be equally satisfied by a fixed version; it is blind to the axis at issue.  Everything else in the test is an import-side assertion (root equality at `:1858-1861`, empty change sets, plain and hashed row spot checks at `:1898`, deep-slot and last-slot readbacks at `:1943-1951`).  Separately, the fixture size is too small to fail on: 262,151 slots is about 16 MiB of raw key and value bytes, which no CI box notices even if an assertion existed.

`export_roundtrips_accounts_storage_and_code` (`:1055`) is the other relevant test.  It deliberately exercises the memo, using `contract` and `contract_twin` sharing the same `CODE` with the comment "shares the same code as `contract`, exercising the bytecode memo" (`:1058`), but it asserts only that the bytes round-trip.  Retention is not observed.  Across the whole file there is no occurrence of "peak" or "OOM", and the only three case-insensitive hits for "memory" are `:327`, `:581` and `:587`, all doc prose (`:581` and `:587` being the `import_state` block quoted above).

What a covering test should assert:

- Export an account whose storage spans several `STORAGE_CHUNK_SLOTS` chunks and assert that the exporter's largest live slot buffer is chunk-bounded, not account-bounded.  A structural assertion is cheaper and more deterministic than a peak-RSS probe: after Option A, assert the export issues one header plus the expected chunk count for that account, using a writer test double or a counting wrapper, so the test fails if a future change reverts to collecting the whole account first.
- Assert a stated ceiling on bytecode retention: export a fixture with more distinct bytecodes than the chosen budget and assert the cache never exceeds it (Option B with an LRU), or assert the memo is gone and the `Bytecodes` read count equals the contract-account count (Option B by deletion).
- Regression guard on output equivalence: the pack produced after the change must be byte-identical to the pack produced before it for the same state, so `export_roundtrips_accounts_storage_and_code` and `import_streams_whale_storage_in_chunks` must both pass unchanged, with no edit to their assertions.
- Confirm by mutation: revert the fix so the loop rebuilds the full per-account map, confirm the new working-set test fails, then restore.  A test never seen to fail is not evidence, and given that the existing whale test passes under the current code, that check matters more than usual here.

## Acceptance criteria

- Exporting an account with N non-zero storage slots holds at most a `STORAGE_CHUNK_SLOTS`-bounded slot buffer in the exporter's working set, for any N; no container allocated by `export_state_pack` is sized by a single account's slot count.
- Bytecode retained across an export is bounded by a stated budget or is not retained at all, so exporter memory no longer scales with the number of distinct code hashes in state.
- The pack bytes produced for a given state are unchanged by the fix: same account order, same slot order, same chunk boundaries, with the existing roundtrip and whale tests passing without edits to their assertions.
- A test drives the export of a multi-chunk account and fails if the exporter materializes the account's full storage, and that failure has been observed by reverting the fix.
- `export_state_pack` carries a `# Memory profile` doc block stating its bound, matching the one `import_state` carries at `crates/tn-reth/src/snapshot.rs:581-587`.

## Related, not duplicate

- **#1076 (open)**, "state-export completion task runs an unbounded `std::fs::copy` (plus sync dir removals) on tokio runtime worker threads", which **PR #1081 (open)** is filed to close.  Same feature and the same `--enable-state-export` opt-in, different resource and different file: #1076 is blocking file I/O on runtime worker threads in the completion task in `close_epoch.rs`, this is resident memory in the plain-state walk in `tn-reth`.  Neither fix touches the other's code.  They do compound operationally, since both fire at the same epoch boundary on the same node.
- **PR #1087 (open)**, "bound per-boundary state-export work with a certified-prefix watermark and incremental bundles", is the nearest in-flight work and is worth reading alongside this.  It bounds the *consensus-bundle* side of the same epoch-boundary export (the `0..tip_epoch` cert scan and the per-boundary bundle rebuild, in `close_epoch.rs` and `crates/storage/src/epoch_records.rs`).  It does not touch `snapshot.rs`, so the plain-state walk's working set is exactly as described above after it lands.
- No open PR touches `crates/tn-reth/src/snapshot.rs` or `crates/storage/src/exec_state_pack.rs` (checked file-by-file across all 22 PRs open on 2026-08-04).  PRs **#1072**, **#1081** and **#1087** touch `crates/node/src/manager/node/close_epoch.rs`, which contains the trigger cited above but not the export walk itself.

## References

- `crates/tn-reth/src/snapshot.rs:157-248` (`export_state_pack`), `:11-16` (module doc: `disable_long_read_transaction_safety`, roughly five-minute MDBX reader timeout), `:167-169` (memo rationale), `:170` (`code_cache` declaration), `:182-243` (merge-join loop), `:185` (per-account `BTreeMap`), `:186-202` (slot drain, `:194` zero filter, `:200` break to next account), `:204-226` (bytecode resolution), `:206` (memo read), `:209-217` (`Bytecodes` point read), `:219` (memo insert), `:228-237` (`ExecStateAccount` build, `:234` storage move), `:238` (single `append_account` call), `:581-587` (`import_state` `# Memory profile`), `:327` and `:587` (the file's only "memory" occurrences, both prose)
- `crates/tn-reth/src/snapshot.rs:1055` (`export_roundtrips_accounts_storage_and_code`), `:1058` (memo-exercising comment), `:1259-1265` (`export_pack` helper), `:1696` (`MULTI_CHUNK_SLOTS = 4 * 64 * 1024 + 7`), `:1716-1717` (`import_streams_whale_storage_in_chunks`), `:1823` (that test's export call), `:1838-1841` (`big_chunks > 1`, the only export-side assertion), `:1858-1861`, `:1898`, `:1943-1951` (import-side assertions)
- `crates/storage/src/exec_state_pack.rs:65` (`STORAGE_CHUNK_SLOTS = 64 * 1024`, private), `:265-297` (`append_account`, with `:269-271` pointing callers at the streaming pair and `:279-295` doing the on-disk chunking), `:299-302` (streaming-path doc: "the caller never needs to hold an account's full storage in memory"), `:303-321` (`append_account_header`), `:326-331` (`append_storage_chunk`)
- `crates/node/src/manager/export_exec.rs:17-19` (exporter spawned on `--enable-state-export`, triggered at each epoch boundary; on-demand trigger noted as future work)
- `crates/node/src/manager/node/run_epoch.rs:417-418` (unconditional `export_epoch_state` call on the epoch-boundary success path)
- `crates/node/src/manager/node/close_epoch.rs:324-326` (export runs on the exporter's background thread), `:340-347` (no-op only when no exporter was spawned)
- `crates/node/src/manager/node.rs:766` (exporter spawned only when the flag is set)
- `crates/telcoin-network-cli/src/node.rs:61-64` (`--enable-state-export`, `default_value_t = false`)

All line numbers hold on `origin/main` `1f33e42c`.  The sweep read the tree at `48e00b4b`; `main` has advanced two commits since (`f8bdb648`, then `1f33e42c` = the merge of #1025), and none of the files cited above changed in either, so every anchor above is byte-identical at both commits.

MavenRain:
Closes #1091.

## Problem

`PinnedStateView::export_state_pack` held two allocations sized by chain state, with no bound and no eviction:

1. A per-account `storage: BTreeMap<B256, B256>` that materialized every non-zero slot of an account before a single `append_account` call, so peak memory tracked the largest account's slot count (a whale contract's storage was fully resident before the writer saw its first byte).
2. A whole-export `code_cache: HashMap<B256, Bytes>` memo that was never cleared, so peak memory also tracked the number of distinct bytecodes in all of plain state.

Both sinks grow with state that anyone can extend for gas (slots on an attacker-deployed contract; one deployment per distinct bytecode), while the walk itself runs with MDBX's long-read guard disabled, so the retention window is the entire export. The blast radius is availability only, on the opt-in `--enable-state-export` path: an OOM kill takes down the whole node process, and if that node is a validator its liveness is what is lost until restart. Consensus decisions and pack contents are unaffected, and a node without the flag is untouched.

This also contradicted the discipline the same file already documents for the opposite direction: `import_state`'s `# Memory profile` promises at most one header plus one `STORAGE_CHUNK_SLOTS`-bounded chunk resident at a time.

## Fix

Drive the export through the same chunk-bounded streaming the import already has, and delete the memo:

- The merge join over `PlainAccountState` / `PlainStorageState` now streams each account's slots straight from the dup cursor into per-chunk buffers flushed at `STORAGE_CHUNK_SLOTS`, emitting one `append_account_header` plus bounded `append_storage_chunk_owned` calls. No container is ever sized by an account's slot count.
- `code_cache` is gone. Bytecode is one `Bytecodes` point read per contract account against the already-pinned transaction (a B-tree lookup on a transaction that is walking both full tables anyway), so nothing retained scales with distinct-bytecode count. This is the smaller-diff variant of the issue's Option B; if export throughput on a proxy-heavy state ever demands the dedup back, a byte-budgeted LRU can be layered on later.
- The chunk bound is one shared constant on both sides: `STORAGE_CHUNK_SLOTS` (now `pub` in `tn-storage`) is enforced by `append_account`'s internal chunker, by the new streaming walk, and matches what the reader streams per chunk, so writer and reader stay symmetric and the pack bytes are unchanged: same account order, same slot order, same chunk boundaries, same stats tallies.
- The walk verifies (rather than assumes) that the dup-sorted cursor yields strictly ascending slot keys per account, erroring instead of silently re-sorting, because chunk boundaries (and therefore pack bytes) depend on that order.
- `export_state_pack` now carries a `# Memory profile` doc block mirroring `import_state`'s: at most one account header, one chunk buffer (~4 MiB), and one bytecode resident at a time.

## Changes

- [x] `crates/tn-reth/src/snapshot.rs`: rewrote the export walk as `export_plain_state` / `stream_account_slots` over a small `StatePackSink` trait (production impl: `ExecStatePackWriter`), deleted `code_cache` and the per-account `BTreeMap`, added the `# Memory profile` block, added an out-of-order-slot error path.
- [x] `crates/storage/src/exec_state_pack.rs`: made `STORAGE_CHUNK_SLOTS` and `append_storage_chunk_owned` public (doc'd) so a streaming exporter can flush at exactly the writer's bound with no per-chunk copy; no behavior change.
- [x] Tests: a `RecordingSink` double stamps every chunk with the source-pull count at emission, which is the observable that separates a streaming exporter from a materialize-first one. Four new unit tests cover chunk cadence on a multi-chunk whale, per-account bytecode resolution without retention, rejection of out-of-order and duplicate slot keys, and orphan-row / zero-slot merge-join semantics.
- [x] `MULTI_CHUNK_SLOTS` in the whale-import test now derives from the public constant (same value, `4 * 64 * 1024 + 7`); no existing test assertion was edited.

## Testing

Pinned-toolchain commands run in the worktree:

- `cargo check` over buildplan's 21-crate reverse-dependency closure of the two edited crates: green.
- `cargo nextest run -p tn-storage exec_state_pack` and `cargo nextest run -p tn-reth snapshot`: 17/17 snapshot tests pass, including `import_streams_whale_storage_in_chunks` (262,151-slot whale, assertions untouched) and `export_roundtrips_accounts_storage_and_code`, which pin the byte-identical output claim end to end.
- Confirmed by mutation: reverting `stream_account_slots` to collect-then-emit makes `export_streams_whale_storage_without_materializing` fail on its first `pulls_at_emit` stamp (first chunk only after all 131,079 pulls instead of at 65,536), then the restore goes green again.
- `cargo +nightly fmt` clean; nightly `clippy --workspace --all-features` run before push.

Rebased onto `origin/main` `19290b11`. #1103 rewrote a large part of `snapshot.rs` (+461/-279), but entirely outside the export path this PR touches: every conflict was an import block (the top-level `use std` / `use tn_types` pair, and the test module's), each resolved as the union of both sides with the entries no longer referenced pruned by a usage scan. Re-deriving this PR's diff against the new main yields the same 676-line patch as before, differing only in those import hunks and in line offsets, and `git diff origin/main` still lists only the two files above.

### DIFF
--- a/crates/storage/src/exec_state_pack.rs
+++ b/crates/storage/src/exec_state_pack.rs
@@ -59,10 +59,14 @@ pub const EXEC_STATE_PACK_VERSION: u16 = 1;
 /// Name of the data file inside the pack directory.
 const DATA_NAME: &str = "state_data";
 
-/// Maximum storage slots per [`ExecStateRecord::Storage`] chunk. Each slot is 64 bytes, so 64k
+/// Maximum storage slots per `Storage` chunk record. Each slot is 64 bytes, so 64k
 /// slots is ~4 MiB decompressed — comfortably under the container's per-record limit
-/// (`MAX_RECORD_SIZE` = 16 MiB, [`crate::archive::pack_iter`]), leaving margin for BCS overhead.
-const STORAGE_CHUNK_SLOTS: usize = 64 * 1024;
+/// (`MAX_RECORD_SIZE` = 16 MiB, `crate::archive::pack_iter`), leaving margin for BCS overhead.
+///
+/// Public so streaming exporters can flush at exactly this bound — reproducing the chunk layout
+/// [`ExecStatePackWriter::append_account`] emits, byte for byte — and so the per-account
+/// working-set ceiling such an exporter must hold in memory is stated in one place.
+pub const STORAGE_CHUNK_SLOTS: usize = 64 * 1024;
 
 /// Upper bound on the header count a pack may declare, enforced before any allocation sized by the
 /// untrusted `header_count`. A genuine pack carries the snapshot header plus at most
@@ -331,8 +335,11 @@ impl ExecStatePackWriter {
     }
 
     /// Append one owned chunk of storage slots for the current account, moving it into its record
-    /// (no extra copy) and bumping the tally. Empty chunks are ignored.
-    fn append_storage_chunk_owned(
+    /// (no extra copy) and bumping the tally. Empty chunks are ignored. A chunk should hold at
+    /// most [`STORAGE_CHUNK_SLOTS`] slots to stay under the container's record limit; streaming
+    /// callers that flush at exactly that bound reproduce
+    /// [`append_account`](Self::append_account)'s chunk layout byte for byte.
+    pub fn append_storage_chunk_owned(
         &mut self,
         slots: Vec<(B256, B256)>,
     ) -> Result<(), ExecStatePackError> {
--- a/crates/tn-reth/src/snapshot.rs
+++ b/crates/tn-reth/src/snapshot.rs
@@ -22,7 +22,9 @@
 //! # Storage format is the exec state pack (not JSONL)
 //!
 //! [`PinnedStateView::export_state_pack`] writes an [`ExecStatePackWriter`]: the caller-supplied
-//! state root and block header(s), then one [`ExecStateAccount`] per plain-state account. The pack
+//! state root and block header(s), then one
+//! [`ExecStateAccount`](tn_storage::exec_state_pack::ExecStateAccount) per plain-state account. The
+//! pack
 //! is a compact, CRC32- and zstd-framed binary artifact (see `tn_storage::exec_state_pack`) — it
 //! does not go through reth's private JSONL `init_from_state_dump` path at all. The restore side
 //! rebuilds state from the pack directly using reth's *public* state-insertion and trie building
@@ -87,19 +89,15 @@ use reth_revm::{
 use reth_stages_types::{StageCheckpoint, StageId};
 use reth_trie::{IntermediateStateRootState, StateRoot, StateRootProgress};
 use reth_trie_db::DatabaseStateRoot;
-use std::{
-    cmp::Ordering,
-    collections::{BTreeMap, HashMap},
-    path::Path,
-};
+use std::{iter::Peekable, path::Path};
 use tn_storage::exec_state_pack::{
-    ExecStateAccount, ExecStateAccountMeta, ExecStatePackReader, ExecStatePackWriter,
-    ExecStateStats, StateEntry,
+    ExecStateAccountMeta, ExecStatePackReader, ExecStatePackWriter, ExecStateStats, StateEntry,
+    STORAGE_CHUNK_SLOTS,
 };
 use tn_types::{
     gas_accumulator::{entry_fee_for_worker, GasAccumulator},
-    Address, BlockBody, BlockNumHash, Bytes, ExecHeader, GenesisAccount, SealedBlock, SealedHeader,
-    TaskManager, WorkerId, B256, U256,
+    Address, BlockBody, BlockNumHash, Bytes, ExecHeader, SealedBlock, SealedHeader, TaskManager,
+    WorkerId, B256, U256,
 };
 use tracing::{debug, info};
 
@@ -155,10 +153,20 @@ impl PinnedStateView {
     /// one account each, emitted in ascending address order (the `PlainAccountState` cursor order).
     /// Accounts and their storage are produced by a merge join of the `PlainAccountState` and
     /// dup-sorted `PlainStorageState` cursors, walked in lockstep so each table is scanned once;
-    /// bytecode is resolved through a lookup memo keyed by code hash. Zero-valued storage slots are
-    /// omitted (an absent slot is already zero in the trie, so keeping a zeroed row would only
-    /// bloat the pack). Only plain state is written — the restore side rebuilds the hashed,
-    /// history, and trie tables itself.
+    /// bytecode is resolved with one `Bytecodes` point read per contract account against the
+    /// pinned transaction. Zero-valued storage slots are omitted (an absent slot is already zero
+    /// in the trie, so keeping a zeroed row would only bloat the pack). Only plain state is
+    /// written — the restore side rebuilds the hashed, history, and trie tables itself.
+    ///
+    /// # Memory profile
+    ///
+    /// Mirrors the bound [`SnapshotRestorer::import_state`] documents for the opposite direction:
+    /// at most one account header, one [`STORAGE_CHUNK_SLOTS`]-bounded storage chunk buffer
+    /// (~4 MiB), and one account's bytecode are resident at a time. Storage streams from the
+    /// cursor into per-chunk buffers flushed as they fill, so a whale account's storage is never
+    /// fully materialized, and bytecode is re-read per contract account instead of memoized, so
+    /// nothing this function allocates scales with a single account's slot count or with the
+    /// number of distinct bytecodes in state.
     ///
     /// Returns the [`ExecStateStats`] the writer accumulated.
     pub fn export_state_pack(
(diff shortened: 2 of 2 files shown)


grantkee (approved):
LGTM!

PR #1114, opened 2026-08-07
fix(tn-reth): refuse non-public dial targets in observer transaction forwarding

### LINKED ISSUE #1092: `tn-reth`: `WorkerRpcForwarder` forwards to any committee-advertised URL with no host or IP allowlist (blind SSRF from observer nodes)
`tn-reth`: `WorkerRpcForwarder` forwards to any committee-advertised URL with no host or IP allowlist (blind SSRF from observer nodes)

**Severity:** Low.  An observer node auto-dials a URL chosen by a committee member, so a committee member can aim an observer's outbound HTTP at hosts inside that observer's own perimeter.  No consensus impact: no halt, no fork, no state divergence, and no effect on batch or certificate production.  The response body is never returned to the advertiser (`crates/tn-reth/src/forward.rs:210-221` only logs an outcome), so this is blind SSRF, a probe and side-effect trigger rather than direct exfiltration.

**Attacker model:** any current-committee validator, or anyone holding a current-committee validator's BLS key.  Advertised RPC info travels inside a BLS-signed `NodeRecord` (`crates/network-libp2p/src/types.rs:843-855`, `:985`) whose signature is verified at record ingest (`crates/network-libp2p/src/consensus.rs:605-609`), so an arbitrary network peer cannot inject one; kademlia record poisoning is already closed off by that signature.  The victim is a worker node that is not an active committee validator ("observer"), the only role that reaches the forwarder.  There are two entry points, both gated the same way: `WorkerNode::seal` routes to `disburse_txns` when the node has no `quorum_waiter` (`crates/consensus/worker/src/worker.rs:281-285`), and epoch close re-disburses orphaned batches when `is_active_cvv()` is false (`crates/node/src/manager/node/close_epoch.rs:128-145`).

## Summary

`cached_providers` (`crates/tn-reth/src/forward.rs:90-124`) turns every advertised `RpcInfo.http` into a live `RootProvider` with a bare `url.parse()` at `:105` and `RootProvider::new_http` at `:107`, with no check on the host.  `forward_txns` then POSTs `eth_sendRawTransaction` to it (`forward.rs:175`) for every transaction in the sealed batch whose sender maps to that validator's committee slot, and for any other transaction whose earlier-tried endpoints all failed endpoint-locally.  The only gate an advertised endpoint passes before reaching this point is `RpcInfo::validate` (`crates/types/src/committee.rs:46-68`), applied at record-ingest in the peer manager (`crates/network-libp2p/src/peers/manager.rs:962-976`); it checks URL byte length and the `http`/`https` scheme and never inspects the host.  A committee member that advertises `http://169.254.169.254/latest/meta-data/`, or any host reachable only from inside an observer operator's network, gets every observer to issue authenticated-by-nobody POSTs at that target on the observer's own behalf.

The module already reasons explicitly about one security property of these attacker-chosen URLs, namely that a plain-`http://` advertisement carries the raw transaction in cleartext (`forward.rs:12-18`).  The dial target itself is not reasoned about anywhere.

## Problem

The end-to-end path, verified in the tree at commit `48e00b4b`:

1. An observer accepts transactions over its public RPC and seals a batch.  `WorkerNode::seal` has no `quorum_waiter` on a non-validator, so it routes to `disburse_txns` (`crates/consensus/worker/src/worker.rs:281-285`).
2. `disburse_txns` calls `network_handle.get_all_validator_rpcs()` and hands the result straight to `forward_txns` (`worker.rs:253-257`).
3. `get_all_validator_rpcs` is documented as a "Snapshot of the current committee's advertised RPCs", populated by kademlia discovery of peers' node records (`crates/network-libp2p/src/types.rs:824-840`).  The values are authored by other committee members, not by the local node.
4. Before a record is cached, `cache_known_peer` calls `rpc.validate()` and nulls the field on error (`crates/network-libp2p/src/peers/manager.rs:962-976`).  Its own comment states the limit of that gate: "signature verification proves authenticity but not scheme correctness".
5. `RpcInfo::validate` checks length and scheme only:

   ```rust
       pub fn validate(&self) -> Result<(), RpcInfoError> {
           if self.http.as_str().len() > MAX_RPC_URL_LEN {
               return Err(RpcInfoError::UrlTooLong(self.http.as_str().len()));
           }
           match self.http.scheme() {
               "http" | "https" => {}
               scheme => {
                   return Err(RpcInfoError::InvalidHttpScheme(scheme.to_string()));
               }
           }
           // ... elided: the `ws` arm applies the same two checks, then `Ok(())`
   ```

   (`crates/types/src/committee.rs:46-55`; the `ws` arm at `:56-66` is the same two checks.)  There is no host inspection anywhere in the function.

6. `cached_providers` reparses that URL and constructs a live provider from it, with the only rejection path being an unparseable URL:

   ```rust
           validator_rpcs
               .iter()
               .filter_map(|(key, rpc)| {
                   let url = rpc.http.to_string();
                   cache
                       .get(&url)
                       .cloned()
                       .or_else(|| match url.parse() {
                           Ok(parsed) => {
                               let provider = RootProvider::new_http(parsed);
                               cache.insert(url, provider.clone());
                               Some(provider)
                           }
                           // ... elided: the `Err` arm logs and yields `None`
   ```

   (`crates/tn-reth/src/forward.rs:98-110`.)

7. `forward_txns` sends to it: `provider.send_raw_transaction(tx.as_slice())` under a 5s per-endpoint timeout and a 15s per-transaction budget (`forward.rs:173-181`, constants at `:43` and `:49`).  The order it walks is the owning validator first, then every advertised endpoint as fallback (`:149`, `:155-156`), and the walk stops at the first delivery or considered rejection (`:186-193`), so an endpoint late in that order is reached only when the ones before it fail endpoint-locally.

A search across `crates/tn-reth`, `crates/types`, `crates/network-libp2p` and `crates/consensus/worker` for a loopback, RFC1918, link-local or `169.254` host check finds none in this path.  The two `allowlist` concepts that do exist are unrelated: the RPC namespace allowlist (`crates/tn-reth/src/cli.rs:10-12`) and the peer operator allowlist (`crates/network-libp2p/src/peers/peer.rs:61-66`), neither of which constrains a forwarded dial target.

**On the `RpcInfo` doc comment.**  The type's own documentation says these are "application-layer URLs consumed by external clients (wallets/dapps); they are never dialed by the libp2p swarm" (`crates/types/src/committee.rs:32-34`).  That sentence stays literally true, since the swarm does not dial them and `forward.rs` is outside the swarm.  It is worth quoting anyway because it is the only place in the tree that states a dialing policy for this field, and it reads as a reason to treat an advertised URL as client-facing metadata rather than as an attacker-chosen dial target for an unattended machine process.  The forwarder makes it the latter.

**Failure scenario.**  A malicious or compromised committee member advertises an internal-only URL: cloud instance metadata at `169.254.169.254`, a Kubernetes API server, or an internal admin endpoint.  An observer that seals a batch dials it and POSTs a JSON-RPC `eth_sendRawTransaction` body into its own perimeter, from a source address inside that perimeter.  The attacker chooses scheme (`http` or `https`), host, port and path, and can rotate the target across successive node-record republishes.  Reaching the target on the first try is under the attacker's control rather than incidental: routing is by sender slot (`forward.rs:155`), so submitting a transaction whose sender maps to the attacker's own committee slot puts the attacker's advertised URL first in the order.  Absent that, the URL is reached only as a fallback, after the endpoints ahead of it fail endpoint-locally.

Two limits on what this buys the attacker, stated here rather than left for a reviewer to find.  First, the response body never leaves the observer (`:210-221`), so there is no exfiltration channel and no direct read oracle.  An indirect oracle would require the attacker to control a second advertised endpoint later in the fallback order and watch whether the request arrives there, which is not demonstrated in this issue, so any port-scan or reachability-mapping use of this primitive should be read as unproven.  Second, the request is a POST carrying a JSON-RPC body, which makes the cloud-metadata case a reachability example rather than credential theft: IMDS credential reads are GETs, IMDSv2 additionally requires a token obtained by PUT, and the response would be discarded either way.  What the mechanism does earn is unauthenticated, attacker-directed POSTs originating inside the observer's network, against a host, port and path the attacker picks, with a body the attacker can influence by submitting their own transaction to the observer first.

## Why the blast radius is bounded today

1. The advertisement is BLS-signed inside a `NodeRecord` (`crates/network-libp2p/src/types.rs:843-855`) and `get_all_validator_rpcs` returns current-committee members only, explicitly excluding pinned operator peers and previous/next committee members (`types.rs:824-830`).  An unprivileged network peer cannot reach this path at all.
2. Only non-validator workers forward.  A committee validator takes the `quorum_waiter` branch and never calls `disburse_txns` (`crates/consensus/worker/src/worker.rs:281-285`).
3. The outcome is never reflected.  `ForwardOutcome` is matched into log lines and discarded (`crates/tn-reth/src/forward.rs:210-221`), and a delivered forward produces no output at all.
4. Amplification is bounded per transaction by `FORWARD_SEND_TIMEOUT` (5s, `forward.rs:43`) and `FORWARD_TX_BUDGET` (15s, `forward.rs:49`), so a single hostile endpoint cannot hold a forward task open indefinitely.

None of these constrain the destination, which is the gap.

## Suggested fix

Add the host check at dial time, in `cached_providers`, immediately before `RootProvider::new_http`.  It has to live here rather than in `RpcInfo::validate`: `validate` is a type-level well-formedness check in `tn-types` with no operator configuration in scope, and it runs at record-ingest in the peer manager, where a local dev deployment legitimately advertises `127.0.0.1`.  Putting the policy at the dial site keeps `validate` a pure syntactic check and puts the decision where the node is about to act on it.

The minimum version is a literal-IP check on the parsed URL host: when the host is an IP literal, reject it if it is loopback, private (RFC1918), link-local (including `169.254.0.0/16` and IPv6 `fe80::/10`), unique-local (`fc00::/7`), or unspecified.  This is DNS-free, costs nothing on the cache-miss path, and catches the metadata-endpoint and internal-IP cases directly.

The complete version additionally resolves DNS hostnames and applies the same predicate to every resolved address.  That closes the case where a committee member advertises `http://internal.attacker.example/` resolving to `10.0.0.5`.  It costs a resolution on each cache miss and, on its own, is still raceable by a rebinding resolver unless the resolved address is pinned into the HTTP connector rather than re-resolved at connect time.  Recommend shipping the literal-IP check first and tracking the resolving version separately, since the literal check is spec-independent and self-contained while the resolving one changes the connector.

Both versions need an operator opt-in, defaulting to reject, that local and dev network configurations enable, so single-host and docker-compose deployments that legitimately advertise `127.0.0.1` keep working.  Log each suppressed endpoint once at `warn` with the BLS key and URL, mirroring the existing unparseable-URL `debug!` at `forward.rs:111-119`, so an operator can distinguish a filtered endpoint from an unreachable one.

## Test gap

The nearest existing coverage is `cached_providers_resolves_every_advertised_endpoint` (`crates/tn-reth/src/forward.rs:313-326`).  It builds two `RpcInfo` entries pointing at `http://127.0.0.1:8545` and `http://127.0.0.1:8546` and asserts both resolve:

```rust
        let resolved = forwarder.cached_providers(&rpcs);

        assert_eq!(resolved.len(), 2);
        assert_eq!(cached_urls(&forwarder).len(), 2);
```

It cannot catch this, because it pins the current behavior: it asserts that loopback endpoints resolve successfully into live providers with zero host filtering.  It therefore has to be updated alongside any allowlist, either by moving its fixtures to non-loopback hosts or by constructing the forwarder with the local opt-in enabled.  The sibling test `cached_providers_reuses_cache_and_evicts_stale_endpoints` (`forward.rs:328-343`) uses `127.0.0.1` fixtures for the same reason and needs the same treatment.

Upstream, the only tests on the advertised-endpoint gate are `rpc_info_validate_rejects_wrong_scheme` (`crates/types/src/committee.rs:996-1020`) and `rpc_info_validate_rejects_oversized_url` (`committee.rs:1023-1047`).  Both assert scheme and length only.  No test anywhere in the tree asserts anything about the host of an advertised endpoint.

A covering test should assert, in one `cached_providers` call under the default (non-local) configuration:

- a slice containing `http://169.254.169.254/latest/meta-data/`, `http://127.0.0.1:8545`, `http://10.0.0.1:8545` and `http://192.168.1.1:8545` yields no provider for any of them, and leaves the provider cache without those URLs;
- a public host in that same slice still resolves and is still cached, so the assertion cannot pass vacuously by rejecting everything (a reject-all regression would otherwise look identical to a correct filter);
- with the local opt-in enabled, the loopback entries resolve again, so dev deployments are covered explicitly rather than by the absence of a test.

Confirm-by-mutation: remove the host check, run the new test, watch it fail, then restore.  A test never seen to fail is not evidence the check is wired in.

## Acceptance criteria

- An advertised `RpcInfo.http` whose host is a loopback, private, link-local, unique-local or unspecified IP literal never becomes a `RootProvider` and never receives a forwarded transaction, under the default network configuration.
- A public advertised host appearing in the same `validator_rpcs` slice as a rejected one still resolves and still receives forwarded transactions, so the filter is selective rather than blanket.
- Operators of local and dev deployments can opt in to loopback and private endpoints explicitly, and that opt-in is off by default.
- Every suppressed endpoint is logged once with its BLS key and URL, so an operator can tell a filtered endpoint from an unreachable one.
- Removing the host check causes at least one test in the suite to fail.

## Related, not duplicate

- #804 (closed) is the change that introduced observer transaction forwarding; it is cited from the code at `crates/consensus/worker/src/worker.rs:240-243`, `crates/node/src/manager/node/start_epoch.rs:462-464` and `crates/tn-reth/src/forward.rs:5`.  This issue is not about that design, only about the missing constraint on the endpoints it made observers dial.
- #901 (closed) covers the operator-side `keytool` path for setting a node's advertised JSON-RPC endpoint.  Same field, publisher side.  Its diff was not audited for this issue; the consumer-side gate described above is `RpcInfo::validate` as applied in `crates/network-libp2p/src/peers/manager.rs:962-976`.
- No open PR touches `crates/tn-reth/src/forward.rs`.  The open Multi-Worker Support stack (#563 through #568) does touch `crates/types/src/committee.rs`, but the fix proposed here does not require changing that file.

## Provenance

Surfaced by a static hardening sweep over `crates/engine` and `crates/tn-reth` at commit `48e00b4b` (2026-08-04): 16 finder lenses, each with an exclusion list built from the live backlog, and every candidate re-checked by an independent adversarial refuter plus a prior-art auditor.  Two caveats from that sweep apply here and are stated rather than hidden.  First, no code in the sweep was built or executed, so every claim in this issue is static; the suggested fix has not been compiled.  Second, the sweep dropped 6 lower-ranked candidates un-verified at a dedup cap and had 6 agents return empty results, so the absence of a neighbouring finding is not evidence of absence.  This particular item did not lose a verifier: every line cited above was read directly in the tree rather than taken on an agent's word, and each citation was re-confirmed against `48e00b4b`.

## References

- `crates/tn-reth/src/forward.rs:1-25` (module doc; the acknowledged plain-`http://` cleartext property at `:12-18`), `:43` (`FORWARD_SEND_TIMEOUT`), `:49` (`FORWARD_TX_BUDGET`), `:90-124` (`cached_providers`; `url.parse()` at `:105`, `RootProvider::new_http` at `:107`, unparseable-URL skip at `:111-119`), `:127-225` (`forward_txns`; fallback set at `:149`, `:156`, `send_raw_transaction` at `:175`, outcome logging at `:210-221`), `:313-326` (`cached_providers_resolves_every_advertised_endpoint`), `:328-343` (`cached_providers_reuses_cache_and_evicts_stale_endpoints`)
- `crates/types/src/committee.rs:27-28` (`MAX_RPC_URL_LEN`), `:30-41` (`RpcInfo` and its "never dialed by the libp2p swarm" doc at `:32-34`), `:46-68` (`validate`), `:996-1020` and `:1023-1047` (the only two `validate` tests)
- `crates/network-libp2p/src/peers/manager.rs:957-986` (`cache_known_peer`; the only `rpc.validate()` call on the remote-record ingest path, at `:966-976`, and its "authenticity but not scheme correctness" comment at `:963-965`).  The other two `rpc.validate()` call sites in the tree validate the node's OWN configured endpoint, not a peer's, and so do not gate this path: `crates/node/src/manager/node.rs:1160-1164` at startup and `crates/telcoin-network-cli/src/keytool/set_rpc.rs:39`
- `crates/network-libp2p/src/types.rs:824-840` (`get_all_validator_rpcs`, current-committee scope and kad sourcing), `:843-855` (`NodeRecord` and its BLS signature), `:985` (`NetworkInfo.rpc`)
- `crates/consensus/worker/src/worker.rs:240-257` (`disburse_txns`), `:281-285` (the non-validator branch in `seal`).  `disburse_txns` has one other caller, `crates/node/src/manager/node/close_epoch.rs:144`, which re-disburses orphaned batches on epoch close and is gated on the same non-CVV condition (`close_epoch.rs:128-145`)
- `crates/node/src/manager/node/start_epoch.rs:462-466` (forwarder construction; note that the construction itself is unconditional, and it is the two call sites above that restrict forwarding to non-CVV nodes)
- `crates/tn-reth/src/cli.rs:10-12` and `crates/network-libp2p/src/peers/peer.rs:61-66` (the two unrelated existing "allowlist" concepts, neither of which constrains a dial target)

All line numbers are at commit `48e00b4b`.  `main` has advanced two commits past it since the sweep.  Of the files cited here, only `crates/node/src/manager/node/start_epoch.rs` and `crates/node/src/manager/node/close_epoch.rs` changed in those commits: on current `main` the forwarder construction is at `start_epoch.rs:493-497` and the orphan-batch `disburse_txns` call is at `close_epoch.rs:145`.  Every other file cited above is byte-identical at both commits, so the rest of the anchors hold on `main` as written.

MavenRain:
Closes #1092.  (Cantina #11)

## Problem

An observer (a worker that is not an active committee validator) forwards every transaction it seals to the JSON-RPC endpoint the owning validator advertised on its worker record. The dial target is therefore chosen by a committee member, not by the observer, and `cached_providers` turned any advertised URL into a live `RootProvider` with no check on the host at all.

A committee member that advertises `http://169.254.169.254/latest/meta-data/`, a Kubernetes API server, or any host reachable only from inside an observer operator's network gets every observer to POST a JSON-RPC body at that target from a source address inside that perimeter. The attacker picks scheme, host, port and path, and can rotate the target on each node-record republish. Because routing is by sender slot, submitting a transaction whose sender maps to the attacker's own committee slot puts the attacker's URL first in the order rather than leaving it as a fallback.

Threat model. The advertisement rides inside a BLS-signed `NodeRecord` whose signature is verified at record ingest, so an arbitrary network peer cannot inject one and kademlia record poisoning is already closed off. The attacker is a current-committee validator, or anyone holding a current-committee validator's BLS key. There is no consensus impact: no halt, no fork, no state divergence, and the response body is never returned to the advertiser, so this is blind SSRF (a probe and side-effect trigger, not an exfiltration channel). Both entry points are gated the same way: `WorkerNode::seal` routes to `disburse_txns` when the node has no `quorum_waiter`, and epoch close re-disburses orphaned batches when `is_active_cvv()` is false.

## Root cause

The only gate an advertised endpoint passed before becoming a dial target was `RpcInfo::validate`, applied at record ingest in the peer manager. It checks URL byte length and the `http` / `https` scheme and never inspects the host. Its own call site says as much: "signature verification proves authenticity but not scheme correctness".

`crates/tn-reth/src/forward.rs` already reasoned explicitly about one security property of these attacker-chosen URLs (that a plain-`http://` advertisement carries the raw transaction in cleartext). The dial target itself was not reasoned about anywhere.

## Fix

The check lives at the dial site rather than in `RpcInfo::validate`. `validate` is a type-level well-formedness check in `tn-types` with no operator configuration in scope, and it runs at record ingest, where a local dev deployment legitimately advertises `127.0.0.1`. Putting the policy where the node is about to act on the URL keeps `validate` a pure syntactic check.

- **`crates/tn-reth/src/forward.rs`**: new `ForwardTargetPolicy` (`PublicOnly` default, `AllowPrivate` opt-in), applied in `cached_providers` immediately before `RootProvider::new_http`. `PublicOnly` refuses loopback, RFC 1918 private-use, link-local (RFC 3927 `169.254.0.0/16` and RFC 4291 `fe80::/10`), RFC 4193 unique-local `fc00::/7`, RFC 6598 shared address space `100.64.0.0/10`, `0.0.0.0/8`, and multicast / broadcast / reserved. The refusal reason is a closed `RefusedTarget` sum type rather than a message string, so the log line and the tests agree on it.
- **Bypass closure.** Four spellings of an internal address that an IP-literal check can miss are unmapped and re-checked against the IPv4 predicate: IPv4-compatible `::a.b.c.d`, IPv4-mapped `::ffff:a.b.c.d`, the RFC 6052 NAT64 well-known prefix `64:ff9b::/96`, and the RFC 3056 6to4 prefix `2002::/16`. The first two are what `Ipv6Addr::to_ipv4` recognizes; the last two it does not, yet a NAT64 gateway (the default egress path for IPv6-only cloud and container subnets) translates `http://[64:ff9b::a9fe:a9fe]/` straight back to `169.254.169.254`, so without this the literal `http://169.254.169.254/` is refused while its NAT64 spelling is dialed. RFC 6052 network-specific prefixes use an operator's own prefix and cannot be enumerated statically, which is stated at the function. The IPv6-specific blocks are tested before the fallback so `::1` reports as loopback rather than as the `0.0.0.1` it maps to. The alternative IPv4 spellings (`http://2130706433/`, `http://0177.0.0.1/`) are normalized to `Host::Ipv4` by URL host parsing before the predicate sees them, and are covered by test.
- **Reserved names.** `localhost`, `*.localhost` (RFC 6761 section 6.3) and `*.local` (RFC 6762) are refused by name, in relative, absolute and subdomain form. Without this, `http://localhost:8545/` walks straight past an IP-literal-only filter. The check is deliberately DNS-free: it costs nothing on the cache-miss path and cannot be steered by a resolver the advertiser controls. Refusing an arbitrary hostname that merely resolves inside the perimeter needs resolution plus address pinning in the connector (a re-resolving check is raceable by a rebinding resolver), which changes the connector and is left to a follow-up, as the issue recommends.
- **`crates/tn-reth/src/forward.rs`**: `EndpointCache` now holds the provider map plus a `refused` set. Both are evicted against the current advertisement set on every forward, so the refusal warning fires once per advertisement instead of once per batch seal, and a re-advertised endpoint warns again. A refused endpoint is never inserted into the provider cache, so it can never be served from cache on a later seal.
- **`crates/config/src/node.rs`**: new `Parameters::allow_private_forward_targets`, `#[serde(default)]` to `false`. Absent from a `parameters.yaml` it deserializes to the refusing policy, so the secure value is the one an un-migrated operator gets.
- **`crates/node/src/manager/node/start_epoch.rs`**: reads the parameter and passes the policy into `WorkerRpcForwarder::new`.
- **`crates/e2e-tests/src/lib.rs`**: the local test network enables the opt-in on the shared parameters before they are distributed. Every node in that harness advertises `http://127.0.0.1:<port>`, which is exactly the single-host case the opt-in exists for. Without this the observer forwarding tests would go silently dead rather than fail loudly.
- **`crates/telcoin-network-cli/README.md`**: parameter table row plus a section stating the default, what it refuses, and the narrow condition under which an operator should turn it on.
- **`crates/tn-reth/README.md`**: the transaction-forwarding section already reasoned about one security property of these advertised URLs (that a plain-`http://` advertisement carries the raw transaction in cleartext) and said nothing about the dial target. It now states the policy, the default, the opt-in, and the DNS limitation alongside it.

Each suppressed endpoint is logged once at `warn` with the advertising validator's BLS key, the URL, and the refusal reason, mirroring the existing unparseable-URL `debug!`, so an operator can tell a filtered endpoint from an unreachable one.

## Testing

Pinned toolchain throughout (`rust-toolchain.toml` = 1.94; rustfmt on the pinned nightly, since `rustfmt.toml` sets `wrap_comments` which stable `cargo fmt` skips).

- `cargo +nightly-2026-03-20 fmt --all -- --check`: clean.
- `cargo +1.94 clippy --workspace --all-targets`: exit 0, zero warnings attributed to any file in this diff.
- `cargo +1.94 test --no-run --workspace`: exit 0.
- `cargo +1.94 test -p tn-reth --lib forward::`: 9 passed, 0 failed.

New coverage in `crates/tn-reth/src/forward.rs`:

- `cached_providers_refuses_non_public_endpoints_but_keeps_public_ones` puts `169.254.169.254`, `127.0.0.1`, `10.0.0.1` and `192.168.1.1` in the same slice as a public host and asserts only the public one becomes a provider and only its URL is cached. The public entry is the guard against a vacuous pass: a reject-everything regression would otherwise look identical to a correct filter.
- `public_only_refuses_every_non_global_host_form` covers each refused block with its expected `RefusedTarget`, including the IPv4-mapped, NAT64 and 6to4 IPv6 spellings and the reserved names.
- `public_only_admits_public_hosts` walks the address just outside each refused block (`172.32.0.1`, `100.128.0.1`, `169.253.0.1`, `126.255.255.255`, `223.255.255.255`, `[fbff::1]`, `[fe7f::1]`), so an off-by-one in a mask fails here rather than silently over-rejecting operators. It also carries a NAT64 and a 6to4 address whose *embedded* address is public, plus `[64:ff9c::a9fe:a9fe]` (one bit off the NAT64 well-known prefix), so the unmapping is shown to be an exact prefix match rather than a wildcard that swallows public IPv6.
- `cached_providers_never_caches_a_refused_endpoint` asserts a refused endpoint stays out of the provider cache across seals, is memoized once for the warning, and is forgotten when it stops being advertised.
- `cached_providers_dials_private_endpoints_under_the_local_opt_in` covers the dev deployment explicitly rather than by the absence of a test.
- The two pre-existing `cached_providers` tests moved off `127.0.0.1` fixtures onto public hosts. They pinned the old behavior (that loopback endpoints resolve with zero host filtering), so they had to be updated alongside the filter; they still test cache reuse and stale-endpoint eviction.

Confirmed by mutation. Each mutant was applied to a byte-exact copy, the suite was watched going red, and the file was restored and re-verified by SHA-256:

| Mutant | Result |
| --- | --- |
| `PublicOnly` arm returns `Ok(())` (host check removed) | killed, 3 failed |
| IPv4-mapped IPv6 unmapping removed | killed, 1 failed |
| `localhost` / `*.localhost` arm disabled | killed, 1 failed |
| refusal memo always re-warns | killed, 2 failed |
| `refused` set not evicted with the advertisement | killed, 1 failed |
| NAT64 well-known prefix not unmapped | killed, 1 failed |
| 6to4 prefix not unmapped | killed, 1 failed |
| `embedded_ipv4` result discarded entirely | killed, 1 failed |

Eight mutants, eight kills, no survivors.

## Scope notes

- The resolving version of the check (resolve a hostname and apply the same predicate to every resolved address, with the address pinned into the connector so a rebinding resolver cannot race it) is deliberately not in this PR. It changes the HTTP connector, whereas the literal check is self-contained. The issue recommends the same split. The practical consequence is worth stating plainly rather than leaving for a reviewer to find: a committee member that advertises `http://evil.example.com/` with an A record pointing at `169.254.169.254` still gets dialed. This change raises the cost of the attack from "advertise a literal" to "control a domain", and closes the literal-address case completely; it is not complete egress filtering, and the README now says so.
- RFC 6052 permits NAT64 network-specific prefixes at `/32`, `/40`, `/48`, `/56`, `/64` and `/96`. Those use an operator's own prefix, which a static check cannot enumerate, so only the well-known `64:ff9b::/96` is matched.
- `RpcInfo::validate` is unchanged, so record ingest still accepts a loopback advertisement. That is intentional: a validator on a local network legitimately advertises one, and the consumer decides whether to dial it.
- The `ws` field of `RpcInfo` is not covered, because the forwarder never dials it. If a future consumer dials `RpcInfo::ws`, it needs the same gate.
- `chain-configs/mainnet/parameters.yaml` and `chain-configs/testnet/parameters.yaml` are unchanged. Both omit the key and therefore get the refusing default, matching how `parallel_fetch_request_delay_interval` is already handled in the mainnet file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11939,6 +11939,7 @@ dependencies = [
  "tn-worker",
  "tokio",
  "tracing",
+ "url",
 ]
 
 [[package]]
--- a/crates/config/src/node.rs
+++ b/crates/config/src/node.rs
@@ -280,6 +280,21 @@ pub struct Parameters {
     /// certificates.
     #[serde(default = "Parameters::default_parallel_fetch_request_delay_interval")]
     pub parallel_fetch_request_delay_interval: Duration,
+
+    /// Allow observer transaction forwarding to dial advertised JSON-RPC endpoints whose host is
+    /// not a public internet address.
+    ///
+    /// An observer forwards each transaction it seals to the endpoint the owning validator
+    /// advertised on its node record, so the dial target is chosen by a committee member rather
+    /// than by this node. Default `false`: an endpoint on a loopback, private, link-local,
+    /// unique-local, shared-address-space or unspecified address is refused, so a committee member
+    /// cannot aim this node's outbound HTTP at hosts inside its own perimeter (issue #1092).
+    ///
+    /// Set `true` only where every committee member is under the same operator as this node --
+    /// single-host and docker-compose deployments, where validators legitimately advertise
+    /// `127.0.0.1` -- since it restores dialing of arbitrary internal addresses.
+    #[serde(default = "Parameters::default_allow_private_forward_targets")]
+    pub allow_private_forward_targets: bool,
 }
 
 impl Parameters {
@@ -327,6 +342,11 @@ impl Parameters {
     fn default_parallel_fetch_request_delay_interval() -> Duration {
         Duration::from_secs(5)
     }
+
+    /// Refuse non-public forwarding targets unless the operator opts in.
+    fn default_allow_private_forward_targets() -> bool {
+        false
+    }
 }
 
 /// Admin server settings.
@@ -366,6 +386,7 @@ impl Default for Parameters {
             basefee_address: None,
             parallel_fetch_request_delay_interval:
                 Parameters::default_parallel_fetch_request_delay_interval(),
+            allow_private_forward_targets: Parameters::default_allow_private_forward_targets(),
         }
     }
 }
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -12,7 +12,7 @@ use std::{
     sync::OnceLock,
 };
 use telcoin_network_cli::{genesis::GenesisArgs, keytool::KeyArgs, node::NodeCommand, NoArgs};
-use tn_config::{Config, ConfigFmt, ConfigTrait, KeyConfig};
+use tn_config::{Config, ConfigFmt, ConfigTrait, KeyConfig, Parameters};
 use tn_node::launch_node;
 use tn_reth::init_txpool_defaults;
 use tn_types::{test_utils::CommandParser, Address, Genesis, GenesisAccount};
@@ -242,6 +242,16 @@ fn config_local_testnet_inner(
     }
     let create_committee_command = CommandParser::<GenesisArgs>::parse_from(genesis_args);
     create_committee_command.args.execute(shared_genesis_dir.clone())?;
+
+    // Every node in this harness runs on one host and advertises `http://127.0.0.1:<port>`, which
+    // the observer's transaction forwarder refuses by default (issue #1092). This is the
+    // single-host deployment the opt-in exists for, so enable it on the shared parameters before
+    // they are distributed, rather than leaving observer forwarding silently dead in e2e tests.
+    let parameters_path = shared_genesis_dir.join("parameters.yaml");
+    let mut parameters: Parameters = Config::load_from_path(&parameters_path, ConfigFmt::YAML)?;
+    parameters.allow_private_forward_targets = true;
+    Config::write_to_path(&parameters_path, &parameters, ConfigFmt::YAML)?;
+
     // If provided optional accounts then hack them into genesis now...
     if let Some(accounts) = accounts {
         let data_dir = shared_genesis_dir.join("genesis/genesis.yaml");
--- a/crates/node/src/manager/node/start_epoch.rs
+++ b/crates/node/src/manager/node/start_epoch.rs
@@ -46,7 +46,7 @@ use tn_reth::{
         ConsensusRegistry::{self, EpochInfo},
         EpochState,
     },
-    WorkerRpcForwarder,
+    ForwardTargetPolicy, WorkerRpcForwarder,
 };
 use tn_rpc::RpcNodeInfo;
 use tn_types::{
@@ -503,9 +503,15 @@ where
 
         // Observer transaction forwarding: a non-committee worker forwards each transaction it
         // accepts to the JSON-RPC endpoint of the validator that owns it, discovered over
-        // kademlia (issue #804).
-        let forwarder =
-            Arc::new(WorkerRpcForwarder::new(network_handle.get_task_spawner().clone()));
+        // kademlia (issue #804). The endpoint is chosen by a committee member, so the policy
+        // decides which advertised hosts this node is willing to dial (issue #1092); it refuses
+        // non-public hosts unless the operator opted in for a single-host deployment.
+        let forwarder = Arc::new(WorkerRpcForwarder::new(
+            network_handle.get_task_spawner().clone(),
+            ForwardTargetPolicy::from_allow_private(
+                consensus_config.parameters().allow_private_forward_targets,
+            ),
+        ));
 
         let worker = WorkerNode::new(
             worker_id,
--- a/crates/telcoin-network-cli/README.md
+++ b/crates/telcoin-network-cli/README.md
@@ -479,6 +479,15 @@ The `parameters.yaml` file controls consensus timing and behavior. If the file i
 | `batch_vote_timeout`                    | `10s`    | Timeout for batch voting requests                   |
 | `basefee_address`                       | none     | Address that receives transaction base fees         |
 | `parallel_fetch_request_delay_interval` | `5s`     | Delay between parallel certificate fetch requests   |
+| `allow_private_forward_targets`         | `false`  | Let observer forwarding dial non-public RPC hosts   |
+
+### `allow_private_forward_targets`
+
+An observer forwards each transaction it accepts to the JSON-RPC endpoint the owning validator advertised on its node record, so the dial target is chosen by a committee member rather than by this node. Left at the default `false`, an advertised endpoint on a loopback, private (RFC 1918), link-local, unique-local, shared-address-space or unspecified address is refused and logged once at `warn` with the advertising validator's BLS key, so a committee member cannot direct this node's outbound HTTP at hosts inside its own perimeter.
+
+The check reads the host as written and never resolves DNS. It refuses IP literals in every spelling (dotted-quad, decimal, octal, hex, IPv4-mapped and NAT64/6to4 IPv6) and the names reserved to resolve locally (`localhost`, `*.localhost`, `*.local`), but a hostname that merely resolves to a private address is still dialed. Treat this as a guard against an advertised internal address, not as complete egress filtering. Operators who need the stronger property should restrict outbound traffic from observer nodes at the network layer.
+
+Set it to `true` only when every committee member is under the same operator as this node - single-host and docker-compose deployments, where validators legitimately advertise `127.0.0.1`. On a public network it re-enables dialing arbitrary internal addresses.
 
 Example (testnet configuration):
 
--- a/crates/tn-reth/Cargo.toml
+++ b/crates/tn-reth/Cargo.toml
@@ -66,6 +66,7 @@ alloy = { workspace = true, features = ["genesis"] }
 alloy-evm = { workspace = true }
 tempfile = { workspace = true }
 serde_json = { workspace = true }
+url = { workspace = true }
 tokio = { workspace = true, features = ["macros", "time"] }
 reth-ethereum-primitives = { workspace = true }
 secp256k1 = { workspace = true, optional = true }
--- a/crates/tn-reth/README.md
+++ b/crates/tn-reth/README.md
@@ -310,6 +310,14 @@ no other validator is tried. The advertised endpoint may be `http` or `https`
 transit plaintext HTTP if a validator advertises it; TN adds no authentication or encryption of
 its own on this path.
 
+The dial target is likewise chosen by a committee member rather than by this node, so
+`ForwardTargetPolicy` gates it at the point an advertised URL becomes a provider. `PublicOnly`
+(the default) refuses loopback, private, link-local, unique-local, shared-address-space and
+unspecified IP literals in every spelling, plus the names reserved to resolve locally, and logs
+each refusal once at `warn` with the advertising validator's BLS key. Operators of single-host
+deployments opt back in with the `allow_private_forward_targets` parameter. The check never
+resolves DNS, so a hostname that resolves to an internal address is still dialed.
+
 ### `SYSTEM_ADDRESS` is unreachable by users
 
 `SYSTEM_ADDRESS` (`0xff…fe`, `src/system_calls.rs`) is crate-private and has no known private
--- a/crates/tn-reth/src/forward.rs
+++ b/crates/tn-reth/src/forward.rs
@@ -9,7 +9,7 @@
 //! not advertised an endpoint (or is momentarily unreachable) is skipped in favor of one that
 //! has.
 //!
-//! Two properties worth knowing at this boundary:
+//! Three properties worth knowing at this boundary:
 //!
 //! - Transport security is whatever the validator advertised. Endpoint validation accepts both `https://`
 //!   and plain `http://` URLs, and the forwarder adds no encryption of its own, so a plain-HTTP
@@ -21,6 +21,12 @@
 //!   share consensus state, so the rejection would repeat everywhere. Only endpoint-local failures
 //!   (timeout, transport error, full pool, internal error) fall through to the next advertised
 //!   validator, and "already known" counts as delivered.
+//! - The dial target is chosen by a committee member, not by this node. The endpoint arrives inside
+//!   a BLS-signed node record, so an arbitrary network peer cannot inject one, but a committee
+//!   member can still advertise any host it likes and every observer will dial it unattended.
+//!   [`ForwardTargetPolicy`] therefore refuses non-public hosts at the dial site by default, so a
+//!   committee member cannot aim an observer's outbound HTTP at hosts inside that observer's own
+//!   perimeter (issue #1092).
 //!
 //! [`submit_txn_if_mine`]: tn_types::BatchValidation::submit_txn_if_mine
 
@@ -31,12 +37,14 @@ use alloy::{
 };
 use std::{
     collections::{BTreeMap, BTreeSet},
+    net::{Ipv4Addr, Ipv6Addr},
     sync::{Arc, Mutex},
     time::Duration,
 };
 use tn_types::{BlsPublicKey, RpcInfo, TaskSpawner, TxnForwarder};
 use tokio::time::timeout;
 use tracing::{debug, warn};
+use url::{Host, Url};
 
 /// Bounds a single validator's `eth_sendRawTransaction` round-trip so one unresponsive endpoint
 /// cannot stall the fallback chain; on timeout the forward tries the next validator.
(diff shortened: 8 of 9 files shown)


grantkee (approved):
LGTM!

Haxatron:
Fix verified. All common private IP addresses are covered,

two caveats:

1) a committee member that advertises http://evil.example.com/ with an A record pointing at 169.254.169.254 still gets dialed, this is acknowledged in both the PR and the README and acknowledged as increasing the cost for the attacker
2) HTTP redirects to internal IP addresses are still possible

Having said that, the issue is limited in nature due to the fact that it is blind (no response is returned) and only committee validators can trigger it.  So this is a proper hardening fix but does not fully cover the issue - operators can and should also configure external firewalls to fully lockdown this vector, based on the environment they are running in

PR #1115, opened 2026-08-07
fix(config): refuse an unset basefee_address at the production entry points

### LINKED ISSUE #1113: `config`: `basefee_address` is consensus-critical but node-local, and an omitted key silently selects a different fee recipient
**Class:** config handling / consensus safety.  Confirmed on `origin/main` (`efc104cc`).  `parameters.basefee_address` selects the account that every transaction credits with base fees, so it enters the state root.  It is the only field in `Parameters` that declares no serde default, and neither `Parameters::validate` nor `Parameters::validate_operational_floors` checks it.  A node whose `parameters.yaml` omits the key parses without error and credits a different address than its peers, so its state roots diverge.  **This does not affect `--chain mainnet` or `--chain adiri`**, because `Config::load_mainnet` and `Config::load_adiri` build `Parameters` from embedded constants and never read the on-disk file.  It applies to networks founded through the genesis ceremony, which read the operator's own file.  Severity is therefore **Medium**: the consequence is a chain split, but the trigger needs a deployment that does not use a named chain preset.  The repo already documents the consequence itself at `crates/tn-reth/src/lib.rs:180` and `:197`;  what is missing is any mechanism that stops two honest operators from disagreeing.

## Summary

The EVM handler credits the base-fee portion of gas to the configured address on every transaction (`crates/tn-reth/src/evm/handler.rs:182-185`):

```rust
context
    .journal_mut()
    .load_account_mut(self.basefee_address)?
    .incr_balance(U256::from(basefee.saturating_mul(gas_used)));
```

That balance change flows into the state root.  The crate says so in its own words at `crates/tn-reth/src/lib.rs:177-180`:

```
/// (`transact_raw`) reads this global via [`basefee_address`] and credits the base-fee
/// portion of gas (`reward_beneficiary`) plus the gas-limit penalty
/// (`reimburse_caller`) to it. Those balance changes flow into the state root, which
/// makes this value CONSENSUS-CRITICAL.
```

and at `:196-197`:

```
/// address, its state roots diverge from peers honoring the configured address, and
/// the node FORKS off the network.
```

The value reaches the handler from each node's own `Parameters` (`crates/node/src/manager/node.rs:1105`).  Its declaration carries no default and no documentation of the fallback (`crates/config/src/node.rs:277-278`):

```rust
/// If set the Address that will recieve basefees.
pub basefee_address: Option<Address>,
```

Every other field in the struct declares its default explicitly.  The immediate neighbours are `#[serde(default = "Parameters::default_batch_vote_timeout")]` at `:275` and `#[serde(default = "Parameters::default_parallel_fetch_request_delay_interval")]` at `:281`.  `gc_depth`, the closest analogue, declares `#[serde(default = "Parameters::default_gc_depth")]` at `:258`.

Because serde deserializes a missing `Option<T>` field to `None` without an explicit `#[serde(default)]`, a `parameters.yaml` that omits the key parses cleanly.  `None` then resolves to `GOVERNANCE_SAFE_ADDRESS` three crates away (`crates/tn-reth/src/env/mod.rs:107-109`, falling back through `crates/tn-reth/src/lib.rs:209` to `crates/config/src/genesis.rs:37`, `0x00000000000000000000000000000000000007a0`).  Nothing logs the substitution and nothing reports the effective address at startup.

## The defect

For a network that reads its own `parameters.yaml`, all three rows below start a node successfully:

| `parameters.yaml` | parsed value | address credited by the EVM | agrees with a peer holding `0x9999...9999` |
| --- | --- | --- | --- |
| `basefee_address: "0x9999...9999"` | `Some(0x9999...9999)` | `0x9999...9999` | yes |
| key absent | `None` | `0x...07a0` (`GOVERNANCE_SAFE_ADDRESS`) | **no, state roots diverge** |
| `basefee_address: "0xdead...beef"` | `Some(0xdead...beef)` | `0xdead...beef` | **no, state roots diverge** |

Divergence begins at the first transaction that pays a non-zero base fee.  There is no error, no warning, and no log line that distinguishes the three rows.  The two nodes agree on the batch, agree on the certificate, and agree on the commit order.  They disagree only on the resulting state root, which is the hardest class of split to diagnose from logs.

Both mainnet and adiri commit an explicit value, and they differ from each other:

```
chain-configs/mainnet/parameters.yaml:14:basefee_address: "0x9999999999999999999999999999999999999999"
chain-configs/testnet/parameters.yaml:14:basefee_address: "0x00000000000000000000000000000000000007a0"
```

Adiri's value equals `GOVERNANCE_SAFE_ADDRESS`, so an adiri node that lost the key would still agree.  Mainnet's does not, so mainnet is the case where an omitted key produces a silent split.

## Why the named chains are safe today

`Config::load_mainnet` (`crates/config/src/node.rs:146`) and `Config::load_adiri` (`:117`) build their `Parameters` solely from the embedded `MAINNET_PARAMETERS` and `TESTNET_PARAMETERS` constants (`:154` and `:125`).  Neither function reads the operator's `parameters.yaml` at any point;  the only file either reads from the datadir is `node_info_path()`.  A mainnet or adiri node therefore takes this value from the binary, and an operator cannot change it by editing a file.  That is a real guard and this issue does not claim otherwise.

The exposure is the path that has no preset.  `crates/telcoin-network-cli/src/genesis/mod.rs:45-53` gives the genesis ceremony a `--basefee-address` flag with `default_value_t = GOVERNANCE_SAFE_ADDRESS`, and `:321` writes `parameters.basefee_address = Some(self.basefee_address);` into the generated file.  So a freshly generated `parameters.yaml` does carry the key.  The gap is everything after generation: nothing checks that all validators on the network hold the same value, nothing rejects a file that lost the key through a hand edit or a partial copy, and nothing surfaces the effective value so an operator could compare it against a peer.

`Parameters::validate` (`crates/config/src/node.rs:388-402`) and `Parameters::validate_operational_floors` (`:425`) bound `gc_depth`, `max_header_num_of_batches`, and `header_num_of_batches_threshold`.  Neither mentions `basefee_address`.  This is the same family as #954, which covers `gc_depth` and `max_header_num_of_batches` as node-local consensus parameters with only ceiling checks;  `basefee_address` was not part of it, and it is the more direct of the three, because it changes the state root on the first transaction rather than only under a sustained asynchronous period.

## A second, currently unreachable hazard in the same field

`crates/tn-reth/src/lib.rs:182-197` documents a first-write-wins `OnceLock`: the first `RethEnv` built in a process pins `BASEFEE_ADDRESS`, and `set_basefee_address` discards the result (`crates/tn-reth/src/env/mod.rs:109`, `let _ = BASEFEE_ADDRESS.set(...)`).  Two constructors pass `None`: `RethEnv::new_for_temp_chain` and `SnapshotRestorer::open`.

I did not find a production path where either runs before `RethEnv::new` inside a single node process, so this is not a live defect.  It is worth a regression test, because the invariant that keeps it safe is process ordering that no test pins.

## Suggested fix

**Option A (recommended): source the address from genesis so a mismatch is impossible.**  `crates/tn-reth/src/lib.rs:205-206` already states the intended model: *"Per chain the address is fixed and only changes via a hard fork."*  A value with those semantics belongs in the artifact every node already agrees on, which is how the mainnet and adiri presets treat it in effect.  Reading it from a per-node file contradicts the documented model for exactly the deployments that have no preset.  This removes the failure mode rather than reporting it.

**Option B (smaller, detection rather than prevention): make the fallback explicit and observable.**  Three parts, all load-bearing:

1. Declare the default at the field, so `crates/config/src/node.rs:278` states its own fallback like every neighbour:

   ```rust
   /// The address that receives all transaction base fees.
   ///
   /// CONSENSUS-CRITICAL: this address is credited on every transaction, so its balance
   /// enters the state root. All nodes on a network must hold the same value or their
   /// state roots diverge. See the `BASEFEE_ADDRESS` docs in `tn-reth`.
   #[serde(default = "Parameters::default_basefee_address")]
   pub basefee_address: Option<Address>,
   ```

   with `fn default_basefee_address() -> Option<Address> { Some(GOVERNANCE_SAFE_ADDRESS) }`.  This changes no behavior.  It moves the fallback from serde's implicit `Option` handling to a named function at the declaration site, which is where a reader looks.

2. Log the effective address once at startup, at INFO, beside the other consensus parameters.  Today nothing prints it, so two diverging operators have no way to compare without reading each other's files.

3. Add `basefee_address` to `Parameters::validate_operational_floors` and reject `None` at the production entry points.  A production node that reached the fallback implicitly is far more likely to have lost the key than to have chosen it.  The named-chain presets set the field before validation runs, so this does not affect mainnet or adiri.

Option B still lets two operators configure two different non-`None` addresses.  Only Option A closes that.

## Suggested test design

- **Deserialization:** a `parameters.yaml` with the key absent parses to the documented default, asserting the concrete address rather than only that the field is `Some`.  A test that asserts `is_some()` passes vacuously under either behavior.
- **Consensus-criticality:** execute one identical batch under two `Parameters` that differ only in `basefee_address` and assert the two state roots differ.  This pins the claim the doc comment makes, so a later change cannot quietly treat the field as cosmetic.
- **`OnceLock` ordering:** assert that a `None`-passing constructor running first in a process pins the fallback and that a later `RethEnv::new` with a configured address does not override it.  This records the current behavior so a future refactor that makes a `None` path reachable in the node process fails a test instead of forking a chain.
- **Confirm-by-mutation:** set `chain-configs/mainnet/parameters.yaml:14` to `GOVERNANCE_SAFE_ADDRESS` and confirm the consensus-criticality test fails.  A test that passes with the mutation in place is pinning nothing.

## Acceptance criteria

- `basefee_address` is sourced from genesis for networks with no chain preset, and a node whose local value disagrees is either overridden or refused at startup, pinned by a test confirmed by mutation.
- Or the field declares its default at `crates/config/src/node.rs:278`, the node logs the effective address at startup, `validate_operational_floors` rejects `None` on the production entry points, and a test pins that two different addresses give two different state roots.
- Or this is closed as wontfix with a one-line note that every validator on a custom network receives one canonical `parameters.yaml` by operational procedure, so the assumption is written down where the next reader of `node.rs:278` will find it.

In all cases `make attest` stays green on the pinned 1.94 toolchain.

## Provenance

Found on 2026-08-06 during a fork-risk audit of `origin/main` `efc104cc`, which asked what in the tree could currently cause unintended forking.  Of the candidates that survived that audit, this is the only one with no existing issue or PR.

Not a duplicate.  `gh search issues` and `gh search prs` for `basefee_address` return no open item.  The only matches are merged #1047 and closed #1046, which changed saturation behavior inside `reward_beneficiary` and did not touch where the address comes from.  #954 covers `gc_depth` and `max_header_num_of_batches` under the same node-local-consensus-parameter theme and is the natural place to fold this if the maintainers prefer one issue for the family;  it does not mention this field today.

One question this issue cannot answer from the repo alone: whether real non-preset deployments distribute a single canonical `parameters.yaml` to every validator.  If they do, this is latent rather than live, and acceptance criterion three is the correct outcome.  If each operator generates or edits their own, it is live.

MavenRain:
Closes #1113.

## Problem

`parameters.basefee_address` selects the account that the EVM credits with the base-fee portion of
every transaction.  That balance enters the state root, so the value is consensus-critical.  The
field was an `Option` with no serde default, and no validation step examined it.

A `parameters.yaml` that omits the key parsed without an error.  serde deserializes a missing
`Option<T>` to `None`, and `None` selects `GOVERNANCE_SAFE_ADDRESS` three crates away, inside
`set_basefee_address`.  Nothing logged the substitution and nothing reported the effective address.

Two operators on the same network can therefore run with two different fee recipients.  The nodes
agree on the batch, on the certificate, and on the commit order.  They disagree only on the
resulting state root, which is the most difficult class of split to diagnose from logs.

`--chain mainnet` and `--chain adiri` are not affected.  `Config::load_mainnet` and
`Config::load_adiri` build their `Parameters` from the embedded constants and never read the
operator's file.  The exposure is the path with no chain preset, which is the path that networks
founded through the genesis ceremony use.

## Root cause

The fallback was correct but invisible.  It lived in `tn-reth`, three crates away from the
declaration an operator reads, and no validation step connected the two.  An absent key and a
deliberate choice of the governance address produced the same result, so the configuration layer
could not tell a lost key from an intended value.

## Fix

Make the key required at the type, and log the effective value.

Per the review discussion, `basefee_address` is now a plain `Address` with no serde default.  A
`parameters.yaml` that omits the key fails to deserialize with an error that names the field.
That refusal holds for every constructor and entry point, not only the guarded production ones,
so the runtime floor check an earlier revision of this PR added is gone as dead code.

The field deliberately declares no `#[serde(default)]`.  A default would fill a fallback on a
missing key in silence, which is the exact failure mode of #1113 moved one layer up.

`Parameters::default()` states `Some`thing concrete: `GOVERNANCE_SAFE_ADDRESS`, the address an
unset `Option` historically resolved to.  `Default` is the test-fixture path;  production reads a
file, and a file without the key does not parse, so this value never reaches a production node.

The value stays data rather than a const: the two shipped presets commit two different addresses,
`telcoin-network genesis --basefee-address` sets a per-network value at the ceremony, and a preset
edit before genesis must not require a coordinated binary upgrade.

No behavior changes for `--chain mainnet`, for `--chain adiri`, or for any file that already
carries the key.  The genesis ceremony already writes the key, so it produces a valid file.

## Changes

- [x] `crates/config/src/node.rs`: `basefee_address` is a plain `Address` documented as
      consensus-critical, with no serde default, so an absent key is a hard deserialize error.
- [x] `crates/config/src/node.rs`: `Parameters::default()` states the governance address, which
      keeps every test fixture on its old effective value.
- [x] `crates/config/src/node.rs`: `Parameters::tracing()` logs the effective base-fee recipient at
      INFO, beside the other consensus parameters, so two operators can compare the value without
      reading each other's files.
- [x] `crates/node/src/manager/node.rs`: the production caller passes `Some(address)` to
      `RethEnv::new`, whose `Option<Address>` parameter remains for the test-facing constructors.
- [x] `crates/tn-reth/src/evm/handler.rs`: the reward test helper takes the collector address, so a
      test can execute the same transaction under two different addresses.
- [x] `crates/telcoin-network-cli/README.md` and `crates/tn-reth/README.md`: record that the key is
      required and that a file without it fails to parse, and fix a stale "if the file is absent,
      defaults are used" lead-in that contradicted the loader.
- [x] `crates/tn-reth/src/env/mod.rs` and `crates/tn-reth/src/lib.rs`: document exactly which
      constructors still pass `None` (genesis tooling, `db load-state`, and tests) and that the
      production node start always supplies the configured address.

## Testing

Four tests.  All of them fail if the change is reverted.

- `absent_basefee_address_key_fails_to_deserialize` pins the parse failure of a file that omits
  the key and requires the error to name the field.  Every other key carries a serde default, so
  the partial file isolates the one required key;  a `#[serde(default)]` reintroduced on the field
  turns this test red.
- `shipped_chain_presets_pin_their_basefee_address` pins both committed addresses by value and
  confirms both presets satisfy the operational floors.  The addresses are consensus-critical, so a
  change to either is a hard fork for that chain, and updating this test is the intended cost of
  making one.
- `differing_basefee_addresses_credit_different_accounts` executes one identical transaction under
  two different base-fee addresses and asserts that the two runs credit two different accounts.
  The credited balances are part of the state, so the two runs produce two different state roots.
  This pins the consensus-criticality claim that the `BASEFEE_ADDRESS` documentation makes, so a
  later change cannot treat the field as cosmetic without a test failure.
- `default_parameters_pass_operational_floors` keeps the fixture default valid against the floors.

## Alternative considered

Issue #1113 also proposes sourcing the address from genesis, which removes the failure mode
instead of reporting it.  That is the stronger fix, and `tn-reth` already documents the matching
model: "Per chain the address is fixed and only changes via a hard fork."  It changes the genesis
artifact and the ceremony, so it is a maintainer decision rather than something to fold into this
change.  This PR takes the smaller option and leaves that door open.  This PR does not stop two
operators from configuring two different addresses.  Only the genesis-sourced option closes that.

One adjacent gap stays open at the ceremony layer: `telcoin-network genesis` gives
`--basefee-address` a clap default of `GOVERNANCE_SAFE_ADDRESS`, so a founder who omits the flag
still writes the governance safe with no prompt.  The written file is explicit and every founder
gets the same value, so it is not a fork risk, but making the flag required would close the last
silent-choice path.  That is a CLI-behavior decision, so it is left out of this change.

### DIFF
--- a/crates/config/src/node.rs
+++ b/crates/config/src/node.rs
@@ -282,14 +282,18 @@ pub struct Parameters {
     /// build different state roots from the same committed output and the network splits. See the
     /// `BASEFEE_ADDRESS` documentation in `tn-reth` for the execution-side detail.
     ///
-    /// An absent key deserializes to `None`. That is declared here rather than left to serde's
-    /// implicit handling of `Option`, because the fallback it selects is three crates away.
-    /// `None` stays legal for test fixtures, but [`Parameters::validate_operational_floors`]
-    /// refuses it at the production entry points, so an operator must state the address. The
-    /// chain presets set it from their embedded parameters, so `--chain mainnet` and
-    /// `--chain adiri` never reach this path.
-    #[serde(default)]
-    pub basefee_address: Option<Address>,
+    /// The key is required: a parameters file that omits it fails to deserialize with an error
+    /// that names the field, wherever that file is parsed. (The one adjacent gap is a missing
+    /// FILE under [`Config::load_or_default`], which substitutes the complete
+    /// [`Parameters::default`] and writes it back to disk; its only caller today is a test
+    /// helper, and the production node start loads through `Config::load`, which errors on a
+    /// missing file.) This field deliberately declares no serde default. A default here would fill
+    /// a fallback on a missing key in silence, which is the exact failure mode of issue #1113
+    /// moved one layer up. Test fixtures obtain an address through [`Parameters::default`],
+    /// which supplies `GOVERNANCE_SAFE_ADDRESS`. The chain presets and the genesis ceremony
+    /// both write the key, so `--chain mainnet`, `--chain adiri`, and ceremony-founded
+    /// networks all parse.
+    pub basefee_address: Address,
     /// The default duration between parallel/fallback fetch requests to peers for missing
     /// certificates.
     #[serde(default = "Parameters::default_parallel_fetch_request_delay_interval")]
@@ -377,11 +381,11 @@ impl Default for Parameters {
             max_batch_delay: Parameters::default_max_batch_delay(),
             max_concurrent_requests: Parameters::default_max_concurrent_requests(),
             batch_vote_timeout: Parameters::default_batch_vote_timeout(),
-            // Explicit, and identical in effect to the previous `None`: an unset address already
-            // resolved to `GOVERNANCE_SAFE_ADDRESS` inside `set_basefee_address`. Stating it here
-            // keeps the default satisfying `validate_operational_floors`, which now refuses
-            // `None`.
-            basefee_address: Some(GOVERNANCE_SAFE_ADDRESS),
+            // The test-fixture path. Production reads a parameters file, and a file without the
+            // key fails to deserialize, so this value never reaches a production node. The
+            // governance safe is the address an unset `Option` historically resolved to inside
+            // `set_basefee_address`, so fixtures keep their old effective value.
+            basefee_address: GOVERNANCE_SAFE_ADDRESS,
             parallel_fetch_request_delay_interval:
                 Parameters::default_parallel_fetch_request_delay_interval(),
         }
@@ -440,13 +444,10 @@ impl Parameters {
     ///   but includes at most `max_header_num_of_batches` of them, so a zero threshold seals empty
     ///   headers on the fast path and a threshold above the max makes the two conditions mutually
     ///   inconsistent.
-    /// - `basefee_address` must be set. The EVM credits this account on every transaction, so its
-    ///   balance enters the state root and every node on the network must agree on it. A
-    ///   `parameters.yaml` that omits the key parses without error, deserializes to `None`, and
-    ///   silently selects `GOVERNANCE_SAFE_ADDRESS`, which splits state roots against every peer
-    ///   holding a different value. Refusing `None` here converts that silent split into a startup
-    ///   error that names the field. The chain presets and the genesis ceremony both write the key,
-    ///   so this rejects a file that lost it, not a supported configuration.
+    ///
+    /// `basefee_address` needs no floor here: the field is a required key with no serde default,
+    /// so a `parameters.yaml` that omits it already fails to deserialize, before any constructor
+    /// runs, with an error that names the field.
     pub fn validate_operational_floors(&self) -> eyre::Result<()> {
         eyre::ensure!(
             self.gc_depth > tn_types::GC_ACTIVITY_BUFFER,
@@ -470,11 +471,6 @@ impl Parameters {
             self.header_num_of_batches_threshold,
             self.max_header_num_of_batches,
         );
-        eyre::ensure!(
-            self.basefee_address.is_some(),
-            "basefee_address must be set explicitly: it receives every transaction's base fee, so \
-             its balance enters the state root and all nodes on the network must agree on it",
-        );
         Ok(())
     }
 
@@ -494,7 +490,7 @@ impl Parameters {
         // reading each other's files.
         info!(
             "Basefee address set to {} (consensus-critical: all nodes must agree)",
-            self.basefee_address.unwrap_or(GOVERNANCE_SAFE_ADDRESS)
+            self.basefee_address
         );
     }
 }
@@ -608,28 +604,32 @@ mod test {
         );
     }
 
-    /// A parameters file that omits `basefee_address` parses without error and yields `None`.
+    /// A parameters file that omits `basefee_address` must fail to deserialize.
     ///
-    /// This asserts the concrete value rather than `is_some()`, which would pass under either
-    /// behavior and pin nothing.
+    /// Every other key in the file carries a serde default, so this partial file isolates the one
+    /// required key. Without this pin, a `#[serde(default)]` reintroduced on the field would fill
+    /// a fallback on a missing key in silence, which is the failure mode of issue #1113.
     #[test]
-    fn absent_basefee_address_key_deserializes_to_none() {
-        let params: Parameters =
-            serde_yaml::from_str("gc_depth: 50").expect("a partial parameters file parses");
-        assert_eq!(
-            params.basefee_address, None,
-            "an omitted key must stay observable as None so the floors can refuse it"
+    fn absent_basefee_address_key_fails_to_deserialize() {
+        let err = serde_yaml::from_str::<Parameters>("gc_depth: 50")
+            .expect_err("a parameters file without basefee_address must not parse");
+        assert!(
+            err.to_string().contains("basefee_address"),
+            "the error must name the field an operator has to fix: {err}"
         );
     }
 
-    /// A production node must not start on a file that lost the key. Without this check the node
-    /// starts, credits the fallback address, and splits state roots against its peers in silence.
+    /// A parameters file with an explicit `basefee_address: null` must also fail to deserialize.
+    ///
+    /// This is the upgrade shape, not a hypothetical: before this change `Parameters::default()`
+    /// held `None`, and `load_from_path_or_default` writes the default back to disk when the file
+    /// is absent, so a datadir created under the old code can hold `basefee_address: null`. That
+    /// file takes serde's type-error path, not the missing-field path the previous test pins, so
+    /// it needs its own pin.
     #[test]
-    fn operational_floors_reject_absent_basefee_address() {
-        let params = Parameters { basefee_address: None, ..Default::default() };
-        let err = params
-            .validate_operational_floors()
-            .expect_err("a production node must not start without an explicit basefee address");
+    fn null_basefee_address_key_fails_to_deserialize() {
+        let err = serde_yaml::from_str::<Parameters>("gc_depth: 50\nbasefee_address: null")
+            .expect_err("a parameters file with a null basefee_address must not parse");
         assert!(
             err.to_string().contains("basefee_address"),
             "the error must name the field an operator has to fix: {err}"
@@ -648,12 +648,11 @@ mod test {
             serde_yaml::from_str(TESTNET_PARAMETERS).expect("adiri parameters parse");
         assert_eq!(
             mainnet.basefee_address,
-            Some(address!("0x9999999999999999999999999999999999999999")),
+            address!("0x9999999999999999999999999999999999999999"),
             "mainnet's committed basefee address changed"
         );
         assert_eq!(
-            adiri.basefee_address,
-            Some(GOVERNANCE_SAFE_ADDRESS),
+            adiri.basefee_address, GOVERNANCE_SAFE_ADDRESS,
             "adiri's committed basefee address changed"
         );
         mainnet
--- a/crates/config/src/node.rs
+++ b/crates/config/src/node.rs
@@ -1,6 +1,6 @@
 //! Configurations for the Telcoin Network.
 
-use crate::{ConfigFmt, ConfigTrait, NodeInfo, TelcoinDirs};
+use crate::{ConfigFmt, ConfigTrait, NodeInfo, TelcoinDirs, GOVERNANCE_SAFE_ADDRESS};
 use reth_chainspec::ChainSpec;
 use serde::{Deserialize, Serialize};
 use std::{fs::File, io::Write, time::Duration};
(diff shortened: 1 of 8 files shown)


crates/config/src/node.rs
```diff
@@ -445,6 +470,11 @@ impl Parameters {
             self.header_num_of_batches_threshold,
             self.max_header_num_of_batches,
         );
+        eyre::ensure!(
+            self.basefee_address.is_some(),
+            "basefee_address must be set explicitly: it receives every transaction's base fee, so \
+             its balance enters the state root and all nodes on the network must agree on it",
+        );
```
grantkee: Is there any benefit to leaving this as an `Option` anymore? I think tests could use `default` just as easily as `None`

grantkee (dismissed):
LGTM - should we just remove the `Option` type and make this a const? Originally, we were avoiding this for tests (I think?) but now we are much more confident in how we're using this basefee address and it isn't really configurable without causing forks

MavenRain:
I'm going to say 'yes' to removing the `Option`, 'no' to a const.

**Remove the `Option`.**  I can make the field a plain `Address` with no serde default.  A missing yaml key then fails at deserialize time with a hard error.  That is earlier and stronger than the runtime floor check this PR adds, and it holds for every constructor, not only the guarded entry points.  Test fixtures keep working through `Parameters::default()`, which supplies `GOVERNANCE_SAFE_ADDRESS`.  You are right that tests do not need `None` for this.

**Keep it data, not a const.**  Three facts in the tree block a const today:

1. The two shipped presets commit two different addresses.  Mainnet pins `0x9999...9999` and testnet (adiri) pins `0x...07a0` (`chain-configs/*/parameters.yaml:14`).  The new pin test in this PR asserts both values.  One const cannot hold both.
2. Chains founded with `telcoin-network genesis` set their own address with `--basefee-address`.  A const removes that knob, and the non-named-chain path (`Config::load`) has no chain-id dispatch to hang per-chain consts on.
3. `chain-configs/README.md` says the presets are placeholders, so the mainnet value will likely change before launch.  As yaml data, that change is a preset edit before genesis.  As a const, it becomes a coordinated binary upgrade, and one straggler node on the old binary forks silently.

One trap to avoid in the follow-up: a `#[serde(default = ...)]` on the plain field.  That would silently fill a fallback on a missing key, which is the exact failure mode of #1113 moved one layer up.  So: required key, hard error.

On the history: the field was already `Option` in the earliest reachable commit, so I could not confirm the original reason.  The OnceLock comment from that same commit ("this will break some test environments") supports your recollection.

If this sounds right I will push the `Option` removal onto this branch.

grantkee:
> I'm going to say 'yes' to removing the `Option`, 'no' to a const.
> 
> **Remove the `Option`.** I can make the field a plain `Address` with no serde default. A missing yaml key then fails at deserialize time with a hard error. That is earlier and stronger than the runtime floor check this PR adds, and it holds for every constructor, not only the guarded entry points. Test fixtures keep working through `Parameters::default()`, which supplies `GOVERNANCE_SAFE_ADDRESS`. You are right that tests do not need `None` for this.
> 
> **Keep it data, not a const.** Three facts in the tree block a const today:
> 
> 1. The two shipped presets commit two different addresses.  Mainnet pins `0x9999...9999` and testnet (adiri) pins `0x...07a0` (`chain-configs/*/parameters.yaml:14`).  The new pin test in this PR asserts both values.  One const cannot hold both.
> 2. Chains founded with `telcoin-network genesis` set their own address with `--basefee-address`.  A const removes that knob, and the non-named-chain path (`Config::load`) has no chain-id dispatch to hang per-chain consts on.
> 3. `chain-configs/README.md` says the presets are placeholders, so the mainnet value will likely change before launch.  As yaml data, that change is a preset edit before genesis.  As a const, it becomes a coordinated binary upgrade, and one straggler node on the old binary forks silently.
> 
> One trap to avoid in the follow-up: a `#[serde(default = ...)]` on the plain field. That would silently fill a fallback on a missing key, which is the exact failure mode of #1113 moved one layer up. So: required key, hard error.
> 
> On the history: the field was already `Option` in the earliest reachable commit, so I could not confirm the original reason. The OnceLock comment from that same commit ("this will break some test environments") supports your recollection.
> 
> If this sounds right I will push the `Option` removal onto this branch.

Good point re not using a const and just removing the `Option`. I think that's the best path forward

sstanfield (approved):

Issue #1106, opened 2026-08-05
node: bound derive_idle_worker_fee_at so history retention depth is expressible


MavenRain:
## Problem

`derive_idle_worker_fee_at` (`crates/node/src/manager/node.rs`) makes the safe history-retention
depth of a Telcoin node unbounded, which in turn makes "how much history does a node actually need?" unanswerable.

The function loops `k = epoch, epoch-1, ...`, performing a **pinned historical state read** at each prior epoch's closing block (`get_worker_fee_configs_at_block`, `get_epoch_info_at_block`, both of which route through `pinned_state_and_env` in `crates/tn-reth/src/env/epoch.rs`) until it finds one of: a Static-strategy anchor, a genuine produced block, a freshly created slot, or epoch 0.

Its own doc comment states the bound honestly: "k strictly decreases to an anchor or epoch 0, so the walk terminates in at most `epoch` steps". Termination is not the issue. The issue is that the terminating condition is a property of *worker activity*, not of a fixed depth: for a chronically idle non-Static worker the walk runs all the way to epoch 0, i.e. a state read at the chain's full lifetime.

This is not a rare path. It runs automatically at every epoch entry through
`derive_base_fees_for_entered_epoch`.

## Why it matters

Five of the six pinned-read families behind `pinned_state_and_env` are bounded to the canonical or finalized tip, or exactly one epoch boundary back. This walk is the sole exception, and it is
therefore the *only* reason archive mode has to be the operating assumption rather than a generous retention window.

@sstanfield made the point while reviewing #1060: "a prune config that was so aggressive it removes blocks you still need is not very useful ... I don't know why we would not be using it on very recent history." He is right about the other five families. This walk is what prevents that observation from being turned into a configuration.

Concretely, #1060 had to enforce archive mode with a blanket "any pruning configuration refuses to start" check, because there is no defensible retention number to enforce instead. Bounding this walk is the precondition for replacing that ban with a depth.

It also interacts with the snapshot-restore gap: a restored node whose workers idle across several epochs after entry can walk back past `window_start`, where pinned reads currently resolve to `Ok(None)` silently. `SnapshotRestorer::check_fee_precondition` does not cover this, since it validates only the single epoch preceding the entered epoch.

## Proposed direction

Give the walk an explicit floor rather than "an anchor or epoch 0", and define what happens when the floor is reached (a documented fallback fee, or a hard error).

This is design work rather than a drive-by change: the walk feeds base-fee derivation, so any
[...]

PR #1131, opened 2026-08-10
fix(tn-reth): refuse pinned reads below the restored-state floor (#1105)

### LINKED ISSUE #1105: tn-reth: pinned reads below a restored snapshot window resolve silently to Ok(None)
## Problem

A snapshot-restored node silently answers pinned consensus-registry reads that fall below the
shipped history window, and nothing in the node notices.

`db load-state` (#985) and the export side (#973) scaffold the chain below the shipped window: every
block below `window_start` is written as a placeholder `ExecHeader { number, ..Default::default() }`
with hash `B256::ZERO`, and only the window's anchor block carries real account and storage state
(`crates/tn-reth/src/snapshot.rs`). The restorer writes ordinary `StageCheckpoint`s and never a
`PruneCheckpoint`.

That combination lands in reth's checkpoint-less branch. In
`crates/storage/provider/src/providers/state/historical.rs`, `is_account_history_available` is
`self.account_history_block_number.map(|b| b <= at).unwrap_or(true)`, so with no prune checkpoint it
reports *available* and no `StateAtBlockPruned` is ever raised. The lookup then falls through to
`HistoryInfo::NotYetWritten` (the `else` arm, since `lowest_available_block_number.is_some()` is
false) and the read returns `Ok(None)`. The account or storage slot reads as **never written**
rather than as its value at the pinned block. No error, no log, no metric.

Every pinned read in `crates/tn-reth/src/env/epoch.rs` routes through `pinned_state_and_env`, so
this affects committee construction, epoch entry, epoch records, and worker fee configs alike.

`RethConfig::ensure_archive_mode` (added in #1060) cannot see this: it keys off
`NodeConfig::prune_config()`, and no pruning is configured anywhere on this path. The source guard
`etc/archive-mode-guard.sh` cannot see it either, since no pruner exists. This is the gap
@sstanfield identified while reviewing #1060 ("if you try to use a pinned block that is not in your
history then why not let the node die a noisy death?").

`SnapshotRestorer::check_fee_precondition` is not a general answer. It verifies only that every
Eip1559 worker produced a genuine block in the single epoch preceding the entered epoch, which is
scoped to seeding the first entered epoch's base fees.

## Why the obvious fix does not work

Checking reth's prune checkpoints at the read site is redundant *and* blind. Redundant because reth
already performs exactly that comparison and already returns `StateAtBlockPruned` when it fails.
Blind because the checkpoint row this case needs does not exist, which is precisely why the
watermark defaults to "available".

The finer-grained check is also unreachable through the handle tn-reth holds. reth does expose
`HistoricalStateProviderRef::account_history_lookup` returning `ProviderResult<HistoryInfo>`, but
`RethEnv::read_only_state_db` returns a `StateProviderBox` (`Box<dyn StateProvider>`), and
`StateProvider` does not carry those methods. Per-key "is this address's history actually present at
block N" would mean giving up the boxed trait object.

## Proposed fix

Use a TN-owned floor rather than a reth-derived one. We write the scaffold, so we know
`window_start` at restore time and do not need to infer it.

1. Persist the restored window's lower bound as node-local metadata at
   `SnapshotRestorer::finish` time.
2. Have `pinned_state_and_env` (`crates/tn-reth/src/env/epoch.rs`) reject
   `header.number < floor` with `StateReadError::Provider`.

This is checkpoint-independent, so it is not blind where reth's watermark is, and it slots into the
existing error arm: `StateReadError::Provider` is already the node-local classification, and the
call sites already implement a documented spread from hard halt (`close_epoch.rs` halts on both
error classes) through fail-open for `ChainGlobal` only (`run_epoch.rs`). No new error variant and
no new call-site policy is required.

Cost is one integer comparison per pinned read. Every traced caller runs at epoch-boundary or
startup cadence except the RPC-driven batched registry read, so this is negligible.

## Open question

Should a pinned read below the shipped window be a hard halt, or is `Ok(None)`-and-continue
acceptable for some readers? That turns on whether a snapshot-restored node is a supported
validator configuration, which is a call for @sstanfield rather than something to decide in the
implementation.

## References

- Reviewed out of #1060, from @sstanfield's review comment.
- reth v1.11.3 `crates/storage/provider/src/providers/state/historical.rs`
- `crates/tn-reth/src/snapshot.rs`, `crates/tn-reth/src/env/epoch.rs`

MavenRain:
Closes #1105.

## Problem

A snapshot-restored node silently answers pinned consensus-registry reads that fall below the
restored-state floor.  The restore scaffold imports state only at the snapshot's final block
`B`: window headers below `B` carry real, resolvable hashes but no state and no history, and
blocks below the window are zero-hash placeholders.  The restorer writes no `PruneCheckpoint`,
so reth's availability watermark (`is_account_history_available . . . unwrap_or(true)`) reports
the history as available.  The per-account history walk then resolves
`HistoryInfo::NotYetWritten` and the read returns `Ok(None)`: every account reads as never
written.  No error, no log, no metric.

Every pinned read routes through `pinned_state_and_env` (`crates/tn-reth/src/env/epoch.rs`), so
committee construction, epoch entry, epoch records, and worker fee configs are all exposed.  The
failure needs no attacker: an idle-worker epoch walk-back (#1106) or an RPC-driven historical
registry read is enough to pin a block below the floor.

## Fix

A TN-owned floor, as proposed in #1105, with two corrections:

1. The floor is the snapshot's final block `B`, not the window's first block.  The issue
   proposed `window_start`, but a pin strictly inside the window (`window_start <= n < B`)
   carries a real header hash, resolves normally, and then reads the same silent empty state:
   the datadir holds no state below `B` anywhere.  `B` is the first block whose state the
   datadir holds, so `B` is the floor.
2. `SnapshotRestorer::import_chain_scaffold` persists `B` into a `restored-state-floor` marker
   file at the datadir root, durably (write a temp file, fsync it, rename into place, fsync the
   datadir), BEFORE any chain data commits.  The mdbx and static-file commits that follow are
   fsynced, so ordering the bare-file marker first (and making its write atomic) means a crash
   at any point cannot surface committed restored data that is missing its floor, and a torn
   write cannot surface a half-written marker.

`RethEnv::new` loads the marker once into the env (`None` on a normally-synced node;  a corrupt
marker is a hard startup error with a remedy message, never `None`).  `pinned_state_and_env`
refuses `header.number < floor` with `StateReadError::Provider` before any state provider is
built.  `clean_created_chain_data` (the failed-import cleanup in `telcoin-network-cli`) removes
the marker LAST, after the chain-data dirs, so an aborted import cannot leave an orphan floor
that poisons a later normal sync of the same datadir, and a mid-cleanup crash never leaves
restored data floorless.

The classification is deliberate: the gap is this node's datadir, not the chain.  A peer with
full history answers the same read correctly, so the failure is not committee-deterministic and
callers must not fail open on it.  The existing call-site policy spread already handles this
class (`close_epoch.rs` halts on both error classes;  `run_epoch.rs` retries `Provider` and
fails open for `ChainGlobal` only), so no new error variant and no new call-site policy is
needed.  Cost on the hot path is one integer comparison per pinned read.

## Why not a reth `PruneCheckpoint`

Writing the missing checkpoint row at restore time looks like the reth-native fix (it would also
cover arbitrary historical RPC reads), but it is unsound here.  reth v1.11.3's
`HistoryInfo::from_lookup` returns `InChangeset(C)` instead of `NotYetWritten` whenever a prune
checkpoint EXISTS, without comparing the target block against the watermark
(`crates/storage/provider/src/providers/state/historical.rs`).  A checkpoint row would therefore
push in-window reads below `B` into change-set lookups, and the import deliberately writes no
change sets for `B` (see the "Why the block-`B` change sets are skipped" section on
`import_state`, whose soundness argument depends on the ABSENT prune boundary).  The TN-owned
floor stays outside reth's history resolution entirely.

## Changes

- [x] `crates/tn-reth/src/env/mod.rs`: `RESTORED_STATE_FLOOR_FILE` marker constant,
      `restored_state_floor` field on `RethEnvInner`, marker load in `RethEnv::new`
      (missing file is `Ok(None)`;  unreadable or unparseable is a hard startup error with a
      remedy message), the public `restored_state_floor_marker` path helper, and the durable
      `write_restored_state_floor` (temp file, fsync, rename, fsync the parent).  Module docs
      extended with the new construction invariant.
- [x] `crates/tn-reth/src/env/epoch.rs`: the floor guard in `pinned_state_and_env`, refusing
      below-floor pins as `StateReadError::Provider` with a message that names the floor, plus
      the RESTORED-DATADIR FLOOR doc section beside the existing ARCHIVE-MODE assumption.
- [x] `crates/tn-reth/src/snapshot.rs`: the scaffold writes the `B` marker before any chain
      data commits;  restore-flow and scaffold doc comments record the new step.
- [x] `crates/telcoin-network-cli/src/db.rs`: `clean_created_chain_data` removes the marker
      after the chain-data dirs it already removes.
- [x] Tests (see below).

## Scope notes

- `read_contract_inner` (`crates/tn-reth/src/helpers.rs`) builds its state provider directly
  via `read_only_state_db`, bypassing `pinned_state_and_env`, so ExEx and `eth_call`-style
  historical reads are not covered by the floor.  The consensus-critical pinned reads all route
  through the guarded path;  extending the floor to the direct path is a follow-up.
- Datadirs restored by binaries that predate this change carry no marker and keep the old
  behavior;  nothing on disk records their window, so there is no back-fill.  Re-running
  `db load-state` on a fresh datadir writes the marker.
- Whether a below-floor pin should hard-halt the node instead of surfacing through the existing
  per-caller `Provider` policy remains the #1105 open question (supported validator
  configuration);  this change keeps the documented policy spread untouched.

## Testing

- `scaffold_persists_restored_state_floor_and_env_refuses_below_it` (new, `snapshot.rs`): a
  window of blocks 3..=5 persists marker `5` (the snapshot's final block, NOT the window
  start);  a fresh env over the datadir refuses a pinned read at block 2 (below the window) as
  `Provider` with the floor message, before the pin hash is resolved;  a pinned read at block 3
  (INSIDE the window, real resolvable hash) is refused with the same message, which is exactly
  the case a `window_start` floor would miss;  at the floor itself the guard admits the pin
  (boundary control: a random hash at block 5 then fails with reth's unresolvable-hash message,
  not the floor's).
- `entry_readiness_accepts_epoch_boundary_snapshot` (extended, `snapshot.rs`): after a real
  export/restore/finish round trip, a fresh env loads the floor, refuses a genesis pin (genesis
  resolves in the datadir but its state was cleared by the scaffold) with the floor message, and
  still answers the same reader at `B` (positive control: the entered epoch reads back).
- Regression guards run unchanged: `restore_roundtrips_state_from_export`,
  `entry_readiness_rejects_non_boundary_final_block`,
  `test_state_read_classifies_provider_vs_chain_global`, and
  `epoch_state_at_epoch_start_epoch_zero_pins_genesis` (a normal node has no floor and genesis
  pins keep working).
- Commands: `cargo +nightly-2026-03-20 fmt -p tn-reth -p telcoin-network-cli`, `cargo +1.94
  check -p tn-reth -p telcoin-network-cli --all-targets`, `cargo +nightly-2026-03-20 clippy -p
  tn-reth -p telcoin-network-cli --all-targets --all-features`, `cargo +1.94 test -p tn-reth`
  (named tests), `cargo +1.94 test --no-run --workspace -j 2` in an isolated target dir, all on
  the repo-pinned toolchains.  All gates pass: fmt and clippy are clean, the six named tests
  pass 6/6, and the workspace test build compiles.
- The new tests were confirmed by mutation, three mutants, each restore byte-compared against
  the pre-mutation sources and the restored tree re-run green:
  - Guard removed (`epoch.rs`): both floor tests fail on the refusal-message assertion;  the
    error that surfaces without the guard is an incidental provider fault, which is the issue's
    premise.
  - Marker write deleted (`snapshot.rs`): both floor tests fail at the marker read because the
    file does not exist.
  - Floor written as `window_start` instead of `B` (`snapshot.rs`): the scaffold test fails at
    the marker-content assertion (the marker reads `3`, the window start, instead of `5`),
    upstream of the window-interior probe that the same floor would also fail by admitting the
    pin at block 3.

### DIFF
--- a/crates/telcoin-network-cli/src/db.rs
+++ b/crates/telcoin-network-cli/src/db.rs
@@ -316,10 +316,15 @@ fn chain_data_dirs(datadir: &Path) -> [PathBuf; 3] {
     [datadir.join("db"), datadir.join("static_files"), datadir.join("consensus-db")]
 }
 
-/// Best-effort removal of the chain-data dirs a failed import created — every [`chain_data_dirs`]
-/// entry not already present in `preexisting`. Skipping pre-existing dirs is a safety guard: even
-/// if `db load-state` were pointed at a populated datadir, this never deletes data the import did
-/// not create. Removal errors are logged, not fatal (the import is already failing).
+/// Best-effort removal of the chain data a failed import created — every [`chain_data_dirs`]
+/// entry not already present in `preexisting`, then the restored-state floor marker
+/// (`import_chain_scaffold` writes it before any chain data commits, so a failed import can
+/// leave it orphaned; a stale floor would poison a later normal sync of this datadir by
+/// refusing pinned reads it can actually serve). The marker goes last: removing it before the
+/// dirs could leave restored chain data without its floor if this cleanup itself dies midway.
+/// Skipping pre-existing dirs is a safety guard: even if `db load-state` were pointed at a
+/// populated datadir, this never deletes data the import did not create. Removal errors are
+/// logged, not fatal (the import is already failing).
 fn clean_created_chain_data(datadir: &Path, preexisting: &HashSet<PathBuf>) {
     for dir in chain_data_dirs(datadir) {
         if preexisting.contains(&dir) {
@@ -333,6 +338,12 @@ fn clean_created_chain_data(datadir: &Path, preexisting: &HashSet<PathBuf>) {
             }
         }
     }
+    let marker = tn_reth::RethEnv::restored_state_floor_marker(datadir);
+    fs::remove_file(&marker)
+        .or_else(|e| (e.kind() == std::io::ErrorKind::NotFound).then_some(()).ok_or(e))
+        .unwrap_or_else(|e| {
+            eprintln!("warning: could not remove {} during import cleanup: {e}", marker.display())
+        });
 }
 
 /// Reject an export bundle that cannot produce a resumable node, returning `Err` before any datadir
--- a/crates/tn-reth/src/env/epoch.rs
+++ b/crates/tn-reth/src/env/epoch.rs
@@ -760,10 +760,33 @@ impl RethEnv {
     /// `HistoryInfo::MaybeInPlainState`, silently falling back to TIP state for this "pinned"
     /// read — exactly the nondeterminism pinning exists to prevent. Revisit every pinned read
     /// before enabling pruning.
+    ///
+    /// RESTORED-DATADIR FLOOR: a datadir bootstrapped from a snapshot holds state ONLY at the
+    /// snapshot's final block `B` — window headers below `B` carry real, resolvable hashes but
+    /// no state and no history, and blocks below the window are zero-hash placeholders (see
+    /// `SnapshotRestorer::import_chain_scaffold` in `snapshot.rs`). reth's checkpoint-less
+    /// history walk answers reads anywhere below `B` with `Ok(None)` per account ("never
+    /// written") — no error, no log — so pins below the persisted floor (= `B`) are refused
+    /// HERE, before any state is built, as [`StateReadError::Provider`]: the gap is this node's
+    /// datadir, not the chain, and a peer with full history answers the same read correctly.
     fn pinned_state_and_env(
         &self,
         header: &SealedHeader,
     ) -> StateReadResult<(State<StateProviderDatabase<StateProviderBox>>, EvmEnv)> {
+        // refuse pins below the restored-state floor before touching the provider: below it the
+        // datadir holds headers but no state, and the read would resolve silently to "never
+        // written" values instead of failing
+        self.inner.restored_state_floor.filter(|floor| header.number < *floor).map_or(
+            Ok(()),
+            |floor| {
+                Err(StateReadError::Provider(format!(
+                    "pinned read at block {} is below this datadir's restored-state floor \
+                     (block {floor}, the snapshot's final block): the snapshot shipped no state \
+                     below it",
+                    header.number
+                )))
+            },
+        )?;
         let db = self.read_only_state_db(header.hash()).map_err(|e| {
             StateReadError::Provider(format!("state provider at {}: {e}", header.hash()))
         })?;
--- a/crates/tn-reth/src/env/mod.rs
+++ b/crates/tn-reth/src/env/mod.rs
@@ -14,8 +14,14 @@
 //!   the LOCAL chain spec in the supplied config — genesis is never fetched from the network.
 //! - [`RethEnv::new`] has two process-global side effects (base-fee address pinning and metrics
 //!   registration) — see its docs.
+//! - Opening an env loads the datadir's restored-state floor marker when one exists (a datadir
+//!   bootstrapped from a snapshot); pinned state reads below the floor are refused — see
+//!   `pinned_state_and_env` in `env/epoch.rs`.
 
-use std::{path::Path, sync::Arc};
+use std::{
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
 use reth_chainspec::ChainSpec as RethChainSpec;
 use reth_db::{init_db, DatabaseEnv};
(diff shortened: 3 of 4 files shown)


sstanfield (approved):
This is good.

PR #1137, opened 2026-08-11
fix(tn-reth): enforce the restored-state floor in read_only_state_db

### LINKED ISSUE #1136: tn-reth: read_contract_at_block bypasses the restored-state floor (#1131)
## Problem

`RethEnv::read_contract_at_block` (`crates/tn-reth/src/env/helpers.rs`) is a pinned read
that does not route through `pinned_state_and_env`: it builds its state from
`read_only_state_db` directly, via `read_contract_inner`. That means the restored-state
floor #1131 added is not consulted on this path. On a snapshot-restored datadir, a call
pinned below the restored window's floor resolves through reth's checkpoint-less history
walk and returns `Ok(None)`-shaped state (every account and slot reads as never written)
with no error, which is exactly the silent outcome the floor exists to refuse.

## Severity

Latent, not live. `read_contract_at_block` has no production caller today: repo-wide the
only call sites are test utilities (`crates/tn-reth/tests/it/pipeline_helpers.rs`,
`crates/tn-reth/src/env/helpers.rs` tests). But it is a public method on `RethEnv`, the
doc comment in `crates/exex/src/context.rs` recommends it as the pinned alternative to
tip reads, and `crates/tn-reth/src/evm/mod.rs` documents its gas ceiling, so the next
consumer inherits the gap silently.

## Proposed fix

Move the floor comparison below both entry points. Two candidate shapes:

1. Check `restored_state_floor` in `read_only_state_db` itself, keyed on the resolved
   block number, so every consumer of a pinned state provider inherits the refusal. This
   is the chokepoint the docs already name as the helper all pinned reads share.
2. Narrower: add the same floor rejection at the head of `read_contract_inner`, mirroring
   the check at the head of `pinned_state_and_env`.

Option 1 covers future bypasses by construction; it needs a look at whether any
`read_only_state_db` caller legitimately reads below the floor (none known).

## Related doc nit (same feature, could ride along)

The doc comment on `import_chain_scaffold` (`crates/tn-reth/src/snapshot.rs:524`) says the
floor marker is `window[0].number` (the window's first block). The code writes
`final_state.number` = `B` (the window's last block, `snapshot.rs:594`), and the rationale
comment directly above the write explains why `B` is correct. The one-line summary at :524
should say `B` as well.

## References

- #1131 (restored-state floor), #1105 (the underlying silent-read issue), #1060 (the
  archive-mode enforcement PR whose review surfaced this family)
- Found while reconciling #1060's archive-mode notes with the post-#1131 tree: the note
  claimed "every pinned read funnels through `pinned_state_and_env`", and auditing that
  claim exposed the one that does not.

MavenRain:
Closes #1136.

## Problem

`RethEnv::read_contract_at_block` is a pinned read that does not route through
`pinned_state_and_env`: it builds its state from `read_only_state_db` directly, via
`read_contract_inner`. The restored-state floor added in #1131 was enforced only in
`pinned_state_and_env`, so this path never consulted it. On a snapshot-restored datadir, a call
pinned below the floor resolves through reth's checkpoint-less history walk and reads every
account and slot as "never written". No error, no log. That is exactly the silent outcome the
floor exists to refuse (#1105).

No production caller hits this today (the only call sites are test utilities), but the method is
public on `RethEnv`, the exex docs recommend it as the pinned alternative to tip reads, and
`evm/mod.rs` documents its gas ceiling . . . the next consumer inherits the gap silently.

## Root cause

The floor comparison lived one layer too high. `pinned_state_and_env` checked it, but the shared
state constructor beneath it (`read_only_state_db`) did not, so a second entry point into that
constructor bypassed the guard.

## Fix

Move the floor refusal into `read_only_state_db` itself (option 1 from the issue), so every
consumer of a pinned state provider inherits it by construction:

- `read_only_state_db` now takes the full `&SealedHeader` instead of a bare hash. The header ties
  the pinned number to the pinned hash, so the floor comparison cannot be handed a mismatched
  pair. Pins below the floor are refused before any provider state is touched, as a new
  `RestoredStateFloorError` carried through `ProviderError::other`.
- The refusal stays classified node-local (a `Provider` fault): the gap is this node's datadir,
  not the chain, and a peer with full history answers the same read correctly. No attacker is
  required for the failure this guards against; the operator's own pinned read is the victim, and
  the floor turns silent empty state into a loud error.
- `pinned_state_and_env` drops its now-duplicate check and inherits the refusal through the
  shared constructor. Its `StateReadError::Provider` classification and message substring are
  unchanged (existing floor tests pass unmodified).
- Surveyed every `read_only_state_db` caller for a legitimate below-floor read (the issue's open
  question on option 1): `pinned_state_and_env`, `read_contract_inner`, and the test-only
  `tn_evm`. None reads below the floor legitimately. The tip-read path (`read_contract`) now also
  flows through the check; a restored datadir's canonical tip starts at the floor `B`, so tip
  reads are always admitted.

## Changes

- [x] `crates/tn-reth/src/env/mod.rs`: floor check moved into `read_only_state_db`; signature
  takes `&SealedHeader`; docs name it the ONE enforcement point
- [x] `crates/tn-reth/src/error.rs`: new `RestoredStateFloorError`, carried via
  `ProviderError::other`
- [x] `crates/tn-reth/src/env/epoch.rs`: duplicate check removed from `pinned_state_and_env`;
  floor docs redirected to the chokepoint
- [x] `crates/tn-reth/src/env/helpers.rs`: `read_contract_inner` passes the header through;
  `read_contract_at_block` documents the refusal
- [x] `crates/tn-reth/src/test_utils.rs`: `tn_evm` resolves a `SealedHeader` and passes it through
- [x] `crates/tn-reth/src/snapshot.rs`: doc nit from the issue (the floor marker is `B`, the
  window's final block, not `window[0].number`) plus the regression tests below

## Testing

- New regression coverage in the two existing floor tests (`snapshot.rs`):
  `read_contract_at_block` pinned below the floor is refused on a real restored datadir (genesis
  pin) and on a scaffold-only datadir (window-interior pin at a real, resolvable hash), with the
  refusal naming the floor; pinned AT the floor the same path is admitted (a codeless address
  answers empty output, the positive control).
- Both new assertions were confirmed by mutation: with the floor check removed from
  `read_only_state_db`, both tests fail at exactly the new `expect_err` lines; restored, all 127
  `tn-reth` lib tests pass.
- `cargo +1.94 check --all-targets` over buildplan's full reverse-dependency closure (21
  crates): green.
- `cargo +1.94 test -p tn-reth --lib`: 127 passed, 0 failed.
- `cargo +nightly-2026-03-20 fmt -- --check`: clean.
- `cargo +nightly-2026-03-20 clippy --workspace` (default and `--all-features`, `-D warnings`)
  and the attest-exact `cargo +1.94 test --no-run --workspace -j 2` in an isolated target dir:
  PENDING (update before push once gates.log shows ALL-GATES-PASS).

### DIFF
--- a/crates/tn-reth/src/env/epoch.rs
+++ b/crates/tn-reth/src/env/epoch.rs
@@ -766,28 +766,17 @@ impl RethEnv {
     /// no state and no history, and blocks below the window are zero-hash placeholders (see
     /// `SnapshotRestorer::import_chain_scaffold` in `snapshot.rs`). reth's checkpoint-less
     /// history walk answers reads anywhere below `B` with `Ok(None)` per account ("never
-    /// written") — no error, no log — so pins below the persisted floor (= `B`) are refused
-    /// HERE, before any state is built, as [`StateReadError::Provider`]: the gap is this node's
-    /// datadir, not the chain, and a peer with full history answers the same read correctly.
+    /// written"), with no error and no log, so pins below the persisted floor (= `B`) are
+    /// refused in `read_only_state_db` (`env/mod.rs`), the shared state constructor this helper
+    /// builds on, before any state is built. The refusal surfaces here as
+    /// [`StateReadError::Provider`]:
+    /// the gap is this node's datadir, not the chain, and a peer with full history answers the
+    /// same read correctly.
     fn pinned_state_and_env(
         &self,
         header: &SealedHeader,
     ) -> StateReadResult<(State<StateProviderDatabase<StateProviderBox>>, EvmEnv)> {
-        // refuse pins below the restored-state floor before touching the provider: below it the
-        // datadir holds headers but no state, and the read would resolve silently to "never
-        // written" values instead of failing
-        self.inner.restored_state_floor.filter(|floor| header.number < *floor).map_or(
-            Ok(()),
-            |floor| {
-                Err(StateReadError::Provider(format!(
-                    "pinned read at block {} is below this datadir's restored-state floor \
-                     (block {floor}, the snapshot's final block): the snapshot shipped no state \
-                     below it",
-                    header.number
-                )))
-            },
-        )?;
-        let db = self.read_only_state_db(header.hash()).map_err(|e| {
+        let db = self.read_only_state_db(header).map_err(|e| {
             StateReadError::Provider(format!("state provider at {}: {e}", header.hash()))
         })?;
         let evm_env = self.inner.evm_config.evm_env(header).map_err(|e| {
--- a/crates/tn-reth/src/env/helpers.rs
+++ b/crates/tn-reth/src/env/helpers.rs
@@ -317,7 +317,9 @@ impl RethEnv {
     /// # Errors
     ///
     /// Same split as [`Self::read_contract`]; additionally, `block_hash` failing to resolve to
-    /// a known sealed header is an [`EvmReadError::Internal`].
+    /// a known sealed header is an [`EvmReadError::Internal`], and so is a pin below a restored
+    /// datadir's state floor, refused by `read_only_state_db` (`env/mod.rs`) before any state
+    /// is read instead of silently answering from the missing pre-snapshot history (#1136).
     pub fn read_contract_at_block(
         &self,
         block_hash: B256,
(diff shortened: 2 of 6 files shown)


sstanfield (approved):
Looks good.

grantkee (approved):
Good catch

PR #1140, opened 2026-08-11
fix(network-libp2p): issue at most one in-flight kad query per missing authority

### LINKED ISSUE #1135: Duplicate kad queries for missing committee members
## Problem

`current_committee_rpcs` triggers discovery for unknown committee members on every call
(`crates/network-libp2p/src/peers/manager.rs:1021-1023`). The filter inside
`trigger_missing_authorities` checks `known_peers`, which records completed outcomes rather
than queries in flight (`manager.rs:787-793`), so a member whose record has not arrived yet is
re-reported missing every time.

Each `MissingAuthorities` event issues a fresh `get_record` and inserts a new entry into
`kad_record_queries`, keyed by query id (`crates/network-libp2p/src/consensus.rs:1726-1731`).
Entries are removed only when their own query completes (`consensus.rs:2141-2150`). A caller
polling at N calls per second with M unknown members therefore accumulates on the order of N×M
duplicate in-flight queries on the swarm loop, all resolving the same records.

The worst window is right after a committee rotation: `prune_known_peers` (`manager.rs:740`,
`manager.rs:762-767`) has just evicted rotated-out members, so M peaks exactly when callers
start polling for the new committee's endpoints.

## Solution

Track in-flight discovery per BLS key: a `HashSet<BlsPublicKey>` consulted in
`trigger_missing_authorities` (`manager.rs:788-789`) before reporting a member missing, with
entries cleared when the record arrives or the query completes.

One design decision to settle first: the filter lives in `PeerManager` but the query map lives
in `ConsensusNetwork`, so the set has to be owned by one side. Either the manager owns it and
clears entries via `add_discovered_peer` plus a completion signal for failed queries, or the
network dedupes against `kad_record_queries` values before issuing `get_record`.

MavenRain:
Closes #1135.

## Problem

`current_committee_rpcs` triggers discovery for unknown committee members on every call.  The filter inside `trigger_missing_authorities` checks `known_peers`, which records completed outcomes rather than queries in flight, so a member whose record has not arrived yet is re-reported missing on every call.

Each `MissingAuthorities` event issued a fresh `get_record` and inserted a new entry into `kad_record_queries`, keyed by query id.  Entries are removed only when their own query completes.  A caller polling at N calls per second with M unknown members therefore accumulated on the order of N x M duplicate in-flight queries on the swarm loop, all resolving the same records.

No attacker is required.  The worst window is right after a committee rotation: `prune_known_peers` has just evicted rotated-out members, so M peaks exactly when callers start polling for the new committee's endpoints.

## Fix

- [x] `crates/network-libp2p/src/consensus.rs`: the `PeerEvent::MissingAuthorities` arm skips any key that already has a query tracked in `kad_record_queries`, so each missing member has at most one live `get_record`.

The dedupe source is the query map itself, not an auxiliary set (the alternative raised in the issue).  The map is authoritative for in-flight state: every terminal query path (record found, invalid record, no record, query error) removes its entry via `close_kad_query`, so a skipped key becomes queryable again as soon as its current query ends.  The removal there runs before any result filtering, so even a query whose record is dropped as stale or non-committee re-arms the key.  There is no second collection to keep in sync across the manager/network boundary, and no starvation window longer than one query lifetime (bounded by the kad query timeout).  In-flight `get_record` queries are now bounded by the number of distinct unresolved keys (at most the tracked committees), independent of caller polling rate.

One deliberate behavior change: while a query is in flight, the per-key retry cadence becomes once per completed query (typically the kad peer timeout, with the query timeout as the ceiling) instead of once per caller poll.  This is the point of the fix, and the suppressed branch logs at `trace` level so the state is observable.  Members we actually connect to still push their records directly on first connect, so the slower re-sample cadence only affects members never connected to.

## Testing

- [x] New regression test `test_missing_authorities_dedupes_inflight_kad_queries` drives `process_peer_manager_event` directly on an un-spawned `ConsensusNetwork`.  The swarm is never polled, so no query can terminate mid-test and every observed entry is in flight by construction (deterministic, no timing assumptions).  It asserts: a first-time key issues exactly one query (positive control), a polling re-report issues none, a new key alongside a re-report still issues its own, steady-state polling does not accumulate entries, and (re-arm) once `close_kad_query` ends a key's query, the next report issues a fresh query with a new query id.
- [x] Test confirmed by mutation: reverting the guard makes it fail;  restoring makes it pass.
- [x] `cargo +nightly fmt -- --check`, `cargo +1.94 clippy --all-targets`, and `cargo +1.94 test --no-run --workspace -j 2` (isolated target dir) green on the pinned toolchain.
- [x] `cargo +1.94 test -p tn-network-libp2p` green.

### DIFF
--- a/crates/network-libp2p/src/consensus.rs
+++ b/crates/network-libp2p/src/consensus.rs
@@ -1724,10 +1724,24 @@ where
                 self.swarm.behaviour_mut().gossipsub.remove_blacklisted_peer(&peer_id);
             }
             PeerEvent::MissingAuthorities(missing) => {
+                // Polling callers such as `current_committee_rpcs` report a member as
+                // missing on every call until its record lands in `known_peers`, so the
+                // same key arrives here repeatedly while its lookup is still in flight.
+                // Issue at most one live `get_record` per key: skip keys already tracked
+                // in `kad_record_queries` (issue #1135). The map is safe as the dedupe
+                // source because every terminal query path removes its entry (see
+                // `close_kad_query`), so a skipped key becomes queryable again as soon
+                // as its current query ends. The removal there runs before any result
+                // filtering, so even a query whose record is dropped as stale or
+                // non-committee re-arms the key.
                 for bls_key in missing {
-                    let key = kad::RecordKey::new(&bls_key);
-                    let query_id = self.swarm.behaviour_mut().kademlia.get_record(key);
-                    self.kad_record_queries.insert(query_id, bls_key.into());
+                    if self.kad_record_queries.values().all(|q| q.request != bls_key) {
+                        let key = kad::RecordKey::new(&bls_key);
+                        let query_id = self.swarm.behaviour_mut().kademlia.get_record(key);
+                        self.kad_record_queries.insert(query_id, bls_key.into());
+                    } else {
+                        trace!(target: "network-kad", ?bls_key, "kad record query already in flight");
+                    }
                 }
             }
             PeerEvent::Discovery => {
--- a/crates/network-libp2p/src/tests/network_tests.rs
+++ b/crates/network-libp2p/src/tests/network_tests.rs
@@ -2543,6 +2543,70 @@ async fn test_node_record_validation() {
     assert!(network.peer_record_valid(&peer_record).is_none());
 }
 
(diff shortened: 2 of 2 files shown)


grantkee (approved):
LGTM!

sstanfield (approved):

PR #1141, opened 2026-08-11
fix(types): add one-shot ShutdownNotifier so a subscribe after notify resolves

### LINKED ISSUE #1134: Notifier::subscribe after notify never resolves, orphaning late-spawned tasks
## Problem

`Notifier` keeps its subscribers in a vec (`crates/types/src/notifier.rs:26-28`). `notify`
drains the vec and flags only the subscribers that existed at that moment
(`notifier.rs:45-59`). A `subscribe` that happens after `notify` (`notifier.rs:37-42`) starts
unflagged, and nothing ever flags it, so the returned future is pending forever
(`notifier.rs:76-85`).

For shutdown this is a real orphan path. The only place `local_shutdown.notify` runs is
`TaskManager::Drop` (`crates/types/src/task_manager.rs:143-147`), and `create_task` subscribes
at spawn time (`task_manager.rs:188`). A task spawned through a `TaskSpawner` whose manager has
already dropped therefore gets a shutdown future that never fires: the task runs untracked and
cannot be cancelled.

Stale spawners exist across epoch boundaries. The worker network handle's spawner is replaced
only at the next epoch start (`crates/node/src/manager/node/start_epoch.rs:445`), and the RPC
forwarder captures its spawner once at construction (`start_epoch.rs:497`) and never refreshes
it. Consumers that spawn through these handles: `crates/consensus/worker/src/worker.rs:253`,
`worker.rs:290`, and `crates/consensus/worker/src/network/mod.rs:211`, `mod.rs:239`,
`mod.rs:277`, `mod.rs:314`.

The code already works around the race in one place: the epoch runner subscribes early
(`crates/node/src/manager/node/run_epoch.rs:343-344`) and carries the noticer to the join, with
a comment explaining why (`run_epoch.rs:451-454`).

## Solution

Not a persistent flag on `Notifier` itself. The certifier re-arms its notifier per proposal:
`new_proposal` (`crates/consensus/primary/src/certifier.rs:110`) is notified for every incoming
header (`certifier.rs:543`) and re-subscribed on every proposal (`certifier.rs:275`,
`certifier.rs:293`, `certifier.rs:412`; the note at `certifier.rs:433` leans on re-arming). A
sticky flag would auto-cancel every proposal after the first header.

Instead, a one-shot variant (say `ShutdownNotifier`) whose flag persists after `notify`, so a
late subscribe resolves immediately. Adopt it where the semantics are genuinely one-shot:

- `TaskManager::local_shutdown` (`task_manager.rs:138`)
- the consensus config's shutdown (`crates/config/src/consensus.rs:160`)
- the certificate fetcher's per-fetch cancel signal
  (`crates/consensus/primary/src/certificate_fetcher.rs:463`)

`Notifier` stays as-is for the certifier's re-armable use.

## Notes

Two small riders in `task_manager.rs` worth folding in:

- The `try_send` failure log in `create_task` (`task_manager.rs:199-202`) reads as a
  bookkeeping error. The consequence is worse than it sounds: the task was already spawned, so
  it keeps running untracked and will never be joined or aborted. The log should say that.
- Two doc comments still mention "submanagers" (`task_manager.rs:309-311`,
  `task_manager.rs:412-414`), but the struct has no submanager field
  (`task_manager.rs:131-141`).

## Suggested test

For the new type: subscribe after notify and assert the future resolves. For `Notifier`, pin
the re-arm behavior the certifier depends on: subscribe, notify, subscribe again, notify again,
and assert both subscriptions resolved.

MavenRain:
Closes #1134.

## Problem

`Notifier` keeps its subscribers in a vec, and `notify` flags only the subscribers that exist at that moment. A `subscribe` that happens after `notify` starts unflagged, and nothing ever flags it, so the returned future is pending forever.

For shutdown this is a real orphan path. The only place `local_shutdown.notify` runs is `TaskManager::Drop`, and `create_task` subscribes at spawn time. A task spawned through a `TaskSpawner` whose manager has already dropped gets a shutdown future that never fires: the task runs untracked and cannot be cancelled. Stale spawners exist across epoch boundaries (the worker network handle's spawner is replaced only at the next epoch start; the RPC forwarder captures its spawner once at construction and never refreshes it), so no attacker is required . . . an ordinary epoch transition can leak such tasks.

## Fix

- [x] `crates/types/src/notifier.rs`: add `ShutdownNotifier`, a one-shot variant whose notified state is sticky. A `subscribe` after `notify` returns a `Noticer` that resolves immediately. `Notifier` keeps its re-arming semantics for the certifier's per-proposal cancel (`new_proposal`), which is re-subscribed on every proposal and must not auto-cancel after the first header.
- [x] Adopt `ShutdownNotifier` where the semantics are genuinely one-shot: `TaskManager::local_shutdown` (and its `TaskSpawner` clone), the consensus config's shutdown, the certificate fetcher's per-fetch cancel signal, the epoch manager's `node_shutdown`, and the worker gateway's shutdown. The `TaskManager` join family (`join`, `join_until_exit`, `until_exit`, `until_task_ends`) now takes `ShutdownNotifier`: `until_exit_internal` notifies the passed signal on exit, so every signal that flows through it is one-shot by contract. Per-epoch safety holds because each epoch constructs a fresh `ConsensusConfig` (and so a fresh notifier); stickiness cannot leak across epochs.
- [x] `crates/types/src/task_manager.rs`: the `try_send` failure log in `create_task` now states the real consequence (the task is already spawned but untracked: it can never be joined or aborted, and only its shutdown notice can end it). The two doc comments that referenced nonexistent submanagers (`set_join_wait`, `abort_all_tasks`) are corrected.
- [x] Test infrastructure adapted to the one-shot contract. `committed_round_after_restart` reused one fixture config across six restart iterations and re-armed it with `notify_shutdown`; it now builds a fresh per-iteration config over the shared store, the same pattern `restart_with_new_committee` already uses. `test_equivocation_protection_after_restart` restarted its proposer against the same latched config, which made the second proposer exit at startup and the test hang at the digest ack; its restart phase now builds a fresh config over the same store, so the equivocation guard still sees the first proposal. `test-utils` `default_test_execution_node` and the `tn-reth` forwarder test fixture passed a temporary `TaskManager` whose immediate drop now latches shutdown; both leak the manager instead so tasks spawned through the retained spawner still run.

## Testing

- New `test_shutdown_notifier_late_subscribe_resolves`: an early subscriber parks on its `Noticer` in a spawned task (covering the waker path), then a subscribe after `notify` resolves immediately. Confirmed by mutation twice: with the sticky-flag seed reverted the test fails on `late.noticed()`, and with the wake call deleted it fails on the parked subscriber timeout.
- New `test_notifier_rearms_after_notify`: pins the re-arm behavior the certifier depends on. Subscribe, notify, subscribe again, notify again; both subscriptions resolve, the second one only on the second notify.
- `cargo +1.94 check --all-targets` green over the full 25-crate reverse-dependency closure; `cargo +1.94 test -p tn-types` (notifier and task manager suites) and `cargo +1.94 test -p tn-primary` (full proposer suite plus `committed_round_after_restart`, `restart_with_new_committee`) green. The proposer restart hang was reproduced against the unadapted test and resolves with the fresh-config restart.
- `make attest` gates on the pinned toolchain: nightly fmt check, nightly `clippy --workspace` (default and `--all-features`), and `cargo +1.94 test --no-run --workspace` in an isolated target dir.

### DIFF
--- a/bin/worker-gateway/src/app.rs
+++ b/bin/worker-gateway/src/app.rs
@@ -4,7 +4,7 @@
 use std::sync::Arc;
 
 use reqwest::Client;
-use tn_types::{Notifier, TaskManager};
+use tn_types::{ShutdownNotifier, TaskManager};
 use tracing::info;
 
 use crate::{
@@ -81,7 +81,7 @@ pub(crate) async fn run(settings: Settings) -> eyre::Result<()> {
             .unwrap_or(u64::MAX),
     );
     let spawner = task_manager.get_spawner();
-    let shutdown = Notifier::new();
+    let shutdown = ShutdownNotifier::new();
 
     // Optional Prometheus scrape endpoint on its own listener. Reuses the node's
     // `tn-metrics` recorder + server so the gateway's `tn_worker_gateway_*` series
--- a/crates/config/src/consensus.rs
+++ b/crates/config/src/consensus.rs
@@ -9,7 +9,7 @@ use std::{
 use tn_network_types::local::LocalNetwork;
 use tn_types::{
     Authority, AuthorityIdentifier, BlsPublicKey, Certificate, Committee, Database, Epoch,
-    EpochDigest, Hash as _, HeaderDigest, Multiaddr, NetworkPublicKey, Notifier,
+    EpochDigest, Hash as _, HeaderDigest, Multiaddr, NetworkPublicKey, ShutdownNotifier,
 };
 use tracing::info;
 
@@ -43,7 +43,7 @@ struct ConsensusConfigInner<DB> {
 #[derive(Debug, Clone)]
 pub struct ConsensusConfig<DB> {
     inner: Arc<ConsensusConfigInner<DB>>,
-    shutdown: Notifier,
+    shutdown: ShutdownNotifier,
 }
 
 impl<DB> ConsensusConfig<DB>
@@ -194,7 +194,7 @@ where
         let primary_public_key = key_config.primary_public_key();
         let authority = committee.authority_by_key(&primary_public_key);
 
-        let shutdown = Notifier::new();
+        let shutdown = ShutdownNotifier::new();
         let genesis = Certificate::genesis(&committee)
             .into_iter()
             .map(|cert| (cert.digest(), cert))
@@ -221,7 +221,7 @@ where
     ///
     /// The shutdown notifier can be used to either subscribe to shutdown events
     /// or trigger shutdown across consensus components.
-    pub fn shutdown(&self) -> &Notifier {
+    pub fn shutdown(&self) -> &ShutdownNotifier {
         &self.shutdown
     }
 
--- a/crates/consensus/primary/src/certificate_fetcher.rs
+++ b/crates/consensus/primary/src/certificate_fetcher.rs
@@ -21,7 +21,7 @@ use tn_network_libp2p::types::NetworkResult;
 use tn_storage::CertificateStore;
 use tn_types::{
     validate_fetched_certificate, AuthorityIdentifier, BlsPublicKey, Certificate, Committee,
-    Database, Header, Noticer, Notifier, Round, TaskManager, TaskSpawner, TnReceiver,
+    Database, Header, Noticer, Round, ShutdownNotifier, TaskManager, TaskSpawner, TnReceiver,
 };
 use tokio::{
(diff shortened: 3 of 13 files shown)


sstanfield (approved):
Looks like a solid fix.

PR #1142, opened 2026-08-11
Added some comparative benchmarks for cache DB backends.

sstanfield:
Added some DB benches to compare backends.  Should run it with a release build.  Can run it with ```cargo test --release -p tn-storage db_backend_comparison -- --ignored --nocapture --test-threads 1```

@MavenRain Note that commenting out (since this is a test MDBX will have sync mode off) does not seem to effect the MDBX times:
        {
            use reth_libmdbx::{EnvironmentFlags, Mode};
                mode: Mode::ReadWrite { sync_mode: BUILD_SYNC_MODE },
                ..Default::default()
            });
        }
We may not need to do this?  Also, not sure how this would not matter or even how MDBX seems to "persist" durably as fast it does...  Still some questions to answer here.

The layered DB does still seem to help for most benches and costs almost nothing in the few it does not.  So probably worth keeping for now.

### DIFF
--- a/crates/storage/src/db_bench.rs
+++ b/crates/storage/src/db_bench.rs
@@ -1,0 +1,456 @@
+//! Comparative benchmark harness for the consensus [`Database`] backends.
+//!
+//! [`crate::test::db_simp_bench`] times a single backend on one big 50k insert/commit. That hides
+//! what dominates real node DB cost: **many small, frequently-committed, durable transactions** on
+//! the consensus hot path (per-certificate ~3 inserts, externalized under a durability barrier),
+//! plus large epoch-boundary `clear_table`s. This module runs a *list* of backends through a
+//! battery of **basic** and **representative-of-real-usage** benchmarks and prints one easy-to-read
+//! side-by-side comparison, so the current DB implementation's performance can be examined.
+//!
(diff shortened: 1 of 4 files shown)


crates/storage/src/db_bench.rs
```diff
@@ -0,0 +1,456 @@
...
+    // Basic ops share `BenchEpoch` state in sequence, like `db_simp_bench`.
+    vec![
+        ("bulk_insert 50k", bench_bulk_insert(rt, db)),
+        ("iter_forward 50k", bench_iter_forward(db)),
+        ("iter_reverse 50k", bench_iter_reverse(db)),
+        ("point_get_txn 50k", bench_point_get_txn(db)),
+        ("clear_table 50k", bench_clear_table(rt, db)),
+        // Representative node-usage patterns (each resets its own table first).
+        ("small_commits 500 x3", bench_small_commits(rt, db)),
+        ("durable_commits 50 x3", bench_durable_commits(rt, db)),
```
MavenRain, line 294: Should be 500 x3, I believe

MavenRain (approved):
Just one nit

grantkee (approved):

PR #1143, opened 2026-08-12
fix(storage): report a queued save failure from CertificatePack flush replies (#1138)

### LINKED ISSUE #1138: CertificatePack: shutdown and persist return Ok after a failed background save (same shape as #1065)
All line references are against `main` at `77a0c1fe`.  File: `crates/storage/src/certificate_pack.rs` unless stated.

## Summary

A certificate save runs on a background thread.  When the save fails, the error latches into a watch slot, and no flush reply ever carries it.  The end-of-epoch path `shutdown()` never consults the slot: it replies with the result of `inner.persist()` alone.  So the final flush reports `Ok(())` while a certificate is missing from the pack, and the one error log the caller checks can never fire for that failure.

This is the same sample-before-enqueue shape that #1065 reported for `EpochRecordDb` and that #1094 (commit `901544f3`) fixed there.  `CertificatePack` still has it.

## Mechanism

1. The `Save` arm (`:61-65`) is fire-and-forget.  On failure it latches the error: `tx_error.send_replace(Some(e))` (`:63`).
2. `get_error` (`:134-139`) borrows the slot and does not clear it.  The error stays latched for later calls.
3. The write-side entry points sample the slot before they enqueue: `save` (`:144`), `try_save` (`:156`), `persist` (`:189`).
4. `persist` (`:188-196`) samples at `:189`, then enqueues `Persist` (`:191`).  The `Persist` arm (`:72-74`) replies `let _ = tx.send(inner.persist());` and never reads the watch.  An error that lands between the sample and the reply is not in the reply.
5. `shutdown` (`:200-224`) takes no sample at all.  It sends `ShutdownAsync` (`:205`) and propagates only the flush result (`:206`).  The `ShutdownAsync` arm (`:80-83`) also replies `inner.persist()` without reading the watch.
6. The `Shutdown` arm (`:75-79`, reached from `Drop` via `:96`) discards even the flush result.
7. The actor drains one FIFO `mpsc::channel(1000)` (`:105`, `:120`) with a single `blocking_recv` consumer (`:59`).  A `Save` queued ahead of the shutdown message always processes first.  Its failure is therefore already latched when the flush replies `Ok`.  This is deterministic ordering, not an unlucky schedule.

## Reachability

- `try_save` caller: `crates/consensus/primary/src/consensus/state.rs:474`.  It logs the error and continues.  The comment at `:476-479` treats the pack as non-critical.
- `shutdown` caller: `state.rs:443-448`.  It logs a shutdown error (`:446`).  Today that log cannot fire for a latched save failure, because the reply never carries it.
- The gap is sharpest for the last save of an epoch.  When that save fails, no later `try_save` samples the slot, `shutdown()` returns `Ok(())`, and the failure is reported nowhere.
- `persist` has no production caller today (only the module test at `:408`).  It is public API, so any future caller inherits the `persist`-side window in point 4.

## Impact

Modest, and scoped to reporting.  The pack is a per-epoch certificate cache (opened at `crates/node/src/manager/node/run_epoch.rs:417`), and the caller already treats it as non-critical.  No consensus data is lost.  The defect makes a pack that silently dropped a certificate read as cleanly shut down.  Consumers later see the drop only as a plain lookup miss: `get` returns `None`, `contains` returns `false`, with no earlier signal that a write failed.

## Suggested direction

#1094 fixed the same shape in `EpochRecordDb` by folding the pending watch error into the flush reply.  The same pattern fits the `Persist` and `ShutdownAsync` arms here.  This watch does not clear on read, so the arms only need to read the slot when they build the reply.  I am glad to send a PR that mirrors #1094 if that is welcome.

## Note

The `Drop`-path `Shutdown` arm discards the flush result (`:76`).  That may be intentional for `Drop`.  I mention it for completeness;  the suggested change above does not touch it.

MavenRain:
Closes #1138.

## Problem

A `CertificatePack` save runs fire-and-forget on a background thread. When a save fails, the error latches into a watch slot that `get_error()` reads without clearing, and the write-side entry points sample that slot only before they enqueue. The `Persist` and `ShutdownAsync` actor arms replied with `inner.persist()` alone and never read the slot, and `shutdown()` takes no sample at all. So a save failure that was still queued at sample time (or that landed after the last sample of the epoch) never reached a flush reply: the end-of-epoch path returned `Ok(())` for a pack that silently dropped a certificate, and the shutdown error log the caller checks (`consensus/state.rs:446`) could never fire for that failure.

No attacker is required and no unlucky schedule is needed. The actor drains a single FIFO `mpsc` channel with one `blocking_recv` consumer, so a `Save` queued ahead of the flush message always fails, and latches, before the flush replies. The consequence is scoped to reporting: consumers later see the dropped certificate only as a plain lookup miss (`get` returns `None`, `contains` returns `false`) with no earlier signal that a write failed.

This is the same sample-before-enqueue shape #1065 reported for `EpochRecordDb`, fixed there by #1094 (commit 901544f3). `CertificatePack` still had it.

## Fix

- [x] `crates/storage/src/certificate_pack.rs`: the `Persist` and `ShutdownAsync` arms now read the latched error slot and fold it into their reply (`pending.map_or(flushed, Err)`), mirroring the #1094 `Persist` arm. The flush itself still runs; the only behavioural change is that a previously silent failure now reaches the caller.
- [x] Unlike `EpochRecordDb`, this watch does not clear on read, so the arms read without consuming and the failure stays visible to `get_error()` afterwards.
- [x] When the flush itself also fails after an earlier failed save (a write error moves the pack to its `failed` state, which refuses the commit), the reply carries the earlier save error, the root cause, matching the #1094 fold precedence.
- [x] Doc comments on the struct, `persist`, and `shutdown` now state the durability-barrier contract explicitly, including the pre-existing short-circuit: an error already latched at call time returns from `persist()` before a flush is enqueued.
- [x] The `Drop`-path `Shutdown` arm intentionally stays untouched, as noted in the issue.

## Testing

Two regression tests inject a deterministic save failure by reopening an existing pack read-only, so the queued `Save` fails in the actor with `AppendError::ReadOnly` (`Inner::persist` on a read-only pack skips the commit and returns `Ok`, so the reply error can only come from the fold). Each test queues the failing `Save` directly on the channel ahead of the flush message; single-consumer FIFO ordering makes the interleaving deterministic rather than a race.

- [x] `persist_reports_a_save_that_failed_while_the_flush_was_queued`: the `Persist` reply carries the queued failure, and the slot stays latched for `get_error()` afterwards.
- [x] `shutdown_reports_a_save_that_failed_while_the_shutdown_was_queued`: `shutdown()` returns the failure on the end-of-epoch path that previously reported `Ok(())`.
- [x] Both tests confirmed by mutation: reverting the two arm folds makes both fail; restoring makes them pass.
- [x] `cargo +1.94 test -p tn-storage` green.
- [x] `cargo +nightly fmt -- --check`, `cargo +1.94 clippy --all-targets`, and `cargo +1.94 test --no-run --workspace -j 2` (isolated target dir) green.

### DIFF
--- a/crates/storage/src/certificate_pack.rs
+++ b/crates/storage/src/certificate_pack.rs
@@ -35,6 +35,11 @@ enum PackMessage {
 
 /// Manages a pack file of [`Certificate`] data, indexed by certificate digest.
 /// Note, lack of Clone to make shutdown easier to manage.
+///
+/// Saves run on a background thread. A failed save latches into an error slot that
+/// [`get_error`](Self::get_error) reads without clearing. [`persist`](Self::persist) and
+/// [`shutdown`](Self::shutdown) are the durability barriers: a latched save failure reaches
+/// their return value, even when that save was still queued at the time of the call.
 #[derive(Debug)]
 pub struct CertificatePack {
     tx: Sender<PackMessage>,
@@ -70,7 +75,14 @@ fn run_pack_loop(
                 let _ = tx.send(inner.contains(digest));
             }
             PackMessage::Persist(tx) => {
-                let _ = tx.send(inner.persist());
+                // Fold a save that failed while this persist was queued into the reply.
+                // `persist()` samples the error slot before enqueueing, and saves are
+                // fire-and-forget, so a save that fails after that sample but before this arm
+                // would otherwise be acknowledged as a successful flush. The slot does not
+                // clear on read, so a plain read is enough here.
+                let pending = tx_error.borrow().clone();
+                let flushed = inner.persist();
+                let _ = tx.send(pending.map_or(flushed, Err));
             }
             PackMessage::Shutdown => {
                 let _ = inner.persist();
@@ -78,7 +90,13 @@ fn run_pack_loop(
                 break;
             }
             PackMessage::ShutdownAsync(tx) => {
-                let _ = tx.send(inner.persist());
+                // Same fold as `Persist`: `shutdown()` never samples the error slot, so this
+                // reply is the only place a save that failed earlier in the queue can surface.
+                // Without it the final flush of an epoch reports `Ok(())` for a pack that
+                // silently dropped a certificate.
+                let pending = tx_error.borrow().clone();
+                let flushed = inner.persist();
+                let _ = tx.send(pending.map_or(flushed, Err));
                 break;
             }
         }
@@ -185,6 +203,12 @@ impl CertificatePack {
     }
 
     /// Flush the pack file and index to disk.
+    ///
+    /// Returns `Err` if any background save queued before this call failed. An error that is
+    /// already latched when this is called short-circuits before the flush is enqueued (the
+    /// slot does not clear on read). A save still queued at that sample is processed before
+    /// the flush by the actor's single FIFO channel, and its failure is folded into the flush
+    /// reply.
     pub async fn persist(&self) -> Result<(), PackError> {
         self.get_error()?;
         let (tx, rx) = oneshot::channel();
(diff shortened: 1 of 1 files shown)


grantkee (approved):
LGTM!

sstanfield (approved):
Nice catch.

PR #1144, opened 2026-08-12
fix(worker): keep observer txs pooled when a batch forward is refused

### LINKED ISSUE #1132: Dropped forward txs are evicted from the observer's own pool and reported as sealed
**Location**: `crates/consensus/worker/src/worker.rs:253` → `crates/batch-builder/src/lib.rs:326`

On an observer, when a batch fails to forward for *any* reason, its transactions are removed from
the observer's own transaction pool anyway, and the node emits a success metric. The chain:

1. `disburse_txns` returns `Ok(())` at `worker.rs:276`, immediately after `spawn_task` at
   `worker.rs:253`. Nothing is awaited — the return value is fully decoupled from forward success.
2. `seal` early-returns that `Ok(())` when `quorum_waiter` is `None` (`worker.rs:282-285`).
3. The batch builder takes its success branch: increments `tn_batch_builder.batches_sealed_total`
   (`crates/batch-builder/src/lib.rs:184`) and sends
   `MinedBatchResult { mined_transactions, changed_accounts }` (`lib.rs:186`), where
   `mined_transactions` is every hash in the batch (`crates/batch-builder/src/batch.rs:112`).
4. Because `mined_transactions` is non-empty, the run loop falls past the guard at `lib.rs:309-316`
   and calls `pool.update_canonical_state(..)` (`lib.rs:326-332`).
5. That builds a `CanonicalStateUpdate` with `update_kind: PoolUpdateKind::Commit` carrying the
   hashes (`crates/tn-reth/src/txn_pool.rs:180-192`) and hands it to reth's
   `on_canonical_state_change`, which prunes each hash from the pool entirely.

**No recovery exists for observers.** `OurNodeBatchesCache`'s only writer is `worker.rs:294`, which
sits *after* the observer's early return at `worker.rs:284` — so `orphan_batches`
(`crates/node/src/manager/node/close_epoch.rs:117-163`) has nothing to re-inject.

**The validator path is correct by contrast.** `seal` returning `Err` maps to
`mined_transactions: vec![]` (`lib.rs:207`) with the explicit in-code note *"this will apply no
changes to transaction pool"*; the run loop then skips `update_canonical_state` at `lib.rs:309-316`
and the transactions stay pending.

**The intended contract exists but is unreachable.** `BlockSealError::NotValidator` is defined
(`crates/types/src/error.rs:167`), is handled as a keep-the-transactions outcome (`lib.rs:202`),
and has a metrics label (`crates/batch-builder/src/metrics.rs:42`) — but a workspace-wide search
returns exactly those three references. **It is never constructed anywhere.** The contract was
written, then bypassed.

**Exposure is the default configuration.** `worker.rpc` is `Option<RpcInfo>` with
`#[serde(default)]` (`crates/types/src/committee.rs:98`), so it is unset unless an operator sets
it. The repo states the consequence outright at `crates/e2e-tests/tests/it/common.rs:433-437`:
*"The genesis ceremony leaves `p2p_info.worker.rpc` unset ... with none advertised,
observer-submitted transactions are dropped."*

**Secondary amplification**: `changed_accounts` carries `nonce: max_nonce + 1`
(`batch.rs:132-139`), so after a dropped forward the sender's next transaction parks as
nonce-gapped and is never batched either.

**What this PR changes.** The eviction mechanism predates the commit. This PR *widens* it by adding
two new drop circumstances: capacity shed (`forward.rs:334-343`, a path that did not exist before)
and budget abandonment (`forward.rs:286-297`, replacing an eventual-completion path). Both are
correct as memory bounds and the author's rationale is sound — *"shedding at the door is strictly
better than shedding by OOM"* (`forward.rs:73-77`). What was not accounted for is that shed and
abandoned transactions are **also evicted from the local pool**, so they are permanently lost
rather than retried on the next build. That trade may well still be the right one; it should be
deliberate rather than unnoticed.

**Proposed fix** — make *admission* synchronous and reuse the already-wired `NotValidator` path.
Same edit as finding 1's Option A; implement once.

```diff
--- a/crates/types/src/worker/sealed_batch.rs
+++ b/crates/types/src/worker/sealed_batch.rs
 pub trait TxnForwarder: Send + Sync + Debug {
+    /// Returns `true` if the batch was admitted to a forward task. `false` means it was dropped
+    /// at the door and the caller still owns these transactions. Admission is not delivery.
     fn forward_txns(
         &self,
         transactions: Vec<Vec<u8>>,
         committee_slots: Vec<BlsPublicKey>,
         validator_rpcs: Vec<(BlsPublicKey, RpcInfo)>,
-    );
+    ) -> bool;
 }
```

```diff
--- a/crates/consensus/worker/src/worker.rs
+++ b/crates/consensus/worker/src/worker.rs
-        self.network_handle.get_task_spawner().spawn_task("disburse-txns", async move { … });
-        Ok(())
+        let admitted = match self.network_handle.get_all_validator_rpcs().await {
+            Ok(rpcs) if !rpcs.is_empty() => {
+                self.forwarder.forward_txns(transactions, self.committee_slots.clone(), rpcs)
+            }
+            Ok(_) => { warn!(target: "worker::batch_provider",
+                "no committee validator has advertised a JSON-RPC endpoint"); false }
+            Err(err) => { warn!(target: "worker::batch_provider", ?err,
+                "failed to discover validator JSON-RPC endpoints"); false }
+        };
+        // A batch that could not even be handed to a forward task must leave its transactions
+        // in the pool: reporting `Ok` here is what evicts them via `update_canonical_state`.
+        if admitted { Ok(()) } else { Err(BlockSealError::NotValidator) }
```

Re-forwarding later is safe: the forwarder classifies `"already known"` as `Delivered`
(`forward.rs:423-425`). Regression test: mirror `test_all_possible_error_outcomes`
(`lib.rs:567-732`) with `quorum_waiter: None` and a forwarder that admits nothing; assert
`pool_size().pending` is unchanged.

MavenRain:
Closes #1132.

## Problem

An observer worker does not wait for its transaction forward to start.  `disburse_txns` spawned a background task and returned `Ok(())` at once (`crates/consensus/worker/src/worker.rs`).  `seal` returned that `Ok(())` when `quorum_waiter` is `None`.  The batch builder then treated every observer batch as sealed:  it incremented `batches_sealed_total` and reported every hash in the batch as mined.  The pool update pruned those hashes from the observer's own pool (`crates/batch-builder/src/lib.rs`, `crates/tn-reth/src/txn_pool.rs`).

When the forward is dropped for any reason, the transactions are lost.  No recovery path exists for observers:  the `OurNodeBatchesCache` write sits after the observer's early return, so `orphan_batches` has nothing to re-inject.  The default configuration hits this.  `worker.rpc` is `#[serde(default)]` and the genesis ceremony leaves it unset, so no endpoint is advertised and every observer forward is dropped.  A second effect follows:  `changed_accounts` advances the sender's tracked nonce, so the sender's next transaction parks as nonce-gapped and is never batched either.

## Root cause

The intended contract exists but was unreachable.  `BlockSealError::NotValidator` is defined, the batch builder handles it as a keep-the-transactions outcome, and it has a metrics label.  No code constructed it.  Admission was decided inside a spawned task, after `seal` had already reported success, so the success report and the forward outcome were fully decoupled.

## Fix

Make admission synchronous and construct `NotValidator` on refusal.  Delivery stays best-effort on a background task, so batch production is never stalled by a slow or unreachable validator.

- [x] `crates/types/src/worker/sealed_batch.rs`: `TxnForwarder::forward_txns` now returns `bool`.  `true` means the batch was admitted to a forward task.  `false` means it was dropped at the door and the caller still owns the transactions.  Admission is not delivery.  `NoopTxnForwarder` returns `false`.
- [x] `crates/consensus/worker/src/worker.rs`: `disburse_txns` awaits endpoint discovery inline.  The discovery call returns a snapshot of already-fetched kademlia records, so it is not a blocking network round-trip.  When discovery fails, when no endpoint is advertised, or when the forwarder refuses the batch, `disburse_txns` returns `Err(BlockSealError::NotValidator)`.
- [x] `crates/tn-reth/src/forward.rs`: `WorkerRpcForwarder::forward_txns` returns `false` on every refuse path (nothing to route, no usable provider after the target policy runs, or no free forward permit) and `true` once the forward task is spawned.

The batch builder already maps `NotValidator` to an empty mined set, which skips `update_canonical_state`.  The transactions stay pending and the next build retries them.

Threat model notes.  The failure needs no attacker:  the default genesis configuration alone drops every observer forward today.  The retry is bounded by the builder's own backoff (`max_batch_delay`, default 1s), so a refused observer re-mines about once per second and does not spin.  Re-forwarding a transaction that did reach a validator is safe:  the forwarder classifies an `already known` response as delivered.  Committee rotation windows self-correct:  seals return `NotValidator` while discovery converges, and the transactions wait in the pool instead of vanishing.  The validator seal path is untouched:  with a quorum waiter present, `seal` never reaches `disburse_txns`.

This branch sits on the forward hardening that landed on main (per-batch budget, a node-wide cap on live forward tasks, forwarder metrics).  The two admission gates compose:  a batch that finds every forward permit taken is refused with `NotValidator` instead of being dropped, so capacity shedding becomes lossless backpressure and the transactions stay pooled for the next build.

Residuals kept out of scope on purpose (tracked in a follow-up issue):  admission checks that a usable endpoint is advertised, not that it answers, so an advertised but dead endpoint still admits (and can still lose) a batch;  the epoch-close orphan-batch path (`close_epoch.rs`) discards the new refusal signal, a pre-existing loss path this PR makes observable but does not close;  and a permanently refused observer re-mines at ~1 Hz with an error log and a seal-failure metric increment per tick, so the follow-up proposes a refusal-specific backoff.

## Testing

- [x] New regression tests, each confirmed by mutation (fix arm reverted, test seen to FAIL, fix restored, test green):
  - `crates/consensus/worker/tests/it/batch_provider_tests.rs`:  an observer seal (`quorum_waiter: None`) with no admitted forward returns `Err(BlockSealError::NotValidator)`;  an empty batch still returns `Ok(())`;  no `NodeBatchesCache` entry is written.  Fails when the admission gate in `disburse_txns` is neutralized.
  - `crates/batch-builder/src/lib.rs`:  an observer builder whose every seal is refused keeps all pending transactions in the pool and keeps running (`NotValidator` is non-fatal).  Fails under the same gate mutation.
  - `crates/tn-reth/src/forward.rs`:  `forward_txns` returns `false` when no endpoint is usable and `true` when a provider resolves.  The refusal test fails when the no-provider branch is flipped to `true`.
- [x] `cargo +nightly fmt -- --check`
- [x] `cargo +1.94 check` and `cargo +1.94 clippy --all-targets -- -D warnings` over the touched crates
- [x] `cargo +1.94 test -p tn-types -p tn-reth -p tn-worker -p tn-batch-builder` (one crate at a time;  all green)
- [x] Attest proxy:  `cargo +1.94 test --no-run --workspace -j 2` in an isolated target dir
- [x] `cargo +nightly clippy --workspace --all-features -- -D warnings`

### DIFF
--- a/crates/batch-builder/src/lib.rs
+++ b/crates/batch-builder/src/lib.rs
@@ -514,6 +514,142 @@ mod tests {
         assert_eq!(pending_pool_len, 3);
     }
 
+    /// An observer worker (no quorum waiter) whose forwarder admits nothing must keep its
+    /// transactions pending: every seal attempt returns `NotValidator`, the pool is not
+    /// drained, and the builder task stays alive to retry on a later build.
+    ///
+    /// The worker's batch channel is interposed so each seal attempt is observable: the
+    /// test receives each sealed batch, runs the real observer `seal` (which runs
+    /// `disburse_txns`), and forwards the result on the ack channel. Two full cycles prove
+    /// the builder processed the first refusal as non-fatal and kept looping.
+    #[tokio::test]
+    async fn test_observer_refused_forward_keeps_txs_in_pool() {
+        let tmp_dir = TempDir::new().unwrap();
+        let TestTools { mut tx_factory, execution_components, task_manager } =
+            get_test_tools(tmp_dir.path());
+        let TestExecutionComponents { reth_env, txpool, chain, .. } = execution_components;
+        let address = Address::from(U160::from(33));
+        let client = LocalNetwork::new_with_empty_id();
+        let worker_to_primary = Arc::new(MockWorkerToPrimaryHang {});
+        client.set_worker_to_primary_local_handler(worker_to_primary);
+        let temp_dir = TempDir::new().unwrap();
+        let store = open_db(temp_dir.path());
+        let seal_timeout = Duration::from_secs(5);
+        let (to_worker, mut from_batch_builder) = tokio::sync::mpsc::channel(2);
+
+        // The real observer worker: no quorum waiter, a forwarder that admits nothing, and
+        // a test network handle that discovers no validator RPC endpoints.
+        let block_provider = Worker::new(
+            0,
+            None::<TestMakeBlockQuorumWaiter>,
+            client,
+            store,
+            seal_timeout,
+            WorkerNetworkHandle::new_for_test(task_manager.get_spawner()),
+            Arc::new(NoopTxnForwarder),
+            Vec::new(),
+        );
+
+        // build execution block proposer with the interposed channel and a short delay so
+        // the retry after a refused seal arrives quickly
+        let batch_builder = BatchBuilder::new(
+            &reth_env,
+            txpool.clone(),
+            to_worker,
+            address,
+            Duration::from_millis(100),
+            task_manager.get_spawner(),
+            0,
+            MIN_PROTOCOL_BASE_FEE,
+            0,
+        )
+        .expect("batch builder");
+
+        let gas_price = reth_env.get_gas_price().unwrap();
+        let value = U256::from(10).checked_pow(U256::from(18)).expect("1e18 doesn't overflow U256");
+
+        // create 3 transactions
+        let transaction1 = tx_factory.create_eip1559(
+            chain.clone(),
+            None,
+            gas_price,
+            Some(Address::ZERO),
+            value, // 1 TEL
+            Bytes::new(),
+        );
+
+        let transaction2 = tx_factory.create_eip1559(
+            chain.clone(),
+            None,
+            gas_price,
+            Some(Address::ZERO),
+            value, // 1 TEL
+            Bytes::new(),
+        );
+
+        let transaction3 = tx_factory.create_eip1559(
+            chain.clone(),
+            None,
+            gas_price,
+            Some(Address::ZERO),
+            value, // 1 TEL
+            Bytes::new(),
+        );
+
+        let added_result = tx_factory.submit_tx_to_pool(transaction1.clone(), txpool.clone()).await;
+        assert_matches!(added_result, hash if &hash == transaction1.hash());
+
+        let added_result = tx_factory.submit_tx_to_pool(transaction2.clone(), txpool.clone()).await;
+        assert_matches!(added_result, hash if &hash == transaction2.hash());
+
+        let added_result = tx_factory.submit_tx_to_pool(transaction3.clone(), txpool.clone()).await;
+        assert_matches!(added_result, hash if &hash == transaction3.hash());
+
+        // txpool size
+        let pending_pool_len = txpool.pool_size().pending;
+        assert_eq!(pending_pool_len, 3);
+
+        // spawn batch_builder once worker is ready
+        let batch_builder_task = tokio::spawn(batch_builder.run());
+
+        // plenty of time for each build attempt
+        let duration = std::time::Duration::from_secs(5);
+
+        // First attempt: the builder proposes all 3 transactions and the real observer seal
+        // refuses the batch because it was not admitted to a forward task.
+        let (sealed_batch, ack) = timeout(duration, from_batch_builder.recv())
+            .await
+            .expect("first batch built in time")
+            .expect("batch builder sender open");
+        assert_eq!(sealed_batch.batch().transactions().len(), 3);
+        let res = block_provider.seal(sealed_batch).await;
+        assert!(matches!(res, Err(BlockSealError::NotValidator)));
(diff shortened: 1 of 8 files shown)


grantkee (approved):
Good work @MavenRain LGTM

sstanfield (approved):

PR #1146, opened 2026-08-12
fix(metrics): count transaction forwarding drops by reason

### LINKED ISSUE #1133: metrics: Transaction forwarding drop paths are warn-only
## Problem

The forwarder counts drops on two paths: a batch shed at admission
(`ForwarderMetrics::record_batch_shed`, `crates/tn-reth/src/metrics.rs:124-126`, sole caller
`crates/tn-reth/src/forward.rs:342`) and transactions abandoned at the batch budget
(`record_txns_abandoned`, `metrics.rs:133-135`). Every other way the pipeline can drop
transactions only warns:

- worker side, no committee validator advertised an endpoint
  (`crates/consensus/worker/src/worker.rs:258-264`)
- worker side, endpoint discovery failed (`worker.rs:265-271`)
- forwarder side, endpoints were advertised but none resolved to a usable provider
  (`forward.rs:315-326`)
- the admission gate `if let Some(providers)` has no else arm (`forward.rs:333-354`), and an
  empty committee is a silent no-op (`forward.rs:312-314`)

There is no tracing-to-metrics bridge in the codebase, so these warns reach no dashboard. An
observer that is dropping every transaction it accepts looks healthy on metrics.

## Solution

One reason-labeled counter for the forwarding pipeline, e.g.
`tn_reth.forwarded_txns_dropped_total{reason=...}`, registered at zero per reason in
`ForwarderMetrics::init` for the same reason the existing counters register eagerly: an absent
series is indistinguishable from a broken exporter.

For the worker-side sites, a `WorkerMetrics::record_forward_dropped(reason, count)` modeled on
the existing labeled recorders `record_quorum_failure`
(`crates/consensus/worker/src/metrics.rs:66-81`) and `record_batches_fetched`
(`metrics.rs:94-104`). The transaction count has to be captured in the `worker.rs:246-252`
window, before the vec moves into the spawned task.

`ForwarderMetrics` can stay `pub(crate)`.

Worth adding in the same pass: a queued-side base series (`record_txns_queued` at the point a
batch is handed to the forwarder), so the drop and abandon counters have a denominator.

## Notes

Sequencing: the fix for observer seal admission (companion issue, link when filed) reshapes the
worker-side sites. Those drops become counted refusals that keep their transactions rather than
losses, so design the counter set once, after that lands, and the reason labels will describe
the final behavior.

MavenRain:
Closes #1133.

## Summary

The forwarding pipeline counted drops on two paths (the admission-time batch shed and the batch-budget abandon) and only warned on every other one.  There is no tracing-to-metrics bridge in the codebase, so those warns reach no dashboard:  an observer that is dropping every transaction it accepts looks healthy on metrics.  This PR gives the pipeline one reason-labeled drop counter per scope, plus the queued-side base series the issue asked for, so the losses have both a signal and a denominator.

## Changes

- [x] `crates/tn-reth/src/metrics.rs`:  a `ForwardDropReason` sum type labeling `tn_reth.forwarded_txns_dropped_total{reason}` with `empty_committee`, `no_usable_endpoint`, `batch_shed`, `rejected`, and `unreached`, each documented with its alertability;  a queued-side base series `tn_reth.forwarded_txns_queued_total`;  both registered at zero in `ForwarderMetrics::init` for the reason the existing counters register eagerly (an absent series is indistinguishable from a broken exporter, and a `rate()` whose first sample is nonzero renders no step).
- [x] `crates/tn-reth/src/forward.rs`:  `forward_txns` counts the queued intake before any admission decision, counts the previously silent empty-committee no-op, counts the advertised-but-unusable endpoint drop beside its existing warn, and counts the shed batch at transaction grain beside the existing batch-grain counter;  `spawn_forward` tallies `rejected` and `unreached` outcomes into the same labeled counter at batch end.  `queued - dropped - abandoned` is now the number a dashboard can trust as delivered.
- [x] `crates/consensus/worker/src/metrics.rs`:  `WorkerMetrics::record_forward_dropped(reason, count)` modeled on the existing labeled recorders `record_quorum_failure` and `record_batches_fetched` (a zero count records nothing, so a path that never dropped creates no series), labeling `tn_worker.forwarded_txns_dropped_total{worker,reason}` with a worker-side `ForwardDropReason` sum type.
- [x] `crates/consensus/worker/src/worker.rs`:  the two discovery drop arms in `disburse_txns` (`no_endpoint_advertised`, `discovery_failed`) record the whole-batch loss;  the transaction count is captured before the vec moves into the spawned task, and the metrics handle moves in beside it.

Cardinality is bounded by construction:  every `reason` value comes from a closed sum type, so the series set is a compile-time constant and no attacker-influenced string (endpoint URL, error text) ever becomes a label value.

## Testing

Run under the pinned toolchain with the OOM cap:

- `cargo +1.94 test -j 2 -p tn-reth --lib -- metrics forward`:  new coverage for the drop reasons registering at zero and moving independently (`test_forward_drop_reasons_register_at_zero_and_count_separately`), both previously silent routing dead ends (`forward_txns_counts_a_batch_with_no_committee_as_dropped` and `forward_txns_counts_a_batch_with_no_advertised_endpoint_as_dropped`), the unusable-endpoint drop (`forward_txns_counts_a_batch_with_no_usable_endpoint_as_dropped`), and the shed path extended to assert the transaction-grain counter and the queued denominator.
- `cargo +1.94 test -j 2 -p tn-worker --lib -- metrics`:  `record_forward_dropped` asserted per `worker` and `reason` labels, with the zero-count negative half asserted the same way the fetch recorder's is (a zero count must not register a series).
- `cargo +nightly fmt` and `cargo +1.94 check -j 2 --all-targets -p tn-reth -p tn-worker`.

The `rejected` and `unreached` sites record from inside the spawned forward task, where a test-local recorder cannot be observed (the existing budget-abandon test asserts timing for the same reason);  their recorder helper is covered by the unit test above.

## Notes

- Based on current `main`, which now carries the admission-control forwarder (GHSA-7v6m) this instruments.  The companion fix for #1132 will reshape the worker-side sites into counted refusals that keep their transactions;  the reasons here are labeled per site, so that change moves the worker-side counter's semantics (from lost to refused) without invalidating the forwarder-side set.
- `ForwarderMetrics` stays `pub(crate)`, as the issue suggested.
- The existing `tn_reth.forwarded_batches_shed_total` and `tn_reth.forwarded_txns_abandoned_total` series are unchanged;  the abandon counter stays separate from the drop counter so the two subtract from the queued denominator without double counting.

### DIFF
--- a/crates/consensus/worker/src/metrics.rs
+++ b/crates/consensus/worker/src/metrics.rs
@@ -8,6 +8,30 @@ use reth_metrics::{
 use std::time::Duration;
 use tn_types::WorkerId;
 
+/// Why the worker dropped a sealed batch's transactions instead of handing them to the
+/// transaction forwarder.
+///
+/// Both variants are whole-batch losses on the observer path (`disburse_txns`): the
+/// transactions were accepted by this node's RPC and never reached the forwarder, so the
+/// forwarder-side series (`tn_reth.forwarded_txns_*`) cannot see them (issue #1133).
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum ForwardDropReason {
+    /// No committee validator advertised a JSON-RPC endpoint to forward to.
+    NoEndpointAdvertised,
+    /// Discovering the committee's advertised endpoints failed outright.
+    DiscoveryFailed,
+}
+
+impl ForwardDropReason {
+    /// The `reason` label value this variant records under.
+    const fn label(self) -> &'static str {
+        match self {
+            Self::NoEndpointAdvertised => "no_endpoint_advertised",
+            Self::DiscoveryFailed => "discovery_failed",
+        }
+    }
+}
+
 /// Derive-backed metric handles for the worker, labeled per `worker`.
 #[derive(Metrics, Clone)]
 #[metrics(scope = "tn_worker")]
@@ -107,6 +131,24 @@ impl WorkerMetrics {
     pub(crate) fn record_batch_fetch_duration(&self, elapsed: Duration) {
         self.handles.batch_fetch_duration_seconds.record(elapsed);
     }
+
+    /// Record transactions dropped by the worker before they reached the transaction
+    /// forwarder, labeled by [`ForwardDropReason`].
+    ///
+    /// Modeled on [`Self::record_batches_fetched`]: a zero count records nothing, so a path
+    /// that never dropped creates no series. Every call site pairs with a `warn!` line; the
+    /// counter is what makes the loss visible on a dashboard, where an observer that drops
+    /// every transaction it accepts otherwise looks healthy (issue #1133).
+    pub(crate) fn record_forward_dropped(&self, reason: ForwardDropReason, count: usize) {
+        if count > 0 {
+            metrics::counter!(
+                "tn_worker.forwarded_txns_dropped_total",
+                "worker" => self.worker_id.to_string(),
+                "reason" => reason.label(),
+            )
+            .increment(u64::try_from(count).unwrap_or(u64::MAX));
+        }
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


grantkee (dismissed):
Great idea!

sstanfield (dismissed):

grantkee (approved):

PR #1150, opened 2026-08-12
fix(storage): latch the first background write failure, not the last

### LINKED ISSUE #1148: storage: queued `CertificatePack` saves after a failed write overwrite the latched root cause with a spurious "read only" error
**Class:** diagnosability / error-reporting.  The fail-stop itself is correct and intact: after a write failure the pack refuses every later append and commit.  The defect is that the surviving diagnostic names a condition that does not exist on a read-write pack, and the real IO error is lost.  Confirmed on `origin/main` `c3947d2c`.  Related: #1138 (surfaces this latch in flush replies, so what the latch holds becomes what operators see), #1094 (the `EpochRecordDb` precedent for draining this latch into a reply).

## Summary

`run_pack_loop` processes messages one at a time and latches every save failure with an unconditional overwrite (`crates/storage/src/certificate_pack.rs:59-64`):

```rust
while let Some(msg) = rx.blocking_recv() {
    match msg {
        PackMessage::Save(cert) => {
            if let Err(e) = inner.save(&cert) {
                tx_error.send_replace(Some(e));
            }
        }
```

`watch::Sender::send_replace` is last-write-wins.  There is no first-error guard at any of the latch's write sites, and nothing ever resets the slot to `None`.

The overwrite becomes systematic after a real write failure:

1. A `Save` fails in `PackInner::append` with `AppendError::WriteDataError` (an IO error during the data write, `crates/storage/src/archive/error/insert.rs:34-38`).  The append path sets `self.failed = true` (`crates/storage/src/archive/pack.rs:269`).  This is the root cause, and it carries the IO error text.
2. `run_pack_loop` latches it: `tx_error.send_replace(Some(e))` (`certificate_pack.rs:63`).
3. Every `Save` already queued behind it then hits the poisoned-pack guard (`archive/pack.rs:262`):

   ```rust
   if self.read_only || self.failed {
       return Err(AppendError::ReadOnly);
   }
   ```

   The pack was opened read-write (`Inner::open(path, false)`, `certificate_pack.rs:108`), but each queued save now fails with the read-only error.
4. Each of those failures maps through `PackError::Append(e.to_string())` (`certificate_pack.rs:283`) and overwrites the latch again.  The last queued save decides the final value: `PackError::Append("read only")`, displayed as `Data Append Error (read only)`.
5. `CertificatePack::get_error` reads with `borrow()` and never clears (`certificate_pack.rs:133-139`).  So the mask is permanent: every later `save`/`try_save`/`persist` caller sees only "read only".  The original `WriteDataError` text is gone.

The only way the root cause survives is when exactly one `Save` is in flight at failure time.  With two or more queued, the loop overwrites it synchronously, before any reader can observe it.

Notes on scope:

- Only `WriteDataError` sets `failed` (`DuplicateKey`/`CrcError`/`SerializeValue` do not, `archive/pack.rs:271-274`), so the cascade needs a genuine IO write failure first.  `failed` is never reset.
- The same guard shape sits in `PackInner::commit` (`archive/pack.rs:298-305`, `CommitError::ReadOnly`) and in the digest-index leg (`HdxIndex::save`, `crates/storage/src/archive/digest_index/index.rs:762-765`), so a flush queued behind a failed save also reports "read only" rather than the cause.
- Independent of the cascade, the latch loses history for any two failures in general: last write always wins.  The `ReadOnly` cascade is the sharpest form because the follow-on failures are guaranteed and mislabeled.

## `EpochRecordDb` has the same latch primitive, with softer dynamics

`crates/storage/src/epoch_records.rs` uses the identical shape: `error: watch::Sender<Option<EpochDbError>>` (`epoch_records.rs:99`), written unconditionally with `send_replace` from all four save arms (`epoch_records.rs:115`, `121`, `127`, `133`).  Two queued failing saves with no reader between them lose the first error the same way: the client-side `self.get_error()?` pre-check on each save API (for example `epoch_records.rs:445`) samples the slot at enqueue time, before the actor has attempted the earlier write, so it cannot close the window.

The dynamics differ in two ways:

- Its `get_error` drains the slot on read (`send_replace(None)`, `epoch_records.rs:424-431`).  A mask lasts only until the next read, not forever.
- The two races that already have targeted fixes are save-then-persist (the `Persist` arm drains the slot and folds it into the reply, `epoch_records.rs:163-171`, test at `epoch_records.rs:1775-1803`) and read-vs-write (`peek_error`, test at `epoch_records.rs:1749-1773`).  Save-then-save has no fix and no test.

## Suggested change

Either of the first two options fixes the misdiagnosis; the first is the smallest.

1. **Keep the first error.**  In each failure arm, write the latch only when the slot is empty, for example `send_if_modified` with a closure that refuses to replace `Some`, or a `borrow()` emptiness check before `send_replace`.  The root cause then survives any number of follow-on failures.  For `EpochRecordDb` this composes with the clearing `get_error`: draining still acknowledges exactly one failure; first-write-wins only changes which failure that is.
2. **Name the poisoned state honestly.**  Give the post-failure guard its own variant (for example `AppendError::Failed`) instead of reusing `ReadOnly`, so even a last-write-wins latch never claims a read-only condition on a read-write pack.  Larger surface: new variant, `Display`, and the matching `CommitError` leg.
3. Both, if the extra variant is wanted for its own sake.

**Regression tests** (both stores): enqueue two failing saves back to back with no `get_error`/`persist` between them, then assert the surviving latched error names the original write failure (options 1/3), or the poisoned state (option 2 alone), and not "read only".  For `EpochRecordDb` the failing-write injection already used by `persist_reports_a_write_that_failed_while_the_flush_was_queued` should extend directly.  For `CertificatePack` the injector needs care: the read-only-reopen injector from the #1138 tests produces `AppendError::ReadOnly` without setting `failed`, so it cannot start this cascade; the first failure must be a real `WriteDataError` (fs-level fault injection, or a test hook that sets the pack's `failed` flag directly).

MavenRain:
Closes #1148.

## Problem

After a real IO write failure, `CertificatePack` correctly refuses every later append and commit. The surviving diagnostic is wrong. Each save queued behind the failure hits the poisoned-pack guard and fails with `AppendError::ReadOnly`, and `run_pack_loop` latched every failure with an unconditional overwrite. The last queued save decided the final value, so `get_error` reported `Data Append Error (read only)` for a pack that was opened read-write, and the original `WriteDataError` text was gone permanently. No attacker is required; one disk-level write failure with two or more saves in flight is enough.

`EpochRecordDb` has the same latch primitive. Its `get_error` drains the slot on read, so the mask is transient there, but two queued failing saves with no reader between them lose the first error the same way.

## Root cause

`watch::Sender::send_replace(Some(e))` is last-write-wins, and no latch write site had a first-error guard. The root cause (the only failure that carries the IO error text, and the only one that sets `failed = true`) is exactly the error that follow-on failures overwrite, because every save queued behind it is guaranteed to fail with the mislabeled `ReadOnly` error.

## Fix

- [x] `crates/storage/src/error_latch.rs` (new): `latch_first_error`, a shared first-error-wins write for the background error latches. It fills the slot only when the slot is empty, via `send_if_modified`.
- [x] `crates/storage/src/certificate_pack.rs`: all three latch write sites (the `Save` arm and both open-failure arms) go through `latch_first_error`, and the `Save` arm now logs each failure the way `EpochRecordDb` already does, so the errors the latch does not keep stay visible. The `get_error` docs state the first-error contract.
- [x] `crates/storage/src/epoch_records.rs`: all four save arms go through `latch_first_error`. Draining reads are unchanged: `get_error` and the `Persist` reply still acknowledge exactly one failure; first-write-wins only changes which failure that is, which in the poisoned-pack cascade is the root cause.
- [x] `crates/storage/src/archive/pack.rs`: a `cfg(test)` injector (`fail_next_append_for_test`) makes the next append fail with `AppendError::WriteDataError`, the same classification a real IO write failure gets, so the tests can start the cascade from a genuine poisoned state. It compiles to a no-op outside tests.

This composes with #1143 (fold the latched error into `CertificatePack` flush replies): what the latch holds is what those replies report, so keeping the first error makes them name the real failure instead of the spurious read-only condition.

## Testing

Regression tests, one per store, both confirmed by mutation (revert the latch to `send_replace`, watch the test fail, restore):

- `certificate_pack::test::queued_saves_after_a_failed_write_keep_the_root_cause`: arms the injector, drives `run_pack_loop` directly, and queues two saves back to back with no reader between them. The latch must keep the injected write failure and must not report `read only`. The test pins the error variant and carries a positive control for the negative assertion (the guard error really renders as `read only`), so a later `Display` reword cannot make it pass vacuously.
- `epoch_records::test::queued_save_failures_keep_the_first_error`: queues two out-of-order saves (distinguishable errors) and a persist behind them. The persist reply drains the latch and must carry the first failure.

Both tests order events through the actor's single FIFO channel, so the failure order is deterministic rather than a race. Full battery on the pinned toolchain in an isolated target dir: nightly `cargo fmt -- --check`, nightly `clippy --workspace` and `clippy --workspace --all-features` with `-D warnings`, `cargo +1.94 test --no-run --workspace`, and `cargo +1.94 test -p tn-storage` to run the storage suite.

### DIFF
--- a/crates/storage/src/archive/pack.rs
+++ b/crates/storage/src/archive/pack.rs
@@ -90,6 +90,15 @@ where
         self.inner.append(value)
     }
 
+    /// Test-only failure injector: make the next append fail with
+    /// [`AppendError::WriteDataError`], the same classification a real io write failure gets.
+    /// The append path then marks the pack failed, which is the poisoned state the queued-save
+    /// regression tests start from.
+    #[cfg(test)]
+    pub(crate) fn fail_next_append_for_test(&mut self) {
+        self.inner.fail_next_append = true;
+    }
+
     /// Return the DB version.
     pub fn version(&self) -> u16 {
         self.inner.version()
@@ -165,6 +174,9 @@ where
     failed: bool,
     read_only: bool,
     uid_idx: u64, // Store for opening an iterator.
+    /// Test-only: when set, the next append fails as if the data write hit an io error.
+    #[cfg(test)]
+    fail_next_append: bool,
     _value: PhantomData<V>,
 }
 
@@ -202,6 +214,8 @@ where
             failed: false,
             read_only,
             uid_idx,
+            #[cfg(test)]
+            fail_next_append: false,
             _value: PhantomData,
         })
     }
@@ -233,9 +247,26 @@ where
         Ok(bytes)
     }
 
+    /// Test-only injection point: fail the append the way a real io write failure fails.
+    /// Armed by [`Pack::fail_next_append_for_test`]; disarms after one use.
+    #[cfg(test)]
+    fn injected_append_failure(&mut self) -> Result<(), AppendError> {
+        std::mem::take(&mut self.fail_next_append)
+            .then(|| AppendError::WriteDataError(io::Error::other("injected write failure")))
+            .map_or(Ok(()), Err)
+    }
+
+    /// Outside tests the injection point compiles to a no-op.
+    #[cfg(not(test))]
+    fn injected_append_failure(&mut self) -> Result<(), AppendError> {
+        Ok(())
+    }
+
     /// Do the actual insert so the public function can rollback easily on an error.
     fn append_inner(&mut self, value: &V) -> Result<u64, AppendError> {
         let record_pos = self.data_file.len();
+
+        self.injected_append_failure()?;
 
         write_value(
             value,
--- a/crates/storage/src/archive/pack.rs
+++ b/crates/storage/src/archive/pack.rs
@@ -83,9 +83,10 @@ where
     ///   - key data
     ///   - value data
     ///
-    /// For the erros IndexCrcError, IndexOverflow, WriteDataError or KeyError the DB will move to a
-    /// failed state and become read only.  These errors all indicate serious underlying issues that
-    /// can not be trivially fixed, a reopen/repair might help.
+    /// A WriteDataError moves the DB to a failed state.  While the DB is failed, each append
+    /// and each commit returns a copy of the error that caused the failed state.  This error
+    /// indicates a serious underlying issue that can not be trivially fixed, a reopen/repair
+    /// might help.
     pub fn append(&mut self, value: &V) -> Result<u64, AppendError> {
         self.inner.append(value)
     }
@@ -117,6 +118,8 @@ where
     /// Flush any caches to disk and sync the data and index file.
     /// All data should be safely on disk if this call succeeds.
     /// Note this is an expensive call (syncing to disk is not cheap).
+    /// On a pack in the failed state this returns [`CommitError::Failed`] with a copy of the
+    /// error that caused the failed state.
     pub fn commit(&mut self) -> Result<(), CommitError> {
         self.inner.commit()
     }
@@ -171,7 +174,10 @@ where
     value_buffer: Vec<u8>,
     /// Used as a second buffer for compress and decompress operations on records.
     compression_buffer: Vec<u8>,
-    failed: bool,
+    /// Root cause of the failed state: a copy of the io error from the append that failed
+    /// the pack. While this is `Some`, each append and each commit returns a copy of this
+    /// error so callers see the root cause and not a generic guard error.
+    failed: Option<io::Error>,
     read_only: bool,
     uid_idx: u64, // Store for opening an iterator.
     /// Test-only: when set, the next append fails as if the data write hit an io error.
@@ -211,7 +217,7 @@ where
             data_file,
             value_buffer: Vec::new(),
             compression_buffer: Vec::new(),
-            failed: false,
+            failed: None,
             read_only,
             uid_idx,
             #[cfg(test)]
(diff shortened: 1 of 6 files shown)


grantkee (dismissed):
Good catch

crates/storage/src/archive/pack.rs
```diff
@@ -233,10 +247,27 @@ where
...
+    /// Armed by [`Pack::fail_next_append_for_test`]; disarms after one use.
+    #[cfg(test)]
+    fn injected_append_failure(&mut self) -> Result<(), AppendError> {
+        std::mem::take(&mut self.fail_next_append)
+            .then(|| AppendError::WriteDataError(io::Error::other("injected write failure")))
+            .map_or(Ok(()), Err)
+    }
+
+    /// Outside tests the injection point compiles to a no-op.
+    #[cfg(not(test))]
```
sstanfield: There is no reason to compile and call this for prod.  I think it should be removed and a cfg(test) added to the call in append_inner() below.  Not a big deal but still less prod code to produce/run.

sstanfield (dismissed):
Good catch.  I am ok with this but consider making the "failed" field in Pack an Option<AppendError> (or Option<std::io::Error>) and fixing this at the pack level.  It would make more sense for the Pack file, once in an error state, to always return the error that caused that state vs ReadOnly.  CommitError may need a new discriminate to capture this but I think this might be cleaner and since once in a failed state the pack writes would always return that error.  The error latch might be good to keep as defense in depth any way.

MavenRain:
Done. `failed` is now `Option<io::Error>` holding the error that moved the pack to the
failed state. While the pack is failed, append and commit both return a copy of that
error (kind and message; `io::Error` is not `Clone`), so callers see the root cause
instead of `ReadOnly`. Commit reports it through a new `CommitError::Failed(io::Error)`
discriminant. I kept the first-error latch as defense in depth for failures that do not
share one root cause, as you suggested. A new pack-level regression test
(`failed_pack_replays_the_root_cause`) pins the replay on both paths, including the
error kind, with a read-only reopen as the positive control for the guard error. While
touching the `append` docs I also corrected them to match the classifier: only
`WriteDataError` moves the pack to the failed state; the old text listed error names
that do not exist on `AppendError`.

Two observations from working in this area, both pre-existing behavior this PR does not
change: `flush()` is not guarded by the failed state (only append and commit are), and
`DataFile::drop` still flushes and syncs unconditionally, so a poisoned pack's buffered
bytes can reach disk through those paths. Also, append replays the root cause through
`AppendError::WriteDataError` while commit got its own discriminant; if you would rather
have a symmetric `AppendError::Failed` so a replayed error is distinguishable from a
fresh write failure in the logs, happy to follow up.

crates/storage/src/archive/pack.rs
MavenRain: Done. The `#[cfg(not(test))]` no-op is removed and the call in `append_inner()` is now
`#[cfg(test)]`, so no injector code compiles for prod.

grantkee (approved):
Thanks @MavenRain - good work

PR #1151, opened 2026-08-12
test(e2e): cut eject restart test runtime by about 36%

### LINKED ISSUE #1147: Speed up e2e test_committee_member_restarted_mid_epoch_after_ejection by another ~36%: give the warmup epoch its own short duration via upgradeStakeVersion
## Problem

`e2e-tests::it eject::test_committee_member_restarted_mid_epoch_after_ejection` runs about 84s (solo, prebuilt `TN_BIN_PATH`).  After #1009 (entry guard, one warmup epoch removed), #1117 (follow-up transfer ends the post-transfer idle under skip-empty-execution), and #1120 (teardown), the profile is simple: two full wall-clock epochs at `RESTART_EPOCH_DURATION = 40` (`crates/e2e-tests/tests/it/eject.rs:71`) plus about 4s of real work.  A fresh audit of every awaited wait in the test found no remaining non-boundary wait above 2s.  The epoch schedule is the whole remaining cost.

The two epochs do different jobs:

- Epoch 0 exists only so that epoch record 0 is produced and certified.  The kill gate needs the kill target to serve certified record 0 (`eject.rs:740-742`).
- Epoch E hosts the whole in-epoch sequence: rotation, burn, execution pin, kill, restart, mid-epoch rejoin.  The doc contract budgets ~29s worst case plus margin (`eject.rs:63-70`).  This epoch must stay at 40s.  A uniform duration cut erodes the rejoin-regression margin and was already rejected for that reason.

Genesis seeds one duration for every epoch, so today the warmup epoch pays the 40s that only the sequence epoch needs.

## Proposal

Give the two epochs different durations with the governance path that already exists.  No node changes are needed.

1. Seed genesis with a short duration, `S = 8`.  Precedent: the sibling test ships 6s epochs (`EJECT_EPOCH_DURATION`, `eject.rs:82`), and `basefee.rs` / `epochs.rs` ship 5s, the documented consensus minimum.
2. Concurrently with startup, the governance wallet (seed 33, the registry owner via `create_genesis_for_test`, `eject.rs:711-718`) sends `upgradeStakeVersion(newConfig)`, where `newConfig` copies the current `StakeConfig` and changes only `epochDuration` to 40.  Reuse `send_owner_tx` (`common.rs:1055`) with the one-retry boundary-orphan pattern of `burn_with_retry` (`eject.rs:104-123`).  The `send_owner_tx` confirmation ceiling stays keyed on the module `EPOCH_DURATION` and is untouched.
3. The contract stamps the new duration into the next epoch at its conclusion.  `_updateEpochInfo` reads `getCurrentStakeConfig()` fresh when it opens epoch N+1 (tn-contracts @ `dc248c0d`: `ConsensusRegistry.sol:1297-1316`;  `StakeManager.sol:79-84` returns `versions[stakeVersion]` with no delay;  `upgradeStakeVersion` at `ConsensusRegistry.sol:1560-1581` is `onlyOwner` and validates only a nonzero duration).  `futureEpochInfo` pre-pins committees, not config, so the new duration is not deferred to N+2.
4. The node already reads the duration from chain state for every epoch: `epoch_boundary = epoch_start + epoch_info.epochDuration` (`crates/node/src/manager/node/run_epoch.rs:193-195`), fed by a `getCurrentEpochInfoCall` read at the epoch-start header (`crates/tn-reth/src/env/epoch.rs:860-878`), with epoch 0 pinned to the genesis header through the same path (`epoch.rs:339-352`).  The binding's doc note states the per-epoch stamp explicitly (`crates/tn-reth/src/system_calls.rs:112-116`).  No static duration flag exists in the node.
5. Key the sequence epoch off the upgrade receipt.  With the upgrade confirmed in epoch R, every epoch from R+1 on runs at 40s.  Wait for a fresh entry into epoch R+1 and run the existing `SEQUENCE_BUDGET_SECS = 33` guard there (`eject.rs:756-758`), with its fallback kept as a belt.  Do not assert "still in epoch 0": under CI load the receipt can land in epoch 1, and that path must stay green.
6. Spawn `fetch_verified_epoch_record(0)` concurrently and await it just before the kill, so record-0 certification is not squeezed by the short epoch 0.
7. Every timeout ceiling keeps its current anchor (module `EPOCH_DURATION` or `RESTART_EPOCH_DURATION`).  Rewrite the `RESTART_EPOCH_DURATION` doc contract (`eject.rs:63-70`) to separate its two roles: the genesis seed becomes `S`, while the constant stays the anchor for the sequence budget and the ceilings.

## Expected runtime

| Path | Schedule | Total | Cut vs ~84s |
| --- | --- | --- | --- |
| Nominal: receipt lands in epoch 0 | 8 + 40 + ~4-6s tail | ~52-54s | ~36-38% |
| Degraded: receipt lands in epoch 1 | 8 + 8 + 40 + tail | ~60-62s | ~26-29% |

`S = 8` is deliberate.  At `S = 10` the degraded path lands at ~64-66s and misses a 25% floor.

## Coverage is unchanged

- Certified epoch records embed no duration.  Record verification checks digests, the BLS quorum, and the final state, so mixed durations cannot invalidate any record the test asserts on.
- The test asserts no duration or stake-version constancy.  The only such assertion lives in `loop_epochs`, which this test does not call.
- Reward math reads the per-validator genesis stake version, so the global version bump does not change reward weights.  Copying the three non-duration `StakeConfig` fields verbatim keeps stake and issuance byte-identical.
- The entry guard stays sound under mixed durations: `epoch_seconds_remaining` computes from the entered epoch's own on-chain duration snapshot (`common.rs:1232-1241`), not from a compile-time constant.
- The regression assertions (`build_epoch_record` divergence, mid-epoch rejoin, six-node boundary close, record verification on all six nodes) run exactly as today, in a 40s epoch with the same 33s budget.
- Bonus coverage: the run now exercises `upgradeStakeVersion` end to end (config authored in epoch N activates at N+1).  No e2e test calls it today.

## Prerequisite

Two additive signatures in the `sol!` block in `crates/tn-reth/src/system_calls.rs`: `upgradeStakeVersion(StakeConfig)` and `getCurrentStakeConfig()`.  The block already binds `struct StakeConfig` (`system_calls.rs:145-154`) and `burn` for this same test (`:223`), and it has `getCurrentStakeVersion` (`:236`) but neither of the two needed here.

## Alternatives considered

- Uniform cut of `RESTART_EPOCH_DURATION` (40 to 32/35): rejected before and still rejected.  It shrinks the margin that keeps a CI-slow restart from turning the rejoin-regression assert into a flake.
- Tightening any poll or ceiling: the ceilings return early and cost ~0;  the wait audit found no non-boundary slack above 2s left after #1009 / #1117 / #1120.

Prior art: #1009 (entry guard), #1116 / #1117 (post-transfer idle), #1121 / #1124 (6s epochs for the sibling test), #1119 / #1120 (boundary-kill speedup).  This issue applies the same wall-clock-boundary methodology to the last reducible boundary in the module.

MavenRain:
Closes #1147.

## Summary

`eject::test_committee_member_restarted_mid_epoch_after_ejection` ran about 84s. After #1009, #1117, and #1120 the profile is two full wall-clock epochs at `RESTART_EPOCH_DURATION = 40` plus about 4s of real work. Only the sequence epoch needs the 40 seconds. The warmup epoch exists only to produce and certify epoch record 0 for the kill gate. This PR gives the two epochs different durations with the governance path that already exists (`upgradeStakeVersion`). Test-only plus two additive bindings; no node changes.

## Changes

- [x] `crates/e2e-tests/tests/it/eject.rs`: new `RESTART_WARMUP_EPOCH_DURATION = 8` genesis seed. The `RESTART_EPOCH_DURATION = 40` doc contract now separates its two roles: it stays the anchor for every budget and ceiling, and is no longer the genesis seed. 8s sits above the documented 5s consensus minimum (`basefee.rs` / `epochs.rs` ship 5s, the sibling eject test 6s).
- [x] `crates/e2e-tests/tests/it/eject.rs`: new `upgrade_duration_with_retry` helper. At startup the governance wallet (nonce 0) sends `upgradeStakeVersion` with the current `StakeConfig` copied verbatim except `epochDuration = 40`. Same one-retry boundary-orphan pattern as `burn_with_retry`.
- [x] `crates/e2e-tests/tests/it/eject.rs`: the test enters the first epoch observed running the 40s duration, not "receipt epoch + 1". A receipt that lands in epoch 1, or in a closing block (the stamp then defers one epoch), only adds one more 8s warmup epoch and stays green. The `SEQUENCE_BUDGET_SECS = 33` runway guard and its fresh-boundary fallback are unchanged.
- [x] `crates/e2e-tests/tests/it/eject.rs`: the record-0 kill gate is spawned concurrently at startup and awaited before the sequence epoch is entered, so the short warmup does not squeeze record-0 certification, gate latency stays outside the measured 33s sequence budget, and a gate failure aborts before the irreversible governance burn. Its 120s ceiling starts at process launch, so it also covers multi-process cold boot.
- [x] `crates/tn-reth/src/system_calls.rs`: two additive `sol!` bindings, `upgradeStakeVersion(StakeConfig)` and `getCurrentStakeConfig()`, with doc comments. Verified against tn-contracts @ dc248c0d: `upgradeStakeVersion` is owner-only, bumps the version immediately, and `_updateEpochInfo` stamps `epochDuration` fresh from `getCurrentStakeConfig()` when it opens the next epoch (`futureEpochInfo` pre-pins committees, not config).

## Coverage is unchanged

- Certified epoch records embed no duration, and the test asserts no duration constancy. The regression assertions (mid-epoch rejoin, boundary close on all six nodes, record verification 0..=E with hash equality, liveness through the restarted node) run exactly as today, in a 40s epoch with the same 33s budget.
- Reward math reads the per-validator genesis stake version. Copying the three non-duration `StakeConfig` fields verbatim keeps stake and issuance byte-identical.
- The entry guard stays sound under mixed durations: `epoch_seconds_remaining` computes from the entered epoch's own on-chain duration snapshot, not a compile-time constant.
- An ineffective upgrade cannot pass silently: the duration-keyed entry wait times out at `RESTART_EPOCH_DURATION * 2` and fails the test loudly. Confirmed by mutation (see Verification).
- Bonus coverage: this is the only e2e exercise of `upgradeStakeVersion` end to end (config authored in epoch N activates at N+1).

## Verification

Measured on 80171ff9 (12 cores, prebuilt `TN_BIN_PATH`, nextest, solo):

| Variant | Runs |
| --- | --- |
| baseline | about 84s (issue #1147 wait audit; profile: two 40s epochs plus about 4s of work) |
| patched | 55.996s, 53.936s, 53.633s |

All three patched runs sit on the nominal schedule (8s warmup + 40s sequence + tail). The degraded path (upgrade receipt in epoch 1) would read about 60-62s and stays green by design.

- Mutation confirmation: with the upgrade made a no-op (it writes the warmup duration again), the duration-keyed entry wait timed out and the test failed loudly (rc=100). The file was then restored from the index and re-verified byte-identical.
- Gates, all green: nightly `fmt --check`, `+1.94 clippy --all-targets`, nightly `clippy --workspace --all-features`, and the attest proxy `+1.94 test --no-run --workspace -j 2` in an isolated target dir.

### DIFF
--- a/crates/e2e-tests/tests/it/eject.rs
+++ b/crates/e2e-tests/tests/it/eject.rs
@@ -60,15 +60,32 @@ use tracing::info;
 /// (genesis creation requires the extra node) but never starts it.
 const GENESIS_OBSERVER: &str = "genesis-observer";
 
-/// Per-test epoch duration (seconds) for the mid-epoch restart regression.
+/// Sequence-epoch duration (seconds) for the mid-epoch restart regression.
 ///
 /// [`test_committee_member_restarted_mid_epoch_after_ejection`] must fit its whole in-epoch
-/// sequence — fresh-boundary entry, one full leader rotation, the governance burn, the burn-block
-/// execution pin, kill, restart, and the mid-epoch consensus rejoin — inside ONE epoch. That
-/// sequence's worst case is ~29s, so the module-wide [`EPOCH_DURATION`] (10s) cannot host it;
-/// 40s banks comfortable margin without stretching the test into extra epochs. Every timeout in
-/// that test which scales with epoch length derives from this constant, not [`EPOCH_DURATION`].
+/// sequence (one full leader rotation, the governance burn, the burn-block execution pin, kill,
+/// restart, and the mid-epoch consensus rejoin) inside ONE epoch. That sequence's worst case is
+/// ~29s, so the module-wide [`EPOCH_DURATION`] (10s) cannot host it; 40s banks comfortable
+/// margin without stretching the sequence across extra epochs.
+///
+/// This constant is the anchor, not the genesis seed: genesis ships the short
+/// [`RESTART_WARMUP_EPOCH_DURATION`] and the test restores this duration on chain (via
+/// `upgradeStakeVersion`) for the epoch hosting the sequence, because only that epoch needs the
+/// 40s. Every timeout in the test which scales with epoch length keeps deriving from this
+/// constant, not [`EPOCH_DURATION`] and not the warmup seed.
 const RESTART_EPOCH_DURATION: u64 = 40;
+
+/// Genesis epoch duration (seconds) for the mid-epoch restart regression's warmup epochs.
+///
+/// The warmup epochs before the sequence epoch have one job: produce and certify epoch record 0,
+/// which the kill gate requires the kill target to serve. Genesis seeds ONE duration for every
+/// epoch, so without a second lever the warmup would pay the 40s that only the sequence epoch
+/// needs; instead genesis seeds this short duration and the test's startup `upgradeStakeVersion`
+/// restores [`RESTART_EPOCH_DURATION`] from the next epoch opening on. 8s sits above the
+/// documented 5s consensus minimum (`basefee.rs` / `epochs.rs` ship 5s, the sibling test 6s per
+/// [`EJECT_EPOCH_DURATION`]) with margin for the upgrade to confirm inside epoch 0 on the
+/// nominal path.
+const RESTART_WARMUP_EPOCH_DURATION: u64 = 8;
 
 /// Per-test epoch duration (seconds) for the mid-epoch current-committee ejection test.
 ///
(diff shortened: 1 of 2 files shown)


grantkee (approved):
This is great!

sstanfield (approved):

PR #1152, opened 2026-08-12
test(storage): add TN_TEST_MDBX_SYNC runtime override and scheduled Durable e2e lane

### LINKED ISSUE #1149: test(storage): add a TN_TEST_MDBX_SYNC runtime override and a scheduled Durable e2e lane (follow-up to #982)
## Summary

PR #982 switched the test-build consensus MDBX environments to `SafeNoSync` at compile time.  The switch is a cfg-gated constant (`BUILD_SYNC_MODE`) in `MdbxDatabase::open`.  In the #982 review thread, @sstanfield raised a concern:

> Not sure about this.  We might mask a timing issue caused by syncing that the e2e tests would fail to identify.

The reply in that thread offered a runtime override: "SafeNoSync stays the default for the fast path, but `TN_TEST_MDBX_SYNC=durable` forces Durable at `open()` with no rebuild", plus a scheduled Durable e2e lane.  The thread ended on the question "fold that into this PR, or land this as-is and track the Durable lane separately?".  The PR merged as-is.  No tracking issue exists.  This issue is that tracker.

PR #1142 makes this timely.  Its benchmark comparison had to comment out the cfg block in the source to A/B the two sync modes.  A runtime override gives the same A/B with no source edit.  It also produces the one #917 acceptance item that nobody measured: the wall-clock delta on a real CI disk.

## Background

- Issue #917 proposed the SafeNoSync switch.  Its acceptance criteria include: "Report the wall-clock delta on a slow-disk CI runner (where the win is real)."  That measurement was never reported on #917 or #982.
- PR #982 (merge e95d73b8) landed the compile-time switch.  A guard test (`test_open_uses_safe_no_sync_under_test_cfg`) pins the const and reads the mode back from the live environment.
- The e2e suite now never runs the Durable timing regime.  Under Durable, the externalization barriers (`persist().await` before a certificate, header, or vote goes on the wire) wait on a real per-commit flush.  That wait races the proposer timers (1000/2500 ms).  SafeNoSync removed this regime from every test run.
- PR #1142 (open) adds `db_backend_comparison` and asks whether the SafeNoSync block is needed.  The bench numbers alone cannot answer that.  The answer needs the e2e delta on a CI disk.

## Proposed change

**1.  Runtime override, test builds only.**  Read `TN_TEST_MDBX_SYNC` inside the existing `#[cfg(any(test, feature = "test-utils"))]` block in `MdbxDatabase::open` (`crates/storage/src/mdbx/database.rs:168-175`).  The vendored `reth_libmdbx::SyncMode` already implements `FromStr` and accepts `durable` and `safe-no-sync` (`libmdbx-rs/src/flags.rs:124-137`).  When the variable is absent, keep `BUILD_SYNC_MODE` (`SafeNoSync`).  When the value does not parse, return an error from `open`.  A silent fallback could run the wrong mode and show a green Durable lane that never ran Durable.  Sketch:

```rust
{
    use reth_libmdbx::{EnvironmentFlags, Mode, SyncMode};
    let sync_mode = std::env::var("TN_TEST_MDBX_SYNC")
        .ok()
        .map(|raw| raw.parse::<SyncMode>().map_err(|e| eyre::eyre!(e)))
        .transpose()?
        .unwrap_or(BUILD_SYNC_MODE);
        mode: Mode::ReadWrite { sync_mode },
        ..Default::default()
    });
}
```

No plumbing is required.  The e2e harness spawns nodes through `std::process::Command` (`crates/e2e-tests/src/lib.rs`, `TestBinary::command`), and no call clears the child environment.  Export the variable in the CI step and every spawned node sees it.  Keep the `BUILD_SYNC_MODE` const assert, so the compiled default still cannot drift to `UtterlyNoSync`.  Production builds compile the whole block out, so the variable has no effect on a release binary.

**2.  Durable e2e lane.**  Run the restart and epoch e2e tests with `TN_TEST_MDBX_SYNC=durable` on a schedule.  A nightly cron job and a separate non-blocking PR job both work;  maintainer preference decides.  The lane keeps the Durable timing regime under test without a wall-clock cost on PR CI.

**3.  Report the measurement.**  The first lane runs give the Durable-vs-SafeNoSync wall-clock delta on the real CI disk.  Post the numbers on this issue.  That closes the open #917 acceptance item and gives #1142 a measured keep-or-drop answer for the SafeNoSync default.

## Acceptance criteria

- [ ] `TN_TEST_MDBX_SYNC=durable` forces `Durable` at `open()` in test builds.  An unset variable keeps `SafeNoSync`.  An invalid value is a hard error, not a fallback.
- [ ] The guard test covers both the default mode and the override, with a live read-back of the opened environment.
- [ ] Production builds do not read the variable.  The env read stays inside the cfg gate.
- [ ] A scheduled CI lane runs the restart/epoch e2e tests with the override set.
- [ ] The measured wall-clock delta is posted on this issue.

## Risk and scope

- Test-only.  The production open path does not change.
- The diff is small: one env read, one guard-test extension, one CI workflow entry.
- The override also removes the hand edit that #1142 used for its A/B comparison.

Line numbers are from `main` (c3947d2c);  re-confirm before implementation.

Refs: #917, #982, #1142.

MavenRain:
Closes #1149.

## Summary

PR #982 switched test-build MDBX environments to `SafeNoSync` at compile time.  That removed the Durable timing regime from every e2e run: under `Durable`, the externalization barriers (`persist().await` before a certificate, header, or vote goes on the wire) wait on a real per-commit flush that races the proposer timers (1000/2500 ms).  The #982 review thread asked for a runtime override plus a scheduled Durable lane;  this PR adds both.  `TN_TEST_MDBX_SYNC` now overrides the compiled default at `MdbxDatabase::open` in test builds, and a nightly workflow runs the restart and epoch e2e suites with `TN_TEST_MDBX_SYNC=durable`.  The lane's step timings give the #917 acceptance measurement (Durable vs SafeNoSync wall clock on a CI disk), and the override replaces the source edit #1142 needed for its A/B comparison.

## Changes

- [x] `crates/storage/src/mdbx/database.rs`: `resolve_sync_mode` reads `TN_TEST_MDBX_SYNC` inside the existing `#[cfg(any(test, feature = "test-utils"))]` block.  `durable` forces `Durable` at `open()`;  `safe-no-sync` selects the compiled default explicitly;  an unset variable keeps `BUILD_SYNC_MODE` (`SafeNoSync`).  A set-but-invalid value (a spelling `SyncMode::from_str` rejects, or a non-UTF-8 value) is a hard error from `open`, never a fallback: a silent fallback could green-run a "Durable" lane that never ran Durable.  Every spelling the parser accepts maps to `Durable` or `SafeNoSync`, so no environment value can reach `UtterlyNoSync`, and the `BUILD_SYNC_MODE` const assert still pins the compiled default.  Production builds compile the whole block out, variable read included.
- [x] `crates/storage/src/mdbx/database.rs`: the #982 guard test now covers the default, the override, and the invalid value, each against a live `open` (the invalid case proves `open` fails;  the other two read the opened environment back).  Every live read-back runs in a child process of the test binary (`--exact` plus a controlled child env), so the parent never mutates process env: sibling tests opening databases on other threads (plain `cargo test`) cannot observe a half-set variable, a panicking phase cannot leak a poisoned value, and the spawn path exercises the same env inheritance the Durable lane relies on.  Each child run asserts "1 passed" from the harness output, so a drifted test name cannot pass vacuously.  The override read-back is exact (`Durable` sets no no-sync bits, so the UtterlyNoSync mask caveat does not apply), and the test pins the literal variable name against const drift.
- [x] `Makefile`: new `test-epochs` target, the epoch twin of `test-restarts` (same filter as the `public-tests` epoch line).
- [x] `.github/workflows/durable-e2e.yaml`: nightly scheduled lane (08:17 UTC) running `make test-restarts` and `make test-epochs` with `TN_TEST_MDBX_SYNC=durable` set on the two suite steps (step-scoped, not job-wide, so a future "variable must be unset" test never inherits it).  An untimed build step ahead of them absorbs the node-binary build and every workspace test-binary compile and validates the committed lockfile (`cargo fetch --locked`, plus a final `git diff --exit-code -- Cargo.lock` assertion, matching pr.yaml's `--locked` discipline), so the timed steps measure test execution only and the A/B pair stays comparable even when one side runs cold.  Every spawned node inherits the variable through `std::process::Command` (the e2e harness never clears the child env), so no plumbing is needed.  `workflow_dispatch` takes a `sync_mode` choice (`durable` / `safe-no-sync`) so the same lane produces both sides of the A/B measurement on the same runner class.  The lane is scheduled and non-blocking;  PR CI cost is zero.

## Acceptance criteria (from #1149)

- [x] `TN_TEST_MDBX_SYNC=durable` forces `Durable` at `open()` in test builds;  unset keeps `SafeNoSync`;  invalid is a hard error, not a fallback.
- [x] The guard test covers the default mode and the override, with a live read-back of the opened environment.
- [x] Production builds do not read the variable (the read sits inside the cfg gate and release builds compile it out).
- [x] A scheduled CI lane runs the restart and epoch e2e tests with the override set.
- [ ] The measured wall-clock delta is posted on #1149 (first scheduled runs after merge;  compare a `durable` dispatch against a `safe-no-sync` dispatch).

## Verification

- `cargo +nightly fmt -- --check` clean;  `cargo +nightly clippy -p tn-storage -p telcoin-network --features tn-storage/test-utils -- -D warnings` clean (the diff only touches `tn-storage`, and the cfg-gated block enters its lib target under `test-utils`, so this lints every changed line;  selecting `telcoin-network` alongside reproduces the `build-e2e-bin` feature resolution).
- `cargo +1.94 nextest run -p tn-storage` green, including the extended guard test and its three child processes.
- Mutation-confirmed, three mutants, each seen to fail then pass after byte-exact restore: (1) a resolver that ignores the parsed override, (2) a drifted variable-name const, (3) an `open` that ignores the resolver (killed only by the live child read-backs, proving the override genuinely reaches the opened environment).
- Attest proxy: `cargo +1.94 test --no-run --workspace -j 2` green in an isolated target dir.

### DIFF
--- a/.github/workflows/durable-e2e.yaml
+++ b/.github/workflows/durable-e2e.yaml
@@ -31,6 +31,11 @@ on:
           - safe-no-sync
         default: durable
 
+# The job only reads the repository (checkout + build + test); the token gets no
+# write scopes (least privilege, CodeQL actions/missing-workflow-permissions).
+permissions:
+  contents: read
+
 env:
   CARGO_INCREMENTAL: 0
   RUSTFLAGS: "-D warnings -D unused_extern_crates"
--- a/.github/workflows/durable-e2e.yaml
+++ b/.github/workflows/durable-e2e.yaml
@@ -1,0 +1,92 @@
+# Scheduled Durable-mode e2e lane (#1149, follow-up to #982 / #917).
+#
+# PR CI runs the e2e suites with the test-build MDBX default (`SafeNoSync`, no hot-path
+# fsync; see crates/storage/src/mdbx/database.rs). That removed the Durable timing regime
+# -- where the externalization barriers (`persist().await` before a certificate, header,
+# or vote goes on the wire) wait on a real per-commit flush that races the proposer
+# timers -- from every test run. This lane restores that regime on a schedule by
+# exporting TN_TEST_MDBX_SYNC=durable, which every spawned node process inherits, at
+# zero wall-clock cost to PR CI. An invalid value fails MdbxDatabase::open outright, so
+# a typo here cannot green-run the wrong sync regime.
+#
+# The timings of the two suite steps, compared against a `safe-no-sync` dispatch run on
+# the same runner class, give the #917 acceptance measurement (Durable-vs-SafeNoSync
+# wall clock on a CI disk): post the first numbers on #1149. The untimed build step ahead
+# of them absorbs every compile (node binary and all test binaries) plus the cache state,
+# so the timed steps measure test execution only and the A/B pair stays comparable even
+# when one side runs cold.
+name: Durable e2e
+
+on:
+  schedule:
+    # Nightly at 08:17 UTC, off the top of the hour to dodge the scheduled-load spike.
+    - cron: "17 8 * * *"
+  workflow_dispatch:
+    inputs:
+      sync_mode:
+        description: "TN_TEST_MDBX_SYNC value for the spawned nodes"
+        type: choice
+        options:
+          - durable
+          - safe-no-sync
+        default: durable
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-D warnings -D unused_extern_crates"
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
+  CARGO_PROFILE_DEV_DEBUG: 0
+  CARGO_PROFILE_TEST_DEBUG: 0
+
+jobs:
+  durable-e2e:
+    runs-on: ubuntu-latest
+    # Covers a cold cache: full workspace test-binary build plus both e2e suites run
+    # serialized (nextest pins the e2e group to max-threads 1) in the slower Durable regime.
+    timeout-minutes: 240
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 1
+
+      - name: Install Rust toolchain
+        run: rustup show active-toolchain || rustup toolchain install
+
+      - name: Install cargo-nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-nextest
+
+      - name: Cache dependencies
+        uses: Swatinem/rust-cache@f0deed1e0edfc6a9be95417288c0e1099b1eeec3
+        with:
+          shared-key: durable-e2e-cache
+          cache-on-failure: "true"
+
+      # Untimed: absorbs the node-binary build and every workspace test-binary compile, and
+      # validates the committed lockfile (`fetch --locked` fails on drift, matching the
+      # `--locked` discipline of pr.yaml), so the timed suite steps below are compile-free.
+      # TN_TEST_MDBX_SYNC is scoped to the suite steps, not exported job-wide: a build needs
+      # no override, and a future "variable must be unset" test (the
+      # TN_SEED_SIGNATURE_FORK_EPOCH pattern in the Makefile) must not inherit it.
+      - name: Build node binary and test binaries (untimed)
+        run: |
+          cargo fetch --locked
+          make build-e2e-bin
+          cargo nextest run --no-run --locked --workspace
+
+      - name: Run restart e2e tests
+        env:
+          # Scheduled runs get `durable`; a manual dispatch may select either side of the A/B.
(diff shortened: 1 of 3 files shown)


sstanfield (approved):
LGTM!

grantkee (approved):
I like the idea and look forward to the results

PR #1153, opened 2026-08-12
fix(worker): close the observer transaction loss residuals behind forward admission

### LINKED ISSUE #1145: Observer transaction loss residuals after #1132
#1132 made observer seal admission synchronous:  `disburse_txns` returns `Err(BlockSealError::NotValidator)` when a forward is refused, and the batch builder keeps the transactions pending.  Three adjacent gaps remain.  Each was found in review of that fix and left out of its scope on purpose.

## 1. Epoch-close orphan batches ignore the refusal signal

`crates/node/src/manager/node/close_epoch.rs` collects `OurNodeBatchesCache` into a local vec and clears the table.  The non-CVV branch then calls `let _ = worker.disburse_txns(batch.seal(digest)).await;` and discards a refusal.  The transactions are then in no pool and no table, so they are lost with no retry.  This loss path predates #1132:  the old code spawned the forward and dropped the transactions silently under the same conditions.  #1132 makes the refusal observable, but this caller does not use it.  The `is_cvv` branch shows one correct shape:  `recover_raw_transaction` plus `pool.add_recovered_transaction_external`, which hands the transactions to the normal build/seal/retry loop.

## 2. Admission passes on a parsed URL, not a reachable endpoint

`WorkerRpcForwarder::cached_providers` builds a lazy HTTP client (`RootProvider::new_http`) and never dials.  One advertised but dead endpoint therefore admits every batch:  `forward_txns` returns `true`, the builder prunes the transactions as mined, and the background task fails every send with only a `warn!`.  The provider cache only evicts a URL when it leaves the advertisement set, so a stale advertisement repeats the loss for every later batch.  Closing this needs a delivery signal:  evict or demote a provider on send failure, or requeue the batch transactions on `ForwardOutcome::NoEndpointReached`.

## 3. A permanent refusal retries at about 1 Hz with error-level logs and no backoff

A refused seal maps to an empty mined set, one skipped builder iteration (`defer_build`), and a retry on the next `max_batch_delay` tick (default 1 s).  The refusal condition in #1132 (no advertised `worker.rpc`) can persist for a whole epoch.  Each tick re-walks and re-encodes the whole pending set, emits `error!("error while sealing batch")`, and increments the seal-failure metric.  The retry itself is the intended recovery.  The cost is the per-tick log and metric noise plus the pool re-encode.  A refusal-specific backoff and a state-change-gated log line would fix the hygiene without touching the recovery.

MavenRain:
Closes #1145.

Stacked on #1144: this branch is based on `1132-observer-forward-admission-gate` and depends on the synchronous admission signal introduced there.  If #1144 changes under review, this PR rebases;  if it merges first, this PR retargets `main` cleanly.

## Problem

#1132 made observer seal admission synchronous, but three adjacent loss paths stayed open (each found in that review and deliberately left out of its scope):

1. `orphan_batches` (epoch close) collects `OurNodeBatchesCache` into a vec, clears the table, and the non-CVV branch discards the `disburse_txns` result.  A refused disbursal leaves those transactions in no pool and no table.
2. Admission passes on a parsed URL, not a reachable endpoint.  `RootProvider::new_http` never dials, so one advertised but dead endpoint admits every batch: the builder prunes the transactions as mined and the background task fails every send with only a `warn!`.  The provider cache only evicts a URL when it leaves the advertisement set, so a stale advertisement repeats the loss for every later batch.
3. A permanent refusal (no advertised `worker.rpc`, possible for a whole epoch) retries at about 1 Hz: each tick re-walks and re-encodes the whole pending set and emits `error!` plus a metric increment.

No attacker required for any of the three: the default genesis ceremony leaves `worker.rpc` unset, and an operator restart or a stale advertisement produces the dead-endpoint case.  The dial target is still committee-chosen and policy-gated (#1092);  nothing here widens who gets dialed, it only changes what this node does when dialing fails.

## Fix

- [x] `node/src/manager/node/close_epoch.rs`: the non-CVV branch checks the `disburse_txns` result;  on refusal it falls back to the CVV shape through the new shared `repool_batch_txns` helper, so the transactions re-enter the normal build/seal/retry loop instead of vanishing after the table clear.
- [x] `tn-reth/src/forward.rs`: a new `Disposition::Unreachable` marks endpoints that produced no JSON-RPC verdict (send timeout, transport failure, unparseable or null response).  Local faults (`SerError`, `LocalUsageError`) stay `TryNext` and never demote a remote endpoint;  server-answered transients (pool full, internal error) also stay `TryNext`.  The classification match is exhaustive over `RpcError`.
- [x] `tn-reth/src/forward.rs`: an unreachable endpoint is demoted at the failure itself, so batches sealed while the forward task is still walking are already refused at admission, and the same task stops re-dialing it for its remaining transactions (no more one send timeout per transaction against a proven-dead endpoint).  A later delivery through the endpoint lifts the demotion, so one timed-out send against a slow but healthy validator cannot hold it out of admission.  `cached_providers` skips demoted endpoints until `UNREACHABLE_COOLDOWN` (30s) elapses or the endpoint leaves the advertisement set, which keeps the map bounded and probes a recovered endpoint within 30s.
- [x] `tn-reth/src/forward.rs`: every transaction that gets no verdict (unreached, or cut off by the batch budget;  never a considered rejection) is returned to the worker's own pool, per transaction as each verdict lands, so an epoch-boundary abort of the task loses at most the unwalked tail.  The first requeue waits out a one-second grace point (`REQUEUE_GRACE`) that orders it after the builder's prune of the same batch, closing the requeue-then-prune race.  A requeued transaction that was in fact delivered is dropped by the destination pool as a duplicate on the next forward ("already known").
- [x] `tn-reth/src/metrics.rs`: two new counters, `tn_reth.forward_endpoints_demoted_total` and `tn_reth.forwarded_txns_requeued_total`, registered at zero from `ForwarderMetrics::init`;  the requeue counter moves per transaction so an aborted task cannot erase requeues that already happened.  The counter unit test now pins all four forwarder series and their independence.
- [x] `batch-builder/src/lib.rs`: the build channel now carries `BuildOutcome::{Mined, Refused, Failed}`.  `NotValidator` maps to `Refused`, which arms a geometric backoff: base = the configured batch delay, doubling per consecutive refusal, capped at 30s (deliberately equal to the endpoint cooldown, so one recovery probe is not refused against a demotion window that already expired).  The backoff gates build starts, so canonical-state wake-ups cannot bypass it.  Only a seal that actually pruned transactions ends the backoff (an empty build proves nothing about admission), and a stale episode (backoff long expired with no build in between) restarts from the base delay instead of doubling from old state.  Logging is state-change-gated: one `warn!` entering refusal, `debug!` repeats, one `info!` with the refused-seal count on recovery.  The seal-failure metric still counts every refusal, and the pending-pool gauge now refreshes on every loop wake so it stays live while the gate is closed.

## Testing

- `forward_txns_demotes_an_unreachable_endpoint_then_refuses_the_next_batch` (tn-reth): a dead loopback endpoint is admitted once, the task's connection failures demote it, and the next batch is refused.
- `demoted_endpoint_reenters_admission_after_the_cooldown` (tn-reth, paused clock): admission refuses during the cooldown and re-admits after it, deterministically.
- `test_forwarder_metrics_register_at_zero_and_count_separately` (tn-reth): all four forwarder counters register at zero and move independently.
- `forwarder_requeues_undelivered_transactions_into_the_pool` (batch-builder): three signed transactions forwarded to a dead endpoint land back in a real `WorkerTxPool` as pending.
- `refused_seals_back_off_geometrically_and_recover` (batch-builder, paused clock): 100/200/400 ms progression, cap at 30s across 12 refusals, gate holds after a refusal, recovery clears it.
- The #1132 regression test (`test_observer_refused_forward_keeps_txs_in_pool`) still passes against the new backoff.
- Every new regression test was confirmed by mutation (classification, admission skip, requeue, backoff doubling each seen to fail under a targeted mutant and pass after restore).
- Commands: `cargo test -p tn-reth -j 2 --lib` and `cargo test -p tn-batch-builder -j 2 --lib` (all green), plus the attest-order battery on the pinned toolchain: nightly fmt check, `cargo +1.94 clippy --all-targets`, and `cargo +1.94 test --no-run --workspace -j 2` in an isolated target dir.

## Known residuals (reviewed, out of scope)

- Forward tasks run on the per-epoch task manager, so an epoch boundary can still abort one mid-walk.  Per-transaction requeueing shrinks that loss to the unwalked tail at the abort instant (before this PR the whole batch was already lost on this path);  a drain-before-abort at epoch close would remove it entirely and is left as a follow-up.
- A server-answered transient (`-32003` pool full, `-32603`) neither demotes nor refuses, so a committee whose only advertised endpoint has a saturated pool cycles build/forward/requeue at batch cadence.  Nothing is lost (the transactions stay pooled), but a transient-specific backpressure signal is a candidate follow-up.
- Recovery is admission evidence, not delivery evidence: against a permanently dead endpoint, each cooldown expiry admits one probe batch, so the log shows one refusal/recovery pair per ~30s.  That is the intended probe cadence;  acking delivery before the builder prunes stays out of scope, as in #1145's framing.


sstanfield (approved):

grantkee (approved):
LGTM

PR #1154, opened 2026-08-12
Harden the gas penalty against EIP-7702 authorization padding

### LINKED ISSUE #1139: EIP-7702 authorization padding bypasses the gas over-reservation penalty
## Problem

A TN worker packs transactions into a batch under a hard 30,000,000-gas ceiling, but it does not execute them while packing — it only packs. So the only measure of how much of a batch a transaction consumes is its self-declared `gas_limit`; real usage is unknown until after consensus certifies the batch, which is too late to reject an over-reserver. Issue #424 added the countermeasure in `reimburse_caller`: a transaction using at least 10% of its gas limit pays no penalty, but below 10% a penalty grows quadratically with the shortfall, approaching confiscation of the entire unused reservation. Usage is read off `gas.spent()`. The effect is that reserving a batch slot costs about as much as the slot, so setting `gas_limit = 30,000,000` for a 21,000-gas transfer is no longer nearly free.

The penalty silently assumes `gas.spent()` reflects real work that justifies the reservation. EIP-7702 breaks that assumption. Each authorization tuple in a set-code transaction is charged a flat 25,000 gas of intrinsic cost (`PER_EMPTY_ACCOUNT_COST`) from the tuple count alone — before any validity check or signature recovery — for roughly 75 bytes on the wire. Padding the authorization list with about 120 junk tuples (~9 KB) drives `gas.spent()` over 10% of a 30,000,000-gas limit, which collapses the quadratic penalty to near zero. The tuples need not be valid: revm skips a tuple whose chain id does not match before it ever recovers a signature, but the 25,000-gas charge has already landed. The net result is that a sender reserves the full batch while paying a small fraction of the intended penalty — the DoS lever #424 was meant to close, reopened through a dimension the penalty does not account for.

The transaction-type allowlist merged in #1104 admits only legacy, EIP-2930, and EIP-1559 transactions and is enforced at the gateway, the batch builder, and the batch validator, so an EIP-7702 transaction cannot enter a batch and never reaches `reimburse_caller` in production today. The numerator flaw reactivates the moment EIP-7702 is added to the allowlist. It is worth fixing now, and locking with a regression test, so that enabling 7702 support later cannot silently reintroduce the bypass.

## Solution

The penalty basis should exclude the EIP-7702 authorization intrinsic: compute the penalty on `gas.spent()` minus `authorization_count × 25,000`, so a padded authorization list no longer buys penalty-free headroom. Only the authorization intrinsic should be excluded. Calldata is priced at 16 gas per byte, which fairly prices the batch bytes a transaction occupies and is too expensive to abuse for this purpose, and the 21,000-gas base is negligible — so a legitimate low-execution transfer keeps the same basis it has today and sees no behavior change.

Subtracting from the basis alone is unsafe, because it makes the penalty larger, and in the padded case larger than the transaction's unused gas. `reimburse_caller` credits the full penalty to the basefee address while flooring the caller refund at zero, so an uncapped penalty would pay out more than the sender prepaid — minting value in consensus-critical code. The penalty should therefore also be capped at unused gas, which restores exact conservation: refund plus penalty plus used gas equals the gas limit. Because that cap can only bind when the authorization count is nonzero, the change is a provable no-op for every non-7702 transaction, so existing traffic and the existing penalty and balance tests are unaffected. Both steps belong together in the pure penalty-and-refund calculation, leaving the `calculate_gas_penalty` curve itself untouched.

grantkee:
- Exclude the EIP-7702 authorization intrinsic (25,000 gas per tuple) from the over-reservation penalty basis, so a padded authorization list can no longer push `gas.spent()` past the 10% threshold and collapse the penalty. Only the authorization intrinsic is excluded — calldata already prices batch bytes fairly and the 21,000 base is negligible — so no non-7702 transaction changes behavior.
- Cap the penalty at unused gas in the same step. Without the cap, the enlarged penalty can exceed unused gas, and since the handler credits the full penalty to the basefee address while flooring the caller refund at zero, it would pay out more than the sender prepaid. The cap restores exact conservation: `refund + penalty + gas_used == gas_limit`.
- The math moves into one pure `gas_penalty_and_refund` helper; the `calculate_gas_penalty` curve is untouched, so its unit tests and the `economics_props` proptests pass unchanged. The cap only ever binds at a nonzero authorization count, making this a provable no-op for other transaction types.
- Tests: a padding-capable `create_eip7702_with_authorizations` factory; unit tests pinning the padded 120-tuple case (penalty `0 → 26,979,000`, capped at unused gas, refund `0`) and proving zero drift for non-7702 spends; and an end-to-end test that routes a padded 7702 transaction reserving the full 30M batch through `execute_payload_and_update_canonical_chain` (the allowlist would otherwise drop it) and asserts the capped penalty lands in a real block — governance revenue `210,000,000` with the fix versus `21,147,000` without it.

closes #1139


grantkee:
### Padded 7702 authorization list gives free ~1,200x ECDSA-recovery amplification
- **Category**: Security
- **Location**: `crates/tn-reth/src/txn_pool.rs:149` (removed gate), claim at `crates/tn-reth/src/txn_pool.rs:142-148`
- **Claim**: reth calls `recover_authorities` — one secp256k1 recovery per authorization tuple — at `validate/eth.rs:582`, after the balance check at `eth.rs:571-573`. `EthPooledTransaction`'s cost is `max_fee_per_gas * gas_limit + value`, so a tuple-padded transaction with `max_fee_per_gas = 0` has cost 0 and passes the balance check from a zero-balance address. Tuple count is bounded only by intrinsic gas (25,000/tuple) against a 30M block gas limit (~1,199 tuples) and by `max_tx_input_bytes = 128 KB` (~1,400). Validation runs on exactly two blocking workers. `main`'s `.no_eip7702()` rejected type `0x04` at `eth.rs:316-323` inside `validate_stateless`, before any recovery. The comment at `txn_pool.rs:146-148` asserting a set-code transaction "has no analogue of the blob disk-exhaustion vector" is false: the analogue is CPU and RSS, not disk.
- **Key Question**: Does `recover_authorities` run for a transaction with `max_fee_per_gas = 0` from a zero-balance sender — i.e. is `validate_sender_balance` the only economic gate ahead of it?
- **Relevant Files**: `crates/tn-reth/src/txn_pool.rs`, reth `crates/transaction-pool/src/validate/eth.rs`, reth `crates/transaction-pool/src/traits.rs`, reth `crates/transaction-pool/src/validate/constants.rs`
- **Failure scenario**: One throwaway zero-balance key signs a type-`0x04` transaction with `gas_limit = 29_996_000`, `max_fee_per_gas = 0`, `value = 0`, 1,199 tuples of random `(r,s,v)` (~110 KB). Submitted repeatedly over `eth_sendRawTransaction` (the gateway forwards it — allowlisted at `proxy.rs:246`). Each request costs one HTTP POST and ~1,199 recoveries (~50-60 ms) pinned to one of two validation workers. A few requests/second saturate both and stall admission of all legitimate transactions. No funding, no gas ever paid.
- **Source**: tn-review

#### VERDICT: CONFIRMED — High. Three corrections, all making it worse.

Every asserted step holds against pinned reth v1.11.3.
Recovery is `validate/eth.rs:743-748` (`auths.iter().flat_map(|auth| auth.recover_authority())`, no caching or dedup) and is the **last** step of `validate_stateful`, after `validate_sender_balance` at `eth.rs:570`.
Cost is `max_fee_per_gas * gas_limit + value` (`traits.rs:1411-1417`); `validate_sender_balance` (`eth.rs:655-670`) rejects only on `cost > balance`, and `0 > 0` is false.

**Correction A — effective validation concurrency is 1, not 2.**
Both spawned tasks are `.run()` clones of the *same* `ValidationTask`, and `task.rs:62-66` holds the `MutexGuard` across `task.await` (it is a `while let` scrutinee temporary).
The job channel has capacity 1, and every submitter awaits the same lock, so a saturated worker stalls **all** pool admission — RPC, gossip, and forwarder alike.
Raising `additional_validation_tasks` does not help.

**Correction B — the 7-wei base fee makes it worse, not better.**
`MIN_PROTOCOL_BASE_FEE` does reject the transaction, but at `pool/txpool.rs:1997-1999`, inside `TxPool::add_transaction`, called from `pool/mod.rs:592` on an **already-computed** `Valid` outcome.
Recovery has already happened.
Because the transaction is never inserted, the `contains(tx.hash())` dedup at `txpool.rs:750` never sees it — so **the attacker replays identical bytes forever**.
A funded attacker paying 7 wei would be deduped; the free one is not.

**Correction C — JSON-RPC batches bypass the gateway pre-flight.**
`proxy.rs:228` only matches a top-level `"method"`, and reth sets no batch limit, so one 15 MB request carries ~66 such transactions ≈ **79,000 recoveries per single rate-limited request**.
The per-IP limit (100 req/s) meters requests, not tuples, and does not bind.

**Correction D (in the finding's favor, minor)** — `max_tx_input_bytes` *does* bound the authorization list: for non-blob types the check is on `transaction.encoded_length()`, the whole RLP transaction (`eth.rs:362-372`), not `input()`.
It simply is not binding at ~92 B/tuple.

Gates checked and cleared: outer signature (1 recovery, signed once offline and replayed), nonce (`eth.rs:637-652` rejects only `tx_nonce < state_nonce`), outer chain-id (attacker sets it correctly; **authorization-tuple chain IDs are never checked**, so `chain_id = 0` is 1 byte), EIP-7825 `tx_gas_limit_cap` (inert — `osaka_time = None` at `crates/types/src/genesis.rs:74`), `max_tx_gas_limit` (`None`, check at `eth.rs:397-408` inert), `tx_fee_cap` (`cost - value = 0`), `minimum_priority_fee` (`None`, and skipped for `is_local`), per-sender slots (post-validation *and* skipped for `is_local`; RPC submits as `TransactionOrigin::Local`).

**Verifier's own economics**: ~50 us/recovery x 1,199 = ~60 ms CPU per transaction, single-threaded.
Saturating the one effective worker needs ~17 tx/s ~= 1.9 MB/s (~15 Mbps) upload — well under the per-IP limit.
With batching, ~0.25 req/s suffices.
Attacker cost: one key, zero balance, zero gas, one signing pass reused indefinitely.

Not consensus safety — nothing is mined, no state diverges, no disk grows.
It is a sustained, free, remote **availability** attack on transaction admission against every reachable committee validator.

**Proposed fix — one line in `crates/tn-reth/src/txn_pool.rs` beside `.no_eip4844()`:** `.with_max_tx_gas_limit(Some(TN_MAX_TX_GAS_LIMIT))`.
Builder method verified at `reth .../validate/eth.rs:1164-1167`; the check it arms (`eth.rs:397-408`) runs inside `validate_stateless`, **before** any state provider is opened and before `recover_authorities`.
At 3,000,000 the tuple ceiling is 119 (mainnet 7702 transactions carry 1) and per-transaction recovery work drops ~10x.
Blunt: it caps all transactions, so pick it above TN's heaviest legitimate call.
Defense in depth: `.with_max_tx_input_bytes(32 * 1024)` (`eth.rs:1142`), the *first* size check in `validate_stateless`, capping tuples at ~350.

**No upstream knob exists for authorization-list length** — the verifier grepped reth's entire `transaction-pool`; the only `authorization_list().len()` is the gas calculation at `eth.rs:1370`.
A precise cap needs a thin `TransactionValidator` wrapper rejecting `ty() == 0x04 && authorization_list().len() > MAX_AUTH_TUPLES` before delegating, which means moving `WorkerTxPool` off the `EthTransactionPool` alias / `Pool::eth_pool` onto `Pool::new`.
**Do not** use `with_minimum_priority_fee` (never runs for `is_local`) or raise `additional_validation_tasks` (the `task.rs` guard serializes workers regardless).

grantkee:
### 4. Recovered authority addresses are interned permanently; unbounded RSS growth
- **Category**: Memory & Safety
- **Location**: `crates/tn-reth/src/txn_pool.rs:149`; reth `crates/transaction-pool/src/pool/mod.rs:590`, `crates/transaction-pool/src/identifier.rs:31-38`
- **Claim**: Every recovered authority address is interned via `authority_ids: authorities.map(|auths| self.get_sender_ids(auths))` → `sender_ids_or_create`, which inserts into both `address_to_id` and `sender_to_address`. `identifier.rs` contains no `remove`, `clear`, `retain`, or `shrink`, so entries survive eviction, mining, and every canonical update for the process lifetime. Pre-7702, growth was one entry per distinct sender; a padded authorization list multiplies that by up to ~1,199 per transaction on the same cost-0 path as finding 1. By contrast the blob vector `.no_eip4844()` closes *is* reclaimed on restart — `DiskFileBlobStore::open` calls `inner.delete_all()`.
- **Key Question**: Does any code path in reth v1.11.3 remove entries from `SenderIdentifiers`, or is `remove_auths` (`pool/txpool.rs:1786-1796`) the only cleanup — which prunes the `auths` map but not the interner?
- **Relevant Files**: reth `crates/transaction-pool/src/identifier.rs`, reth `crates/transaction-pool/src/pool/mod.rs`, reth `crates/transaction-pool/src/pool/txpool.rs`
- **Failure scenario**: Attacker streams cost-0 padded 7702 transactions at 1/second, each with 1,199 fresh random authority addresses. RSS grows ~1,199 x 2 hash-map entries (~64 B each) ~= 150 KB/s ~= 13 GB/day, monotonically, reclaimed only by restart.
- **Source**: tn-review

#### VERDICT: CONFIRMED — upgraded to High. One of my framings was backwards.

Interning path confirmed: `validate/eth.rs:582` → `pool/mod.rs:590` → `sender_ids_or_create` (`pool/mod.rs:219`) → inserts into **both** maps (`identifier.rs:31-38`).
`recover_authorities` (`eth.rs:743-747`) applies no chain-id, nonce, or validity filter — every ECDSA-recoverable tuple yields an interned address.

**Stronger than filed**: interning at `pool/mod.rs:590` happens **before** `pool.add_transaction` at `pool/mod.rs:592`.
A transaction the pool itself rejects — nonce gap, `ExceededAccountSlots`, underpriced, pool full — has **already interned all its authorities**.
The attacker never needs the transaction accepted, let alone mined.

No removal path: `SenderIdentifiers` has four methods, none of them `remove`/`clear`/`retain`/`shrink`/`drain`, and no capacity bound.
Exhaustive grep for `address_to_id`/`sender_to_address` across all of reth returns only `identifier.rs:14-35`.
`remove_auths` (`pool/txpool.rs:1786-1798`) prunes only the separate `auths` map; `TxPool` does not own the interner (it lives on `PoolInner`), so eviction, mining, and `discard_worst` cannot reach it.

**Epoch reset refuted** — the most promising escape, and it fails.
`crates/node/src/engine/mod.rs:194-196` states it outright: "A worker's components (RPC server + transaction pool) are created once on the node's first epoch and are not torn down across epoch transitions" (`crates/node/src/manager/node/start_epoch.rs:459-470` takes the `respawn_worker_network_tasks` branch thereafter).
The pool, and its interner, is process-lifetime.

**Correction to my own framing.**
I wrote that blob sidecars are reclaimed on restart "while this interner is not" — that asymmetry does not exist.
The interner is an in-memory `HashMap`; it is also destroyed on restart, and my phrasing contradicted the finding's own headline.
The correct analogy is the one TN's code already makes at `txn_pool.rs:129-139`: both grow without bound **at runtime** with no reclamation path, and TN judged that severe enough to refuse blobs outright.
Also worth stating: upstream reth has the identical leak, so TN's exposure is **not** caused by skipping `maintain_transaction_pool` — that loop never touched `SenderIdentifiers` either.

**Verifier's own arithmetic** (rejecting my point estimate): `AddressMap` is a hashbrown SwissTable; `address_to_id` and `sender_to_address` bucket pairs cost 66 B, and at hashbrown's 43.75-87.5% live occupancy that is **75-151 B per live address**, ~105 B mid-cycle.
At 1,199 tuples x 86,400 s = 103,593,600 addresses/day → **~8-16 GB/day, centered ~11 GB/day**, plus a transient ~2x spike at each doubling rehash (itself an OOM trigger at multi-GB sizes).
My 13 GB/day sits inside the band; the range is the honest number.

**Proposed fix**: the interner is upstream and unpatched, so prevent `TransactionValidationOutcome::Valid { authorities: Some(..) }` from being produced for a padded list.
(1) Wrap the validator built at `crates/tn-reth/src/txn_pool.rs:125-153` with a thin `TransactionValidator` rejecting `authorization_list().len() > N` (N ~= 16-32) *before* delegating — `authorities` is only populated on the `Valid` arm, so nothing is interned; cuts per-transaction interning 40-75x.
(2) Cumulative growth is still unbounded: upstream a refcounted/GC-able `SenderIdentifiers`, or intern lazily since `auths` is the only consumer and is already pruned by `remove_auths`.
(3) Observability gap — no metric today would reveal this growing in production; add a gauge on `identifiers` length.

grantkee:
### 5. The honest-7702 e2e test cannot fail; its stated mutation claim is false
- **Category**: Consensus Safety
- **Location**: `crates/engine/tests/it/main.rs:2767` (doc claim `:2758-2765`, assertion `:2905-2909`); mirror at `crates/tn-reth/src/evm/utils.rs:506`
- **Claim**: `test_honest_7702_sender_pays_no_penalty` uses `TX_GAS_LIMIT = 146_000`. Because `146_000 <= MIN_GAS_LIMIT_THRESHOLD (210_000)`, `calculate_gas_penalty` returns 0 on its first guard for every intrinsic-subtraction variant, so the penalty is 0 with the two-argument subtraction, with numerator-only subtraction, and with no subtraction at all. The doc comment's claim that "re-inflating the basis assesses a penalty and changes governance revenue, breaking the 817,600 pin" is false. `honest_small_delegation_exact_estimate_pays_no_penalty` pins the identical shape and is equally insensitive.
- **Key Question**: With `gas_limit = 146_000`, does any of `gas_penalty_and_refund(146_000, 146_000, 116_800, A)` for `A` in `{125_000, 0}` — or the numerator-only variant `calculate_gas_penalty(146_000, 21_000)` — return a non-zero penalty?
- **Relevant Files**: `crates/engine/tests/it/main.rs`, `crates/tn-reth/src/evm/utils.rs`
- **Failure scenario**: Computed: both-args `(0, 29_200)`; `A = 0` → `(0, 29_200)`; numerator-only → `0`. Deleting the `gas_spent.saturating_sub(auth_intrinsic)` argument from `gas_penalty_and_refund` — reverting the entire change this branch exists to make — leaves this e2e test green. A sensitive shape does exist at unit level (`honest_exact_estimate_with_capped_refund_pays_no_penalty`, N=20/521,000: numerator-only yields 104,200 vs the pinned 0), so the gap is in the e2e layer and the 146,000 mirror. Raising `NUM_AUTHS` to 20 makes the limit 521,000 and the test load-bearing.
- **Source**: tn-review

#### VERDICT: CONFIRMED — Medium. All five assertions are insensitive.

Guard is `<=` against 210,000 (`utils.rs:22,61-65`); the handler passes the raw transaction `gas_limit` as the first argument (`handler.rs:166`).
For `gas_penalty_and_refund(146_000, 146_000, 116_800, 125_000)`, `unused_gas = 29_200` in all variants: production reduces the limit to 21,000, the other two leave it at 146,000 — and **146,000 <= 210,000 either way**, so every variant returns `(0, 29_200)` on the first guard before any ratio is computed.
A swapped spent/used mutant also returns 0.

The verifier read the whole body (`main.rs:2767-2967`) and found every assertion insensitive: `block.gas_used == 116_800` (revm's post-refund gas; `reimburse_caller` only moves balances), the `gas_penalty_and_refund` helper pin, governance revenue `817_600` (= 7 x 116,800 exactly, penalty 0 in every variant), `sender_delta == 817_600`, and the authority nonce/delegated-code checks (which guard batch admission and revm tuple application, not the penalty basis).
The doc comment at `main.rs:2761-2763` — "Swapping spent/used or re-inflating the basis assesses a penalty and changes governance revenue, breaking the 817,600 pin below" — is factually false; neither mutation moves any pinned value.
The unit mirror at `utils.rs:506` is equally insensitive.

**Correction to my proposed fix.**
I suggested `NUM_AUTHS = 20`; the verifier derived **`NUM_AUTHS = 12`** as the minimum that works, with parameters checked: `TX_GAS_LIMIT = 21,000 + 12 x 25,000 = 321_000` (clears 210,000); refund `12 x 12,500 = 150,000` capped by EIP-3529 at `321,000/5 = 64,200` → `EXPECTED_GAS_USED = 256_800`; governance pin becomes `7 x 256,800 = 1_797_600`.
Production still yields 0 (reduced limit 21,000 → guard), while the numerator-only mutant now gets `calculate_gas_penalty(321_000, 21_000)` → usage 65,420,560 < 10^8 → **penalty 35,872 != 0**, so the test fails under the mutant.

**Structural limit worth knowing**: an exact-estimate honest case can *never* catch two mutants.
`A = 0` gives `gas_spent == gas_limit` → 100% usage → 0 by construction (already covered by `test_padded_7702_authorization_list_pays_near_maximal_penalty` at `main.rs:2644`, which asserts `gas_used > GAS_LIMIT / 10` at `:2690-2692` precisely so the pre-fix path computes 0 against the pinned 26,560,959).
Catching the spent/used swap needs non-auth work above 210,000 *and* a capped refund above 90% of it, forcing N >= 30 plus substantial calldata — a redesign, not a tweak.
So `NUM_AUTHS = 12` closes the one gap the doc comment names, and **the doc comment must be corrected regardless** to say the pin guards numerator-only inflation and to point at the padded test for the `A = 0` direction.

Incidental: `honest_exact_estimate_with_capped_refund_pays_no_penalty` (`utils.rs:493`) has the same defect on the production path (`521,000 - 500,000 = 21,000` short-circuits) but *does* catch the numerator-only mutant, so it is not fully vacuous.

grantkee:
### 6. Penalty basis silently becomes post-refund gas when the EIP-7623 floor binds over a refund
- **Severity (initial)**: Low
- **Category**: State & Funds
- **Location**: `crates/tn-reth/src/evm/handler.rs:113`, doc claim at `handler.rs:86-99`
- **Claim**: `revm-handler-15.0.0/src/post_execution.rs:12-20` enters the floor branch when `spent_sub_refunded() < floor_gas` and calls `Gas::set_spent(floor_gas)`, lowering `gas.spent()` whenever `floor_gas < spent`. That condition is `0 < spent - floor_gas < refunded`, reachable for any refunding transaction. The handler's `gas.spent()` read therefore returns a floor-clamped, post-refund value, contradicting the doc's "penalty is computed from pre-refund gas".
- **Key Question**: For `spent = 300_000, refunded = 60_000, floor_gas = 250_000`, does `exec_result.gas().spent()` return 300,000 or 250,000 inside `reimburse_caller`?
- **Relevant Files**: `crates/tn-reth/src/evm/handler.rs`, `revm-handler-15.0.0/src/post_execution.rs`, `revm-interpreter-32.0.0/src/gas.rs`
- **Failure scenario**: Non-7702 transaction, `gas_limit = 30_000_000`, heavy zero calldata putting `floor_gas = 250_000`, pre-refund spend 300,000, 60,000 SSTORE refund. Handler sees `gas_spent = 250_000` → penalty ~24.99M instead of ~24.06M: ~0.93M gas of extra confiscation. Deterministic on every node, so an over-charge rather than a fork. Pre-existing on `main`; flagged because the new clamp reasons about `gas_spent` as if it were the true pre-refund value.
- **Source**: tn-review

grantkee:
### 7. Published floor-clamp recipe is conditional; the node applies it unconditionally
- **Severity (initial)**: Low
- **Category**: State & Funds
- **Location**: `docs/gas-penalty.md:261-266`, `docs/gas-penalty.md:200-203`, `docs/gas-penalty.md:275-279`
- **Claim**: `reimburse_caller` (`crates/tn-reth/src/evm/handler.rs:157-163`) applies `effective_auth_intrinsic(A, gas_spent, floor_gas) = A.min(gas_spent - floor_gas)` for every 7702 transaction whenever EIP-7623 is enabled, not only when the floor binds. The doc instructs integrators to subtract the full `25_000 * N` and use `min(...)` only "if the transaction carries enough calldata for the EIP-7623 floor to bind". Whenever `0 < gas_spent - floor_gas < 25_000 * N` the floor does not bind yet the clamp still reduces `A`. The doc also contradicts itself between line 185 (correct universal form) and lines 200-203/261-266.
- **Key Question**: For `gas_spent = 226_000`, `floor_gas = 221_000`, `A = 25_000` (floor does not bind), does the handler subtract 25,000 or 5,000?
- **Relevant Files**: `docs/gas-penalty.md`, `crates/tn-reth/src/evm/handler.rs`, `crates/tn-reth/src/evm/utils.rs`
- **Failure scenario**: One tuple, 20,000 zero calldata bytes, 100,000 gas execution, `gas_limit = 2,260,000`. `floor_gas = 221,000`; spent 226,000 > floor, so floor does not bind. Node: `A_eff = min(25,000, 5,000) = 5,000` → penalty 809 gas. Wallet following step 3: full 25,000 → penalty 20,613 gas — a 25x fee over-estimate.
- **Source**: tn-review

MavenRain (commented):
Reviewed at f78af76f with a local `cargo +1.94 check --all-targets` pass over the touched crates (green).  The penalty math checks out: the clamp keeps the numerator at `max(spent - A, floor)` in both branches of the 7623 floor check, and the cap restores exact gas conservation.  Three inline comments below: one on the activation path for 0x04 admission, two on test rigor.

crates/types/src/lib.rs
```diff
@@ -74,15 +75,61 @@ pub use alloy::{
...
+/// binary — one without the two-argument authorization-intrinsic subtraction —
+/// computes a different post-execution balance for the same certified batch
+/// and forks state at the live edge. Chain history is clean (no 7702
+/// transaction has ever been certified), so there is no replay risk; the risk
+/// window is only a mixed-version fleet during rollout. Every validator must
+/// run a binary at or beyond this change before the first 7702 transaction is
+/// admitted.
 pub fn batch_allowlisted_tx_type<T: Typed2718>(tx: &T) -> bool {
-    tx.is_legacy() || tx.is_eip2930() || tx.is_eip1559()
+    tx.is_legacy() || tx.is_eip2930() || tx.is_eip1559() || tx.is_eip7702()
```
MavenRain, line 100: This line turns on type 0x04 admission for the whole network, with the LOCKSTEP note above as the only protection.  A mixed-version fleet fails in two ways during rollout.  An old validator rejects any batch that carries a 0x04 transaction with `UnsupportedTxType`, which penalizes the honest upgraded proposer.  Since every TN chain activates Prague at genesis, an old binary executes a certified 0x04 transaction with the old penalty math and forks state silently.  Would a protocol-level gate (fork height or chain-config flag) make sense here, so admission fails loudly on an old binary instead of relying on operational lockstep?  #1247 proposes the same shape of gate for a consensus-visible change.

crates/tn-reth/tests/it/economics_props.rs
```diff
@@ -4,14 +4,35 @@
...
 const MIN_GAS_LIMIT_THRESHOLD: u64 = 210_000;
 /// 10% threshold for penalty calculation
 const USAGE_THRESHOLD_PCT: u64 = 10;
+/// Production batch gas limit (`tn_types::max_batch_gas`) — the most gas a single batch can grant.
+/// The anti-padding invariant below is exact only up to this bound: the penalty's derivative in the
+/// gas limit is `1 - 120r^2 + 200r^3` for `r = real_work / gas_limit`, which approaches 1 as
+/// `r -> 0`, so the real-valued margin approaches zero and the floor division in
+/// `usage_ratio_scaled` decides by one gas. Sparse one-gas inversions exist at limits in the low
+/// 40-millions and beyond — see `padding_is_one_gas_cheaper_above_the_batch_gas_limit`.
+const MAX_BATCH_GAS: u64 = 30_000_000;
```
MavenRain, line 32: This is a hardcoded copy of the 30M limit.  The compile-time asserts in this file compare against this local constant, not against `tn_types::max_batch_gas`, so raising the real cap would pass here silently.  `max_batch_gas(_epoch)` is not const, so a test-time `assert_eq!(MAX_BATCH_GAS, max_batch_gas(0))` would tie the copy to the source of truth.

crates/tn-reth/tests/it/economics_props.rs
```diff
@@ -152,3 +440,66 @@ fn test_quadratic_scaling() {
...
+///
+/// This locks the exact-estimate-pays-zero contract against any future "claw back the applied-tuple
+/// lever" formula change. Every formula in the `A - refunded` family — excusing only the unrefunded
+/// portion of the intrinsic, or scaling `A` by the refund ratio — provably breaks this lock on a
+/// large fraction of the domain, because the refund lands in the `gas_spent - gas_used` gap that
+/// the penalty basis never sees. That impossibility is why report finding 2a was accepted and
+/// pinned rather than fixed; if this sweep ever fails, the accepted tradeoff has been traded for a
+/// worse one.
+#[test]
+fn pure_delegation_exact_estimates_pay_no_penalty_for_every_tuple_count() {
```
MavenRain, line 486: With `gas_limit == gas_spent == s`, any formula that subtracts the same clamped value from both `calculate_gas_penalty` arguments produces a 100% usage ratio and zero penalty.  That includes the `A - refunded` claw-back variants that the doc in utils.rs says this lock refutes; the sweep only discriminates against numerator-only subtraction.  The refund element of the tuple is also true by construction rather than an independent oracle.  Sweeping with `gas_limit > gas_spent` slack would make the lock bind, or the doc claim could be narrowed to what the sweep actually pins.

MavenRain (dismissed):
Worth having a look at the comments, but nothing I'd block

grantkee:
@MavenRain can you look into a stacked PR for this to address some of these security concerns at the RPC level? I'm thinking we might be able to wrap the RPC logic and enforce our own auth list len - need to update batch validator too

PR #1161, opened 2026-08-12
fix(tn-reth): apply node_config.rpc.eth_config() when building the EthApi

### LINKED ISSUE #1156: RPC: node_config.rpc.eth_config() is not applied when the EthApi is built
## Summary

`RethEnv::get_rpc_server` builds the `eth` API with `EthApi::builder(...).build()`.  It does not apply `node_config.rpc.eth_config()`.  The CLI exposes five eth-config flags, and the node parses them, but the core `EthApi` ignores them.  Reth's own launch path applies these values.  See `EthApiCtx::eth_api_builder` in reth v1.11.3, `crates/node/builder/src/rpc.rs:1180-1196`.

## Problem

`crates/tn-reth/src/lib.rs:1279-1285`:

```rust
let eth_api = EthApi::builder(
    self.inner.blockchain_provider.clone(),
    transaction_pool,
    network,
    self.inner.evm_config.clone(),
)
.build();
```

`.build()` keeps the `EthConfig` defaults.  The neighbor lines do consume `node_config.rpc`: line 1278 for module selection (`transport_rpc_module_config()`) and line 1298 for the server config (`rpc_server_config()`).  Only the eth-config values are dropped.

TN defines its own `RpcServerArgs` subset (`crates/tn-reth/src/rpc_server_args.rs:50-193`), converts it into reth's `RpcServerArgs` (`impl From`, line 230), and stores it in `NodeConfig.rpc`.  `RethCommand` flattens this subset (`crates/tn-reth/src/lib.rs:103,263-275`).  The subset keeps five eth-config flags.  Each one parses, then has no effect:

- `--rpc.gascap` (`rpc_server_args.rs:148-156`): the `eth_call` / `eth_estimateGas` cap stays at the default 50M
- `--rpc.max-simulate-blocks` (`:168-174`): `eth_simulateV1` stays at the default 256
- `--rpc.eth-proof-window` (`:176-184`): `eth_getProof` keeps the default window of 0 (latest state only)
- `--rpc.proof-permits` (`:186-188`)
- `--rpc.state-cache.*` (`:190-192`): the eth state cache keeps the default sizes

These flags are not affected, because they reach the handlers through `transport_rpc_module_config()`, which embeds `RpcModuleConfig::new(self.eth_config())`:

- `--rpc.max-tracing-requests` (blocking task guard, reth `crates/rpc/rpc-builder/src/lib.rs:540`)
- `--rpc.max-trace-filter-blocks` (`TraceApi`, reth `crates/rpc/rpc-builder/src/lib.rs:789-795`)
- `--rpc.max-blocks-per-filter`, `--rpc.max-logs-per-response` (`EthFilter`, reth `crates/rpc/rpc-builder/src/eth.rs:29`)

The remaining `EthConfig` knobs (gas price oracle, pending block kind, EVM memory limit, max blocking IO requests) have no TN CLI flag, so they always keep their defaults.

## Impact

A run without these flags does not change: the clap defaults and the `EthConfig` defaults come from the same reth constants.  The problem hits operators who set one of the five flags.  The node accepts the flag and shows no warning.  Example: with `--rpc.gascap 100000000` the node starts, but `eth_estimateGas` still caps at 50M gas.

## Suggested fix

Mirror `EthApiCtx::eth_api_builder` (reth v1.11.3, `crates/node/builder/src/rpc.rs:1180-1196`) in `get_rpc_server`:

```rust
let eth_config = self.inner.node_config.rpc.eth_config();
let eth_api = EthApi::builder(
    self.inner.blockchain_provider.clone(),
    transaction_pool,
    network,
    self.inner.evm_config.clone(),
)
.gas_cap(eth_config.rpc_gas_cap.into())
.max_simulate_blocks(eth_config.rpc_max_simulate_blocks)
.eth_proof_window(eth_config.eth_proof_window)
.eth_state_cache_config(eth_config.cache)
.fee_history_cache_config(eth_config.fee_history_cache)
.proof_permits(eth_config.proof_permits)
.gas_oracle_config(eth_config.gas_oracle)
.max_blocking_io_requests(eth_config.max_blocking_io_requests)
.pending_block_kind(eth_config.pending_block_kind)
.raw_tx_forwarder(eth_config.raw_tx_forwarder)
.evm_memory_limit(eth_config.rpc_evm_memory_limit)
.build();
```

`node_config.rpc.eth_config()` picks up the five supported flags and returns defaults for the knobs TN does not expose, so the full chain is safe.  `EthApiBuilder` has each setter (reth `crates/rpc/rpc/src/eth/builder.rs:283-368,587`), including `eth_state_cache_config`, so no manual cache construction is needed.  An alternative is to review each value and pin TN-specific choices, in the same style as the `NetworkArgs` block in `RethConfig::new`.

## Credit

Reported by Haxatron during the collaborative review.

MavenRain:
Closes #1156.  (Cantina #14)

## Problem

`RethEnv::get_rpc_server` builds the `eth` API with `EthApi::builder(...).build()` and never applies `node_config.rpc.eth_config()`. The neighbor lines do consume `node_config.rpc` (module selection via `transport_rpc_module_config()`, server limits via `rpc_server_config()`), so only the eth-config values were dropped. The five eth-config flags TN exposes parse cleanly and then silently do nothing:

- `--rpc.gascap` . . . `eth_call` / `eth_estimateGas` stay capped at the default 50M
- `--rpc.max-simulate-blocks` . . . `eth_simulateV1` stays at the default 256
- `--rpc.eth-proof-window` . . . `eth_getProof` keeps the default window of 0
- `--rpc.proof-permits` . . . concurrent proof permits keep the default
- `--rpc.state-cache.*` . . . the eth state cache keeps the default sizes

These flags are resource bounds on the public RPC surface. No attacker is required: an operator who lowers a bound to harden a public endpoint (or raises one for an internal tool) gets the reth default instead, with no warning. The filter/pubsub half of the config was never affected, because `transport_rpc_module_config()` already embeds `RpcModuleConfig::new(self.eth_config())`; only the `EthApi` itself was unconfigured.

## Fix

- [x] `crates/tn-reth/src/env/rpc.rs`: add `apply_eth_config`, a mirror of reth's own `EthApiCtx::eth_api_builder` (reth v1.11.3, `crates/node/builder/src/rpc.rs:1180-1196`), and chain it into `get_rpc_server` between `EthApi::builder(...)` and `.build()`. The config source is `self.node_config().rpc.eth_config()`, the same call the module-selection line already uses, so the five supported flags flow through and every knob TN does not expose keeps its reth default (identical values to today).
- [x] The mirror applies the full `EthConfig`, not only the five exposed flags: `gas_cap`, `max_simulate_blocks`, `eth_proof_window`, `eth_state_cache_config`, `fee_history_cache_config`, `proof_permits`, `gas_oracle_config`, `max_batch_size`, `max_blocking_io_requests`, `pending_block_kind`, `raw_tx_forwarder`, `send_raw_transaction_sync_timeout`, `evm_memory_limit`, `force_blob_sidecar_upcasting`. Three deltas against reth's reference, called out in the doc comment: the state cache goes in as `eth_state_cache_config` (reth passes a pre-spawned shared cache; TN has no other cache consumer, so `build()` spawning one is equivalent), `send_raw_transaction_sync_timeout` is applied here though reth's reference skips it (no CLI flag feeds it yet, so it re-applies the default), and TN keeps the builder's default task executor where reth injects its own (same behavior as before this fix). The extras are no-ops today (the values equal the defaults) and keep this call site correct if TN exposes more flags later.

## Testing

- [x] `test_apply_eth_config_overrides_every_builder_default`: builds an `EthApiBuilder` over a temp `RethEnv`, applies an `EthConfig` whose asserted fields all differ from the defaults, and reads each value back through the builder's getters (gas cap, simulate blocks, proof window, state cache, fee history cache, proof permits, gas oracle, batch size, pending block kind, raw tx forwarder). `max_blocking_io_requests`, `send_raw_transaction_sync_timeout`, `evm_memory_limit`, and `force_blob_sidecar_upcasting` have no builder getter, so only the setter calls cover them.
- [x] `test_tn_rpc_args_reach_eth_config`: drives the five TN CLI flags through the TN-args-to-reth-args conversion into `eth_config()` and asserts each lands in the `EthConfig`.
- [x] Both tests confirmed by mutation: reverting a setter in `apply_eth_config` fails the first test; the second fails if a flag is dropped from the conversion.
- [x] `cargo +1.94 test -p tn-reth` green on the pinned toolchain.
- [x] attest proxy battery: `cargo +nightly fmt -- --check`, `cargo +1.94 clippy --all-targets`, and `cargo +1.94 test --no-run --workspace -j 2` in an isolated target dir, all green.
- [x] Default-equivalence audit: for every field `apply_eth_config` applies, `EthConfig::default()` equals the builder's own default (checked per field against the reth v1.11.3 source), so a node run without `--rpc.*` flags gets a byte-identical `EthApi` before and after this change.
- Known limit: the two tests cover the helper and the config chain; the single wiring expression in `get_rpc_server` is covered by review, not by test. A testable seam would need a custom `NodeConfig` through the test env constructors . . . happy to do that as a follow-up if wanted.

## Out of scope

- `--rpc.txfeecap` is the same bug class on the pool side (parses, then nothing applies it; reth wires it only in its `PoolBuilder` launch path, which TN does not run). Filed separately so this PR stays a mirror of the reth eth-API path.
- Injecting `RethEnv`'s task spawner into the `EthApi` builder (reth's reference does; TN today keeps the builder's default executor for the fee-history cache task). Behavior-neutral for this fix, worth its own decision.

## Credit

Reported by Haxatron during the collaborative review (#1156).

### DIFF
--- a/crates/tn-reth/src/env/rpc.rs
+++ b/crates/tn-reth/src/env/rpc.rs
@@ -2,8 +2,8 @@
 //!
 //! TN assembles reth's RPC builder stack over the worker's transaction pool and the
 //! [`WorkerNetwork`] shim; there is no engine namespace because consensus drives execution
-//! directly. Namespace exposure and transport limits come from the CLI allowlist
-//! (`crate::cli`) via `node_config().rpc`.
+//! directly. Namespace exposure, transport limits, and the `eth` API knobs (gas cap, proof
+//! window, cache sizes) come from the CLI allowlist (`crate::cli`) via `node_config().rpc`.
 //!
 //! Note: [`RethEnv::get_rpc_server`] swallows a failure to merge the TN-specific module at
 //! `error!` level — the server still starts and serves the standard namespaces with the TN
@@ -14,9 +14,11 @@ use std::sync::Arc;
 use jsonrpsee::Methods;
 use reth::rpc::{
     builder::{config::RethRpcServerConfig as _, RpcModuleBuilder, RpcServerHandle},
-    eth::EthApi,
+    eth::{EthApi, EthApiBuilder},
 };
 use reth_provider::providers::BlockchainProvider;
+use reth_rpc_eth_api::RpcNodeCore;
+use reth_rpc_eth_types::EthConfig;
 use reth_transaction_pool::{blobstore::DiskFileBlobStore, EthTransactionPool};
 
 use crate::{
@@ -26,6 +28,41 @@ use crate::{
     worker::WorkerNetwork,
     RethEnv, RpcServer, WorkerTxPool,
 };
+
+/// Apply the operator's [`EthConfig`] values to the `eth` API builder.
+///
+/// Mirrors `EthApiCtx::eth_api_builder` (reth v1.11.3, `crates/node/builder/src/rpc.rs`),
+/// which TN's launch path does not run. Without this, each `--rpc.*` eth flag parses and
+/// then silently keeps the reth default (#1156).
+///
+/// Three deltas against reth's reference:
+/// - cache: TN spawns no shared state cache up front, so the configured sizes go in as
+///   `eth_state_cache_config` for `build` to spawn from. Equivalent for TN, which has no other
+///   consumer of the cache.
+/// - `send_raw_transaction_sync_timeout`: applied here, absent from reth's reference. No CLI flag
+///   feeds it yet, so it re-applies the default (forward compatibility).
+/// - task spawner: reth injects its own; TN keeps the builder's default executor, the same behavior
+///   as before this fix.
+fn apply_eth_config<N: RpcNodeCore, Rpc, NextEnv>(
+    builder: EthApiBuilder<N, Rpc, NextEnv>,
+    config: EthConfig,
+) -> EthApiBuilder<N, Rpc, NextEnv> {
+    builder
+        .gas_cap(config.rpc_gas_cap.into())
+        .max_simulate_blocks(config.rpc_max_simulate_blocks)
+        .eth_proof_window(config.eth_proof_window)
+        .eth_state_cache_config(config.cache)
+        .fee_history_cache_config(config.fee_history_cache)
+        .proof_permits(config.proof_permits)
+        .gas_oracle_config(config.gas_oracle)
+        .max_batch_size(config.max_batch_size)
+        .max_blocking_io_requests(config.max_blocking_io_requests)
+        .pending_block_kind(config.pending_block_kind)
+        .raw_tx_forwarder(config.raw_tx_forwarder)
+        .send_raw_transaction_sync_timeout(config.send_raw_transaction_sync_timeout)
+        .evm_memory_limit(config.rpc_evm_memory_limit)
+        .force_blob_sidecar_upcasting(config.force_blob_sidecar_upcasting)
+}
 
 impl RethEnv {
     /// Build and return the RPC server for the instance.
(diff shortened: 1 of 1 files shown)


sstanfield (approved):

grantkee (approved):
LGTM!

Issue #1067, opened 2026-07-29
BatchValidator is fork-blind and denylists only EIP-4844, so type-0x04 (EIP-7702) transactions certify on mainnet and are silently dropped at execution


MavenRain:
**Suggested title:** `BatchValidator` is fork-blind and denylists only EIP-4844, so type-0x04 (EIP-7702) transactions certify on mainnet and are silently dropped at execution

**Severity:** Low / Defense-in-depth and observability (**no halt, no fork, no state divergence**).  The benign outcome is real but it is not enforced by the validator;  it rests on an upstream error-classification detail in `alloy-evm`.
**Attacker model:** A Byzantine (or key-compromised) member of the current on-chain committee.  Every unprivileged route is closed: the RPC/pool path is Prague-gated by stock reth, and `process_report_batch` rejects any reporter outside the committee with `NonCommitteeBatch`.

## Summary
`BatchValidator::validate_batch` never consults the chain spec or the fork schedule.  Its only transaction-type discriminator is `tx.is_eip4844()`.  On mainnet, whose genesis declares `shanghaiTime: 0` and no `cancunTime`/`pragueTime`, EIP-7702 is inactive, yet a batch containing a type-0x04 transaction passes all eight of the validator's checks, is voted on, and is certified.  At execution revm rejects it with `Eip7702NotSupported`, and that rejection lands in the executor's explicit skip-and-continue arm: the node logs `warn!("skipping invalid transaction")`, increments `invalid_txs_skipped_total`, and finalizes the block without the transaction.

So the transaction is dropped rather than executed, which is consistent with the drop-rather-than-halt stance documented for `unrecoverable_txs_dropped_total` (issues #933 / #938).  The gap is that the drop happens two layers below where it should, in a counter that is explicitly documented as *not* an alert signal, and only because of how a third-party crate happens to classify the error.

## Mechanism (verified end to end)
1. **Mainnet is pre-Prague.** `chain-configs/mainnet/genesis.yaml:3,15` declares chainId 487 with `shanghaiTime: 0` and no cancun/prague key, and `Config::load_mainnet` deserializes that file verbatim (`crates/config/src/node.rs:155-156`).  The Prague-enabling default `prague_time = Some(0)` at `crates/types/src/genesis.rs:73` is reachable only from `test_genesis` and the genesis-ceremony CLI, never from the mainnet load path.  Testnet (chainId 2017) does carry cancun and prague, so this is mainnet-only.
[...]

PR #1169, opened 2026-08-13
fix(tn-reth): reject blob and set-code txns at pool admission (#1159)

### LINKED ISSUE #1159: DoS: unbounded on-disk blob-sidecar growth (blob store never cleaned under TN pool maintenance)
## Summary

A remote, unprivileged user can exhaust a validator node's disk and crash it. The transaction pool admits EIP-4844 (blob) transactions and writes each blob sidecar to the on-disk `DiskFileBlobStore`. TN drives pool maintenance with its own `process_canon_state_update` and does not run reth's `maintain_transaction_pool` loop. In the pinned reth v1.11.3, the blob store uses deferred deletion: `delete` / `delete_all` only mark hashes for removal, and the only code that removes files from disk is `cleanup()`, reachable only through the maintenance loop. Because that loop never runs in TN, and blob transactions are never mined, no runtime path removes the sidecars. The on-disk blob store grows without bound.

Surfaced by Haxatron in the collaborative review ("The blob store can still be written to during reth txn pool ... it is possible for a user to submit many blob transactions into the reth pool and write unbounded data to the blob store which can exhaust disk and crash nodes"). This issue records the verified mechanism and impact. Verified at `origin/main` (1c73e8a3) against reth tag `v1.11.3`.

## Details

1. The pool admits blob transactions and writes the sidecar to disk at admission. The validator builds the stock reth blob-capable validator with a real `DiskFileBlobStore` and never disables type 3 (`crates/tn-reth/src/txn_pool.rs:112-123`). Blob validation has no fee floor: it checks fork activation, sidecar presence and version, and the KZG proof only (reth `validate/eth.rs:601-739`). The `pending_block_blob_fee = u128::MAX` that TN passes (`txn_pool.rs:245`) is a subpool-placement input, not an admission gate: it only prevents `TxState::ENOUGH_BLOB_FEE_CAP_BLOCK` (reth `pool/txpool.rs:1972-1984`), which parks the transaction in the blob subpool. On any successful add, the sidecar is written unconditionally, for every subpool: `on_added_transaction` -> `insert_blob` -> `blob_store.insert` (reth `pool/mod.rs:713-719,1259-1267`).

2. Blob transactions are never mined in TN, so the mined-removal path never touches them. The batch builder strips them (`crates/batch-builder/src/batch.rs`, `remove_eip4844_txs`) and the batch validator rejects any batch that carries one (`crates/batch-validator/src/validator.rs`). Their hashes therefore never appear in the `mined_transactions` set that `on_canonical_state_change` uses to drop mined blobs.

3. The blob store uses deferred deletion. `delete` / `delete_all` only insert the hash into an in-memory `txs_to_delete` set (reth `blobstore/disk.rs:150-164`). The only code that calls `fs::remove_file` is `cleanup()` (`disk.rs:166-189`, unlink at :174). The file doc states this: the store "uses deferred deletion ... it's expected that the maintenance task will call `BlobStore::cleanup`" (`disk.rs:28-29`).

4. `cleanup()` runs only from the maintenance loop. Its only Pool entry is `Pool::cleanup_blobs` (reth `pool/mod.rs:1280`), and the only production caller of that is `maintain_transaction_pool`, in its finalized-block branch (reth `maintain.rs:239-248`).

5. TN does not run that loop, and its own path never reaches `cleanup()`. TN drives the pool with `process_canon_state_update` -> `update_canonical_state` -> `Pool::on_canonical_state_change` (`crates/tn-reth/src/txn_pool.rs:149-249`). That function only marks discarded blobs through `delete_discarded_blobs` (reth `pool/mod.rs:508-532`, mark at :528; `delete_blobs` -> `delete_all`, mark only). A tree-wide search finds no call to `maintain_transaction_pool`, `cleanup_blobs`, or `BlobStoreCanonTracker` anywhere in TN.

6. No other mechanism frees the files at runtime. There is no `Drop` impl on `DiskFileBlobStore`. Subpool eviction and `discard_worst` / `truncate_pool` funnel back through `delete_blobs` (mark only). There is no disk-size cap: `blob_cache_size` bounds the in-memory decoded-sidecar cache, `blob_limit` bounds in-memory pool accounting, and `size_tracker` is a metrics gauge decremented only inside `cleanup()`. `DiskFileBlobStore::open` does wipe the whole blob directory (`disk.rs:47` -> `:344-345`, `remove_dir_all`), but that fires only at process start. A restart recovers the disk, but growth during a run is unbounded and the attacker resumes after the restart.

7. Ingress is unfiltered on the validator's own RPC. `add_raw_transaction_external` hands the transaction straight to `add_transaction(TransactionOrigin::External)` with no type-3 filter (`crates/tn-reth/src/txn_pool.rs:275-287`). The only allowlist screen lives in the separate, optional `bin/worker-gateway` binary (`proxy.rs:203-249`); a direct connection to a validator's own listener bypasses it.

## Attacker cost

Near zero. There is no minimum priority fee (`minimum_priority_fee: None`, reth `validate/eth.rs:968`). The only affordability gate is `cost <= balance`, where `cost` is computed from the transaction's own declared fees (reth `validate/eth.rs:654-670`). Set all fee fields and value to zero, and `cost = 0`, so a fresh unfunded account is admitted. Because the transaction is never mined, the sender is never charged. Distinct keypairs mint unbounded distinct hashes, and each type-3 transaction carries up to about 128 KB per blob (up to 6 blobs). Disk fills quickly at no cost.

## Impact

Availability / denial of service. Single-node disk exhaustion crashes the node. Consensus tolerates fewer than one third faulty, so a single victim is a liveness hit on that validator, not an instant chain halt. The startup wipe makes it recoverable by restart, but the attacker resumes immediately, so it is a persistent, near-free griefing vector that forces operator intervention.

Rating: Medium. It escalates toward High if validators expose unfiltered RPC (gateway not interposed, internal, or misconfigured) and several validators are targeted together toward a liveness-threatening fault fraction. Not consensus or fund safety.

## Suggested fix

Reject EIP-4844 at admission, the same way reth's own node builder does. In `crates/tn-reth/src/txn_pool.rs`, add `.no_eip4844()` on the validator builder (and `.no_eip7702()` for the parallel unsupported type). Once type 3 is rejected before insertion, the `kzg_settings` call and the `DiskFileBlobStore` can be dropped. Keep the batch-builder sweep and the batch-validator rejection as defense in depth.

This is the same remediation proposed in the companion pool-admission-allowlist issue. The two share one fix; this issue tracks the distinct availability impact.

Add a test that submits a blob transaction through the production ingress path (`add_raw_transaction_external`) and asserts rejection at submission.

## References

- Finding surfaced by Haxatron (collaborative review).
- TN `origin/main` (1c73e8a3): `crates/tn-reth/src/txn_pool.rs:112-123` (pool build), `:149-249` (`process_canon_state_update`, `pending_block_blob_fee = u128::MAX`), `:275-287` (unfiltered external ingress), `:334-337` (`remove_eip4844_txs`).
- reth v1.11.3: `blobstore/disk.rs:28-29,150-189,344-345`; `pool/mod.rs:508-532,713-719,1259-1303`; `maintain.rs:239-248`; `validate/eth.rs:601-739,654-670,968`; `pool/txpool.rs:1972-1984`.
- `crates/batch-builder/src/batch.rs`, `crates/batch-validator/src/validator.rs` (blob transactions never mined).
- Companion draft: `issue-pool-admission-allowlist.md` (shared fix). Precedent: PR #1104, issue #1067 (allowlist symmetry).

MavenRain:
## Summary

Reject EIP-4844 (blob) and EIP-7702 (set-code) transactions at transaction-pool admission by
adding `.no_eip4844()` and `.no_eip7702()` to the validator builder in `WorkerTxPool::new`. This
closes the unbounded on-disk blob-sidecar denial-of-service described in #1159 and brings pool
admission in line with the batch builder and batch validator, which already refuse both types.

Closes #1159. (Cantina #1)

## Problem

TN admits blob transactions and writes each sidecar to the on-disk `DiskFileBlobStore`, but nothing
removes those files at runtime, so a remote unprivileged sender can exhaust a validator's disk and
crash the node.

The decisive chain, verified at `origin/main` (1c73e8a3) against the pinned reth `v1.11.3`:

- The validator is built with a real `DiskFileBlobStore` and does not disable type 3
  (`crates/tn-reth/src/txn_pool.rs`). Blob validation has no fee floor, so a fresh unfunded key with
  all-zero fees is admitted (`cost <= balance` with `cost = 0`), and because the transaction is
  never mined the sender is never charged.
- On any successful add, reth writes the sidecar unconditionally, for every subpool
  (`on_added_transaction` -> `insert_blob` -> `blob_store.insert`).
- The reth blob store uses deferred deletion: `delete` / `delete_all` only mark hashes in an
  in-memory set; the sole `fs::remove_file` lives in `cleanup()`, which is reachable only through
  `Pool::cleanup_blobs`, whose only production caller is reth's `maintain_transaction_pool` loop.
- TN does not run that loop. It drives the pool itself via `process_canon_state_update` ->
  `on_canonical_state_change`, which only marks discarded blobs, never calls `cleanup()`. A tree-wide
  search finds no call to `maintain_transaction_pool`, `cleanup_blobs`, or `BlobStoreCanonTracker`.
- Blob transactions are never mined (the batch builder strips them, the batch validator rejects any
  batch that carries one), so the mined-removal path never fires either.
- The startup wipe in `DiskFileBlobStore::open` (`remove_dir_all`) only runs at process start, so a
  restart recovers the disk but the attacker resumes immediately.

Each type-3 transaction carries up to ~128 KB per blob (up to 6 blobs), and distinct keys mint
distinct hashes, so disk fills quickly at near-zero cost.

Finding surfaced by Haxatron in the collaborative review. This PR records the verified reth wiring
and closes the vector.

## Fix

Add `.no_eip4844()` and `.no_eip7702()` on the `EthTransactionValidatorBuilder` in
`WorkerTxPool::new`. In reth `v1.11.3` this flips `EthTransactionValidatorBuilder::eip4844` /
`eip7702` (default `true`) to `false`, so `validate_one` rejects the type before any subpool or
blob-store side effect: the `EIP4844_TX_TYPE_ID` arm returns
`InvalidTransactionError::Eip4844Disabled` and the `EIP7702_TX_TYPE_ID` arm returns
`Eip7702Disabled`. That check runs ahead of origin branching, so it rejects local, external, and
gossipped submissions alike. With no blob transaction ever inserted, no sidecar is ever written, and
the DoS surface is gone.

EIP-7702 is included because TN already treats it as an unsupported type (the batch builder strips it
and the batch validator maps it to `UnsupportedTxType`); rejecting it at the pool is the same policy
applied one layer earlier, not a new design decision.

The change is intentionally minimal. The existing batch-builder sweep and batch-validator rejection
are kept as defense in depth. The `DiskFileBlobStore` and `kzg_settings` are left in place: with
blobs rejected before insertion the store is never written, and removing it would change the pool's
public type parameters for no security benefit.

## Threat model

- Attacker: any party that can reach a validator's transaction ingress (its own RPC, or the
  gossip/forwarding path). No funds, no privilege, no prior state.
- Before: a stream of valid zero-fee blob transactions writes sidecars that are never reclaimed at
  runtime, growing the validator's disk without bound until the node crashes; the attacker resumes
  after any restart. Single-node availability / liveness impact (Medium; escalates if several
  validators are targeted toward a liveness-threatening fault fraction).
- After: every blob and set-code transaction is rejected at `validate_one`, before it can be
  inserted or write a sidecar, on every ingress path.

## Testing

Built against an isolated target with the pinned toolchain (`cargo +1.94`, sccache, `-j 2`).

New unit tests in `crates/tn-reth/src/txn_pool.rs`:

- `test_pool_rejects_blob_transaction`: funds the sender at genesis and builds a valid-KZG blob
  transaction, so the only remaining ground for rejection is the `.no_eip4844()` type gate.
  `add_transaction_local` returns an error and the pool stays empty (`pending`, `blob`, and `queued`
  all `0`). An empty pool proves no sidecar reached the on-disk blob store, which is the exact DoS
  surface #1159 describes.
- `test_pool_rejects_eip7702_transaction`: funds the sender, submits a fork-valid EIP-7702 set-code
  transaction through the external raw-ingress path (`add_raw_transaction_external`), and asserts the
  same error-and-empty-pool result via the `.no_eip7702()` gate.

Suites:

- `cargo +1.94 test -p tn-reth --lib`: `141 passed; 0 failed`, including both new tests.
- `cargo +1.94 test -p tn-batch-builder --test it test_batch_builder_produces_valid_batches`:
  `1 passed; 0 failed`. This existing test previously submitted a blob and expected the batch builder
  to discard it. With blobs rejected at admission the blob setup is removed and the assertions now
  track the observed pool state: the initial pending count drops from 4 to 3 (no blob in the pool),
  the mid-run pending count stays 4 (the three funded EIP-1559 transactions remain pending and a
  fourth is submitted before the canonical-state ack), and the first batch carries the three original
  transactions.

Formatting: `cargo +nightly-2026-03-20 fmt --check` clean on both crates.
Lint: `cargo +nightly-2026-03-20 clippy -p tn-reth -p tn-batch-builder --all-features` clean (exit 0, no warnings).

## References

- reth `v1.11.3`: `crates/transaction-pool/src/validate/eth.rs` (`no_eip4844` / `no_eip7702`
  builder methods; the `EIP4844_TX_TYPE_ID` / `EIP7702_TX_TYPE_ID` reject arms in `validate_one`);
  `crates/transaction-pool/src/blobstore/disk.rs` (deferred deletion, `cleanup`);
  `crates/transaction-pool/src/maintain.rs` (`maintain_transaction_pool`).
- TN: `crates/tn-reth/src/txn_pool.rs` (pool build and `process_canon_state_update`);
  `crates/batch-builder/src/batch.rs`, `crates/batch-validator/src/validator.rs` (blob/7702 never
  mined).
- Precedent: PR #1104, issue #1067 (batch-validator tx-type allowlist).

### DIFF
--- a/crates/batch-builder/tests/it/build_batches.rs
+++ b/crates/batch-builder/tests/it/build_batches.rs
@@ -19,7 +19,7 @@ use tn_test_utils::wait_until;
 use tn_types::{
     gas_accumulator::GasAccumulator, test_genesis, Address, Batch, BatchValidation, Bytes,
     Certificate, CertifiedBatch, CommittedSubDag, ConsensusHeaderDigest, ConsensusOutput, Database,
-    Encodable2718, NoopTxnForwarder, ReputationScores, SealedBatch, TaskManager,
+    Encodable2718, GenesisAccount, NoopTxnForwarder, ReputationScores, SealedBatch, TaskManager,
     MIN_PROTOCOL_BASE_FEE, U160, U256,
 };
 use tn_worker::{test_utils::TestMakeBlockQuorumWaiter, Worker, WorkerNetworkHandle};
@@ -201,11 +201,11 @@ async fn test_make_batch_el_to_cl() -> eyre::Result<()> {
     Ok(())
 }
 
-/// Create 4 valid EIP-1559 transactions.
+/// Create 4 valid EIP-1559 transactions and 1 valid EIP-4844 (blob) transaction.
 ///
-/// No blob transaction is involved: the pool now rejects EIP-4844 transactions at admission,
-/// so none can reach the batch builder. First 3 transactions mined in the first batch.
-/// Before a canonical state change, mine the 4th transaction in the next batch.
+/// The pool rejects the blob transaction at admission (issue #1159), so it never reaches the
+/// batch builder. First 3 transactions mined in the first batch. Before a canonical state
+/// change, mine the 4th transaction in the next batch.
 #[tokio::test]
 async fn test_batch_builder_produces_valid_batches() {
     //
@@ -213,6 +213,14 @@ async fn test_batch_builder_produces_valid_batches() {
     //
     // adiri genesis with TxFactory funded
     let genesis = test_genesis();
+
+    // fund a second random factory for the eip-4844 transaction, so its rejection below can
+    // only come from the pool's type gate, never from insufficient balance
+    let mut blob_tx_factory = TransactionFactory::new_random();
+    let genesis = genesis.extend_accounts([(
+        blob_tx_factory.address(),
+        GenesisAccount::default().with_balance(U256::MAX),
+    )]);
     let chain: Arc<RethChainSpec> = Arc::new(genesis.into());
     let address = Address::from(U160::from(333));
     let tmp_dir = TempDir::new().unwrap();
@@ -278,10 +286,14 @@ async fn test_batch_builder_produces_valid_batches() {
     let added_result = tx_factory.submit_tx_to_pool(transaction3.clone(), txpool.clone()).await;
     assert_matches!(added_result, hash if &hash == transaction3.hash());
 
-    // txpool size
+    // submit a valid eip-4844 blob transaction: the pool must reject it at admission (#1159)
+    let blob_pooled = blob_tx_factory.create_eip4844_pooled(chain.clone(), None, gas_price);
+    let blob_result = txpool.add_transaction_local(blob_pooled).await;
+    assert!(blob_result.is_err());
+
+    // txpool size: the rejected blob left nothing behind
     let pool_size = txpool.pool_size();
     assert_eq!(pool_size.pending, 3);
-    // blobs are rejected at pool admission, so none can be present
     assert_eq!(pool_size.blob, 0);
 
     // spawn batch_builder once worker is ready
--- a/crates/batch-builder/tests/it/build_batches.rs
+++ b/crates/batch-builder/tests/it/build_batches.rs
@@ -19,7 +19,7 @@ use tn_test_utils::wait_until;
 use tn_types::{
     gas_accumulator::GasAccumulator, test_genesis, Address, Batch, BatchValidation, Bytes,
     Certificate, CertifiedBatch, CommittedSubDag, ConsensusHeaderDigest, ConsensusOutput, Database,
-    Encodable2718, GenesisAccount, NoopTxnForwarder, ReputationScores, SealedBatch, TaskManager,
+    Encodable2718, NoopTxnForwarder, ReputationScores, SealedBatch, TaskManager,
     MIN_PROTOCOL_BASE_FEE, U160, U256,
 };
 use tn_worker::{test_utils::TestMakeBlockQuorumWaiter, Worker, WorkerNetworkHandle};
@@ -201,12 +201,11 @@ async fn test_make_batch_el_to_cl() -> eyre::Result<()> {
     Ok(())
 }
 
-/// Create 5 transactions.
+/// Create 4 valid EIP-1559 transactions.
 ///
-/// First 4 mined in first batch.
-/// One of the transactions is EIP-4844 blob which is discarded.
-/// (only 3 valid txs in first batch)
-/// Before a canonical state change, mine the 5th transaction in the next batch.
+/// No blob transaction is involved: the pool now rejects EIP-4844 transactions at admission,
+/// so none can reach the batch builder. First 3 transactions mined in the first batch.
+/// Before a canonical state change, mine the 4th transaction in the next batch.
 #[tokio::test]
 async fn test_batch_builder_produces_valid_batches() {
     //
@@ -214,15 +213,6 @@ async fn test_batch_builder_produces_valid_batches() {
     //
     // adiri genesis with TxFactory funded
     let genesis = test_genesis();
-
-    // create random tx factory for eip-4844 transaction reth does not allow
-    // different tx types in pool at same time from same address
-    // see Err: ExistingConflictingTransactionType
-    let mut blob_tx_factory = TransactionFactory::new_random();
-    let genesis = genesis.extend_accounts([(
-        blob_tx_factory.address(),
-        GenesisAccount::default().with_balance(U256::MAX),
-    )]);
     let chain: Arc<RethChainSpec> = Arc::new(genesis.into());
     let address = Address::from(U160::from(333));
     let tmp_dir = TempDir::new().unwrap();
@@ -288,15 +278,10 @@ async fn test_batch_builder_produces_valid_batches() {
     let added_result = tx_factory.submit_tx_to_pool(transaction3.clone(), txpool.clone()).await;
     assert_matches!(added_result, hash if &hash == transaction3.hash());
 
-    // submit eip-4844 blob transaction
-    let _ = blob_tx_factory
-        .create_and_submit_eip4844(chain.clone(), None, gas_price, txpool.clone())
-        .await;
-
     // txpool size
     let pool_size = txpool.pool_size();
-    assert_eq!(pool_size.pending, 4);
-    // ensure blob is not under valued
+    assert_eq!(pool_size.pending, 3);
+    // blobs are rejected at pool admission, so none can be present
     assert_eq!(pool_size.blob, 0);
 
     // spawn batch_builder once worker is ready
@@ -326,7 +311,8 @@ async fn test_batch_builder_produces_valid_batches() {
         )
         .await;
 
-    // assert 4 txs in pending pool - blob should be removed while creating first batch
+    // assert 4 txs in pending pool - the 3 original txs are not mined until the ack,
+    // plus the 1 new tx submitted above
     let pool_size = txpool.pool_size();
     assert_eq!(pool_size.pending, 4);
     assert_eq!(pool_size.blob, 0);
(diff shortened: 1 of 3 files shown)


crates/batch-builder/tests/it/build_batches.rs
```diff
@@ -288,15 +278,10 @@ async fn test_batch_builder_produces_valid_batches() {
     let added_result = tx_factory.submit_tx_to_pool(transaction3.clone(), txpool.clone()).await;
     assert_matches!(added_result, hash if &hash == transaction3.hash());
 
-    // submit eip-4844 blob transaction
-    let _ = blob_tx_factory
-        .create_and_submit_eip4844(chain.clone(), None, gas_price, txpool.clone())
-        .await;
-
```
grantkee, line 295: Instead of removing this logic and asserting L283, assert that this is an err? As it stands, this doesn't really test anything

crates/tn-reth/src/txn_pool.rs
```diff
@@ -114,6 +114,19 @@ impl WorkerTxPool {
...
+        // For blobs this is also a denial-of-service fix. On a successful add reth writes the blob
+        // sidecar to the on-disk DiskFileBlobStore, but that store uses deferred deletion whose
+        // only unlink runs in reth's maintain_transaction_pool loop. TN drives pool
+        // maintenance itself and never runs that loop, so nothing removes the sidecars at
+        // runtime and a remote unprivileged sender could grow a validator's disk without
+        // bound. Rejecting both unsupported types here, before insertion, closes that
+        // vector and mirrors reth's own node builder for a chain that supports neither
+        // type. See issue #1159.
+        .no_eip4844()
+        .no_eip7702()
```
grantkee, line 129: This is just a headsup: we plan to support this tx type in #1154. This PR will likely land first and be addressed in the follow-up. I think `no_eip7702` is in-scope for this issue/PR and no change is necessary

crates/batch-builder/tests/it/build_batches.rs
MavenRain: Done

grantkee (approved):
LGTM

sstanfield (approved):

Haxatron:
fix verified. no_eip4844 rejects blob txs from ever entering the mempool, preventing any sidecars from being stored on disk.

Issue #1158, opened 2026-08-12
Optimistic U256::MAX balance in batch builder amplifies invalid transactions into follow-up batches (griefing/DoS)


MavenRain:
## Summary (Cantina #16)

The batch builder reports an optimistic `balance: U256::MAX` for every sender that had a transaction mined in a batch.  Reth's transaction pool treats that sender as able to pay for all of its queued transactions and promotes them from the parked sub-pool to `pending`.  The batch builder then loops immediately and packs those promoted transactions into the next batch, before the engine's real canonical update corrects the balance.  Peer batch validation is purely structural, so the follow-up batch reaches quorum.  At execution the sender's real balance is already spent, so those transactions fail stateful validation and are skipped for free.

A funded attacker pays for about one successful transaction per sender but forces the network to gossip, decode, recover signatures for, quorum-certify, store, and execution-validate up to 255 extra invalid transactions per sender at no cost.  This is a low-cost griefing and resource-amplification vector.

Originally reported by **Haxatron** in the Telcoin Network collaborative review (Cantina) and acknowledged by the maintainers.  Filing here because platform findings are not being tracked separately during this review.

All line references are at commit `7891daa5`.

## Affected code

- `crates/batch-builder/src/batch.rs:128-139` builds the synthetic account update.
- `crates/batch-builder/src/lib.rs:318-343` applies it and loops without waiting for the engine.
- `crates/batch-validator/src/validator.rs:35-81` validates peer batches without any balance or nonce check against state.
- `crates/tn-reth/src/env/execution.rs:160-179` skips invalid transactions at execution with no fee.
- `crates/tn-reth/src/cli.rs:98` sets the per-sender pool cap to 256 (Reth's own default is 16).

## Mechanism

**1. The batch builder reports an unlimited balance.**

After a batch reaches quorum, `build_batch()` emits one `ChangedAccount` per sender that had a transaction in the batch:

```rust
// crates/batch-builder/src/batch.rs:132
let changed_accounts: Vec<ChangedAccount> = sender_nonces
    .into_iter()
    .map(|(address, max_nonce)| ChangedAccount {
        address,
        nonce: max_nonce + 1, // next expected nonce
        balance: U256::MAX,   // corrected by engine's canonical update
    })
    .collect();
```

**2. The pool promotes the sender's parked transactions.**

[...]

PR #1170, opened 2026-08-13
fix(batch-builder): carry real sender balance in optimistic pool update

MavenRain:
## Summary

The batch builder's optimistic post-mining pool update reported `balance: U256::MAX` for every sender that had a transaction mined in a batch. That inflated balance let Reth's transaction pool promote a sender's parked, insufficient-funds transactions into the very next batch, which the batch builder builds immediately inside the optimistic window before the engine's canonical update corrects the balance. Because peer batch validation is purely structural, the follow-up batch reaches quorum, is gossiped and stored, and is then skipped for free at execution.

This PR sets the optimistic balance to the sender's **real canonical balance minus the cost of the transactions just mined for that sender** instead of `U256::MAX`, so a sender that cannot fund its remaining transactions keeps them parked. The nonce advance that keeps legitimate sequential senders flowing is unchanged.

Fixes the Cantina finding reported by **Haxatron** and tracked in #1158.

## Problem

After a batch reaches quorum, `build_batch()` emits one `ChangedAccount` per sender that had a transaction in the batch, and `lib.rs` feeds those into `pool.update_canonical_state(...)` → Reth's `on_canonical_state_change`. When a sender's reported balance jumps to `U256::MAX`, Reth reclassifies that sender's queued (insufficient-funds) transactions into `pending`. The run loop then continues **without** waiting for the engine's authoritative canonical stream (this is deliberate, to avoid stalling legitimate sequential senders), so the next batch is built from transactions that only *look* affordable.

`validate_batch()` checks the digest, worker id, epoch, byte size, decoding, blob exclusion, gas limit, and base fee — but no balance or nonce against execution state — so the follow-up batch of underfunded transactions still receives quorum certification. At execution the sender's real balance is already spent, so those transactions fail stateful validation and are skipped with no fee and no state change.

Net effect: a funded attacker pays for roughly one successful transaction per sender but forces the network to gossip, decode, recover signatures for, quorum-certify, store, and execution-validate up to `TN_TXPOOL_MAX_ACCOUNT_SLOTS_PER_SENDER - 1` (255 at the current cap) extra invalid transactions per sender, at no cost. The documented "No spam vector" defense in the README covers a *different* claim — new transactions entering the pool are validated against real balance at entry — but the optimistic `changed_accounts` balance governs promotion/demotion of transactions **already** in the pool, which is exactly what the attack exploits.

## Fix

- **`TxPool::get_account_balance(address) -> U256`** (`crates/tn-reth/src/txn_pool.rs`): new trait method returning a sender's canonical balance as of the latest committed block. `WorkerTxPool` now retains a `BlockchainProvider` handle and answers via `basic_account`. A missing account or a state-read error yields `U256::ZERO`, the conservative choice: it can only keep a sender's remaining transactions parked, never promote an unfunded one, and the engine's authoritative update corrects it the same round.
- **`build_batch`** (`crates/batch-builder/src/batch.rs`): while packing, it accumulates each sender's mined-transaction cost, and the optimistic `ChangedAccount.balance` is now `pool.get_account_balance(address).saturating_sub(mined_cost)` instead of `U256::MAX`. The nonce advance (`max_nonce + 1`) is unchanged.
- **README** (`crates/batch-builder/README.md`): the balance-rationale section is rewritten to explain the real-balance-minus-mined-cost update, why `U256::MAX` was an amplification vector, and the known residual below.

### Why this is safe and preserves throughput

- **The burst is closed.** Debiting the just-mined cost means a sender whose real balance cannot cover its remaining transactions keeps them parked. The immediate amplification in the issue's PoC — queue up to 255 transactions, mine one, watch the rest promote into the very next batch — no longer occurs.
- **Legitimate senders unaffected.** A sender who actually funded its transactions keeps them in `pending` (real balance minus the mined spend still covers the rest); only genuinely underfunded transactions are parked, which is correct. The debit uses each transaction's max cost, so it errs toward parking rather than over-promoting, and the engine's canonical update restores the true balance within the same consensus round.
- **Entry validation unchanged.** New transactions are still validated against the real on-chain balance by Reth's `TransactionValidationTaskExecutor`.

### Known residual (deliberately not fixed here)

The optimistic balance is re-derived from the last committed canonical balance on every in-window rebuild and does **not** accumulate debits across successive rebuilds before the engine's canonical update lands. A carefully drip-fed sender can therefore still have on the order of one transaction promoted per rebuild (rather than the full parked set at once) until the canonical update corrects the balance. Fully closing this requires tracking cumulative optimistic spend across rebuilds, or a state-aware check on the follow-up batch path (directions 3 and 4 in the issue). Both are larger design choices with throughput trade-offs that the maintainers are best placed to make; this PR is the minimal, local change that removes the immediate burst without touching the consensus or execution paths. It implements directions 1 and 2 and strictly improves on the current behavior in the meantime.

## Testing

- New unit test `crates/batch-builder/src/batch.rs::tests::changed_accounts_carry_real_balance_and_advance_nonce`: builds two sequential transactions from one sender, has the pool report a large finite balance, calls `build_batch`, and asserts the produced `changed_accounts` entry carries `real_balance - mined_cost` (not `U256::MAX`) while still advancing the nonce past the highest mined nonce.
- `cargo +1.94 test -p tn-batch-builder --features test-utils --lib`
- `cargo +1.94 fmt --check`, nightly `clippy --workspace --all-features`, and `make attest` on the pinned toolchain.

## Notes

- All code references are on current `main`; the issue's line numbers are at its reference commit `7891daa5`.
- Related but out of scope: the per-sender slot cap of 256 (`crates/tn-reth/src/cli.rs`, up from Reth's default 16) sets the amplification factor and is left unchanged here.

### DIFF
--- a/crates/batch-builder/README.md
+++ b/crates/batch-builder/README.md
@@ -59,21 +59,24 @@ This allows the batch builder to immediately loop and build the next batch from 
 
 The pool receives updates from **two independent sources**:
 
-1. **Batch builder** (`lib.rs` → `update_canonical_state()`): Called immediately after quorum. Provides an early, optimistic update with mined transaction hashes and nonce advances. Uses `U256::MAX` for balances (see below).
+1. **Batch builder** (`lib.rs` → `update_canonical_state()`): Called immediately after quorum. Provides an early, optimistic update with mined transaction hashes and nonce advances. Carries each sender's real canonical balance (see below).
 
 2. **Canonical pool task** (`txn_pool.rs` → `process_canon_state_update()`): A separate background task subscribed to the engine's canonical state stream. Called after the engine executes the committed batches and produces a new canonical block. Provides authoritative nonce, balance, and mined transaction data derived from actual EVM execution.
 
 Both call `on_canonical_state_change()` on the underlying Reth pool. The engine's update overwrites the batch builder's optimistic state with the real post-execution values.
 
-### Why `U256::MAX` for Balance
+### Which Balance the Optimistic Update Carries
 
 The batch builder does not execute transactions and cannot know the post-execution balance.
-The balance field in `ChangedAccount` is set to `U256::MAX` for the following reasons:
+The balance field in `ChangedAccount` is set to the sender's **real canonical balance minus the cost of the transactions just mined for that sender** (balance read via `TxPool::get_account_balance`), not `U256::MAX`.
 
-- **Pre-execution balance is equally wrong.** After mining 8 transactions that each transfer value, the sender's real balance is lower than the pre-execution balance. Neither `U256::MAX` nor the pre-execution balance reflects reality — only the post-execution balance does, which requires EVM execution.
-- **No spam vector.** New transactions entering the pool are validated against the real on-chain balance by Reth's `TransactionValidationTaskExecutor`. The `changed_accounts` balance only affects whether the pool *demotes existing* transactions after mining — it does not bypass entry validation.
-- **Short persistence window.** The `U256::MAX` balance only persists until the engine's canonical update arrives (same consensus round), which overwrites it with the real post-execution balance.
-- **Prevents unnecessary demotion.** Using a lower balance could cause the pool to evict transactions that are actually valid, degrading throughput for legitimate senders.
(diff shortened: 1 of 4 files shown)


sstanfield (approved):

grantkee (approved):
Overall improvement from applying unbounded `U256::MAX` updates

Haxatron:
Fix verified the real sender balance (sender balance - maximum sum of transaction fees) is now reported in the canon notification so that the pool can update / clear any stale transactions accordingly

PR #1171, opened 2026-08-13
fix(tn-reth): try the next validator on node-local -32000 refusals

### LINKED ISSUE #1168: `tn-reth`: `classify_server_error` treats node-local reth `-32000` refusals (underpriced, replacement, fee cap, address reserved) as chain-wide verdicts and drops transactions other validators would accept
## Summary

`classify_server_error` in `crates/tn-reth/src/forward.rs` decides between "verdict on the transaction" and "validator-local condition" by JSON-RPC error code alone:

```rust
/// JSON-RPC error codes reth returns for a validator-local, transient condition (a full pool,
/// `-32003`, or an internal/IO error, `-32603`) rather than a verdict on the transaction itself.
/// The forward falls through to the next advertised validator on these, since another validator
/// may still accept the transaction.
const TRANSIENT_RPC_CODES: [i64; 2] = [-32003, -32603];
```

```rust
fn classify_server_error(code: i64, message: String) -> Disposition {
    if TRANSIENT_RPC_CODES.contains(&code) {
        Disposition::TryNext(message)
    } else if message.to_ascii_lowercase().contains(ALREADY_KNOWN_MESSAGE) {
        Disposition::Delivered
    } else {
        // Every validator shares consensus state, so a considered rejection repeats everywhere.
        Disposition::Rejected(format!("code {code}: {message}"))
    }
}
```

The code axis cannot express that split.  In the pinned reth (tag `v1.11.3`), both `-32000` and `-32003` mix node-local conditions with chain-wide verdicts.  The result is a liveness gap: a node-local refusal that arrives as `-32000` becomes `Disposition::Rejected`, which breaks the fallback chain in `forward_txns`, and the transaction is dropped even when the next advertised validator would accept it.  The submitting client already received a success response from the observer's RPC, so the loss is silent apart from a `warn!` log.

## Evidence: the reth v1.11.3 code assignment

All citations are to the reth checkout that `Cargo.toml` pins (`tag = "v1.11.3"`).

`From<RpcPoolError> for ErrorObject`, `crates/rpc/rpc-eth-types/src/error/mod.rs:1016-1042`:

- Only `TxPoolOverflow` maps to `-32003` (`EthRpcErrorCode::TransactionRejected`).
- `AlreadyKnown`, `InvalidSender`, `Underpriced`, `ReplaceUnderpriced`, `ExceedsGasLimit`, `MaxTxGasLimitExceeded`, `ExceedsFeeCap`, `NegativeValue`, `OversizedData`, `ExceedsMaxInitCodeSize`, `PoolTransactionError`, `Eip4844`, `Eip7702`, and `AddressAlreadyReserved` all map to `-32000` (`EthRpcErrorCode::InvalidInput`).
- Only `RpcPoolError::Other` maps to `-32603`.

`RpcInvalidTransactionError::error_code`, same file, lines 743-756: `InvalidChainId`, `GasTooLow`, `GasTooHigh`, `GasRequiredExceedsAllowance`, `NonceTooLow`, `NonceTooHigh`, `FeeCapTooLow`, `FeeCapVeryHigh` map to `-32000`; `Revert` maps to `3`; **everything else falls through a `_` arm to `-32003`**, including `InsufficientFunds`, `TxTypeNotSupported`, `NonceMaxValue`, and the out-of-gas variants.

reth's `send_raw_transaction` handler surfaces pool errors through `EthApiError::PoolError(err) => err.into()` (mod.rs:293) with no other mapping layer, so a validator's advertised endpoint returns exactly these codes. telcoin-network builds that endpoint from stock reth components (`RpcModuleBuilder` + `EthApi` in `crates/tn-reth/src/lib.rs`, stock `EthTransactionPool` with reth's own validator in `crates/tn-reth/src/txn_pool.rs`), so nothing rewrites the error surface in between.

## Consequence 1: node-local refusals terminate the fallback chain

These conditions are properties of one validator's pool contents or one operator's configuration, not of the transaction.  Each arrives as `-32000`, misses the `already known` substring, and stops the chain as `Rejected`:

| Condition | Message | Why it is node-local |
|---|---|---|
| `ReplaceUnderpriced` | `replacement transaction underpriced` | Depends on what sits in that validator's pool and on its `--txpool.pricebump`. A fallback validator without the original pooled would accept the same bytes. |
| `Underpriced` (via `FeeCapBelowMinimumProtocolFeeCap`) | `transaction underpriced` | `--txpool.minimal-protocol-fee` is per-operator CLI config, enforced as a hard reject in reth's pool insert. |
| `ExceedsFeeCap` | `tx fee (...) exceeds the configured cap (...)` | `--rpc.txfeecap` is per-node RPC config (see #1160 for its current enforcement status;  once enforced, a capped validator refuses what an uncapped one accepts). |
| `AddressAlreadyReserved` | `address already reserved` | Blob-vs-regular exclusivity against that validator's current pool contents. |
| `OversizedData`, `MaxTxGasLimitExceeded`, `ExceedsMaxInitCodeSize` | various | Pool-config limits.  Uniform by default, divergent only under config skew, listed for completeness. |

By contrast the pool-full case the constant was written for (`TxPoolOverflow`, `-32003`) is already handled correctly, as is `RpcPoolError::Other` (`-32603`).

## Consequence 2: the `-32003` arm retries chain-wide verdicts

Because reth's `RpcInvalidTransactionError::error_code` sends its remainder to `-32003`, deterministic verdicts such as `InsufficientFunds` and `TxTypeNotSupported` are classified `TryNext` and re-sent to every advertised validator before being dropped as `NoEndpointReached`.  This is bounded (committee size, `FORWARD_TX_BUDGET`) and harmless to correctness, but it contradicts the doc comment on `TRANSIENT_RPC_CODES` (`-32003` is not "a full pool";  it is also reth's catch-all invalid-transaction code), and it makes the terminal log misleading (`could not forward transaction to any advertised validator RPC` for a transaction that received a definitive verdict).

## Suggested direction

Options, not mutually exclusive:

1. Treat the known node-local `-32000` messages as `TryNext` by substring, the same mechanism already used for `already known`: `transaction underpriced`, `replacement transaction underpriced`, `exceeds the configured cap`, `address already reserved`.
2. Alternatively invert the default for unrecognized codes to `TryNext` (fail open). The chain is already bounded by `FORWARD_TX_BUDGET`, so the worst case for a true verdict is a few redundant round-trips, while the worst case of the current fail-closed default is a silently dropped transaction.
3. Document that committee validators must run uniform admission config (`--txpool.pricebump`, `--txpool.minimal-protocol-fee`, `--rpc.txfeecap`, pool size limits), which shrinks but does not eliminate the class: `ReplaceUnderpriced` and `AddressAlreadyReserved` stay node-local under identical config because pools differ.
4. Correct the doc comment either way: `-32003` is `TxPoolOverflow` plus reth's invalid-transaction catch-all, not "a full pool".

## Relation to existing work

- #1145 / PR #1153 cover admission, requeue, and endpoint demotion around the forwarder.  #1153's description states the `RpcError` match is exhaustive over transport shapes and defers a backpressure signal for the `-32003`/`-32603` transients;  it does not change how server-answered codes are split into `Rejected` vs `TryNext`, which is what this issue is about.  If #1153 lands first, the cited lines move but the classifier text is unchanged.
- #1160 tracks `--rpc.txfeecap` enforcement.  Its landing makes the `ExceedsFeeCap` row above reachable in practice.

## Provenance

Verified against reth `v1.11.3` (checkout `d6324d63e`) at the cited lines, against `crates/tn-reth/src/forward.rs` (`classify_server_error`, `TRANSIENT_RPC_CODES`, `forward_txns`), and against the RPC assembly in `crates/tn-reth/src/lib.rs` / `txn_pool.rs`.  Each claim was independently re-read adversarially before drafting;  no telcoin-side re-mapping of the reth error surface exists on the `eth_sendRawTransaction` path.

MavenRain:
Closes #1168. (Cantina #19)

## Problem

`classify_server_error` in `crates/tn-reth/src/forward.rs` splits a server answer into a verdict on the transaction (`Rejected`, stop the fallback chain) or a validator-local condition (`TryNext`, try the next validator). The split uses the JSON-RPC error code alone. The pinned reth (`v1.11.3`) does not give node-local refusals their own code. Reth maps `Underpriced`, `ReplaceUnderpriced`, `ExceedsFeeCap`, and `AddressAlreadyReserved` to `-32000` (`crates/rpc/rpc-eth-types/src/error/mod.rs`, `From<RpcPoolError> for ErrorObject`). The same code carries chain-wide verdicts such as `nonce too low`.

A node-local refusal that arrives as `-32000` becomes `Disposition::Rejected`. The forward stops the fallback chain and drops the transaction. The next advertised validator would accept the same bytes. The submitting client already holds a success response from the observer RPC, so the loss is silent apart from one `warn!` line.

## Root cause

The code axis cannot express the node-local versus chain-wide split for `-32000`. Only `TxPoolOverflow` maps to `-32003`, and only `RpcPoolError::Other` maps to `-32603`. `TRANSIENT_RPC_CODES: [-32003, -32603]` therefore covers one node-local pool refusal and misses the other four.

## Fix

Classify the known node-local `-32000` refusals by message substring, with the same mechanism the function already uses for `already known`. A new `NODE_LOCAL_MESSAGES` table holds the four reth messages:

- `transaction underpriced` (`RpcPoolError::Underpriced`: below that operator's `--txpool.minimal-protocol-fee`)
- `replacement transaction underpriced` (`RpcPoolError::ReplaceUnderpriced`: below that validator's `--txpool.pricebump` against its current pool)
- `exceeds the configured cap` (`RpcPoolError::ExceedsFeeCap`: above that node's `--rpc.txfeecap`)
- `address already reserved` (`RpcPoolError::AddressAlreadyReserved`: blob-versus-regular sender exclusivity against that pool)

A match returns `Disposition::TryNext`, so the forward tries the next advertised validator. The carve-out is scoped to code `-32000` (`INVALID_INPUT_RPC_CODE`), so the same substring under any other code keeps its terminal meaning. An unrecognized `-32000` message keeps the terminal `Rejected` classification: chain-wide verdicts (bad nonce, wrong chain, malformed) dominate that code, and every validator repeats them.

The threat model does not change. Each attempt stays bounded by `FORWARD_SEND_TIMEOUT`, and the whole chain stays bounded by `FORWARD_TX_BUDGET`. A reclassified refusal costs at most one bounded pass over the committee. The failure this fixes needs no attacker: two honest validators with different pool contents (a same-nonce replacement) or different admission config (`--rpc.txfeecap`, `--txpool.pricebump`, `--txpool.minimal-protocol-fee`) are sufficient.

Out of scope: the `-32003` arm still returns `TryNext` for the deterministic verdicts reth folds into its catch-all (for example insufficient funds). That retry pass is bounded, and the corrected `TRANSIENT_RPC_CODES` doc comment now states the mix.

## Changes

- [x] `crates/tn-reth/src/forward.rs`: add `NODE_LOCAL_MESSAGES` and a `TryNext` arm for the four node-local `-32000` refusal messages in `classify_server_error`, scoped to `INVALID_INPUT_RPC_CODE` (`-32000`).
- [x] `crates/tn-reth/src/forward.rs`: correct the `TRANSIENT_RPC_CODES` doc comment. `-32003` is `TxPoolOverflow` plus reth's catch-all for invalid-transaction variants with no explicit code, not only a full pool.
- [x] `crates/tn-reth/src/forward.rs`: align the `Disposition` variant docs and the forward-loop comment with the new split.
- [x] `crates/tn-reth/src/forward.rs`: add `classify_server_error_tries_next_on_node_local_refusals` with a case-insensitivity row and two negative controls: an adjacent chain-wide verdict (`gas required exceeds allowance`) and a carve-out substring under a different code (code 3).

## Testing

- `cargo +1.94 test -p tn-reth --lib -- forward::tests::classify_server_error` runs both classifier tests in an isolated target dir.
- Mutation checks, each seen to fail: the `NODE_LOCAL_MESSAGES` arm disabled fails the new test; the lowercase fold removed fails the case-insensitivity row; the scoping constant skewed to `-32001` fails the positive rows. Restored, all pass.
- Attest proxy on the pinned toolchains in `.tn-1168-iso`: `cargo +nightly-2026-03-20 fmt`, `cargo +nightly-2026-03-20 clippy --workspace` and `--workspace --all-features`, `cargo +1.94 test --no-run --workspace -j 2`.

The four message strings match the pinned reth `v1.11.3` source verbatim (`crates/rpc/rpc-eth-types/src/error/mod.rs`, checkout `d6324d63e`).

### DIFF
--- a/crates/tn-reth/src/forward.rs
+++ b/crates/tn-reth/src/forward.rs
@@ -92,15 +92,44 @@ const FORWARD_BATCH_BUDGET: Duration = Duration::from_secs(120);
 /// and a tighter cap would shed batches an unstressed node could have delivered.
 const MAX_CONCURRENT_FORWARDS: usize = 64;
 
-/// JSON-RPC error codes reth returns for a validator-local, transient condition (a full pool,
-/// `-32003`, or an internal/IO error, `-32603`) rather than a verdict on the transaction itself.
-/// The forward falls through to the next advertised validator on these, since another validator
-/// may still accept the transaction.
+/// JSON-RPC error codes reth reserves for conditions that are validator-local or indeterminate,
+/// never cleanly a verdict on the transaction: `-32003` is a full pool (`TxPoolOverflow`) plus
+/// reth's catch-all for invalid-transaction variants with no explicit code of their own (for
+/// example insufficient funds), and `-32603` is an internal/IO error. The forward falls through
+/// to the next advertised validator on these: retrying a deterministic verdict caught in the
+/// `-32003` mix costs at most one bounded pass over the committee, while stopping on a full pool
+/// would drop a transaction another validator may still accept.
 const TRANSIENT_RPC_CODES: [i64; 2] = [-32003, -32603];
 
 /// Substring of reth's `eth_sendRawTransaction` error message when the transaction is already in a
 /// validator's pool (`code -32000`). Treated as a successful delivery, not a failure to retry.
 const ALREADY_KNOWN_MESSAGE: &str = "already known";
+
+/// The JSON-RPC code reth answers with for both chain-wide verdicts and the node-local refusals
+/// in [`NODE_LOCAL_MESSAGES`] (`EthRpcErrorCode::InvalidInput`). The message carve-out is scoped
+/// to this code so a matching substring under any other code keeps its terminal meaning.
+const INVALID_INPUT_RPC_CODE: i64 = -32000;
+
+/// Substrings of reth `eth_sendRawTransaction` error messages (code `-32000`) that describe one
+/// validator's pool contents or one operator's admission config rather than a verdict on the
+/// transaction itself. Reth maps almost every pool refusal to `-32000`, the same code it uses
+/// for chain-wide verdicts, so the code axis cannot make this split; the message can. The
+/// forward tries the next advertised validator on these, since a validator with different pool
+/// contents or config may accept the same bytes.
+const NODE_LOCAL_MESSAGES: [&str; 4] = [
+    // `RpcPoolError::Underpriced`: priced below that operator's `--txpool.minimal-protocol-fee`.
+    "transaction underpriced",
+    // `RpcPoolError::ReplaceUnderpriced`: a same-nonce transaction already sits in that
+    // validator's pool and the fee bump is below its `--txpool.pricebump`. Subsumed by the
+    // entry above under `contains`, pinned separately so a reth rewording of one message
+    // cannot silently drop coverage of the other.
+    "replacement transaction underpriced",
+    // `RpcPoolError::ExceedsFeeCap`: the fee is above that node's `--rpc.txfeecap`.
+    "exceeds the configured cap",
+    // `RpcPoolError::AddressAlreadyReserved`: blob-vs-regular sender exclusivity against that
+    // validator's current pool contents.
+    "address already reserved",
+];
 
 /// Whether the forwarder may dial an advertised endpoint whose host is not a public internet
 /// address.
@@ -511,8 +540,9 @@ impl WorkerRpcForwarder {
                                 result = ForwardOutcome::Rejected(reason);
                                 break;
                             }
-                            // Endpoint-local problem (timeout, transport error, full pool): the
-                            // transaction's fate is unknown here, so try the next validator.
+                            // Endpoint-local problem (timeout, transport error, full pool, or a
+                            // node-local refusal): the transaction's fate is unknown here, so
+                            // try the next validator.
                             Disposition::TryNext(reason) => {
                                 debug!(
                                     target: "worker::forward",
@@ -683,10 +713,11 @@ enum Disposition {
     /// The validator accepted the transaction, or it was already in a validator's pool.
     Delivered,
     /// The validator returned a considered rejection of the transaction itself (bad nonce,
-    /// underpriced, invalid, wrong fork). No other validator will accept it either.
+    /// invalid, wrong fork). No other validator will accept it either.
     Rejected(String),
-    /// This endpoint gave no verdict (timeout, transport error, or a transient full pool); the
-    /// transaction may still be accepted by another validator.
+    /// This endpoint gave no verdict (timeout, transport error, a transient full pool, or a
+    /// refusal tied to this validator's own pool contents or admission config); the transaction
+    /// may still be accepted by another validator.
     TryNext(String),
 }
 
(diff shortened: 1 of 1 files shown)


grantkee (approved):
Good idea

PR #1172, opened 2026-08-13
fix(tn-reth): seed reth's process-global RPC defaults and disable IPC for temp chains

### LINKED ISSUE #1165: Temp-chain nodes still bind reth's /tmp/reth.ipc: reth's process-global RpcServerArgs default is never seeded
## Summary

#1155 aligned TN's `RpcServerArgs::default()` with the CLI default (`/tmp/tn.ipc`).  The IPC socket a temp-chain node actually opens does not go through TN's struct at all.  `RethEnv::new_for_temp_chain` fills its config from reth's `NodeConfig::default()`, whose `rpc` field is reth's own `RpcServerArgs`.  Reth's `Default` clones the process-global `DefaultRpcServerArgs`, and the workspace never initializes that global, so the config keeps reth's values: `ipcdisable: false`, `ipcpath: "/tmp/reth.ipc"`.

## Problem

- `crates/tn-reth/src/env/genesis.rs` line 124: `..NodeConfig::default()` fills the temp-chain config from reth defaults.
- The comment one screen up (lines 103 to 107) documents this exact process-global hazard for the txpool and seeds `init_txpool_defaults()` first.  There is no RPC analog anywhere in the workspace (zero hits for `DefaultRpcServerArgs`), so reth's global keeps reth's own default.
- `crates/tn-reth/src/env/rpc.rs` line 73: `start_rpc` builds the server config from `self.node_config().rpc`, the reth-side struct.
- `crates/node/src/engine/inner.rs` line 167 starts that RPC server unconditionally during worker component init, and per the module docs on `rpc_server_args.rs` the IPC server is on unless `--ipcdisable` is passed.

## Impact

Test environments only.  Production nodes parse the CLI, and TN's `From` impl carries TN's ipcpath into the reth-side config.  But every test env built through `new_for_temp_chain` (for example `default_test_execution_node`, `crates/test-utils/src/execution.rs` line 42) binds a unix domain socket at the fixed global path `/tmp/reth.ipc`, outside the test's `TempDir`, and leaves the file behind.  Two concurrent test runs on one host race for the same path, and a locally running reth node collides with the test suite.  After #1155 the divergence is also confusing on its face: the unit tests assert the default IPC endpoint is TN's, while the only construction path that opens a socket without CLI parsing still opens reth's.

## Fix

Two candidate remediations, in the spirit of the existing txpool seeding:

1. Seed reth's global next to `init_txpool_defaults()`: initialize `DefaultRpcServerArgs` with TN's `DEFAULT_IPC_ENDPOINT` (and consider `ipcdisable: true` for throwaway chains).  Ordering matters for the same reason the pool comment gives: the global is get-or-init, and any `RpcServerArgs::default()` call locks it, so the seed must run first.
2. Alternatively, set the `rpc` field explicitly in the temp-chain `NodeConfig` instead of taking it from `..NodeConfig::default()`, with `ipcdisable: true` or a per-tempdir ipcpath.

A regression test can pin the invariant: build a temp-chain `RethEnv` and assert its node config's ipcpath is not reth's `/tmp/reth.ipc` (or that `ipcdisable` is set).

## Credit

Surfaced during adversarial review of #1164, the remediation for #1155.

MavenRain:
Closes #1165.

## Problem

`RethEnv::new_for_temp_chain` fills its config from `NodeConfig::default()`, and reth's `RpcServerArgs::default()` clones a process-global `DefaultRpcServerArgs` that the workspace never seeded. The global therefore kept reth's own values (`ipcdisable: false`, `ipcpath: "/tmp/reth.ipc"`), and the worker RPC server starts unconditionally during component init. Every test env built through `new_for_temp_chain` (for example `default_test_execution_node`) bound a unix domain socket at that fixed global path: outside the test's `TempDir`, left behind after the run, raced by any concurrent test run on the same host, and colliding with a locally running reth node. No attacker required. #1164 (the #1155 remediation) fixed TN's own `RpcServerArgs::default()`, but the temp-chain construction path never goes through TN's struct.

## Fix

- [x] `crates/tn-reth/src/cli.rs`: rename `init_txpool_defaults` to `init_reth_defaults` and seed reth's `DefaultRpcServerArgs` (ipcpath = TN's `DEFAULT_IPC_ENDPOINT`) next to the existing txpool seeding, in the spirit of that precedent. The ordering contract is the same one the pool comment documents: the global is get-or-init and any reth-side `RpcServerArgs::default()` locks it, so the seed must run first. Renaming the one choke point means every existing call site (both CLI parse entry points, e2e, test-utils, temp chains) now seeds both globals with no new call to forget.
- [x] `crates/tn-reth/src/env/genesis.rs`: `new_for_temp_chain` additionally sets `rpc: RpcServerArgs { ipcdisable: true, ..Default::default() }` explicitly. Seeding alone would only move the collision from `/tmp/reth.ipc` to `/tmp/tn.ipc` (and make the test suite race a production TN node instead of a reth one). A throwaway chain parses no CLI and nothing in the workspace connects to one over IPC (the tests that exercise IPC spawn nodes through the CLI with a per-tempdir `--ipcpath`), so it must not bind any fixed global path at all.
- [x] Call-site comments, module docs, and the `lib.rs` re-export updated for the rename across `telcoin-network-cli`, `e2e-tests`, `test-utils`, and `tn-reth`.

## Testing

- [x] New regression test `test_temp_chain_disables_ipc` (tn-reth `env/genesis.rs` tests): a temp-chain env's node config has `ipcdisable` set. It pins the order-independent half of the invariant; the doc comment explains why the path itself is not asserted there (which path the global holds depends on which test in the binary reads reth's RPC defaults first). Confirmed by mutation: reverting the `rpc:` field fails it.
- [x] New integration test binary `crates/tn-reth/tests/reth_defaults.rs`, owning its process for the same reason `telcoin-network-cli/tests/txpool_defaults.rs` does: after seeding, a reth-side `RpcServerArgs::default()` carries TN's `DEFAULT_IPC_ENDPOINT`, differs from reth's hard-coded `/tmp/reth.ipc`, and keeps the IPC server enabled for parsed nodes (only temp chains disable it). Confirmed by mutation: dropping the `with_ipcpath` seed fails it.
- [x] `cargo +1.94 test -p tn-reth --lib` (146 passed), `cargo +1.94 test -p tn-reth --test reth_defaults`, and `cargo +1.94 test -p telcoin-network-cli --test txpool_defaults` green on the pinned toolchain.
- [x] `cargo +nightly fmt -- --check` and `cargo +1.94 clippy --all-targets` green in an isolated target dir; `make attest` run before push.

### DIFF
--- a/crates/e2e-tests/src/lib.rs
+++ b/crates/e2e-tests/src/lib.rs
@@ -14,7 +14,7 @@ use std::{
 use telcoin_network_cli::{genesis::GenesisArgs, keytool::KeyArgs, node::NodeCommand, NoArgs};
 use tn_config::{Config, ConfigFmt, ConfigTrait, KeyConfig, Parameters};
 use tn_node::launch_node;
-use tn_reth::init_txpool_defaults;
+use tn_reth::init_reth_defaults;
 use tn_types::{test_utils::CommandParser, Address, Genesis, GenesisAccount};
 use tracing::{error, info};
 // unused deps warnings
@@ -319,8 +319,9 @@ pub fn spawn_local_testnet(
             ipc_path: ipc_path_str.clone(),
         });
 
-        // seed reth's pool defaults before the parse resolves `--txpool.max-account-slots`
-        init_txpool_defaults();
+        // seed reth's process-global defaults before the parse resolves
+        // `--txpool.max-account-slots`
+        init_reth_defaults();
         let command = NodeCommand::parse_from([
             "tn",
             "--http",
--- a/crates/telcoin-network-cli/src/cli.rs
+++ b/crates/telcoin-network-cli/src/cli.rs
@@ -9,7 +9,7 @@ use clap::{Parser, Subcommand};
 use std::{ffi::OsString, fmt, path::PathBuf, str::FromStr};
 use tn_config::KeyConfig;
 use tn_node::engine::TnBuilder;
-use tn_reth::{dirs::DEFAULT_ROOT_DIR, init_txpool_defaults, LogArgs};
+use tn_reth::{dirs::DEFAULT_ROOT_DIR, init_reth_defaults, LogArgs};
 use tokio::{runtime::Builder, task::JoinHandle};
 use tracing::info_span;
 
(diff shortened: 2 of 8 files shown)


grantkee (approved):
Good work

PR #1175, opened 2026-08-13
fix(tn-reth): hold a forwarded tx rejection until a second validator confirms it

### LINKED ISSUE #1167: Observer forwarding: a single byzantine validator can silently censor an account's transactions
## Summary

The observer transaction forwarder in `crates/tn-reth/src/forward.rs` sends each transaction first to the one validator that owns the sender's committee slot.  It then stops the fallback chain as soon as that validator reports `Delivered` or `Rejected`.  A byzantine owning validator can report either result without doing the work.  So it can drop the transaction and stop the observer from trying any honest validator.  This censors every observer-forwarded transaction from the accounts that map to that validator's slot.

This is a low-severity hardening note.  It does not break a stated guarantee.  The project already records that a byzantine committee member can censor and keep full committee power (`reputation.md`, lines 231-235).  The purpose of this issue is to name a sharper and silent form of that gap on the observer-forward path, and to suggest small hardening steps.

Reference: `origin/main` at `09ee8f61` (2026-08-12).  The forwarding logic is unchanged from the version under review.

## Threat model context

Byzantine committee validators are in scope.  The consensus code assumes the standard BFT bound (f < n/3).  The committee is a closed, permissioned, staked set of well-known operators.  The project does not claim per-transaction or per-account censorship resistance anywhere.  `crates/consensus/primary/src/consensus/reputation.md:231` states the gap directly:

> Reputation measures liveness only.  Byzantine faults are filtered before scoring (Section 1), and the sole consequence of a low score is losing leader turns (Section 3).  A byzantine node that equivocates, censors, or selectively withholds votes but still gets its certificates committed retains a maximal score and full committee power.

So a byzantine validator that censors is a known and accepted condition at the consensus layer.  This issue is about how the observer-forward path makes that condition worse for a targeted account.

## Mechanism

`owning_validator` maps the sender to one committee slot.  The forwarder tries that owner first, then the other advertised validators as fallbacks:

`crates/tn-reth/src/forward.rs:476`
```rust
let ordered = owner.into_iter().chain(fallbacks.iter().cloned());
```

The per-transaction loop stops on the first `Delivered` or `Rejected`:

`crates/tn-reth/src/forward.rs:507` and `:511`
```rust
Disposition::Delivered => {
    result = ForwardOutcome::Delivered;
    break;
}
Disposition::Rejected(reason) => {
    result = ForwardOutcome::Rejected(reason);
    break;
}
```

A byzantine owner controls both results:

1. **Silent drop.**  The owner accepts the connection and returns any success from `eth_sendRawTransaction`.  An `Ok` maps straight to `Delivered`:

   `crates/tn-reth/src/forward.rs:500`
   ```rust
   |res| res.err().map_or(Disposition::Delivered, classify_error),
   ```
   The observer records success.  It writes no warning.  It never tries a fallback.  The transaction never reaches an honest validator.

2. **Fake rejection.**  The owner returns any JSON-RPC error whose code is not `-32003` or `-32603`, and whose message does not contain "already known".  `classify_server_error` then returns `Rejected`, which stops the chain:

   `crates/tn-reth/src/forward.rs:99` and `:103`
   ```rust
   const TRANSIENT_RPC_CODES: [i64; 2] = [-32003, -32603];
   const ALREADY_KNOWN_MESSAGE: &str = "already known";
   ```

The observer cannot tell an honest result from a false one.  It does no cross-check with a second validator.

The design comment states the trust assumption directly:

`crates/tn-reth/src/forward.rs:19`
> A considered rejection stops the fallback chain.  If a validator's RPC rejects the transaction itself (bad nonce, underpriced, invalid), no further validators are tried: all validators share consensus state, so the rejection would repeat everywhere.

This holds only if the owner is honest.  A byzantine owner breaks it.

## Why the fallback does not help

The fallback chain protects availability, not integrity.  It helps when the owner has not advertised an endpoint, or is unreachable.  It does not help against an owner that answers and lies.

The observer-forward path is the only path for an observer's accepted transactions.  Observers do not gossip their own batches.  The forward replaces the old libp2p broadcast (`crates/consensus/worker/src/worker.rs:240`, issue #804).  Observers also do not subscribe to the worker batch topic (`crates/consensus/worker/src/network/handler.rs:142`, issue #960).  A committee validator publishes its batch through `publish_batch` (`crates/consensus/worker/src/worker.rs:330`).  An observer never reaches that path.  So a dropped forward has no second delivery route.

Routing is a pure function of the sender and the committee.  So a user cannot avoid a byzantine owner by choosing a different observer.  Every observer routes that account to the same validator.

## Impact and severity

- **Class:** targeted censorship and liveness loss.  Not a fund loss.  Not a consensus-safety break.
- **Precondition:** one byzantine committee validator.
- **Scope:** transactions that enter through observers, for the accounts whose sender maps to that validator's slot.
- **Duration:** while that validator owns that sender's slot.  The slot can change when the committee rotates at an epoch boundary.
- **Signal:** the silent-drop variant produces no operator-visible warning.  The observer logs success.

Severity is low.  Users keep other routes, such as direct submission to a validator RPC.  The base gap is already documented.  The two new points are:

1. One assigned validator becomes the sole gatekeeper for an account.  This is stronger than the general BFT censorship gap, which needs a threshold to target one sender.
2. The silent-drop variant leaves no trace on the observer.

## Suggested hardening (each has a cost)

1. **Do not let one owner's `Rejected` stop the chain.**  Treat the owner's `Rejected` as advisory.  Try at least one other validator before you conclude a rejection.  Or require the same considered rejection from a second validator.  Cost: more RPC load.  It weakens the "the rejection would repeat everywhere" optimization.

2. **Reduce the silent-drop risk.**  An honest success and a false success look the same at the RPC layer.  So the only defenses are redundancy or verification:
   - Forward to more than one validator, not the owner alone.
   - Or check later that the transaction appears in a committed batch, and re-forward if it does not.

   Cost: more RPC load, or added inclusion-tracking state.

3. **If the team accepts the current behavior, record the byzantine case in the code.**  Add a note next to the design comment at `crates/tn-reth/src/forward.rs:19`.  The comment now reads as if it holds under byzantine validators.  It does not.

## Notes and scope

- This angle appears un-raised.  The related dial-target trust boundary was raised and fixed for SSRF (#1092 and #1114).  The byzantine-rejection angle was not.
- The sender-routing itself is intentional and correct for nonce ordering (`crates/tn-reth/src/forward.rs:8`).  This issue does not propose to remove it.

MavenRain:
Closes #1167.  (Cantina #10)

## Problem

The observer transaction forwarder sends each transaction to the validator that owns the sender's committee slot.  The fallback chain stops on the first `Delivered` or `Rejected`.  A byzantine owning validator controls both results.  It can return a fabricated rejection, or a fabricated success, and the observer never tries an honest validator.  Sender routing makes that one validator the sole gatekeeper for every account that maps to its slot, and every observer routes those accounts to it.  The silent-drop variant leaves no operator-visible signal.

## Root cause

The stop-on-rejection rule in `forward.rs` assumed every verdict is honest: "all validators share consensus state, so the rejection would repeat everywhere."  That holds only for honest validators.  The design comment did not state the byzantine case, and the code gave one validator's word final authority over the transaction's fate.

## Fix

A considered rejection is now advisory until a second validator confirms it.

- The per-transaction walk (`walk_fallback_chain`, extracted from `spawn_forward`) holds the first `Rejected` verdict and keeps walking.
- A second considered rejection from a different validator confirms it, and the chain stops as before.  Honest rejections cost at most one extra RPC round-trip, and only on the rejection path.  That path is rare: every forwarded transaction already passed this observer's own pool validation, so a considered rejection normally means a nonce race, not spam.
- A later delivery overrides the held rejection.  The override is logged with the rejecting validator's key and counted in a new `tn_reth.forwarded_rejections_overridden_total` series, so a lying or state-divergent validator is visible to operators.
- A rejection that ends the chain unconfirmed (no other validator reachable to weigh in) still stands.  The same rule survives the per-transaction budget: the held verdict is banked in a slot the caller reads back when the timeout cancels the walk mid-chain, so a fast rejection followed by unresponsive fallbacks is tallied as rejected, not as unreached or abandoned (both of whose docs claim the transaction was never handed to a validator).

Threat model: byzantine committee members are in scope (f < n/3), and `reputation.md` already records that a censoring validator keeps full committee power.  This change does not add a quorum guarantee.  It raises the collusion a fabricated-rejection censorship needs from one validator to two, and it makes verdict divergence observable.  The per-transaction cost stays bounded: the walk still visits each advertised validator at most once under the same `FORWARD_TX_BUDGET`, so the worst case adds one verdict-bearing round-trip, not a new unbounded path.  The symmetric fabrication (answer success, drop the transaction) is deliberately not defended: it is indistinguishable from an honest delivery at this boundary, and catching it would cost redundant delivery or inclusion tracking on every honest-path forward.  The module docs now record that residual trust next to the routing rationale, as the issue asked.

## Changes

- [x] `crates/tn-reth/src/forward.rs`: extract the per-transaction fallback walk into `walk_fallback_chain`;  hold the first considered rejection until a second validator confirms or contradicts it;  log and count contradictions;  bank the held verdict in a slot `expired_walk_outcome` reads back when the budget cancels the walk;  update the module boundary docs and the `Disposition` / `ForwardOutcome` docs.
- [x] `crates/tn-reth/src/metrics.rs`: add `tn_reth.forwarded_rejections_overridden_total`, registered eagerly in `ForwarderMetrics::init` like the shed / abandoned series so the series exists from node start.

## Testing

- New deterministic unit tests drive `walk_fallback_chain` against scripted local JSON-RPC endpoints (canned verdict payloads, request ids echoed, one request per connection so per-validator hit counts are exact):
  - `a_lone_rejection_is_overridden_by_a_later_delivery`: the issue's attack;  the fallback is consulted, delivery wins, and the override counter reads 1.
  - `a_second_rejection_confirms_and_stops_the_walk`: two rejections stop the chain, and the accepting validator behind them is never dialed.
  - `a_sole_unconfirmed_rejection_still_stands`: one rejection plus an unreachable fallback still reports `Rejected`.
  - `a_transient_error_does_not_confirm_a_rejection`: a `-32003` between the two verdicts neither confirms nor stops the walk.
  - `a_budget_cut_walk_still_reports_the_held_rejection`: an expired budget reads the banked verdict back (and, with nothing banked, still reports `NoEndpointReached`).
- Each new behavior was confirmed by mutation: reverting to stop-on-first-rejection fails the override and transient tests;  erasing the held rejection at chain end fails the sole-rejection test;  dropping the budget read-back fails the budget-cut test;  dropping the slot write fails the sole-rejection slot assertion;  the suite was restored to green after each mutation.
- `make attest` green on the pinned 1.94 toolchain (fmt + nightly clippy + workspace test build).


sstanfield (dismissed):

grantkee (dismissed):
LGTM! @MavenRain thank you for consistently breaking issues into separate PRs. I noticed that this one addresses another bug for the forwarder so there's merge conflicts with #1153 

I'm in favor of addressing residual, related issues in the same PR with separate issues. What do you think? 

For example, this patch was a single commit. It's easy to isolate during the review process as a separate commit and provides full context for reviewing the complete solution. As it stands now, the code that changed in this PR may be affected by upstream #1153 after it's merged and will need rereview. I think this could speed up the merge process and reduce review fatigue, but interested to know any objections to this idea. Tagging @sstanfield for input as well

MavenRain:
> LGTM! @MavenRain thank you for consistently breaking issues into separate PRs. I noticed that this one addresses another bug for the forwarder so there's merge conflicts with #1153
> 
> I'm in favor of addressing residual, related issues in the same PR with separate issues. What do you think?
> 
> For example, this patch was a single commit. It's easy to isolate during the review process as a separate commit and provides full context for reviewing the complete solution. As it stands now, the code that changed in this PR may be affected by upstream #1153 after it's merged and will need rereview. I think this could speed up the merge process and reduce review fatigue, but interested to know any objections to this idea. Tagging @sstanfield for input as well

I think that, wherever this situation can easily be identified ahead of time, this kind of a merge is a good idea.  In this particular case, the issues were addressed far enough apart that the earlier PR could have merged before the later PR goes through review.  Although, in that case, the issue of a merge conflict is moot.

Before I open a PR, I can do a quick check to see how related to an earlier open PR that the one I am about to open is, and merge under certain conditions.  Those conditions are necessarily fuzzy, so we can try it for a time and, if PR size and scope starts to get unruly, we could always revert to keeping them separate again.

grantkee:
Good point @MavenRain - you are crushing the PRs and sometimes it's hard to keep up! I'll continue to try and improve on my own review turnaround times. Keep up the great work 🙌

grantkee (approved):

Haxatron:
This PR reduces the censorship risk by forwarding to a second validator if the first fails, but does not completely eliminate it. A byzantine validator could also return a false acceptance and the prevent the second forwarding from occuring

PR #1176, opened 2026-08-13
fix: enforce --rpc.txfeecap on RPC transaction submission

### LINKED ISSUE #1160: The --rpc.txfeecap flag is parsed and documented as a spend guard but never reaches any enforcement point
## Summary

The `--rpc.txfeecap` flag is parsed, documented as an active spend guard, and forwarded into `reth::args::RpcServerArgs`. But no code consumes the value. The node hand-builds its transaction pool and never calls `set_tx_fee_cap`. The result: the operator-facing flag is dead configuration, and the module doc claims a protection that does not exist. Verified at `origin/main` (1c73e8a3) against the pinned reth v1.11.3 source.

Surfaced by Haxatron's feedback on the pool construction in the collaborative review; this issue records the verified mechanism.

## Details

1. The flag is documented as an enforced spend guard. `crates/tn-reth/src/rpc_server_args.rs:16`:

   > `- Spend: transactions submitted through this RPC are refused above a fee cap (--rpc.txfeecap, default 1.0 in native-token units; 0 disables the cap).`

   It is parsed at `rpc_server_args.rs:175-183` (default `1.0`), defaulted at line 238 (`DEFAULT_TX_FEE_CAP_WEI`), and forwarded verbatim into reth's `RpcServerArgs` at line 273.

2. Nothing consumes the forwarded value. In reth v1.11.3 the only enforcement consumer of `rpc_tx_fee_cap` is `EthereumPoolBuilder`, which threads it into the pool validator (`crates/ethereum/node/src/node.rs:495`, `.set_tx_fee_cap(ctx.config().rpc.rpc_tx_fee_cap)`). TN does not use `EthereumPoolBuilder`. TN builds its own validator in `crates/tn-reth/src/txn_pool.rs:113-120` and never calls `set_tx_fee_cap` (`git grep set_tx_fee_cap` over the tree: zero hits). There is no RPC-layer consumer in reth either; `eth_config()` never reads the field.

3. The fallback default cannot protect production traffic. `EthTransactionValidatorBuilder::new` defaults `tx_fee_cap` to 1 ether (reth `crates/transaction-pool/src/validate/eth.rs:973`), but the check (eth.rs:418-439) runs only when `local_transactions_config.is_local(origin, sender)` is true. Every production ingress path submits with `TransactionOrigin::External`: reth's default `send_raw_transaction` tags `External` (`crates/rpc/rpc-eth-api/src/helpers/transaction.rs:79-88`), and TN's `add_transaction_local` / `add_transaction_and_subscribe_local` are called only from `crates/tn-reth/src/test_utils.rs`. So no fee cap applies to any production submission, at any flag value.

## Impact

Informational / configuration correctness. An operator who sets `--rpc.txfeecap` gets no cap, silently. The module doc asserts a spend protection that does not hold. No consensus or fund-safety impact.

## Suggested fix

Decide the intended semantics first, because plain wiring is not sufficient:

- To enforce the cap: pass the parsed value into `WorkerTxPool::new` and call `.set_tx_fee_cap(value)` on the validator builder. Then also decide whether the cap must cover `External` submissions, because reth's check is local-origin only. Covering public RPC traffic needs either a local-address configuration or a TN-side check on the `add_raw_transaction_external` path.
- If the cap will not be enforced: remove the flag and the `Spend:` claim from the module doc, so operators are not misled.

Either way, add a test that submits an over-cap transaction through the production ingress path and asserts the intended outcome.

## References

- `crates/tn-reth/src/rpc_server_args.rs:16,175-183,238,273` (origin/main 1c73e8a3)
- `crates/tn-reth/src/txn_pool.rs:113-120` (validator construction, no cap call)
- reth v1.11.3 `crates/ethereum/node/src/node.rs:489-499` (reference wiring), `crates/transaction-pool/src/validate/eth.rs:418-439,973` (check + default)

MavenRain:
Closes #1160.  (Cantina #12)

## Problem

`--rpc.txfeecap` is parsed and documented as a spend guard, but no code path enforces it. TN builds its transaction pool by hand and never calls `set_tx_fee_cap`, so the validator keeps reth's hardcoded default of 1 ETH. That default also cannot fire in production: reth's validator checks the fee cap only for transactions it treats as local (`LocalTransactionConfig::is_local`), and raw RPC submissions arrive with `TransactionOrigin::External`. An operator who sets a strict cap still accepts transactions with an arbitrarily large fee through `eth_sendRawTransaction`. No attacker is required: one honest fee typo through public RPC is enough to spend funds the flag promises to guard.

## Root cause

Two half-wired guards:

- The pool validator has a fee-cap check, but TN never passes the parsed flag to the pool builder, and the check runs only for local-treated transactions.
- The production RPC ingress is reth's stock `eth_sendRawTransaction` / `eth_sendRawTransactionSync`, which submit with `External` origin, so the validator check never runs for public RPC traffic.

## Fix

Enforce the flag at both points. The RPC guard is cheap-first: it decodes the raw bytes and does saturating integer arithmetic before any pool or signer-recovery work, and it delegates to the unchanged reth handlers when the transaction is under the cap.

- [x] `crates/tn-reth/src/txn_pool.rs`: wire `.set_tx_fee_cap(node_config.rpc.rpc_tx_fee_cap)` into the validator builder. This matches upstream reth node construction and covers Local-origin submissions, for example `--txpool.locals` senders.
- [x] `crates/tn-reth/src/rpc_fee_cap.rs` (new): a guard that replaces `eth_sendRawTransaction` and `eth_sendRawTransactionSync` on the built server. It decodes the raw bytes without signer recovery, computes the same quantity the validator compares (`max_fee_per_gas * gas_limit`, plus the blob fee bound for EIP-4844 transactions, in saturating `U256` arithmetic), and refuses an over-cap transaction with reth's own error object (`RpcPoolError::ExceedsFeeCap`). Both methods need replacement because the sync variant calls the trait method internally, not RPC dispatch.
- [x] `crates/tn-reth/src/env/rpc.rs`: install the guard with `add_or_replace_if_module_configured(RethRpcModule::Eth, ...)`, which respects the operator's transport module selection. `get_rpc_server` now returns `eyre::Result`: the node must not start serving submission methods without the configured spend guard.
- [x] `crates/node/src/engine/inner.rs`: propagate the new error.
- [x] `crates/tn-reth/src/rpc_server_args.rs`: document both enforcement points on the flag.
- [x] `crates/tn-reth/src/env/genesis.rs`, `crates/tn-reth/src/worker.rs`: test seams (`new_for_temp_chain_with_rpc_args`, `WorkerNetwork::new_for_test`). Default behavior is unchanged.
- [x] `crates/tn-reth/Cargo.toml`, `Cargo.lock`: jsonrpsee `server`/`macros` features and `async-trait` for the guard trait.

Semantics: the flag value parses in native-token units (`parse_ether_value`, default 1.0) and `0` disables the check. This matches the flag's documented semantics and the validator's own `Some(0) | None` skip. `eth_sendTransaction` is not touched: it submits with Local origin, so the validator wiring covers it (and TN builds `EthApi` with no signers, so the method cannot currently submit at all). `--txpool.nolocals` empties the local class, which leaves only the RPC-boundary check; the flag docs now say so. One documented divergence from reth: a transaction that is both over the cap and carries an invalid signature returns the fee-cap error, where reth reports the signature error first. A consequence is that a client can learn the configured cap without a funded key; the cap is operator configuration, not a secret.

Operational note: committee validators should run a uniform cap. The transaction forwarder treats one validator's fee-cap refusal as final for that transaction, so a lax edge node in front of a strict committee can accept a transaction the committee then silently drops. The flag docs state this; the forwarder-side classification is a follow-up issue, since `forward.rs` has several PRs in flight.

## Testing

All commands ran on the pinned toolchain (1.94).

- `cargo +1.94 test -p tn-reth --lib`: 10 new tests.
  - Unit (`rpc_fee_cap`): over-cap refusal with the exact reth error string, fee equal to cap accepted (boundary), zero cap disables, empty bytes and undecodable bytes return reth's exact errors, EIP-4844 blob fee counted (393,216 wei blob bound on a 147,000 wei gas fee).
  - Integration (`env::rpc`): built through `get_rpc_server` (the production registration) and called via the registered `eth_sendRawTransaction` / `eth_sendRawTransactionSync` methods. Under-cap transaction lands in the pool; over-cap transaction is refused with the reth error message and never reaches the pool; zero cap accepts a 7,000,000 wei fee. A separate case serves HTTP alone (IPC disabled) and confirms the guard on the HTTP registration, the transport the forwarder uses.
  - Validator wiring (`txn_pool`): a local submission over the cap is rejected with `ExceedsFeeCap`.
- Confirm-by-mutation, each seam seen to fail: removed the `set_tx_fee_cap` wiring (validator test fails), forced the RPC guard to always pass (integration test fails), flipped the boundary `<=` to `<` (boundary unit test fails), dropped the cap from the HTTP-only registration (HTTP transport test fails). Restored after each run.
- Attest battery: nightly `fmt --check`, `cargo +1.94 clippy --all-targets`, nightly `clippy --workspace --all-features`, and `cargo +1.94 test --no-run --workspace -j 2` in an isolated target dir.


sstanfield (dismissed):

crates/tn-reth/src/rpc_fee_cap.rs
```diff
@@ -0,0 +1,249 @@
...
+    Api: EthTransactions
+        + LoadReceipt
+        + EthApiTypes<NetworkTypes = alloy::network::Ethereum>
+        + Clone
+        + Send
+        + Sync
+        + 'static,
+    jsonrpsee::types::ErrorObject<'static>: From<<Api as EthApiTypes>::Error>,
+{
+    async fn send_raw_transaction(&self, bytes: Bytes) -> RpcResult<B256> {
```
grantkee, line 134: I like this. There are some other eth api methods we likely want to intercept given the nature of TN BFT consensus

grantkee (commented):
In general, I fully support this feature. However, I'm concerned about setting the default to 1.0 native token. At current prices, that is $0.0019 - I think `0` (disabled) is better for now until we have data

crates/tn-reth/src/rpc_fee_cap.rs
MavenRain: I agree.  I filed the audit as #1231.  Methods that diverge today: `eth_feeHistory`'s next-block base fee entry (reth steps EIP-1559;  TN base fees are per-epoch from `WorkerConfigs`), `eth_syncing` (the network shim hardcodes `false`), the blob-fee surface (TN execution ignores EIP-4844 but the handlers still quote blob fees), and the pending-block env placeholders.  The issue also pins the interception pattern from this PR: exact method names, install via `add_or_replace_if_module_configured`, repeat reth's request-trace line, reuse reth's error objects, delegate for everything unchanged.  `eth_gasPrice`, the `safe`/`finalized` tags, and the uncle methods checked out correct as-is.

MavenRain:
The default is now `0` (cap disabled).  Reth's `1.0` was calibrated for ether;  1.0 TEL is about $0.002 today, so as a default it is not a useful bound.

The change covers both default paths: the clap `default_value` and the manual `Default` impl.  A mirror-parity test pair (like the existing `ipcpath` pair) keeps the two from drifting apart or back to reth's constant.  The flag doc, the module threat-model doc, and the CLI README now state the `0` default.  Operators can opt in with `--rpc.txfeecap <value>` once we have fee data.

grantkee (approved):
LGTM!

Haxatron:
Fix verified. The fee cap is now applied for all locally submitted transactions and any submissions via external RPC methods (`eth_sendRawTransaction*`) to reject any transactions exceeding the cap before reaching the mempool.

Note that for remote external RPC transactions, stock Reth 1.11.3 does not check fee cap as well, it was eventually fixed here https://github.com/paradigmxyz/reth/pull/25412, pushed to Reth 2.4.x

PR #1180, opened 2026-08-13
fix(tn-reth): rotate the observer forward fallback dial order

### LINKED ISSUE #1173: forward: fallback dial order is a deterministic BLS-pubkey byte sort with no per-node rotation, which concentrates overflow traffic on the lowest-keyed validator
## Classification
- Severity: Low. This is a fairness and liveness item, with a soft-censorship surface during an outage window.
- Component: `crates/tn-reth/src/forward.rs`.
- Class: networking fairness and retry concentration.
- Anchored to `origin/main` at commit `31cc4e90`.

## Summary
When an observer forwards a transaction, it dials the transaction's owning validator first. If the owner is not reached, the observer falls back to every other committee endpoint in order. The fallback order is the `BTreeMap<BlsPublicKey, _>` key order. That order is a raw byte-lexicographic sort of the 96-byte BLS public key. There is no per-node or per-epoch rotation or shuffle on this path. So every observer uses the same fallback order for every transaction.

## Effect
All observers share one deterministic fallback order. The lowest-keyed reachable validator is always the first fallback. When the owner of some account's transactions is not reached, which is an ordinary partial outage, every observer's redirected traffic for that owner converges on the same single validator at the same time. A validator can also grind its BLS keypair offline, so its key sorts first, to make itself the permanent first fallback.

## Honest scope
Do not overstate this item. The fallback fires only when the owner is not reached. The loop stops at the first delivery or considered rejection (`forward.rs:503-513`). So the sink absorbs the redirect share of one downed owner, which is about `1/committee_size` of the load, not all traffic. It also still delivers the transaction. There is no fund loss and no consensus-safety impact. The concern is fairness, because one node carries a disproportionate retry load, and a soft-censorship surface during an outage, because a malicious first fallback that lies through `classify_error` could drop redirected transactions. See the related classification issue #1168 and the observer-censorship issue #1167.

## Current behavior (code)
The fallback list is the map's keys, in ascending key order (`forward.rs:446-448`), over `BTreeMap<BlsPublicKey, RootProvider>` (`forward.rs:444`):

```
446	        // Fallback order: every usable endpoint, so a transaction whose owning validator has
447	        // not advertised (or is unreachable) can still reach the committee.
448	        let fallbacks: Vec<BlsPublicKey> = providers.keys().cloned().collect();
```

The dial order is owner first, then the fallbacks in ascending key order (`forward.rs:475-484`):

```
475	                let owner = owning_validator(tx, committee_size, &committee_slots);
476	                let ordered = owner.into_iter().chain(fallbacks.iter().cloned());
...
484	                    for key in ordered {
```

`BlsPublicKey` ordering is raw byte-lexicographic on the 96-byte pubkey (`crates/types/src/crypto/bls_public_key.rs:158-162`, backed by the array `cmp` at `:34-38`):

```
158	impl Ord for BlsPublicKey {
159	    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
160	        self.as_ref().cmp(other.as_ref())
161	    }
162	}
```

There is no rotation or shuffle on this path. The only `rand` import is test-only (`forward.rs:727-731`, inside `#[cfg(test)] mod tests`). A file-wide search returns zero `shuffle` hits and zero `rotate` hits.

## Suggested fix
Rotate the fallback tail per node and per epoch, so redirected traffic spreads across the committee. Keep the owner-first behavior. Only the fallback tail changes. One low-cost, total (no panic on empty) form:

```
let start = (epoch as usize).wrapping_rem(fallbacks.len().max(1));
let rotated = fallbacks
    .iter()
    .cycle()
    .skip(start)
    .take(fallbacks.len())
    .cloned();
let ordered = owner.into_iter().chain(rotated);
```

Seed `start` (or a shuffle) from a value the attacker cannot pin by key-grinding, for example `hash(node_id, epoch)` rather than the raw key order alone. A per-epoch rotation already breaks the fleet-wide convergence, because different observers no longer share one first fallback for the same failed owner.

MavenRain:
Closes #1173.

## Problem

An observer forwards each transaction to the validator whose committee slot owns the sender.  When the owner has not advertised an endpoint, or is unreachable, the forwarder dials the remaining committee endpoints in a fallback order.  That order is the iteration order of a `BTreeMap` keyed by BLS public key, so it is the byte sort of the committee's keys and identical on every observer.  All redirected traffic converges on the lowest-keyed reachable validator.

Two consequences follow.  No attacker is required for the first: any owner outage sends every observer's redirects to the same single validator, a load hotspot the committee never chose.  The second needs one dishonest validator: BLS keypair generation is cheap, so a validator can grind keys offline until its key sorts below every other committee key, and the first-fallback position is then permanent.  The holder sees redirected transactions first, which is a standing censorship and front-running vantage over exactly the traffic whose owner is down.

## Fix

`WorkerRpcForwarder` now carries a node-wide rotation counter (`Arc<AtomicU64>`).  The counter starts at a random per-process value and advances once per spawned forward.  Each forward rotates the fallback list left by the counter reduced modulo the list length, so consecutive forwards start their dial walk at consecutive endpoints.

Owner-first routing is unchanged: the rotation only reorders the tail that is tried after the owner, so nonce ordering per sender account is unaffected.  The rotation is independent of key bytes, so key grinding cannot buy a position in it.  The random seed means observers restarted together do not share one phase.  The seed provides fairness, not secrecy: predicting it moves no trust boundary, because every endpoint in the list is a committee member already authorized to receive the transaction.

## Changes

- [x] `crates/tn-reth/src/forward.rs`: add the `fallback_rotation` counter to `WorkerRpcForwarder`, seeded from process entropy in `new()`.
- [x] `crates/tn-reth/src/forward.rs`: rotate the fallback dial order once per spawned forward via the total helpers `rotated_fallbacks` and `rotation_start` (modulo guarded against an empty list, `u64::MAX` wraps cleanly).
- [x] `crates/tn-reth/src/forward.rs`: module docs state the rotation invariant.

## Testing

- `rotation_start_wraps_and_is_total`: modulo reduction, `u64::MAX`, single-entry, and empty-list cases.
- `rotated_fallbacks_rotates_left_without_losing_entries`: each rotation is a permutation of the input, no entry lost or duplicated, across counters 0 through 3 and the empty list.
- `spawned_forwards_rotate_the_first_fallback_dialed`: three local counting JSON-RPC endpoints paired to a sorted committee, one unrecoverable transaction per forward so no owner is ever dialed, and the per-endpoint hit counts must move `[1,0,0]` then `[1,1,0]` then `[1,1,1]` then `[2,1,1]` across four forwards.  The first dial provably rotates.  Completion is awaited by draining the forward semaphore, not by sleeping.
- Both regression tests were confirmed by mutation.  Discarding the counter reproduces the reported concentration (`[2,0,0]`, every forward dials the lowest-keyed endpoint) and fails both rotation tests.  Never advancing the counter fails the endpoint test.
- Gates on the pinned toolchain: nightly `cargo fmt --check`, nightly `cargo clippy --workspace --all-features -- -D warnings`, `cargo +1.94 test --no-run --workspace` in an isolated target dir, and the full `tn-reth` suite.


grantkee (dismissed):
Merge conflict - but the solution seems solid and should support multi-worker too. Good work

grantkee (approved):

PR #1181, opened 2026-08-13
fix(snapshot): enforce the BLOCKHASH lookback floor on restore windows and packs

### LINKED ISSUE #1174: snapshot restore: import_chain_scaffold accepts a header window shorter than the 256-block BLOCKHASH lookback and zero-fills the gap
## Classification
- Severity: Low. This is defense-in-depth on the restore trust boundary.
- Component: `crates/tn-reth/src/snapshot.rs`, function `import_chain_scaffold`.
- Class: consensus state-root divergence caused by a wrong `BLOCKHASH` result.
- Anchored to `origin/main` at commit `31cc4e90`.

## Summary
The snapshot restore path checks the header window for internal consistency. It does not check that the window is long enough to cover the EVM `BLOCKHASH` lookback of 256 blocks. A pack with a short window passes all of the current checks. The restore code then fills the missing ancestor slots with zero hashes. On the affected node, `BLOCKHASH` over the gap can return zero. This can make the node's state root different from the rest of the fleet.

## Why the severity is Low
The honest exporter always ships the anchor plus up to 256 real ancestor headers. See `gather_headers` in `crates/node/src/manager/export_exec.rs:270-286`, which walks down to `max(1, anchor - BLOCKHASH_ANCESTORS)`:

```
270	/// Collect the snapshot header (first) plus up to [`BLOCKHASH_ANCESTORS`] real ancestor headers
271	/// (excluding genesis), so the pack is self-sufficient for `BLOCKHASH` and the restore scaffold.
272	fn gather_headers(reth_env: &RethEnv, anchor: &SealedHeader) -> eyre::Result<Vec<ExecHeader>> {
...
278	    let lowest = anchor.number.saturating_sub(BLOCKHASH_ANCESTORS).max(1);
279	    for number in (lowest..anchor.number).rev() {
```

For a genuine pack, the first restored block `B+1` reads `BLOCKHASH` only over `[B-255, B]`. That range is inside the real hashes. The zero dummies sit below `B-256` and are never reached on the honest path. The in-code comment already states this (`snapshot.rs:623-625`):

```
623	        // headers 1..=B-1: real headers with real hashes inside the window, zero-hash dummies below
624	        // it. the dummies are never referenced by BLOCKHASH within the shipped window, so a zero
625	        // hash there is inert.
```

The gap becomes reachable only if a node imports a truncated or corrupted pack. The impact is then limited to that one node, which forks itself off the fleet. It is not a fleet-wide consensus break. The pack tip is pinned to a super-quorum certificate, but the pack content (the header count) is not certified.

## Current behavior (code)
`import_chain_scaffold` runs four checks and no length floor (`snapshot.rs:537-587`): non-empty window, tip equals `final_state`, block-number contiguity, and parent-hash linkage. `window_start` is checked only for `!= 0`:

```
537	        let header_b = window
538	            .last()
539	            .ok_or_else(|| eyre!("snapshot restore: cannot scaffold an empty header window"))?;
...
548	        if header_b.number != b || header_b.hash() != final_state.hash {
...
556	        let window_start = window[0].number;
557	        if window_start == 0 {
558	            return Err(eyre!("snapshot restore: window must not include genesis (block 0)"));
559	        }
560	        for (i, header) in window.iter().enumerate() {
...
576	        window.iter().zip(window.iter().skip(1)).try_for_each(|(parent, child)| {
```

Missing ancestors below `window_start` are written with `B256::ZERO` (`snapshot.rs:628-636`):

```
628	            for number in 1..b {
629	                match header_at(number) {
630	                    Some(header) => {
631	                        headers_writer.append_header(header.header(), &header.hash())?
632	                    }
633	                    None => {
634	                        let dummy = ExecHeader { number, ..Default::default() };
635	                        headers_writer.append_header(&dummy, &B256::ZERO)?;
636	                    }
```

The lookback constant lives only on the export side (`crates/node/src/manager/export_exec.rs:37`), value `256`. The restore side has no matching floor:

```
37	const BLOCKHASH_ANCESTORS: u64 = 256;
```

The pack-level guard bounds the header count from above only. In `crates/storage/src/exec_state_pack.rs`, `MAX_HEADER_COUNT = 4096` (line 77, enforced at 413), and the only lower guard is `!= 0` (line 407). A pack that declares `header_count == 1` passes both `open` and `import_chain_scaffold`. The CLI consumer passes the pack's headers through `scaffold_window`, which only sorts and drops genesis (`crates/telcoin-network-cli/src/db.rs:760-766`), then straight into the scaffold (`db.rs:551-555`).

## Impact
A node that restores from an untrusted or corrupted pack with `header_count < 256` (down to `1`) scaffolds zero-hash ancestors below `window_start`. If a transaction in a later block executes `BLOCKHASH(n)` for `n` in `[B-255, window_start-1]`, the node reads `0x00` while honest peers read the real ancestor hash. The node's execution result and state root diverge. The node forks itself off. This is an availability and liveness problem for the affected node.

## Proof of concept (sketch)
1. Take a valid exec state pack for block `B` (with `B > 256`).
2. Truncate the header set to the tip only, so `header_count = 1`, and keep the certified `final_state` tip.
3. Import the pack. `import_chain_scaffold` passes, because the tip matches and a one-header window is trivially contiguous and hash-linked.
4. Blocks `1..B-1` are scaffolded with `B256::ZERO`.
5. Produce or replay a block at `B+1` whose transactions call `BLOCKHASH(B-255)`. The node returns `0x00`, not the real ancestor hash. The resulting state root does not match the fleet.

## Suggested fix
Add a lower-bound check in `import_chain_scaffold`, so the window must cover the full lookback or reach genesis:

```
let floor = b.saturating_sub(BLOCKHASH_ANCESTORS - 1).max(1);
if window_start > floor {
    return Err(eyre!(
        "snapshot restore: window start {window_start} does not cover the BLOCKHASH lookback \
         (need <= {floor} for tip {b})"
    ));
}
```

Add a matching lower bound on `header_count` in `crates/storage/src/exec_state_pack.rs`, so a short pack is rejected at `open()` as well as at scaffold time. Share the `BLOCKHASH_ANCESTORS` constant between `export_exec.rs` and the restore side, so the export bound and the restore floor cannot drift apart.

MavenRain:
Closes #1174.

## Problem

The snapshot restore path checks the header window for internal consistency (non-empty, tip matches the certified `final_state`, block-number contiguity, parent-hash linkage) but never checks that the window is long enough to cover the EVM `BLOCKHASH` lookback of 256 blocks. `import_chain_scaffold` fills every block below the window with a zero-hash dummy. A truncated or corrupted pack with a short window (down to a single header) passes every current check; on the restored node, `BLOCKHASH(n)` for `n` inside the lookback of the first post-restore block then returns `0x00` while honest peers return the real ancestor hash. The node's execution result and state root diverge, and it forks itself off the fleet.

The pack tip is pinned to a super-quorum certificate, but the pack content (the header count) is not certified, so the length of the window is untrusted input on the restore boundary. The honest exporter always ships the full window (`gather_headers` walks down to `max(1, anchor - 256)`), so no attacker is required to hit the gap: a truncated download or a short copy is enough. The impact is limited to the restoring node (availability and liveness), not fleet-wide consensus.

## Fix

- [x] `crates/storage/src/exec_state_pack.rs`: `BLOCKHASH_ANCESTORS` (256) now lives here, shared by the export side and the restore side, so the export bound and the restore floor cannot drift apart. `ExecStatePackReader::open` refuses a pack whose declared `header_count` is below `min(block_number, BLOCKHASH_ANCESTORS)` with a new `InsufficientHeaders` error, before any header record is read and before any datadir mutation.
- [x] `crates/tn-reth/src/snapshot.rs`: `import_chain_scaffold` refuses a window whose first block is above `max(1, B - (BLOCKHASH_ANCESTORS - 1))`, so every zero-hash dummy it writes sits strictly below the `BLOCKHASH` reach of all post-restore blocks.
- [x] `crates/node/src/manager/export_exec.rs`: the local `BLOCKHASH_ANCESTORS` constant is removed in favor of the shared one.

Both checks are cheap integer comparisons. The pack floor refuses pre-flight, before any datadir mutation; the scaffold floor refuses before any chain data commits (the floor marker and the block-write transaction both come after the check). The floor is enforced symmetrically on the producing side (the exporter's `gather_headers` bound) and the consuming side (pack open plus scaffold), derived from the real EVM parameter rather than guessed. The scaffold floor is the consensus requirement (the first post-restore block `B + 1` reads `BLOCKHASH` over `[B - 255, B]`), deliberately one block looser than the exporter's own bound (the exporter ships one extra ancestor, down to `B - 256`, as headroom), so a pack at the exact floor is still accepted.

## Testing

- `cargo +nightly fmt -- --check`
- `cargo +1.94 test -p tn-storage exec_state_pack -j 2`: new `open_rejects_header_count_below_blockhash_floor`, a boundary pair at tip 300 (255 headers refused, 256 accepted).
- `cargo +1.94 test -p tn-reth snapshot -j 2`: new `import_rejects_a_window_short_of_the_blockhash_lookback`, a boundary pair at tip 300 (window start 45 accepted, start 46 refused), with the accepted run as the positive control.
- `cargo +1.94 test -p telcoin-network-cli pack -j 2` and `cargo +1.94 test -p tn-node export -j 2` for the downstream consumers.
- Both new tests were confirmed by mutation: with the corresponding floor check neutralized each test fails, and each passes again once the check is restored.
- Existing test fixtures that declared a tip block above their header coverage (tip 5 with 1 header, tip 100 with 3 headers) were re-anchored to compliant shapes (tip 1, tip 3), and the restored-floor scaffold test's window (blocks 3..=5 at tip 5) was extended to reach block 1; each test's original concern is unchanged.
- Out of scope: no test drives the real exporter (`gather_headers`) through the restore floors at a tip above 256, so the exporter bound (`B - 256`) and the restore floor (`B - 255`) are asserted independently rather than against each other; left as a follow-up.


crates/storage/src/exec_state_pack.rs
```diff
@@ -68,9 +68,17 @@ const DATA_NAME: &str = "state_data";
 /// working-set ceiling such an exporter must hold in memory is stated in one place.
 pub const STORAGE_CHUNK_SLOTS: usize = 64 * 1024;
 
+/// Depth of the EVM `BLOCKHASH` lookback window: the opcode resolves the 256 most recent block
+/// hashes. The export side embeds this many real ancestor headers below the snapshot header
+/// (`gather_headers` in `tn-node`), and the restore side refuses a pack (and a scaffold window,
+/// `SnapshotRestorer::import_chain_scaffold` in `tn-reth`) that does not cover this lookback
+/// below its tip. One shared constant keeps the export bound and the restore floor from drifting
+/// apart (issue #1174).
+pub const BLOCKHASH_ANCESTORS: u64 = 256;
```
grantkee, line 83: TN is on pectra fork, so the EVM still relies on this historic data. However osaka introduced [eip-2935](https://eips.ethereum.org/EIPS/eip-2935) in anticipation of endorsing [eip-7709](https://eips.ethereum.org/EIPS/eip-7709) in a future fork which will make this value `0`

grantkee (dismissed):
Merge conflict, but code is great. Good catch

grantkee (approved):

PR #1186, opened 2026-08-13
docs(evm): document and surface the gas over-reservation penalty

### LINKED ISSUE #1182: docs(evm): document and surface the gas over-reservation penalty
## Summary

`calculate_gas_penalty` charges a penalty for over-reserved gas. The charge is not visible to the transaction sender. The receipt does not change. No event or RPC field reports the penalty. The only documentation is contributor-facing.

## Problem

Evidence at `31cc4e90`:

- `TNEvmHandler::reimburse_caller` computes the penalty and splits the unused-gas refund (`crates/tn-reth/src/evm/handler.rs:126-148`). The caller receives `(unused_gas - penalty_gas) * effective_gas_price` in wei (`handler.rs:129-139`). The penalty portion goes to the base-fee address (`handler.rs:142-148`).
- The receipt `gas_used` comes from `result.gas_used()` and never includes the penalty (`crates/tn-reth/src/evm/block.rs:1414-1428`). A penalized transaction and an unpenalized one produce identical receipts.
- The only reporting is two node-side `tracing::debug!` calls (`crates/tn-reth/src/evm/utils.rs:81-88`, `handler.rs:132`). There is no EVM log, no event, and no RPC field.
- The only documentation is in `crates/tn-reth/README.md` (formula, threshold, and fund destination, in the ~359-397 range). There is no `docs/` directory. `chain-configs/` does not mention the penalty.

Effect: a sender sees a smaller refund than `(gas_limit - gas_used) * effective_gas_price` with no stated reason. Wallets and integrators cannot explain the balance delta from receipt data alone.

## Origin

Raised as comments in the 2026 collaborative review (comments do not become tracked findings on their own).

Haxatron: "There can be some UX issues here associated with this penalty. It should be documented."

shotes: "For UX, I think it's weird that the gas penalty happens silently and does not get reported back to gas used. It just gets lopped off and the user doesn't have reasoning, they just see the lopped off gas refund. This might fall to wallet UX aside from being outlined in protocol documentation though."

## Suggested fix

1. Add user-facing protocol documentation for the penalty: the 10 percent usage threshold, the `MIN_GAS_LIMIT_THRESHOLD` exemption (210,000), the quadratic formula, and the destination of the funds.
2. Consider a discoverable per-transaction signal. Options: an RPC extension field, or a documented convention for the base-fee address credit.
3. Coordinate the documentation with PR #1154, which changes the penalty basis for EIP-7702 transactions (#1139), so the documented formula does not go stale on merge.

MavenRain:
Closes #1182.  (Cantina #20)

## Summary

`calculate_gas_penalty` charges senders for over-reserving gas, but nothing user-facing explains the charge. The receipt does not change, no event or RPC field reports it, and the only documentation is contributor-facing (`crates/tn-reth/README.md`). Wallets and integrators cannot explain the balance delta from receipt data alone.

This PR adds user-facing documentation and a documented detection convention. Of the issue's two suggested per-transaction signals, it implements the "documented convention" branch; an RPC extension field stays open as a follow-up (see Out of scope).

## Changes

- [x] `docs/gas-penalty.md` (new): the user-facing specification. It covers the two activation conditions (gas limit greater than 210,000 and usage below 10 percent), the exact integer formula with its 10^9 fixed-point precision, the pre-refund vs post-refund gas distinction, the worked-examples table (mirrors the `calculate_gas_penalty` doc table, whose penalty values are pinned by `test_gas_penalty_calculation`), the refund split, the base-fee-address destination, and two detection recipes for wallets: exact from balances, and exact-or-upper-bound from recomputing the formula with receipt `gasUsed` (upper bound because the penalty is non-increasing in gas used, and receipt `gasUsed` never exceeds pre-refund gas).
- [x] `README.md`: a short "Gas over-reservation penalty" section, so the charge is discoverable from the repo front page.
- [x] `crates/tn-reth/README.md`: a cross-link from the contributor-facing "Fee economics" bullet to the user-facing page, with a keep-in-sync note.
- [x] `crates/tn-reth/src/evm/utils.rs`: one-character doc-comment correction found while fact-checking the table. The `10,000,000` row's "Penalty % of Unused" reads 95.6%, but `9,564,282 / 9,979,000 = 95.8%`. The penalty value itself is correct and unchanged; only the percentage label was off. No code change.

## Out of scope

- An RPC extension field that reports `penalty_gas` per transaction (the `tn` namespace in `crates/execution/tn-rpc` is a natural home). The documented balance-delta convention covers detection today; a live field is a separate API-design decision.
- PR #1154 changes the penalty basis for EIP-7702 transactions (#1139). This page documents current `main`. If #1154 merges, the page needs one added sentence on the EIP-7702 basis, in whichever order the two land.

## Verification

Documentation-only diff: 4 files, +164/-1. The single Rust delta is one character inside an existing `///` doc-comment table row (same width, rustfmt-stable, no code or doctest affected).

- Every numeric row in the examples table was recomputed by hand with the integer formula and matches the unit-test expectations in `crates/tn-reth/src/evm/utils.rs` (`test_gas_penalty_calculation`). That recomputation is what surfaced the 95.6% vs 95.8% label typo, fixed in both places.
- Behavioral claims (pre-refund penalty basis, exemption at `gas_limit <= 210_000`, exemption at usage of 10 percent or more, penalty capped at unused gas so the refund cannot go negative, base-fee-address credit, system-transaction exemption, penalty applies to reverted transactions) were adversarially fact-checked line-by-line against `calculate_gas_penalty` and `TNEvmHandler::reimburse_caller`.
- A repo-wide sweep confirmed no production receipt, log, event, or RPC field references the penalty outside `utils.rs`, `handler.rs`, and tests, which is the gap this page documents.
- Relative links verified to resolve from each file's location.


grantkee (approved):
I think the changes are solid. The documentation is a larger decision, we have outlined this on https://docs.telcoin.network hosted with gitbook. Maybe we move that as a submodule into this repo to ensure docs stay consistent? This information is also subject to change with #1154

sstanfield (approved):

MavenRain:
> I think the changes are solid. The documentation is a larger decision, we have outlined this on https://docs.telcoin.network hosted with gitbook. Maybe we move that as a submodule into this repo to ensure docs stay consistent? This information is also subject to change with #1154

I'd be in favor of moving that as a submodule/crate into this repo to ensure consistency.

PR #1188, opened 2026-08-14
fix(node): deliver epoch votes across the epoch-open certificate wait

### LINKED ISSUE #1187: Epoch-open deadlock when the seed-signature fork is active: the certified-anchor wait blocks before the epoch-vote consumer spawns (adiri halt at epoch 383)
## Summary

All five adiri validators halted at the 383 to 384 epoch boundary on 2026-08-13, at about 20:21 UTC.  The cause is a circular wait in the epoch-open path.  The wait arms when `seed_signature_active(current_epoch)` is true.

At epoch open, `open_epoch_pack` blocks until the prior epoch record is certified.  `create_consensus` runs only after `open_epoch_pack` returns, and it spawns the per-epoch "primary network events" task.  That task is the only consumer that moves peer `EpochVote` gossip into the vote collector.  Each validator waits for a certificate that needs votes, and the votes wait for a consumer that has not spawned yet.  The wait is symmetric across the committee, so the whole network halts.  Restarts do not clear it.

## Environment

- Network: adiri
- Image: `v0.13.0-adiri`, commit `09ee8f616d811b398228d844241bab898b5eea46`
- Seed-signature fork epoch: 383
- Committee: 5 validators, quorum threshold 4

## Observed behavior

We confirmed the same sequence in the logs of all five validators.  The timestamps below are from validator-1.  Validators 2 to 5 match within gossip latency, and all five fail on the same record digest.

1. Consensus is healthy to the end of epoch 383.  The final rounds contain certificates from all five authorities.
2. 20:21:01.148: `epoch boundary success`, `Epoch Task Manager: All tasks shutdown`, and `publishing epoch record GNEs7wdXUaCqnpX6` appear in the same millisecond.
3. 20:21:01.344: `Starting epoch` 384.  Then 20:21:01.367: WARN `previous epoch record not yet certified, waiting for its certificate`.
4. For 65 seconds the log shows no epoch-vote receipt and no rejection.  65 s = (24 + 2) x `EPOCH_VOTE_RECV_TIMEOUT` of 2.5 s (`MAX_EPOCH_VOTE_TIMEOUTS` = 24;  epoch_votes.rs:21-28).
5. 20:22:06.183: ERROR `failed to reach quorum on epoch close` for `GNEs7wdXUaCqnpX6`.  All five validators report this error with the same digest.  So the record itself is deterministic and consistent;  only vote delivery failed.
6. Five by-digest recovery requests to peers fail.  The process exits about 2 minutes after the boundary.  (This is earlier than `CERTIFIED_ANCHOR_WAIT`;  node supervision likely explains the early exit.  The deadlock analysis does not depend on this detail.)
7. After a restart, a node does not publish its epoch-383 vote again.  It enters the same wait and dies again.  The fleet crash-loops.

## Root cause

All code references are to commit `09ee8f61`.

1. `run_epoch.rs` opens each epoch through `open_epoch_pack` (run_epoch.rs:288-289).  When `seed_signature_active(current_epoch)` is true (run_epoch.rs:659), it calls `certified_prior_epoch_anchor` (run_epoch.rs:671).  That function logs the WARN (run_epoch.rs:771) and then blocks on `certified_record_by_epoch_with_timeout(prev_epoch, CERTIFIED_ANCHOR_WAIT)` (run_epoch.rs:779).
2. `create_consensus` (run_epoch.rs:384) runs only after `open_epoch_pack` returns.  It leads to `spawn_primary_network_for_epoch` (start_epoch.rs:401), which spawns the "primary network events" task (primary/src/network/mod.rs:1234) on the per-epoch task manager.
3. That task is the only drain of the `primary_network_events` channel, and it is the only path that forwards a peer `EpochVote` to the vote collector (handler.rs:535).  The channel is `new_always_subscribed` (consensus_bus.rs:408), so incoming votes queue across the boundary, but nothing drains them.
4. The wait in step 1 can never complete.  The certificate needs a quorum of votes.  The votes need the consumer.  The consumer spawns after the wait.  Every validator blocks in the same way, so the committee halts.

## Why the first fork boundary (382 to 383) survived

The gate applies to the epoch that opens, so the open of epoch 383 also took the blocking branch (it waited on the epoch-382 certificate).  It survived because of a timing race, not because the path is safe.  `write_epoch_record` spawns `manage_epoch_votes` (run_epoch.rs:491) before `abort_all_tasks` (run_epoch.rs:529), and the node-lifetime vote collector accepts votes for `latest + 1`.  A validator that closes slightly later than its peers can collect quorum before its own teardown, and it then serves the certificate to the others.  At the 383 close, all five validators published and tore down within gossip latency of each other.  Validator-1 shows publish and teardown in the same millisecond.  The race was lost fleet-wide.  Which fork-era boundary halts is luck.

## Why restarts cannot recover the network

- A restarted node does not re-publish its epoch vote.
- A peer serves an epoch-record request only when the record is certified;  when the certificate is missing, it retries and then returns `UnavailableEpoch` (handler.rs:1161-1181).  No certified record for epoch 383 exists anywhere.
- The per-epoch consumer is dead on every peer, so new votes are not processed anywhere.

No restart order can build the missing certificate.  Recovery needs an out-of-band step: a patched binary, or an injected certificate.  A rollback of the fork epoch is not safe: headers committed at or past the fork epoch do not decode under the legacy seven-field layout (forks.rs:150-159).

## Causes we ruled out

- A missing or mixed-version validator.  All five authorities produced certificates in the final rounds.  The `quorum reached` lines with `4/4` are at-threshold logging.
- Loss of the gossip topic.  The epoch-vote topic is static (config/src/network.rs:180-181) and is never unsubscribed.
- Vote rejection.  Gossip rejections log at WARN through `process_gossip`;  the logs contain none.
- Fork-gated encoding or digest divergence.  All five validators fail on the same digest.

## Secondary hardening suggestions

These made the halt hard to diagnose from logs:

- `handle_new_vote` (epoch_votes.rs:294-316) drops mismatched votes with no log line (the silent return is at line 314).
- No INFO-level receipt log exists for epoch votes.  One line per received vote would make a halt like this diagnosable from a single validator log.
- handler.rs:542-546 drops votes for an unknown committee epoch at debug level only.

## Suggested fix directions

- Spawn the "primary network events" consumer, or a minimal gossip-to-vote-collector pump, before the certified-anchor wait in `open_epoch_pack`.  Or:
- Make the epoch-vote delivery path node-lifetime, so it does not depend on epoch-open ordering.
- Have a node re-publish its own epoch vote on restart, so a partial outage can heal.

MavenRain:
## Problem

All five adiri validators crash-loop at the epoch 383 -> 384 boundary.  Each node closed epoch 383, durably persisted the epoch record, published it, and then blocked in `certified_prior_epoch_anchor` waiting for the epoch-383 certificate.  The certificate never forms, and restarts do not help.

## Root cause

The certificate is aggregated from `EpochVote` gossip.  The only production path that moves that gossip into the vote collector is the per-epoch primary network task, which `create_consensus` spawns after `open_epoch_pack` returns.  The epoch open itself blocks inside `open_epoch_pack` waiting for the certificate.  Votes therefore queue on `primary_network_events` with no consumer, on every node at once.  Restarting makes it worse: nothing on the restart path re-publishes a node's own vote, re-fires `epoch_record_watch`, or re-subscribes the epoch-vote gossip topic, so the collector never starts aggregating for epoch 383 at all.

## Fix

Two pieces, both reusing the existing vote machinery.

**1. A scoped epoch-vote pump around the certificate wait (`EpochVotePump`).**  Before awaiting the certificate, `certified_prior_epoch_anchor` takes the `primary_network_events` subscription and spawns a pump task.  The pump decodes each gossip event.  A signature-checked `EpochVote` goes into `new_epoch_votes` (the same send the gossip handler performs), and every other event is buffered.  When the wait returns, on both the success and the error path, the pump stops and releases the subscription; the buffered events are then re-queued from a detached task (see the threat model for why the re-queue must not run on the caller's critical path).  This closes the delivery gap at every future epoch boundary, not just for this outage.

**2. Resume the epoch-close vote protocol on startup.**  Right after the node-lifetime vote collector spawns, `EpochManager::run` reads the latest persisted epoch record.  If it exists, belongs to an already-closed epoch (`record.epoch < current_epoch`), and has no certificate, the node first subscribes the epoch-vote gossip topic itself, then re-fires `epoch_record_watch` with the persisted record.  The subscription is load-bearing: the only per-epoch subscribe runs in `spawn_primary_network_for_epoch`, after the epoch open this recovery unblocks, and until it runs gossipsub neither fans our publishes out nor accepts peer votes (`verify_gossip` rejects messages on a topic with no `authorized_publishers` entry).  The publisher window is derived from the persisted record: `committee` (the outgoing voters) plus `next_committee`, the previous/current half of the previous/current/next window the epoch-start path builds as `boundary_publishers`.  The later per-epoch subscribe replaces that publisher set outright (the swarm's `Subscribe` handler is a map insert) and a repeated gossipsub subscribe is idempotent, so the recovery-time subscription cannot pollute normal operation.  Firing the watch replays the exact close-path handoff: the collector re-signs the same persisted bytes (the vote is a pure function of the record and the loaded key), publishes the vote, re-publishes it on every 2.5s recv-timeout tick, and drives the normal aggregation, certificate write, and peer-recovery machinery.  An INFO line marks the resume with the epoch and digest, and a DEBUG line marks the certified-record skip.

Deployed to the five validators, each restarted node re-publishes its epoch-383 vote (piece 2) and delivers its peers' votes while blocked at the epoch-384 open (piece 1).  Any 4-of-5 quorum then certifies the record, the anchor wait polls the certificate out of storage, and state-sync serves it to any node still behind.

### Threat model

- **Forwarding unverified gossip cannot mint a bad certificate.**  The pump skips the handler's topic, committee, and dedup gates, but keeps the per-vote `check_signature` the collector's aggregation loop documents as its precondition.  Aggregation itself only counts votes whose `epoch_hash` matches the collector's record and whose signer is in that record's committee, dedups per authority (`committee_keys.remove` plus the `signed_authorities.insert` guard), and verifies the assembled certificate against a 2/3+1 super quorum (`EpochRecord::verify_with_cert`) before anything is stored.
- **The receiver take/drop ordering is sound.**  `primary_network_events` allows one live subscription, and a second live subscribe panics.  `QueChanReceiver::drop` returns the inner receiver to the channel's slot.  The pump task's stream owns the receiver and drops it before the task yields its buffer, and `stop_and_replay` joins the task before returning, on both exit paths of the wait.  The real consumer takes the subscription only later, in `spawn_primary_network_for_epoch`.
- **The replay is detached, deliberately.**  The channel is bounded and continuously fed by the node-lifetime swarm task, while the only future drainer (the per-epoch network task) cannot spawn until `certified_prior_epoch_anchor` returns.  Awaiting a backpressuring re-queue send on that critical path could therefore suspend forever.  The detached task's sends always complete because the next consumer drains the channel independently; if the process exits first the task dies with it, losing only events a dead process could not have handled.  Replayed events may interleave with newer traffic, a reordering the network layer tolerates (gossip is unordered and requests carry their own deadlines).  Undecodable or bad-signature gossip is replayed rather than consumed, so penalty attribution stays with the real handler.
- **The resume cannot certify an in-progress epoch.**  The `record.epoch < current_epoch` gate keeps the epoch-0 dummy record, and any record for the epoch still running, away from the collector.  Certifying the dummy would permanently block the real epoch-0 record behind `write_epoch_record`'s certificate short-circuit.  The persisted record is used verbatim, never rebuilt, so the re-signed vote digest matches the one peers voted on.

## Changes

- [x] `crates/consensus/primary/src/network/mod.rs`: add `EpochVotePump` (scoped subscribe, signature-checked vote forwarding, buffering, stop-then-detached-replay).
- [x] `crates/node/src/manager/node/run_epoch.rs`: run the pump around the `certified_record_by_epoch_with_timeout` wait in `certified_prior_epoch_anchor`, stopping it on both the Ok and the Err path.
- [x] `crates/node/src/manager/epoch_votes.rs`: add `resume_epoch_certification` (latest-record read, certificate gate, recovery-time `subscribe_with_publishers` on the epoch-vote topic with the record-derived publisher window, watch re-fire, INFO/DEBUG ops logging).
- [x] `crates/node/src/manager/node.rs`: invoke the resume in `EpochManager::run`, right after `spawn_epoch_vote_collector`.
- [x] `crates/node/src/manager/mod.rs`: re-export the resume helper.
- [x] Tests: pump forward/buffer/detached-replay plus subscription release (`network_tests.rs`), and the resume trigger matrix in `epoch_votes.rs` (uncertified record fires the watch and subscribes with the right topic and publisher window, certified record skips, empty store skips, epoch-0 dummy skips, future-epoch record skips).

## Deployment note

Restart at least 4 of the 5 validators within about a minute of each other.  A restarted node re-publishes its epoch-383 vote every 2.5 seconds, but only for roughly 65 seconds (the collector's 24-plus-2 recv-timeout windows) before it falls back to certificate peer recovery, and a publish that finds no subscribed peer yet fails silently until the next tick.  A round that misses the window is safely retried by restarting the fleet together again;  the resume re-fires on every restart for as long as the record remains uncertified.

## Testing

- `cargo +nightly fmt --all --check` (the toolchain CI's fmt lane uses): clean, workspace-wide.
- `cargo +1.94 clippy --no-deps` for the two touched crates: DEFERRED to CI.  The build machine was below the local disk floor (11 GiB free), so no local build, clippy, or test run was attempted.  CI and the second machine own the dynamic tier for this change.
- New unit tests, to run in CI: `test_epoch_vote_pump_forwards_votes_and_replays_the_rest`, `test_resume_fires_watch_for_uncertified_latest_record`, `test_resume_skips_certified_latest_record`, `test_resume_skips_in_progress_epoch_dummy`, `test_resume_skips_record_from_a_future_epoch`.
- The resume-to-certificate composition is covered by the existing collector tests, which drive `manage_epoch_votes` through the same `epoch_record_watch` fire the resume performs.

Closes #1187

### DIFF
--- a/crates/consensus/primary/src/network/mod.rs
+++ b/crates/consensus/primary/src/network/mod.rs
@@ -18,8 +18,8 @@ use tn_network_libp2p::{
     error::NetworkError,
     read_frame,
     types::{
-        IntoResponse as _, NetworkCommand, NetworkEvent, NetworkHandle, NetworkResponseMessage,
-        NetworkResult,
+        GossipPayload, IntoResponse as _, NetworkCommand, NetworkEvent, NetworkHandle,
+        NetworkResponseMessage, NetworkResult,
     },
     write_frame, DenyReason, GossipMessage, Penalty, PrimarySyncRequest, ResponseChannel, Stream,
     StreamError, SyncFrame, SyncFrameError,
@@ -31,12 +31,12 @@ use tn_storage::{
     PayloadStore,
 };
 use tn_types::{
-    encode, BlsPublicKey, BlsSignature, Certificate, ConsensusHeaderDigest, ConsensusOutput,
-    ConsensusResult, Database, Epoch, EpochCertificate, EpochDigest, EpochRecord, EpochVote,
-    Header, HeaderDigest, Round, TaskError, TaskSpawner, TnReceiver, TnSender, Vote,
+    encode, try_decode, BlsPublicKey, BlsSignature, Certificate, ConsensusHeaderDigest,
+    ConsensusOutput, ConsensusResult, Database, Epoch, EpochCertificate, EpochDigest, EpochRecord,
+    EpochVote, Header, HeaderDigest, Round, TaskError, TaskSpawner, TnReceiver, TnSender, Vote,
 };
 use tokio::sync::{mpsc, oneshot, OwnedSemaphorePermit, Semaphore};
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 pub mod handler;
 mod message;
 mod sync_codec;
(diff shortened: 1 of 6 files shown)


MavenRain:
Closed in favor of https://github.com/Telcoin-Association/telcoin-network/pull/1196

PR #1192, opened 2026-08-14
fix(storage): enforce STORAGE_CHUNK_SLOTS on read and write paths

### LINKED ISSUE #1178: snapshot restore does not re-validate storage-chunk length against STORAGE_CHUNK_SLOTS
## Summary

The snapshot restore path reads storage chunks from a state pack. It does not
check that each chunk holds at most `STORAGE_CHUNK_SLOTS` slots. The
`STORAGE_CHUNK_SLOTS` bound is a writer-side convention only. A hand-crafted pack
can carry storage chunks that are larger than the intended working-set size. Each
such chunk drives a heap allocation before restore recomputes and checks the state
root.

## Location

- `crates/tn-reth/src/snapshot.rs` — `import_state` loop (near line 791),
  `write_storage_chunk` (near line 998).
- `crates/storage/src/exec_state_pack.rs` — `next_entry` (near line 519),
  `STORAGE_CHUNK_SLOTS` (line 69).

## Details

The writer splits each account's storage into chunks of at most
`STORAGE_CHUNK_SLOTS` slots. The constant and its stated intent:

```rust
// crates/storage/src/exec_state_pack.rs:64
/// (`MAX_RECORD_SIZE` = 16 MiB, `crate::archive::pack_iter`), leaving margin for BCS overhead.
pub const STORAGE_CHUNK_SLOTS: usize = 64 * 1024;
```

The reader returns a storage chunk with no length check:

```rust
// crates/storage/src/exec_state_pack.rs:527
Some(Ok(ExecStateRecord::Storage(chunk))) => Some(Ok(StateEntry::Storage(chunk))),
```

The restore loop validates address order and record order. It does not validate
chunk length:

```rust
// crates/tn-reth/src/snapshot.rs:812
StateEntry::Storage(chunk) => {
    // a Storage record can only follow an Account record; ...
    let address = current.ok_or_else(|| { /* ... */ })?;
    Self::write_storage_chunk(&provider, address, chunk, b)?;
}
```

`write_storage_chunk` then allocates vectors sized to `chunk.len()`:

```rust
// crates/tn-reth/src/snapshot.rs:1015
let plain_storage: Vec<(U256, U256)> = chunk
    .iter()
    .map(|(slot, value)| (U256::from_be_bytes(slot.0), U256::from_be_bytes(value.0)))
    .collect();
```

The only remaining ceiling is the container record cap
`MAX_RECORD_SIZE = 16 MiB` (`crates/storage/src/archive/pack_iter.rs:22`). That
cap is about 4 times the documented `STORAGE_CHUNK_SLOTS` working set (65,536
slots is near 4 MiB). So a malformed pack can push chunks up to about 4 times the
intended size through the restore allocation.

## Reachability and severity

Severity: low.

The path is not network-reachable. An operator invokes restore with
`tn db load-state` (`crates/telcoin-network-cli/src/db.rs`) against a pack file
that the operator supplies. The module documents that restore proves internal
consistency, not provenance:

```
// crates/tn-reth/src/snapshot.rs:43
Restore proves internal consistency, not provenance.
```

The restore streams one chunk at a time, so peak memory stays bounded to about one
`MAX_RECORD_SIZE` record. This is not a large-scale memory DoS. It is a
defense-in-depth gap: a writer-side invariant is trusted on the read path without
a re-check, so a corrupt or hostile pack can exceed the documented bound.

## Suggested fix

Reject any storage chunk whose length is greater than `STORAGE_CHUNK_SLOTS`.
Return the corrupt-pack error, the same as the existing order checks. Add the check
in `next_entry` (or in the `import_state` loop). This makes the documented bound an
enforced bound.

## Known-issue check

This is distinct from #1174 (BLOCKHASH lookback zero-fill in the same module). I
found no open issue or PR that covers a storage-chunk length re-check on the
restore path.

MavenRain:
Closes #1178.

## Problem

The snapshot restore path reads storage chunks from a state pack without checking that each chunk holds at most `STORAGE_CHUNK_SLOTS` (64k) slots. The bound is a writer-side convention only: `ExecStatePackWriter::append_account` splits at exactly that size, but `ExecStatePackReader` returns whatever slot count a record carries. A hand-crafted or corrupt pack can therefore carry chunks up to the container's 16 MiB record cap (about 4x the documented working set), and each one drives allocations sized to the oversized chunk (`import_state` / `write_storage_chunk` in `crates/tn-reth/src/snapshot.rs`) before the state-root re-check runs.

Threat model: the pack is operator-supplied (`tn db load-state`), not network-reachable, and streaming keeps peak memory near one record. This is a defense-in-depth gap (a trusted-writer invariant on an untrusted read path), not a large-scale DoS. The fix makes the documented bound an enforced bound.

## Fix

One shared constant, enforced symmetrically on the producing and reconstructing paths (all in `crates/storage/src/exec_state_pack.rs`):

- [x] New `check_chunk_bound` helper and a dedicated `ExecStatePackError::OversizedStorageChunk { len, max }` variant, mirroring the existing `TooManyHeaders` bound error.
- [x] **Reader**: `ExecStatePackReader::pull` re-checks every `Storage` record it pulls, so every read path (`next_entry`, `next_account`, `accounts`) rejects an oversized chunk before any consumer sizes an allocation to it. `verify` now walks records through `pull` instead of the raw container iterator, so a pack that `verify` accepts is one the import side will not reject on structural grounds.
- [x] **Writer**: `append_storage_chunk_owned` (and via it `append_storage_chunk`) rejects an oversized chunk at append time, so a buggy streaming exporter fails at write time rather than producing a pack that fails restore. The internal chunker (`append_account`) splits at exactly the bound, so no legitimate writer is affected.
- [x] Doc comments updated on `STORAGE_CHUNK_SLOTS`, both append fns, `pull`, and `verify` to state the bound is enforced, not advisory.

The restore loop itself needs no change: `import_state` consumes `next_entry` and inherits the rejection, failing loudly the same way the existing order checks do.

## Testing

- `reader_rejects_oversized_storage_chunk`: hand-crafts a raw pack (bypassing the writer's new check) whose single chunk holds `STORAGE_CHUNK_SLOTS + 1` slots; all three read paths (`next_account`, `next_entry`, `verify`) reject it with `OversizedStorageChunk`. Positive control: the same fixture at exactly `STORAGE_CHUNK_SLOTS` streams through all three paths, so the rejection is attributable to the bound check, not to the raw-crafted fixture.
- `writer_rejects_oversized_storage_chunk`: an append at exactly the bound succeeds, one slot more is rejected without a partial tally, and the surviving pack still verifies.
- Both regression tests confirmed by mutation: with the guard short-circuited the oversized-rejection assertions fail, and with `<` in place of `<=` the exact-bound positive control fails; the fix restored, everything is green.
- `cargo +1.94 test -p tn-storage exec_state_pack` green under the pinned toolchain; `make attest` (nightly fmt + 1.94 clippy + workspace test-build in an isolated target dir).

### DIFF
--- a/crates/storage/src/exec_state_pack.rs
+++ b/crates/storage/src/exec_state_pack.rs
@@ -66,6 +66,12 @@ const DATA_NAME: &str = "state_data";
 /// Public so streaming exporters can flush at exactly this bound — reproducing the chunk layout
 /// [`ExecStatePackWriter::append_account`] emits, byte for byte — and so the per-account
 /// working-set ceiling such an exporter must hold in memory is stated in one place.
+///
+/// The bound is enforced on both ends: the writer rejects a longer chunk at
+/// [`ExecStatePackWriter::append_storage_chunk_owned`], and the reader rejects one on every read
+/// path (as [`ExecStatePackError::OversizedStorageChunk`]) before any consumer sizes an
+/// allocation to it, so a corrupt or hand-crafted pack cannot push chunks past the documented
+/// working-set ceiling through a restore.
 pub const STORAGE_CHUNK_SLOTS: usize = 64 * 1024;
 
 /// Upper bound on the header count a pack may declare, enforced before any allocation sized by the
@@ -208,6 +214,19 @@ fn decode_header(bytes: &[u8]) -> Result<ExecHeader, ExecStatePackError> {
     ExecHeader::decode(&mut &bytes[..]).map_err(|e| ExecStatePackError::HeaderRlp(e.to_string()))
 }
 
+/// Enforce the documented [`STORAGE_CHUNK_SLOTS`] bound on one storage chunk's slot count.
+///
+/// A conforming writer keeps the bound by construction ([`ExecStatePackWriter::append_account`]
+/// splits at exactly this size), so a longer chunk can only come from a non-conforming writer or
+/// a corrupt / hand-crafted pack. Checked at both ends: on append (so a buggy exporter fails at
+/// write time, not at restore time) and on every record pulled by the reader (so the read side
+/// never trusts the writer-side convention).
+fn check_chunk_bound(len: usize) -> Result<(), ExecStatePackError> {
+    (len <= STORAGE_CHUNK_SLOTS)
+        .then_some(())
+        .ok_or(ExecStatePackError::OversizedStorageChunk { len, max: STORAGE_CHUNK_SLOTS })
+}
+
 /// Summary returned by a successful [`ExecStatePackReader::verify`] pass.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct VerifyReport {
@@ -325,8 +344,8 @@ impl ExecStatePackWriter {
     }
 
     /// Append one chunk of storage slots for the current account. Empty chunks are ignored. A chunk
-    /// should hold at most [`STORAGE_CHUNK_SLOTS`] slots to stay under the container's record
-    /// limit.
+    /// must hold at most [`STORAGE_CHUNK_SLOTS`] slots to stay under the container's record
+    /// limit; a longer chunk is rejected as [`ExecStatePackError::OversizedStorageChunk`].
     pub fn append_storage_chunk(
         &mut self,
         slots: &[(B256, B256)],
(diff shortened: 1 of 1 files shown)


grantkee (approved):
LGTM

PR #1194, opened 2026-08-14
fix(network-libp2p): cap per-entry multiaddrs on discovery ingest

### LINKED ISSUE #1183: network-libp2p: PeerExchange ingest does not cap per-peer multiaddr count, enabling one-message dial amplification
## Summary

The PeerExchange ingest path stores a peer's advertised multiaddr list with no cap on the
number of addresses per peer entry. The equivalent kad `NodeRecord` path caps this list at
`MAX_ADVERTISED_MULTIADDRS` (16). Because the cap was not mirrored, one unauthenticated,
non-committee peer can send a single small PeerExchange response that maps a fake `peer_id` to
tens of thousands of multiaddrs. The victim node stores the list and later tries to dial every
address. The attacker chooses the addresses, so the victim connects to third-party IP:port
targets that the attacker selects.

This is a griefing and connection-amplification vector. It is not a chain-halt.

- Cap that IS applied (kad record path):
  `crates/network-libp2p/src/consensus.rs:616`
  `if node_record.info.multiaddrs.len() > MAX_ADVERTISED_MULTIADDRS` (constant at
  `consensus.rs:81`, value 16), inside `peer_record_valid`.
- Sibling path with NO per-peer address cap (PeerExchange):
  `crates/network-libp2p/src/peers/manager.rs:652` `process_peer_exchange`.
  The address list is copied verbatim at `manager.rs:663-664`
  (`PeerInfo { addrs: addrs.into_iter().collect() }`) and stored at `manager.rs:687`
  (`self.discovery_peers.insert(peer.peer_id, peer.addrs)`).
  `eligible_for_discovery` then calls `has_valid_unbanned_ips` (`manager.rs:1058-1071`), which
  requires only that at least ONE address in the list is a valid, unbanned IPv4 or IPv6
  address. It never checks the count, and it does not filter the other addresses. Unroutable
  or private (RFC1918) addresses in the same list still pass and are still dialed.
- Dial path (confirmed at `31cc4e90`):
  `discovery_heartbeat` (`manager.rs:1111-1168`) selects stored entries and calls
  `dial_peer(peer, addrs, None)` (`manager.rs:1138`) with the full stored address vector.
  `dial_peer` (`manager.rs:181-251`) gates only on peer identity status (local, banned,
  dialing, connected), then pushes `DialRequest { peer_id, multiaddrs, reply }`
  (`manager.rs:249-250`). `peers/behavior.rs:195-209` turns this into a real Swarm dial:
  `ToSwarm::Dial { opts: DialOpts::peer_id(peer_id).condition(PeerCondition::Disconnected).addresses(multiaddrs).build() }`.
  The full, uncapped address list reaches libp2p's dial machinery, which attempts the
  addresses (batched by its dial-concurrency factor) until one succeeds or the list ends.

## Root cause

A resource cap was applied to one collection and was not mirrored onto its sibling. The kad
`NodeRecord` path bounds the per-record multiaddr count. The PeerExchange path bounds only the
number of peer entries (`peers_to_take = max_discovery_peers - current_count; .take(peers_to_take)`
at `manager.rs:684-685`). It does not bound the number of multiaddrs inside each entry.

## Threat model

A byzantine, non-committee peer that can complete a connection and send a PeerExchange response.
The attacker does not need any validator key.

## Impact

The victim performs a large number of outbound dial attempts against attacker-chosen third-party
addresses. Each stored entry is dialed across its full address list when the heartbeat selects it
(libp2p batches the attempts by its dial-concurrency factor). This wastes the victim's connection
resources and turns the honest node into a connection-reflection and involuntary port-scan proxy
against third parties. The per-entry address count is bounded only by the message-size limit. A
fake peer entry never completes a handshake, so it stays in `discovery_peers` (up to
`max_discovery_peers`) and is dialed again across later heartbeats, so the effect is sustained, not
one-shot.

Suggested severity: High for a networking layer, because the honest node is used as an attack proxy
against parties it did not choose. It does not corrupt consensus and does not halt the chain.

## Reproduction (sketch)

1. Connect to a target node as a non-committee peer.
2. Send a PeerExchange response with one entry: a random `peer_id` and a multiaddr list filled to
   the message-size limit with addresses that point at a chosen third-party IP:port.
3. Observe the target dial every address in the list.

## Suggested fix

Apply a per-entry address cap in `process_peer_exchange`, the same way the kad path applies
`MAX_ADVERTISED_MULTIADDRS`. Reject or truncate any entry whose address count is over the cap. A
dedicated constant (for example `MAX_PEER_EXCHANGE_MULTIADDRS`) is fine if the value should differ
from the kad limit.

## Scope

The rest of the PeerExchange path is already hardened. The number of peer entries is capped, and
each address is checked for validity and ban status. The only missing control is the per-entry
address count.

MavenRain:
Closes #1183.

## Problem

`process_peer_exchange` stores a peer's advertised multiaddr list with no cap on the number of
addresses per entry. The kad `NodeRecord` path caps its list at `MAX_ADVERTISED_MULTIADDRS`
inside `peer_record_valid`, but the PeerExchange path caps only the number of peer entries, not the
addresses inside each entry.

One unauthenticated, non-committee peer can send a single PeerExchange response that maps a fake
`peer_id` to an address list bounded only by the message-size limit. The victim stores the list in
`discovery_peers`, and the discovery heartbeat later dials the full list. The attacker chooses the
addresses, so the victim opens connection attempts against third-party IP:port targets the attacker
selects. A fake entry never completes a handshake, so it stays in `discovery_peers` and is dialed
again on later heartbeats. This turns an honest node into a sustained dial-amplification and
involuntary port-scan proxy. No validator key is required.

## Root cause

A resource cap was applied to one ingest path and was not mirrored onto its sibling.
`eligible_for_discovery` checks identity, IP validity, and ban status, but never the address count.

## Fix

- [x] `crates/network-libp2p/src/peers/manager.rs`: `eligible_for_discovery` now rejects any entry
  whose address list exceeds `MAX_MULTIADDRS_PER_PEER`.
- [x] `crates/network-libp2p/src/peers/peer.rs`: `MAX_MULTIADDRS_PER_PEER` lowered from 32 to 1
  (review follow-up, see below); doc note on its second role as the discovery per-entry ceiling.
- [x] `crates/network-libp2p/src/consensus.rs`: `MAX_ADVERTISED_MULTIADDRS` (was 16) is now tied
  to `MAX_MULTIADDRS_PER_PEER`, so validation and storage agree on how many addresses one peer may
  present.
- [x] `crates/network-libp2p/src/peers/mod.rs`: the cap is re-exported crate-wide (it was
  test-only) so the record validation can reference it.

The value follows the protocol assumption, agreed in review and at stand-up: there is exactly one
address per peer. A `NodeRecord` advertises one address (`NodeRecord::build`), a committee entry
carries one (`network_address`), and every consumer acts on a single address for a peer. So the
stored set holds only the address the peer most recently presented. The set is keyed on exact
`Multiaddr` equality and one endpoint reaches it in two syntactic forms, with and without the
`/p2p/<peer_id>` suffix (advertised as the operator configured it, dialed always with `/p2p` because
libp2p-swarm appends it to every dial, seen bare again on an inbound connection because libp2p-quic
dials from the listening socket); under a cap of 1 those forms replace each other instead of
accumulating, and either form dials the same endpoint. An eviction only trims the peer-exchange
payload (`exchange_info`): dialing reads `known_peers` / kad / `discovery_peers` and banning reads
`observed_ip_addresses`, never this set. The cap also bounds the dial fan-out of one discovery
entry to a single address.

Mixed-version note: a sender still on the pre-#1194 cap (32) may hold both forms of one honest
endpoint for a peer it dialed, so its exchange entry for that peer carries 2 addresses and a node on
this branch rejects it as ineligible for discovery. That only removes a discovery candidate during
the rollout window (kad closest-peers and committee dialing are unaffected); once senders run this
branch every honest entry carries exactly one address.

A legitimate exchange entry is a copy of the sender's stored set for that peer, so an honest sender
cannot exceed the cap. The check is a cheap length compare, runs before any dial resources are
spent, and is attacker-independent.

Enforcing the cap in `eligible_for_discovery` covers all three call sites with one guard:

- `process_peer_exchange`, the uncapped ingest this issue reports,
- `process_peers_for_discovery`, the kad closest-peers ingest. Those entries carry whatever the
  remote responder's routing table holds for a peer (`KadPeer.multiaddrs` in FIND_NODE replies),
  which nothing upstream caps and which never passes through `peer_record_valid`, so this guard is
  the first bound on that path too. Honest entries there hold one `/p2p` form per address, so no
  honest entry is affected.
- the `discovery_heartbeat` retain, which re-checks eligibility, so an oversized entry that reached
  `discovery_peers` through any other path is purged before it is dialed.

Rejection (not truncation) mirrors the kad `peer_record_valid` semantics and does not let the
attacker choose which addresses survive. The cap is also symmetric across producer and reader: the
producing path (`peer_exchange` / `exchange_info`) and the ingesting path bound the per-peer
address list with the same shared constant, and so does record validation.

## Testing

- `test_process_peer_exchange_rejects_oversized_multiaddr_list`: an exchange entry with
  `MAX_MULTIADDRS_PER_PEER + 1` addresses never enters `discovery_peers`; an entry with exactly
  `MAX_MULTIADDRS_PER_PEER` addresses does (the boundary is exact).
- `test_discovery_heartbeat_purges_oversized_multiaddr_entry`: an oversized entry already present
  in `discovery_peers` is purged by the heartbeat and produces no dial request.
- `honest_address_forms_replace_each_other_within_the_cap` (peer.rs): one endpoint advertised
  bare, then dialed with `/p2p`, then seen bare on an inbound connection: the set never grows past
  one entry and always holds the form most recently presented. Confirmed by mutation: cap 1 -> 2
  fails it, restored passes.
- `test_node_record_validation_rejects_oversized_multiaddr_list`: a signed record advertising
  exactly `MAX_MULTIADDRS_PER_PEER` addresses is accepted with every address intact; one address
  more is rejected.
- All four tests are deterministic (fixed TEST-NET-3 addresses, no sleeps, no timing assumptions).
  The two ingest tests were confirmed by mutation (cap guard neutralized: 0 passed, 2 failed;
  restored: 2 passed). The two new tests were confirmed by mutation as well (see the follow-up
  commit for the exact mutants).
- The existing cap tests reference the constant symbolically, so they run unchanged at the new
  value.
- The `tn-network-libp2p` test suite is green: 255 passed, 0 failed (`tn_network_libp2p` lib test binary).
- Nightly `cargo fmt -- --check` green. Scoped nightly
  `clippy -p tn-network-libp2p --no-deps --all-targets --all-features -- -D warnings` reports no
  lints in the changed files (it does surface a pre-existing `collapsible_match` in
  `tests/network_tests.rs`, untouched by this PR, under a nightly newer than CI's).

### DIFF
--- a/crates/network-libp2p/src/peers/peer.rs
+++ b/crates/network-libp2p/src/peers/peer.rs
@@ -24,6 +24,10 @@ use tracing::{error, warn};
 /// over it (`known_ip_addresses`) bounded (GHSA-29v6-gvv5-45gx). A legitimate node advertises a
 /// single address, so the cap is never reached in normal operation, and address refresh and key
 /// rotation stay well within it.
+///
+/// The discovery path reuses this value as the per-entry ceiling in `eligible_for_discovery`:
+/// a PeerExchange entry with more addresses than this set can hold cannot come from an honest
+/// peer's store, so it is rejected before it reaches `discovery_peers` (issue #1183).
 pub(crate) const MAX_MULTIADDRS_PER_PEER: usize = 32;
 
 /// Information about a given connected peer.
--- a/crates/network-libp2p/src/peers/peer.rs
+++ b/crates/network-libp2p/src/peers/peer.rs
@@ -20,15 +20,33 @@ use tracing::{error, warn};
 /// `register_outgoing`) and by the peer's own self-advertised kad `NodeRecord`s (`update_net`).
 /// A non-committee observer can republish its own signed record without bound (a record with a
 /// fresher timestamp is always accepted), so the advertised path is attacker-sustained. Capping
-/// the set keeps a single peer entry bounded in memory and keeps the reputation and ban scans
-/// over it (`known_ip_addresses`) bounded (GHSA-29v6-gvv5-45gx). A legitimate node advertises a
-/// single address, so the cap is never reached in normal operation, and address refresh and key
-/// rotation stay well within it.
+/// the set keeps a single peer entry bounded in memory and keeps the peer-exchange payload built
+/// from it (`exchange_info`) bounded (GHSA-29v6-gvv5-45gx).
+///
+/// The value is derived from the forms one honest address takes in this set. The protocol
+/// advertises exactly one address per node (`NodeRecord::build`, the committee `network_address`),
+/// but the set is keyed on exact `Multiaddr` equality and one endpoint appears in up to two
+/// syntactic forms, with and without the `/p2p/<peer_id>` suffix: the advertised form is whatever
+/// the operator configured (`keytool` writes it with `/p2p`, test fixtures without), the dialed
+/// form always carries `/p2p` (libp2p-swarm appends it to every dial and `register_outgoing`
+/// stores the address the connection was established on), and the inbound observed form
+/// (`send_back_addr`) never does. libp2p-quic dials from the listening socket, so for a peer
+/// whose observed source endpoint matches its advertised one (no NAT rewrite) the inbound form is
+/// its own listen address and adds no third endpoint. Two forms for the current address plus two
+/// for the previous one while an in-process address change is in flight gives 4. Such a peer stays
+/// within it, so nothing is evicted in normal operation. If a peer ever exceeds it (a NAT-rewritten
+/// endpoint adds a third form), eviction only trims the peer-exchange payload (`exchange_info`):
+/// dialing reads `known_peers`, kad and `discovery_peers`, and banning reads
+/// `observed_ip_addresses`, never this set. A tight value also bounds the dial fan-out one
+/// discovery entry can cause.
 ///
 /// The discovery path reuses this value as the per-entry ceiling in `eligible_for_discovery`:
 /// a PeerExchange entry with more addresses than this set can hold cannot come from an honest
-/// peer's store, so it is rejected before it reaches `discovery_peers` (issue #1183).
-pub(crate) const MAX_MULTIADDRS_PER_PEER: usize = 32;
+/// peer's store, so it is rejected before it reaches `discovery_peers` (issue #1183). The record
+/// validation cap `MAX_ADVERTISED_MULTIADDRS` is tied to this value, so validation and storage
+/// agree on how many addresses one peer may present: a record can never carry more addresses than
+/// the store keeps for a peer.
+pub(crate) const MAX_MULTIADDRS_PER_PEER: usize = 4;
 
 /// Information about a given connected peer.
 /// Note that bls_public_key and network_key are Optional.
@@ -156,12 +174,14 @@ impl Peer {
     ///
     /// A newly seen address is always admitted, so the most recent address a peer presents (a
     /// connection witnessed via `register_incoming` / `register_outgoing`, or the address it
-    /// advertises on a rotated network key via `update_net`) is always recorded for the ban path,
-    /// which reads [`Self::known_ip_addresses`]. If admitting it pushes the set over the cap, an
-    /// older address is evicted to restore the bound. Re-recording an address already present is a
-    /// no-op. A self-advertised republish flood therefore churns the set within the cap instead of
-    /// growing it without bound (GHSA-29v6-gvv5-45gx); a legitimate peer never approaches the cap,
-    /// so nothing is ever evicted.
+    /// advertises on a rotated network key via `update_net`) is always part of the peer-exchange
+    /// payload built by [`Self::exchange_info`]. If admitting it pushes the set over the cap, one
+    /// of the other addresses is evicted to restore the bound. Re-recording an address already
+    /// present is a no-op. A self-advertised republish flood therefore churns the set within the
+    /// cap instead of growing it without bound (GHSA-29v6-gvv5-45gx). The forms an honest peer's
+    /// address takes fit within the cap (see [`MAX_MULTIADDRS_PER_PEER`]), so nothing is evicted
+    /// in normal operation, and an eviction can only ever trim that payload: the ban path reads
+    /// [`Self::observed_ip_addresses`], not this set.
     fn note_multiaddr(&mut self, multiaddr: Multiaddr) {
         if self.multiaddrs.insert(multiaddr.clone())
             && self.multiaddrs.len() > MAX_MULTIADDRS_PER_PEER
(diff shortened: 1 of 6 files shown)


crates/network-libp2p/src/peers/peer.rs
```diff
@@ -24,6 +24,10 @@ use tracing::{error, warn};
 /// over it (`known_ip_addresses`) bounded (GHSA-29v6-gvv5-45gx). A legitimate node advertises a
 /// single address, so the cap is never reached in normal operation, and address refresh and key
 /// rotation stay well within it.
+///
+/// The discovery path reuses this value as the per-entry ceiling in `eligible_for_discovery`:
+/// a PeerExchange entry with more addresses than this set can hold cannot come from an honest
+/// peer's store, so it is rejected before it reaches `discovery_peers` (issue #1183).
 pub(crate) const MAX_MULTIADDRS_PER_PEER: usize = 32;
```
grantkee: This was introduced before this PR, but 32 is excessive. I thought the protocol assumes only 1 valid multiaddr? I'm not sure what would happen or how a validator could end up with 2+ valid multiaddr

grantkee (dismissed):
It's a good catch, but the `MAX_MULTIADDRS_PER_PEER` at 32 doesn't do much because the protocol assumes only 1 valid multiaddr unless I'm missing something. Approved since this was an existing const. We should address this value here or soon in a follow up PR @MavenRain

MavenRain:
I agreed that 32 was excessive.  Thanks for pushing on it.  I dug into where the number should come from and pushed a follow-up commit that derives it instead of reusing the old headroom.

The protocol does advertise exactly one address per node, but the stored set is keyed on exact `Multiaddr` equality and one honest endpoint shows up in it in two syntactic forms, with and without the `/p2p/<peer_id>` suffix: the advertised form is whatever the operator configured (`keytool` writes `network_address` with `/p2p`, the test fixtures advertise it bare), the dialed form always carries `/p2p` (libp2p-swarm appends it to every dial and `register_outgoing` stores the address the connection was established on), and the inbound observed form (`send_back_addr`) never does.  Inbound connections add no third endpoint because libp2p-quic dials from the listening socket, so an honest `send_back_addr` is the peer's own listen address (for a peer whose observed source endpoint matches its advertised one, i.e. no NAT rewrite).  I measured this on a real two-node QUIC connection in the test harness: the dialer's store for its peer held `n=2` (the bare advertised form + the `/p2p` dialed form), the listener's store for the dialer held `n=1` (exactly the dialer's advertised address).

So the changes in the new commit:

- `MAX_MULTIADDRS_PER_PEER`: 32 -> 4.  Two forms for the current address plus two for the previous one while an in-process address change is in flight.  An honest peer never crosses it, so nothing is evicted in normal operation.  If a peer ever does exceed it, eviction only trims the peer-exchange payload: dialing reads `known_peers` / kad / `discovery_peers` and banning reads `observed_ip_addresses`, never this set.  It also caps the dial fan-out of one discovery entry at 4 instead of 32.
- `MAX_ADVERTISED_MULTIADDRS`: 16 -> tied to `MAX_MULTIADDRS_PER_PEER`, so validation and storage agree on how many addresses one peer may present.  Same reasoning as yours: a legitimate record carries exactly one address, so 16 was headroom without a derivation.
- I corrected a claim in the PR description: the kad closest-peers ingest was not "record-capped at 16".  Those addresses come from remote responders' routing tables (`KadPeer.multiaddrs` in FIND_NODE replies) and never pass through `peer_record_valid`, so the new `eligible_for_discovery` guard is the first bound on that path too.  Comment in `eligible_for_discovery` and the PR body updated.
- Tests: `honest_address_forms_fit_within_the_cap` (peer.rs) exercises the exact derivation, the forms with and without `/p2p` for two addresses fill the cap and none is evicted, and `test_node_record_validation_rejects_oversized_multiaddr_list` (network_tests.rs) checks the record cap boundary (at cap accepted, cap + 1 rejected).  The existing cap tests all reference the constant symbolically, so they run unchanged at the new value.

If you would rather see the correct count land at 1, the follow-up would be to normalize `/p2p` off on ingest and drop a superseded address when a fresher record arrives;  then the cap could go to 2.  I kept that out of this PR to keep the change small.  I'd be happy to open it as a follow-up if you want it.

MavenRain:
Following up from stand-up: we only ever consume one address per peer, so I pushed a commit that takes `MAX_MULTIADDRS_PER_PEER` down to 1 (and `MAX_ADVERTISED_MULTIADDRS` follows, since it is tied to it).

With the cap at 1 the stored set is just the address the peer most recently presented. The bare and `/p2p/<peer_id>` forms of one endpoint replace each other instead of accumulating (either form dials the same endpoint), and an eviction only trims the peer-exchange payload, since dialing reads `known_peers` / kad / `discovery_peers` and banning reads `observed_ip_addresses`. The discovery per-entry ceiling drops to one address with it, so one PeerExchange entry can now cause at most one dial.

One thing worth a look: a sender still on the old cap (32) can hold both forms of an endpoint for a peer it dialed, so its exchange entry for that peer carries 2 addresses and a node on this branch rejects it as ineligible for discovery. That only removes a discovery candidate during the rollout window (kad closest-peers and committee dialing are unaffected), and once senders run this branch every entry carries exactly one address. If you would rather have ingest tolerate that during rollout (take the first address instead of rejecting the entry), I can add it here.

Test `honest_address_forms_fit_within_the_cap` is replaced by `honest_address_forms_replace_each_other_within_the_cap` (advertised bare, dialed with `/p2p`, seen bare inbound: the set never grows past one and holds the most recent form). Confirmed by mutation (cap 1 -> 2 fails it). The other cap tests reference the constant symbolically and run unchanged.

grantkee (approved):

sstanfield (approved):

PR #1196, opened 2026-08-14
Do not require certification for an epoch record when starting a new epoch.

sstanfield:
This should be safe since we will have generated and saved a record that has to be correct if we have not forked.


crates/node/src/manager/node/run_epoch.rs
```diff
@@ -565,231 +520,75 @@ where
...
-        } else {
-            (previous_epoch_rec.digest(), previous_epoch_rec)
+            // Invariant: every node seals and durably persists epoch N-1's record at that epoch's
+            // close before advancing (`write_epoch_record`), and an imported node receives a
+            // contiguous, certificate-verified record chain from genesis. State sync never runs
+            // execution past the current epoch (`handle_sync_output` no-ops output beyond it), so a
+            // node crosses every boundary through `write_epoch_record`. A missing record here is
+            // therefore a corrupted or incomplete datadir, not a recoverable sync gap - fail loudly
+            // rather than waiting on a peer that cannot supply what a valid node already has.
+            self.consensus_chain.epochs().record_by_epoch(previous_epoch).await.ok_or_else(
```
MavenRain: **HIGH: a crash between `close_epoch()` and `write_epoch_record()` now bricks the node permanently.**

The two writes are sequential and hit two independent stores (run_epoch.rs:437-446, also :253/257 and :503/506). `close_epoch()` commits the boundary block to the reth DB, and `concludeEpoch` runs inside that block, so the canonical tip already reports epoch N+1 when it returns. `write_epoch_record()` persists epoch N's record afterward, and it does several async chain reads first, so the window is not small.

If the node crashes in that window, every restart then fails here: `get_committee_with_epoch_start_info` resolves committee.epoch()=N+1 from the tip, `record_by_epoch(N)` returns `None`, and this hard error fires before any replay arm that could rebuild the record. The `requested_missing_epoch` nudge above fetches certificates only. The base branch handled exactly this gap with a bounded peer wait, which this PR deletes.

The error text also misleads the operator: the datadir is not corrupt, and the message tells them to investigate a corruption that did not happen.

Suggested fix: on a miss, recompute the record locally. It is deterministic from on-chain state, which is this PR's own core argument, and `write_epoch_record` shows every input is re-derivable. A bounded peer-wait fallback also works.

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -565,231 +520,75 @@ where
...
         };
+
+        // Anchor the epoch-close seed message on the digest of the record sealed for epoch N-1
+        // (the default digest for epoch 0, which has no prior record). The record is computed
+        // deterministically from the pinned closing block, so every honest node derives the same
+        // digest and no certificate is needed to trust it here: a divergent record implies a fork,
+        // which a certificate could not repair. Proposer signing/verification of the seed message
+        // stays fork-gated elsewhere; only the "anchor must be certified" coupling is dropped.
+        let prior_epoch_record =
+            if current_epoch == 0 { EpochDigest::default() } else { previous_epoch_rec.digest() };
```
MavenRain: **HIGH: a local record divergence now becomes a silent whole-epoch exclusion instead of a loud halt.**

The deleted threat model still holds at head. This digest is captured once per epoch into the `ConsensusConfig` and is never re-read. The record store is write-once and is never reconciled against peers. If this node's local record diverges for any reason short of a real fork (a local DB fault, an execution bug), the node signs and verifies seed messages against a digest no peer shares. Peers reject with `InvalidSeedSignature`. With n=5 and quorum 4, that node contributes nothing for the whole epoch, and two such nodes stall the epoch. Nothing self-heals, and nothing logs at error level on the divergent node.

The old `certified_prior_epoch_anchor` turned this state into a loud, restartable error. I agree the record is deterministic for honest steady-state nodes (I traced every `build_epoch_record` input). What this PR removes is the tripwire, not happy-path correctness. Consider a cheap self-check at epoch open, for example: compare the local digest against the certificate when one is already stored, and log at error level on mismatch.

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -565,231 +520,75 @@ where
...
-            // to a higher epoch (e.g. 3 while we are opening epoch 3 with previous_epoch=2),
-            // keep the higher value so the collector retries that epoch too.
-            let current = *self.consensus_bus.requested_missing_epoch().borrow();
-            self.consensus_bus.requested_missing_epoch().send_replace(current.max(previous_epoch));
-            rec
-        } else if current_epoch == 0 {
+
+        // Nudge the epoch-record collector to backfill missing epoch *certificates* only.
+        // Self-closed records are stored uncertified; their certs arrive later from epoch-vote
+        // gossip or peer recovery. This is purely cert availability - it does not gate record
```
MavenRain, line 552: **HIGH: after a fleet-wide vote-quorum failure, the certificate gap is now permanent and mostly silent.**

The race that lost the votes at the 383 close is still reachable: nodes do not re-vote after a restart (the vote path is driven by the in-memory `epoch_record_watch`, which nothing re-fires from durable storage), and `manage_epoch_votes` is a one-shot bounded task. This nudge cannot help, because the record collector fetches certified pairs only, and no peer holds a certificate.

Old behavior: loud fleet halt. New behavior: the fleet keeps producing blocks (good, that is the goal), but no mechanism at head can ever produce the missing certificate. Consequences, permanently: peers refuse to serve that epoch's record (`handler.rs:1166` serves certified records only), so new-node onboarding and behind-node catch-up stall at that epoch; state export skips from that epoch onward; `tn_epochRecord` returns NotFound for a record the node holds. Those three gates pre-date this PR, but this PR makes the state they gate on survivable and therefore durable. Operator signal is warn/info noise only.

Suggested fix: land this together with a re-vote-on-restart mechanism (PR #1193 carries one), or add a certificate backfill path.

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -565,231 +520,75 @@ where
...
+        // presence below. Never decrease it: the gossip handler may already have requested a
+        // higher epoch (e.g. 3 while we are opening epoch 3 with previous_epoch=2).
+        let requested = *self.consensus_bus.requested_missing_epoch().borrow();
+        self.consensus_bus.requested_missing_epoch().send_replace(requested.max(previous_epoch));
+
+        let previous_epoch_rec = if current_epoch == 0 {
+            // Genesis: no prior record to seal. Seed the chain from a filler carrying the genesis
+            // committee (identical to the `save_dummy_epoch0` record); the anchor below is the
+            // default digest. Gated on `current_epoch == 0`: by epoch 1 the real epoch-0 record has
```
MavenRain: **MEDIUM (latent): the epoch-0 dummy vs real record distinction is now unenforced.**

`try_record_by_epoch` returns the epoch-0 dummy while the record index is empty, and the caller cannot tell it from a real record. Today this is safe: the epoch loop is strictly sequential (`write_epoch_record` is awaited before `NewEpoch` returns), the dummy resets on DB reopen, and run_epoch.rs:220-226 guards the restart case. But the test doc at `epoch_records.rs:2473-2482` still warns that the dummy answers `record_by_epoch(0)` in exactly the epoch-1-open window, and the deleted certified path was the enforcement. Any future concurrency between the epoch-0 close and the epoch-1 open silently anchors epoch 1 on the dummy digest.

Suggested: add an explicit dummy-digest check here, or at least a load-bearing comment at this read site.

crates/node/src/manager/node/run_epoch.rs
```diff
@@ -565,231 +520,75 @@ where
...
-                .certified_prior_epoch_anchor(&committee, previous_epoch, &task_spawner)
-                .await?;
-            (certified.digest(), certified)
-        } else {
-            (previous_epoch_rec.digest(), previous_epoch_rec)
+            // Invariant: every node seals and durably persists epoch N-1's record at that epoch's
+            // close before advancing (`write_epoch_record`), and an imported node receives a
+            // contiguous, certificate-verified record chain from genesis. State sync never runs
+            // execution past the current epoch (`handle_sync_output` no-ops output beyond it), so a
+            // node crosses every boundary through `write_epoch_record`. A missing record here is
```
MavenRain: **MEDIUM: the new behaviors have zero test coverage.**

The PR deletes two tests and adds none. Untested: this hard-error path on a missing prior record, a post-fork epoch open with an uncertified anchor, and the de-recursed poll loop in `epoch_records.rs`. The crash-window scenario from my comment at line 569 is directly testable: persist the boundary block, skip the record write, restart, and assert the node recovers instead of crash-looping.

crates/storage/src/epoch_records.rs
```diff
@@ -688,15 +675,15 @@ impl EpochRecordDb {
...
-            let retry = outcome.as_ref().err().is_some_and(|e| e.is_retryable())
+            let mut outcome = self.certified_record_by_epoch(epoch).await;
+            let mut retry = outcome.as_ref().err().is_some_and(|e| e.is_retryable())
                 && tokio::time::Instant::now() < deadline;
-            if retry {
-                tokio::time::sleep(Duration::from_millis(200)).await;
-                self.certified_record_poll(epoch, deadline).await
-            } else {
-                outcome
+            while retry {
```
MavenRain, line 692: **MEDIUM: the de-recursion dropped the 200ms sleep, so this loop busy-spins.**

The base version slept 200ms before each retry. This loop retries immediately, so it hammers the DB actor in a hot loop until the deadline. The recursion removal itself is correct and closes the stack-overflow failure seen in production.

Mitigating: after this PR the `certified_record_by_epoch_with_timeout` family has no production callers left (test-only). So either restore the sleep (`POLL_INTERVAL` is still live, `poll_until_deadline` shows the idiom) or delete the now-dead family outright.

MavenRain (commented):
Full review of this PR at head 9fce2d3a. The core determinism argument holds: I traced every `build_epoch_record` input and all of them come from consensus-agreed state. The removal of the deadlock is real. But the PR deletes three safety nets and does not replace them, and one deletion adds a new permanent-failure mode. Six inline comments below carry the specific findings. One more finding has no diff anchor:

**Doc drift in three files the diff does not touch.** (1) `crates/node/src/manager/node/start_epoch.rs:307-311` still says `open_epoch_pack` guarantees a certificate-backed digest. This PR removes that guarantee. (2) `crates/storage/src/epoch_records.rs:45-46` and `:650` still link to the deleted `record_by_epoch_with_timeout`. These rustdoc links are now broken. (3) `crates/types/src/forks.rs:273` still says epoch 1 takes the post-fork certified-anchor path. That path no longer exists.

Also note: CI skipped all real gates on this PR (clippy, fmt, test). The green check did not compile this diff.

crates/node/src/manager/node/run_epoch.rs
sstanfield: Yep, tests found this and fixing it now.

crates/node/src/manager/node/run_epoch.rs
sstanfield: Not sure I agree with this.  The examples of a local DB fault or execution bug will both leave a node in an unusable state, no amount of asking peers for epoch records will fix that.  I think this is like saying if the consensus DB or Reth DB get corrupted the node should try to recover on it's own.  It can't and trying to do seems like a lot of wasted code.  Well maybe it could by wholesome wiping of data and rebuilds but that seems like a task to be done by an operator intentionally.  I don't think this at all invalidates the core premise of deterministic EpochRecords.  Maybe I am missing something.

crates/node/src/manager/node/run_epoch.rs
sstanfield: I think revoting is out of scope for this PR.  This PR is basically decoupling the consensus EpochRecord logic from the network which I think is a big win.  We can address this in another network related PR.  It may also be that if we ever have a missing cert then it might require manual intervention from the old committee nodes (in an extreme case).  Building an epoch re-vote in as part of a network refactor is a good idea though I just don't think it belongs here.

crates/node/src/manager/node/run_epoch.rs
sstanfield: I have add a check for the dummy epoch 0 record and use it to filter out the dummy record here and when trying to rebuild the epoch record.

crates/node/src/manager/node/run_epoch.rs
MavenRain: Fair enough . . . opened https://github.com/Telcoin-Association/telcoin-network/issues/1198 to support this

MavenRain:
One robustness ask on the new recovery guard in `recover_previous_epoch_record` (close_epoch.rs):

When record N-1 is absent and the executed tip is not exactly epoch N-1's closing block, the guard hard-errors out of `run_epoch` on every boot.  I agree fail-loud is right here: state-sync cannot create that state (subscriber.rs no-ops sync output beyond the current epoch), records are write-once, and `open_epoch_pack` gates epoch entry on record presence.  So, the state is reachable only through corruption or import damage.

The ask is only about the message.  The guard's error ("refusing to re-derive from an unexpected block") does not carry the operator guidance the sibling corruption path has ("corrupted or incomplete datadir, do NOT delete chain-data").  An operator who hits this in a boot loop gets no hint that deleting chain-data makes it worse. I suggest harmonizing the message with the sibling path, or falling through to the standard corrupted-datadir error.

MavenRain:
One latent-invariant hardening ask on the recovery path's `parent_state` pin:

The load-bearing comment at close_epoch.rs:328-333 enumerates three `write_epoch_record` call sites, each guarded by `wait_for_consensus_execution`.  The new recovery call site is a fourth, and it has no such wait.  Here is why it is still correct today: `try_restore_state` (node.rs:717) seeds the `recent_blocks` watch from the persisted canonical tip, and nothing reaches the engine before the recovery call in `run_epoch` (`to_engine` moves into `run_epochs` at node.rs:941, and its first send is the replay in run_epoch.rs:258).  The pinned parent state is exactly the closing block.

That ordering is enforced nowhere.  If a future change reorders recovery after `replay_missed_consensus`, the pin silently points at a newer block. `build_epoch_record` copies `final_state` through unchecked (the continuity check only covers committee handoff), so the failure mode is a divergent record, not a halt.

The cheap fix, both values are already in hand: in `recover_previous_epoch_record`, assert `parent_state == closing_block.num_hash()`, and add the recovery call site to the comment's enumeration.  That converts a silent future divergence into a loud error at the point of breakage.

MavenRain:
Small error-reporting nit in `boundary_consensus_digest`.

`closing_block.parent_beacon_block_root.unwrap_or_default()` turns a missing root into the zero digest.  The nonce check can still pass, and the subsequent lookup then reports "boundary consensus header 0x00..00 is absent".  That message points the operator at a missing header when the actual problem is a header field that was never set.

Suggest making a missing `parent_beacon_block_root` its own explicit error so the two failure modes stay distinguishable.

MavenRain (dismissed):
See prior comments, but nothing blocking

MavenRain (approved):

grantkee (approved):
Good work @sstanfield

Issue #1086, opened 2026-08-04
Rollout plan: versioned header + #1062 seed signature on adiri


MavenRain:
Scope:  ship PR #1062 (mandatory `seed_signature` on the primary header) together with the green-lit header versioning, cut the wire over cleanly, and migrate persisted data so no heuristic legacy decode lives in the node runtime.  File and line references are against current main unless marked (1032) for the PR branch.

## 0. Summary of the proposal

1. **Version the header in #1062** with a one-byte BCS enum envelope inside `Header`'s existing custom serde impls:  `V0` is the legacy field layout, `V1` is the #1062 layout.  Digests stay per-variant over the untagged inner bytes, so every historical digest, vote signature, aggregate signature, and consensus-chain parent link keeps verifying.
2. **Yes to the standup question:  scheduled downtime plus a one-time rewrite is better than carrying fallback behavior**, with one precision:  the rewrite must be a *rewrap* (decode old bytes with the legacy layout, re-emit them tagged as `V0`), not a conversion to `V1`.  Converting old headers to `V1` is not available to us:  digests are recomputed from decoded bytes on every deserialize (header.rs:193-203), so changing historical bytes to a new shape severs every vote, aggregate, and parent-hash link in the archive.  After the rewrap, all persisted bytes are uniformly version-tagged, the node runtime does pure version dispatch (no byte-sniffing dual decode), and the only code that ever parses untagged legacy bytes is the offline migration tool itself.
3. **Wire cutover by partition, #970-style:**  bump the req-res protocol string `/0.0.2` to `/0.0.3`, bump the sync protocol `/0.0.1` to `/0.0.2`, and version the gossip topic string.  Old and new binaries then never exchange a header-bearing message on any channel, so no wire fallback exists either.
4. **Coordinated single-boundary flip on adiri** (5 validators):  stop each node right after a clean epoch close, run the migration CLI, start the new binary.  Observers and RPC nodes run the same CLI, so "everyone else must resync" becomes optional rather than forced.

## 1. The double-check:  effect of #1062 on consensus-DB deserialization

Full table-by-table audit (14 tables, crates/storage/src/lib.rs:82-109, three physical sub-DBs selected per-table in composite_db.rs:39-44), plus the pack stores.  Verified by an adversarial re-read pass.

**Header-bearing stores and their failure modes under a #1062 binary without versioning:**

[...]

PR #1197, opened 2026-08-15
fix(consensus): fork-gate strictly increasing commit timestamps

### LINKED ISSUE #1191: Same-second ConsensusOutputs overwrite each other's ConsensusHeader digest in the repurposed EIP-4788 beacon-roots ring buffer
**Suggested title:** Same-second `ConsensusOutput`s overwrite each other's `ConsensusHeader` digest in the repurposed EIP-4788 beacon-roots ring buffer

**Severity:** Informational / design-level protocol correctness.  No production code in this repo reads the beacon-roots contract storage today (the only in-repo readers are test assertions at `crates/engine/tests/it/main.rs:82,95`).  The loss hits the intended external use of the buffer: trustless in-EVM proofs of consensus output digests (bridges, light clients, future contracts).
**Attacker model:** None required.  The collision happens organically whenever two consecutive outputs commit in the same wall-clock second.  A Byzantine leader can also force it (timestamps are only validated as `>=` the previous header), but honest operation alone produces it.

**Credit:** surfaced by Haxatron in the current collaborative review.  This issue records the finding with repo-level evidence.

## Summary
Telcoin Network repurposes the EIP-4788 pre-block system call: each executed block's `parent_beacon_block_root` header field carries the `ConsensusHeader` digest of the output that produced it (`crates/tn-reth/src/payload.rs:136`), and once per output (first batch only) that digest is written into the canonical EIP-4788 beacon-roots contract (`crates/tn-reth/src/evm/block.rs:1141-1192`, gated by `first_batch()` at `:1274-1278`).

The contract keys its ring buffer purely by timestamp: it stores the timestamp in slot `timestamp % 8191` and the root in slot `timestamp % 8191 + 8191`.  The block timestamp is `output.committed_at()` in whole seconds (`payload.rs:114`), and `commit_timestamp` is only monotonic **non-decreasing**: `previous_sub_dag_ts.max(leader.created_at())` (`crates/types/src/primary/output.rs:399`, with an explicit auto-correct log for the tie/decrease case at `:401-404`).  With `min_header_delay = 1000ms` / `max_header_delay = 2500ms` defaults (`crates/config/src/node.rs:250-255,327-334`) and second-granularity stamps, two consecutive outputs routinely share a second.

When outputs O1 and O2 share timestamp T, their digests D1 != D2 (the digest hashes the chain-incrementing `number` field, `crates/types/src/primary/block.rs:64`), but both writes target the same two storage slots.  O2's write silently replaces D1.  From any later state, the contract's `get(T)` returns only D2.  D1 is permanently unretrievable through the standard EIP-4788 in-EVM proof path, which is the trustless, archive-free mechanism the repurposing is meant to provide.

## Why Ethereum does not have this problem
On L1, header validity requires `timestamp > parent.timestamp`, so a same-key different-root overwrite is impossible;  the only overwrites are ring-buffer aging after 8,191 seconds.  Telcoin bypasses reth's header-chain validation entirely (`crates/tn-reth/src/traits.rs:6,54,82-92`;  blocks are BFT-final and canonicalized directly), and consensus header validation deliberately accepts `>=` timestamps so that sub-second header production is not blocked (`crates/consensus/primary/src/network/handler.rs:837-851`).  Equal consecutive timestamps are therefore a designed-for condition, not a corner case: every block inside one multi-batch output already carries the identical timestamp (`crates/engine/src/payload_builder.rs:189-214`).

## Evidence that the deployed contract has last-write-wins semantics
- `chain-configs/mainnet/genesis.yaml:61-64` and `chain-configs/testnet/genesis.yaml:57-60` deploy the byte-for-byte canonical EIP-4788 runtime bytecode at `0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02`.  The bytecode has no versioning or collision handling;  a repeated-timestamp `set()` overwrites both slots.
- The repo's own test suite encodes the slot arithmetic (`crates/engine/tests/it/main.rs:53,76-95`) and even constructs the exact collision scenario unintentionally: the multi-output tests (around `:1069-1121`, `:1600-1637`, `:1984-2012`) build consecutive outputs whose `commit_timestamp` values are equal with distinct `consensus_header_digest` values.

## Secondary concern: window length at TN cadence
The buffer is keyed by wall-clock second, so it covers at most 8,191 distinct seconds of history, about 2 hours 16 minutes, regardless of how many outputs commit.  Ethereum's 12-second slots stretch the same buffer to about 27.3 hours.  Any consumer that needs a longer proving window than ~2.27 hours cannot rely on the contract at all.  Same-second collisions shrink the effective coverage further because tied outputs collapse to one retrievable digest.

## Impact
- The in-EVM proof path (`get(T)` on the beacon-roots contract) can attest at most one consensus output per second.  Digests of same-second predecessors are silently dropped.  Two honest parties can hold conflicting "the digest at T" views depending on when they queried.
- The per-block header field is untouched: O1's blocks keep `parent_beacon_block_root = D1` forever, and node-internal consumers (`crates/node/src/engine/inner.rs:62`, catch-up in `crates/consensus/primary/src/consensus_bus.rs:786`) read the header field, not the contract, so consensus and recovery are unaffected.
- An indirect in-EVM proof of D1 remains possible through the EIP-2935 blockhashes buffer (keyed by unique block number) plus a header preimage, within its own 8,191-block window.  This is a workaround, not the advertised 4788 interface.

## Suggested directions
1. Make the commit timestamp strictly increasing at the consensus layer: `commit_timestamp = max(previous_sub_dag_ts + 1, leader.created_at())`.  One line, but consensus-visible (fork required), and it lets the recorded time drift ahead of wall clock while sustained commits arrive faster than one per second.
2. Replace the timestamp key with an always-unique key for the repurposed buffer: deploy a modified ring-buffer contract keyed by `nonce` (`epoch << 32 | round`) or by block number instead of the canonical 4788 bytecode.  This also fixes the window concern (8,191 outputs instead of 8,191 seconds) at the cost of tooling compatibility with stock EIP-4788.
3. If neither change is wanted, document the limitation explicitly: the buffer attests "a digest that committed at second T", not "the digest of output N", and consumers needing per-output attestation must use the header field via the EIP-2935 path.

MavenRain:
Closes #1191. (Cantina #7)

## Problem

Two consecutive `ConsensusOutput`s routinely commit in the same wall-clock second: `commit_timestamp` is only monotonic non-decreasing (`previous_sub_dag_ts.max(leader.created_at())`), consensus header validation deliberately accepts `>=` timestamps, and the default header delays are 1000/2500 ms against second-granularity stamps. The repurposed EIP-4788 pre-block system call writes each output's `ConsensusHeader` digest into the canonical beacon-roots contract, which keys its ring buffer purely by `timestamp % 8191`. When outputs O1 and O2 share second T, both writes target the same two storage slots and O2 silently replaces O1's digest. From any later state, `get(T)` returns only O2; O1's digest is permanently unretrievable through the standard in-EVM proof path that the repurposing is meant to provide. No attacker is required . . . honest operation alone produces the tie (a Byzantine leader can also force it, but the fix does not depend on who created it).

## Root cause

Ethereum never hits this because L1 header validity requires `timestamp > parent.timestamp`, so a same-key different-root overwrite is impossible. Telcoin bypasses reth's header-chain validation (blocks are BFT-final) and accepts equal consecutive timestamps by design, but kept the L1 contract's timestamp-keyed buffer unchanged. The uniqueness assumption the contract encodes was never re-established at the consensus layer.

## Fix

- [x] `crates/types/src/forks.rs`: new epoch fork gate `strict_commit_timestamp_active(epoch)` with `STRICT_COMMIT_TIMESTAMP_FORK_EPOCH = u32::MAX` on adiri (dormant; standard two-phase rollout, a follow-up PR sets the concrete epoch exactly as the seed-signature fork did in #1086). Non-adiri builds are active from genesis and carry no legacy period.
- [x] `crates/types/src/primary/output.rs`: the commit-timestamp derivation moves into a pure helper `resolved_commit_timestamp`. Post-fork the floor is `previous + 1` (saturating), so no two chained commits of an epoch share a second; pre-fork stays byte-identical to the historical clamp, so already-committed sub-dag digests are reproduced exactly.
- [x] The gate keys on `leader.epoch()` (the epoch inside the value being derived), never node-local committee state, for the same reason documented on `seed_signature_active`: the timestamp feeds the sub-dag digest, so a mixed fleet would fork on the first tied pair without an epoch-boundary cutover. `CommittedSubDag::new` is the single production derivation point (`bullshark.rs`); replay and pack verification consume the stored value and never re-derive it.
- [x] Observability: a tie-bump now emits a `debug!` line (the existing `warn!` only covers the decrease case, so the exact scenario this fork corrects would otherwise never log). Debug rather than warn because ties are routine during faster-than-one-per-second bursts.
- [x] Test-only fork override, mirroring the seed-signature fork: `TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH` (`crates/types/src/forks.rs`, compiled only under `test-utils`; production binaries cannot be repointed). The e2e harness pins the fork dormant on every spawned node (`TestBinary::command`, next to the existing seed-signature pin) and the Makefile forwards the variable on the e2e lanes, defaulting to dormant.

The recorded time can run ahead of wall clock while commits arrive faster than one per second, and the drift shrinks only when a real second passes with no commit. Production cadences (a 1 s minimum header delay, one leader per two rounds) do not sustain faster-than-one-per-second commits, so the floor is inert there in practice. The e2e networks do sustain it (250-500 ms header delays), which made chain time run about twice as fast as wall time and broke the harness's host-clock-vs-block-timestamp phase math (`wait_for_mid_epoch` computes `now.saturating_sub(block_timestamp)`, which pins at 0 once chain time passes the host clock); that is why the e2e lane runs the fork dormant via the override. Every downstream consumer (block timestamps, epoch-boundary checks) reads this same chain; no consumer re-validates it against a local clock.

## Out of scope (documented residuals)

- Cross-epoch ties: the first commit of an epoch has no previous sub-dag (the per-epoch consensus pack starts empty), so it can still tie the prior epoch's closing commit. Documented in the fork const and the helper docs; closing it needs the epoch manager to thread the prior epoch's closing timestamp into consensus startup, which deserves its own PR.
- The buffer still covers at most 8,191 seconds (about 2 h 16 min) of history; that window is a property of the deployed contract, not of this formula.
- The EIP-2935 blockhashes path (keyed by unique block number) is untouched and remains the indirect proof route for pre-fork history.
- While chain time is ahead of wall clock (sustained sub-second commit cadences only), guards built on `TimestampSec::elapsed()` saturate to zero and pass vacuously (for example the CVV rejoin staleness check in `subscriber.rs`). Production cadences do not sustain that regime and the e2e harness pins the fork dormant, so this stays a documented property rather than a change in observed behavior; tightening those guards for future-dated timestamps deserves its own issue.

## Testing

- `crates/types/src/primary/output.rs` unit tests pin both formula arms case by case (tie, decrease, advance, no-previous, `u64::MAX` saturation) and drive `CommittedSubDag::new` end to end with a tied chained pair, split per build: default-feature builds must bump, adiri builds must keep the legacy tie while dormant. `crates/types/src/forks.rs` pins the gate's rollout contract per build. These run in both CI lanes (default workspace nextest and the adiri feature lane, which builds `tn-types`).
- New engine e2e test `test_same_second_outputs_keep_both_beacon_roots_digests`: executes two chained single-batch outputs whose leaders share one fixed second and asserts BOTH outputs' timestamp and digest slots from the FINAL canonical state. The existing per-block `assert_eip4788` checks read each block's own historical state, where the overwritten digest was still visible, so they could never catch the collision; the final-state read is where it used to land. Deterministic fixed second (no wall clock).
- Gate rollout-contract tests assert against the override-free build decision, and a new `test-utils` test pins the override inert when the variable is unset (both mirroring the seed-signature fork's tests). The `tn-types` suite also runs under `--features test-utils` locally to compile the override arm.
- Updated the pre-existing `tn-primary::it` test `test_monotonically_incremented_commit_timestamps`: it pinned the legacy clamp (a round-4 leader with an older `created_at` commits with a timestamp equal to the previous commit's), which the strict formula deliberately changes to `previous + 1`. The test now derives its expected value from `forks::strict_commit_timestamp_active`, so it pins the correct arm under both CI lanes.
- Mutation-confirmed: reverting the strict arm to the legacy clamp fails the tn-types tests and the engine e2e test; forcing the non-adiri gate dormant fails the gate-contract and end-to-end tests. Sources restored byte-exact afterwards.
- Local static ladder: nightly `fmt --all -- --check`, scoped clippy (`-p tn-types --no-deps --all-features --all-targets`, `-p tn-engine --no-deps` at attest scope, plus nightly passes on `tn-types` default and `--features adiri`), `nextest -p tn-types` (default and `--features adiri`), `nextest -p tn-primary --test it`, the engine test above, plus a multi-lens diff review. The dynamic tier (workspace nextest, nightly workspace clippy, the pinned `+1.94` attest) runs on CI and the second machine per the 2026-08-13 policy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,19 +144,29 @@ E2E_BIN := $(E2E_TARGET_ROOT)/e2e/telcoin-network
 # seed_signature_override_is_inert_when_unset requires a process WITHOUT the variable.
 TN_SEED_SIGNATURE_FORK_EPOCH ?= 4294967295
 
+# Strict-commit-timestamp fork epoch for the e2e lanes (#1191), same shape as the
+# seed-signature pin above: u32::MAX = fork DORMANT. The e2e harness itself defaults the
+# spawned nodes to dormant (TestBinary::command), because sub-second e2e header delays make
+# the strict floor advance chain time faster than wall clock; this variable exists so a
+# fork-active lane can override that default. Only test-utils builds consult it
+# (tn_types::forks::strict_commit_timestamp_fork_epoch_override). Same caveat as above:
+# never set it on a bare --workspace run (strict_commit_timestamp_override_is_inert_when_unset
+# requires a process WITHOUT the variable).
+TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH ?= 4294967295
+
 # run restart integration tests
 test-restarts: build-e2e-bin
-	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) cargo nextest run --run-ignored all test_restarts ;
+	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH=$(TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH) cargo nextest run --run-ignored all test_restarts ;
 
 # run epoch integration tests (same filter as the public-tests epoch line). The scheduled
 # Durable e2e lane (#1149) runs this and test-restarts with TN_TEST_MDBX_SYNC=durable
 # exported so every spawned node opens MDBX in Durable.
 test-epochs: build-e2e-bin
-	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) cargo nextest run -p e2e-tests --test it --run-ignored all test_epoch ;
+	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH=$(TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH) cargo nextest run -p e2e-tests --test it --run-ignored all test_epoch ;
 
 # run e2e tests
 test-e2e: build-e2e-bin
-	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features ;
+	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH=$(TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH) cargo nextest run -p e2e-tests --run-ignored ignored-only --all-features ;
 
 # run tests with coverage (using llvm-cov + nextest)
 coverage:
@@ -232,8 +242,8 @@ revert-submodule:
 # workspace tests that don't require faucet credentials
 public-tests: build-e2e-bin
 	TN_BIN_PATH="$(E2E_BIN)" cargo nextest run --workspace --exclude tn-faucet --no-fail-fast ;
-	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) cargo nextest run -p e2e-tests --test it --run-ignored all test_epoch ;
-	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) cargo nextest run --run-ignored all test_restarts ;
+	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH=$(TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH) cargo nextest run -p e2e-tests --test it --run-ignored all test_epoch ;
+	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH=$(TN_STRICT_COMMIT_TIMESTAMP_FORK_EPOCH) cargo nextest run --run-ignored all test_restarts ;
 
 # local checks to ensure PR is ready
 pr:
--- a/crates/consensus/primary/tests/it/output_tests.rs
+++ b/crates/consensus/primary/tests/it/output_tests.rs
@@ -94,9 +94,16 @@ fn test_monotonically_incremented_commit_timestamps() {
         tn_types::EpochSeedChainValue::genesis_placeholder(),
     );
 
-    // THEN the latest sub dag should have the highest committed timestamp - basically the
-    // same as the previous commit round
-    assert_eq!(sub_dag_round_4.commit_timestamp(), sub_dag_round_2.commit_timestamp());
+    // THEN the commit timestamp never decreases. While the strict fork is dormant (adiri
+    // builds pre-fork) the value clamps to the previous commit's timestamp. Once the fork is
+    // active (non-adiri builds are active from genesis) the floor is `previous + 1`, so every
+    // commit in an epoch records a unique timestamp (#1191).
+    let expected = if tn_types::forks::strict_commit_timestamp_active(0) {
+        sub_dag_round_2.commit_timestamp() + 1
+    } else {
+        sub_dag_round_2.commit_timestamp()
+    };
+    assert_eq!(sub_dag_round_4.commit_timestamp(), expected);
 }
 
 #[test]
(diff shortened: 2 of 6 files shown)


grantkee (changes requested):
I understand the issue, but this solution won't work. If we have 10 node committees, 2 rounds, and 5 batches, the subdag timestamp would end up +50s with this solution. I'm not sure this is worth solving or that it's necessarily an issue, but let's discuss more. If necessary, we can write a different contract that measures something other than timestamp?

grantkee:
My understanding of EIP-4788 is to make consensus data natively accessible to the EVM. Ethereum uses their beacon root, and TN uses the consensus chain. The EVM header fields are repurposed to include consensus chain data, so I think overwriting is sufficient for our needs. The last executed block of each consensus output is stored on the beacon root contract, which supports confirming the latest consensus block root on-chain. Consequently, I'm in favor of leaving the behavior as-is

MavenRain:
Thanks Grant.  I agree with this.  I will close this PR.

One clarification:  The `+1` floor applies once per committed sub-dag (one per `ConsensusOutput`), not per batch or per block: every block of an output carries the same `output.committed_at()` (`crates/tn-reth/src/payload.rs:114`).  So, committee size and batch count do not multiply the drift.  What multiplies it is commit cadence, and there your point stands: `should_create_header` proposes as soon as `header_num_of_batches_threshold` (default 5) digests are queued, without waiting `min_header_delay` (`crates/consensus/primary/src/proposer.rs:905-907`).  Under sustained load rounds are latency-bound, commits can exceed one per second, and the strict floor then advances chain time by one second per commit no matter how little wall time passed.  That drift only unwinds during idle seconds.  The PR body's claim that production cadence cannot sustain that was wrong.

On the 4788 point: agreed.  Nothing in-repo reads the contract for a historical timestamp, and node restore reads `parent_beacon_block_root` from the block header, not from contract storage, so a same-second overwrite only affects an external prover that wants the earlier output's digest through `get(T)`.  If that ever becomes a product need, a TN contract keyed by `ConsensusHeader.number` (something other than timestamp, as you said) seems right.

MavenRain:
Closing as a result of review discussion; see #1191.

PR #1201, opened 2026-08-17
fix(tn-reth): enforce payability in the TEL precompile dispatcher

### LINKED ISSUE #1190: tel precompile: dispatcher enforces staticness but not payability, so value attached to any selector is stranded at 0x7e1
## Summary

The TEL precompile dispatcher (`telcoin_precompile`, `crates/tn-reth/src/evm/tel_precompile/mod.rs:146`) enforces static-call write protection itself (`mod.rs:157-161`), because registering the address as a precompile bypasses the interpreter's checks.  The same reasoning applies to payability, and the dispatcher does not enforce it.  No code in the module reads `PrecompileInput::value`.  Every selector, including the view `totalSupply()`, accepts a nonzero call value.

The EVM credits the value to `0x7e1` before the precompile runs.  On the pinned dependency versions:

- `revm-handler` 15.0.0 journals the caller-to-target `BalanceTransfer` in `make_call_frame` (`frame.rs:170-180`) before it invokes `precompiles.run` (`frame.rs:192`).  On precompile success, `checkpoint_commit` keeps the transfer (`frame.rs:195`).  Only a precompile error reverts it (`frame.rs:200`).
- `alloy-evm` 0.27.3 populates the field the dispatcher would need: `PrecompileInput { .., value: inputs.call_value(), .. }` (`src/precompiles.rs:468-472`, field at `:594`).  The guard is one comparison away, in the same place as the `is_static` guard.

So a successful `totalSupply()` call with attached value both returns the supply and permanently credits `0x7e1`'s native balance.

## Why the stranded value is unrecoverable

No selector moves `0x7e1`'s native balance out to any address:

- `handle_burn` only destroys from it: `balance_decr(internals, TELCOIN_PRECOMPILE_ADDRESS, amount)` (`burnable.rs:376`), then decrements the supply slot (`burnable.rs:306-312` pattern, slot 100 per `mod.rs:111`).
- `handle_claim` credits the recipient without debiting `0x7e1` (`burnable.rs:294`).
- `handle_mint` touches no balances.

Value that lands at `0x7e1` outside the burn lifecycle can only ever be destroyed by a future governance `burn`, never returned.  `TOTAL_SUPPLY_SLOT` is a separate counter and is not corrupted by a stray transfer, but `0x7e1`'s balance stops corresponding to any tracked quantity.

Note the asymmetry with plain sends: a bare value transfer with empty calldata does revert, because the dispatcher's `input.data.len() < 4` check (`mod.rs:147-149`) errors the frame and the journal revert returns the value.  The naive send path is already rejected.  A value-bearing function call is the inconsistent survivor.  (`SELFDESTRUCT` force-sends remain possible independently of this issue: EIP-6780 keeps the balance-transfer effect, and `pragueTime: 0` is set from genesis, `chain-configs/mainnet/genesis.yaml:15-17`.)

## Impact

Low.  There is no attacker profit: the sender loses funds and no party can extract them.  The realistic harm is an integration footgun.  A wallet, contract, or the governance flow itself that attaches `msg.value` to a precompile call destroys those funds irrecoverably, where a Solidity contract with the same ABI would revert.  It also breaks expected function-mutability semantics: the module already encodes the `view` half of Solidity semantics as a first-class dispatcher concept, and omits the payability half.

## Suggested fix

Mirror the existing `read_only` classification (`mod.rs:157-161`) with a payability rule, next to the static gate:

- Reject `input.value != U256::ZERO` for every selector not explicitly named payable.
- The payable list is empty today, or contains only `burnCall` if governance intends to fund the burn pool by attaching value to `burn(amount)` calls.  The e2e governance forwarder already anticipates forwarding call value to the precompile (`crates/tn-reth/tests/it/pipeline_helpers.rs:62`), so this is a maintainer decision about the intended burn funding path.  If everything rejects value, the only remaining inlet to `0x7e1` is a `SELFDESTRUCT` force-send.

Either variant changes no existing behavior: no handler reads `input.value`, and no test sends nonzero value (the unit harness never sets `.value(..)` on `TxEnv`, `precompile_test_utils.rs:209-244`).  Both keep the fail-safe direction the static gate established: a selector added later rejects value unless it is explicitly named payable.

A regression test needs a value-bearing variant of the harness `exec`.  Assert the value-bearing call reverts and `0x7e1`'s balance is unchanged, and keep a zero-value positive control that succeeds.

## Environment

- main `648b6f81`
- `alloy-evm` 0.27.3, `revm` 34.0.0, `revm-handler` 15.0.0 (Cargo.lock)

## Credit

Reported by researcher shotes in the Cantina collaborative review, as a comment on the dispatcher doc block (`mod.rs:141`).

MavenRain:
Closes #1190.  (Cantina #21)

## Problem

The dispatcher (`telcoin_precompile`) enforces static-call write protection itself, because the precompile registration bypasses the interpreter's checks. The same reasoning applies to payability, and the dispatcher does not enforce it. No code in the module reads `PrecompileInput::value`. Every selector, including the view `totalSupply()`, accepts a nonzero call value.

The EVM credits the attached value to `0x7e1` before the precompile runs. `revm-handler` journals the caller-to-target transfer before it invokes the precompile, and it commits the transfer when the call succeeds. A successful `totalSupply()` call with attached value therefore both returns the supply and credits `0x7e1`. No selector moves native balance out of `0x7e1`, so the value is stranded. Only a future governance `burn` can destroy it, and no party can recover it. A Solidity contract with the same ABI would revert these calls through its compiled `CALLVALUE` guard.

There is no attacker profit: the sender loses the funds, and nobody can extract them. The harm is an integration footgun. A wallet, a contract, or the governance flow that attaches `msg.value` destroys those funds, where Solidity semantics promise a revert.

## Fix

- [x] `crates/tn-reth/src/evm/tel_precompile/mod.rs`: add a payability gate next to the static gate. The gate rejects nonzero `input.value` for every selector that is not named in a payable list. The rejection errors the frame, so the journal revert returns the value to the caller.
- [x] The payable list holds one selector: `burn(uint256)`. `handle_burn` destroys from the precompile's own balance. A plain value send reverts on the calldata-length check, and the genesis account starts with zero balance. Attached value on `burn(amount)` is therefore the sanctioned inlet for the burn pool: `msg.value == amount` funds and burns in one transaction, and excess value stays in the pool for later burns. The e2e governance forwarder (`pipeline_helpers.rs`) already anticipates this path.
- [x] The payable exemption also requires `input.target_address == TELCOIN_PRECOMPILE_ADDRESS`. Under `DELEGATECALL` or `CALLCODE` the `value` field carries the calling frame's apparent `msg.value` and nothing is transferred to the precompile. Apparent value never funds the pool, so it cannot authorize a draw from it, and such calls are rejected.
- [x] The classification is a payable list, not a reject list. This keeps the fail-safe direction the static gate established: a selector added later rejects value unless it is explicitly named payable, with no attacker required to hit the failure.
- [x] Doc comments: the module doc, the access-control list, and the dispatcher doc gain a payability section. The section records the journal mechanism and the `DELEGATECALL`/`CALLCODE` semantics.
- [x] `crates/tn-reth/src/evm/precompile_test_utils.rs`: add `exec_value_to` and `exec_with_value` to the shared harness. `exec_to` now delegates with `U256::ZERO`, which builds a byte-identical `TxEnv` (the builder default is already zero).

If governance prefers to keep every selector non-payable, the change is one line: replace the `matches!` arm with `false`. Then the only remaining inlet to `0x7e1` is a `SELFDESTRUCT` force-send.

Scope note: the gate is ungated by fork, the same way the static-call gate shipped. A value-bearing call to `0x7e1` that succeeded before this change now reverts. If sealed history on a live chain contains such a call, replay of that history diverges. No in-repo caller sends value to `0x7e1`, so the exposure is limited to third-party transactions already on chain; please confirm none exist or gate the check on a fork if they do.

## Testing

New regression tests in `tel_precompile/mod.rs`, all run with `cargo +1.94 test -p tn-reth --lib tel_precompile -j 2`:

- `test_read_only_selector_rejects_value`: `totalSupply()` with 1 wei attached reverts, and neither balance moves. The same call with zero value succeeds as the positive control.
- `test_mutating_selector_rejects_value`: `mint(uint256)` with 1 wei attached reverts. The same call with zero value succeeds.
- `test_burn_value_funds_pool_and_burns`: the pool starts empty, which mirrors the mainnet genesis alloc. `burn(500)` with zero value fails (negative control). `burn(500)` with 500 wei attached succeeds, the pool returns to zero, the caller pays exactly 500 wei, and `totalSupply` decreases by 500.
- `test_burn_excess_value_stays_in_pool`: `burn(200)` with 500 wei attached succeeds, leaves 300 wei in the pool for later burns, and decreases `totalSupply` by the burn amount (200), never by the attached value.
- `test_delegatecall_apparent_value_rejects_burn`: a failure-propagating `DELEGATECALL` relay forwards `burn(100)` while the transaction attaches 100 wei. The precompile sees apparent value with no transfer, the call reverts, and no balance moves. The zero-value delegatecall burn succeeds as the positive control.

Mutation confirmation, one in-diff mutant per guard clause: with the gate neutralized (`payable || true`) the two reject tests fail. With the payable list emptied (`let payable = false`) the two burn tests fail. With the `target_address` conjunct dropped, the delegatecall test fails. The restored file passes the full `tel_precompile` battery. Static ladder run locally (nightly `fmt --check`, scoped nightly `clippy -p tn-reth --all-features --all-targets --no-deps -D warnings`). CI and the pinned `+1.94` attest own the dynamic tier.

### DIFF
--- a/crates/tn-reth/src/evm/precompile_test_utils.rs
+++ b/crates/tn-reth/src/evm/precompile_test_utils.rs
@@ -221,6 +221,21 @@ impl TestEnv {
         calldata: Vec<u8>,
         gas_limit: u64,
     ) -> TestResult {
+        self.exec_value_to(caller, target, calldata, gas_limit, U256::ZERO)
+    }
+
+    /// Execute a transaction targeting `target` with the given gas limit and attached call value.
+    ///
+    /// Automatically increments the caller's nonce. Nonzero `value` exercises the precompile
+    /// dispatcher's payability gate.
+    pub fn exec_value_to(
+        &mut self,
+        caller: Address,
+        target: Address,
+        calldata: Vec<u8>,
+        gas_limit: u64,
+        value: U256,
+    ) -> TestResult {
         let nonce = self.nonces.entry(caller).or_insert(0);
         self.evm.ctx.set_tx(
             TxEnv::builder()
@@ -229,6 +244,7 @@ impl TestEnv {
                 .data(calldata.into())
                 .gas_limit(gas_limit)
                 .nonce(*nonce)
+                .value(value)
                 .build()
                 .unwrap(),
         );
@@ -236,6 +252,19 @@ impl TestEnv {
         // use mainnet handler with default gas to 0 for simpler test logic
         // pipeline tests use TN tools and account for gas usage
         MainnetHandler::default().run(&mut self.evm)
+    }
+
+    /// Execute a precompile call with the given gas limit and attached call value.
+    ///
+    /// [`exec`](Self::exec) with `value` wei attached; targets [`TELCOIN_PRECOMPILE_ADDRESS`].
+    pub fn exec_with_value(
+        &mut self,
+        caller: Address,
+        calldata: Vec<u8>,
+        gas_limit: u64,
+        value: U256,
+    ) -> TestResult {
+        self.exec_value_to(caller, TELCOIN_PRECOMPILE_ADDRESS, calldata, gas_limit, value)
     }
 
     /// Execute a precompile call with the default gas limit (100,000).
--- a/crates/tn-reth/src/evm/tel_precompile/mod.rs
+++ b/crates/tn-reth/src/evm/tel_precompile/mod.rs
@@ -11,7 +11,10 @@
 //! intercepted and routed to [`telcoin_precompile`] instead of executing bytecode. Bypassing
 //! bytecode execution also bypasses the interpreter's static-call write protection, so
 //! [`telcoin_precompile`] enforces that itself: inside a `STATICCALL` frame it serves the
-//! read-only selectors and refuses every state-mutating one.
+//! read-only selectors and refuses every state-mutating one. Payability follows the same
+//! reasoning: the compiled `CALLVALUE` guard a Solidity dispatcher would run never executes
+//! here, so [`telcoin_precompile`] rejects nonzero call value on every selector except the
+//! explicitly payable `burn(uint256)`.
 //!
 //! # Module structure
 //!
@@ -42,6 +45,10 @@
 //! - **Any account**: can call `totalSupply()` (read-only).
 //! - **Static frames**: regardless of caller, only the read-only selectors (`totalSupply()`, plus
 //!   `hasMintRole` with the faucet feature) are served inside a `STATICCALL`.
+//! - **Call value**: only `burn(uint256)` accepts nonzero value, and only when the precompile is
+//!   the actual transfer target (a plain `CALL`): the attached wei tops up the pool that `burn`
+//!   destroys from, and any excess stays in the pool for later burns. Every other value-bearing
+//!   call is rejected, so `burn` is the single value inlet at [`TELCOIN_PRECOMPILE_ADDRESS`].
 //!
 //! # Feature flags
 //!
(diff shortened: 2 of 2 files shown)


sstanfield (approved):

grantkee (approved):
LGTM

PR #1202, opened 2026-08-17
fix(network): gate kad provider capacity check on new keys only

### LINKED ISSUE #1195: network-libp2p: kad add_provider rejects refresh/self-provide once max_provided_keys is reached
## Summary

The kad `add_provider` path runs the `max_provided_keys` capacity check before it determines whether
the provider key already exists. The sibling `put` path was fixed so its capacity check runs only when
the key is new. Because the fix was not mirrored, a saturated provider table rejects every later
`add_provider`, including a legitimate refresh of an existing key and the node's own startup
self-provide.

This is an availability griefing vector. It is not a chain-halt.

- Correct gating (`put`):
  `crates/network-libp2p/src/kad.rs:461-468`. The capacity check is inside `if new_record { ... }`.
  `new_record` is computed at `kad.rs:451-458`. An update of an existing key is allowed even at
  capacity.
- Missing gating (`add_provider`):
  `crates/network-libp2p/src/kad.rs:513-520`. The check
  `if self.num_providers >= self.config.max_provided_keys { self.evict_expired_providers(); }`
  followed by
  `(self.num_providers < self.config.max_provided_keys).then_some(()).ok_or(Error::MaxProvidedKeys)?;`
  runs unconditionally. The key lookup that computes `row_exists` happens later, at `kad.rs:522-535`.

## Root cause

The capacity-check ordering was corrected for `put` and was not mirrored onto `add_provider`. The
provider path checks capacity before it knows whether the key is new, so it cannot distinguish a new
insertion from a refresh of an existing key.

## Threat model

Byzantine peers that fill the provider table with distinct keys within the record TTL. No validator key
is needed.

## Impact

Once byzantine peers saturate the `max_provided_keys` table (for example 1024 keys), every later
`add_provider` is rejected with `MaxProvidedKeys`. This includes:

- A legitimate refresh of a key the node already provides. The refresh is needed to keep the record
  live, so the node's own valid provider records can expire.
- The node's own startup self-provide, which has no retry. The node then does not announce itself as a
  provider for its intended keys.

Suggested severity: Medium. Availability griefing of an honest node's own provider records.

## Reproduction (sketch)

1. Fill the provider table to `max_provided_keys` with distinct keys from byzantine peers.
2. Trigger a legitimate refresh (or restart the honest node so it runs its self-provide).
3. Observe the refresh or self-provide rejected with `MaxProvidedKeys`, even though the key already
   exists or belongs to the node.

## Suggested fix

Gate the `max_provided_keys` check on whether the provider key is new, the same way `put` gates its
`max_records` check. Compute the key lookup first, then apply the capacity check only when the key is
new. This allows a refresh or self-provide of an existing key at capacity.

## Scope

The capacity cap itself is correct and needed. The only defect is the order of the check relative to
the new-versus-existing decision, which `put` already handles correctly.

MavenRain:
Fixes #1195.

## Problem

`add_provider` runs the `max_provided_keys` capacity check before it determines whether the provider key already exists. The sibling `put` path gates its `max_records` check on `new_record` (an update of an existing key is allowed at capacity), but the ordering was not mirrored onto the provider path.

Byzantine peers that fill the provider table with distinct keys within the record TTL need no validator key. Once the table holds `max_provided_keys` entries (1024 by default), every later `add_provider` returns `Err(MaxProvidedKeys)`, including:

- A legitimate refresh of a key the node already provides. The refresh keeps the record live, so the node's own valid provider records can expire.
- The node's own startup self-provide, which has no retry. The node then does not announce itself as a provider for its intended keys.

This is availability griefing of an honest node's own provider records, not a chain halt.

## Root cause

The capacity-check ordering was corrected for `put` and not mirrored onto `add_provider`. The provider path checks capacity before it knows whether the key is new, so it cannot tell a new insertion from a refresh of an existing key.

## Fix

- [x] `crates/network-libp2p/src/kad.rs`: `add_provider` computes the key lookup (`row_exists`) first, then applies the expired-group eviction attempt and the capacity check only when the key is new, inside `if !row_exists { ... }`. This mirrors `put`'s `if new_record { ... }` gate exactly.

The cap itself is unchanged and still binds: an overwrite of an existing, already-counted row cannot grow the table, and a brand-new key at capacity is still rejected after the eviction attempt. The check now gates the same event on both store paths (`put` for records, `add_provider` for provider keys), so the two capacity bounds behave symmetrically.

## Testing

- [x] `test_kad_add_provider_refresh_allowed_at_capacity`: fills the primary and worker tables to a small cap, refreshes an existing key 10 times at capacity (must succeed, count unchanged), then confirms a brand-new key is still rejected with `MaxProvidedKeys` (positive control).
- [x] `test_kad_add_provider_evicts_expired_when_full`: saturates the table with expired provider groups and confirms a brand-new live key still succeeds through the eviction attempt, mirroring `test_kad_put_evicts_expired_when_full`.

Commands run under the pinned toolchain (`cargo +1.94 test -p tn-network-libp2p -- test_kad`, shared target, `-j 2`): 15 passed, 0 failed, including the 13 pre-existing kad store tests.

Each new test was confirmed by mutation against the staged fix:

- Gate reverted to unconditional (`if !row_exists` replaced by `if true`, the pre-fix ordering): `test_kad_add_provider_refresh_allowed_at_capacity` fails (2 passed, 1 failed).
- Capacity check deleted (`if false`): both new tests fail, the refresh test at its new-key rejection control (1 passed, 2 failed).
- Eviction call deleted: `test_kad_add_provider_evicts_expired_when_full` fails (2 passed, 1 failed).

The source was byte-restored after each mutant (verified identical to the staged copy) and the full kad filter re-ran green (15 passed, 0 failed).

Static ladder also green: nightly `fmt --check` clean; in-crate `clippy --no-deps --all-features --all-targets` clean; ctxcat-review over the staged diff returned no findings; an independent logic review confirmed the counter invariant on every provider path and verified against the lockfile-pinned `libp2p-kad 0.48.0` that upstream `MemoryStore::add_provider` applies `max_provided_keys` only to a vacant key, so this change restores parity with the upstream store contract as well as with `put`.

### DIFF
--- a/crates/network-libp2p/src/kad.rs
+++ b/crates/network-libp2p/src/kad.rs
@@ -510,15 +510,6 @@ impl<DB: Database> RecordStore for KadStore<DB> {
     }
 
     fn add_provider(&mut self, record: ProviderRecord) -> libp2p::kad::store::Result<()> {
-        if self.num_providers >= self.config.max_provided_keys {
-            // Try to free a slot by evicting fully-expired provider key groups.
-            self.evict_expired_providers();
-        }
-        // Re-check after the eviction attempt; still full is a hard error.
-        (self.num_providers < self.config.max_provided_keys)
-            .then_some(())
-            .ok_or(Error::MaxProvidedKeys)?;
-
         let key = self.key_to_hash(&record.key);
         let kr: KadProviderRecord = record.into();
         let stored = match self.kad_type {
@@ -533,6 +524,23 @@ impl<DB: Database> RecordStore for KadStore<DB> {
         // genuinely new key it is already counted in `num_providers`, so only a missing
         // row bumps the gauge (issue #999).
         let row_exists = stored.is_some();
+
+        // The capacity check applies only to a brand-new key, mirroring `put`'s
+        // `new_record` gate: an overwrite of an existing, already-counted row cannot
+        // grow the table, so a refresh of a key this node already stores (including
+        // its own self-provide) must succeed even when byzantine peers have
+        // saturated the table with other keys (issue #1195).
+        if !row_exists {
+            if self.num_providers >= self.config.max_provided_keys {
+                // Try to free a slot by evicting fully-expired provider key groups.
+                self.evict_expired_providers();
+            }
+            // Re-check after the eviction attempt; still full is a hard error.
+            (self.num_providers < self.config.max_provided_keys)
+                .then_some(())
+                .ok_or(Error::MaxProvidedKeys)?;
+        }
+
         let merged = stored
             .as_deref()
             .and_then(|raw| decode_providers(&key, raw))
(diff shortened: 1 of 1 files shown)


sstanfield (approved):

grantkee (approved):
Good catch

PR #1203, opened 2026-08-17
fix(node): re-vote stored-but-uncertified epoch record on startup

### LINKED ISSUE #1198: Permanent epoch-record certificate gap after fleet-wide vote-quorum failure (deferred from #1196)
## Summary

After #1196, a fleet-wide vote-quorum failure at an epoch close leaves a permanent certificate gap for that epoch's record. No mechanism at head can manufacture the missing certificate afterward. The fleet keeps producing blocks (the goal of #1196, achieved), but new and behind nodes can never sync past that epoch.

This is a tracking issue. Steve called re-vote out of scope for #1196 ("decoupling EpochRecord logic from the network") and pointed to a future network refactor PR, with manual old-committee intervention in extreme cases. That scope call is reasonable. The gap itself is real and needs an owner.

## Mechanism

1. Each node persists its epoch record locally, then votes on it. The vote collector (`crates/node/src/manager/epoch_votes.rs:57-266`) is a one-shot, bounded collector driven by `epoch_record_watch`.
2. `epoch_record_watch` is in-memory only. Nothing re-fires it from durable storage for a record that is already persisted. A node that persisted its record but lost the vote race (for example, all nodes tore down within gossip latency at the epoch close, as in the epoch-383 incident shape) never votes again after restart.
3. If no quorum of votes ever aggregates, no certificate for record N exists anywhere.
4. Peers serve only certified records (`crates/consensus/primary/src/network/handler.rs:1166`). The state-sync collector fetches only certified pairs and retries every 5 seconds forever (`crates/state-sync/src/epoch.rs:150-190`).
5. Result: onboarding and catch-up stall permanently at epoch N. The operator signal is warn/info noise only.

## What #1196 already mitigates

The `recover_previous_epoch_record` path added in #1196 re-derives a missing record N-1 and re-fires `epoch_record_watch`, so the vote collector re-votes it. This covers the crash window where the record was never written. It does not cover the persisted-but-uncertified state: recovery fires only when the record is missing, so a node that has its record on disk but no certificate stays silent.

## Suggested direction

Re-vote from durable storage on restart: when the current committee's record exists locally but no certificate for it exists, re-fire the vote path. #1193's Bug-D re-vote is the existing companion change for this. An equivalent certificate backfill in the network refactor would also close it.

## References

- #1196 (where the trade was made: loud fleet halt became quiet permanent degradation)
- #1193 (Bug-D re-vote on restart, the companion fix)
- Review notes: v1 finding 3 and v2 disposition at head ace934b6

MavenRain:
Closes #1198.

## Problem

The vote collector is armed only by `epoch_record_watch` (`crates/node/src/manager/epoch_votes.rs`). Only the epoch-close paths write that channel, and the channel is in-memory. A node that persists its epoch record and then shuts down before a vote quorum aggregates never fires the channel again. After a restart the node holds a record with no certificate, and it never re-votes. No mechanism at head can manufacture the missing certificate afterward.

The failure needs no attacker. The epoch-383 incident shape produces it: every node tears down within gossip latency at an epoch close, each node persists its record first, and no vote quorum ever forms. Peers serve only certified records (`primary/src/network/handler.rs`), and the state-sync collector fetches only certified pairs and retries forever (`state-sync/src/epoch.rs`), so onboarding and catch-up stall permanently at that epoch.

The anchor wait already assumes this cannot happen: `CERTIFIED_ANCHOR_WAIT` is sized on the basis that "the vote collector and this wait start from nearly the same instant" (`run_epoch.rs`). On the restart path that assumption is false today. The wait starts, but nothing arms the collector.

## Fix

Re-vote from durable storage on restart.

- [x] `crates/node/src/manager/epoch_votes.rs`: new `revote_uncertified_epoch_record_on_startup(db, watch)`. It reads the latest stored epoch record. When the store holds no certificate for that record, it re-publishes the record on `epoch_record_watch`. That re-arms the existing collector round: self-sign when this node is in the record's committee, collect peer votes, then the bounded peer-fetch fallback.
- [x] `crates/node/src/manager/node.rs`: call the hook in `launch()` immediately after `spawn_epoch_vote_collector`. The collector subscribes to the watch inside the spawn call, and a value sent before that subscription is treated as already seen, so the ordering is load-bearing and documented at both sites.

Threat model and bounds:

- The re-publish is idempotent. Votes re-sign deterministically (BLS over the record digest), the vote handler verifies and deduplicates them, and certificate writes are append-once per digest.
- The added work is bounded and attacker-independent: at most one re-armed round per process start, and `manage_epoch_votes` already bounds every round by `MAX_EPOCH_VOTE_TIMEOUTS` vote windows plus `EPOCH_CERT_RECOVERY_ATTEMPTS` peer fetches.
- Committee authority is unchanged. A node outside the record's committee signs nothing; its re-armed round only listens for votes and then tries the peer fetch.
- Epoch 0 is re-armed like any other epoch. The uncertifiable epoch-0 dummy record is not observable by this hook: `run_epochs` seeds it in-memory after the hook runs and `save_dummy_epoch0` never persists it, so a stored epoch-0 record at startup is a genuine closed-epoch-0 record. Without this, a fresh network that closes epoch 0 and restarts before a vote quorum aggregates would stall at its very first boundary, where no peer can supply the missing cert either.
- A certified latest record and an empty store both change nothing (both negative paths are tested).

Known residual scope, deliberately left to the network refactor per the issue: the hook heals the latest stored record only, once per process start. A historical gap behind an already certified later record needs the refactor's certificate backfill, and a bounded round that expires before enough peers are back up is retried only by the next restart. Both are stated in the function docs.

Scope note: this covers the persisted-but-uncertified state that #1196's `recover_previous_epoch_record` does not reach (that recovery fires only when the record is missing). #1193 carries the equivalent re-vote on the adiri hotfix line (its Bug-D). The full re-vote/backfill story stays with the planned network refactor; this hook is the minimal fix at head.

## Testing

- [x] 5 new deterministic tests in `epoch_votes.rs`:
  - `test_startup_revote_fires_for_uncertified_record` (positive: watch re-armed with the stored record)
  - `test_startup_revote_skips_certified_record`
  - `test_startup_revote_fires_for_uncertified_epoch_zero_record` (genuine epoch-0 record heals; the dummy is unreachable at startup)
  - `test_startup_revote_skips_empty_store`
  - `test_startup_revote_reaches_quorum_and_stores_cert` (end to end: the startup hook alone, with no manual watch write, drives the collector to re-sign, aggregate peer votes, and store the certificate; convergence waits use `wait_until`, no fixed sleeps)
- [x] Command: `cargo test -p tn-node epoch_vote_collector_tests`
- [x] Every new load-bearing test confirmed by in-diff mutation: three mutants (watch re-arm deleted, store read replaced with `None`, missing-cert guard inverted), each killed by the predicted tests, original restored and verified byte-identical to the staged bytes.
- [x] Static ladder: nightly `fmt -- --check` green; scoped clippy on the buildplan-emitted 14-crate `-p` list with `--all-features --no-deps -- -D warnings` green.
- Dynamic tier: GitHub CI fires on push; the pinned `+1.94` attest runs on the second machine.

### DIFF
--- a/crates/node/src/manager/epoch_votes.rs
+++ b/crates/node/src/manager/epoch_votes.rs
@@ -12,7 +12,10 @@ use tn_types::{
     BlsAggregateSignature, BlsPublicKey, BlsSignature, Epoch, EpochCertificate, EpochDigest,
     EpochRecord, EpochVote, Noticer, TaskSpawner, TnReceiver as _,
 };
-use tokio::sync::mpsc::{self, Receiver, Sender};
+use tokio::sync::{
+    mpsc::{self, Receiver, Sender},
+    watch,
+};
 use tracing::{error, info, warn};
 
 type VoteQueue = VecDeque<(Epoch, Sender<EpochVote>, Option<Receiver<EpochVote>>)>;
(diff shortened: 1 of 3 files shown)


sstanfield (approved):
Looks good.  I was thinking we could just do this on epoch start as a fallback (vs just process start).  This should fix the observed problem though and as noted in the comment the other case should be handled by normal networking (i.e. make sure we don't orphan the legit votes somehow).

grantkee (approved):
LGTM - this closes the remaining issue for network-wide cold restart. I will follow up with some specific issues for hardening this since network-wide cold restart for mainnet is not an option. Good work @MavenRain

PR #1207, opened 2026-08-17
Don't verify the parent hash for consensus block 1- no need and its subject to change.

sstanfield:
Specifically keeping this check will require a bunch of fork code.  It does not matter since an invalid hash here will cause overall verification to fail so we can safely skip it to keep the verify code simpler in this on tiny edge case.

### DIFF
--- a/crates/storage/src/consensus_pack.rs
+++ b/crates/storage/src/consensus_pack.rs
@@ -951,8 +951,9 @@ impl Inner {
         .map_err(OpenError::IndexFileOpen)?;
         let mut parent_digest_expectation = if epoch == 0 {
             // Don't worry about consensus block 1 in epoch 0, if it is invalid other verifications
-            // will fail. This can be set but doing so leaves potential footguns around
-            // and this check is not really useful in this case.
+            // will fail (for instance epoch 0 final state will not verify). This can be set but
+            // doing so forces fork aware code here and verification will fail with an invalid value
+            // either way.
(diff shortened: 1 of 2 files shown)


MavenRain:
Nits, none blocking:

1. `pack_validate.rs:227-234`: the same edit lands in the offline validator, but `db validate` (telcoin-network-cli/src/db.rs:116) has no terminal digest gate, so "other verifications will fail" is not true there. A corrupted epoch-0 block-1 parent now reports `VALID`. Suggest keeping a best-effort `Some(ConsensusHeader::default().digest())` in this file (commented as build-flavor specific), or adding an informational note to the report.
2. `consensus_pack.rs:963-966` loop comment still says the link "is verified INSIDE the decoder via `HeaderExpectation::Parent`" and refers to `parent_digest`.
3. `pack_validate.rs:179-182` rustdoc says `previous` anchors the first header; the epoch-0 branch ignores `previous`.
4. The new comments say "footguns" without the reason. Suggest naming it: `ConsensusHeader::default().digest()` is build-flavor and fork dependent (types/primary/block.rs:207-235), and the write side still mints it at genesis (state-sync/src/lib.rs:156-166), so anchoring here needs fork-aware code and the terminal `final_consensus` check is the backstop.

MavenRain (dismissed):

MavenRain (approved):

grantkee (approved):

PR #1208, opened 2026-08-17
fix(tn-reth): pin the blob gas price in evm_env to match block execution

### LINKED ISSUE #1205: tn-reth: evm_env derives blob_gasprice from the header (= 1) while block execution pins it to u128::MAX, so BLOBBASEFEE disagrees between RPC simulation and execution
## Summary

`TnEvmConfig` builds the EVM block environment in two places and they disagree on the EIP-4844 blob gas price:

- `next_evm_env` (used for block building and execution) hard-codes `BlobExcessGasAndPrice { excess_blob_gas: 0, blob_gasprice: u128::MAX }` with the comment "eip4844 transactions are ignored" (`crates/tn-reth/src/evm/config.rs:178-181`).
- `evm_env(header)` (used for RPC reads and simulation) derives the price from the header: `header.excess_blob_gas.zip(chain_spec.blob_params_at_timestamp(ts)).map(|(g, p)| p.calc_blob_fee(g))` (`crates/tn-reth/src/evm/config.rs:124-130`).

Every TN header carries `excess_blob_gas: Some(0)` (`crates/tn-reth/src/evm/block.rs:1487`; the reth genesis header does the same when Cancun is active at genesis) and TN activates Cancun and Prague at timestamp 0 (`chain-configs/mainnet/genesis.yaml:16-17`, `crates/types/src/genesis.rs:72-74`), so `blob_params_at_timestamp` always returns the Prague params and `calc_blob_fee(0)` evaluates to `BLOB_TX_MIN_BLOB_GASPRICE = 1` (alloy-eips 1.8.3, `eip4844/mod.rs:84`, `eip7840.rs:156`).

Result: the `BLOBBASEFEE` opcode (`block.blobbasefee` in Solidity) returns `1` inside `eth_call`, `eth_estimateGas`, `eth_createAccessList`, `debug_trace*` and `trace_*`, but returns `340282366920938463463374607431768211455` (`u128::MAX`) inside real block execution.  A contract that branches on `block.blobbasefee` can pass simulation and behave differently on chain.

Severity: informational / low.  Not consensus-critical (see below).  Raised by Haxatron in the Cantina collaborative review; this issue records the verification and a fix shape.

## Where the two values are observable

| Surface | Env constructor | `blob_gasprice` seen by `BLOBBASEFEE` |
| --- | --- | --- |
| Block building and execution (`RethEnv::build_block_from_batch_payload` -> `builder_for_next_block`, `crates/tn-reth/src/env/execution.rs:110`) | `next_evm_env` | `u128::MAX` |
| `eth_call` / `eth_estimateGas` / `eth_createAccessList` at `latest` or a historical block (reth `evm_env_at` -> `evm_env_for_header`, `crates/rpc/rpc-eth-api/src/helpers/state.rs:277-296`) | `evm_env(header)` | `1` |
| Same calls at `pending`: TN sets a real pending block on every executed block (`crates/engine/src/payload_builder.rs:310`), so reth takes the `ActualPending` branch and calls `evm_env(block.header())` (`helpers/pending_block.rs:76`) | `evm_env(header)` | `1` |
| `debug_traceCall` / `debug_traceTransaction` / `debug_traceBlock`, `trace_*` (all go through `evm_env_at`) | `evm_env(header)` | `1` |
| `eth_simulateV1` (reth calls `next_evm_env` for every simulated block, `helpers/call.rs:130-132`) | `next_evm_env` | `u128::MAX` |
| `eth_blobBaseFee` (reth computes it from `header.excess_blob_gas` independently, `helpers/fee.rs:374-386`) | n/a | `1` |

So today three answers exist for "what is the blob base fee on TN": `1` (most RPC reads and `eth_blobBaseFee`), `u128::MAX` (execution and `eth_simulateV1`).

Both values are pinned by reth v1.11.3, revm 34.0.0 and alloy-eips 1.8.3 from `Cargo.lock`.  `calc_blob_fee(0) = 1` was executed against alloy-eips 1.8.3 in a standalone probe (Cancun, Prague and Osaka params all give `1`).

## Why this is not consensus-critical

- All state-changing execution goes through `next_evm_env` (`builder_for_next_block`), including replay and catch-up through the payload builder.  Batch validation does not execute the EVM at all (`crates/batch-validator/src/validator.rs:39`).
- `evm_env(header)` is reached from the engine and node manager only for read-only registry queries: `epoch_state_at_header` (`crates/tn-reth/src/env/epoch.rs:233`, called from `crates/node/src/manager/node.rs:214`), `worker_fee_configs_inner` (`epoch.rs:594`, called from `run_epoch.rs:1072`) and `classified_registry_batch_at_header` (`epoch.rs:736`, called from `crates/node/src/engine/inner.rs:405-439`).  These are `eth_call`-style view reads against `ConsensusRegistry` and `WorkerConfigs`.  Neither contract, nor any other TN-owned contract under `tn-contracts/src`, reads `block.blobbasefee`, so the value is inert on those paths.  If a future contract upgrade reads it, the two constructors must already agree.
- revm only uses `blob_gasprice` for two things: the `BLOBBASEFEE` opcode (`revm-interpreter-32.0.0/src/instructions/block_info.rs:79-85`) and, for `TransactionType::Eip4844` only, the fee-cap check (`revm-handler-15.0.0/src/validation.rs:86-92,177-189`) and the blob-fee balance deduction (`revm-context-interface-14.0.0/src/transaction.rs:217-227`).  TN never admits blob transactions to a block: the batch builder skips them (`crates/batch-builder/src/batch.rs:94,135`), the batch validator rejects them (`crates/batch-validator/src/validator.rs:213`), and #1169 rejects them at pool admission.  So the price only reaches a non-blob transaction through `BLOBBASEFEE`.

## Impact

- A contract that reads `block.blobbasefee` (fee estimators, L2-style DA cost helpers, any library that guards on it) gets `1` in `eth_estimateGas` / `eth_call` and `u128::MAX` on chain.  Gas estimates and revert predictions can be wrong for such contracts.
- A hand-built type-3 `TransactionRequest` passes the `BlobGasPriceGreaterThanMax` check in `eth_call` / `eth_estimateGas` at `latest` and `pending` (price `1`) but fails it in `eth_simulateV1` and in execution (price `u128::MAX`).
- Two RPC surfaces (`eth_simulateV1` vs everything else) disagree with each other.

No fund is at risk and no attacker profit path exists.

## Suggested fix

Keep the execution semantics (`u128::MAX`) and make `evm_env` agree.  Do not do the reverse: deriving the price from the header in `next_evm_env` would change what `BLOBBASEFEE` returns on chain, which is an execution-semantics change.

1. Extract one private helper that both constructors call, so the two cannot drift again:

```rust
/// Blob gas is not priced on TN: EIP-4844 transactions never reach a block, so
/// `BLOBBASEFEE` reports `u128::MAX` on every path (execution, replay, RPC).
fn tn_blob_excess_gas_and_price() -> BlobExcessGasAndPrice {
    BlobExcessGasAndPrice { excess_blob_gas: 0, blob_gasprice: u128::MAX }
}
```

In `evm_env`, keep the existing `Option` shape (only `Some` when the header carries `excess_blob_gas` and blob params exist at that timestamp) but replace the computed value with the helper's, so the pre-Cancun `None` branch is unchanged:

```rust
let blob_excess_gas_and_price = header
    .excess_blob_gas
    .zip(self.chain_spec().blob_params_at_timestamp(header.timestamp))
    .map(|_| tn_blob_excess_gas_and_price());
```

In `next_evm_env`, use `Some(tn_blob_excess_gas_and_price())`.

2. Add a parity test in `crates/tn-reth/src/evm/config.rs` (the module currently has one test):

```rust
/// `evm_env` (RPC and replay reads) and `next_evm_env` (block execution) must report the
/// same blob gas price, or `BLOBBASEFEE` differs between simulation and execution.
#[test]
fn test_evm_env_blob_gasprice_matches_block_execution() {
    let chain: ChainSpec = tn_types::test_genesis().into();
    let config = TnEvmConfig::new(Arc::new(chain), GasAccumulator::default());
    // every TN header carries `excess_blob_gas: Some(0)` (see `TNBlockAssembler`)
    let header = ExecHeader { excess_blob_gas: Some(0), ..Default::default() };
    let env = config.evm_env(&header).expect("evm env");
    assert_eq!(
        env.block_env.blob_excess_gas_and_price,
        Some(BlobExcessGasAndPrice { excess_blob_gas: 0, blob_gasprice: u128::MAX }),
    );
}
```

On `main` (648b6f81) this assertion fails with `blob_gasprice: 1`.  (Not executed in this report; the `1` follows from the pinned alloy-eips arithmetic above.)

3. Note in the fix PR that RPC behavior changes on purpose: hand-built type-3 requests to `eth_call` / `eth_estimateGas` / `debug_traceCall` will now fail `BlobGasPriceGreaterThanMax` unless the caller sets `maxFeePerBlobGas` to `u128::MAX`, matching execution.

Optional follow-up: `eth_blobBaseFee` will still return `1` because reth computes it from the header, not from `evm_env`.  Overriding it to `u128::MAX` (or documenting that TN has no blob market) is a separate, cosmetic decision.

## History and duplicates

- Both constructors landed together when `crates/tn-reth/src/evm/config.rs` was created (commit 433fe723, #889); no later commit touched the blob lines.  This is not a regression.
- Upstream reth `EthEvmConfig::evm_env` uses the same header-derived computation;  TN's `u128::MAX` in `next_evm_env` is a deliberate TN deviation, so `evm_env` is the copy that was not updated to match.
- No open or closed issue or PR covers this.  Related but distinct: #1159 (blob-sidecar disk growth) and #1169 (reject blob and set-code txs at pool admission).

## Credit

Reported by Haxatron in the Cantina collaborative review (comment on `crates/tn-reth/src/evm/config.rs:128`).

MavenRain:
Closes #1205.  (Cantina #8)

## Problem

`TnEvmConfig` builds the EVM block environment in two places, and they disagreed on the EIP-4844 blob gas price:

- `next_evm_env` (block building and execution, `eth_simulateV1`) pins `BlobExcessGasAndPrice { excess_blob_gas: 0, blob_gasprice: u128::MAX }`.
- `evm_env(header)` (every header-derived environment: `eth_call`, `eth_estimateGas`, `eth_createAccessList`, `debug_trace*`, `trace_*` at `latest`, historical and `pending`; the replay-side read-only registry queries in `env/epoch.rs`; and reth's block re-execution through `executor_for_block`, e.g. `debug_executionWitness` and the test executor) derived the price from the header: `blob_params.calc_blob_fee(header.excess_blob_gas)`.

Every TN header carries `excess_blob_gas: Some(0)` and Cancun/Prague are active at timestamp 0, so the derived value is always `calc_blob_fee(0) = 1`. `BLOBBASEFEE` (`block.blobbasefee`) therefore returned `1` in simulation and re-execution but `u128::MAX` on chain. A contract that branches on it can pass `eth_call` / `eth_estimateGas` and behave differently in the block, a block that runs such a contract re-executes to a different state root than the one the builder produced, and a hand-built type-3 request passed the `BlobGasPriceGreaterThanMax` check in `eth_call` but failed it in execution and `eth_simulateV1`.

Not consensus-critical: every canonical execution, including replay and catch-up, goes through the payload builder and therefore `next_evm_env`; the `evm_env` re-execution path is ad hoc (`debug_executionWitness`, tests). None of the TN-owned contracts (`ConsensusRegistry`, `WorkerConfigs`) that the replay-side `evm_env` reads touch reads `block.blobbasefee`. No fund is at risk. Reported by Haxatron in the Cantina collaborative review.

## Root cause

Both constructors landed together in #889. `next_evm_env` carries the deliberate TN deviation (blob gas is not priced because blob transactions never reach a block), while `evm_env` kept upstream reth's header-derived computation. Two copies of one policy, only one of them updated.

## Fix

Keep the execution semantics (`u128::MAX`) and make `evm_env` agree, through one shared constant so the two cannot drift again:

- `TN_BLOB_EXCESS_GAS_AND_PRICE` (private `const` in `crates/tn-reth/src/evm/config.rs`, documented) holds `{ excess_blob_gas: 0, blob_gasprice: u128::MAX }`.
- `next_evm_env` uses `Some(TN_BLOB_EXCESS_GAS_AND_PRICE)` (same value as before, no behavior change on chain).
- `evm_env` uses `header.excess_blob_gas.zip(blob_params).map(|_| TN_BLOB_EXCESS_GAS_AND_PRICE)`: upstream's `Option` shape is preserved (`None` only for a header without `excess_blob_gas` or a timestamp without blob params, which TN never produces), only the value changes. `blob_params` is the value already computed for `max_blobs_per_tx`, so the fork lookup is not repeated.

The reverse (deriving the price from the header in `next_evm_env`) would change what `BLOBBASEFEE` returns on chain, i.e. an execution-semantics change, so it is not done.

Intended RPC behavior change: `BLOBBASEFEE` now reads `u128::MAX` in `eth_call` / `eth_estimateGas` / `eth_createAccessList` / `debug_trace*` / `trace_*`, matching execution and `eth_simulateV1`. A hand-built type-3 request to those endpoints now fails `BlobGasPriceGreaterThanMax` unless it sets `maxFeePerBlobGas` to `u128::MAX`, exactly as it already did in execution. Non-blob transactions (types 0, 1, 2, 4) are unaffected: revm reads `blob_gasprice` only for `BLOBBASEFEE` and for the EIP-4844 fee-cap check and blob-fee deduction, which are gated on the transaction type.

revm's only other reader of `blob_gasprice` for a non-blob transaction is `calculate_caller_fee`, whose blob term is gated on `TransactionType::Eip4844` and uses saturating / checked arithmetic, so `u128::MAX` cannot overflow or panic there. reth's default local pending-block builder does narrow the block env's `blob_gasprice` to `u64`, but TN configures `PendingBlockKind::None` and serves a real pending block (`env/rpc.rs`), so that path is not reached (and it was already fed `next_evm_env`'s `u128::MAX`).

Out of scope (separate, cosmetic decision): `eth_blobBaseFee` still returns `1` because reth computes it from the header, not from the EVM env. The pool's `pending_block_blob_fee = Some(u128::MAX)` (`txn_pool.rs`, batch builder) already agrees with the pinned value and is untouched. Pool admission itself is not gated on blob transactions until #1169 lands; today the pool prices them out of the pending set and the batch builder / validator keep them out of blocks, which is what the new constant's doc comment says.

## Changes

- [x] `crates/tn-reth/src/evm/config.rs`: new documented `TN_BLOB_EXCESS_GAS_AND_PRICE` constant; `evm_env` and `next_evm_env` both use it; module doc gains a "blob gas is not priced on TN" paragraph.
- [x] `crates/tn-reth/src/evm/config.rs` tests (3 new):
  - `test_evm_env_blob_gasprice_matches_next_evm_env` builds the execution env from the genesis header plus `TNPayload::build_pending_env` and asserts `evm_env(genesis)` carries the same `blob_excess_gas_and_price`, and that it is the pinned `{ 0, u128::MAX }` (fails on `main` with `blob_gasprice: 1`).
  - `test_blobbasefee_opcode_reads_pinned_price_under_both_envs` is the observable form: it runs `BLOBBASEFEE PUSH0 MSTORE PUSH1 32 PUSH0 RETURN` through the real TN EVM (`TNEvmFactory::create_evm` on an `InMemoryDB`, plain non-blob call) under the `evm_env` environment and under the `next_evm_env` environment and asserts both return `u128::MAX` (on `main` the `evm_env` run returns `1`).
  - `test_evm_env_blob_gasprice_none_without_header_excess_blob_gas` pins the preserved `Option` shape.
- [x] `crates/tn-reth/README.md`: the "Fork and EVM configuration" bullet names the shared constant and states that `BLOBBASEFEE` agrees between simulation, re-execution and execution.

## Testing

Local checks under the pinned toolchains:

- `cargo +nightly clippy -p tn-reth --all-features --all-targets --no-deps -j 2 -- -D warnings`: clean.
- `cargo +1.94 test -p tn-reth --lib evm::config::tests -j 2`: 4 passed (3 new + the existing `extra_data` test).
- Mutation confirmation of the new tests (mutate, run, byte-restore): restoring the header-derived price in `evm_env` fails the parity test and the opcode test; pricing the shared constant at `1` fails both as well; making `evm_env` unconditionally `Some` fails the `Option`-shape test.

Dynamic tier: GitHub CI (`nextest --workspace`, workspace clippy, adiri lane) runs on push; `make attest` (pinned 1.94, workspace test build) is run separately.

### DIFF
--- a/crates/tn-reth/src/evm/config.rs
+++ b/crates/tn-reth/src/evm/config.rs
@@ -34,6 +34,12 @@
 //! an epoch-closing block there would price every EIP-1559 worker from one empty slot (zero gas,
 //! `MIN_PROTOCOL_BASE_FEE`) and produce a divergent state root, not merely degraded reward
 //! accounting.
+//!
+//! Blob gas is not priced on TN. Both block-environment constructors (`evm_env` for every
+//! header-derived environment: `eth_call`, tracing, the replay-side registry queries and reth's
+//! block re-execution through `executor_for_block`; `next_evm_env` for block building and
+//! execution) pin [`TN_BLOB_EXCESS_GAS_AND_PRICE`] instead of deriving the price from the header,
+//! so `BLOBBASEFEE` reads the same value in simulation, in re-execution and on chain.
 
 use super::{TNBlockAssembler, TNBlockExecutionCtx, TNBlockExecutorFactory, TNEvmFactory};
 use crate::{error::TnRethError, payload::TNPayload, TNPrimitives};
@@ -49,6 +55,20 @@ use std::sync::Arc;
 use tn_types::{
     gas_accumulator::GasAccumulator, BlockHeader as _, SealedBlock, SealedHeader, B256, U256,
 };
+
+/// The blob gas values every TN block environment carries: no excess blob gas and a blob gas
+/// price of `u128::MAX`.
+///
+/// EIP-4844 transactions never reach a TN block (the batch builder skips them, the batch validator
+/// rejects any batch that carries one, and the pool prices them out of the pending set with
+/// `pending_block_blob_fee = Some(u128::MAX)`), so blob gas has no market and the price is pinned
+/// rather than derived from the header. `evm_env` and `next_evm_env` both use this one value:
+/// deriving it in only one place made `BLOBBASEFEE` return `1` in every `evm_env` environment
+/// (`eth_call` / `eth_estimateGas` / tracing, and a block re-executed through
+/// `executor_for_block`) but `u128::MAX` in block building and execution (#1205). Every TN header
+/// carries `excess_blob_gas: Some(0)` (`TNBlockAssembler`), matching `excess_blob_gas` here.
+const TN_BLOB_EXCESS_GAS_AND_PRICE: BlobExcessGasAndPrice =
+    BlobExcessGasAndPrice { excess_blob_gas: 0, blob_gasprice: u128::MAX };
 
 /// TN-related EVM configuration.
 #[derive(Debug, Clone)]
@@ -119,15 +139,13 @@ impl ConfigureEvm for TnEvmConfig {
             cfg_env.set_max_blobs_per_tx(blob_params.max_blob_count);
         }
 
-        // derive the EIP-4844 blob fees from the header's `excess_blob_gas` and the current
-        // blobparams
-        let blob_excess_gas_and_price = header
-            .excess_blob_gas
-            .zip(self.chain_spec().blob_params_at_timestamp(header.timestamp))
-            .map(|(excess_blob_gas, params)| {
-                let blob_gasprice = params.calc_blob_fee(excess_blob_gas);
-                BlobExcessGasAndPrice { excess_blob_gas, blob_gasprice }
-            });
+        // TN does not price blob gas: use the value block execution pins (see
+        // `TN_BLOB_EXCESS_GAS_AND_PRICE`) instead of `params.calc_blob_fee(excess_blob_gas)`,
+        // which is `1` for every TN header and made `BLOBBASEFEE` disagree between `eth_call` and
+        // execution. Keep upstream's `Option` shape: `None` only when the header carries no
+        // `excess_blob_gas` or the timestamp has no blob params (pre-Cancun, which TN never is).
+        let blob_excess_gas_and_price =
+            header.excess_blob_gas.zip(blob_params).map(|_| TN_BLOB_EXCESS_GAS_AND_PRICE);
 
         let block_env = BlockEnv {
             number: U256::from(header.number()),
@@ -175,10 +193,8 @@ impl ConfigureEvm for TnEvmConfig {
             prevrandao: Some(payload.prev_randao()),
             gas_limit: payload.gas_limit,
             basefee: payload.base_fee_per_gas,
-            blob_excess_gas_and_price: Some(BlobExcessGasAndPrice {
-                excess_blob_gas: 0,       // no excess gas for blobs
-                blob_gasprice: u128::MAX, // eip4844 transactions are ignored
-            }),
+            // eip4844 transactions are ignored: pinned, and shared with `evm_env`
+            blob_excess_gas_and_price: Some(TN_BLOB_EXCESS_GAS_AND_PRICE),
         };
 
         let evm_env = EvmEnv::new(cfg, block_env);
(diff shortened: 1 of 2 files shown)


crates/tn-reth/src/evm/config.rs
```diff
@@ -50,6 +56,20 @@ use tn_types::{
...
+///
+/// EIP-4844 transactions never reach a TN block (the batch builder skips them, the batch validator
+/// rejects any batch that carries one, and the pool prices them out of the pending set with
+/// `pending_block_blob_fee = Some(u128::MAX)`), so blob gas has no market and the price is pinned
+/// rather than derived from the header. `evm_env` and `next_evm_env` both use this one value:
+/// deriving it in only one place made `BLOBBASEFEE` return `1` in every `evm_env` environment
+/// (`eth_call` / `eth_estimateGas` / tracing, and a block re-executed through
+/// `executor_for_block`) but `u128::MAX` in block building and execution (#1205). Every TN header
+/// carries `excess_blob_gas: Some(0)` (`TNBlockAssembler`), matching `excess_blob_gas` here.
+const TN_BLOB_EXCESS_GAS_AND_PRICE: BlobExcessGasAndPrice =
```
grantkee, line 70: great idea

grantkee (approved):
Good work!

Haxatron:
Fix verified: both functions now share the same `TN_BLOB_EXCESS_GAS_AND_PRICE` constant

PR #1209, opened 2026-08-18
Close a race in the basefee restart e2e test.

sstanfield:
Diagnosis (your question)

  Why it failed: The timed-out transfer was #3 (basefee-restart-local) — routed through the just-restarted node2's own RPC. Node2 was still catching up (chasing 4 epochs, briefly stalled fetching a missing consensus pack, its stale-epoch
  proposals rejected by peers) and didn't converge to the live epoch until ~16s after restart. The test's assertion 2b only waited for node2 to serve a historical block (block 5), which it can do while still epochs-behind on live consensus — so
  the tx was submitted through a node whose worker/batch path wasn't participating yet, and the 15s window expired. The live quorum was healthy throughout; transfers #1 and #2 landed fine.

  crates/e2e-tests/tests/it/basefee.rs — before routing transfer #3 through the restarted node, wait for it to reach the network's current epoch (not just serve a historical block), reusing the existing current_epoch / wait_for_epoch_at_least
  helpers:

  let live_epoch = current_epoch(&provider).await?.epoch_id;
  wait_for_epoch_at_least(&restarted_provider, live_epoch).await?;


MavenRain (approved):
Looks good

grantkee (approved):

PR #1214, opened 2026-08-18
fix(tn-reth): refuse DELEGATECALL/CALLCODE frames in the TEL precompile dispatcher

### LINKED ISSUE #1210: tel precompile: dispatcher accepts DELEGATECALL/CALLCODE frames, so input.caller is spoofable to governance; harden by requiring a direct call to 0x7e1
## Summary

`telcoin_precompile` dispatches on the 4-byte selector and gates state mutation on `input.caller` (through `has_governance_role` / `has_mint_role`), but it never checks the call scheme or the frame's target address (`crates/tn-reth/src/evm/tel_precompile/mod.rs:146-199`).  A precompile is looked up by `bytecode_address` (`alloy-evm-0.27.3/src/precompiles.rs:439`: `self.get(&inputs.bytecode_address)`), and a `DELEGATECALL` sets `bytecode_address` to the callee while preserving the parent frame's `caller` (`revm-interpreter-32.0.0/src/instructions/contract.rs`, `delegate_call`: `bytecode_address: to`, `caller: context.interpreter.input.caller_address()`).  So if a contract `A` executes `DELEGATECALL(0x7e1, ...)`, the precompile runs with `input.caller` equal to `A`'s own caller, not `A`.

Because every handler hardcodes `TELCOIN_PRECOMPILE_ADDRESS` for all storage and balance operations (`burnable.rs` `sstore`/`sload`/`balance_decr`;  `helpers.rs:43-77` take the address as an explicit argument), such a `DELEGATECALL` operates on the **canonical** `0x7e1` ledger, not on the delegatecaller's storage.  This is unlike a normal Solidity contract, where `DELEGATECALL` redirects storage to the caller.

The module doc states "Any `CALL`/`STATICCALL` targeting that address is intercepted and routed to `telcoin_precompile`" (`mod.rs:10-11`), but the dispatcher enforces no such restriction: `DELEGATECALL` and `CALLCODE` frames reach it too.

Severity: informational / low, defense-in-depth hardening.  There is no attacker-profit path today (see below) and no known exploit;  the concern is that a spoofable `caller` is a latent coupling that a future handler change could turn live.  Raised by Haxatron in the Cantina collaborative review;  this issue records the verification, corrects the impact framing, and proposes a one-line guard.

## The mechanic (verified)

| Call scheme | `caller` seen by the precompile | `target_address` | Reaches `0x7e1` state? |
| --- | --- | --- | --- |
| `CALL 0x7e1` (direct) | the immediate caller | `0x7e1` (`= bytecode_address`) | yes (intended) |
| `STATICCALL 0x7e1` (direct) | the immediate caller | `0x7e1` (`= bytecode_address`) | yes, read-only selectors only |
| `A` does `DELEGATECALL 0x7e1` | **`A`'s own caller (preserved)** | `A` (`!= bytecode_address`) | **yes, canonical `0x7e1`** |
| `A` does `CALLCODE 0x7e1` | `A` (not preserved past `A`) | `A` (`!= bytecode_address`) | yes, canonical `0x7e1` |

Frame construction (`revm-interpreter-32.0.0/src/instructions/contract.rs`): `call` sets `target_address: to`, `caller: self.target_address()`;  `static_call` sets `target_address: to`, `caller: self.target_address()`, `is_static: true`;  `delegate_call` sets `target_address: self.target_address()` (preserved), `caller: self.caller_address()` (grandparent-preserved), `bytecode_address: to`;  `call_code` sets `target_address: self.target_address()`, `caller: self.target_address()`, `bytecode_address: to`.  Dispatch (`alloy-evm-0.27.3/src/precompiles.rs:439`) keys only on `bytecode_address`, so all four schemes route into `telcoin_precompile`;  `PrecompileInput` (built at `precompiles.rs:468-477`) carries `caller`, `target_address`, and `bytecode_address`, but the TN handlers read only `caller` (for authorization) and never `target_address`.

The one property this delivers to an attacker: under `DELEGATECALL`, `input.caller` is the grandparent caller.  If the Governance Safe (`GOVERNANCE_SAFE_ADDRESS = 0x...7a0`, `crates/config/src/genesis.rs:37`) `CALL`s a contract `A`, and `A` then `DELEGATECALL`s `0x7e1`, the precompile sees `input.caller == GOVERNANCE_SAFE_ADDRESS` and every governance gate passes.

## Why this is not exploitable today

The spoofable `caller` does not yield attacker profit, and the practical precondition is a full governance compromise:

1. **`mint(uint256)` cannot direct funds to an attacker.** The recipient is hardcoded (`burnable.rs:190`: `let recipient = GOVERNANCE_SAFE_ADDRESS;`), and the pending mint is keyed by that address (`amount_slot(recipient)`), ignoring calldata.  `mint` only creates a 7-day-timelocked pending amount for governance.
2. **`claim(address)` cannot pay an attacker.** The recipient is read from calldata (`burnable.rs:267`: `Address::from_slice(&calldata[12..32])`), so an attacker can pass their own address, but only `amount_slot(GOVERNANCE_SAFE_ADDRESS)` is ever seeded, so `amount_slot(attacker)` is zero and the call reverts at "claim: no pending mint" (`burnable.rs:276-278`).  A nonzero claim credits `GOVERNANCE_SAFE_ADDRESS`'s own native balance, not the attacker's.
3. **`burn(uint256)` has no beneficiary.** It decrements `TELCOIN_PRECOMPILE_ADDRESS`'s own native balance and `totalSupply` (`burnable.rs:376`, `:388`) and returns "insufficient balance" if `0x7e1` holds nothing (`helpers.rs:70-72`).  It is pure destruction, so the maximum realized impact is griefing of precompile-held TEL, not theft.
4. **The precondition equals a governance-transaction-execution compromise.** The only route to `input.caller == GOVERNANCE_SAFE_ADDRESS` is a genuine top-level call executed by the Safe.  All protocol-injected EVM calls use `SYSTEM_ADDRESS`, not the governance address (`crates/tn-reth/src/env/epoch.rs`, `block.rs`);  base-fee crediting is a balance hook, not a `CALL`;  and the precompile makes no outbound calls.  An attacker who can make governance execute `CALL A` can equally make it execute `CALL 0x7e1` directly, so `DELEGATECALL` grants no capability over a direct governance call.
5. **The `CALLCODE` and EIP-7702 parts of the original claim do not hold.** `CALLCODE` sets the new frame's `caller` to `self.target_address()` (the immediate caller), not the preserved grandparent, so `caller != GOVERNANCE_SAFE_ADDRESS` unless the Safe itself performs the `CALLCODE`.  EIP-7702 authorizes an EOA to adopt delegated code;  it cannot forge a contract's identity in the `caller` field, and `GOVERNANCE_SAFE_ADDRESS` is a genesis-deployed Safe with no signing key.  Neither is a distinct spoofing vector against this precompile.

## Why it is still worth hardening

- **Doc-vs-code mismatch.** `mod.rs:10-11` asserts CALL/STATICCALL-only interception, but the dispatcher enforces no scheme guard.  Code should match the documented contract.
- **Latent coupling.** `input.caller` is grandparent-preserved and spoofable to governance today;  the only reason it is inert is the hardcoded `mint` recipient plus `claim`'s revert on an empty slot.  Any later change that lets `mint`/`claim` credit a calldata-chosen recipient without re-gating, or that makes a handler read `input.target_address`, converts this into a live unauthorized-mint.  The guard removes that whole class of future footgun for one comparison.
- **Surprising storage semantics.** Because the handlers hardcode `0x7e1`, a `DELEGATECALL` into the precompile hits canonical state rather than the delegatecaller's, which breaks the usual `DELEGATECALL` mental model and is worth refusing explicitly.

## Suggested fix

At the top of `telcoin_precompile`, before selector dispatch, require that the precompile was entered by a direct call to its own address.  A direct `CALL`/`STATICCALL` to `0x7e1` sets `target_address == bytecode_address == 0x7e1`;  `DELEGATECALL` and `CALLCODE` set `target_address` to the caller's own frame address, so `target_address != bytecode_address`.  Add the guard as the first statement of `telcoin_precompile`, ahead of the existing `input.data.len() < 4` check:

```rust
/// Rejection emitted when the precompile is entered via DELEGATECALL/CALLCODE rather than a
/// direct call to its own address.
const INDIRECT_ENTRY: &str = "telcoin precompile: only a direct call to 0x7e1 is permitted";

// Precompiles are dispatched by `bytecode_address`, so a DELEGATECALL/CALLCODE to 0x7e1 reaches
// this code with `input.caller` preserved from the parent frame. Every handler operates on the
// canonical 0x7e1 ledger, so refuse any frame that did not target 0x7e1 directly: a direct
// CALL/STATICCALL has `target_address == bytecode_address`, whereas DELEGATECALL/CALLCODE carry
// the caller's own address in `target_address`.
(input.target_address == TELCOIN_PRECOMPILE_ADDRESS)
    .then_some(())
    .ok_or_else(|| PrecompileError::Other(INDIRECT_ENTRY.into()))?;
```

This rejects `DELEGATECALL` and `CALLCODE` while leaving every direct `CALL`/`STATICCALL` (including `totalSupply()` and, with the faucet feature, `hasMintRole`) working, since those set `target_address == 0x7e1`.  It composes with the existing static-call guard rather than replacing it.

Add a negative test: a frame with `caller == GOVERNANCE_SAFE_ADDRESS` but `target_address != TELCOIN_PRECOMPILE_ADDRESS` must revert without mutating state.  This requires the test harness to set `target_address` independently of `bytecode_address` (today `precompile_test_utils.rs` drives direct calls only).

## Load-bearing assumptions

- The "no escalation beyond a governance compromise" argument (point 4) assumes the production Governance Safe's module and threshold configuration is tight.  The Safe and its modules are not in this repo.  A permissive executor module that let a lower-trust principal make the Safe `CALL` or `DELEGATECALL` arbitrary targets without full owner-threshold sign-off would lower the precondition to that module's bar.  The guard is valuable precisely because it does not depend on that off-repo configuration.
- If the Safe is ever induced to `DELEGATECALL` attacker code directly, that attacker frame's `target_address` becomes the Safe's own address, so a subsequent `CALLCODE` from it would carry `caller == Safe`.  This is a real second hop, but it rides on the same stronger "Safe executes raw `DELEGATECALL`" precondition rather than being an independent `CALLCODE` bug.  The proposed `target_address == 0x7e1` guard rejects that hop too.

## Environment

- Repo: `Telcoin-Association/telcoin-network`, `main` at `648b6f81`.
- Pins from `Cargo.lock`: `alloy-evm 0.27.3`, `revm 34.0.0`, `revm-handler 15.0.0`, `revm-interpreter 32.0.0`, `revm-precompile 32.1.0`.
- Files: `crates/tn-reth/src/evm/tel_precompile/mod.rs`, `burnable.rs`, `helpers.rs`;  `crates/config/src/genesis.rs`.

## Credit

Haxatron (Cantina collaborative review) raised the absence of `DELEGATECALL`/`CALLCODE`/EIP-7702 handling.  This issue confirms the `DELEGATECALL` caller-preservation mechanic, documents why it does not currently yield an unauthorized mint or attacker profit (and why the `CALLCODE`/7702 variants do not apply), and proposes the direct-call guard as defense in depth.

MavenRain:
Closes #1210.  (Cantina #6)

## Problem

`telcoin_precompile` authorizes every state-mutating selector on `input.caller` (through `has_governance_role` / `has_mint_role`) but never checks how the frame reached it. revm looks a precompile up by the frame's `bytecode_address` alone (`PrecompilesMap::get(&inputs.bytecode_address)`, alloy-evm 0.27.3), so `DELEGATECALL` and `CALLCODE` frames whose code address is `0x7e1` run the dispatcher exactly like a `CALL`. Two properties of that frame shape make it unsafe to serve:

- **The caller is spoofable.** `DELEGATECALL` preserves the parent frame's `msg.sender`. A contract `A` that governance chose to `CALL` presents `caller == GOVERNANCE_SAFE_ADDRESS` when it delegates into `0x7e1`, and every governance gate passes.
- **The storage does not follow `DELEGATECALL` semantics.** Every handler passes the literal `TELCOIN_PRECOMPILE_ADDRESS` to `sstore` / `sload` and the balance helpers, so the indirect frame acts on the canonical `0x7e1` ledger, not on `A`'s storage as it would against ordinary bytecode. `test_delegatecall_writes_target_precompile_storage` pinned exactly this: governance -> relay -> `DELEGATECALL 0x7e1` -> `mint(1234)` landed in the canonical pending-mint slot.

The module doc and README both state that "`CALL`/`STATICCALL` targeting that address" is what the dispatcher serves; the code enforced no such restriction.

Threat model. On mainnet this is latent, not live: `mint` hardcodes governance as the recipient, `claim` reverts on an empty pending slot, `burn` has no beneficiary, and presenting a governance `caller` at all needs governance to execute the transaction (at which point a direct call to `0x7e1` is equally available), so there is no attacker-profit path today (the issue records that verification). Under the `faucet` feature it is live: `mint(address, uint256)` credits a calldata-chosen recipient, so any contract a mint-role holder chose to `CALL` could mint to an address of its choosing. Either way the authorization gates were trusting a spoofable identity, and any later handler change that credits a calldata-chosen recipient on mainnet would have turned the latent path live without a compile-time signal. `CALLCODE` presents the calling contract itself as `caller`, so it spoofs nothing on its own, but it is the same indirect frame shape and is refused for the same reason. EIP-7702 is not a vector here (dispatch keys on the EOA's own address, and #1169 rejects set-code transactions at admission).

## Fix

The first statement of `telcoin_precompile`, ahead of the calldata length check and of selector dispatch, refuses any frame that is not a direct call, through the upstream predicate `PrecompileInput::is_direct_call()` (`target_address == bytecode_address`). A direct `CALL`/`STATICCALL` sets both to `0x7e1`; `DELEGATECALL`/`CALLCODE` keep the calling contract's own address as `target_address`. The rejection is `PrecompileError::Other("telcoin precompile: only a direct call to 0x7e1 is permitted")`, which reverts the inner frame like the existing static-call rejection does. Read-only selectors are refused too: an indirect frame has no legitimate use of this precompile, and refusing it whole is cheaper to reason about than a per-selector exemption. The guard is attacker-independent and costs one address comparison per call; it composes with the existing static-call guard rather than replacing it.

## Changes

- [x] `crates/tn-reth/src/evm/tel_precompile/mod.rs`: `INDIRECT_ENTRY` rejection constant; direct-call guard as the first statement of `telcoin_precompile`; module docs (interception paragraph, access-control list) and a `# Direct-call guard` section on the dispatcher doc.
- [x] `crates/tn-reth/tests/it/precompile_relays.rs`: `DELEGATECALL_RELAY_BYTECODE` (`0xf4`, six operands) and `CALLCODE_RELAY_BYTECODE` (`0xf2`, seven operands), byte-identical to the existing `STATICCALL` / `CALL` relays but for the opcode, both returning the inner success flag; module docs extended for the two new frame shapes and their hosting addresses.
- [x] `crates/tn-reth/tests/it/tel_precompile_props.rs`: `test_delegatecall_writes_target_precompile_storage` (which pinned the pre-fix behaviour) replaced by `test_delegatecall_cannot_reach_precompile` (governance -> relay at a fresh address -> `DELEGATECALL` `mint`; inner flag false, pending slot untouched under both `0x7e1` and the relay), `test_delegatecall_refuses_read_only_selectors_too` (`totalSupply()` refused under `DELEGATECALL`, served under `STATICCALL`; scheme is the only variable), and `test_callcode_cannot_reach_precompile` (relay hosted at governance so `caller` genuinely is governance; still refused). The forwarding `RELAY_BYTECODE` that could not distinguish accept from refuse is gone.
- [x] `crates/tn-reth/tests/it/tel_precompile_faucet_props.rs`: `test_delegatecall_cannot_mint_faucet` (governance -> relay -> `DELEGATECALL` `mint(RECIPIENT, 1234)`; refused, recipient balance and `totalSupply` unchanged) and `test_callcode_cannot_mint_faucet` (relay hosted at governance, driven by `USER`; refused), so the faucet build carries the live-path coverage the mainnet-only module cannot, and every relay is exercised under both feature sets.
- [x] `crates/tn-reth/src/evm/tel_precompile/README.md`: interception paragraph, access-control row for `DELEGATECALL` / `CALLCODE`, and the "`DELEGATECALL` semantics under revm" section rewritten as "Direct-call guard (`DELEGATECALL` / `CALLCODE`)" with the mechanic, the mainnet-latent / faucet-live framing, the guard, and the regression tests.

No selector, gas, ABI, or storage-layout change. Direct `CALL`/`STATICCALL` paths are untouched: the existing `CALL` relay positive controls (`test_staticcall_cannot_mutate_precompile_state`, `test_staticcall_cannot_grant_mint_role`), the `STATICCALL` read-only tests, and every top-level unit test in `burnable.rs` / `faucet.rs` still pass.

## Testing

Under the pinned toolchain from `rust-toolchain.toml`:

- `cargo +1.94 test -p tn-reth --lib evm::tel_precompile`: all direct-call unit tests pass (guard leaves top-level calls untouched).
- `cargo +1.94 test -p tn-reth --test it -- delegatecall callcode staticcall`: the three new mainnet relay tests plus the two `STATICCALL` tests (whose `CALL` relay halves are the positive controls) pass.
- `cargo +1.94 test -p tn-reth --test it --features faucet -- delegatecall callcode staticcall`: the two new faucet relay tests plus the two faucet `STATICCALL` tests pass.
- `cargo +nightly clippy -p tn-reth --all-features --all-targets --no-deps -- -D warnings`: clean.
- Mutation confirmation: with the guard removed (the four-line `is_direct_call` statement deleted, nothing else touched), all five new tests fail (`DELEGATECALL`/`CALLCODE` mint accepted, pending slot written or recipient credited; `totalSupply()` served under `DELEGATECALL`); restored, all pass. Each new test has been seen to fail against unfixed code.

CI (nightly fmt / clippy `--all --all-features`, `nextest --workspace`, adiri lane) fires on push; the pinned `cargo +1.94 test --no-run --workspace` attest runs on the second machine.

### DIFF
--- a/crates/tn-reth/src/evm/tel_precompile/README.md
+++ b/crates/tn-reth/src/evm/tel_precompile/README.md
@@ -2,7 +2,7 @@
 
 This directory implements a **native token-issuance precompile** for the Telcoin (TEL) token. The precompile owns the on-chain mint/claim/burn lifecycle and exposes a single read-only view (`totalSupply`). It does **not** expose an ERC-20 transfer/approve/permit surface — those flows live in user-space contracts and rely on native value transfers, which are equivalent to ERC-20 transfers because TEL balances are native account balances.
 
-The precompile is registered as a `DynPrecompile` inside reth's `PrecompilesMap` at address `0x00000000000000000000000000000000000007e1`. Any `CALL` or `STATICCALL` targeting this address is intercepted by the dispatcher in `mod.rs`, which routes on the 4-byte function selector. Routing is not unconditional for a `STATICCALL`: only the read-only selectors are served in a static frame — `totalSupply`, plus `hasMintRole` under the `faucet` feature — and every state-mutating selector is refused before dispatch. See "`STATICCALL` write protection" under Security considerations.
+The precompile is registered as a `DynPrecompile` inside reth's `PrecompilesMap` at address `0x00000000000000000000000000000000000007e1`. Every call frame whose code address is `0x7e1` is intercepted by the dispatcher in `mod.rs`. revm keys that lookup on the frame's `bytecode_address` alone, so `DELEGATECALL` and `CALLCODE` frames arrive there too; the dispatcher refuses those before doing anything else and serves only a direct `CALL` or `STATICCALL`, on which it routes on the 4-byte function selector (see "Direct-call guard" under Security considerations). Routing is not unconditional for a `STATICCALL`: only the read-only selectors are served in a static frame — `totalSupply`, plus `hasMintRole` under the `faucet` feature — and every state-mutating selector is refused before dispatch. See "`STATICCALL` write protection" under Security considerations.
 
 ## Module map
 
@@ -70,6 +70,7 @@ No pending state, no timelock. Mint roles can be granted/revoked by governance.
 | `grantMintRole` / `revokeMintRole` | Governance only (faucet feature)                   |
 | `hasMintRole` / `totalSupply`      | Any account (read-only)                            |
 | Any selector inside a `STATICCALL` | Read-only only — `totalSupply` (plus `hasMintRole` under `faucet`); mutating selectors refused regardless of caller |
+| Any selector via `DELEGATECALL` / `CALLCODE` | Refused before dispatch, regardless of caller and selector; only a direct `CALL`/`STATICCALL` to `0x7e1` reaches a handler |
 
 Governance is identified by `GOVERNANCE_SAFE_ADDRESS` from `tn-config`.
 
@@ -87,9 +88,18 @@ The rationale is that none of those selectors needed to live at the protocol lev
 
 Shrinking the surface to issuance-only is the actual reason for the deletion; the resulting protocol is simpler and exposes less authority than a full ERC-20 implementation would.
 
-### `DELEGATECALL` semantics under revm
+### Direct-call guard (`DELEGATECALL` / `CALLCODE`)
 
-For completeness: a contract that `DELEGATECALL`s into `0x7e1` runs the precompile's logic, but every `SSTORE` the precompile performs targets the literal address argument it passes — `TELCOIN_PRECOMPILE_ADDRESS` — not the calling contract's storage. The precompile dispatcher routes through `EvmInternals::sstore(TELCOIN_PRECOMPILE_ADDRESS, …)`, and revm's journaled state writes to the address argument verbatim with no `DELEGATECALL`-aware rewrite. A regression test in `crates/tn-reth/tests/it/tel_precompile_props.rs` (`test_delegatecall_writes_target_precompile_storage`) pins this behaviour against future revm upgrades.
+revm dispatches a precompile by the frame's `bytecode_address` alone (`PrecompilesMap::get(&inputs.bytecode_address)`), with no check on the call scheme, so a contract that `DELEGATECALL`s or `CALLCODE`s into `0x7e1` runs the dispatcher just as a `CALL` does. Two things make that frame shape unsafe to serve:
+
+- **The `caller` is not the contract in front of the precompile.** `DELEGATECALL` preserves the parent frame's `msg.sender`, so a contract `A` that governance chose to `CALL` presents `caller == GOVERNANCE_SAFE_ADDRESS` when it delegates into `0x7e1`, and every governance gate passes. (`CALLCODE` presents `A` itself as the caller, so it spoofs nothing on its own, but it is the same indirect frame shape and is refused for the same reason.)
+- **The storage does not follow `DELEGATECALL` semantics.** Every handler passes the literal `TELCOIN_PRECOMPILE_ADDRESS` to `EvmInternals::sstore` / `sload` and the balance helpers, and revm's journaled state writes to that address verbatim with no `DELEGATECALL`-aware rewrite. An indirect frame therefore acts on the **canonical** `0x7e1` ledger, not on the delegating contract's storage as it would against ordinary bytecode.
+
+On mainnet the combination is latent rather than live: `mint` hardcodes governance as the recipient and `claim` reverts on an empty pending slot, so a spoofed governance caller cannot direct funds to itself, and presenting a governance `caller` at all requires governance to execute the transaction (at which point a direct call to `0x7e1` is equally available to it). Under the `faucet` feature it is live: `mint(address, uint256)` credits a calldata-chosen recipient, so any contract a mint-role holder chose to `CALL` could mint to an address of its choosing. Either way the authorization gates were trusting a spoofable identity.
+
+The dispatcher therefore refuses every frame that is not a direct call, as its first check and ahead of selector dispatch: `PrecompileInput::is_direct_call()` (`target_address == bytecode_address`) holds for a direct `CALL`/`STATICCALL`, which set both to `0x7e1`, and fails for `DELEGATECALL`/`CALLCODE`, which keep the calling contract's own address as `target_address`. Read-only selectors are refused too; an indirect frame has no legitimate use of this precompile.
+
+Regression tests reach the precompile through the `DELEGATECALL` and `CALLCODE` relays in `crates/tn-reth/tests/it/precompile_relays.rs`: `test_delegatecall_cannot_reach_precompile`, `test_delegatecall_refuses_read_only_selectors_too` and `test_callcode_cannot_reach_precompile` (mainnet, in `tel_precompile_props.rs`), and `test_delegatecall_cannot_mint_faucet` and `test_callcode_cannot_mint_faucet` (`faucet`, in `tel_precompile_faucet_props.rs`). The `CALL` and `STATICCALL` positive controls in the static-call tests below prove the guard leaves direct calls untouched.
 
 ### `STATICCALL` write protection
 
--- a/crates/tn-reth/src/evm/tel_precompile/mod.rs
+++ b/crates/tn-reth/src/evm/tel_precompile/mod.rs
@@ -7,11 +7,15 @@
 //! precompile.
 //!
 //! The precompile is registered as a [`DynPrecompile`] inside a [`PrecompilesMap`] at
-//! [`TELCOIN_PRECOMPILE_ADDRESS`] (`0x7e1`). Any `CALL`/`STATICCALL` targeting that address is
-//! intercepted and routed to [`telcoin_precompile`] instead of executing bytecode. Bypassing
-//! bytecode execution also bypasses the interpreter's static-call write protection, so
-//! [`telcoin_precompile`] enforces that itself: inside a `STATICCALL` frame it serves the
-//! read-only selectors and refuses every state-mutating one.
+//! [`TELCOIN_PRECOMPILE_ADDRESS`] (`0x7e1`). Every call frame whose code address is `0x7e1` is
+//! intercepted and routed to [`telcoin_precompile`] instead of executing bytecode. revm keys that
+//! lookup on the frame's `bytecode_address` alone, so `DELEGATECALL` and `CALLCODE` frames reach
+//! the dispatcher as well as `CALL`/`STATICCALL`. Bypassing bytecode execution leaves two frame
+//! properties for [`telcoin_precompile`] to enforce itself. It refuses any frame that did not
+//! reach `0x7e1` by a direct call: under `DELEGATECALL` the `caller` it would authorize is
+//! inherited from the parent frame, while every handler still writes the canonical `0x7e1`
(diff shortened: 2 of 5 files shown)


grantkee (dismissed):
Code LGTM - merge conflict, but looks like it's an easy "keep both". Good work @MavenRain - we need to test syncing before we deploy this because this could cause a fork on testnet, but I'm pretty sure governance hasn't used these op codes yet

grantkee (approved):

sstanfield (approved):

Haxatron:
Fix verified: the TEL precompile now only accepts direct CALL or STATICCALL via `is_direct_call` which verifies that `self.target_address == self.bytecode_address`

PR #1215, opened 2026-08-18
fix(tn-reth): gate blsVerify on the O(1) gas floor before decoding calldata

### LINKED ISSUE #1211: tn-reth: blsVerify precompile ABI-decodes calldata before any gas gate
## Summary

This is a hardening / cost-asymmetry item. Suggested triage: Medium (borderline Low).

`handle_bls_verify` calls `blsVerifyCall::abi_decode_raw` before the message cap and before the gas check. The decode copies all three `bytes` fields into owned `Bytes` values. That is three allocations plus three memcpy per call. A warm `STATICCALL` that forwards 0 gas still pays that copy. The caller pays about 140 gas per call.

The work is bounded per block by purchased gas. It is transient. It does not corrupt state. Batch validation does not execute transactions, so only executing nodes pay. The fix is an O(1) gas floor before the decode. The fix keeps the Ok-set and `gas_used` byte-identical.

## Where

origin/main `2ea18bec`, `crates/tn-reth/src/evm/bls_precompile/mod.rs`, `handle_bls_verify` at lines 149-168.

Lines 150-158 verbatim:

```rust

    let message_len = decoded.message.len();
    let cost = (message_len <= MAX_BLS_VERIFY_MESSAGE_LEN)
        .then_some(message_len)
        .ok_or_else(|| PrecompileError::Other("blsVerify: message exceeds maximum length".into()))
        .and_then(|len| bls_verify_gas_cost(len).ok_or(PrecompileError::OutOfGas))
        .and_then(|cost| (gas_limit >= cost).then_some(cost).ok_or(PrecompileError::OutOfGas))?;
```

Related lines in the same file:

- Line 74: `const BLS_VERIFY_GAS_COST: u64 = 150_000;`
- Line 92: `const MAX_BLS_VERIFY_MESSAGE_LEN: usize = 4096;`
- Lines 121-122: `fn bls_precompile(input: PrecompileInput<'_>) -> PrecompileResult { dispatch(input.data, input.gas)`
- Lines 143-148: the doc comment that describes the current ordering (quoted under History).
- Line 180: `if compressed_sig.len() != 48 || compressed_pubkey.len() != 96 {` inside `bls_verify`. This gate runs after the decode and after the gas check.

### The decode is the only unmetered O(calldata) work per call

- alloy-evm 0.27.3 `PrecompilesMap::run` (`src/precompiles.rs:433-467`) passes a borrowed `shared_memory_buffer_slice` for `CallInput::SharedBuffer`. It does not copy the calldata.
- revm-handler 15.0.0 `make_call_frame` (`src/frame.rs:183-192`) clones `inputs.input`. For `SharedBuffer` that clone is a `Range<usize>`, so it is O(1).
- revm-handler 15.0.0 has no `gas_limit == 0` short-circuit (`src/frame.rs:192`). alloy-evm has none either (`src/precompiles.rs:445` and `:470`). The precompile runs with 0 forwarded gas.
- alloy-sol-types 1.5.7 copies each `bytes` field: `PackedSeqToken::into_bytes` is `Bytes::copy_from_slice(self.0)` (`src/abi/token.rs:509-512`). `SolType for Bytes` calls `token.into_bytes()` (`src/types/data_type.rs:318-334`).
- All pinned eth precompiles gate an O(1) cost before O(len) work. `identity_run` checks `gas_used > gas_limit` before `Bytes::copy_from_slice` (revm-precompile 32.1.0 `src/identity.rs:23-30`). `sha256_run` and `ripemd160_run` do the same (`src/hash.rs:25-28`, `:41-44`). The cheapest metered memcpy in this EVM is 3 gas per 32-byte word.

So the three `Bytes::copy_from_slice` calls in `abi_decode_raw` are the only unmetered O(len) work per call.

### The message cap does not bound the copy

`signature` and `pubkey` are unbounded `bytes` in the ABI (`mod.rs:56-59`). Their 48 / 96 length gate is inside `bls_verify` (`mod.rs:180`). It runs after the decode and after the gas gate. A multi-MB signature with a small message passes the 4096-byte message cap. Its cost is about 150k gas. With 0 forwarded gas the call returns `OutOfGas` after the full copy.

The decoder accepts overlapping offsets (alloy-sol-types 1.5.7 `src/abi/decoder.rs:124-132`, `child()` slices at any in-range offset). All three fields can point at the same region. That gives up to 3x the bytes per call.

### Attacker cost per call

- The precompile is always warm. `warm_addresses()` (alloy-evm 0.27.3 `src/precompiles.rs:499-501`) feeds `journal.warm_precompiles` in `load_accounts` (revm-handler 15.0.0 `src/pre_execution.rs:18-38`).
- Warm `STATICCALL` base is `WARM_STORAGE_READ_COST = 100` (revm-interpreter 32.0.0 `src/instructions.rs:126`).
- The caller can forward 0 gas. Forwarded gas is `min(63/64 of remaining, stack request)` (revm-interpreter 32.0.0 `src/instructions/contract/call_helpers.rs:83-92`). `STATICCALL` has no value stipend.
- Memory expansion is charged once. `resize_memory` charges only on growth (`call_helpers.rs:34-49`). A loop reuses the expanded region.
- No memory cap blocks a 2.27 MB region. `memory_limit()` returns `u64::MAX` unless the optional `memory_limit` feature is on (revm-context 13.0.0 `src/cfg.rs:491-496`).
- Loop cost estimate: 100 (call) + about 18 (arg restage) + 2 (POP) + about 19 (loop bookkeeping) + 1 (JUMPDEST). That is about 140 gas per iteration. This is arithmetic from the gas constants, not a measured loop.

### Per-block bound (model)

- Block gas: `max_batch_gas` is 30,000,000 (`crates/types/src/worker/sealed_batch.rs:195-199`). The engine builds one execution block per batch at that limit (`crates/engine/src/payload_builder.rs:188-199`).
- Args region of N words costs `E(N) = 3N + N^2/512` gas once. The optimum for `(30M - E(N)) * N` is N of about 71,043 words. That is 2,273,408 bytes. `E(N)` is about 10.07M gas.
- Calls per block: (30M - 10.07M) / 140 is about 142,350.
- Bytes decoded per block: about 324 GB with a single copy. Up to about 970 GB when all three fields alias the same region.

### Bench numbers (with machine caveat)

Micro-bench of `blsVerifyCall::abi_decode_raw` with alloy-sol-types 1.5.7 (`/Users/oobi/Documents/tn-bls-decode-bench`). Apple M2 Pro, 32 GiB, single thread, release build.

| Message | Calldata | Time per call | Rate |
| --- | --- | --- | --- |
| 119 B | 480 B | 0.071 us | 6.8 GB/s |
| 4096 B (cap) | 4448 B | 0.146 us | 30.5 GB/s |
| 64 KiB | 65,888 B | 1.35 us | 48.9 GB/s |
| 1 MiB | 1,048,928 B | 18.3 us | 57.4 GB/s |
| 2.3 MB | 2,300,352 B | 73.9 us | 31.1 GB/s |

The raw floor gate `gas_limit < 150_000` costs about 0.41 ns per call.

Node CPU per attack block, from the 2.3 MB row: 142,350 x 73.9 us is about 10.5 s with a single copy. With 3x aliasing it is about 30 s. The band is about 5.6 s (57 GB/s, cache-resident row) to about 40 s (8 GB/s, conservative).

Machine caveat: the 2.3 MB row is dominated by the macOS large-malloc page-fault path. Linux glibc may reuse mappings and land nearer raw memcpy speed. That is unverified. A normal 30M-gas block was not timed. Under conservative assumptions the amplification is above 10x. It is likely 30x to 300x. Measure one real attack block against one normal block before quoting a ratio.

### Consequence for the node

- The engine executes serially on one blocking task. `MAX_QUEUED_OUTPUTS = 8` (`crates/engine/src/lib.rs:44-52`, `:61-63`). Once the queue is full, the engine stops polling the consensus stream. That pushes backpressure into the `to_engine` channel and the EpochManager forwarder.
- Batches seal at least every 1 s (`chain-configs/{mainnet,testnet}/parameters.yaml`, `max_batch_delay: 1s`). One commit can carry several 30M-gas batches. Seconds of CPU per block therefore matter.
- Batch validators do not execute transactions (`crates/batch-validator/src/validator.rs:39-40`). They vote in the batches.
- Whether the primary itself stalls was not traced. Treat that as unverified.
- Attacker cost is at most 30M gas times the base fee per block. The floor is reth `MIN_PROTOCOL_BASE_FEE` (7 wei) with per-worker epoch adjustment (`crates/tn-reth/src/evm/block.rs:483-521`). That is negligible at any plausible TEL price.

### Severity

Medium (borderline Low). The work is gas-bounded, transient, and self-healing. There is no state corruption and no permanent liveness loss. That keeps it below High. The asymmetry is real: about 140 gas buys an O(MB) unmetered copy, and the pre-fix rate is about 2000x below the MCOPY-purchasable memcpy rate. A sustained attack degrades executed-chain liveness and can backpressure consensus. Haxatron's factual claim is correct. His "not exploitable" view weighs the 100-gas payment; this issue quantifies the ratio.

## History

- Before PR #1051, `handle_bls_verify` began with `if gas_limit < BLS_VERIFY_GAS_COST { return Err(PrecompileError::OutOfGas) }`. That flat floor ran before the decode.
- PR #1051 (merged 2026-08-04) replaced the early floor with decode-then-charge. It added the per-word cost and the message cap.
- The current doc comment (`mod.rs:143-148`) reads: "Decoding precedes the gas gate so the message length - which determines the charge - is known before pricing (the decode is cheap; the previous early gate was only a coarse floor). The message is capped at [`MAX_BLS_VERIFY_MESSAGE_LEN`] and then charged [`bls_verify_gas_cost`] (base plus a per-word surcharge), so a caller supplying a long message pays for the extra hash-to-curve work it induces instead of riding the flat base. An over-cap message is rejected as `Other`; an under-funded call is `OutOfGas`."
- Haxatron surfaced this in a public comment on the Cantina collaborative review on 2026-08-18. He wrote that the allocation is done before any gas is checked or charged. He suggested gas checks first, or light checks on the raw calldata, then `abi_decode_raw`. He judged it not exploitable because the attacker still pays 100 warm `STATICCALL` gas. He called it an improvement. This issue agrees with the factual claim and quantifies the asymmetry.

## Proposed fix

Ordering inside `handle_bls_verify`:

1. `if gas_limit < BLS_VERIFY_GAS_COST { return Err(PrecompileError::OutOfGas); }` This is O(1) and runs before any decode.
2. Optional: an O(1) raw-calldata check (see below).
3. `blsVerifyCall::abi_decode_raw(calldata)`, mapped to `PrecompileError::Other` as today.
4. The `MAX_BLS_VERIFY_MESSAGE_LEN` cap, mapped to `Other` as today.
5. The exact charge `bls_verify_gas_cost(len)` = `150_000 + 12 * ceil(len / 32)`, then `gas_limit >= cost` else `OutOfGas`.
6. `bls_verify`, then `Ok(PrecompileOutput::new(cost, ...))`.

Sketch:

```rust
    // O(1) floor before any allocation: the base charge is the least any accepted call pays.
    (gas_limit >= BLS_VERIFY_GAS_COST)
        .then_some(())
        .ok_or(PrecompileError::OutOfGas)?;

    let message_len = decoded.message.len();
    let cost = (message_len <= MAX_BLS_VERIFY_MESSAGE_LEN)
        .then_some(message_len)
        .ok_or_else(|| PrecompileError::Other("blsVerify: message exceeds maximum length".into()))
        .and_then(|len| bls_verify_gas_cost(len).ok_or(PrecompileError::OutOfGas))
        .and_then(|cost| (gas_limit >= cost).then_some(cost).ok_or(PrecompileError::OutOfGas))?;

}
```

Update the doc comment at `mod.rs:143-148` with the fix. It should say: the O(1) floor runs before the decode; the exact per-word charge runs after the decode.

### Optional raw-calldata ceiling

The largest canonical calldata for `signature` = 48, `pubkey` = 96, `message` = 4096 is 4448 bytes after the selector. Derivation: 3 head words (96) + sig tail (32 + 64 padded) + pubkey tail (32 + 96) + message tail (32 + 4096). With the selector it is 4452 bytes. A `MAX_BLS_VERIFY_CALLDATA_LEN = 4448` ceiling before the decode makes the over-cap path O(1).

Caveat: a total-length ceiling is not behavior-preserving.

- `signature` and `pubkey` are unbounded `bytes`. A canonical call with a 4096-byte message and a 200-byte pubkey is 4576 bytes. Today it returns `Ok(false)` with full gas charged. A ceiling turns it into `Err(Other)`.
- `abi_decode_raw` ignores trailing bytes (alloy-sol-types 1.5.7 `src/abi/decoder.rs:275-280`, no remaining-bytes check). A canonical 480-byte call with junk appended past 4448 bytes succeeds today. A ceiling rejects it.
- Both flips are consensus-visible for any non-`ConsensusRegistry` caller. `ConsensusRegistry` uses fixed 48 / 96 / 119 lengths and is unaffected.

Two safe options:

- Drop the ceiling. The floor gate alone is sufficient (numbers below).
- Replace it with a message-length peek. Read head word 3 as the message offset. Read the 32-byte length word at that offset. If that length exceeds 4096, return the same `Other("blsVerify: message exceeds maximum length")` the post-decode cap returns. The decoder reads the same offset and length, so this never rejects a currently-successful input.

### Error-class change

Only inputs with `gas_limit < 150_000` change variant. Malformed calldata and over-cap messages flip from `PrecompileError::Other` to `PrecompileError::OutOfGas`. Both map to the `return_error!` class (revm-interpreter 32.0.0 `src/instruction_result.rs:166-188`, `PrecompileOOG` and `PrecompileError` together). The caller frame pushes 0 and refunds nothing for either (revm-handler 15.0.0 `src/frame.rs:470-486`). State, gas, and stack are indistinguishable on-chain. Every prior `Ok` had `gas_limit >= cost >= 150_000`, so the Ok-set and `gas_used` are byte-identical.

### Residual work with the fix

Each decode now needs at least 150,000 forwarded gas. With the same 2.27 MB args region: (30M - 10.07M) / 150,100 is 132 calls per block. That is about 300 MB copied per block, or about 900 MB with 3x aliasing. Time: about 5 ms at 57 GB/s, about 10 ms at the 2.3 MB bench row, about 38 ms at 8 GB/s. For comparison, `MCOPY` of the same region costs 213,135 gas, so the same gas buys about 211 MB. The residual is about 1.4x the EVM's own metered memcpy rate. That is the same class as metered work.

## Tests

Ordering test with a positive control. It fails on origin/main and passes after the reorder.

```rust
/// The gas floor is checked before decoding: under-funded calls are OutOfGas even when the calldata
/// would not decode, and the same bytes with the floor met reach the decoder and fail as Other.
#[test]
fn dispatch_gas_floor_precedes_decode() {
    // three head words of 0xff: offsets far past the end, so decode fails with Other
    let malformed = [&blsVerifyCall::SELECTOR[..], &[0xffu8; 96][..]].concat();
    assert!(matches!(dispatch(&malformed, 0), Err(PrecompileError::OutOfGas)));
    assert!(matches!(dispatch(&malformed, BLS_VERIFY_GAS_COST - 1), Err(PrecompileError::OutOfGas)));
    // positive control: with the floor met, the decoder runs and rejects the bytes
    assert!(matches!(dispatch(&malformed, BLS_VERIFY_GAS_COST), Err(PrecompileError::Other(_))));

    // an over-cap 1 MB message under the floor is OutOfGas, not the (post-decode) cap error
    let big_message = vec![0u8; 1 << 20];
    let over = encode_verify(&[0u8; 48], &[0u8; 96], &big_message);
    assert!(matches!(dispatch(&over, 0), Err(PrecompileError::OutOfGas)));

    // a multi-MB signature with a small message passes the cap; under the floor it is OutOfGas
    let big_sig = vec![0u8; 1 << 20];
    let sig_heavy = encode_verify(&big_sig, &[0u8; 96], &[0u8; 119]);
    assert!(matches!(dispatch(&sig_heavy, 0), Err(PrecompileError::OutOfGas)));
}
```

Mutation check: restore decode-first and the `OutOfGas` assertions fail with `Other`. The test pins the ordering.

Existing tests: none pins a variant that flips.

- `dispatch_verify_out_of_gas` (`mod.rs:372-376`) uses `cost - 1` = 150,047. That is above the floor, so it is still `OutOfGas` after the decode.
- `dispatch_rejects_oversized_message` (`mod.rs:437-455`) uses 10,000,000 gas. It is still `Other`.
- `dispatch_rejects_unknown_selector_and_short_input` (`mod.rs:460-469`) fails in `dispatch` before `handle_bls_verify`.

Optional: time one 30M-gas attack block against one normal block in the tn test harness. That locks in the severity band.

## Out of scope / notes

- RPC `eth_call` and `eth_estimateGas` reach the same code path off the consensus path. Their gas caps are node configuration. Track that exposure separately if needed.
- The TEL precompile is untouched.
- Batch validation is untouched. It never executes the EVM.
- Cantina judges may ask for a live PoC timing before they accept Medium. Fall back to `cantina-low` in that case.

Suggested labels: security, bug, cantina-medium

MavenRain:
Closes #1211.

## Problem

`handle_bls_verify` (`crates/tn-reth/src/evm/bls_precompile/mod.rs`) called `blsVerifyCall::abi_decode_raw` before any gas check. The decode copies all three `bytes` arguments into owned `Bytes` (three allocations plus three memcpy). Every other cost in the precompile is charged before the work it covers. Nothing covered this copy:

- The precompile address is warm, so a `STATICCALL` costs the caller 100 gas, and the caller may forward 0 gas. Neither revm nor alloy-evm short-circuits a 0-gas precompile call, and alloy-evm hands the precompile a borrowed slice of shared memory (no copy on the way in).
- `signature` and `pubkey` are unbounded `bytes` in the ABI. Their 48 / 96 length gate is inside `bls_verify`, after the decode. So the 4096-byte `message` cap did not bound the copy: a multi-MB `signature` with a 119-byte message decodes in full and only then fails the exact gas gate.
- Memory expansion is charged once, and the decoder accepts overlapping offsets, so a loop can re-drive the same 2.27 MB region with all three fields aliased. Model from the gas constants: about 140 gas per iteration, about 142k iterations in a 30M-gas block, hundreds of GB of unmetered memcpy per block on every executing node. That is seconds of engine CPU per block for at most 30M gas at the base fee. The engine executes serially behind an 8-deep output queue, so sustained blocks of this shape backpressure into consensus delivery.

The work is transient and gas-bounded per block; there is no state corruption. This is a cost-asymmetry / liveness hardening item (Medium, borderline Low, in the issue). Batch validators do not execute transactions, so only executing nodes pay.

## Root cause

PR #1051 replaced the flat `if gas_limit < BLS_VERIFY_GAS_COST { OutOfGas }` gate with the per-word charge and moved the decode ahead of the gas check so the message length is known before pricing. The exact charge needs the decode; the O(1) floor does not. Dropping the floor left the decode as the one O(calldata) step ahead of any charge.

## Fix

Gate twice, cheapest first (the shape every pinned eth precompile uses: `identity`, `sha256`, `ripemd160` all check an O(1) cost before their O(len) work):

- [x] `handle_bls_verify`: `(gas_limit >= BLS_VERIFY_GAS_COST).then_some(()).ok_or(PrecompileError::OutOfGas)?;` is the first statement, ahead of `abi_decode_raw`. The base charge is the least any accepted call pays, so the floor is attacker-independent and derived from the existing constant, not a new number.
- [x] The exact post-decode charge (`MAX_BLS_VERIFY_MESSAGE_LEN` cap, then `bls_verify_gas_cost(len)`, then `gas_limit >= cost`) is unchanged.
- [x] Doc comments on `handle_bls_verify` and `BLS_VERIFY_GAS_COST` describe the two-gate ordering and why the floor exists.

Behavior: every accepted call already had `gas_limit >= cost >= BLS_VERIFY_GAS_COST`, so the Ok-set and `gas_used` are byte-identical. The only observable change is for calls with `gas_limit < 150_000` whose calldata is malformed or carries an over-cap message: they now fail as `OutOfGas` instead of `Other`. Both are the same error class to the EVM (`return_error!`: the caller frame gets `0`, and the forwarded gas is consumed), so state, gas, and stack are indistinguishable on chain. `ConsensusRegistry` (fixed 48 / 96 / 119 lengths, ample gas) is unaffected.

Not done, on purpose: no raw-calldata length ceiling before the decode. A total-length cap is not behavior-preserving (a canonical call with an oversized `pubkey` returns `Ok(false)` today, and the decoder ignores trailing bytes), and the floor alone brings the residual down to at most about 132 decodes per 30M block, the same class as `MCOPY`-metered work.

## Testing

- New `dispatch_gas_floor_precedes_decode` (unit, `evm::bls_precompile`): malformed calldata (head words of `0xff`) at gas `0` and at `BLS_VERIFY_GAS_COST - 1` is `OutOfGas`; the same bytes at exactly `BLS_VERIFY_GAS_COST` reach the decoder and fail as `Other` (positive control, which also pins the `>=` boundary); a 1 MiB over-cap message at gas `0` is `OutOfGas`, and at the floor it is the cap's `Other`.
- `dispatch_verify_out_of_gas` gained a guard that its `cost - 1` budget clears the floor, so it keeps pinning the exact post-decode gate on its own.
- Mutation-confirmed with both mutants taken from the diff: (A) floor removed (the origin/main ordering) and (B) `>=` flipped to `>`. Each fails exactly `dispatch_gas_floor_precedes_decode`; bytes restored, suite green again.
- Commands (pinned toolchain, `-j 2`): `cargo +1.94 test -p tn-reth --lib evm::bls_precompile` (15 passed, 0 failed) and `cargo +1.94 test -p tn-reth --test it -- bls_precompile` (11 passed, 0 failed; the EVM-level proptests and the `STATICCALL` relay path). `cargo +nightly fmt --all -- --check` and `cargo +nightly clippy -p tn-reth --all-features --all-targets -- -D warnings` are clean.
- Well-formed under-cap inputs are observationally identical under both orderings (only the work differs), so the discriminating inputs are exactly malformed calldata and over-cap messages under the floor. The bench numbers for the sig-heavy payload are in #1211.

### DIFF
--- a/crates/tn-reth/src/evm/bls_precompile/mod.rs
+++ b/crates/tn-reth/src/evm/bls_precompile/mod.rs
@@ -71,6 +71,10 @@ sol! {
 ///
 /// The length-dependent portion of hash-to-curve is charged separately by
 /// [`BLS_VERIFY_PER_WORD_GAS_COST`]; the total for a given message is [`bls_verify_gas_cost`].
+///
+/// The base is also the O(1) gas floor that [`handle_bls_verify`] checks before it decodes the
+/// calldata: no accepted call pays less than this, so a call that cannot fund it is refused before
+/// any of its `bytes` arguments are copied.
 const BLS_VERIFY_GAS_COST: u64 = 150_000;
 
 /// Per-32-byte-word surcharge over the hashed `message`.
@@ -140,13 +144,28 @@ fn dispatch(data: &[u8], gas: u64) -> PrecompileResult {
 
 /// `blsVerify(bytes,bytes,bytes) -> bool`.
 ///
-/// Decoding precedes the gas gate so the message length - which determines the charge - is known
-/// before pricing (the decode is cheap; the previous early gate was only a coarse floor). The
-/// message is capped at [`MAX_BLS_VERIFY_MESSAGE_LEN`] and then charged [`bls_verify_gas_cost`]
-/// (base plus a per-word surcharge), so a caller supplying a long message pays for the extra
-/// hash-to-curve work it induces instead of riding the flat base. An over-cap message is rejected
-/// as `Other`; an under-funded call is `OutOfGas`.
+/// Gas is gated twice, cheapest first.
+///
+/// 1. An O(1) floor of [`BLS_VERIFY_GAS_COST`] runs before the calldata is decoded. The base is the
+///    least any accepted call pays, so a call that cannot fund it is refused before its three
+///    `bytes` arguments are copied out of calldata. That copy is the only O(calldata) work in the
+///    precompile that no earlier charge covers: the precompile address is warm, and a `STATICCALL`
+///    can reach it with 0 forwarded gas, so without the floor a warm call's 100 gas base buys an
+///    arbitrarily large memcpy on every executing node (the `signature` and `pubkey` fields are
+///    unbounded `bytes`, so the message cap alone does not bound the copy).
+/// 2. The decode then yields the message length, which fixes the exact charge. The message is
+///    capped at [`MAX_BLS_VERIFY_MESSAGE_LEN`] and charged [`bls_verify_gas_cost`] (base plus a
+///    per-word surcharge), so a caller supplying a long message pays for the extra hash-to-curve
+///    work it induces instead of riding the flat base.
+///
+/// An over-cap message is rejected as `Other`; an under-funded call is `OutOfGas` at whichever
+/// gate it fails. Every accepted call clears both gates, so the floor changes neither the accepted
+/// set nor `gas_used`. The one observable difference is that an under-funded call whose calldata
+/// is malformed or carries an over-cap message is now `OutOfGas` instead of `Other`; the EVM treats
+/// both the same way (the caller frame gets `0` and its forwarded gas is consumed).
 fn handle_bls_verify(calldata: &[u8], gas_limit: u64) -> PrecompileResult {
+    (gas_limit >= BLS_VERIFY_GAS_COST).then_some(()).ok_or(PrecompileError::OutOfGas)?;
+
     let decoded = blsVerifyCall::abi_decode_raw(calldata)
         .map_err(|e| PrecompileError::Other(format!("blsVerify: {e}").into()))?;
 
@@ -367,13 +386,44 @@ mod tests {
         assert_eq!(out.gas_used, expected);
     }
 
-    /// Verification with less gas than the length-dependent cost is metered as out-of-gas.
+    /// Verification with less gas than the length-dependent cost is metered as out-of-gas. The
+    /// budget here (`cost - 1`) clears the O(1) floor and fails only the exact post-decode charge,
+    /// so this pins the second gate independently of the first.
     #[test]
     fn dispatch_verify_out_of_gas() {
         let v = vector(7, 0x42);
         let cost = bls_verify_gas_cost(v.message.len()).expect("cost fits u64");
+        assert!(cost > BLS_VERIFY_GAS_COST, "budget must clear the floor to reach the exact gate");
         let res = dispatch(&encode_verify(&v.sig, &v.pubkey, &v.message), cost - 1);
         assert!(matches!(res, Err(PrecompileError::OutOfGas)));
+    }
+
+    /// The O(1) gas floor is checked before the calldata is decoded. Under-funded calls are
+    /// `OutOfGas` even when the calldata would not decode or carries an over-cap message, and the
+    /// same bytes with the floor met reach the decoder and fail as `Other`. Restoring decode-first
+    /// flips the under-funded cases to `Other`, so this pins the ordering.
+    #[test]
+    fn dispatch_gas_floor_precedes_decode() {
+        // Three head words of 0xff: offsets far past the end of the data, so decoding fails.
+        let malformed = [blsVerifyCall::SELECTOR.as_slice(), [0xffu8; 96].as_slice()].concat();
+        assert!(matches!(dispatch(&malformed, 0), Err(PrecompileError::OutOfGas)));
+        assert!(matches!(
+            dispatch(&malformed, BLS_VERIFY_GAS_COST - 1),
+            Err(PrecompileError::OutOfGas)
+        ));
+        // Positive control: with the floor met, the decoder runs and rejects the bytes, so the
+        // `OutOfGas` above comes from the floor and not from a decoder that never saw the input.
+        assert!(matches!(
+            dispatch(&malformed, BLS_VERIFY_GAS_COST),
+            Err(PrecompileError::Other(_))
+        ));
+
+        // An over-cap 1 MiB message under the floor is refused by the floor, not by the cap that
+        // runs after the decode; with the floor met the same call reaches the cap.
+        let big_message = vec![0u8; 1 << 20];
+        let over = encode_verify(&[0u8; 48], &[0u8; 96], &big_message);
+        assert!(matches!(dispatch(&over, 0), Err(PrecompileError::OutOfGas)));
+        assert!(matches!(dispatch(&over, BLS_VERIFY_GAS_COST), Err(PrecompileError::Other(_))));
     }
 
     /// `bls_verify_gas_cost` is the flat base plus the per-word surcharge, rounding the message up


grantkee (approved):
LGTM!

sstanfield (approved):

Haxatron:
Fix Review: Currently we check the floor gas constant before the ABI decode allocations. As the allocation work is dynamic in cost it is not sufficient to check against a constant. A better way is to check the length of the raw calldata and the gas before the ABI decode as follows:
```
    let message_len = calldata.len();
    let cost = (message_len <= MAX_BLS_VERIFY_MESSAGE_LEN)
        .then_some(message_len)
        .ok_or_else(|| PrecompileError::Other("blsVerify: message exceeds maximum length".into()))
        .and_then(|len| bls_verify_gas_cost(len).ok_or(PrecompileError::OutOfGas))
        .and_then(|cost| (gas_limit >= cost).then_some(cost).ok_or(PrecompileError::OutOfGas))?;
```

Filed issue #1332 for this.

PR #1216, opened 2026-08-18
Add a test to codify and confirm that network events are not lost between epochs.

sstanfield:
Unit test that verifies and makes sure we don't loose the ability for network events to survive across epoch boundaries.


grantkee (approved):
LGTM

MavenRain (approved):
Cool!

I like the idea of encoding more of what we believe to be true about the system into tests.  I think it'd be great if we did more of this.

PR #1217, opened 2026-08-19
Tighten epoch vote gossip to only allow votes for expected EpochRecords.

sstanfield:


MavenRain:
Non-blocking observation on fork visibility.  The alt-record branch in `manage_epoch_votes` used to give a node with a divergent local record a fast signal: alt votes were tallied, and at quorum it logged "Reached quorum on epoch record ... instead of ..." and broke out early into recovery.  Digest-mismatched votes now stop in the handler, so that branch is starved (the new comment already calls it defense-in-depth).  A forked node discovers the divergence only after the full timeout ladder (26 x 2.5s = 65s) and then recovers through `request_epoch_cert`.  Recovery still works; the operator-facing signal is what got quieter.

Suggestion: the handler is now the one place that still sees mismatched digests, and `InvalidEpochVote` already carries the expected and received digests.  A counter or a rate-limited `warn!` there would restore the fork signal the collector lost.

MavenRain:
Note on the admission window at the epoch boundary.  The old gate resolved membership from the live current/next committee, so peer votes were admitted before our own record existed and were buffered by the collector's one-epoch-ahead queue.  The new gate requires our stored uncertified record, which exists only after our own `close_epoch` saves it (save, persist, arm collector, publish own vote).  Votes from faster peers that arrive before that save are dropped with a `debug!` and come back only on the authors' 2.5s republish timer.  A lagging node's boundary pays at least one republish interval; the worst case degrades to the 65s timeout plus the cert download path.  Bounded and self-healing, so latency-only, but worth a sentence in the PR description so the narrower window is on record as intended.

MavenRain:
Small doc drift from the dedup key change.  The key is now `(author, epoch)`, but two comments still describe the old shape:

- the `record_epoch_vote` doc: "Record a verified `(author, epoch, epoch-record)` vote"
- the gate comment: "for this (author, record)"

MavenRain:
Test suggestion.  `test_epoch_vote_wrong_digest_rejected_before_verify` covers the digest arm well.  The other half of the gate, `Some((record, Some(cert)))` for an already-certified epoch, has no test.  A companion that saves a record plus its certificate and asserts the vote is dropped without an error or penalty would pin the certified branch.

MavenRain (approved):
Several suggestions . . . nothing blocking

sstanfield:
> Note on the admission window at the epoch boundary. The old gate resolved membership from the live current/next committee, so peer votes were admitted before our own record existed and were buffered by the collector's one-epoch-ahead queue. The new gate requires our stored uncertified record, which exists only after our own `close_epoch` saves it (save, persist, arm collector, publish own vote). Votes from faster peers that arrive before that save are dropped with a `debug!` and come back only on the authors' 2.5s republish timer. A lagging node's boundary pays at least one republish interval; the worst case degrades to the 65s timeout plus the cert download path. Bounded and self-healing, so latency-only, but worth a sentence in the PR description so the narrower window is on record as intended.

This is a deliberate tradeoff.  Next PR takes steps to mitigate but I did it this way to greatly simplify bookkeeping and grief/DOS edge cases.  In short there are enough small delays and republishes to mitigate this.  This is not timing critical either now that nodes following the tip do not require certs.

sstanfield:
> Non-blocking observation on fork visibility. The alt-record branch in `manage_epoch_votes` used to give a node with a divergent local record a fast signal: alt votes were tallied, and at quorum it logged "Reached quorum on epoch record ... instead of ..." and broke out early into recovery. Digest-mismatched votes now stop in the handler, so that branch is starved (the new comment already calls it defense-in-depth). A forked node discovers the divergence only after the full timeout ladder (26 x 2.5s = 65s) and then recovers through `request_epoch_cert`. Recovery still works; the operator-facing signal is what got quieter.
> 
> Suggestion: the handler is now the one place that still sees mismatched digests, and `InvalidEpochVote` already carries the expected and received digests. A counter or a rate-limited `warn!` there would restore the fork signal the collector lost.

This is by design.  Nodes that are forked are cooked and can not recover.  Doing a bunch of extra work to detect this early seems like unnecessary complexity.  A forked node will likely crash before long.  The next PR makes this forking explicit and brings the node down (if it managed to still be running).  The idea that a node can have a "oopsie" and generate the wrong epoch record is simply incorrect and I don't think we should code for that anymore.

grantkee (approved):
LGTM

PR #568, opened 2026-02-20
tests: Multi-Worker Support - PR 6: Tests

TanguyDeTaxis:
## Summary
- Add worker-topic isolation and worker-id validation coverage for multi-worker behavior.
- Branch: `stack/559-multi-worker-tests`

Closes #559

PR #563, opened 2026-02-20
feat: Multi-Worker Support - PR 1: Configuration

TanguyDeTaxis:
## Summary
- Lay the multi-worker configuration foundation while keeping default behavior compatible.
- Branch: `stack/554-config-foundation`

Closes #554

Issue #559, opened 2026-02-14
tests: Multi-Worker Support — PR 6: Tests


grantkee:
## Problem

The node currently hardcodes a single worker per validator (`worker_id = 0`). The `EpochManager`, network layer, and several initialization paths assume exactly one worker. This blocks the ability to run independent fee markets, specialized transaction pools, or any form of worker-level parallelism.

## Goal

Refactor the node to support **N independent workers per validator**. Each worker operates as a standalone unit with its own:

- libp2p swarm (dedicated gossip topics, listen address, network key)
- RPC server (unique port)
- Transaction pool
- Batch builder + batch validator
- `LocalNetwork` instance for primary communication

Workers share only the **Primary** (consensus) and the **execution engine** (block production). The `num_workers` count is a consensus-level parameter — all validators must agree on it.

## Why

The immediate motivation is **multiple fee markets**. Once multi-worker is in place, a follow-up (Phase 2) spawns 2 workers by default:

- **Worker 0 (General):** accepts all transactions, standard EIP-1559 fee market
- **Worker 1 (Whitelisted Transfers):** accepts only whitelisted ERC-20 `transfer`/`transferFrom` calls, operates with a reduced base fee

This architecture also enables future process separation — workers can be extracted into standalone processes communicating with the primary over RPC.

## Design Constraints

1. **Workers are fully independent** — no cross-worker shared state. Each worker has its own network identity, pool, and gossip topics.
2. **Per-worker gossip topics** — `tn-worker-{id}` and `tn-txn-{id}` replace the current global `tn-worker` and `tn-txn` topics. This provides network-level isolation.
3. **Per-worker `LocalNetwork`** — each worker gets its own `LocalNetwork` instance for primary communication. The primary registers as the handler on every worker's `LocalNetwork`. This is the seam for future process separation.
4. **`num_workers` is a consensus parameter** — changing it requires a coordinated upgrade across all validators. Defaults to `1` for backward compatibility.
5. **Execution engine is shared** — batches from all workers are processed sequentially by the same engine. Worker ID is already encoded in the block `difficulty` field.
6. **Faucet on worker 0 only** — the testnet faucet attaches to the general-purpose worker.

## Current State

Much of the infrastructure already supports N workers but is only called with `worker_id = 0`:

- `ExecutionNodeInner.workers: Vec<WorkerComponents>` — vec exists, only 1 element
- `GasAccumulator` — supports N workers internally, but initialized with `new(1)`
[...]

PR #1219, opened 2026-08-19
feat(config,types): multi-worker configuration foundation (#554, PR 1)

MavenRain:
Refs #554 (PR 1 of the multi-worker series; the earlier stack #563-#568 was closed for drift).

## Summary

Lays the configuration and type foundation for N workers per validator while keeping every default at one worker. Data structures that hard-wired a single worker (`NodeP2pInfo.worker`, `BootstrapServer.worker`, the global batch topic, `Committee::number_of_workers() { 1 }`) become per-worker collections and a committee-carried count. No runtime path spawns more than `DEFAULT_WORKER_ID` yet: that is #555-#558.

Two points where this PR departs from the issue text, both because of what landed since it was written:

- **No `Parameters.num_workers`.** Since #844/#846 the worker count's source of truth is the on-chain `WorkerConfigs` contract, read at the previous epoch's closing block. Adding a second, operator-editable copy in the node config would create exactly the desync the issue's constraint 4 warns about. The committee reads the same on-chain value instead (grantkee's suggestion on #554).
- **`worker_txn_topic` no longer exists on main**; only the batch topic is per-worker'd.

## Changes

- [x] `crates/types/src/primary/info.rs` . . . `NodeP2pInfo { primary, workers: Vec<P2pNode> }` with `new(primary, workers)`, `worker(id) -> Option<&P2pNode>`, `worker_mut(id)`, `num_workers()`. Deserialization accepts BOTH the legacy `worker: {..}` map and the new `workers: [..]` list (untagged helper + `#[serde(from)]`); serialization writes the new shape only; an empty `workers` list is a serde error. `Default` stays one worker.
- [x] `crates/types/src/committee.rs` . . . `BootstrapServer { primary, workers }` with the same legacy-compatible deserialize and `worker(id)` / `num_workers()`. The legacy fallback is an untagged enum, which needs `deserialize_any`, so it only runs when the deserializer `is_human_readable()` (YAML / JSON files); bcs (consensus store, consensus pack) reads the current shape directly. `CommitteeInner` gains `num_workers: NonZeroUsize` (`serde(default)` = 1 so existing committee files load unchanged; part of `PartialEq`). `Committee::number_of_workers()` returns it instead of the constant; `Committee::with_num_workers(n)` rebuilds the inner. `CommitteeBuilder::with_num_workers(n)`; `add_bootstrap_server` / `add_authority_and_bootstrap` take the worker list.
- [x] `crates/types/src/primary/header.rs` . . . `Header::validate` compares `usize::from(*worker_id) >= committee.number_of_workers()` (drops the naked `as usize`); the `UnkownWorkerId` rejection is unchanged.
- [x] `crates/config/src/genesis.rs` . . . `NodeInfo::worker_network_key(id)` / `worker_network_address(id)` return `Option`, plus `worker_p2p_nodes()` and `num_workers()`. `NetworkGenesis::validate()` now also requires every validator to declare the SAME non-zero worker count (`agreed_num_workers`), and `create_committee` passes the full worker lists and stamps that agreed count on the genesis committee.
- [x] `crates/config/src/network.rs` . . . `LibP2pConfig::worker_batch_topic(chain_id, worker_id)` returns `tn-worker-{chain_id}-{worker_id}`.
- [x] `crates/config/src/{node.rs,consensus.rs}` . . . `Config::update_worker_network_key(worker_id, key)` (error on a missing worker) and `ConsensusConfig::worker_address(worker_id) -> Option<Multiaddr>`.
- [x] `crates/node/src/manager/node.rs` . . . the count read is factored out of `sync_num_workers_from_chain` into `read_num_workers_at_epoch_entry(reth_env, epoch_first_block)` (same pin, same `retry_provider_faults`, same halt-on-failure policy); `sync_num_workers_from_chain` calls it, behaviour unchanged.
- [x] `crates/node/src/manager/node/start_epoch.rs` . . . `create_committee_from_state` reads the on-chain count through that same helper for BOTH arms (epoch 0: file-loaded committee `.with_num_workers(n)`; later epochs: `CommitteeBuilder::with_num_workers(n)`), using `epoch_info.blockHeight` exactly as the accumulator path does. A failed read or a zero count halts epoch entry (no fail-open arm). Worker address / bootstrap-peer / topic sites use `DEFAULT_WORKER_ID`.
- [x] `crates/consensus/worker/src/network/{handle.rs,handler.rs}` . . . `WorkerNetworkHandle` carries its `worker_id` (new ctor parameter, `worker_id()` accessor) so publish and the gossip handler agree on the per-worker topic.
- [x] `crates/test-utils-committee/src/{builder.rs,authority.rs}` . . . `Builder::number_of_workers(n)`; the `assert_eq!(number_of_workers.get(), 1)` is gone; each authority gets `n` worker `P2pNode`s (worker 0 keeps the `KeyConfig` key, extra fixture workers get fresh keys); the fixture committee is stamped with the count.
- [x] `chain-configs/{mainnet,testnet}/committee.yaml` . . . `worker:` -> `workers: [ ... ]`, one element each, values unchanged (loads identically under the legacy path too).
- [x] Worker-0 accessor migration at every other `p2p_info.worker` site: `keytool/{generate,set_rpc,mod}.rs`, `e2e-tests/tests/it/common.rs`, `network-libp2p` and `worker` integration tests, `batch_fetcher.rs`, `worker.rs`.

## Consensus notes

- **The count that validates is the count that executes.** `Header::validate` (first statement of `vote_inner`, re-run on certificate acceptance) rejects any payload `worker_id >= committee.number_of_workers()`. Before this PR that bound was a compile-time 1; now it is the same `WorkerConfigs` read that sizes the `GasAccumulator` for the epoch, taken at the same block, so the id that a header is allowed to reference and the id execution can account for cannot disagree. Changing the count is still an on-chain, epoch-boundary event (`setNumWorkers`, effective next epoch), so it stays a coordinated upgrade as #554 constraint 4 requires.
- **Halt, do not fail open.** If the count cannot be read at epoch entry the node halts instead of entering the epoch with a guessed committee, matching the existing accumulator policy. A zero count is rejected explicitly.
- **Wire compatibility.** The worker batch topic string changes for worker 0 (`tn-worker-{chain}` -> `tn-worker-{chain}-0`), so nodes on either side of this change do not share a batch mesh. This is a coordinated-upgrade change, same class as the chain-id namespacing that preceded it. If a rolling upgrade matters more than a uniform scheme, worker 0 can keep the legacy string; say so and it is a one-line change plus test.
- **Storage compatibility.** The bcs layout of `Committee` changes (`BootstrapServer.worker` becomes a length-prefixed list, `CommitteeInner` gains a trailing `num_workers`). `EpochMeta` records written by earlier builds (consensus pack, consensus store) will not decode under this layout and vice versa. Fresh round-trips are covered by the tn-storage suites; existing on-disk packs are not. Two ways to take this: bump `PACK_VERSION` with a legacy `EpochMeta` decode (one more commit here or a follow-up), or treat it as part of the same coordinated upgrade as the topic change. Say which and I will do it.
- **Config compatibility.** Validator YAML (`p2p_info.worker`) and committee YAML (`bootstrap_servers.*.worker`, no `num_workers` key) written before this PR load unchanged; unit tests pin both legacy shapes against their one-element new-shape equivalents.

## Testing

Pinned toolchain from the worktree, `-j 2`, every step under `gateledger run -- tnquiet -- ...` (tests under `testledger`):

- `cargo +1.94 check --all-targets` on `tn-types tn-config tn-test-utils-committee tn-node tn-worker telcoin-network-cli e2e-tests tn-network-libp2p`: clean.
- `cargo +1.94 clippy --all-targets --no-deps -- -D warnings` on `tn-types tn-config tn-test-utils-committee tn-node tn-worker telcoin-network-cli` and `-p tn-network-libp2p --lib`: clean. (The e2e-tests and libp2p `--all-targets` test targets carry 25 pre-existing lints outside this diff.)
- `cargo +1.94 test`: `tn-types` `committee` + `primary::info` 31/0 (new: `NodeP2pInfo` / `BootstrapServer` round-trip, legacy-shape, empty-rejected, bcs round-trip of `BootstrapServer` and of a `Committee` with bootstrap servers; committee YAML without `num_workers` -> 1; round-trip with 2; `with_num_workers` keeps derived fields and accepts the default committee), `tn-config` 43/0 (new: genesis `validate` rejects mismatched worker counts, `create_committee` stamps the agreed count; topic test updated), `telcoin-network-cli::keytool` 20/0, `tn-worker --test it` 25/0, `tn-node` lib 61/0.
- Mutation-confirmed the five new load-bearing tests (each seen to FAIL under an in-diff mutant, then restored): legacy arm -> empty list; non-empty check disabled; genesis mismatch predicate disabled; serde default -> 2; `with_num_workers` re-running `load()`.
- Multi-agent review over the staged diff (8 raw findings, 1 high refuted on escalation); the four actionable ones are folded in: `with_num_workers` no longer re-runs `load()` (a missing epoch-0 committee file no longer panics), one shared serde shim for both structs, fixture `node_info` workers are the same `P2pNode`s the committee advertises, and a `warn!` when the on-chain count is above 1 (this node version spawns only worker 0).
- Pinned `+1.94` attest on the second machine caught two things the scoped local gates missed, both fixed: a lib-only unused import (`DEFAULT_WORKER_ID` used only under `cfg(test)`), and 20 tn-storage / tn-primary failures from the untagged serde shim under bcs (see the `committee.rs` bullet); `tn-storage` lib (207/0, covers the 18 `consensus` / `consensus_pack` / `pack_validate` failures) and both `tn-primary` restart tests re-run green after the fix.
- Not run locally: e2e suite.

## Out of scope (next PRs in the series)

- Per-worker swarms, listen addresses, RPC ports and `KeyConfig` worker keys (#555); per-worker `LocalNetwork` (#556); the `EpochManager` spawn loop and `Vec<WorkerNetworkHandle>` (#557); engine initialization (#558); multi-worker tests (#559).
- `RpcNodeInfo` (tn-rpc) still advertises the single worker key/address; it maps to worker 0 for now.

### DIFF
--- a/chain-configs/testnet/committee.yaml
+++ b/chain-configs/testnet/committee.yaml
@@ -21,34 +21,34 @@ bootstrap_servers:
     primary:
       network_address: /ip4/35.203.41.59/udp/49590/quic-v1/p2p/12D3KooWCnizjv4ctsBTwwYJxZP3duCuoebJfbk1fUQv6Vg1dVUh
       network_key: 4XTTM47Qeb4XhWEK7ybzj5bGsnVraHKaknpMFPgpg8jHYbGFo
-    worker:
-      network_address: /ip4/35.203.41.59/udp/49594/quic-v1/p2p/12D3KooWHGdbYerasxspuhZW4KqNMbbpqvLNCG6LGpS5yx6b8x2x
-      network_key: 4XTTM8bKFPoKfVL1UwN1vBMjCWCFVKbKpKUha12qr2Bi86ip4
+    workers:
+      - network_address: /ip4/35.203.41.59/udp/49594/quic-v1/p2p/12D3KooWHGdbYerasxspuhZW4KqNMbbpqvLNCG6LGpS5yx6b8x2x
+        network_key: 4XTTM8bKFPoKfVL1UwN1vBMjCWCFVKbKpKUha12qr2Bi86ip4
   rZdhyVGJAtz7kyFtFARWVVcw5DB3o9E4GvPN6zuKfUhUBGFRvEK2PgEV44CJt9wu9GMyssFQBFXRknrenuhJixo9KCADrZzHdtcvAwAYDmcLc7Ugh9epGc5LRZC5TK2hnYA:
     primary:
       network_address: /ip4/34.105.191.161/udp/49590/quic-v1/p2p/12D3KooWBRH7XpYhxCPTrw364UnmaSYDR4hCotoTHxTWTjKfrcUm
       network_key: 4XTTM2jxmNy1nZZX7tbVWBWgbj3Bstjgew7Qh2AsGVxwCpPFs
-    worker:
-      network_address: /ip4/34.105.191.161/udp/49594/quic-v1/p2p/12D3KooWFf6gRgYGnob7zhSZEnswRAHs4tSuh9UZXc5HzeKNut14
-      network_key: 4XTTM6ynLGq1MQAin2MtyMpmmZkwXYZSMpN5oFpV42svusenA
+    workers:
+      - network_address: /ip4/34.105.191.161/udp/49594/quic-v1/p2p/12D3KooWFf6gRgYGnob7zhSZEnswRAHs4tSuh9UZXc5HzeKNut14
+        network_key: 4XTTM6ynLGq1MQAin2MtyMpmmZkwXYZSMpN5oFpV42svusenA
   24JaKrebALhvhXxsAabjd5ZK22k1W4zKwGepKD5pcobQwBkZyiruYyhRM3hkLap8Ssc4qUdcRUJSrs2389hUoV7G83HKYrmkz2pTzCJiQURV5vtkrh8gFRKBQWKFTWLWah9s:
     primary:
       network_address: /ip4/34.102.122.126/udp/49590/quic-v1/p2p/12D3KooWSwK9yPLNtDr116cX74FS4Ehs2Nx2fWAUjpnsGrNkNGKd
       network_key: 4XTTMJFzopXoTVayf2m4wE69GCqMXW3wUnimiU3CdK5zHL36j
-    worker:
-      network_address: /ip4/34.102.122.126/udp/49594/quic-v1/p2p/12D3KooWESQQ5KghAFnUwWJq7niJuVjrLDooNzypjmRfwsySMXhC
-      network_key: 4XTTM5m63vU9mmcv8yAmFEpc946PWotoFWDb4TyqRz7ayKJUJ
+    workers:
+      - network_address: /ip4/34.102.122.126/udp/49594/quic-v1/p2p/12D3KooWESQQ5KghAFnUwWJq7niJuVjrLDooNzypjmRfwsySMXhC
+        network_key: 4XTTM5m63vU9mmcv8yAmFEpc946PWotoFWDb4TyqRz7ayKJUJ
   26L1XwWTr6bBNnxJmDJXJgTqMAyD4ExkqwPYYJqDD3gob3jzejypQoewqRmqb9D2Yudk7zjcPqaTzFcEtXhrYXUtNx2ZYeLKC8V77LL2J136Hvw8eCGXpadGAbymThMhWYAa:
     primary:
       network_address: /ip4/34.141.203.172/udp/49590/quic-v1/p2p/12D3KooWA2LvcWr3dvLx3CFnayoY4Q1mUX1qC2RuA2iAbUXjCpjM
       network_key: 4XTTM1M2aTfK8FHUc4riCi1hNCzfRxC1HKF38tF7vdi9GAbWT
-    worker:
-      network_address: /ip4/34.141.203.172/udp/49594/quic-v1/p2p/12D3KooWDWAL8TE8PikV2Jxt3WSEkXCKcDKipDdVsFJVJMMxEbGz
-      network_key: 4XTTM4pqyybhD15t93yRJAYL4u7qz5tKAwSEjbTiFLayVCN46
+    workers:
+      - network_address: /ip4/34.141.203.172/udp/49594/quic-v1/p2p/12D3KooWDWAL8TE8PikV2Jxt3WSEkXCKcDKipDdVsFJVJMMxEbGz
+        network_key: 4XTTM4pqyybhD15t93yRJAYL4u7qz5tKAwSEjbTiFLayVCN46
   26L3gmkBzxAxEwvVuDrxEmKf2XwYGnYj1EQSiZSX3PBAvRa6Z5tegfmDAGHTsBwdtjFuAd7ayQEemE1AUE5ajSSK8wn3uz1MY5vjbGdeh3RoUv1TKLfbNTTULq9qdtaSTHhL:
     primary:
       network_address: /ip4/34.40.242.29/udp/49590/quic-v1/p2p/12D3KooWMUuChs1v4umTNiHivmD9G764hbcTTooRrMUpLZ9Tgpn5
       network_key: 4XTTMCoarZ1UzgGu7QNk93o6yQhjjBGbub2QfaZtaNnkzebZB
-    worker:
-      network_address: /ip4/34.40.242.29/udp/49594/quic-v1/p2p/12D3KooWE8p62tGetgUeJbKKyctPB2ky1zpahA6wPVr18R7iBtLB
-      network_key: 4XTTM5TVjt2jjW3cJLFmk6enDKdQdVfp2pNiB7iFmAejF9f7H
+    workers:
+      - network_address: /ip4/34.40.242.29/udp/49594/quic-v1/p2p/12D3KooWE8p62tGetgUeJbKKyctPB2ky1zpahA6wPVr18R7iBtLB
+        network_key: 4XTTM5TVjt2jjW3cJLFmk6enDKdQdVfp2pNiB7iFmAejF9f7H
--- a/crates/node/src/manager/node/start_epoch.rs
+++ b/crates/node/src/manager/node/start_epoch.rs
@@ -24,7 +24,7 @@
 //! yet executed is replayed to the engine, with a guard that refuses to cross an
 //! epoch boundary.
 
-use super::run_epoch::retry_provider_faults;
+use super::{read_num_workers_at_epoch_entry, run_epoch::retry_provider_faults};
 use crate::{
     engine::ExecutionNode, manager::EpochManager, primary::PrimaryNode, worker::WorkerNode,
     EngineToPrimaryRpc,
@@ -32,6 +32,7 @@ use crate::{
 use eyre::{eyre, OptionExt, WrapErr as _};
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
+    num::NonZeroUsize,
     sync::Arc,
     time::Duration,
 };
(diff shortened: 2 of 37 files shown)


chain-configs/testnet/committee.yaml
```diff
@@ -21,34 +21,34 @@ bootstrap_servers:
     primary:
       network_address: /ip4/35.203.41.59/udp/49590/quic-v1/p2p/12D3KooWCnizjv4ctsBTwwYJxZP3duCuoebJfbk1fUQv6Vg1dVUh
       network_key: 4XTTM47Qeb4XhWEK7ybzj5bGsnVraHKaknpMFPgpg8jHYbGFo
-    worker:
```
sstanfield, line 24: I think the old version of this file will be on disk for testnet not this version.  It looks like you may be handling the deserialize manually for this, just want to confirm this is true?

crates/node/src/manager/node/start_epoch.rs
```diff
@@ -335,27 +339,53 @@ where
...
     async fn create_committee_from_state(
         &self,
+        reth_env: &tn_reth::RethEnv,
         epoch: Epoch,
+        epoch_first_block: u64,
         validators: HashMap<BlsPublicKey, &ConsensusRegistry::ValidatorInfo>,
     ) -> eyre::Result<Committee> {
         info!(target: "epoch-manager", "creating committee from state");
 
+        let num_workers = read_num_workers_at_epoch_entry(reth_env, epoch_first_block)
```
sstanfield, line 357: I think this will fail at genesis when there are no blocks/state?

sstanfield (commented):
Couple of questions, probably ok but want to confirm before approval.

chain-configs/testnet/committee.yaml
MavenRain: Yeah, it's manual.


Evidence:

The deserialize is manual for human-readable formats: `BootstrapServer` (and `NodeP2pInfo`) go through `deserialize_primary_workers` in `crates/types/src/committee.rs`, an untagged enum that accepts either `workers: [..]` or the legacy `worker: {..}` map and turns the legacy map into a one-element list.  So, the testnet `committee.yaml` on disk loads unchanged, and the edit to the checked-in file is cosmetic (same values, new shape).  Serialization writes only the new shape.

Pinned by `bootstrap_server_legacy_shape_decodes_as_one_worker` and `legacy_shape_decodes_as_one_worker` (tn-types), which feed the literal `primary:/worker:` YAML text and assert one worker.

The one caveat is the binary path: bcs cannot drive an untagged enum, so for non-human-readable formats the shim reads the current shape directly.  That matters only for stored `Committee` values (consensus store, consensus pack), which are written by this same node and so carry the new shape.

crates/node/src/manager/node/start_epoch.rs
MavenRain: It should not, as that is the same read `run_epoch` already does at epoch 0 on main.


Evidence:

`read_num_workers_at_epoch_entry` is `sync_num_workers_from_chain`'s read factored-out, unchanged: `epoch_first_block.saturating_sub(1)`, so at epoch 0 it resolves block 0, which is the genesis header reth writes at chain init, and reads `WorkerConfigs` from genesis state (the contract is in the genesis alloc).  `run_epoch.rs` is untouched by this PR and its epoch-0 arm calls `sync_num_workers_from_chain(&reth_env, &gas_accumulator, epoch_info.blockHeight)` with the same argument I pass here, so if this read failed at genesis the accumulator sizing would already fail today.  `catchup_accumulator` at startup does the same.

`close_record_adjust_and_entry_read_agree_across_boundaries` (tn-node) drives that exact call with block 0 against a fresh temp chain that has only genesis and asserts the count.

What is new is only that a failed read now also halts committee construction rather than only accumulator sizing, and a zero count is rejected (`NonZeroUsize`).

MavenRain:
The binary format did change. As pushed, historic pack files would not be readable on this branch. I should have led with that instead leaving it buried it in the "Storage compatibility" bullet.

What changed, both inside `Committee`, both bcs-visible:

1. `BootstrapServer.worker: P2pNode` -> `workers: Vec<P2pNode>`: bcs writes a ULEB128 length prefix before the element, so one extra byte per bootstrap server.
2. `CommitteeInner` gains a trailing `num_workers: NonZeroUsize`: 8 extra bytes. `#[serde(default)]` only helps self-describing formats; bcs sees a short record and errors.

`EpochMeta` (first record of every consensus pack) embeds `Committee`, so a pack written by today's main does not decode on this branch, and vice versa. The YAML side is fine (legacy `worker:` and a missing `num_workers` both load, tests pin it); this is only the binary side.

I propose a `PACK_VERSION` bump: it is the mechanism that keeps the old packs readable after the layout change. Concretely, in this PR: `PACK_VERSION` 1 -> 2, and decode the `EpochMeta` of version <= 1 packs through the old `Committee` layout (`LegacyBootstrapServer { primary, worker }`, no `num_workers`, converted to the current type with `num_workers = 1`). The pack reader already dispatches on `self.version` (0 vs 1) for batches and output bytes, so this is one more arm plus a round-trip test that writes a v1 `EpochMeta` with the old byte layout and reads it back.

The alternative is to keep the bcs bytes identical (custom `Serialize` that writes the old shape when `!is_human_readable()` and skips `num_workers`). Then no version bump is needed in this PR, but the pack cannot carry the worker count, which PRs 2 to 5 need, so the bump moves to a later PR rather than going away. Happy to go either way; say which and I will push it.

grantkee:
> The binary format did change. As pushed, historic pack files would not be readable on this branch. I should have led with that instead leaving it buried it in the "Storage compatibility" bullet.
> 
> What changed, both inside `Committee`, both bcs-visible:
> 
> 1. `BootstrapServer.worker: P2pNode` -> `workers: Vec<P2pNode>`: bcs writes a ULEB128 length prefix before the element, so one extra byte per bootstrap server.
> 2. `CommitteeInner` gains a trailing `num_workers: NonZeroUsize`: 8 extra bytes. `#[serde(default)]` only helps self-describing formats; bcs sees a short record and errors.
> 
> `EpochMeta` (first record of every consensus pack) embeds `Committee`, so a pack written by today's main does not decode on this branch, and vice versa. The YAML side is fine (legacy `worker:` and a missing `num_workers` both load, tests pin it); this is only the binary side.
> 
> I propose a `PACK_VERSION` bump: it is the mechanism that keeps the old packs readable after the layout change. Concretely, in this PR: `PACK_VERSION` 1 -> 2, and decode the `EpochMeta` of version <= 1 packs through the old `Committee` layout (`LegacyBootstrapServer { primary, worker }`, no `num_workers`, converted to the current type with `num_workers = 1`). The pack reader already dispatches on `self.version` (0 vs 1) for batches and output bytes, so this is one more arm plus a round-trip test that writes a v1 `EpochMeta` with the old byte layout and reads it back.
> 
> The alternative is to keep the bcs bytes identical (custom `Serialize` that writes the old shape when `!is_human_readable()` and skips `num_workers`). Then no version bump is needed in this PR, but the pack cannot carry the worker count, which PRs 2 to 5 need, so the bump moves to a later PR rather than going away. Happy to go either way; say which and I will push it.

The epoch is knowable before the `Committee` bytes are read, on both paths:

- `EpochMeta.epoch` is serialized **first**, before `committee` (`consensus_pack.rs:50-55`), so it is readable in-stream.
- `Inner::stream_import` already receives `epoch: Epoch` as a parameter (`consensus_pack.rs:906-912`), so the receiver knows which epoch's pack it is importing before it decodes anything.
- Nested one level down, `CommitteeInner.epoch` is the second *serialized* field — `authorities`, then `epoch`, then `bootstrap_servers`, then `num_workers` (the three `#[serde(skip)]` fields are absent from the stream) — so both changed fields are decoded after the discriminator.

This matches the rule the existing precedent already states at `crates/types/src/forks.rs:230-233`: *"Callers MUST pass the epoch carried inside the value being encoded or decoded … never `Committee::epoch()` or other node-local state, so that historical values keep their historical layout at any nesting depth."*

With the gate in place: pre-fork epochs encode and decode legacy layout on both binaries, so packs flow in both directions during the rollout; post-fork epochs are new-layout only, and a straggler on an old build fails loudly and drops out — the same behavior `SEED_SIGNATURE_FORK_EPOCH` was designed for.

grantkee (approved):
LGTM

sstanfield (approved):

PR #1220, opened 2026-08-19
tn-reth: surface ConsensusRegistry logs from epoch-boundary system calls

### LINKED ISSUE #1218: tn-reth: epoch-boundary system calls discard ConsensusRegistry logs (NewEpoch, ValidatorActivated/Exited/Slashed never reach receipts)
## Summary

The epoch-boundary system calls discard every log the `ConsensusRegistry` emits. `transact_and_commit_system_call` (`crates/tn-reth/src/evm/block.rs:321-360`) reads only `gas_used` and `gas_refunded` from the `Success` arm, never builds a receipt for the call, and commits the state. `finish()` returns only `self.receipts`, so these logs are absent from the receipts root, the header `logs_bloom`, and `eth_getLogs`. They are visible only in the node's `trace!(?res)` line (block.rs:357).

This is the same treatment the EIP-4788 and EIP-2935 pre-block calls get (block.rs:1185-1229) and it matches Ethereum, where a system call produces no receipt. The difference is that those contracts emit nothing, while the registry emits several documented lifecycle events from its `onlySystemCall` paths.

## Events dropped

Reachable only through a system call (`tn-contracts/src/consensus/ConsensusRegistry.sol`):

| event | emit site | entry point |
|---|---|---|
| `NewEpoch` | L70 | `concludeEpoch` |
| `ValidatorActivated` | L674 (`_activate`) | `concludeEpoch` -> `_updateValidatorQueue` |
| `ValidatorExited` | L693 (`_exit`) | `concludeEpoch` -> `_updateValidatorQueue` |
| `ValidatorSlashed` | L132 | `applySlashes` |
| `ValidatorRetired` | L703 (`_retire`) | `applySlashes` -> `_consensusBurn` |
| `NextCommitteeSizeUpdated` | L765 | `applySlashes` -> `_ejectFromCommittees` |

`applyIncentives` emits nothing.

## Why it matters

`IConsensusRegistry.sol:86-99` documents `ValidatorPendingActivation` and `ValidatorPendingExit` as queue-entry events and `ValidatorActivated` / `ValidatorExited` as their resolution. The queue-entry events come from user transactions and are observable. The resolution events come from `concludeEpoch` and are not. A log-driven indexer, explorer, or staking UI sees every validator enter Pending and never leave it, and never sees an epoch boundary, a slash, or a retirement.

No consensus or safety impact: every node drops the same logs, so state and receipts roots agree. Severity informational.

## Options

1. **Document and accept.** Add a note to the `transact_and_commit_system_call` doc block and to the event natspec in `IConsensusRegistry.sol` that events emitted under `onlySystemCall` are not recorded on-chain. Zero risk; matches Ethereum system-call semantics. Indexers must read registry state (`getValidators`, `getCurrentEpochInfo`) instead of logs for these transitions.
2. **Surface them node-side, off-consensus.** Decode `res.result` logs in the `Success` arm and emit them at `info!`/`debug!` with the event name, or expose them over a `tn_` RPC / metrics, the way `epoch_close_gas` already is. Gives operators visibility without touching receipts.
3. **Record them on-chain.** Build a synthetic receipt for each system call (or one per boundary). Changes `receipts_root` and `logs_bloom`, so this is a hard fork and needs the same code-hash gate as the registry upgrade path.

Option 1 is the floor. Option 2 is cheap and independent of the fork decision. Option 3 is a product call on whether the lifecycle events are part of the public contract.

## Credit

Raised by shotes in the Cantina collaborative review (`block.rs:323`).

MavenRain:
## Summary

Surface the `ConsensusRegistry` logs that the epoch-boundary system calls emit. The node now decodes them and writes one named `info!` line per event; receipts, the header bloom, and `eth_getLogs` are unchanged. Option 2 of #1218. (Cantina #22)

## Problem

`transact_and_commit_system_call` reads only `gas_used` and `gas_refunded` from the `Success` arm and builds no receipt. `finish()` returns only `self.receipts`. The registry's `onlySystemCall` paths emit `NewEpoch`, `ValidatorActivated`, `ValidatorExited`, `ValidatorSlashed`, `ValidatorRetired`, and `NextCommitteeSizeUpdated`, and every one of them is dropped: the only trace was the `trace!(?res)` dump. An operator following the node log sees validators enter `PendingActivation` (a user transaction, so it has a receipt) and never sees the resolution.

Recording them on-chain would change `receipts_root` and `logs_bloom`, which is a hard fork and a separate product decision (option 3 in #1218). This PR keeps consensus untouched.

## Fix

- `system_calls.rs`: bind every event of `IConsensusRegistry.sol` in the `ConsensusRegistry` `sol!` block (the full interface, not only the system-call subset, so a future emit behind `onlySystemCall` decodes by name instead of falling through as an unknown topic). Add `log_registry_event`, an exhaustive match over `ConsensusRegistryEvents` that emits `info!(target: "engine", event = <name>, detail = ?payload, "<call> emitted registry event")`. Exhaustive on purpose: binding a new event without naming it is a compile error.
- `evm/block.rs`: the `Success` arm destructures `logs` and hands them to `surface_system_call_logs`. Each log is classified by `classify_system_call_log`: a registry log decodes to `SystemCallLog::Registry` and is emitted at `info!`; a registry log with an unknown topic, or a log from any other contract (the `WorkerConfigs` write emits `WorkerConfigUpdated`), is emitted raw at `debug!` with address, topics, and data. Nothing is dropped and nothing is promoted to a line that looks decoded.
- Doc blocks on `transact_and_commit_system_call` and the `system_calls` module state that system calls produce no receipt and that these lines are the only place the events surface.

Threat model: the new path is observational. It runs after the result is known `Success`, reads `res.result.logs`, and touches neither `res.state`, the gas accounting, nor the receipts vector. Epoch boundaries are rare and emit a bounded number of events (one per activation or exit plus `NewEpoch`), so `info!` cannot flood the log; a non-closing block never reaches the function.

## Testing

- `registry_event_selectors_match_the_solidity_interface` pins every bound event's topic-0 hash to the `tn-contracts` `artifacts/ConsensusRegistry.json` ABI signature and pins the row count to `ConsensusRegistryEvents::COUNT`, so ABI drift or an unpinned new event fails at test time instead of demoting every `NewEpoch` to "unknown topic" at runtime.
- `registry_log_decodes_to_named_event_and_rejects_unknown_topic` round-trips a `NewEpoch` through `encode_log_data` and `decode_log`, and checks that a foreign topic-0 is rejected rather than mis-decoded.
- `system_call_logs_classify_to_named_registry_events` runs `concludeEpoch` as a real system call against a consensus-registry genesis and asserts the emitted logs classify to `NewEpoch` with the expected epoch id and committee, with zero unknown or foreign logs.
- Static ladder: nightly fmt check, `check -p tn-reth --tests` on the pinned 1.94 toolchain, nightly clippy `-p tn-reth --all-features --all-targets --no-deps -D warnings`. The `tn-reth` crate test suite passes.

Closes #1218 (option 2). Option 1's natspec note in `IConsensusRegistry.sol` is a tn-contracts change and is not part of this PR.

### DIFF
--- a/crates/tn-reth/src/system_calls.rs
+++ b/crates/tn-reth/src/system_calls.rs
@@ -3,13 +3,20 @@
 //! These compile into types for interacting with smart contracts through
 //! System Calls.
 //!
-//! The `sol!` blocks bind three contracts from the tn-contracts submodule:
-//! - `ConsensusRegistry` — validator lifecycle (mint/stake/activate/exit/burn), epoch and committee
-//!   views, the three epoch-boundary mutators the protocol invokes as system calls in the closing
-//!   block (`applyIncentives`, then `applySlashes`, then `concludeEpoch`), and the contract's
-//!   events. System calls build no receipt, so the lifecycle events those mutators emit are
-//!   surfaced node-side only: [`log_registry_event`] names and emits each one at `info!` from the
-//!   system-call path in `evm/block.rs`.
+//! The `sol!` blocks bind four surfaces from the tn-contracts submodule. The registry is split
+//! across two of them — a hand-written CALL surface and a generated LOG surface — because the two
+//! have opposite requirements: calls need richer Rust types than a JSON ABI can express, logs need
+//! coverage a hand-written list cannot guarantee.
+//! - `ConsensusRegistry` — the hand-written call surface: validator lifecycle
+//!   (mint/stake/activate/exit/burn), epoch and committee views, and the three epoch-boundary
+//!   mutators the protocol invokes as system calls in the closing block (`applyIncentives`, then
+//!   `applySlashes`, then `concludeEpoch`). Curated, not exhaustive: it binds what the node calls
+//!   and decodes, in the Rust types the node needs.
+//! - `registry_abi` (crate-private; [`RegistryEvents`] is its public face) — the registry's log
+//!   surface, generated from the compiled Foundry artifact and therefore covering the contract's
+//!   WHOLE event ABI, inherited events included. System calls build no receipt, so the events the
+//!   boundary mutators emit are surfaced node-side only: [`log_registry_event`] names and emits
+//!   each one at `info!` from the system-call path in `evm/block.rs`.
 //! - `LegacyConsensusRegistry` — the pre-fork registry's epoch-close surface, frozen so pre-fork
 //!   epoch closes replay byte-identically (see the code-hash gate in `evm/block.rs`).
 //! - `WorkerConfigs` — per-worker base-fee strategy used for base fee adjustment, plus the
@@ -24,7 +31,10 @@
 //! NOTE: slashing is not live — the protocol calls `applySlashes` with an empty array. Do not
 //! assume validators can currently be slashed in-protocol.
 
-use alloy::{primitives::address, sol};
+use alloy::{
+    primitives::{address, Log},
+    sol,
+};
 use tn_types::{
     gas_accumulator::{WorkerConfigEntry, WorkerFeeConfig},
     Address, Epoch,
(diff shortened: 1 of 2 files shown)


crates/tn-reth/src/system_calls.rs
```diff
@@ -340,6 +432,47 @@ sol!(
...
+/// the decoded payload. Purely observational: nothing here reads or writes state.
+///
+/// Exhaustive over the events enum on purpose: binding a new event in the `sol!` block without
+/// naming it here is a compile error, not a silently unnamed log line.
+pub(crate) fn log_registry_event(
+    description: &str,
+    event: ConsensusRegistry::ConsensusRegistryEvents,
+) {
+    use ConsensusRegistry::ConsensusRegistryEvents as E;
+    let emit = |name: &'static str, detail: &dyn core::fmt::Debug| info!(target: "engine", event = name, ?detail, "{description} emitted registry event");
```
grantkee: nit: prefer generics over dynamic dispatch

```
let emit = |name: &'static str, detail: &T| where T: core::fmt::Debug { ... };
```

crates/tn-reth/src/system_calls.rs
```diff
@@ -340,6 +432,47 @@ sol!(
...
+    match event {
+        E::ValidatorStaked(e) => emit("ValidatorStaked", &e),
+        E::ValidatorPendingActivation(e) => emit("ValidatorPendingActivation", &e),
+        E::ValidatorActivated(e) => emit("ValidatorActivated", &e),
```
grantkee: Instead of typing, can the enum be used? This seems inefficient and fragile

also, this change requires a fork to preserve syncing

sstanfield (dismissed):

grantkee (approved):

PR #1221, opened 2026-08-20
fix(network): cap a kad provider record's address list

### LINKED ISSUE #1185: network-libp2p: kad add_provider does not cap a provider record's address list
## Summary

The kad `add_provider` path stores a provider record's address list verbatim with no cap on the
number of addresses. The `PutRecord` / `NodeRecord` path caps the advertised multiaddr count at
`MAX_ADVERTISED_MULTIADDRS` (16). Because the cap was not mirrored, one byzantine peer can write a
large volume of attacker-controlled address data into the consensus database through repeated
`AddProvider` requests.

This is a single-node disk and memory exhaustion vector. It is not a chain-halt.

- No-cap site (`add_provider`):
  `crates/network-libp2p/src/kad.rs:512-559`. The conversion
  `From<ProviderRecord> for KadProviderRecord` (`kad.rs:107-113`) copies the address list verbatim
  at `kad.rs:111` (`addresses: value.addresses`). Neither `add_provider` nor the `From` impl checks
  `value.addresses.len()`.
- Sibling cap on a different axis:
  `merge_provider` (`kad.rs:370-391`) caps `pruned.len() < self.config.max_providers_per_key` at
  `kad.rs:385`. This bounds the number of distinct providers per key, not the number of addresses
  per provider.
- Contrast anchor (the cap that has no analogue here):
  `crates/network-libp2p/src/consensus.rs:616`, the per-record multiaddr cap on `NodeRecord`.

## Root cause

A per-record address cap was applied to `NodeRecord` and was not mirrored onto `ProviderRecord`.
The provider path bounds the provider count per key and the total provider-key count, but it does not
bound the address-list length inside a single provider record.

## Threat model

A byzantine peer that can complete a connection and send `AddProvider` requests. No validator key is
needed.

## Impact

Each `AddProvider` request can carry a large address list (for example about 16 KB). With up to 1024
requests over one connection, the node writes about 16 MB of attacker-controlled data to the consensus
database. A modest Sybil set raises this to a few hundred MB. The growth is bounded but is disk and
memory pressure that the attacker controls.

Suggested severity: Medium. Bounded single-node resource exhaustion. No chain-halt.

## Reproduction (sketch)

1. Connect to the target as a non-committee peer.
2. Send `AddProvider` requests, each with an address list filled toward the message-size limit.
3. Observe the consensus database grow by the attacker-controlled volume.

## Suggested fix

Cap `ProviderRecord.addresses` length in `add_provider` (or in the `From<ProviderRecord>` impl),
mirroring `MAX_ADVERTISED_MULTIADDRS`. Reject or truncate any record whose address count is over the cap.

## Scope

The provider-count-per-key and provider-key-count limits are already enforced. The only missing control
is the address-list length inside a single provider record.

MavenRain:
Fixes #1185.

## Problem

`add_provider` stores a provider record's address list verbatim. The signed `NodeRecord` path rejects a record that advertises more than `MAX_ADVERTISED_MULTIADDRS` addresses (now tied to `MAX_MULTIADDRS_PER_PEER` = 1) (`peer_record_valid`, GHSA-29v6-gvv5-45gx), but the provider path had no analogue. The existing provider caps bound different axes: `max_providers_per_key` bounds distinct providers per key and `max_provided_keys` bounds the key count. Neither bounds the address list inside one record.

A byzantine peer that completes a connection can send `AddProvider` requests with the address list sized toward the message limit (about 16 KB per request). With up to 1024 requests over one connection the node writes about 16 MB of attacker-controlled data to the consensus database; a modest Sybil set raises this to a few hundred MB. No validator key is needed, and the write happens at admission, before any use of the record. This is bounded, single-node disk and memory pressure, not a chain halt.

## Fix

- [x] `crates/network-libp2p/src/kad.rs`: `add_provider` rejects a record whose address list exceeds `MAX_ADVERTISED_MULTIADDRS`, with a `network-kad` warn log and `Error::ValueTooLarge`. The guard runs first, before the eviction attempt and the capacity check, so a rejected record mutates no store state.
- [x] `crates/network-libp2p/src/consensus.rs`: `MAX_ADVERTISED_MULTIADDRS` becomes `pub(crate)` so both admission points share one constant. One cap, two enforcement sites, no drift.

The cap sits on the single write path into the provider tables (the primary and worker tables both dispatch through `add_provider`), so every stored record is bounded at admission and the read paths need no change. Rows written before this fix are not rescanned; a legitimate overwrite replaces them and expiry retires them.

## Testing

- [x] `test_kad_add_provider_rejects_oversized_address_list`: `MAX_ADVERTISED_MULTIADDRS + 1` addresses. `add_provider` returns `Err(ValueTooLarge)`, the provider count stays 0, `providers()` stays empty.
- [x] `test_kad_add_provider_admits_address_list_at_cap`: exactly `MAX_ADVERTISED_MULTIADDRS` addresses are admitted and stored intact (positive control). The tests reference the constant, so they track the merged retie of the cap from 16 to `MAX_MULTIADDRS_PER_PEER` (= 1) and pin the boundary at 1/2.
- [x] `test_kad_add_provider_worker_rejects_oversized_address_list`: the worker table rejects through the same guard.

Commands run under the pinned toolchain (`cargo +1.94 test -p tn-network-libp2p -- test_kad`, `-j 2`): 16 passed, 0 failed, including the 13 pre-existing kad store tests. Re-run green after merging current `main` (which reties the cap to `MAX_MULTIADDRS_PER_PEER` = 1), alongside a green workspace `cargo +1.94 check --all-targets` on the merged tree.

Each new rejection test was confirmed by mutation: with the guard neutralized (`<= MAX_ADVERTISED_MULTIADDRS` replaced by `<= usize::MAX`) both reject tests fail at their `matches!` assertions while the at-cap control still passes (2 passed, 2 failed); the source was byte-restored (verified identical to the index) and the full kad filter re-ran green (16 passed, 0 failed). The mutation check was repeated after the merge of current `main`, at the retied 1/2 address boundary, with the same result.

Static ladder also green: nightly `fmt -- --check` clean; ctxcat-review over the staged diff plus an independent logic review returned no findings (the logic pass verified against pinned `libp2p-kad-0.48.0` that `start_providing` builds the local self-record with an empty address list, so the cap cannot reject the node's own records, and that a rejected inbound record surfaces as a logged, non-fatal store error).


grantkee (dismissed):
Good catch

sstanfield (approved):

grantkee (approved):

Issue #554, opened 2026-02-14
feat: Multi-Worker Support — PR 1: Configuration


grantkee:
## Problem

The node currently hardcodes a single worker per validator (`worker_id = 0`). The `EpochManager`, network layer, and several initialization paths assume exactly one worker. This blocks the ability to run independent fee markets, specialized transaction pools, or any form of worker-level parallelism.

## Goal

Refactor the node to support **N independent workers per validator**. Each worker operates as a standalone unit with its own:

- libp2p swarm (dedicated gossip topics, listen address, network key)
- RPC server (unique port)
- Transaction pool
- Batch builder + batch validator
- `LocalNetwork` instance for primary communication

Workers share only the **Primary** (consensus) and the **execution engine** (block production). The `num_workers` count is a consensus-level parameter — all validators must agree on it.

## Why

The immediate motivation is **multiple fee markets**. Once multi-worker is in place, a follow-up (Phase 2) spawns 2 workers by default:

- **Worker 0 (General):** accepts all transactions, standard EIP-1559 fee market
- **Worker 1 (Whitelisted Transfers):** accepts only whitelisted ERC-20 `transfer`/`transferFrom` calls, operates with a reduced base fee

This architecture also enables future process separation — workers can be extracted into standalone processes communicating with the primary over RPC. They can even support multiple execution environments.

## Design Constraints

1. **Workers are fully independent** — no cross-worker shared state. Each worker has its own network identity, pool, and gossip topics.
2. **Per-worker gossip topics** — `tn-worker-{id}` and `tn-txn-{id}` replace the current global `tn-worker` and `tn-txn` topics. This provides network-level isolation.
3. **Per-worker `LocalNetwork`** — each worker gets its own `LocalNetwork` instance for primary communication. The primary registers as the handler on every worker's `LocalNetwork`. This is the seam for future process separation.
4. **`num_workers` is a consensus parameter** — changing it requires a coordinated upgrade across all validators. Defaults to `1` for backward compatibility.
5. **Execution engine is shared** — batches from all workers are processed sequentially by the same engine. Worker ID is already encoded in the block `difficulty` field.
6. **Faucet on worker 0 only** — the testnet faucet attaches to the general-purpose worker.

## Current State

Much of the infrastructure already supports N workers but is only called with `worker_id = 0`:

- `ExecutionNodeInner.workers: Vec<WorkerComponents>` — vec exists, only 1 element
[...]

PR #1223, opened 2026-08-20
Fork multi worker

grantkee:
- Adds `MULTI_WORKERS_FORK_EPOCH` to `forks.rs` as a dormant `u32::MAX` placeholder cloned from the seed_signature mechanism: `multi_workers_fork_active(epoch)` gates on the epoch carried inside the value being encoded or decoded, never node-local state. Non-adiri builds are active from genesis (no non-adiri network holds pre-#554 packs); `TN_MULTI_WORKERS_FORK_EPOCH` overrides under test-utils.
- `CommitteeInner` serde is now hand-written with an `is_human_readable` split: YAML/JSON keep the derive's exact behavior (committee files load unchanged), while binary formats branch on the in-band epoch. The pre-fork arm is byte-identical to the pre-#554 derive, verified against shadow structs of origin/main's derives in all four directions (legacy decode, byte-identical re-encode, old binary reads new output, post-fork output unchanged).
- `FIELD_NAMES` stays the 4-entry post-fork superset because bcs sizes its field sequence from the local list; a 3-entry list would silently `None` the post-fork `num_workers` read.
- The pre-fork encoder fails closed on `num_workers != 1` or any multi-worker bootstrap server (surfaces as a panic through `tn_types::encode`); the graceful guards sit earlier: epoch entry errors when the on-chain count exceeds one while the fork is inactive (the post-fork case keeps the existing warn), and genesis generation rejects a multi-worker agreement the epoch-0 committee could not encode. `tn-config` gains an `adiri` feature forward so the pre-fork test lane runs.
- `verify_epoch_meta` now pins `committee.epoch() == meta.epoch`. Local writes derive the record epoch from the committee, so only wire input (peer epoch sync, imported bundles) can diverge, and the committee's own epoch is what selects its decode layout.
- Test architecture is differential, copying the Header/seed_signature precedent: derived legacy shadow reprs with deterministic fixed-byte fixtures, frozen golden hex for both layouts (the legacy golden stays pinned on default builds through the shadow; epoch `u32::MAX` is post-fork even while dormant, so the V1 golden pins on both lanes), a byte-length oracle sweep in tests/it (108 adiri / 72 default grid points, delta measured from encoded parts), and an epoch-corruption test proving a poked epoch fails loudly instead of resuming at a wrong offset.
- First byte fixture in `crates/storage`: a frozen sealed epoch-407 pack drives all read doors under adiri (`open_append_exists` on a bare data file, which is the exact door that exited 1 in production, plus `open_static`, `stream_import` with byte-identity on the imported file, the `open_append` meta-compare arm, and `validate_pack_file`), while the non-adiri lane pins the loud rejection contract.
- e2e: `test_epoch_sync` step 8 grows from existence checks to content proof (fingerprint sealed packs at the kill, revalidate with `validate_pack_file` plus byte equality after restart), and a new wrapper pins the fork at epoch 1 so one datadir holds a legacy epoch-0 pack and a post-fork epoch-1 pack, both revalidated after restart. `TN_MULTI_WORKERS_FORK_EPOCH` is forwarded by `TestBinary::command` and wired into the Makefile fork lanes, never the bare `--workspace` runs.
- A domain review pass (storage/epoch/consensus) found the mechanism sound with no blocker or high findings and produced three fix-forward commits: the legacy goldens moved one epoch below the arming floor (406, so arming at exactly 407 cannot flip them post-fork), the e2e harness pins both fork variables (the pack revalidation also decodes seed_signature-gated headers, and the single-var pin failed misleadingly without the Makefile env), and the CI adiri lane now runs `tn-storage` and `tn-config` with their adiri forwards, which previously compiled the golden-pack door tests and the pre-fork genesis test out of every job.
- Review notes deferred deliberately: the epoch-0 pack is written from the committee file before the epoch-entry guard runs (a mis-set file fails as an encode error, not the guard's message) - resolved structurally by the separate epoch-0-meta branch that defers pack creation to the first `new_epoch`; `num_workers` is a `usize` on the wire (8 fixed bytes, inherited from the base PR) - worth deciding whether a fixed-width type should land before this layout ships anywhere; and the base PR rewrote `chain-configs/*/committee.yaml` to the `workers:` shape, which pre-#554 binaries cannot read (new binaries read both; existing datadirs keep their old file) - one line for the rollout notes.
- Proven against real adiri data: a copied validator `consensus-db` (sealed epochs sampled at 1, 111, 404, and 406, the newest) driven through three `--features adiri` builds. The pre-#554 binary validates every sampled pack and cold-boots through `ConsensusChain::new` to epoch 406; the #554 foundation binary fails every `db validate` and dies in `open_append_exists` with `Epoch Load Error (deserialize value: unknown protocol id: 35)`, the exit-1 failure this fixes; this branch's binary produces validate reports byte-identical to the pre-#554 binary, boots to the same state down to the consensus header digest, and leaves sealed pack bytes hash-unchanged through a writable open.
- Arming is a follow-up PR: retarget the constant (must be >= 407 per `CONSENSUS_REGISTRY_FORK_EPOCH` and comfortably above the live epoch), drop the self-flagged `clippy::absurd_extreme_comparisons` expect, and confirm the mainnet genesis-active assumption.


crates/config/src/genesis.rs
grantkee, line 1: these changes might be overkill, but I'll wait for feedback

grantkee:
Attestation complete - no option to verify through GH

MavenRain:
This looks clean to me, so will pull it into the mix as-is

PR #1224, opened 2026-08-20
fix(engine): credit batch priority fees to the producer beneficiary (#1222)

### LINKED ISSUE #1222: Batch beneficiary is the header author, not the batch producer: a byzantine validator can steal another validator's priority fees
## Summary

The block beneficiary for an executed batch is the execution address of the **sub-DAG header author**, not the validator that actually produced (sealed) the batch.  Batch digests carry no binding to their producer, and header validation checks only that a batch exists in the shared worker lane, not who made it.  A byzantine validator can copy another validator's already-gossiped batch digest into its own certified header.  If that header is ordered at or before the victim's header in the committed sub-DAG, the byzantine validator receives all of the batch's priority fees and the victim receives nothing.

This was raised by **Haxatron** in the Cantina collaborative review:

> Currently the beneficiary is `self.authority_execution_address(header.author())`. A byzantine validator can observe and then insert another validators batch digest into their header and then redirect all priority fees from the victim validator to the byzantine validator, assuming their certified header is ordered first in the subDAG
>
> Byzantine validators can also observe other sealed batches during dissemination and then include their transactions inside them.  This allows for standard frontrunning (reorder) and also theft of priority fees.

We reviewed the code and confirm the first claim.  The second claim (frontrunning via plaintext batch dissemination) is also feasible but is a separate, more generic root cause.  See "Scope" below.

## Severity

High confidence that the mechanism is real and reachable under the standard "up to f byzantine" threat model.  No collusion or 2f+1 stake is needed for the theft.

Economic magnitude is bounded and scales with priority-fee (tip) activity:
- Only the **priority fee** (tip) of the captured batch is redirected.
- The **base fee** is not affected.  It always goes to `basefee_address` (governance safe).
- **Leader block rewards** are not affected.  They are keyed on the deterministic Bullshark leader schedule, not on batch attribution.

The one precondition that is not provable from static reading is the probability that the attacker wins the sub-DAG ordering race against the true producer under real network latency.  The attacker has a plausible causal timing edge (see step 2 below), but "wins on demand" versus "wins some fraction of the time" needs a testnet or simulation to quantify.

## Root cause

The beneficiary is resolved from the header author, in both `iter_to_output` paths.

`crates/storage/src/consensus_pack.rs:1693` (and the legacy path at `:1832`):

```rust
let address = committee.authority(header.author()).map(|a| a.execution_address());
if let Some(address) = address {
    batches.push(CertifiedBatch { address, batches: cert_batches });
}
```

That address is used as the block beneficiary during execution.

`crates/engine/src/payload_builder.rs:203-205`:

```rust
let payload = TNPayload::new(
    canonical_header,
    cert_batch.address,   // beneficiary
    ...
```

`crates/tn-reth/src/evm/handler.rs:171-177` credits the priority fee to that beneficiary:

```rust
// transfer priority fee to coinbase/beneficiary
let coinbase_gas_price = effective_gas_price.saturating_sub(basefee);
context
    .journal_mut()
    .load_account_mut(beneficiary)?
    .incr_balance(U256::from(coinbase_gas_price.saturating_mul(gas_used)));
```

## Why current checks do not prevent it

1. **No producer binding on batch digests.**  `Header::validate` (`crates/types/src/primary/header.rs:123-169`) checks that the author is a committee member and that every `worker_id` in the payload is in range.  It does not check that any payload digest was produced by the author.  The payload is a bare `BTreeMap<digest, worker_id>`.

2. **Vote-time availability check is producer-agnostic.**  `sync_header_batches` (`crates/consensus/primary/src/state_sync/header_validator.rs:99-123`) accepts a payload entry if `contains_payload(digest, worker_id)` holds, that is, if *some* batch with that digest exists under that shared worker lane.  The in-code comment concedes the gap:

   > `header_validator.rs:118` : Note on this note- the soure of batches is now agnostic so this may not be a DOS anymore, still seems like a useful check though...

3. **The worker lane is global.**  `worker_id` is shared across authorities (worker 0 talks to other worker 0s).  A victim's sealed batch is gossiped to the attacker's same-lane worker, so the attacker holds the digest and honest voters hold it too.  The attacker pairs the victim's digest with the shared `worker_id`;  there is no author field to spoof.

4. **The payload store and messages carry no producer identity.**  `PayloadStore` is keyed on `(digest, worker_id)` (`crates/storage/src/stores/payload_store.rs`).  `WorkerOthersBatchMessage` carries only `digest` and `worker_id` (`crates/network-types/src/notify.rs`), and `report_others_batch` writes it straight to the store (`crates/consensus/primary/src/network/mod.rs:1598`).

5. **Ordering gates the reward, and there is no author tiebreak.**  `order_dag` (`crates/consensus/primary/src/consensus/utils.rs:55`) sorts by round ascending, with same-round ties resolved by DAG DFS discovery order.  There is no author-based or producer-based tiebreak.  Execution order equals `sub_dag.headers()` order.  The first header to reference a digest gets the real batch;  a later header that references the same digest replays the same transactions, which fail on nonce and are skipped.

   `crates/tn-reth/src/env/execution.rs:169-175`:
   ```rust
   // allow transaction errors (ie - duplicates)
   // it's possible that another worker's batch included this transaction
   RETH_METRICS.invalid_txs_skipped_total.increment(1);
   continue;
   ```
   A skipped transaction contributes zero `gas_used`, so the later (duplicate) beneficiary earns zero.  Whoever is ordered first collects 100% of the tips.

## Attack walkthrough

1. The victim's worker seals a batch and broadcasts the full batch to all committee workers for the availability quorum (`crates/consensus/worker/src/quorum_waiter.rs`).  The attacker's worker receives it and stores the digest.
2. The attacker learns the digest early.  The quorum broadcast reaches the attacker's worker before the victim's own primary is notified through `report_own_batch`, so the attacker can reference the digest in a header of the same round or earlier.
3. The attacker builds its own header with `payload = { victim_digest -> worker_id }` and requests votes.  Honest voters accept it because `contains_payload(victim_digest, worker_id)` is already true locally.  The header is certified.
4. Both the attacker's header and the victim's header reach the same committed sub-DAG referencing the same digest.
5. During linearization, the attacker's header is at a round less than or equal to the victim's, or wins the same-round DFS tiebreak.  The engine executes the batch under the attacker as beneficiary and credits all priority fees to the attacker.  The victim's later block replays the same transactions, which are skipped, and the victim earns nothing.

## Intent mismatch

The current behavior appears unintended.  The type and one README describe the beneficiary as the batch **producer**, while the code and the other README describe the certificate **author**:

- `crates/types/src/primary/output.rs:34-37` (doc on `CertifiedBatch.address`): "The ECDSA address of the authority that **produced the batch**. This address is used as the block beneficiary during execution."
- `crates/tn-reth/README.md:48`: "`beneficiary` | Execution address of the authority that **produced the batch**;  receives priority fees."
- `crates/engine/README.md:46`: "Is the execution address of the **certificate's author** ... The engine takes the resolved address on trust."

The sealed batch even carries its own `beneficiary` field for the producer, which is discarded downstream in favor of the header author's address.

## Suggested remediation

The fix should bind a batch's fee credit to its producer rather than to the header author.  Options, in rough order of preference:

1. **Attribute fees to the batch producer.**  Carry the producing authority (or its execution address) with the batch, verified at seal or attestation time, and use that as the beneficiary instead of `committee.authority(header.author())`.  The sealed batch already has a `beneficiary` field that can serve this purpose if it is authenticated.
2. **Credit each digest once, to its first legitimate referencer only.**  Deduplicate digests across headers before assigning beneficiaries, so a copied digest in a second header earns nothing even if that header is ordered first.
3. **Bind digests to producers at vote time.**  Extend the availability check so a header may only reference a digest that was produced by the header author's own worker, and record `(digest, producing_authority)` rather than `(digest, worker_id)` alone.  Note the historical DoS concern in the `header_validator.rs` comment when tightening this.

## Scope

Haxatron's second observation (a byzantine worker reads plaintext transactions during dissemination and copies or reorders them into its own batch) is also feasible.  Worker-to-worker batch dissemination sends the full plaintext `SealedBatch` before consensus ordering, with no encryption or commit-reveal (`crates/consensus/worker/src/quorum_waiter.rs`, `crates/consensus/worker/src/network/handler.rs:270`).  This is the standard unencrypted-mempool frontrunning class and is a **separate** root cause from the digest-reuse theft above: it requires fabricating a new batch from copied content, whereas the primary finding reuses an existing digest at zero cost.  Fixing one does not fix the other.  We suggest tracking it separately if it warrants a report.

## Distinct from filed issues

This is distinct from #1205 (blob gasprice in the EVM env), #1211 (blsVerify gas floor), #1218 (system-call logs), #554 (num_workers config), and #1042 (unchecked `u128` multiply in `reward_beneficiary`, an arithmetic bug in the crediting math rather than a who-gets-credited bug).

Credit: reported by Haxatron in the Cantina collaborative review.

MavenRain:
Closes #1222. (Cantina #4)

## Problem

The block beneficiary that receives a batch's priority fees was resolved from the sub-DAG **header author**, not from the validator that produced the batch.  A byzantine validator can copy another validator's already-gossiped batch digest into its own certified header;  if that header is ordered at or before the victim's in the committed sub-DAG, the byzantine validator receives all of the batch's priority fees and the victim receives nothing.

No collusion and no 2f+1 stake are required: the theft works under the standard "up to f byzantine" model.  Only the priority fee (tip) is redirected.  The base fee still goes to `basefee_address`, and leader block rewards are keyed on the Bullshark leader schedule, so neither is affected.

The mechanism was reported by Haxatron in the Cantina collaborative review.

## Root cause

The beneficiary is resolved from `header.author()` at the single execution crediting site and at both `CertifiedBatch` construction sites:

- `crates/engine/src/payload_builder.rs`: `cert_batch.address` was passed to `TNPayload::new` as the block beneficiary.  This is the one crediting site;  both the live consensus path and the historical pack-replay path flow through `execute_consensus_output`.
- `crates/consensus/executor/src/subscriber.rs` and `crates/storage/src/consensus_pack.rs` build `CertifiedBatch { address, .. }` from `committee.authority(header.author()).execution_address()`.

The batch itself already carries the producer's intended fee recipient: `Batch::beneficiary`.  That field is covered by the batch digest (only `received_at` is `#[serde(skip)]`), so it is identical no matter which header references the digest.  Reading the beneficiary from the header author instead of from the batch is what let a copied digest redirect the fees.  The doc on `CertifiedBatch.address` even described it as "the ECDSA address of the authority that produced the batch," which the header-author derivation did not honor.

## Fix

Credit each batch's priority fees to the producer's own digest-bound `Batch::beneficiary`, gated behind a hard fork so already-committed blocks re-execute unchanged.

- [x] `crates/engine/src/payload_builder.rs`: in the non-empty batch loop, resolve the beneficiary as `batch.beneficiary` when the producer-beneficiary rule is active for the output's epoch, else the legacy `cert_batch.address`.  This is the only crediting site, so the change is symmetric across the live and pack-replay paths by construction (both execute here).
- [x] `crates/types/src/forks.rs`: add `BATCH_PRODUCER_BENEFICIARY_FORK_EPOCH` plus `batch_producer_beneficiary_active(epoch)`, mirroring the existing `SEED_SIGNATURE_FORK_EPOCH` block exactly (adiri dormant until the fork epoch, mainnet active from genesis, a `TN_BATCH_PRODUCER_BENEFICIARY_FORK_EPOCH` test-utils override).  The gate reads the epoch carried inside the consensus output (`output.leader().epoch()`), never node-local committee state.
- [x] `crates/types/src/primary/output.rs` and `crates/engine/README.md`: document that `CertifiedBatch.address` is the header author's execution address, used as the beneficiary only before the fork;  from the fork onward each batch's fees follow `Batch::beneficiary`.

### Threat model

- The redirect was caused by reading the beneficiary from the referencing header, not by the referencing itself.  `Batch::beneficiary` is inside the batch digest, so an attacker cannot reference the victim's fee-bearing batch while pointing the fee elsewhere: changing the beneficiary changes the digest, which is a different batch the attacker would have to produce and get certified itself (ordinary inclusion, not theft).  Crediting the in-digest field therefore closes the theft at the root with no attacker-dependent assumptions.
- A vote-time producer-binding check is **not** required and is left out deliberately.  It would reject a header that references a batch it did not produce, which addresses batch-inclusion / DoS concerns (the source-agnostic availability check is intentional per the in-code note at `header_validator.rs`), not the fee credit.  Once the fee follows `Batch::beneficiary`, referencing another validator's digest gains the attacker nothing.
- `Batch::beneficiary` is producer-chosen and unvalidated, so a producer can still direct its own batch's fees anywhere (including giving them away).  That is self-affecting and standard;  it is not theft of another validator's fees and is out of scope for this fix.
- The rule change alters the sealed header's `beneficiary` (block hash) and the priority-fee balance credit (state root) for every priority-fee block, so it is fork-gated: earlier epochs keep resolving from the header author and re-execute byte-identically during sync.  The gate is applied at the single execution site the live and replay paths share, so writer and reader cannot diverge.  The activation epoch value is a placeholder set safely in the future;  re-verify it against the live adiri chain at merge time and raise it in the same PR if that epoch has already begun.

## Testing

Two tests were added and run locally under `testledger` (build plus isolated run), each pinned to the fork contract:

- `crates/types/src/forks.rs`: unit test `batch_producer_beneficiary_boundary_matches_this_builds_rollout_contract`.  Pins the gate to this build's rollout: under `adiri` it is dormant for epochs `0`, `1`, `BATCH_PRODUCER_BENEFICIARY_FORK_EPOCH - 1` and active from `BATCH_PRODUCER_BENEFICIARY_FORK_EPOCH` onward (`>=`, not `>`);  non-adiri builds are active from genesis.  It also asserts the `test-utils` env override is unset in this process, so a latched override cannot make the boundary read as a pass.
  - `testledger -- cargo test -p tn-types --lib -j 2 batch_producer_beneficiary_boundary_matches_this_builds_rollout_contract`
  - Result: `1 passed; 0 failed; 0 ignored`.
- `crates/engine/tests/it/main.rs`: regression test `test_priority_fee_credits_batch_producer_not_header_author`.  Builds the #1222 theft scenario: a batch whose producer beneficiary (`0x1111...1111`) differs from the sub-DAG header author's committee execution address (the attacker, `0x0033a3...665ccb`), then asserts the sealed block beneficiary is the batch producer, not the header author.
  - `testledger -- cargo test -p tn-engine --test it -j 2 test_priority_fee_credits_batch_producer_not_header_author`
  - Result: `1 passed; 0 failed; 0 ignored`.

Mutation confirmation (the regression test is load-bearing).  Reverting the fix in `payload_builder.rs` to the unconditional pre-fork `let beneficiary = cert_batch.address;` (so `batch.beneficiary` no longer appears in the crediting loop) makes the engine test fail as expected:

    thread 'test_priority_fee_credits_batch_producer_not_header_author' panicked at crates/engine/tests/it/main.rs:2619:5:
    assertion `left == right` failed: block beneficiary must be the batch producer (batch.beneficiary), not the header author
      left: 0x0033a370616805b1fd275b7ffab83fc41d665ccb   (header author execution address)
     right: 0x1111111111111111111111111111111111111111   (batch producer beneficiary)

The fix was then restored from backup and verified: the gated beneficiary line is present, no mutation residue, `5 files changed, 409 insertions(+), 7 deletions(-)` across the crate sources.  Without the fix the block credits the header author, which is exactly the redirect #1222 describes.

Static ladder (local): nightly `cargo +nightly fmt -- --check` clean;  `diffclass` verdict BUILD (code changed), so the build and full test matrix are owned by the dynamic tier.

Dynamic tier (owns build plus full test):

- GitHub CI (`.github/workflows/pr.yaml`) runs on the push: nightly `fmt --all -- --check`, nightly `clippy --locked --all --all-features -- -D warnings`, `nextest --locked --workspace --exclude tn-faucet`, and the adiri feature lane.
- The pinned `cargo +1.94 test --no-run --workspace` attest runs on the second machine post-push.

### Merge with main (63780c38) and review follow-up

Main advanced past the branch point (#1219 multi-worker foundation and others).  One conflicted file, `crates/types/src/forks.rs`: both sides appended a new fork block (constant + gate + test-utils override + rollout-contract tests) at the same anchors.  The resolution keeps both additions verbatim;  no line of either side is modified.  The two forks arm independently.

A separate follow-up commit (kept out of the merge so it stays reviewable) aligns this PR's fork with main's fork bookkeeping and addresses the review comment:

- The startup fork-schedule log (#1086) now includes `batch_producer_beneficiary_fork_epoch`, the same way main's merge added `multi_workers_fork_epoch`.
- The e2e harness doc states why `TN_BATCH_PRODUCER_BENEFICIARY_FORK_EPOCH` is deliberately absent from the child fork-pin loop: that loop guards wire-format forks against harness/binary decode mismatches;  this fork gates an execution rule, so spawned nodes keep their build default and the engine tests own both sides of the gate.
- The engine README threat-model table records the transaction-poaching residual from review: certification binds a batch's fees to its producer, but a validator can re-pack another validator's publicly gossiped transactions into its own batch as beneficiary.  That is an inclusion race (public-mempool MEV class);  only the first-ordered copy collects, and closing it fully needs sealed batch content.  Tracked in #1259.


sstanfield (dismissed):
This seems good.  I think there could still be a related attack where the validator poaches individual transactions.  That may be hard to close...

MavenRain:
I agree, and thank you for the review.  The tx-level poach is real and this fix does not close it: crediting `Batch::beneficiary` binds a batch's fees to the validator that built that batch, but nothing stops a validator from reading another validator's gossiped batch and re-packing the same transactions into a batch of its own, with itself as beneficiary.

Two structural points keep it in a different class from the digest-copy redirect this PR closes:

1. The poacher does real work: it must build, gossip, and get its own batch certified through the normal path.  There is no free redirect of someone else's certified work.
2. Only the copy ordered first collects.  The loser's duplicates fail the nonce check at execution, so the poach is an inclusion race on publicly visible transactions, the same economics as public-mempool MEV, not guaranteed theft.

Closing it fully would need the batch contents to be sealed until ordering (commit-reveal or committee threshold encryption), which is a protocol change well beyond this fix.  I have documented the residual in the engine README threat-model table in this PR and filed a tracking issue so it does not get lost: #1259.

sstanfield (dismissed):

grantkee (commented):
I'm not sure this will fork testnet? Also, we should clean up the subscriber since this was the only place that used `cert_batch.address` I think

MavenRain:
That's a fair question, especially after the epoch-383 boundary.  Here is the functional chain, link by link.

**Is the fork functional?**

1. *The gate is wired to the right epoch.*  `execute_consensus_output` derives `epoch` from the output being executed (`output.leader().epoch()`, `crates/engine/src/payload_builder.rs:43`) and the beneficiary selection at `:213` gates on that same value.  It never reads node-local committee state, so live execution and historical pack replay take the same arm for the same output.

2. *The gate itself flips at the boundary.*  `batch_producer_beneficiary_boundary_matches_this_builds_rollout_contract` (`crates/types/src/forks.rs:629`) asserts through the public entry point that adiri builds are dormant at `0`, `1`, and `FORK_EPOCH - 1` and active at `FORK_EPOCH` and `u32::MAX` (`>=`, not `>`), and that non-adiri builds are active from genesis.  The adiri branch is not dead coverage: the `adiri-test` CI lane compiles tn-types with `tn-types/adiri` and runs it (`.github/workflows/pr.yaml:148`).

3. *The active arm works through real execution.*  `test_priority_fee_credits_batch_producer_not_header_author` (`crates/engine/tests/it/main.rs:2447`) executes a real output where the batch producer and the header author differ, and asserts both the sealed header's beneficiary and the balance credit.  Reverting the selection to `cert_batch.address` flips both assertions, so the test kills that mutant;  the run was mutation-confirmed before push.

4. *The dormant arm is the status quo, not new code.*  The `else` branch is the exact rule the chain executes today for every block;  the PR only wraps it in the gate.   I will be direct about the coverage: no engine test executes that `else` arm, because the it-binary is non-adiri (active from genesis) and the env override is OnceLock-latched per process.  The e2e doc overclaimed this ("the engine tests own both sides");  I have corrected it in this PR.  If you want it closed properly, I can add a boundary-crossing engine test in its own test process that pins `TN_BATCH_PRODUCER_BENEFICIARY_FORK_EPOCH` and executes one output on each side of the boundary, asserting the credited account flips;  say the word and I will push it or file it as a follow-up.

5. *It cannot fail the way epoch 383 failed.*  The seed-signature halt was a fork gate coupled to epoch-open, blocking on a certificate that its own not-yet-spawned consumer had to produce.  This gate is a pure predicate on the output's epoch, evaluated inside execution;  it waits on nothing, exchanges no messages, and does not touch epoch open or close.  The worst case on a mixed fleet is state divergence at the boundary, which is the standard hard-fork rollout constraint, not a deadlock.

One expectation to set: at the flip, blocks only change where `batch.beneficiary` differs from the header author's execution address.  Honest validators that reference their own batches with matching config produce identical blocks under both rules, so the fork can be observably quiet at the boundary.  Quiet is the success case, not a sign the gate did not fire.

**Subscriber cleanup.**  The engine's post-fork path is indeed the only consumer that stopped reading `cert_batch.address`, but the field still has one live reader: the pre-fork arm in `crates/engine/src/payload_builder.rs:216`.  Adiri sync from genesis re-executes every pre-fork block under the header-author rule, so the subscriber must keep resolving the address (`crates/consensus/executor/src/subscriber.rs:523`), and the pack-replay reader mirrors the same resolution (`crates/storage/src/consensus_pack.rs:1724`).  Removing it would break historical replay.  On non-adiri builds the read is dead code because the gate is always active, but the type is shared across builds.  If you want, a follow-up could make the field `Option<Address>` populated only for pre-fork epochs, but that buys little for the divergent construction path it adds;  I would leave it until genesis replay is no longer a requirement.

grantkee:
> That's a fair question, especially after the epoch-383 boundary. Here is the functional chain, link by link.
> 
> **Is the fork functional?**
> 
> 1. _The gate is wired to the right epoch._  `execute_consensus_output` derives `epoch` from the output being executed (`output.leader().epoch()`, `crates/engine/src/payload_builder.rs:43`) and the beneficiary selection at `:213` gates on that same value.  It never reads node-local committee state, so live execution and historical pack replay take the same arm for the same output.
> 2. _The gate itself flips at the boundary._  `batch_producer_beneficiary_boundary_matches_this_builds_rollout_contract` (`crates/types/src/forks.rs:629`) asserts through the public entry point that adiri builds are dormant at `0`, `1`, and `FORK_EPOCH - 1` and active at `FORK_EPOCH` and `u32::MAX` (`>=`, not `>`), and that non-adiri builds are active from genesis.  The adiri branch is not dead coverage: the `adiri-test` CI lane compiles tn-types with `tn-types/adiri` and runs it (`.github/workflows/pr.yaml:148`).
> 3. _The active arm works through real execution._  `test_priority_fee_credits_batch_producer_not_header_author` (`crates/engine/tests/it/main.rs:2447`) executes a real output where the batch producer and the header author differ, and asserts both the sealed header's beneficiary and the balance credit.  Reverting the selection to `cert_batch.address` flips both assertions, so the test kills that mutant;  the run was mutation-confirmed before push.
> 4. _The dormant arm is the status quo, not new code._  The `else` branch is the exact rule the chain executes today for every block;  the PR only wraps it in the gate.   I will be direct about the coverage: no engine test executes that `else` arm, because the it-binary is non-adiri (active from genesis) and the env override is OnceLock-latched per process.  The e2e doc overclaimed this ("the engine tests own both sides");  I have corrected it in this PR.  If you want it closed properly, I can add a boundary-crossing engine test in its own test process that pins `TN_BATCH_PRODUCER_BENEFICIARY_FORK_EPOCH` and executes one output on each side of the boundary, asserting the credited account flips;  say the word and I will push it or file it as a follow-up.
> 5. _It cannot fail the way epoch 383 failed._  The seed-signature halt was a fork gate coupled to epoch-open, blocking on a certificate that its own not-yet-spawned consumer had to produce.  This gate is a pure predicate on the output's epoch, evaluated inside execution;  it waits on nothing, exchanges no messages, and does not touch epoch open or close.  The worst case on a mixed fleet is state divergence at the boundary, which is the standard hard-fork rollout constraint, not a deadlock.
> 
> One expectation to set: at the flip, blocks only change where `batch.beneficiary` differs from the header author's execution address. Honest validators that reference their own batches with matching config produce identical blocks under both rules, so the fork can be observably quiet at the boundary. Quiet is the success case, not a sign the gate did not fire.
> 
> **Subscriber cleanup.** The engine's post-fork path is indeed the only consumer that stopped reading `cert_batch.address`, but the field still has one live reader: the pre-fork arm in `crates/engine/src/payload_builder.rs:216`. Adiri sync from genesis re-executes every pre-fork block under the header-author rule, so the subscriber must keep resolving the address (`crates/consensus/executor/src/subscriber.rs:523`), and the pack-replay reader mirrors the same resolution (`crates/storage/src/consensus_pack.rs:1724`). Removing it would break historical replay. On non-adiri builds the read is dead code because the gate is always active, but the type is shared across builds. If you want, a follow-up could make the field `Option<Address>` populated only for pre-fork epochs, but that buys little for the divergent construction path it adds; I would leave it until genesis replay is no longer a requirement.

I agree in general, but what I mean is we've deliberately avoided expanding the initial 5 validator set to this point to ensure forks were kept to a minimum. The 5 validators are all running the same protocol, and I'm confident the operators are running our code (and have not acted maliciously *yet*) - so I expect the value from header and batch to still be the same for all batches executed on testnet with the current validator set. We will ask @blu3chipsx to sync a node before updating the network, but I'm confident this is not a fork on the current testnet.

grantkee (dismissed):
LGTM!

grantkee (approved):

Haxatron:
Fix verified. the correct beneficary (specified by the original batch producer and committed by batch digest) is now used.

PR #1226, opened 2026-08-20
fix(storage): fail open_append on an unreadable first pack record

### LINKED ISSUE #1225: open_append appends a second EpochMeta when the first pack record is unreadable
## Problem

`ConsensusPack::open_append` creates or reopens the data file for an epoch's consensus pack. When the file already holds records, the first record must be the epoch's `EpochMeta`, and the function fetches it to compare against the meta it expects. The fetch result is pattern-matched: a successful read runs the comparison, and any failure falls through to the arm meant for a brand-new file, which appends a fresh `EpochMeta`. That arm cannot tell "no record yet" from "record present but unreadable", and the underlying `Pack::fetch` gives it no way to: a read at end-of-file surfaces as an IO error (`read_exact` hits UnexpectedEof), the same class as a torn record or a failed CRC, never a NotFound. Classification by error kind is impossible at this call site.

So any unreadable first record, whether a torn write from a crash during the very first meta append or a record corrupted at rest, is silently answered by appending a second `EpochMeta` after the bad bytes, and the open reports success. The pack is corrupt from that point on and nothing later can fix it: `trunc_and_heal` only repairs the tail of the file, `pack_validate` flags the duplicated meta as `EpochMetaMismatch`, and the warm-restart path (`open_append_exists`) refuses the file outright. The node meanwhile keeps appending consensus output into the broken container, so the corruption is discovered not at the open that caused it but at some later restart, validation run, or peer sync.

The torn case is a live crash window, not a hypothetical. `open_append` runs at every epoch transition through `new_epoch` and on fresh cold start, and a crash midway through the first meta append followed by a restart lands exactly in the failure arm.

## Solution

Discriminate absence by file length, not by fetch outcome. A data file longer than the pack header must hold record bytes, so the first record must load as a matching `EpochMeta`; a failed load should propagate as `PackError::EpochLoad` and fail the open, mirroring how `open_append_exists` already reports it. A file at exactly header length is legitimately fresh, either newly created or left by a crash between the header write and the meta append, and appending the meta there stays correct and idempotent.

This deliberately changes behavior for a torn or corrupt first record: the open fails instead of "healing". That is the right trade. A pack whose first record is unreadable holds no usable consensus data, tail healing cannot repair it, and failing fast keeps the node from writing more consensus output into a file every later consumer will reject.

Regression coverage should pin all three shapes: a corrupt first record rejected without modifying the file, a torn first record rejected the same way, and the header-only crash window still recovered by appending the meta.

grantkee:
- `ConsensusPack::open_append` now discriminates a missing first record by file length instead of fetch success: record bytes past `DATA_HEADER_BYTES` mean the `EpochMeta` must load, and a failed load propagates as `PackError::EpochLoad` rather than appending a second meta. The existing meta-equality check (`InvalidEpoch` on mismatch) is untouched.
- Error-kind classification was not an option: `Pack::fetch` reports a read at EOF as an IO error (`read_exact` hits UnexpectedEof), never NotFound, so an absent record and a torn one arrive as the same error class.
- Deliberate behavior change: a torn or corrupt first record now fails the open instead of "healing" into a corrupt pack. `trunc_and_heal` only repairs the tail, such a pack holds zero usable consensus data, and the old path left `pack_validate` flagging a duplicate meta while `open_append_exists` refused the file on the next warm restart.
- The header-only crash window (data header written, meta never appended) still gets the meta appended; the fresh-pack path is unchanged and idempotent.
- Three regression tests: `test_open_append_rejects_corrupt_first_record` (byte flipped inside the meta record, `EpochLoad`, file length unchanged), `test_open_append_rejects_torn_first_record` (file truncated mid size-prefix, same outcome), and `test_open_append_appends_meta_to_header_only_file` (meta appended, output saved, clean reopen). The corrupt and torn tests were written first and returned Ok against the old code.

closes #1225

### DIFF
--- a/crates/storage/src/consensus_pack.rs
+++ b/crates/storage/src/consensus_pack.rs
@@ -747,6 +747,9 @@ impl Inner {
 
     /// Opens a new epoch pack for append.  Will create a new set of epoch static
     /// files to write consensus output into if they do not exist.
+    ///
+    /// A data file holding record bytes must begin with a readable [`EpochMeta`] matching this
+    /// epoch; an unreadable first record fails the open rather than appending a second meta.
     fn open_append<P: AsRef<Path>>(
         path: P,
         previous_epoch: &EpochRecord,
@@ -770,8 +773,16 @@ impl Inner {
             genesis_consensus: previous_epoch.final_consensus,
         };
 
-        if let Ok(meta) = data.fetch(DATA_HEADER_BYTES as u64) {
-            let meta = meta.into_epoch()?;
+        // Discriminate a missing meta by length, not by fetch error kind: fetch reports a read
+        // at EOF as an io error, the same class as a torn or corrupt record, so record bytes
+        // past the header mean the first record must load as a matching meta.  Failing the open
+        // here is deliberate; trunc_and_heal only repairs the tail, and a pack whose first
+        // record is unreadable holds no usable consensus data.
+        if data.file_len() > DATA_HEADER_BYTES as u64 {
+            let meta = data
+                .fetch(DATA_HEADER_BYTES as u64)
+                .map_err(|e| PackError::EpochLoad(e.to_string()))?
+                .into_epoch()?;
             if epoch_meta != meta {
                 return Err(PackError::InvalidEpoch(
                     epoch,
@@ -779,6 +790,8 @@ impl Inner {
                 ));
             }
         } else {
+            // Header-only file: brand new, or a crash landed between the header write and the
+            // meta append.  Either way appending the meta initializes the pack.
             data.append(&PackRecord::EpochMeta(epoch_meta.clone()))
                 .map_err(|e| PackError::Append(e.to_string()))?;
         }
@@ -2073,8 +2086,8 @@ pub(crate) mod test {
     };
 
     use crate::{
-        archive::pack::PackCompression,
-        consensus_pack::{max_batches_per_output, ConsensusPack, Inner, PACK_VERSION},
+        archive::pack::{Pack, PackCompression, DATA_HEADER_BYTES},
+        consensus_pack::{max_batches_per_output, ConsensusPack, Inner, PackRecord, PACK_VERSION},
         mem_db::MemDatabase,
     };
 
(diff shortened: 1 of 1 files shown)


crates/storage/src/consensus_pack.rs
```diff
@@ -770,15 +773,25 @@ impl Inner {
...
         };
 
-        if let Ok(meta) = data.fetch(DATA_HEADER_BYTES as u64) {
-            let meta = meta.into_epoch()?;
+        // Discriminate a missing meta by length, not by fetch error kind: fetch reports a read
+        // at EOF as an io error, the same class as a torn or corrupt record, so record bytes
+        // past the header mean the first record must load as a matching meta.  Failing the open
+        // here is deliberate; trunc_and_heal only repairs the tail, and a pack whose first
+        // record is unreadable holds no usable consensus data.
+        if data.file_len() > DATA_HEADER_BYTES as u64 {
```
MavenRain: Question on the fail-closed trade-off here.  A crash during the first meta append leaves partial bytes past the header.  Records are sequential, so nothing behind a torn first record is addressable:  that pack holds zero reachable consensus data, and the only lost artifact is the meta this call is about to rebuild from `previous_epoch` + `committee`.  This path now refuses such a pack until an operator deletes the epoch dir, while the header-only window one instant earlier (header written, no meta bytes) self-heals in the branch below.  Two thoughts:

1. Was auto-truncate-to-header considered for the narrow case where no readable record exists past the bad first record?  That would keep fail-closed for a pack with real data behind a corrupt meta, and self-heal the dataless torn window, consistent with the header-only branch.
2. If fail-closed stays (the PR body makes a reasonable case), could `EpochLoad` carry the file path and the remediation step?  Operators will hit this at epoch start on nodes already damaged by the old append-second-meta behavior, and the log line is all they will have.

crates/storage/src/consensus_pack.rs
```diff
@@ -3240,6 +3253,140 @@ pub(crate) mod test {
...
+        assert!(
+            std::fs::metadata(&data_path).expect("metadata").len() > DATA_HEADER_BYTES as u64,
+            "meta must have been appended"
+        );
+
+        // Reopening finds the appended meta and compares clean: recovering the crash window
+        // is idempotent.
+        let pack = ConsensusPack::open_append(temp_dir.path(), previous_epoch, committee.clone())
+            .expect("reopen append");
+        assert!(pack.get_consensus_output(1).await.is_ok());
```
MavenRain: Optional hardening:  after this reopen, also open the pack read-only and read the output through it:

```rust
let pack = ConsensusPack::open_static(temp_dir.path(), 0).expect("open static after header-only recovery");
assert!(pack.get_consensus_output(1).await.is_ok());
```

The append reopen goes through the heal-time length reconcile, but `open_static` is the path that runs `files_consistent`, so a static reopen pins the recovered pack for read-only consumers as well.

MavenRain (dismissed):
A couple questions and a suggestion . . . nothing blocking

crates/storage/src/consensus_pack.rs
grantkee: Good idea

crates/storage/src/consensus_pack.rs
grantkee: I agree
A.1 — auto-truncate-to-header: yes, possible. Implemented.

The discriminator can't be error-kind sniffing (the parent commit deliberately avoided that). It's two independent witnesses, and neither is sufficient alone:

- record_size(28) → UnexpectedEof. This validates the size prefix and CRC without decompressing, so it means the file physically stops inside record zero. The log is sequential append-only, so nothing can live behind it.
- consensus_pos_idx.is_empty(). The size prefix has no checksum of its own, so a flipped bit can inflate a record's extent past EOF and mimic a tear on a pack that really holds data. Every consensus read resolves through the position index.

Still fail-closed: a complete-but-corrupt meta (CRC or decode failure), and any tear with indexed outputs behind it.

A.2 — diagnostics

All three open doors name the pack file and epoch. The remediation is deliberately not uniform: deleting the epoch dir is safe from open_append (it recreates it) but bricks the node from open_append_exists/open_static (they don't), so only open_append carries the delete advice. I confirmed telcoin-network db validate accepts an epoch-NN directory before putting it in operator text.

MavenRain (approved):

PR #1228, opened 2026-08-20
fix(storage): enforce pack header version in PackIter/AsyncPackIter

### LINKED ISSUE #1227: PackIter and AsyncPackIter skip the pack header version check
## Problem

Every pack data file opens with a header that stamps a format version, and the pack layer treats it as a contract: `PackInner::open_data_file` rejects a file stamped newer than the version the caller supports (`LoadHeaderError::InvalidVersion`) and accepts older ones. The two iterator entries do not hold up their end. `PackIter::open` and `AsyncPackIter::open` check the header's appnum and nothing about the version, and `AsyncPackIter::open` carries a comment claiming it mirrors `open_data_file`'s version/appnum rejection, describing a check the code does not perform.

The gap bites on the peer path. `AsyncPackIter::open` is the entry for `ConsensusPack::stream_import`, which consumes epoch packs streamed by peers during epoch sync. A peer running a newer binary whose pack layout has moved on streams a file stamped with a future version, and the local reader walks it with current-format logic instead of rejecting it at the header: whatever goes wrong then goes wrong deep inside record decoding, on bytes the reader was never meant to interpret, with errors that point at the records rather than at the version mismatch that caused them. Local iteration is only incidentally protected, because `Pack::raw_iter` clones a data file whose header already passed `open_data_file`'s gate on open.

## Solution

Both opens should take the version the reader supports and enforce the `open_data_file` gate before the appnum check: reject files stamped newer, accept equal or older. It should be a parameter rather than a constant baked into the iterator, because the archive layer hosts multiple pack kinds with independent versioning (`open_partial` set that precedent). Accepting older is load-bearing rather than leniency: `stream_import` still decodes v0 legacy streams through a dedicated branch, and those must keep flowing through the same open.

Call sites should pin the version they mean. `stream_import` passes `PACK_VERSION`, which is where the new enforcement actually lands; `Pack::raw_iter` passes its own already-validated header version, so local iteration re-accepts exactly the file it came from and nothing loosens for the existing readers.

On the peer path the rejection should surface as a named read error that callers propagate. `stream_import` wraps open failures in `PackError::ReadError`, which `is_missing_static_files` does not classify as "epoch files absent", and that distinction is worth pinning in a test: a future-version pack from a peer must be reported as unreadable, never treated as an epoch that is simply missing.

Identified during review for #1219

grantkee:
- `PackIter::open` and `AsyncPackIter::open` take a new `version: u16` and reject files stamped newer with `LoadHeaderError::InvalidVersion` before the appnum check, the same gate as `PackInner::open_data_file`. Files stamped at or below the requested version stay readable; `stream_import`'s v0 legacy decode branch depends on older files being accepted.
- Version is a parameter, not a crate constant, because the archive layer hosts multiple pack kinds with independent versions.
- `Pack::raw_iter` passes its own already-validated `header.version()`, so local iteration re-accepts exactly the file it came from and existing readers see no behavioral change. `stream_import` passes `PACK_VERSION`, which is where the enforcement lands: the peer-facing path previously parsed unknown future layouts with current-format logic.
- The stale comment in `AsyncPackIter::open` that claimed version/appnum parity with `open_data_file` now matches the code.
- Three new tests: `test_pack_iter_rejects_newer_version` and `test_async_pack_iter_rejects_newer_version` stamp a file one version ahead and assert a newer file is rejected while the equal-version and newer-reader cases read the record; `test_stream_import_rejects_newer_pack_version` pins the peer-path surface, a `PackError::ReadError` containing "invalid version" with `is_missing_static_files()` false, so callers propagate the error instead of treating the epoch as absent.
- Mutation-checked: with the version guard removed all three tests fail at the enforcement assertion; with it restored `cargo nextest run -p tn-storage` passes 210. Workspace `cargo check` and clippy clean.

closes #1227

### DIFF
--- a/crates/storage/src/archive/pack.rs
+++ b/crates/storage/src/archive/pack.rs
@@ -488,7 +488,9 @@ where
     /// This iterator will not see any data in the write cache.
     fn raw_iter(&self) -> Result<PackIter<V, File>, LoadHeaderError> {
         let dat_file = { self.data_file.try_clone()? };
-        PackIter::open(dat_file, self.uid_idx)
+        // Our own header already passed the version gate on open, so requesting exactly its
+        // version re-accepts this file without loosening the iterator's check.
+        PackIter::open(dat_file, self.uid_idx, self.header.version())
     }
 }
 
@@ -892,7 +894,7 @@ mod tests {
             .create(false)
             .open(tmp_path.path().join("pack_test_one"))
             .unwrap();
-        let mut iter = PackIter::open(data_file, 0).unwrap().map(|r| r.unwrap());
+        let mut iter = PackIter::open(data_file, 0, 0).unwrap().map(|r| r.unwrap());
         let v: TestRec = iter.next().unwrap();
         assert_eq!(v.idx, 1);
         assert_eq!(v.name, "Value One");
@@ -1076,7 +1078,7 @@ mod tests {
         let (tmp_dir, _pos) = build_pack_with_decompression_bomb();
         let path = tmp_dir.path().join("pack_bomb");
         let file = File::open(&path).expect("open file");
-        let mut iter = PackIter::<TestRec, _>::open(file, 0).expect("iter open");
+        let mut iter = PackIter::<TestRec, _>::open(file, 0, 0).expect("iter open");
         match iter.next() {
             Some(Err(FetchError::RequestedDecompressSizeTooLarge(max))) => {
                 assert_eq!(max, MAX_RECORD_SIZE);
@@ -1092,7 +1094,7 @@ mod tests {
         let (tmp_dir, _pos) = build_pack_with_decompression_bomb();
         let path = tmp_dir.path().join("pack_bomb");
         let file = tokio::fs::File::open(&path).await.expect("open file");
-        let mut iter = AsyncPackIter::<TestRec, _>::open(file, 0).await.expect("iter open");
+        let mut iter = AsyncPackIter::<TestRec, _>::open(file, 0, 0).await.expect("iter open");
         match iter.next().await {
             Some(Err(FetchError::RequestedDecompressSizeTooLarge(max))) => {
                 assert_eq!(max, MAX_RECORD_SIZE);
@@ -1123,7 +1125,7 @@ mod tests {
         let (tmp_dir, _pos) = build_pack_with_corrupt_zstd_frame();
         let path = tmp_dir.path().join("pack_corrupt");
         let file = File::open(&path).expect("open file");
-        let mut iter = PackIter::<TestRec, _>::open(file, 0).expect("iter open");
+        let mut iter = PackIter::<TestRec, _>::open(file, 0, 0).expect("iter open");
         match iter.next() {
             Some(Err(FetchError::IO(_))) | Some(Err(FetchError::DeserializeValue(_))) => {}
             other => panic!("expected IO or DeserializeValue error, got {other:?}"),
@@ -1137,7 +1139,7 @@ mod tests {
         let (tmp_dir, _pos) = build_pack_with_corrupt_zstd_frame();
         let path = tmp_dir.path().join("pack_corrupt");
         let file = tokio::fs::File::open(&path).await.expect("open file");
-        let mut iter = AsyncPackIter::<TestRec, _>::open(file, 0).await.expect("iter open");
+        let mut iter = AsyncPackIter::<TestRec, _>::open(file, 0, 0).await.expect("iter open");
         match iter.next().await {
             Some(Err(FetchError::IO(_))) | Some(Err(FetchError::DeserializeValue(_))) => {}
             other => panic!("expected IO or DeserializeValue error, got {other:?}"),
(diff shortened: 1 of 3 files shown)


sstanfield (approved):

crates/storage/src/archive/pack.rs
```diff
@@ -488,7 +488,9 @@ where
     /// This iterator will not see any data in the write cache.
     fn raw_iter(&self) -> Result<PackIter<V, File>, LoadHeaderError> {
         let dat_file = { self.data_file.try_clone()? };
-        PackIter::open(dat_file, self.uid_idx)
+        // Our own header already passed the version gate on open, so requesting exactly its
+        // version re-accepts this file without loosening the iterator's check.
+        PackIter::open(dat_file, self.uid_idx, self.header.version())
```
MavenRain, line 493: This check is a tautology on the local path: the bound comes from the same header the gate compares against, and `open_data_file` already validated it, so the comparison is `v > v` and never fires.  Fine as documented, but if the intent is real defense in depth for `raw_iter` consumers, store the caller-requested version in `PackInner` next to `uid_idx` and pass that instead. This is merely optional polish.

crates/storage/src/archive/pack_iter.rs
```diff
@@ -45,8 +45,14 @@ where
...
     /// Produces an iterator over all the (key, values).  All and records
     /// are returned in insert order.
-    pub fn open(mut reader: R, uid_idx: u64) -> Result<Self, LoadHeaderError> {
+    /// Accepts files stamped at or below `version` and rejects newer ones with
+    /// [`LoadHeaderError::InvalidVersion`], the same gate as `PackInner::open_data_file`.
+    pub fn open(mut reader: R, uid_idx: u64, version: u16) -> Result<Self, LoadHeaderError> {
         let header = DataHeader::load_header(&mut reader, uid_idx)?;
+        if header.version() > version {
+            // Do not allow a newer version than we request but allow an older.
+            return Err(LoadHeaderError::InvalidVersion);
```
MavenRain, line 54: `LoadHeaderError::InvalidVersion` still renders as "invalid version (should be 0)" while `PACK_VERSION == 1`, and this new peer path surfaces that text verbatim to operators via `PackError::ReadError(e.to_string())`.  Consider an `(expected, got)` payload, e.g. `InvalidVersion { expected: u16, got: u16 }`, so the message is correct and the new test can assert on the pair instead of the prose.

crates/storage/src/consensus_pack.rs
```diff
@@ -2989,6 +2990,54 @@ pub(crate) mod test {
         drop(pack);
     }
 
+    /// A peer streaming a pack whose header is stamped NEWER than [`PACK_VERSION`] must be
```
MavenRain, line 2993: This covers only the reject direction.  The accept-older half is untested through `stream_import`: a gate written `!=` instead of `>` would pass every current test while breaking mixed-version import.  Suggest a companion test that builds a pack at version 0 (`open_append_version(..., 0)`), streams its data file through `ConsensusPack::stream_import`, and asserts the imported outputs match.

crates/storage/src/consensus_pack.rs
```diff
@@ -914,9 +914,10 @@ impl Inner {
     ) -> Result<Self, PackError> {
         let base_dir = path.as_ref().join(format!("epoch-{epoch}"));
         let _ = create_dir_synced(&base_dir);
-        let mut stream_iter = AsyncPackIter::<PackRecord, R>::open(stream, epoch as u64)
-            .await
-            .map_err(|e| PackError::ReadError(e.to_string()))?;
+        let mut stream_iter =
+            AsyncPackIter::<PackRecord, R>::open(stream, epoch as u64, PACK_VERSION)
```
MavenRain, line 918: Follow-up candidate, not blocking here: a version rejection arrives at the caller as `EpochPackAttempt::Failed` (primary/src/network/mod.rs:924-936), whose arm re-marks the peer `sync_capability = true`.  After a future `PACK_VERSION` bump, an old node re-probes newer peers on every sync attempt and burns `MAX_EPOCH_SYNC_PROBES` slots.  A distinct error variant would let epoch sync drop such peers without penalizing them (the `ReadError` catch-all also carries genuinely transient I/O).

crates/storage/src/consensus_pack.rs
```diff
@@ -2989,6 +2990,54 @@ pub(crate) mod test {
...
+        match &err {
+            PackError::ReadError(msg) => {
+                assert!(msg.contains("invalid version"), "unexpected message: {msg}");
+            }
+            other => panic!("expected ReadError, got {other:?}"),
+        }
+        // A version rejection means the peer's bytes are unreadable, not that the epoch files
+        // are absent on disk: the miss classifier must propagate it as an error.
+        assert!(
+            !err.is_missing_static_files(),
```
MavenRain, line 3036: Nit: given the `ReadError` match above, `is_missing_static_files` (which only matches `PackError::Open` + NotFound) cannot return true here, so this assert is near-tautological.

MavenRain (approved):
No blockers, just suggestions

PR #1230, opened 2026-08-21
Refactor the epoch vote certification code.

sstanfield:
This should stream line it, make it more DOS resistant and easier to recover if something goes wrong.  It also acknowledges that epoch records are deterministic and drops trying to "recover"- that is a fork.

### DIFF
--- a/crates/consensus/primary/src/tests/network_tests.rs
+++ b/crates/consensus/primary/src/tests/network_tests.rs
@@ -39,9 +39,9 @@ use tn_types::{
     encode, error::HeaderError, now, to_intent_message, AuthorityIdentifier, BlockHash,
     BlockHeader, BlockNumHash, BlsKeypair, BlsPublicKey, BlsSignature, BlsSigner as _, Certificate,
     CommittedSubDag, ConsensusHeaderDigest, ConsensusNumHash, ConsensusResult, Database, Epoch,
-    EpochDigest, EpochRecord, EpochSeedMessage, EpochVote, ExecHeader, Hash as _, HeaderDigest,
-    ReputationScores, Round, SealedHeader, TaskManager, TnReceiver as _, TnSender as _, VoteDigest,
-    VoteInfo, B256,
+    EpochCertificate, EpochDigest, EpochRecord, EpochSeedMessage, EpochVote, ExecHeader, Hash as _,
+    HeaderDigest, ReputationScores, Round, SealedHeader, TaskManager, TnReceiver as _,
+    TnSender as _, VoteDigest, VoteInfo, B256,
 };
 use tracing::debug;
 
@@ -1167,6 +1167,51 @@ async fn test_epoch_vote_wrong_digest_rejected_before_verify() -> eyre::Result<(
     assert!(
         tokio::time::timeout(Duration::from_millis(50), rx.recv()).await.is_err(),
         "a mismatched-digest vote must not reach the collector"
+    );
+    Ok(())
+}
+
+/// The other half of the ingress gate: a vote for an epoch we hold but that is ALREADY CERTIFIED
+/// (`get_epoch_by_number` returns `Some((record, Some(cert)))`) must be dropped silently — no
+/// error, no penalty, not forwarded. Certification is complete so further votes are surplus, and
+/// the benign drop must not be charged to the (honest) relayer on the gossip path. Companion to
+/// `test_epoch_vote_wrong_digest_rejected_before_verify`, which covers the digest arm.
+#[tokio::test]
+async fn test_epoch_vote_already_certified_epoch_dropped_before_verify() -> eyre::Result<()> {
+    let temp_dir = TempDir::new().unwrap();
+    let TestTypes { committee, handler, consensus_bus, task_manager: _task_manager, .. } =
+        create_test_types(temp_dir.path()).await;
+
+    // Seed the epoch-0 record, then store a certificate for it so the gate sees it as certified.
+    // This branch never verifies the cert, so a minimal one (a valid BLS signature reused from a
+    // vote, empty signer bitmap) is enough to flip `get_epoch_by_number`'s `None` -> `Some(cert)`.
+    let record = seed_uncertified_epoch0(&handler, &committee).await;
+    let auth = committee.authorities().next().expect("committee has authorities");
+    let key_config = auth.consensus_config().key_config().clone();
+    let vote = record.sign_vote(&key_config);
+    let cert = EpochCertificate {
+        epoch_hash: record.digest(),
+        signature: vote.signature,
+        signed_authorities: RoaringBitmap::new(),
+    };
+    handler
+        .consensus_chain()
+        .epochs()
+        .save_certificate(record.digest(), cert)
+        .await
+        .expect("store epoch-0 certificate");
+
+    // The vote is otherwise valid (correct digest, real committee member) — it is dropped ONLY
+    // because the epoch is already certified.
+    let mut rx = consensus_bus.subscribe_new_epoch_votes();
+    let res = handler.process_gossip(&epoch_vote_gossip(vote, 0)).await;
+    assert!(
+        res.is_ok(),
+        "a vote for an already-certified epoch must be dropped without error or penalty: {res:?}"
+    );
+    assert!(
+        tokio::time::timeout(Duration::from_millis(50), rx.recv()).await.is_err(),
+        "a vote for an already-certified epoch must not reach the collector"
     );
     Ok(())
 }
--- a/crates/node/src/manager/epoch_votes.rs
+++ b/crates/node/src/manager/epoch_votes.rs
@@ -9,8 +9,8 @@ use tn_config::KeyConfig;
 use tn_primary::{network::PrimaryNetworkHandle, ConsensusBusApp};
 use tn_storage::{consensus::ConsensusChain, epoch_records::EpochRecordDb};
 use tn_types::{
-    BlsAggregateSignature, BlsPublicKey, BlsSignature, Epoch, EpochCertificate, EpochDigest,
-    EpochRecord, EpochVote, Noticer, TaskSpawner, TnReceiver as _,
+    BlsAggregateSignature, BlsPublicKey, BlsSignature, Epoch, EpochCertificate, EpochRecord,
+    EpochVote, ShutdownNotifier, TaskSpawner, TnReceiver as _, TnSender as _,
 };
 use tokio::sync::{
     mpsc::{self, Receiver, Sender},
@@ -18,7 +18,7 @@ use tokio::sync::{
 };
 use tracing::{error, info, warn};
 
-type VoteQueue = VecDeque<(Epoch, Sender<EpochVote>, Option<Receiver<EpochVote>>)>;
+type VoteQueue = VecDeque<(Epoch, Sender<EpochVote>)>;
 
 /// Per-iteration wait for the next epoch vote in [`manage_epoch_votes`]'s collection loop.
 pub(crate) const EPOCH_VOTE_RECV_TIMEOUT: Duration = Duration::from_millis(2500);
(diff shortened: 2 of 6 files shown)


crates/node/src/manager/epoch_votes.rs
```diff
@@ -234,11 +220,15 @@ async fn manage_epoch_votes(
...
                                 target: "epoch-manager",
-                                "Over wrote expected epoch record {epoch_hash} with verified epoch record {new_epoch_hash}",
+                                "Network came to consensus on epoch record {new_epoch_hash} we expected epoch record {epoch_hash}, we have forked!",
                             );
-                            save_and_persist_with_logs(&db, new_epoch_rec, cert).await;
+                            // We generated a different record than the network certified, so we
+                            // have forked. A fork is not something a node can safely recover from
+                            // (records are deterministic, so ours is simply wrong) — fail-stop
+                            // rather than adopt the network's record and pretend to recover.
+                            node_shutdown.notify();
```
MavenRain, line 231: The fail-stop is the right call for a fork, but only this path reaches it, and only committee nodes get here.  Two things I traced in the worktree (all static, worth a second look):

1. `state_sync::spawn_epoch_record_collector` polls `request_epoch_cert` every 5 s on every node (state-sync/epoch.rs:31-77, :150-190).  On a forked node it validates the network (record, cert) and calls `epochs().save`, but `Inner::save_record` is first-write-wins by epoch slot (storage/epoch_records.rs:1275-1300, no digest comparison), so the network record is never written and the cert is appended as an orphan under the network digest (:1305-1322).  `update_finals` still overwrites `final_numbers[N]`.  `get_epoch_by_number(N)` stays `(forked, None)`, so the poller refetches and logs "retrieved cert for epoch N" every 5 s forever.
2. A forked node outside the committee never self-votes (:405) and the handler drops every network vote at the digest check (handler.rs:518-535), so `manage_epoch_votes` is never spawned for that epoch and this `notify()` is unreachable.  That node keeps running on the forked record with only the log loop above.  A forked committee node stops here, but on restart `revote_uncertified_epoch_record_on_startup` sees the forked latest record with no cert, re-arms, and the node stops again ~65 s later.  Nothing durable marks the fork.

Suggestion: make the fork decision in one place that every node class reaches.  The cheapest spot is `collect_epoch_records` / `Inner::save`: a validated (record, cert) for an epoch whose stored record has a different digest is the fork signal.  Here, persist a marker before `notify()` and `return` right after it, so a restart can refuse to run (or print the remediation) without a full ladder to re-detect.  `certified_record_by_epoch` (storage/epoch_records.rs:625) is documented as closing the first-write-wins gap but has no caller outside its own file.

crates/node/src/manager/epoch_votes.rs
```diff
@@ -304,19 +289,38 @@ async fn handle_new_vote(vote: EpochVote, vote_queues: &mut VoteQueue) {
...
+        let Some((epoch_rec, None)) =
+            consensus_chain.epochs().get_epoch_by_hash(vote.epoch_hash).await
+        else {
+            // Missing the record or it is certified.  These were pre-checked when the gossip came
+            // in so this should not happen.
+            return;
+        };
+        // Spawn the vote collector in response to a vote if it was missing.
+        // This allows the possibility of recovering a cert with a republished vote even if stale.
+        task_spawner.spawn_task(format!("epoch votes for epoch {}", epoch_rec.epoch), async move {
```
MavenRain, line 304: Before this PR a collector was spawned once per local watch event and `handle_new_vote` refused votes outside `latest + 1`.  Now any valid vote for any record we still hold uncertified spawns `manage_epoch_votes` (the "even if stale" case above).  Two bounds are missing:

- `TaskSpawner::spawn_task` returns no handle (types/task_manager.rs:163-168), so the `pop_front` eviction at :320-321 only drops the `Sender`.  The task runs on to the end of its ladder (24 x 2.5 s) plus up to 5 x 3 `request_epoch_cert` RPCs (network/mod.rs:755-774).  A later vote for the same epoch then spawns a second collector for it, and the handler doc at handler.rs:276-277 ("the collector dedups per signer downstream") no longer holds once the entry is evicted.
- There is no epoch window, so a republished vote for an old uncertified epoch is enough to start a fresh 65 s task.

In practice the handler LRU (`MAX_EPOCH_VOTES` = 1024, per (author, epoch)) and committee-only authorship keep this small, so low risk.  Suggestion: keep an `AbortHandle` in each `vote_queues` entry and abort on eviction, and limit spawn to the current epoch and N-1 (the failsafe target).

crates/node/src/manager/epoch_votes.rs
```diff
@@ -74,37 +80,24 @@ async fn manage_epoch_votes(
...
     loop {
         tokio::select! {
+            _ = &shutdown => return,
             result = tokio::time::timeout(timeout, vote_rx.recv()) => {
                 match result {
                     Ok(None) => break,  // Channel closed- we are done.
                     Ok(Some(vote)) => {
+                        if vote.public_key == me {
+                            // Record our vote if we see it for this epoch so we can revote if/when needed.
+                            my_vote = Some(vote);
```
MavenRain, line 99: Defense-in-depth only (every producer today keeps `epoch`, `epoch_hash`, and the record consistent, and the handler enforces `record.digest() == vote.epoch_hash`), but two keys are taken from the vote before this collector validates it:

- `my_vote` is captured here, above the `vote.epoch != epoch_rec.epoch` check and the digest arm, although the comment says "for this epoch".  Whatever lands here is what the timeout and at-quorum republishes gossip.
- In `handle_new_vote` the queue entry is keyed by the unsigned `vote.epoch` (:323) while the collector is built from the record found by `vote.epoch_hash` (:295-296).

Suggestion: `if vote.public_key == me && vote.epoch_hash == epoch_hash { my_vote = Some(vote); }` after the epoch check, and push `(epoch_rec.epoch, epoch_vote_tx)` captured before the move.

crates/node/src/manager/epoch_votes.rs
```diff
@@ -173,6 +153,12 @@ async fn manage_epoch_votes(
             target: "epoch-manager",
             "reached quorum on epoch close for {}/{epoch_hash}", epoch_rec.epoch
         );
+        // Republish our vote one final time.  This is not strictly needed but if we were the first
+        // validator to close the epoch and we got no timeouts the laggy validators may have
+        // missed our vote- give them one more chance.
+        if let Some(vote) = my_vote {
+            let _ = primary_network.publish_epoch_vote(vote).await;
```
MavenRain, line 160: This republish sits between reaching quorum and `aggregate` / `save_certificate` / `persist`.  It is a round trip through the network command channel, and the comment says it is not strictly needed.  A slow network task delays the durable step by that amount, and a crash in the widened window loses an assembled quorum (the respawn then re-collects from scratch).

Suggestion: aggregate and persist first, then republish, or spawn the republish as a task like the delayed initial publish at :414-421.

crates/node/src/manager/epoch_votes.rs
```diff
@@ -304,19 +289,38 @@ async fn handle_new_vote(vote: EpochVote, vote_queues: &mut VoteQueue) {
...
-        }
+        // We do not have a collector for this epoch so start one and send it this vote.
         let (epoch_vote_tx, epoch_vote_rx) = mpsc::channel(10_000);
+        // If we recieve a valid vote and aren't collecting votes to certify then start.
+        let Some((epoch_rec, None)) =
+            consensus_chain.epochs().get_epoch_by_hash(vote.epoch_hash).await
+        else {
+            // Missing the record or it is certified.  These were pre-checked when the gossip came
+            // in so this should not happen.
+            return;
```
MavenRain, line 300: Our own vote now goes through this path, so a record that is not yet in the DB (or already certified) drops it with no log and no collector.  Production ordering is right (`save_record` -> `persist` -> `epoch_record_watch`), and the test edits in this PR show the dependency (each test now seeds `save_record` + `persist` before `send_replace`).  Since the comment says this should not happen, a `warn!` with the epoch and digest would make the failure visible.  Same for the `let _ =` on the bus sends at :388 and :412: a failed send of our own vote is worth a log line.

crates/node/src/manager/epoch_votes.rs
```diff
@@ -354,32 +359,69 @@ pub(crate) fn spawn_epoch_vote_collector(
...
+                        // republishes cover any that still miss it.
+                        tokio::time::sleep(INITIAL_VOTE_PUBLISH_DELAY).await;
+                        let _ = primary_network_clone.publish_epoch_vote(epoch_vote).await;
                         Ok(())
-                    },
-                );
+                    });
+                }
             }
+            // See spawn_epoch_record_collector() for how syncing nodes can keep up there epoch certs if not following the tip.
```
MavenRain, line 424: Small doc items left by the refactor:

- "there epoch certs" -> "their epoch certs" here; "recieve" (:294), "committe" (:392), "it's task" (:265, "its").
- The `revote_uncertified_epoch_record_on_startup` docstring below still says "The vote collector is armed only by `epoch_record_watch`" (now armed by the first vote on the bus) and that it "heals the latest stored record only" (the N-1 failsafe at :370-395 now also heals the record before the latest).  The `CERTIFIED_ANCHOR_WAIT` reference at :437 names a constant that does not exist in the tree (pre-existing, not from this PR).
- storage/epoch_records.rs: the single-round-trip rationale for `EpochByHash` is stated three times (:75-78, :885-887); one reference is enough.

crates/consensus/primary/src/tests/network_tests.rs
```diff
@@ -1171,6 +1171,51 @@ async fn test_epoch_vote_wrong_digest_rejected_before_verify() -> eyre::Result<(
...
+    let auth = committee.authorities().next().expect("committee has authorities");
+    let key_config = auth.consensus_config().key_config().clone();
+    let vote = record.sign_vote(&key_config);
+    let cert = EpochCertificate {
+        epoch_hash: record.digest(),
+        signature: vote.signature,
+        signed_authorities: RoaringBitmap::new(),
+    };
+    handler
+        .consensus_chain()
```
MavenRain, line 1198: Good test, and it does pin the property (kp1 is needed to reach super_quorum 3 of 4, so a mode gate would fail it).  The new paths that still have no test, in risk order:

1. The fail-stop branch (`node_shutdown.notify()` on a digest mismatch at :221-231), which is the headline behaviour change.  The `PrimaryNetworkHandle::new_for_test` mock can answer `RequestEpochCert` with a different anchored record, and a `ShutdownNotifier` subscriber can assert the stop.
2. Dead-collector respawn in `handle_new_vote` (send fails -> entry removed -> new spawn).
3. A vote for a certified or unknown record dropped in `handle_new_vote`.
4. The at-quorum republish and the 500 ms delayed publish (assert two publishes reach the mock network).
5. `my_vote` captured from the stream and republished on a timeout.

MavenRain (dismissed):
GTG, just some considerations

MavenRain (approved):

crates/node/src/manager/epoch_votes.rs
sstanfield: I added a check to see if vote_rx closed in the network download loop.  If the sender closes then the task should end, I think that will handle lingering tasks well enough.

The second item is NOT an issue, that is by design.  It is supposed to make it easier to recover in case we do wind up with a gap.  So unless you see an actual issue this creates I am not changing it.

crates/node/src/manager/epoch_votes.rs
sstanfield: I am planning to split this into a new PR.  It may be larger than expected and we might want a general "node is in a forked/failed state" system.  Should never need it but...

crates/consensus/primary/src/tests/network_tests.rs
sstanfield: More tests added.

PR #1232, opened 2026-08-21
fix(tn-reth): answer eth_syncing from consensus catch-up state (#1231)

### LINKED ISSUE #1231: eth API methods served by stock reth handlers diverge from TN consensus (follow-up to #1176)
## Summary

PR #1176 replaced `eth_sendRawTransaction` / `eth_sendRawTransactionSync` on the built RPC server because the stock reth handlers did not enforce a TN flag.  The same review (grantkee on #1176) noted: "There are some other eth api methods we likely want to intercept given the nature of TN BFT consensus." This issue is that audit.  It lists the eth namespace methods whose stock reth handlers answer with Ethereum consensus assumptions that TN does not satisfy, the methods that were checked and are already correct, and the interception pattern each fix must repeat.

References are pinned to PR #1176's branch (`fix/1160-enforce-txfeecap` @ c5e0dcea, main merged at 5f1c0b49) and reth v1.11.3 (`d6324d63`, the lockfile pin).

### 1.  `eth_feeHistory`: the next-block base fee entry uses Ethereum's EIP-1559 step

Reth appends one extra entry to `baseFeePerGas` for the block after the newest returned block, computed as `chain_spec.next_block_base_fee(last_header, timestamp)` (reth `crates/rpc/rpc-eth-api/src/helpers/fee.rs:175-181`, second site at 229-237).  That is mainnet's EIP-1559 schedule: the predicted fee moves whenever the last block is off the 50% gas target.

TN does not derive base fees that way.  Worker base fees come from the on-chain `WorkerConfigs` contract and apply per epoch (`crates/tn-reth/src/env/epoch.rs:590` `worker_fee_configs_inner`;  `crates/types/src/gas_accumulator.rs` `BaseFeeContainer`, adjusted only at epoch close).  Within an epoch the real next-block base fee is the current one, flat.  TN's own code already documents this exact mismatch for a test helper: `RethEnv::get_gas_price` says its EIP-1559 computation "is NOT how TN derives base fees" (`crates/tn-reth/src/env/helpers.rs:219-224`).

Impact: wallets derive `maxFeePerGas` from exactly this entry.  After a full block the predicted fee is up to 12.5% above the real next-block fee;  after an empty block, 12.5% below, and a transaction priced from the low quote waits until its sender re-prices.  No attacker profit;  it is a correctness gap in the fee quote API.

Fix shape: intercept `eth_feeHistory`, delegate to the stock handler, then overwrite the final `baseFeePerGas` entry with the worker's current epoch base fee.  The worker's `BaseFeeContainer` has to be plumbed into `get_rpc_server` the same way the fee cap was in #1176.  Note the per-worker dimension: each worker has its own container (`GasAccumulator::base_fee(worker_id)`), so the multi-worker RPC track (#554-#559) needs each worker's server to hold its own handle.

### 2.  `eth_syncing`: hardcoded `false`

`WorkerNetwork::is_syncing` and `is_initially_syncing` return `false` unconditionally (`crates/tn-reth/src/worker.rs:160-166`).  A TN node that is catching up on consensus output (state sync, epoch catch-up) still reports fully synced.  Load balancers, wallets, and exchange infra gate traffic on `eth_syncing`;  on TN the gate never closes, so a syncing node serves stale `latest` state as if current.

Fix shape: this one does not need an RPC intercept.  Wire TN's sync state into the `WorkerNetwork` shim (a watch/flag set by the consensus sync path) so the stock handler answers correctly.  An intercept is the fallback if the sync signal cannot reach the shim cleanly.

### 3.  Blob fee surface: TN execution ignores EIP-4844, RPC quotes blob fees anyway

TN's EVM env pins `excess_blob_gas: 0` and `blob_gasprice: u128::MAX` with the comment "eip4844 transactions are ignored" (`crates/tn-reth/src/evm/config.rs:178-181`).  But the stock handlers still serve `eth_blobBaseFee` and `eth_feeHistory`'s `baseFeePerBlobGas` from header math, quoting blob space as available and cheap.  A client that trusts the quote submits a blob transaction that TN will never execute as quoted.

Fix shape: decide the contract first (does TN reject 4844 at the pool boundary, or accept and ignore?), then make `eth_blobBaseFee` / the blob entries report it honestly.  This needs its own small design note before code.

### 4.  Pending-block env placeholders

`BuildPendingEnv for TNPayload` fills the pending env with `beneficiary: Address::ZERO`, `worker_id: 0`, `timestamp: parent + 1`, and the parent's base fee (`crates/tn-reth/src/payload.rs:184-203`).  So `eth_call` / `eth_estimateGas` / `eth_getBlockByNumber` on the `pending` tag simulate a block no TN worker would produce: zero beneficiary (COINBASE-reading contracts see address zero), single-worker view, and a base fee that is stale across an epoch boundary.  This is an approximation by design;  the ask is to document it on the flag/API surface and decide whether `PendingBlockKind::None` should be TN's default (the test helper at `crates/tn-reth/src/env/rpc.rs:334` already uses it).

### 5.  `NetworkStatus` placeholders (`eth_protocolVersion`, `admin_nodeInfo`)

`network_status` hardcodes `protocol_version: 1` and `head: Default::default()` (zero hash) (`crates/tn-reth/src/worker.rs:143-152`).  Cosmetic, but any tooling that reads the head from node info gets a zero hash.  Lowest priority;  fix opportunistically when touching the shim for item 2.

## Checked and already correct

- `safe` / `finalized` block tags: `finalize_block` drives `set_finalized` / `set_safe` on the provider (`crates/tn-reth/src/env/execution.rs:209-215`).
- `eth_gasPrice` / `eth_maxPriorityFeePerGas`: reth adds the suggested tip to the latest header's own base fee, no EIP-1559 stepping (reth `fee.rs:364-368`), so the answer is right within an epoch.  Only the epoch-boundary block can be misquoted, same window as item 1's fix.
- Uncle methods: reth returns empty;  correct, TN has no uncles.
- `eth_chainId` / `net_version`: wired to the chain spec (`crates/tn-reth/src/worker.rs:155-157`).
- `eth_sendTransaction`: unreachable, TN builds `EthApi` with no signers (documented in #1176).
- Raw submission fee cap: #1176.

## The interception pattern (from #1176)

Every intercepted method must repeat what the fee-cap guard does, so operators and clients cannot tell an intercepted method from a stock one except by the corrected answer:

1.  A jsonrpsee `#[rpc(server, namespace = "eth")]` trait whose method names match reth's `EthApiServer` exactly.
2.  Installed with `add_or_replace_if_module_configured(RethRpcModule::Eth, ...)` so the operator's transport module selection is respected, and the install failure is an error, not a log line (`crates/tn-reth/src/env/rpc.rs:124-127`).
3.  Repeat reth's request-trace line verbatim (`tracing::trace!(target: "rpc::eth", ...)`) before doing anything else;  operators grep that target (`crates/tn-reth/src/rpc_fee_cap.rs:134-137`).
4.  Reuse reth's exact error objects and error precedence wherever the intercept decides an error itself.
5.  Delegate to the unchanged stock handler for everything the intercept does not correct.
6.  Tests go through `get_rpc_server` and call the registered method over a real transport, so the production registration is what is tested.

## Suggested order

Item 1 (fee history) and item 2 (syncing) first;  both are quote/health surfaces external infra consumes today.  Item 3 needs a design decision.  Items 4 and 5 are documentation-or-opportunistic.

MavenRain:
Closes #1231 (items 2 and 5 of the audit;  items 1, 3, and 4 land in their own PRs).

## Problem

`eth_syncing` on a TN node always answers `false`: `WorkerNetwork::is_syncing` and `is_initially_syncing` are hardcoded (`crates/tn-reth/src/worker.rs`).  A validator that has fallen behind and is streaming missed consensus output to rejoin the committee still reports fully synced.  Load balancers, wallets, and exchange infra gate traffic on `eth_syncing`;  on TN the gate never closes, so a catching-up node serves stale `latest` state as if current.  No attacker is required: an ordinary crash-recovery window misroutes traffic to the stale node.

The same shim also leaves the `network_status` head defaulted to the zero hash and hardcodes `protocol_version: 1` without explanation (item 5 of the audit, fixed here opportunistically as the issue suggests).

## Fix

Wire the node's consensus catch-up state into the shim so the stock reth handler answers correctly.  No RPC intercept: reth's `eth_syncing` derives its whole answer from `NetworkInfo::is_syncing` (reth `rpc-eth-api/src/helpers/spec.rs:46-72` at the v1.11.3 pin), so the flag is the complete surface.

- [x] `crates/tn-reth/src/worker.rs`: `WorkerNetwork` holds `SyncFlags` (same `Arc<RwLock>` shape as the peer-count field).  `set_syncing` records the state;  `is_syncing` reads it back.  The first caught-up report latches `completed_initial_sync`, so `is_initially_syncing` distinguishes the first catch-up of the process from a later mid-epoch fall-behind.  Also adds `new_for_test` (the same constructor PR #1176 adds on its branch) so tests can build the shim without a live worker network handle.
- [x] `crates/node/src/manager/node/start_epoch.rs`: `node_mode_is_syncing`, an exhaustive match on `NodeMode` (a new variant fails to compile until this decision is made for it): `CvvInactive` is the demoted catch-up mode, so it is the one mode that reports syncing.  `CvvActive` is optimistic-current by definition (the node is demoted when that turns out false).  An epoch-scoped task subscribes to the `ConsensusBus::node_mode` watch and pushes every change into the shim;  it dies with the epoch task manager and is respawned on rollover alongside the shim's peer-count task, re-reading the mode `identify_node_mode` just published.
- [x] `crates/node/src/engine/{mod,inner}.rs`: `set_workers_syncing` fans the state out to every worker's shim.
- [x] `crates/tn-reth/Cargo.toml`: dev-only `jsonrpsee` `server`/`macros` features so the new test calls through the production RPC registration.
- [x] **Item 5** (`crates/tn-reth/src/worker.rs` + `crates/node/src/engine/inner.rs`): the shim holds the execution env (`Option<RethEnv>`, `None` only in tests) and answers the `network_status` head from `canonical_tip().hash()`;  the test constructor falls back to the genesis hash, the true tip of a fresh chain.  `protocol_version: 1` stays but is now documented as a fixed placeholder: TN speaks no eth wire protocol, and `capabilities` is honestly empty.

The mode watch is the signal the consensus sync path already maintains: demotion to `CvvInactive` happens where the node detects it is outside the GC window ("we are behind, go to catchup mode", `primary/src/network/handler.rs`), and promotion back to `CvvActive` happens only when catch-up reaches a consensus header fresh within 5 seconds (`executor/src/subscriber.rs`).  An `Observer` follows consensus output continuously and has no behind/caught-up signal today;  it keeps answering not-syncing, unchanged from before this PR.  A consensus-height-gap signal that would cover observer catch-up is left as a follow-up on #1231.

## Testing

- `crates/tn-reth/src/worker.rs`: unit test for the flag lifecycle, including the initial-sync latch surviving a later fall-behind.
- `crates/tn-reth/src/env/rpc.rs`: builds the production RPC server over a temp env via `get_rpc_server`, calls the registered `eth_syncing` through its methods, and watches the answer flip: `false` at rest, a sync-progress object (with `currentBlock == highestBlock`, reth's shape when only the flag is known) while the shim reports syncing, `false` again after.
- `crates/node/src/manager/node/start_epoch.rs`: exhaustive `node_mode_is_syncing` expectations per mode.
- Item 5 (`crates/tn-reth/src/worker.rs`): `network_status` over a temp execution env answers the canonical tip's hash as its head (asserted non-zero, so the old defaulted head fails the test);  the env-less test constructor answers the genesis hash.
- Local static ladder green (nightly fmt check, scoped check/clippy with `--no-deps`, scoped unit tests under the pinned queue discipline);  each new load-bearing test confirmed by in-diff mutation (flag read reverted to the hardcoded `false`, head reverted to the default, tests fail, fix restored).  CI's workspace lanes and the pinned `+1.94` attest run on push.

Out of scope (tracked in #1231): item 1 (`eth_feeHistory` next-block base fee entry) depends on the interception infrastructure PR #1176 introduces and stacks on it once that merges;  item 3 (blob fee surface) needs its design decision;  item 4 (pending-block default) is its own small PR.  Note for review: #1176 also touches `crates/tn-reth/src/worker.rs` (it adds the same `new_for_test`);  whichever lands second resolves a one-constructor overlap (now a one-line field-init difference).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -111,12 +111,13 @@ jobs:
       - name: Run tests
         run: cargo nextest run --locked --workspace --exclude tn-faucet --no-fail-fast --profile ci
 
-  # Adiri feature lane: the adiri-gated suites (the legacy header wire-format pins in
-  # tn-types, the consensus-registry fork tests in tn-reth, and the pre-fork entry-fee pin
-  # in tn-node) only compile under the adiri features, so the default-feature test job
-  # above never builds or runs them. tn-storage is excluded: two of its consensus_pack
-  # tests (dup_batch_across_certs, v0_shared_batch_served_as_v1_bytes) fail under the
-  # adiri features today, independent of this lane.
+  # Adiri feature lane: the adiri-gated suites only compile under the adiri features, so
+  # the default-feature test job above never builds or runs them - the legacy header
+  # wire-format pins and the frozen pre-fork committee vectors in tn-types, the
+  # consensus-registry fork tests in tn-reth, the pre-fork entry-fee pin in tn-node, the
+  # frozen pre-fork consensus-pack fixtures in tn-storage, and the pre-fork genesis
+  # rejection in tn-config. Every crate that forwards `adiri` and has adiri-gated tests
+  # belongs on this line: one left off silently stops testing its pre-fork lane.
   adiri-test:
     needs: skip-maintainers
     if: >-
@@ -144,7 +145,7 @@ jobs:
           cache-on-failure: "true"
 
       - name: Run adiri-gated tests
-        run: cargo nextest run --locked -p tn-types -p tn-reth -p tn-node --features tn-types/adiri,tn-reth/adiri,tn-reth/test-utils,tn-node/adiri --no-fail-fast --profile ci
+        run: cargo nextest run --locked -p tn-types -p tn-reth -p tn-node -p tn-storage -p tn-config --features tn-types/adiri,tn-reth/adiri,tn-reth/test-utils,tn-node/adiri,tn-storage/adiri,tn-config/adiri --no-fail-fast --profile ci
 
   # Guard: rocksdb (librocksdb-sys, the entire RocksDB C++ engine and the single largest
   # native compile in the tree) must stay out of the default feature graph. It is gated
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3142,6 +3142,7 @@ dependencies = [
  "tn-config",
  "tn-node",
  "tn-reth",
+ "tn-storage",
  "tn-test-utils",
  "tn-types",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11883,6 +11883,7 @@ version = "0.1.0"
 dependencies = [
  "alloy",
  "alloy-evm",
+ "async-trait",
  "clap",
  "dirs-next",
  "eyre",
--- a/Makefile
+++ b/Makefile
@@ -144,19 +144,29 @@ E2E_BIN := $(E2E_TARGET_ROOT)/e2e/telcoin-network
 # seed_signature_override_is_inert_when_unset requires a process WITHOUT the variable.
 TN_SEED_SIGNATURE_FORK_EPOCH ?= 4294967295
 
+# Multi-workers fork epoch for the e2e lanes (#554). Same shape as the seed-signature
+# variable above and armed independently of it: defaults to u32::MAX so the default lanes run
+# the fork DORMANT (legacy single-worker committee layout, what adiri carries on disk);
+# non-adiri builds are otherwise active from genesis. Override for a fork-active lane:
+#   TN_MULTI_WORKERS_FORK_EPOCH=1 make test-epochs
+# Only test-utils builds consult it (tn_types::forks::multi_workers_fork_epoch_override).
+# Set only on name-filtered nextest lines, never a bare --workspace run: tn-types'
+# multi_workers_override_is_inert_when_unset requires a process WITHOUT the variable.
+TN_MULTI_WORKERS_FORK_EPOCH ?= 4294967295
+
 # run restart integration tests
 test-restarts: build-e2e-bin
-	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) cargo nextest run --run-ignored all test_restarts ;
+	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) TN_MULTI_WORKERS_FORK_EPOCH=$(TN_MULTI_WORKERS_FORK_EPOCH) cargo nextest run --run-ignored all test_restarts ;
 
 # run epoch integration tests (same filter as the public-tests epoch line). The scheduled
 # Durable e2e lane (#1149) runs this and test-restarts with TN_TEST_MDBX_SYNC=durable
 # exported so every spawned node opens MDBX in Durable.
 test-epochs: build-e2e-bin
-	TN_BIN_PATH="$(E2E_BIN)" TN_SEED_SIGNATURE_FORK_EPOCH=$(TN_SEED_SIGNATURE_FORK_EPOCH) cargo nextest run -p e2e-tests --test it --run-ignored all test_epoch ;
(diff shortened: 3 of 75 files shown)


grantkee (approved):
LGTM

sstanfield (approved):

PR #1235, opened 2026-08-21
fix(tn-reth): serve eth_feeHistory's next-block base fee from the epoch schedule (#1231)

MavenRain:
Toward #1231 (items 1 and 3 of the audit: the `eth_feeHistory` next-block base fee, and the blob-fee quote surface).  Follow-up to #1176;  sibling of #1232 (item 2).

## Problem

Reth's stock `eth_feeHistory` appends one extra `baseFeePerGas` entry for the block after the newest returned block and computes it with Ethereum's EIP-1559 step (`next_block_base_fee`, reth `crates/rpc/rpc-eth-api/src/helpers/fee.rs:175-181` and `229-237` at the lockfile pin `d6324d63`).  TN does not derive base fees that way: each worker's base fee comes from the on-chain `WorkerConfigs` contract and is flat within an epoch.  The predicted entry is therefore wrong whenever the newest returned block is off the EIP-1559 gas target: up to 12.5% above the real next-block fee after a full block, 12.5% below after an empty one.  Wallets derive `maxFeePerGas` from exactly this entry, so a transaction priced from the low quote waits until its sender re-prices.  No attacker is required;  every client that trusts the quote misprices.

The blob surface has the same shape (item 3, second commit).  TN's pool refuses EIP-4844 transactions at admission (#1159), so there is no blob market.  But TN headers pin `excess_blob_gas` to zero, and the stock handlers quote that as the EIP-4844 minimum: `eth_blobBaseFee` answers 1 wei and `eth_feeHistory` fills `baseFeePerBlobGas` with 1s — an advertisement of cheap blob space the pool refuses to sell.  The contract decision the issue asked for is already made by #1159 (reject at the pool boundary);  this PR makes the quotes match it.  Zero is the answer reth itself serves for a chain without blob fields (pre-Cancun headers take `unwrap_or_default` through the same paths), so zero is what a client library already understands as "no blob market".

## Fix

Intercept `eth_feeHistory` with the #1176 pattern: delegate to the unchanged reth handler, then overwrite only the final `baseFeePerGas` entry with TN's real answer.

- [x] `crates/tn-reth/src/rpc_fee_history.rs` (new): a jsonrpsee `eth`-namespace trait whose method name matches reth's `EthApiServer` exactly, repeating reth's `rpc::eth` request-trace line verbatim before delegating.  The corrected entry is (a) the recorded header base fee of block `newest + 1` when that block is already in the database (historical ranges), else (b) the worker's current epoch base fee from the shared `BaseFeeContainer`.  The database read has no in-memory overlay, so a not-yet-persisted block falls back to the container value;  within an epoch the two agree, and the answer self-heals once the block persists.  Empty histories (`blockCount` 0) pass through untouched, as do all other fields except the blob columns.
- [x] Item 3, same intercept (second commit): the handler zeroes `baseFeePerBlobGas` and `blobGasUsedRatio` after the delegate returns (lengths preserved, so the response shape stays exactly reth's), and the trait gains `eth_blobBaseFee` answering a constant zero with reth's verbatim trace line.  No delegation for `eth_blobBaseFee`: the honest answer is a constant, so no provider fault can change it.
- [x] `crates/tn-reth/src/env/rpc.rs`: `get_rpc_server` takes the worker's `BaseFeeContainer` and installs the corrected method with `add_or_replace_if_module_configured(RethRpcModule::Eth, ...)`, returning an error instead of logging on install failure.  Coverage is structural: the same module config drives the module build and the replacement, so every transport that serves the eth namespace gets the corrected method.
- [x] `crates/node/src/engine/inner.rs`, `crates/node/src/engine/mod.rs`, `crates/node/src/manager/node/start_epoch.rs`: `initialize_worker_components` now receives the live container from `gas_accumulator.base_fee(worker_id)` instead of a `u64` snapshot;  the pool still initializes from a snapshot of it.  The container clone stays live across epochs (slot resizes are index-stable and epoch-boundary adjustments store into the same `Arc`), so the RPC answer tracks each epoch's fee without re-running worker init.  Each worker's server holds its own worker's container, which is the fee its clients price against (multi-worker track #554-#559).

## Testing

Static ladder (all via `gateledger run`): nightly `fmt` + `fmt -- --check`, `cargo check -p tn-reth -p tn-node --tests`, `cargocho clippy -p tn-reth -p tn-node --all-targets --no-deps -D warnings`, all green.

- [x] `cargo test -p tn-reth --lib rpc_fee_history::tests`: 7 unit tests on the corrections (next-block arithmetic, recorded-fee precedence, tip fallback, empty-history no-op, blob-column zeroing with lengths and the gas column preserved).
- [x] `cargo test -p tn-reth --lib env::rpc::tests`: 6 transport tests through the production `get_rpc_server` registration (`methods_by`), asserting the final entry equals the container fee for both `latest` and `pending`, that returned-block entries stay untouched, that both blob columns come back zero (the stock values for these headers are 1 wei, so the zeros are the intercept's), and that `eth_blobBaseFee` answers zero.
- [x] `cargo test -p tn-node --test it worker_pool_base_fee`: the pool still initializes from the container snapshot.
- [x] Mutation-confirmed, six mutations killed: dropping the overwrite, ignoring the recorded header fee, correcting the wrong index, dropping the blob-column zeroing, answering `eth_blobBaseFee` with the stock 1 wei, and skipping the ratio column each make the new tests fail;  restores byte-verified.

Dynamic tier: CI on push;  pinned `+1.94` attest on the second machine.

## Merge coordination

`crates/tn-reth/src/env/genesis.rs` (RPC-args test seam), the `WorkerNetwork::new_for_test` hunk in `worker.rs`, and the `Cargo.toml` jsonrpsee/async-trait hunks are taken verbatim from open #1176 (also replicated by open #1232), so those files merge cleanly whichever PR lands first.  `env/rpc.rs` overlaps #1176's fee-cap install and #1232's test module;  the conflicts are mechanical (adjacent installs and adjacent tests).  `start_epoch.rs` and `engine/inner.rs` overlap #1232 in adjacent regions.


crates/node/src/engine/inner.rs
```diff
@@ -137,7 +138,7 @@ impl ExecutionNodeInner {
         worker_id: WorkerId,
         network_handle: WorkerNetworkHandle,
         engine_to_primary: EP,
-        base_fee: u64,
+        base_fee: BaseFeeContainer,
```
grantkee, line 141: `initialize_worker_components` runs once per node lifetime and moves a `BaseFeeContainer` clone into the RPC server, which holds it forever. `GasAccumulator::set_num_workers` truncates slots on shrink and creates fresh default containers on regrow, so for any worker id ≥ 1 a shrink→regrow across epoch boundaries leaves the still-running RPC server holding a container that `EpochBaseFees::apply` no longer writes to — `eth_feeHistory` would quote a permanently stale fee for that worker. Unreachable today: only `DEFAULT_WORKER_ID` (0) is initialized, the count is clamped to ≥ 1, and grows are index-stable, so slot 0's container is stable for the node lifetime. The multi-worker foundation ( #554 / #1219 ) is actively landing, and the comment at `start_epoch.rs:439` ("otherwise, loop here") marks this exact call site as the future multi-worker loop.

grantkee (approved):
LGTM - noted one change that may affect multi-worker support, but not a blocker

crates/node/src/engine/inner.rs
MavenRain: Good callout . . . create a follow-up issue for this: https://github.com/Telcoin-Association/telcoin-network/issues/1282

PR #1237, opened 2026-08-21
fix(tn-reth): default the RPC pending block kind to none (#1231)

MavenRain:
Addresses the pending-block item (item 4) of #1231;  the other items land in their own PRs (#1232 and the feeHistory branch).

## Problem

TN's `RpcServerArgs` has no pending-block flag, so reth's clap default (`full`) wins and `pending`-tag RPC queries answer from a simulated pending block that no worker ever produces.  Its env (`BuildPendingEnv for TNPayload`, `crates/tn-reth/src/payload.rs`) is built from placeholder approximations: the beneficiary is the zero address (COINBASE-reading contracts see address zero), the worker id is 0, the timestamp is parent + 1, and the base fee is the parent's, which goes stale across an epoch boundary (TN fees are epoch-flat).  No attacker is required: any `pending`-tag caller reads these placeholder values as if real.

## Fix

- [x] `crates/tn-reth/src/rpc_server_args.rs`: add `--rpc.pending-block` (`rpc_pending_block: PendingBlockKind`) with a TN default of `none`, set it in the manual `Default` impl, and carry it through the `From` conversion into reth's args.  With `none`, reth's pending-block builders return no pending block, so `pending`-tag queries answer null or fall back to the latest state . . . honest answers instead of placeholder ones.  An operator who wants the simulated block opts in with `--rpc.pending-block full`, which survives natively.
- [x] `crates/tn-reth/src/payload.rs`: document the placeholder approximations on the `BuildPendingEnv` impl and point at the flag.

## Testing

- `pending_block_defaults_none_and_explicit_full_survives` (`rpc_server_args.rs`): the clap default and the manual `Default` impl agree on `none`, and an explicit `--rpc.pending-block full` survives the parse and the TN-args-to-reth-args conversion into `EthConfig`.
- `test_tn_rpc_args_reach_eth_config` (`env/rpc.rs`) gains a `pending_block_kind == PendingBlockKind::None` assert.  Only the conversion carrying TN's field can produce `None` here, since reth's own default is `full`.
- Local static ladder green (nightly fmt check, scoped clippy with `--no-deps`, scoped unit tests under the pinned queue discipline);  each new load-bearing assert confirmed by in-diff mutation (TN default reverted to `full`, tests fail, fix restored).  CI's workspace lanes and the pinned `+1.94` attest run on push.


sstanfield (dismissed):

grantkee (dismissed):
LGTM!

grantkee (approved):

PR #1240, opened 2026-08-21
TEL precompile follow-ups: event mirror, harness fidelity, rejection gas

### LINKED ISSUE #1239: Value-funded `burn` moves TEL into the precompile with no event
## Problem

TEL is the chain's native token, so holdings are native account balances rather than an ERC-20 ledger. The precompile at `0x7e1` bridges that gap by emitting ERC-20-shaped events for issuance and destruction: `Transfer(0x0, recipient, amount)` on `claim`, `Transfer(0x7e1, 0x0, amount)` on `burn`. The module README tells off-chain indexers to reconstruct TEL as an ERC-20 from exactly those events.

The payability gate added for #1190 makes `burn(uint256)` the single payable selector and the sanctioned way to fund the pool that `burn` destroys from. The wei moves from the caller to `0x7e1` through the EVM's own value transfer, which runs before the precompile does and emits nothing. The handler then emits only the outbound `Transfer(0x7e1, 0x0, amount)`.

An indexer rebuilding balances from these logs watches the pool pay out wei it was never seen receiving. Its mirrored balance for `0x7e1` goes negative, and the funding caller is never debited, so the reconstruction diverges from the chain on every value-bearing burn.

Two call shapes make it worse. `burn{value: V}(amount)` with `V > amount` parks `V - amount` at `0x7e1` with no event at all, invisible until some later burn destroys it. `burn{value: V}(0)` funds the pool while emitting `Burn(0)` and a zero-value `Transfer`, which reads as a no-op.

## Solution

`handle_burn` should emit `Transfer(caller, 0x7e1, msg.value)` ahead of the burn events whenever the call carries value. The payability gate already guarantees everything that log would claim: value reaches the handler for `burn` alone, and only when the precompile is the real transfer target, so the caller is always the funder and the wei is already in the pool. Delegatecall and callcode frames, where the apparent value never moved, are rejected before dispatch.

The alternative is to require `msg.value == 0 || msg.value == amount` in the gate, which closes the excess case by forbidding it. That removes multi-transaction top-up of the pool and forces the dispatcher to decode `burn`'s calldata a second time to compare against the amount. Mirroring the transfer keeps the documented design and still closes the hole.

The extra log costs roughly 1,756 gas on the value-bearing path. `burn` charges a flat 8,000 and its worst case already sits above that, so the constant should absorb the log rather than move: burning is governance-only, so the subsidy stays out of reach of untrusted callers. The gas table in the module README should record the new worst case.

grantkee:
- `burn` with value now emits `Transfer(caller, 0x7e1, msg.value)` before the burn events. The EVM's own transfer into the pool emits nothing, so an indexer reconstructing TEL as an ERC-20 from precompile logs saw the pool pay out wei it was never seen receiving, and attaching more value than the burned amount left the excess with no event at all. Two tests pin the log set for value-funded and zero-value burns.
- The flat 8,000 gas charge for `burn` is unchanged and absorbs the extra log. Worst case moves from 10,362 to 12,118 (0.77x to 0.66x headroom); burning is governance-only, so the subsidy is not reachable by untrusted callers. README gas table updated.
- The test harness now builds `0x7e1` with the `0xfe` code mainnet genesis gives it instead of an EIP-158-empty account at zero balance. No assertion could observe the difference, since `MainnetHandler::run` never finalizes the journal, but pool-draining tests such as `test_burn_value_funds_pool_and_burns` end with the pool at exactly zero, which is the state the code exempts from clearing in production.
- Documented that a precompile rejection halts rather than reverts, consuming all the gas the frame was given: a rejected sub-call loses the 63/64 it was forwarded, a rejected top-level transaction is charged its full limit, and the `false` it pushes arrives with 1/64 of the gas left to act on it. Unchanged from the static-call refusal, which has always worked this way, but the payability gate exposes it to a new class of caller.

closes #1239

### DIFF
--- a/crates/tn-reth/src/evm/tel_precompile/burnable.rs
+++ b/crates/tn-reth/src/evm/tel_precompile/burnable.rs
@@ -4,7 +4,12 @@
 //! 1. **`mint(uint256)`** — governance creates a pending mint with a 7-day timelock.
 //! 2. **`claim(address)`** — only governance can finalize the mint after the timelock expires,
 //!    crediting governance safe's native balance and incrementing `totalSupply`.
-//! 3. **`burn(uint256)`** — governance destroys tokens held by the precompile account.
+//! 3. **`burn(uint256)`** — governance destroys tokens held by the precompile account. `burn` is
+//!    the one payable selector: the EVM credits attached `msg.value` to the precompile before the
+//!    handler runs, topping up the pool the burn draws from, so `msg.value == amount` funds and
+//!    burns in one transaction and any excess stays in the pool for a later `burn`. That transfer
+//!    happens inside the EVM and emits nothing, so the handler mirrors it as an inbound
+//!    `Transfer(caller, precompile, msg.value)` log ahead of the burn events.
 //!
 //! The timelock provides a safety window for governance to cancel malicious mints before
 //! tokens enter circulation. A second `mint` call **overwrites** any pending mint, which
--- a/crates/tn-reth/src/evm/tel_precompile/burnable.rs
+++ b/crates/tn-reth/src/evm/tel_precompile/burnable.rs
@@ -55,7 +55,8 @@ sol! {
     event Claim(address indexed recipient, uint256 amount);
     /// Emitted when tokens are burned.
     event Burn(uint256 amount);
-    /// Emitted when tokens are minted (from address(0)) or burned (to address(0)).
+    /// Emitted when tokens are minted (from address(0)), burned (to address(0)), or sent to the
+    /// precompile as call value funding a `burn`.
     event Transfer(address indexed from, address indexed to, uint256 value);
 }
 
@@ -342,6 +343,14 @@ pub(super) fn handle_claim(
 ///
 /// Decrements the precompile's native balance by `amount`, then decrements `totalSupply`.
 ///
+/// `value` is the call value the EVM credited to the precompile before this handler ran. The
+/// dispatcher's payability gate admits it for this selector alone, and only when the precompile is
+/// the actual transfer target, so nonzero `value` is always `caller` topping up the pool this
+/// handler draws from. That transfer happens inside the EVM and emits nothing, so the handler
+/// mirrors it as `Transfer(caller, precompile, value)` ahead of the burn events: an indexer
+/// reconstructing TEL as an ERC-20 from these logs would otherwise watch tokens leave a pool they
+/// were never seen entering, and drift negative.
+///
 /// # Access control
 /// Governance-only via [`has_governance_role`].
 pub(super) fn handle_burn(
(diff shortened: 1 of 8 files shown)


crates/tn-reth/src/evm/tel_precompile/burnable.rs
```diff
@@ -389,6 +401,22 @@ pub(super) fn handle_burn(
         .sstore(TELCOIN_PRECOMPILE_ADDRESS, TOTAL_SUPPLY_SLOT, new_supply)
         .map_err(|e| PrecompileError::Other(format!("sstore failed: {e:?}").into()))?;
 
+    // Emit Transfer(caller, precompile, value) — mirrors the EVM's own silent value transfer
+    // that funded this burn, keeping the event log a complete account of the pool's balance.
+    if !value.is_zero() {
```
MavenRain, line 417: Question on consensus gating: this mirror log is emitted unconditionally, and `0x7e1` is live from mainnet genesis. Comparable changes to the state transition are gated on fork epochs (`CONSENSUS_REGISTRY_FORK_EPOCH`, `SEED_SIGNATURE_FORK_EPOCH` in `crates/types/src/forks.rs`), and an extra log changes the receipts root.

The payability gate (#1201, 5fb93f76) is not in any release tag yet, so no historical block can contain a value-funded burn and re-sync divergence is not possible today. The residual exposure is a mixed-version window: if a future value-funded burn executes while some validators run this build and others run a pre-#1240 build, the two groups compute different receipt roots for the same block.

Is the deployment expected to upgrade all validators before value-funded burns are used, or should this emission sit behind a fork epoch?

crates/tn-reth/src/evm/tel_precompile/burnable.rs
```diff
@@ -342,21 +343,32 @@ pub(super) fn handle_claim(
 ///
 /// Decrements the precompile's native balance by `amount`, then decrements `totalSupply`.
 ///
+/// `value` is the call value the EVM credited to the precompile before this handler ran. The
+/// dispatcher's payability gate admits it for this selector alone, and only when the precompile is
+/// the actual transfer target, so nonzero `value` is always `caller` topping up the pool this
+/// handler draws from. That transfer happens inside the EVM and emits nothing, so the handler
+/// mirrors it as `Transfer(caller, precompile, value)` ahead of the burn events: an indexer
+/// reconstructing TEL as an ERC-20 from these logs would otherwise watch tokens leave a pool they
+/// were never seen entering, and drift negative.
```
MavenRain, line 357: One inlet makes this invariant incomplete: `SELFDESTRUCT` with beneficiary `0x7e1` credits the pool's balance directly.  It is permissionless, it does not route through the precompile, and post-EIP-6780 it still transfers the full balance even when the contract is not destroyed.  Wei that enters that way is never mirrored, so an indexer can still watch the pool pay out wei it was not seen receiving.

For the other paths the claim holds: claim and mint credit the recipient, never the pool; a bare value send halts and unwinds; the basefee address is `0x9999...`, not `0x7e1`.  `SELFDESTRUCT` is the only silent inlet I could find.  Suggest qualifying this doc (and the README) to name it, rather than any code change.

crates/tn-reth/src/evm/tel_precompile/mod.rs
```diff
@@ -321,6 +349,59 @@ mod tests {
...
+        assert_eq!(
+            logs[0].topics(),
+            [
+                burnable::Transfer::SIGNATURE_HASH,
+                GOVERNANCE_SAFE_ADDRESS.into_word(),
+                TELCOIN_PRECOMPILE_ADDRESS.into_word(),
+            ]
+        );
+        // The mirror reports the attached value, not the burned amount.
+        assert_eq!(U256::from_be_slice(logs[0].data.data.as_ref()), value);
```
MavenRain, line 461: Two assertions would pin this test down further. It never checks `logs[0].address == TELCOIN_PRECOMPILE_ADDRESS`, and `U256::from_be_slice` accepts fewer than 32 bytes, so neither the emitter nor the 32-byte ABI encoding is asserted. A regression from `to_be_bytes_vec()` to `to_be_bytes_trimmed_vec()` would still pass. Suggest asserting the address and `logs[0].data.data.len() == 32` (and the same for `logs[2]`).

MavenRain (dismissed):
Nothing blocking . . . just a few points of interest

crates/tn-reth/src/evm/tel_precompile/burnable.rs
grantkee: Yes, and no burn has happened on testnet thus no fork necessary for s yncing testnet nodes

crates/tn-reth/src/evm/tel_precompile/burnable.rs
grantkee: True, but this opcode is deprecated and access is behind a multisig. Governance would have to go out of its way to create this giant mess. I'm not sure it's necessary or relevant

crates/tn-reth/src/evm/tel_precompile/mod.rs
grantkee: Good idea - added + docs

MavenRain (approved):

PR #1243, opened 2026-08-22
fix(tn-reth): detect canon-state broadcast lag in pool maintenance and resync drifted senders

### LINKED ISSUE #1236: tn-reth: worker pool maintenance silently loses canonical updates when the canon-state broadcast lags (no Lagged detection, no drift resync)
## Summary

The worker's transaction-pool maintenance task consumes canonical-state updates through `canonical_state_stream()`. That stream is backed by a bounded tokio broadcast channel (256 entries in the pinned reth v1.11.3), and the stream wrapper silently skips every notification lost to lag. The pool task has no way to observe the gap and no resync path, so a lag event permanently corrupts the worker's pool state: mined transactions from the skipped rounds stay pending, and changed account nonces/balances are never refreshed.

Credit: surfaced by Haxatron's Cantina collaborative-review comment on `notify_canon_state()`.

## Details

Producer: `RethEnv::announce_executed_output` broadcasts one `CanonStateNotification::Commit` per consensus round:

- `crates/tn-reth/src/env/execution.rs:338` - `self.canonical_in_memory_state().notify_canon_state(notification);`

Channel and stream semantics (reth v1.11.3, tag `v1.11.3`):

- `crates/chain-state/src/in_memory.rs:28` - `CANON_STATE_NOTIFICATION_CHANNEL_SIZE: usize = 256`
- `crates/chain-state/src/notifications.rs:70` - `CanonStateNotificationStream` maps `BroadcastStreamRecvError::Lagged` to a debug-level log and continues with the oldest retained notification. The consumer never sees the error.

TN subscriber with no lag story: the worker pool maintenance task.

- `crates/tn-reth/src/txn_pool.rs:149-173` - critical task; applies each `Commit` via `process_canon_state_update`
- `crates/tn-reth/src/txn_pool.rs:215-249` - collects the notification's `changed_accounts` and `mined_transactions`, then calls `on_canonical_state_change` synchronously on the loop thread

Because the handler runs inline in the stream loop, sustained per-round processing cost above the round production rate lets the receiver fall behind. Once it is more than 256 notifications behind, the skipped `Commit`s are gone: the next received notification carries only its own blocks and state. The pool tip and base fee self-heal (each notification carries the latest tip), but mined-transaction removal and account refresh for the skipped rounds never happen, and nothing later reconciles them.

The other two in-process subscribers already tolerate or surface lag:

- Batch builder (`crates/batch-builder/src/lib.rs:290-294`) only snapshots the tip of whatever notification arrives, so it self-heals.
- ExEx manager (`crates/exex/src/manager.rs`) consumes the raw receiver, detects gaps, and surfaces an explicit `TnExExNotification::Lagged` marker so stateful ExExes reconcile.

Upstream contrast: reth pairs the same lossy stream with the drift machinery in `maintain_transaction_pool` (`MaintainedPoolState::Drifted` plus bounded `reload_accounts`, reth `crates/transaction-pool/src/maintain.rs`). TN's replacement maintenance task dropped that recovery path.

## Impact

- Already-mined transactions stay in the pool as pending and get re-proposed in later batches. Batch validation (`crates/batch-validator/src/validator.rs:63-82`) checks size, decode, tx-type allowlist, gas, and base fee, not nonces, so stale batches pass quorum and waste batch space and execution throughput as nonce-too-low failures.
- Stale account nonce/balance snapshots cause wrong promote/demote decisions and pool bloat, which adds eviction pressure on honest transactions on the affected worker.
- The failure is silent: the only signal is a debug-level log inside reth.

Likelihood is low in normal operation (the handler is normally fast), but the window is attacker-influenced: spam that inflates per-round changed-account sets raises the handler's cost per notification while rounds keep arriving. This is a defense-in-depth gap with a permanent, silent failure mode.

## Suggested fix

1. Subscribe with `subscribe_to_canonical_state()` (the raw `broadcast::Receiver`) instead of `canonical_state_stream()` so `RecvError::Lagged(n)` is observable.
2. On `Lagged`, treat the pool as drifted and resync: reload on-chain account state for the pool's senders (bounded, like reth's `max_reload_accounts`) and drop transactions whose nonce is below the canonical account nonce, mirroring reth's `maintain.rs` drift handling.
3. Log at `warn!` and add a metrics counter so operators can see lag events.

MavenRain:
Closes #1236.  (Cantina #3)

## Problem

The worker pool maintenance task consumes canonical-state updates through `canonical_state_stream()`. That stream is backed by a bounded tokio broadcast channel (256 entries in the pinned reth v1.11.3), and the stream wrapper maps every `Lagged` error to a debug-level log and skips ahead, so the task can never observe a gap. Once the task falls more than 256 notifications behind, the skipped `Commit`s are gone: mined transactions from the lost rounds stay pending, and changed account nonces and balances are never refreshed. The pool tip and base fee self-heal on the next notification; the per-round data does not, and nothing later reconciles it.

The damage compounds silently. Stale pending transactions get re-proposed in later batches, and batch validation (`crates/batch-validator/src/validator.rs`) checks size, decode, tx-type allowlist, gas, and base fee, not nonces, so the stale batches pass quorum and waste batch space and execution throughput as nonce-too-low failures. Stale account snapshots cause wrong promote/demote decisions and pool bloat on the affected worker.

No attacker is required: any sustained per-round processing cost above the round production rate opens the window. The window is attacker-influenced, though . . . spam that inflates per-round changed-account sets raises the handler's cost per notification while rounds keep arriving.

Upstream reth pairs the same lossy stream with drift recovery in `maintain_transaction_pool` (`MaintainedPoolState::Drifted` plus bounded `reload_accounts`); TN's replacement maintenance task dropped that path. The other two TN subscribers already tolerate or surface lag (batch builder snapshots only the tip; the ExEx manager surfaces an explicit `Lagged` marker).

## Fix

- [x] `crates/tn-reth/src/txn_pool.rs`: the maintenance task subscribes via `subscribe_to_canonical_state()` (the raw broadcast receiver) wrapped in a `BroadcastStream`, the same shape the ExEx manager uses, so `Lagged` surfaces as an error item instead of being swallowed.
- [x] On `Lagged`: `warn!`, increment the new `tn_reth.canon_state_lagged_total` counter, and mark every pool sender dirty (`unique_senders()`, reth's `Drifted` analogue).
- [x] Dirty senders reload from canonical state in bounded chunks (`MAX_RELOAD_ACCOUNTS = 100` per loop iteration, mirroring reth's `max_reload_accounts`) and apply via `update_accounts`, which refreshes each sender's nonce/balance snapshot and discards transactions whose nonce is below the canonical account nonce, i.e. the transactions mined in the lost rounds. The bound is enforced on every iteration, so recovery cannot stall the already-behind loop; addresses whose state read fails stay dirty and retry. The reload reads latest state while older retained notifications may still be queued; draining them can transiently re-apply an older snapshot, but the drain's own tail restores exact state (the backlog's final `Commit` is authoritative for every sender it touches, and senders touched only in lost rounds are never overwritten by the backlog), and discarded transactions cannot be resurrected.
- [x] `crates/tn-reth/src/metrics.rs`: alertable `canon_state_lagged_total` counter on `RethEnvMetrics` so operators see lag events (previously the only signal was a debug log inside reth).
- [x] `WorkerTxPool::new` split into `build` (pool construction) plus `spawn_maintenance_task`; `RethEnv::init_txn_pool_without_maintenance` (`cfg(test)`) is the seam that lets a test reproduce a pool that missed a commit.
- [x] `Commit` / `Reorg` handling is unchanged in behavior: `Reorg` is still skipped with a warning (TN never reorgs), and the task still exits with a `TaskError` when the stream closes.

## Testing

- `test_lag_resync_discards_transactions_mined_in_lost_rounds`: builds the pool without its maintenance task, admits a transaction, then commits a canonical block that mines it. That is byte-for-byte the state a lagged worker is left in. The test first proves the drift is real (canonical nonce advanced to 1 while the mined transaction is still pending in the pool), then runs the lag path (`mark_drifted` + `reload_dirty_accounts`) and asserts the mined transaction is discarded and the dirty set drains.
- `test_reload_dirty_accounts_is_bounded`: three dirty senders with `max_reload = 1` drain exactly one per call.
- Local: nightly `fmt --check` and a `-p tn-reth --all-targets` check green; the `txn_pool` test battery green under the pinned toolchain. CI runs the full workspace ladder on push; the pinned `+1.94` attest runs post-push.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sstanfield (dismissed):

crates/tn-reth/src/txn_pool.rs
```diff
@@ -176,32 +208,139 @@ impl WorkerTxPool {
...
+    /// [`MAX_RELOAD_ACCOUNTS`]); addresses whose state read fails stay dirty and are
+    /// retried on a later iteration.
+    ///
+    /// The reload reads the LATEST canonical state while older retained notifications may
+    /// still be queued behind the `Lagged` marker. Draining those can transiently re-apply
+    /// an older snapshot for a sender, but the drain's own tail restores exact state: the
+    /// backlog's final `Commit` is authoritative for every sender it touches, and senders
+    /// touched only in the lost rounds are never overwritten by the backlog at all.
+    /// Discarded transactions cannot be resurrected by the transient regression.
+    fn reload_dirty_accounts(&self, dirty: AddressSet, max_reload: usize) -> AddressSet {
```
grantkee, line 329: ### 1. Lag recovery performs blocking storage reads inline on the async runtime

- **Verdict**: CONFIRMED (all mechanical claims) — **Confidence**: HIGH
- **Severity**: Medium (initial) → **Low** (verified). Efficiency and latency hygiene, not DoS.
- **Category**: Concurrency
- **Location**: `crates/tn-reth/src/txn_pool.rs:305-316` (`reload_dirty_accounts`),
  `crates/tn-reth/src/txn_pool.rs:323-348` (`load_changed_account`), driven from
  `crates/tn-reth/src/txn_pool.rs:234-253`

`reload_dirty_accounts` is a plain `fn`, and its read loop at `txn_pool.rs:307-309` is a fully synchronous `.map().collect()` with no `await` or yield point; grep for `spawn_blocking|block_in_place|yield_now` in `txn_pool.rs` returns no matches. `spawn_critical_task` is `tokio::spawn` (`crates/types/src/task_manager.rs:189`), not the blocking pool — contrast `spawn_blocking_task` at `task_manager.rs:210-219`. The node runs one shared multi-thread runtime for everything (`crates/telcoin-network-cli/src/cli.rs:161-165`).

`basic_account` does terminate in synchronous MDBX I/O. Traced: `blockchain_provider.rs:773-775` → `ConsistentProvider::new` → `consistent.rs:71-78` calls `database_provider_ro()` → `database/mod.rs:238-250` → `self.db.tx()?` → `mdbx_txn_begin_ex`. Then `latest_ref()` → `history_by_block_hash_ref` → `try_into_history_at_block` (`database/provider.rs:850-890`), which itself does `best_block_number()` plus two `get_prune_checkpoint()` reads plus a hash→number lookup. Roughly 4-5 MDBX lookups plus a transaction begin/abort per address.

reth does offload the equivalent work: `maintain.rs:232` is `task_spawner.spawn_blocking_task(fut)`, and `maintain.rs:58,76` confirm reth's `max_reload_accounts: 100` — TN's `MAX_RELOAD_ACCOUNTS` matches exactly.

**Why the severity dropped from Medium to Low.** The verifier found several facts the original
analysis did not weigh:

- **Recovery-only, and zero-cost on the happy path.** On a normal `Commit`, `apply_canon_notification` returns `AddressSet::default()` and `dirty_addresses` starts empty, so `chunk` is empty and **no reads happen at all** (`txn_pool.rs:236-248`). The only producer of dirty addresses is `mark_drifted`, reachable solely via `Lagged`, which requires falling more than 256 notifications behind.
- **Occupies one worker, does not stall the runtime.** The runtime is multi-thread with `num_cpus` workers and work stealing; there is no single-worker configuration in the node path.
- **Not usefully attacker-steerable.** An attacker must both induce the lag and have funded distinct senders filling the pool, and the reload rate is fixed at 100/iteration regardless.
- **reth is not categorically strict about this either.** reth calls the same `load_accounts` inline and synchronously in its async loop body on the reorg path (`maintain.rs:360`), with an unbounded `missing_changed_acc`.
- **This is the house pattern, and larger unbounded instances already run on the hot path.** `crates/batch-builder/src/batch.rs:172-182` calls `pool.get_account_balance(address)` once per unique sender per batch — the same `BlockchainProvider::basic_account` MDBX read — inside a plain async task (`crates/batch-builder/src/lib.rs:206-212`), unbounded, every round. The repo already carries an acknowledging TODO at `txn_pool.rs:365-368`.

**Proposed fix.** Two independent changes; do (a) first, it delivers the larger real-world win and is shared with finding 2. See finding 2 for the code. Then optionally (b): offload the chunk at the call site so the tests' direct calls stay synchronous.

```rust
let to_reload: AddressSet = dirty_addresses.into_iter().chain(newly_dirty).collect();
if to_reload.is_empty() {
    dirty_addresses = to_reload;
    continue;
}

// The reload is synchronous MDBX I/O. Run it on the blocking pool so it never occupies one
// of the async worker threads this runtime also uses for consensus and networking (reth
// offloads the same work: `maintain_transaction_pool`'s `spawn_blocking_task`).
let pool = txn_pool_clone.clone();
let retained = to_reload.clone();
dirty_addresses = tokio::task::spawn_blocking(move || {
    pool.reload_dirty_accounts(to_reload, MAX_RELOAD_ACCOUNTS)
})
.await
.unwrap_or_else(|error| {
    // the blocking task was dropped or panicked: keep every sender dirty so the next
    // notification retries the reload
    warn!(target: "txpool", ?error, "dirty-account reload task failed; retrying next round");
    retained
});
```

crates/tn-reth/src/txn_pool.rs
```diff
@@ -176,32 +208,139 @@ impl WorkerTxPool {
...
+        }
+        pending.chain(loaded.iter().filter_map(|result| result.as_ref().err().copied())).collect()
+    }
+
+    /// Read `address`'s canonical account and shape it as a [`ChangedAccount`] for the pool.
+    ///
+    /// A missing account maps to [`ChangedAccount::empty`] (nonce 0, zero balance), matching
+    /// reth's `load_accounts`; a provider read error returns the address so the caller keeps
+    /// it dirty and retries.
+    fn load_changed_account(&self, address: Address) -> Result<ChangedAccount, Address> {
```
grantkee: ### 2. A fresh `ConsistentProvider` + MDBX read transaction is built per address

- **Verdict**: PARTIALLY_VALID — **Confidence**: HIGH
- **Severity**: Medium (initial) → **Low** (verified)
- **Category**: Optimization
- **Location**: `crates/tn-reth/src/txn_pool.rs:323-348`, called per-address from
  `crates/tn-reth/src/txn_pool.rs:307-309`

**The mechanism is confirmed exactly, step for step.** `self.1` is
`BlockchainProvider<TelcoinNode>` (`txn_pool.rs:117-120`), not a `StateProvider`. Each
`basic_account` call constructs a fresh `ConsistentProvider` (`blockchain_provider.rs:124-127`),
which opens a new MDBX read transaction _and_ snapshots the in-memory head
(`consistent.rs:66-80`), then builds a `MemoryOverlayStateProviderRef` via `latest_ref()`
(`consistent.rs:107-118`, `409-417`). So a 100-address chunk is **100 `ConsistentProvider`s, 100
MDBX read transactions, 100 memory overlays, and roughly 300 extra table lookups**. reth's
`load_accounts` acquires one provider at `maintain.rs:682` and loops addresses against it at
`maintain.rs:686-687`. The claim's arithmetic is right.

**Falsified sub-claim: the consistency argument.** The original finding asserted that reading
each address at a potentially different canonical tip makes the chunk "not a single consistent
snapshot," implying a correctness problem. The verifier established this is mechanically true
but **not a defect**:

- The pool has no cross-address invariant. `update_accounts` refreshes each sender's
  nonce/balance independently; a slightly newer read for sender B is a _valid_ state for B.
- A newer read only discards more already-mined transactions (correct); an older read only leaves
  transactions parked, which the next `Commit` corrects.
- The file already documents and accepts a strictly _larger_ version of this skew at
  `txn_pool.rs:299-304` — the reload reads LATEST while older notifications are still queued.
  Per-address tip skew is a subset of that already-analyzed window.

**Cost is also more modest than implied.** `MemoryOverlayStateProvider` holds `Arc`s
(`chain-state/src/in_memory.rs:753-763`); `block()` is an `Arc` clone
(`in_memory.rs:628-630`), no block-data copy. An MDBX RO transaction begin is single-digit
microseconds. Realistic overhead for a 100-address chunk is ~0.5-2 ms, on a path that only runs
after a lag event. Tidiness, not a latency hazard.

**Proposed fix** (this is the convergent one — three verifiers proposed it independently, and it
also addresses most of finding 1's cost). Add `StateProviderBox, StateProviderFactory as _` to
the `reth_provider` import list at `txn_pool.rs:38-41`; both are already used this way in
`crates/tn-reth/src/env/helpers.rs:39-45,293-295`.

```rust
fn reload_dirty_accounts(&self, dirty: AddressSet, max_reload: usize) -> AddressSet {
    let mut pending = dirty.into_iter();
    let chunk: Vec<Address> = pending.by_ref().take(max_reload).collect();

    // Acquire ONE state provider for the whole chunk, like reth's `load_accounts`.
    // `BlockchainProvider::basic_account` builds a fresh `ConsistentProvider` on every call
    // -- an MDBX read transaction plus a `MemoryOverlayStateProvider` over the in-memory
    // canonical blocks -- so a per-address read repeats that setup `max_reload` times.
    let state = match self.1.latest() {
        Ok(state) => state,
        Err(error) => {
            debug!(target: "txpool", ?error, "failed to acquire canonical state for pool resync");
            // nothing was reloaded: every address stays dirty and is retried next iteration
            return pending.chain(chunk).collect();
        }
    };

    let loaded: Vec<Result<ChangedAccount, Address>> =
        chunk.into_iter().map(|address| Self::load_changed_account(&state, address)).collect();
    let accounts: Vec<ChangedAccount> =
        loaded.iter().filter_map(|result| result.as_ref().ok().copied()).collect();
    if !accounts.is_empty() {
        self.0.update_accounts(accounts);
    }
    pending.chain(loaded.iter().filter_map(|result| result.as_ref().err().copied())).collect()
}

fn load_changed_account(
    state: &StateProviderBox,
    address: Address,
) -> Result<ChangedAccount, Address> {
    state
        .basic_account(&address)
        .map(|maybe_account| {
            maybe_account
                .map(|account| ChangedAccount {
                    address,
                    nonce: account.nonce,
                    balance: account.balance,
                })
                .unwrap_or_else(|| ChangedAccount::empty(address))
        })
        .map_err(|error| {
            debug!(target: "txpool", ?address, ?error, "failed to reload account state for pool resync");
            address
        })
}
```

`StateProviderBox = Box<dyn StateProvider + Send>` and `StateProvider: AccountReader`
(`storage-api/src/state.rs:30,34-42`), so `state.basic_account(&address)` resolves with the
`AccountReader as _` import already present at `txn_pool.rs:39`.

**Collateral effect**: the test at `txn_pool.rs:826` calls `pool.load_changed_account(...)` and
must be updated with the signature change:

```rust
let state = pool.1.latest().expect("state provider for canonical tip");
let account = WorkerTxPool::load_changed_account(&state, tx_factory.address()).unwrap();
```

crates/tn-reth/src/txn_pool.rs
```diff
@@ -176,32 +208,139 @@ impl WorkerTxPool {
...
-                    // Reorg notification here is a bug upstream. Skip it rather than panic —
-                    // this runs inside a critical task, and aborting it would take down the
-                    // whole node over a pool-maintenance miss.
-                    _ => warn!(
-                        target: "txpool",
-                        "unexpected canonical state notification (TN never reorgs); skipping \
-                         transaction pool update"
-                    ),
-                }
+                let newly_dirty = update
```
grantkee, line 238: TLDR: consider re-arming reload dirty accounts for residual drains in the event tx volume spikes then is very low

### 3. Drift recovery advances only one chunk per incoming notification

- **Verdict**: PARTIALLY_VALID — **Confidence**: HIGH
- **Severity**: Low (initial) → **Low** (verified), but the practical impact is far smaller
  than the finding argued
- **Category**: Bugs / Liveness
- **Location**: `crates/tn-reth/src/txn_pool.rs:236-249`

**Confirmed.** `reload_dirty_accounts` has exactly one production call site, `txn_pool.rs:245`,
inside the `while let Some(update) = state_stream.next().await` body opening at line 236. The
other hits (833, 855-859) are `#[cfg(test)]`. `dirty_addresses` is a task-local `let mut` at
line 235 reassigned only at 245; `reload_dirty_accounts` takes one chunk (line 307) and returns
the remainder (line 315) without looping. There is no interval, tick, or backup branch. The reth
comparison is also accurate: `maintain.rs:205-233` re-arms the reload at the top of every loop
iteration, `maintain.rs:258-261` selects on it, `maintain.rs:321` loops straight back on a
reload-only wakeup, and `maintain.rs:275` adds a `stale_eviction_interval` tick as an independent
liveness path. TN has neither.

**Confirmed: the dirty set can greatly exceed 100.** `mark_drifted` returns
`unique_senders()` (`txn_pool.rs:286`), which spans every subpool (reth
`pool/txpool.rs:189-191`). TN does not override subpool limits — `build` uses
`node_config.txpool.pool_config()` (`txn_pool.rs:155`) and `init_reth_defaults`
(`cli.rs:131-146`) seeds only `max_account_slots` (256) and `ipcpath` — so reth's
`TXPOOL_SUBPOOL_MAX_TXS_DEFAULT = 10_000` (`transaction-pool/src/config.rs:15`) stands across
three reachable subpools: **up to 30,000 senders** (the blob pool is unreachable via
`.no_eip4844()` at `txn_pool.rs:172`).

**The mitigation the finding missed — and it is decisive.** Tokio's broadcast semantics mean
`Lagged(n)` is produced _only_ when the ring has been overwritten, i.e. the buffer is full and
the receiver's cursor resets to the oldest of **256 immediately-available messages**
(`chain-state/src/in_memory.rs:28`). `BroadcastStream` yields `Err(Lagged)` and then those 256
`Ok(Commit)`s back-to-back **with no await on new consensus**. So the real timeline is:

- iteration 1: `mark_drifted` → N dirty, reload 100
- iterations 2..257: 256 buffered `Commit`s, each reloading another 100 → **25,600 more, with
  zero waiting**
- only the residual (N − 25,700) waits on fresh notifications

Even at the theoretical 30,000-sender maximum the residual is ~4,300, i.e. ~43 further rounds. At
the configured cadence (`max_header_delay` 2500 ms / `min_header_delay` 1000 ms,
`crates/config/src/node.rs:327-333`, one notification per executed consensus output,
`crates/tn-reth/src/env/execution.rs:338`) that is **~45-110 seconds, not indefinite**.

**Harm is bounded and self-healing.** A drifted sender means already-mined transactions stay
pending and reach the batch builder. Those are executed and **skipped, not fatal** —
`crates/tn-reth/src/env/execution.rs:163-179` matches `BlockValidationError::InvalidTx` and
`continue`s, with `invalid_txs_skipped_total` documented as expected in normal operation. The
batch builder's own post-quorum update (`crates/batch-builder/src/lib.rs:473-479` →
`update_canonical_state`) removes those transactions once the batch is admitted, so drifted
senders also self-clear through the normal build path. Net: wasted batch space and a misleading
RPC pool view for at most tens of seconds. No fork risk, no incorrect state.

**Proposed fix** (optional — reasonable to defer). Add a self-re-arming branch so the residual
drains at reload speed rather than round speed, keeping the chunk cap:

```rust
/// Interval at which residual dirty senders are re-chunked between canonical-state
/// notifications.
///
/// Without it the drain advances only when the stream yields, so a dirty set larger than the
/// post-lag backlog (256 buffered notifications x [`MAX_RELOAD_ACCOUNTS`]) waits on consensus
/// rounds to finish resyncing. reth re-arms its `reload_accounts_fut` from the same `select!`
/// for the same reason.
const RELOAD_RETRY_INTERVAL: Duration = Duration::from_millis(100);
```

```rust
let mut reload_interval = tokio::time::interval(RELOAD_RETRY_INTERVAL);
reload_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
loop {
    let newly_dirty = tokio::select! {
        // canonical updates first: they are what keeps the pool correct, and a post-lag
        // backlog should be applied before spending reads on the residual
        biased;

        update = state_stream.next() => {
            let Some(update) = update else { break };
            update
                .map(|notification| {
                    txn_pool_clone.apply_canon_notification(notification);
                    AddressSet::default()
                })
                .unwrap_or_else(|BroadcastStreamRecvError::Lagged(missed)| {
                    txn_pool_clone.mark_drifted(missed)
                })
        }

        // drain the residual without waiting on the next consensus round; the guard parks
        // this branch whenever the pool is in sync
        _ = reload_interval.tick(), if !dirty_addresses.is_empty() => AddressSet::default(),
    };
    dirty_addresses = txn_pool_clone.reload_dirty_accounts(
        dirty_addresses.into_iter().chain(newly_dirty).collect(),
        MAX_RELOAD_ACCOUNTS,
    );
}
```

Requires adding `Duration` to `use std::{sync::Arc, time::Instant};` at `txn_pool.rs:52`.

crates/tn-reth/src/txn_pool.rs
```diff
@@ -176,32 +208,139 @@ impl WorkerTxPool {
...
+            .map(|maybe_account| {
+                maybe_account
+                    .map(|account| ChangedAccount {
+                        address,
+                        nonce: account.nonce,
+                        balance: account.balance,
+                    })
+                    .unwrap_or_else(|| ChangedAccount::empty(address))
+            })
+            .map_err(|error| {
```
grantkee, line 404: ### 4. Provider read failures during resync are logged at `debug!`

- **Verdict**: PARTIALLY_VALID — **Confidence**: HIGH
- **Severity**: Low (initial) → **Low** (verified). Diagnostics; no correctness impact.
- **Location**: `crates/tn-reth/src/txn_pool.rs:335-343`

**Confirmed facts.** `map_err` at `txn_pool.rs:335-343` logs `debug!` and returns the address;
the neighbouring lag log at `txn_pool.rs:279-284` is `warn!`, so the level inconsistency within
one recovery path is real. There is **no failed-reload metric** — grep for `canon_state_lagged`
returns only `txn_pool.rs:285` and `metrics.rs:123`. The retry is genuinely unbounded and no
address is dropped: `txn_pool.rs:315` returns `pending.chain(failures).collect()` and
`txn_pool.rs:245-248` unions it back into the `AddressSet`. One caveat the original finding did
not state: retries are notification-driven, not timer-driven, so an idle stream stalls the
resync (this is finding 3).

**Falsified sub-claim: "no operator-visible signal at default log levels."** The verifier traced
the actual tracing setup rather than assuming. `crates/telcoin-network-cli/src/cli.rs:72`
flattens reth's `LogArgs`; `cli.rs:188` calls `init_tracing_with_layers`. In reth v1.11.3
`crates/node/core/src/args/log.rs`, stdout verbosity defaults to INFO (lines 217, 232-238), but
`log_file_filter` defaults to `"debug"` (line 34) with `log_file_max_files` = 5 (line 51) — **the
file layer is on by default at DEBUG**. `layers.rs:122-133` builds it as
`build_env_filter(None, "debug")`, appending the bare `debug` directive last; tracing-subscriber
0.3.23 `Ord for Directive` ignores level (`filter/env/directive.rs:300-328`), so `DirectiveSet::add`
replaces a bare `info` from `RUST_LOG` with the bare `debug`. Even under the `RUST_LOG=info` that
`etc/compose.yaml:106,129,152,175` sets, this line lands in `<datadir>/reth/logs/reth.log`.

So: **no metric signal — true; no log signal at default levels — false.** It is captured, but
buried in a full debug firehose across a 5×200 MB rotation. And the one-shot `warn!` at line 279
does reach stdout when the lag begins; it simply never repeats to say the resync is stuck.

**Upstream comparison weakens it further.** reth's own `load_accounts`
(`maintain.rs:684-694`) pushes a failed address to `failed_to_load` with **no log at all**, and
logs whole-state failure at `debug!` (`maintain.rs:310`). TN's per-address `debug!` on the same
target is strictly more observable than the code it mirrors.

**House idiom strengthens the metric half.** This repo has a stated, twice-implemented convention
that an absorbed provider fault gets a counter: `crates/node/src/metrics.rs:38-48` ("Once such a
fault is survivable it becomes invisible until it is not") and
`crates/engine/src/metrics.rs:35-37` (`persist_provider_fault_retries_total`). The resync failure
path is the odd one out. Note also that `tn_reth_canon_state_lagged_total` is **not** in
`etc/grafana/telcoin-node-metrics.json` (only `tn_reth_unrecoverable_txs_dropped_total` is) and is
not in `crates/tn-reth/README.md`, so even the counter that does exist has no dashboard or alert
wired up.

**Proposed fix.** Add a counter and one _aggregated_ `warn!` per iteration — per-address `warn!`
would flood at up to 100 addresses per round. Keep the existing per-address `debug!`: it carries
the address and the `ProviderError`, which the aggregate cannot. In `metrics.rs`:

```rust
/// Canonical account reads that failed while resyncing pool senders after a lag event.
///
/// Alert on a sustained rate, not on any nonzero value. A failed read leaves its sender in the
/// dirty set to be retried on the next notification, so isolated increments are the intended
/// absorption of a transient provider fault. A rate that never returns to zero means the
/// resync [`RethEnvMetrics::canon_state_lagged_total`] started is not finishing, and the pool
/// is serving stale nonce/balance snapshots for every sender still queued.
pub(crate) canon_state_resync_failures_total: Counter,
```

and in `txn_pool.rs`, replacing the tail of `reload_dirty_accounts` (line 315):

```rust
let failed: Vec<Address> =
    loaded.iter().filter_map(|result| result.as_ref().err().copied()).collect();
if !failed.is_empty() {
    // Bounded by `max_reload`, so one line per iteration rather than one per address.
    RETH_METRICS.canon_state_resync_failures_total.increment(failed.len() as u64);
    warn!(
        target: "txpool",
        failed = failed.len(),
        reloaded = accounts.len(),
        "canonical account reads failed during pool resync; those senders stay dirty and are \
         retried on the next canonical notification"
    );
}
pending.chain(failed).collect()
```

crates/tn-reth/src/txn_pool.rs
```diff
@@ -176,32 +208,139 @@ impl WorkerTxPool {
...
+        }
+    }
+
+    /// Record a canonical-state broadcast lag event and return the sender set to resync.
+    ///
+    /// The skipped `Commit`s are gone from the broadcast channel, so their mined
+    /// transactions and changed accounts can never be replayed: treat every sender with
+    /// transactions in the pool as dirty, exactly like reth's
+    /// `MaintainedPoolState::Drifted`.
+    fn mark_drifted(&self, missed: u64) -> AddressSet {
```
grantkee, line 301: TLDR: metric for missed canon state notifications

### 5. Lag metric discards the `missed` notification count

- **Verdict**: PARTIALLY_VALID — the `increment(1)` choice is a correct DESIGN_DECISION; the
  missing magnitude series is a real but additive gap. **Confidence**: HIGH
- **Severity**: Low (initial) → **Informational** (verified)
- **Category**: Observability
- **Location**: `crates/tn-reth/src/txn_pool.rs:278-287`, `crates/tn-reth/src/metrics.rs:115-123`

**Confirmed facts.** `missed` appears only as a `warn!` field (`txn_pool.rs:281`); line 285 is
`increment(1)`, unconditional. The doc does say "Alert on any nonzero value" and "sustained lag
warrants investigation" (`metrics.rs:115-123`). The emitted series carries no magnitude
information, and magnitude is not recoverable from any other metric — there is no counter of
notifications received or applied by the pool task to difference against.

**Falsified framing: `increment(missed)` would be wrong.** The field is named
`canon_state_lagged_total` and its doc declares "Canonical-state broadcast lag **events**" with
"Each increment means the pool task fell more than the broadcast channel's capacity behind"
(`metrics.rs:115,118-119`). Incrementing by `missed` would silently redefine the series while
keeping a name that says otherwise, and would break "alert on any nonzero value" thresholding by
magnitude. The house idiom already distinguishes the two shapes: event counters use
`increment(1)` (`metrics.rs:255,312`), magnitude counters take a count
(`metrics.rs:264,274,282,290,299`; `crates/state-sync/src/lib.rs:276-278` increments
`headers_fetched_total` by a height gap), and the repo already ships paired
denominator/numerator series (`forwarded_txns_queued_total` vs `forwarded_txns_dropped_total`,
`metrics.rs:285-300`). Magnitude belongs in a **separate** series.

**Mitigation the finding understated**: magnitude _is_ recoverable from the log. The `warn!` at
`txn_pool.rs:279-284` carries `missed`, and WARN passes the default stdout INFO filter, so an
operator investigating has the number in the first place they will look.

Severity drops to Informational because the dashboard consequence is currently moot:
`tn_reth_canon_state_lagged_total` does not appear in `etc/grafana/telcoin-node-metrics.json` or
`crates/tn-reth/README.md`, so no query today would be misled.

**Proposed fix** — additive only; leave line 285 as `increment(1)`. In `metrics.rs`:

```rust
/// Canonical-state notifications the worker pool maintenance task lost to broadcast lag.
///
/// The magnitude behind [`RethEnvMetrics::canon_state_lagged_total`], which counts lag EVENTS:
/// one increment there reads identically whether one notification was dropped or ten thousand
/// were, and the count otherwise exists only in the `missed` field of the accompanying `warn!`.
/// Read the two together - `rate(missed) / rate(lagged)` is the average gap per event, and it
/// is the gap, not the event count, that says how much sender state the pool had to rebuild.
pub(crate) canon_state_notifications_missed_total: Counter,
```

and in `txn_pool.rs`, immediately after line 285:

```rust
RETH_METRICS.canon_state_notifications_missed_total.increment(missed);
```

### 6. `metrics.rs` module and `init()` docs are stale after adding a third counter

- **Verdict**: CONFIRMED — **Confidence**: HIGH (verified by direct inspection of the diff, and
  independently re-reported by the finding 4/5 verifier as an incidental observation)
- **Severity**: Informational
- **Category**: Architecture
- **Location**: `crates/tn-reth/src/metrics.rs:3`, `crates/tn-reth/src/metrics.rs:73`,
  `crates/tn-reth/src/metrics.rs:76`, `crates/tn-reth/src/metrics.rs:84`

The diff for `0cb74e21` touches only `metrics.rs:87-121`: it updated the **struct-level** doc
(line 90-95, now "plus the worker pool's canonical-state subscription health") and added the
counter. Three neighbouring statements were not updated and now contradict the code:

- `metrics.rs:3` — "The counter series are exported under the `tn_reth` scope, in two unrelated
  groups", then enumerates only block-building and observer-forwarding. There are now three
  groups.
- `metrics.rs:73` — "Register **both** counters with the global recorder" — there are three.
- `metrics.rs:76-77` — "the only sites that force it are drop paths a healthy node never takes."
  `mark_drifted` (`txn_pool.rs:285`) is a new non-drop force site. This sentence is
  load-bearing: it is the justification for why `init()` exists at all.
- `metrics.rs:84` — "both counters baseline at zero from process start".

This module's docs encode the recorder-ordering invariant that makes these counters work, so
drift here costs more than ordinary comment rot.

**Proposed fix**: update line 3 to "in three unrelated groups" and add a sentence naming the pool
canonical-state group; change "both counters" → "all three counters" at lines 73 and 84; and
amend line 76 to note that `mark_drifted` is a non-drop force site — which, if anything,
_strengthens_ the argument for `init()`, since a lag event is exactly the moment an operator
needs the series to already exist with a zero baseline for `rate()` to render a step.

crates/tn-reth/src/metrics.rs
grantkee, line 1: ### 6. `metrics.rs` module and `init()` docs are stale after adding a third counter

- **Verdict**: CONFIRMED — **Confidence**: HIGH (verified by direct inspection of the diff, and
  independently re-reported by the finding 4/5 verifier as an incidental observation)
- **Severity**: Informational
- **Category**: Architecture
- **Location**: `crates/tn-reth/src/metrics.rs:3`, `crates/tn-reth/src/metrics.rs:73`,
  `crates/tn-reth/src/metrics.rs:76`, `crates/tn-reth/src/metrics.rs:84`

The diff for `0cb74e21` touches only `metrics.rs:87-121`: it updated the **struct-level** doc
(line 90-95, now "plus the worker pool's canonical-state subscription health") and added the
counter. Three neighbouring statements were not updated and now contradict the code:

- `metrics.rs:3` — "The counter series are exported under the `tn_reth` scope, in two unrelated
  groups", then enumerates only block-building and observer-forwarding. There are now three
  groups.
- `metrics.rs:73` — "Register **both** counters with the global recorder" — there are three.
- `metrics.rs:76-77` — "the only sites that force it are drop paths a healthy node never takes."
  `mark_drifted` (`txn_pool.rs:285`) is a new non-drop force site. This sentence is
  load-bearing: it is the justification for why `init()` exists at all.
- `metrics.rs:84` — "both counters baseline at zero from process start".

This module's docs encode the recorder-ordering invariant that makes these counters work, so
drift here costs more than ordinary comment rot.

**Proposed fix**: update line 3 to "in three unrelated groups" and add a sentence naming the pool
canonical-state group; change "both counters" → "all three counters" at lines 73 and 84; and
amend line 76 to note that `mark_drifted` is a non-drop force site — which, if anything,
_strengthens_ the argument for `init()`, since a lag event is exactly the moment an operator
needs the series to already exist with a zero baseline for `rate()` to render a step.

grantkee (dismissed):
I raised a few minor concerns regarding efficiency and completeness. I also filed a separate issues #1302 and #1303 that were uncovered during the review. Overall, good work @MavenRain 

I'm happy to rereview changes in this PR or a follow-up PR - up to you

MavenRain:
I filed a follow-up: https://github.com/Telcoin-Association/telcoin-network/issues/1304

The rest of the findings have been addressed in this PR

grantkee (approved):

Haxatron:
This fix LGTM, one improvement is that instead of covering only 100 accounts on every event, we could do it repeatedly in case a missed `Lagged` event could accumulate more than 100 accounts but I noticed this is already covered and actively tracked in https://github.com/Telcoin-Association/telcoin-network/issues/1304

PR #1245, opened 2026-08-23
fix(node): keep mid-epoch-ejected members in the boundary gossip window (#1244)

### LINKED ISSUE #1244: Mid-epoch ejection permanently bans the ejected validator network-wide: boundary epoch vote trips #1217's fatal InvalidEpochVote, so it cannot follow as an observer (deterministic eject_restart e2e failure)
## Summary

A validator that is ejected on-chain mid-epoch is permanently banned by every peer at the next epoch close.  The ban fires on the `tn-epoch-vote` gossip topic through the `FatalAuthor` penalty path.  After the ban the ejected node has zero connected primary peers, so it cannot fetch epoch records and cannot follow consensus as an observer.  Its canonical head freezes at the ejection boundary.

This breaks the documented contract of `eject::test_committee_member_restarted_mid_epoch_after_ejection` (the ejected validator must keep following the chain as an observer), and the test now fails deterministically: 2 runs on 2 machines, both timing out at `crates/e2e-tests/tests/it/eject.rs:1098` waiting for the ejected node's head.

The fatal path is new in #1217 (`59b2e9f9`, merged 2026-08-21): the `InvalidEpochVote` variant it introduced (vote digest does not match the locally expected `EpochRecord`) is classified `is_author_content_fault() == true`, which the gossip reject path maps to `Penalty::Fatal`, a permanent ban.  Before #1217 there was no fatal outcome on this topic for a current-or-recent committee member: a member's vote for an alternative record digest was admitted (the equivocation path needs it) and a non-member vote was the benign, non-penalizing `PeerNotInCommittee`.

## Observed failure

Both runs fail at the same assertion, `wait_for_head_at_least(&victim_url, tx_block + 1, ...)` at `eject.rs:1098`, where `victim_url` is `endpoints[4]`, the ejected validator-5, not the restarted validator-2:

- Machine 1 (`make attest` e2e): `Error: timed out waiting for condition: http://127.0.0.1:16628 head to reach block 6` after 194.9s.
- Machine 2 (macOS, isolated rerun with `--ignored --exact`): same error shape, `head to reach block 7`, after 915.5s.

Both runs sat on top of main `59b2e9f9` (#1217).  The only extra commit in the tree (a `tn-reth` startup storage-consistency change, PR for #1238) is exonerated by the logs: the restarted validator-2 healed cleanly and advanced from execution tip 2 to canonical head 12 by epoch 7 with zero storage errors, and it served the post-restart transfer through its own RPC.

## Timeline from the node logs

All lines below are from one failing local run (timestamps 2026-08-22, ANSI stripped).

The ejection lands in the final block of epoch 2.  Validator-5 records the shrink and still reaches quorum on the same epoch-2 record digest as its peers:

```
13:21:50.590 WARN  committee shrank mid-epoch: validator(s) ejected on-chain; recording shrunken committee ejected=[sPQpDF8...] epoch=2   (validator-5)
13:21:50.906 INFO  reached quorum on epoch close for 2/GKV5bydhCU6ghWvs   (validator-5)
13:21:50.908 INFO  reached quorum on epoch close for 2/GKV5bydhCU6ghWvs   (validator-1)
```

Within 200ms, validator-1 classifies a `tn-epoch-vote` gossip authored by validator-5 as an author content fault and bans it permanently:

```
13:21:51.089 WARN  unauthorized-author gossip - applying fatal penalty to the author, not the forwarding relayer: PeerId("12D3KooWKmyW...") topic=TopicHash { hash: "tn-epoch-vote-911329" }   (validator-1)
13:21:51.093 WARN  peer banned peer_id=PeerId("12D3KooWKmyW...")   (validator-1)
```

Every peer converges on the same ban (the author re-gossips its vote on a timer, so each recipient applies the penalty independently).  From the next heartbeat onward, validator-1 and the genesis observer each show `banned_count=1` on the primary network, and validator-5 shows zero primary peers for the rest of the run:

```
13:22:01.648 INFO  peer metrics heartbeat connected_count=0 connected_or_dialing=0 banned_count=0   (validator-5, primary; was 5/5 at 13:21:31)
```

Validator-5's epoch-record collector then requests the epoch-3 record every 5 seconds from 13:21:56 until teardown at 13:25:17, and every attempt fails, even though the live committee certified epoch 3 at 13:22:30 and certified every epoch through 7 on schedule:

```
13:24:57.594 INFO  failed to retrieve epoch from a peer 3: Could not get the epoch record!   (validator-5, ~40th attempt)
13:22:30.845 INFO  reached quorum on epoch close for 3/DwW5oBUm3YPw936F   (validator-1)
```

The control group proves the records are servable: the genesis observer, never in any committee, retrieves every record promptly and follows to canonical head 12 in epoch 7.  Validator-5's head freezes at 5 (epoch 2, 13:21:50) and never moves again.  The test then times out polling exactly that node.

## Root cause analysis

The fatal chain, all sites on current main:

1.  `crates/consensus/primary/src/network/handler.rs` (post-#1217): an epoch vote is checked against the locally stored uncertified record for that epoch.  A digest mismatch is `ensure!(record.digest() == vote.epoch_hash, PrimaryNetworkError::InvalidEpochVote(..))`.
2.  `crates/consensus/primary/src/error/network.rs`: `InvalidEpochVote` is in the `is_author_content_fault()` true-set (introduced by #1217).
3.  `crates/network-libp2p/src/consensus.rs:1174`: an author content fault on gossip resolves to `RejectPenalty::FatalAuthor` and `process_penalty(author_id, Penalty::Fatal)`.  The ban is permanent; there is no path back.

The triggering vote's digest is not logged, so the exact mismatch is inferred, but the timing supports a vote/shrink race at the boundary block: the ejection executes in the final block of epoch 2, and each node rebuilds its expected epoch-2 record (shrunken committee) at its own moment inside that second.  A vote built from the pre-shrink record state, or received by a peer whose expected record briefly differs, has a mismatched digest.  Pre-#1217 that vote was admitted or dropped benignly.  Post-#1217 it is a permanent network-wide ban of an honest node.

Note the asymmetry with the neighboring checks, which #1217 deliberately kept benign: a non-member vote is still non-penalizing `PeerNotInCommittee` (the GHSA-j2g4-553f-875r relayer-safety analysis), and a vote for an already-certified epoch is dropped with no penalty.  The digest-mismatch arm is the one new fatal outcome, and an honest node can hit it with no attacker present.

## Impact

- Any mid-epoch ejection can permanently partition the ejected node from the primary network.  The intended degradation (committee member becomes a following observer) is unreachable once every peer holds a fatal ban.
- No attacker is required.  The race is between an on-chain ejection landing near an epoch close and the ejected member's own honest epoch-close vote.
- `Penalty::Fatal` does not self-heal, so the operator of an ejected validator cannot recover the node by waiting or restarting it; every peer must forget the ban.
- e2e signal: `eject::test_committee_member_restarted_mid_epoch_after_ejection` fails deterministically, currently masking any other regression in the eject path.  The `#[ignore]`d e2e suite does not run in the PR CI lane, which is how #1217 landed green.

## Repro

```
cargo test -p e2e-tests --test it -- eject::test_committee_member_restarted_mid_epoch_after_ejection --exact --ignored --nocapture
```

Then read `crates/e2e-tests/test_logs/eject_restart/nodevalidator-5-run1.log` (frozen head, endless epoch-record retries) against `nodevalidator-1-run1.log` (the `unauthorized-author gossip` and `peer banned` pair seconds after the `committee shrank mid-epoch` warning).

## Related

- #1217 introduced the fatal digest-mismatch arm this report suspects.
- #1212 / #1213: the one-shot epoch-vote round and epoch-scoped gossip task make the boundary window this race lives in.
- #500: the standing concern that ejection interacts badly with epoch record validation.
- #873: the penalty.md Fatal-site catalog; this adds an unlisted honest-node path to the FatalAuthor row.

## Suggested direction

Keep #1217's tightening for true strangers, but make the digest-mismatch arm benign (drop, no penalty) when the vote's author is a member of either the expected record's committee or the pre-shrink committee for that epoch.  An in-flight epoch-close vote from a just-ejected member is then dropped like `PeerNotInCommittee` instead of banned.  Alternatively, downgrade `InvalidEpochVote` from the fatal set to a severe, decaying penalty, so a boundary race costs reputation but not permanent connectivity.  Either way the eject e2e test is the regression gate.

MavenRain:
Closes #1244.

## Problem

A validator ejected on-chain mid-epoch is permanently banned by every peer at the next epoch close.  After the ban the ejected node has zero connected primary peers, cannot fetch epoch records, and its canonical head freezes at the ejection boundary.  This breaks the documented contract of `eject::test_committee_member_restarted_mid_epoch_after_ejection` (the ejected validator must keep following the chain as an observer), and that e2e test fails deterministically (2 runs on 2 machines, both timing out at `eject.rs:1098` waiting for the ejected node's head).

No attacker is required.  The race is between an on-chain ejection landing near an epoch close and the ejected member's own honest epoch-close vote.

## Root cause

The failing chain runs through the network layer's boundary window, not the epoch-vote handler that #1244 first suspected (that handler arm maps `InvalidEpochVote` to a Mild penalty, which does not ban):

1.  A mid-epoch ejection mutates the CURRENT epoch's on-chain validator array in place.  The next epoch's pinned entry read (`run_epoch`, `validators_for_epochs_at_header`) therefore returns the previous epoch's post-ejection (shrunken) committee.
2.  `spawn_primary_network_for_epoch` derives both the boundary-publisher allowlist for `epoch_vote_topic` / `consensus_output_topic` and the peer manager's penalty-exemption slots from that set (the previous/current/next window, issues #715, #898, #912).  The ejected member is in no slot, so both protections miss it.
3.  The ejected member's final epoch-close vote is still in flight at the swap (the author gossips it moments before peers record the shrink, and gossipsub re-propagates it).  `verify_gossip` rejects it as `RejectReason::UnauthorizedAuthor`, which resolves to `RejectPenalty::FatalAuthor` and `process_penalty(author, Penalty::Fatal)` (`network-libp2p/src/consensus.rs`).  `Penalty::Fatal` never decays, so every peer converges on a permanent ban.

The log pair from the failing run confirms the site: `committee shrank mid-epoch: ... recording shrunken committee` followed within 200ms by `unauthorized-author gossip - applying fatal penalty to the author ... topic=tn-epoch-vote` and `peer banned`.

## Fix

Restore the pre-shrink membership when entering an epoch.  Epoch N-2's sealed record carries epoch N-1's committee as it stood at N-1's start (`next_committee`), before any mid-epoch ejection.  `run_epoch` now unions that set into the previous-committee keys threaded to the network layer.

- [x] `node/src/manager/node/run_epoch.rs`: new `widen_previous_committee_with_preshrink` helper; the entry path fetches epoch N-2's record and unions its `next_committee` into `previous_committee_keys` (epochs 0 and 1 have no grandparent record and are unchanged).  A missing record fails open with a `warn!` (the widening is network hygiene, not a consensus input;  `resolve_prior_epoch_record` still owns the halt for a corrupt record chain).
- [x] `node/src/manager/node/start_epoch.rs`: boundary-window and `create_consensus` doc comments record the new invariant.

Threat model.  The widened set feeds only gossip publisher authorization and penalty exemption, exactly the two mechanisms that were already deliberately spanning the previous/current/next window so in-flight boundary traffic from a rotating-out committee is not rejected (#912).  A member ejected mid-epoch is an outgoing committee member whose exit came early; it gets the same one-epoch grace every outgoing member already gets, and never-committee peers stay excluded.  It regains no consensus authority: quorum and voting resolve through the epoch's `ConsensusConfig` committee, and the epoch-vote handler still checks authorship against the sealed record's post-shrink committee, so its stale vote is dropped as the benign, non-penalizing `PeerNotInCommittee`.  The derivation is deterministic on every node (sealed record chain), so the allowlist and exemption windows stay identical across honest peers.

## Testing

- New unit tests (`cargo +1.94 test -p tn-node --lib widen_`): the ejected member is restored from the grandparent record, a missing grandparent is identity, and a no-shrink epoch is identity.  The restore test was mutation-confirmed: with the widening call reverted to identity, it fails; restored, it passes.
- Regression gate GREEN locally: `cargo +1.94 test -p e2e-tests --test it -- eject::test_committee_member_restarted_mid_epoch_after_ejection --exact --ignored` (the deterministic failure this fixes) passes on this branch: `test result: ok. 1 passed; 0 failed` in 2458.9s, with the ejected validator following as an observer past the boundary.
- Static ladder locally (nightly `fmt --check`, scoped `clippy -p tn-node --no-deps -- -D warnings`); the workspace CI lanes and the pinned `+1.94` attest run remotely on push.

### DIFF
--- a/crates/node/src/manager/node/run_epoch.rs
+++ b/crates/node/src/manager/node/run_epoch.rs
@@ -329,6 +329,40 @@ where
                 (previous.into_iter().collect(), next)
             };
 
+        // A mid-epoch on-chain ejection mutates the CURRENT epoch's validator array in place,
+        // so the pinned entry read above returns the previous epoch's post-ejection (shrunken)
+        // committee. The ejected member's boundary traffic is still in flight, though: its final
+        // epoch-close vote is authored moments before the shrink lands and re-propagates into
+        // this epoch. `spawn_primary_network_for_epoch` derives both the boundary-publisher
+        // allowlist and the peer manager's penalty-exemption slots from this set, so leaving the
+        // ejected member out has `verify_gossip` rejecting that vote as `UnauthorizedAuthor` and
+        // every peer applying `Penalty::Fatal`: the honest, just-ejected node is permanently
+        // banned network-wide and cannot degrade to a following observer (issue #1244). Restore
+        // the pre-shrink membership from the sealed record chain: epoch N-2's record carries
+        // epoch N-1's committee as it stood at N-1's start (`next_committee`), before any
+        // mid-epoch ejection. Epochs 0 and 1 have no grandparent record (epoch 0's committee is
+        // genesis-configured), so an ejection inside epoch 0 stays out of scope here.
+        //
+        // A missing grandparent record is a corrupt datadir for `resolve_prior_epoch_record`,
+        // but this widening is network hygiene, not a consensus input: fail open (no widening)
+        // and let the stricter path own the halt.
+        let previous_committee_keys = if entered >= 2 {
+            let grandparent = self.consensus_chain.epochs().record_by_epoch(entered - 2).await;
+            if grandparent.is_none() {
+                warn!(
+                    target: "epoch-manager",
+                    entered,
+                    missing_record = entered - 2,
+                    "grandparent epoch record missing or unreadable (record_by_epoch collapses \
+                     storage errors to None); cannot restore pre-shrink members to the boundary \
+                     gossip window"
+                );
+            }
+            widen_previous_committee_with_preshrink(previous_committee_keys, grandparent.as_ref())
+        } else {
+            previous_committee_keys
+        };
+
         let consensus_config = self
             .configure_consensus(network_config, committee, next_committee_keys, prior_epoch_record)
             .await?;
@@ -1022,6 +1056,42 @@ async fn resolve_prior_epoch_record(
     Ok((rec, digest))
 }
 
+/// Restore mid-epoch-ejected members of the previous epoch to its committee-key set.
+///
+/// `grandparent_record` is epoch N-2's sealed record when entering epoch N; its `next_committee`
+/// is epoch N-1's committee as it stood at N-1's start, before any mid-epoch on-chain ejection
+/// shrank the registry's array in place. The union is everyone who operated as a committee member
+/// at any point during epoch N-1: exactly the population whose boundary gossip (epoch-close
+/// votes, the epoch's final consensus output) is still in flight when epoch N's network comes up,
+/// and therefore the population `verify_gossip` and the peer manager's penalty exemption must not
+/// treat as never-committee strangers (issue #1244).
+///
+/// Scope: the returned set feeds ONLY the network layer (gossip publisher authorization and
+/// penalty-exemption slots). Quorum and voting always resolve through the epoch's
+/// `ConsensusConfig` committee, and the epoch-vote handler checks authorship against the sealed
+/// record's (post-shrink) committee, so an ejected member regains no consensus authority. The
+/// grace is deliberate and epoch-bounded: like any rotated-out member in the previous slot, the
+/// restored member is also unbanned and score-primed by `update_committees` for this one epoch,
+/// then ages out of the window at the next boundary.
+fn widen_previous_committee_with_preshrink(
+    onchain_previous: HashSet<BlsPublicKey>,
+    grandparent_record: Option<&EpochRecord>,
+) -> HashSet<BlsPublicKey> {
+    let preshrink: HashSet<BlsPublicKey> = grandparent_record
+        .map(|record| record.next_committee.iter().copied().collect())
+        .unwrap_or_default();
+    let ejected: Vec<BlsPublicKey> = preshrink.difference(&onchain_previous).copied().collect();
+    if !ejected.is_empty() {
+        warn!(
+            target: "epoch-manager",
+            ?ejected,
+            "previous committee shrank mid-epoch: keeping the ejected member(s) in the boundary \
+             gossip window so their in-flight epoch-close traffic is not fatally penalized"
+        );
+    }
+    onchain_previous.into_iter().chain(ejected).collect()
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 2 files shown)


grantkee (changes requested):
This behavior is intended. A forced-exit of a validator mid-epoch is an emergency boot from governance that's only reserved for the most egregious validator offenses. If a validator is ejected, they should have all authority revoked and should be blocked from participating with the protocol in any capacity.

MavenRain:
Thanks for the clarification.  I am closing this PR: restoring gossip authority for an ejected member is the opposite of the intent.

Two follow-ups:

1. The deterministic e2e failure that motivated #1244 remains on main. `eject::test_committee_member_restarted_mid_epoch_after_ejection` asserts the old contract (the burned validator keeps following the chain as an observer) and times out at its final head wait.  PR #1271 reworks the test to match the intended semantics: the five live nodes keep every detector, the burned validator's post-ejection progress is no longer asserted, and a new deterministic check requires the burned validator's BLS key to be absent from the burn epoch's certified record.

2. One scope question on "blocked from participating with the protocol in any capacity": today the network-wide ban is incidental rather than enforced.  It fires only when the ejected member's in-flight epoch-close vote is delivered after peers record the shrink.  When that vote lands before the swap, no penalty fires and the ejected node keeps following the chain and serving records indefinitely.  `test_validator_ejected_from_current_committee_mid_epoch` exercises exactly that path today and passes.  If read-only following by an ejected validator should also be blocked, peers would need to ban ejected members proactively at the epoch boundary.  I can file that as a follow-up issue if you want it enforced.

grantkee:
> 2. One scope question on "blocked from participating with the protocol in any capacity": today the network-wide ban is incidental rather than enforced. It fires only when the ejected member's in-flight epoch-close vote is delivered after peers record the shrink. When that vote lands before the swap, no penalty fires and the ejected node keeps following the chain and serving records indefinitely. test_validator_ejected_from_current_committee_mid_epoch exercises exactly that path today and passes. If read-only following by an ejected validator should also be blocked, peers would need to ban ejected members proactively at the epoch boundary. I can file that as a follow-up issue if you want it enforced.

Thanks @MavenRain - I think we continue to harden against malicious actors in general which will cover this type of scenario. Consensus tolerates up to 1/3 malicious committee members, and I think the lift is non-trivial. It may be something that comes up later, but for now I think we can safely ignore

PR #1246, opened 2026-08-23
fix(tn-reth): run reth's storage consistency check at startup (#1238)

### LINKED ISSUE #1238: Startup skips reth's static-file/DB consistency check, so a crash in the persist commit window leaves the node stuck on restart
## Summary

`persist_executed_output` commits blocks through reth's non-atomic multi-backend
commit (static files, RocksDB, MDBX). Its own doc comment says a crash in that
window leaves the static-file writer ahead of the database, and that "reth's
startup consistency check reconciles the divergence on restart." That check is
never wired into Tn-reth startup. After such a crash the node cannot heal the
divergence and gets stuck re-hitting a terminal persist error on every restart.

## Location

- Persist path: `crates/tn-reth/src/env/execution.rs:252` (`persist_executed_output`),
  commit at `:280` (`provider_rw.commit()`).
- Durability contract that relies on the startup check:
  `crates/tn-reth/src/env/execution.rs:243-247`.
- Startup provider construction: `crates/tn-reth/src/env/mod.rs:300`
  (`init_provider_factory`). It builds `ProviderFactory::new(...)` and calls
  `init_genesis` only. No consistency check.
- Retry classifier that makes the divergence terminal:
  `crates/engine/src/payload_builder.rs:355` (`persist_output_with_retry`),
  `:387` (`retryable_persist_fault`).

## The gap

`provider_rw.commit()` is reth's `DatabaseProviderRW::commit`, which is
non-atomic across three backends:

```
self.static_file_provider.finalize()?;   // independent filesystem fsync
self.rocksdb_provider.commit_batch()?;   // independent RocksDB commit
self.tx.commit()?;                        // independent MDBX transaction commit
```

Each backend is individually durable, but the group is not atomic. A crash after
the static-file finalize/fsync but before the MDBX commit leaves static files at
block N while MDBX is at N-1.

Stock reth heals this on startup. `ProviderFactory::check_consistency`
(reth v1.11.3, `crates/storage/provider/src/providers/database/mod.rs:349`) heals
file-level inconsistencies and returns unwind targets for the layers that ran
ahead; the `NodeBuilder` launch path
(`crates/node/builder/src/launch/common.rs:507-568`) runs it and unwinds to the
minimum consistent block before the node serves anything.

Tn-reth does not use `NodeBuilder`. `init_provider_factory` builds the
`ProviderFactory` directly and never calls `check_consistency`. A grep of the
whole workspace finds no call to it. So the reconciliation the persist doc
comment depends on does not run.

## Sequence

1. `persist_executed_output` for block N: `save_blocks` appends N to static files
   and the finalize step fsyncs them; then the process crashes before the MDBX
   commit.
2. Durable state after the crash: static files highest = N, MDBX tip = N-1.
   The finalized/safe marker write was in the same uncommitted MDBX transaction,
   so the on-chain marker is also N-1.
3. Restart: `init_provider_factory` opens the factory with no consistency check.
   Consensus catch-up re-delivers output from the committed marker (N-1) and the
   engine re-executes and re-persists block N.
4. `save_blocks` appends N to static files that already hold N. reth returns
   `ProviderError::UnexpectedStaticFileBlockNumber`.
5. `persist_output_with_retry` calls `retryable_persist_fault`, which classifies
   `UnexpectedStaticFileBlockNumber` as non-retryable (terminal). The persist
   fails.
6. Every subsequent restart repeats steps 3-5. The node cannot make progress.

## Impact

Liveness / durability. A single crash inside the per-commit window (power loss,
OOM kill, panic between the static-file fsync and the MDBX commit) leaves the node
unable to restart cleanly. Recovery needs manual static-file surgery (prune the
ahead static-file blocks back to the MDBX tip). No funds are at risk, but a
validator is down until an operator intervenes by hand.

## Fix

Run reth's reconciliation at startup, mirroring `launch/common.rs:507-568`. In
`init_provider_factory`, after building the `ProviderFactory` and before the node
serves, call `factory.check_consistency()` and unwind to the minimum of the
returned `(rocksdb_unwind, static_file_unwind)` targets. At minimum, heal the
static-file and RocksDB layers down to the committed MDBX tip on open. Once wired
in, the persist doc comment's stated safety net actually holds.

## Notes

- reth pin: `paradigmxyz/reth` tag `v1.11.3` (matches the cited launcher lines).
- Origin: Haxatron's Cantina collaborative-review comment on
  `persist_executed_output` (review closes 2026-09-02).
- No existing issue found for this gap (searched check_consistency / static file
  consistency / durability). Confirm before filing.

MavenRain:
Closes #1238.  (Cantina #2)

## Problem

`persist_executed_output` commits blocks through reth's non-atomic multi-backend commit: static files finalize and fsync, RocksDB commits its batch, then the MDBX transaction commits last.  Each backend is individually durable, but the group is not atomic.  A crash after the static-file finalize but before the MDBX commit durably leaves static files at block N while MDBX (and the finalized/safe markers, written in the same MDBX transaction) sit at N-1.

The persist path's own doc comment says reth's startup consistency check reconciles that divergence on restart.  It never ran: Tn-reth does not use reth's `NodeBuilder` launch path, and `init_provider_factory` built the `ProviderFactory` directly and only called `init_genesis`.  No call to `ProviderFactory::check_consistency` existed anywhere in the workspace.

The failure loop needs no attacker: after the crash, consensus catch-up re-delivers output from the committed marker (N-1), the engine re-executes block N, `save_blocks` re-appends N into static files that already hold it, reth returns `ProviderError::UnexpectedStaticFileBlockNumber`, and `retryable_persist_fault` correctly classifies that as terminal.  Every restart repeats the loop.  The validator stays down until an operator prunes the static files by hand.

## Fix

- [x] `crates/tn-reth/src/env/mod.rs`: `init_provider_factory` now runs `provider_factory.check_consistency()` after factory construction and before `init_genesis`, the same ordering as reth's launcher (`launch/common.rs`).  The check heals the crash-window direction in place: it prunes the leading static-file and RocksDB layers down to the MDBX stage checkpoints, which `save_blocks` advances inside the same MDBX transaction as the blocks, so the checkpoints always name the committed tip.
- [x] A surviving unwind target is the opposite divergence: MDBX committed past a backend that lost data.  The commit ordering cannot produce that state (MDBX commits last);  only corruption or manual datadir changes can.  Reth heals it with an unwind pipeline, but TN has no stage pipeline to unwind MDBX, so startup now refuses to start and names the lagging backends and their unwind targets instead of serving a datadir whose backends disagree.
- [x] `crates/tn-reth/src/env/execution.rs`: the module and method doc comments that promised the startup check now point at the wiring in `init_provider_factory`.

## Testing

- [x] New regression test `test_startup_heals_static_files_ahead_of_database` in `crates/tn-reth/src/env/mod.rs` simulates the exact crash window: persist block 1 fully, `save_blocks` for block 2, commit only the headers static-file writer, drop the MDBX transaction uncommitted, then rebuild a `RethEnv` over the same datadir.  Asserts the restart succeeds, the headers static-file tip healed back to the MDBX tip, and block 2 re-executes and persists cleanly.  The mid-test database-side oracle reads the `StageId::Headers` stage checkpoint (exactly what `check_consistency` compares), not `last_block_number`, which resolves through the headers static file and cannot see the divergence.
- [x] Mutation-confirmed red-first: with the `check_consistency` call removed, the same test fails at the heal assertion (the headers static-file tip stays at 2 instead of pruning back to 1);  with the fix restored it passes.
- [x] Local gates green under the pinned repo setup: the new test via `cargo test -p tn-reth --lib`, nightly `fmt -- --check`, scoped `clippy -p tn-reth --no-deps` with `-D warnings`.
- [x] CI (fmt, workspace clippy, nextest, adiri lane) fires on push;  the pinned `+1.94` attest runs on the second machine post-push.
- [x] Known pre-existing failure, not caused by this change: `eject::test_committee_member_restarted_mid_epoch_after_ejection` (the `#[ignore]`d e2e lane) times out deterministically on top of main `59b2e9f9`, tracked in #1244.  The failing wait polls the ejected validator-5, which is permanently banned network-wide when its boundary epoch-close vote trips #1217's new fatal `InvalidEpochVote` penalty.  The restarted validator-2, the only node this change touches differently, heals cleanly to canonical head 12 with zero consistency-check errors in the same runs.


sstanfield (approved):
I don't think this hurts anything but we are supposed to be explicitly not using RocksDB and I don't see a mention of MDBX.  It this actually working and/or needed in our config?  I get the rocks unwind thing may just be an artifact or something not sure how useful it will be (maybe it includes mdcx somehow?).

MavenRain:
That's the correct instinct, and no RocksDB is in play.  In our build this check is MDBX vs static files only.  The RocksDB names come from reth's API surface, not from our storage.

Details:

- We pin reth v1.11.3 with `default-features = false` and the `rocksdb` feature off. `RocksDBProvider` compiles to reth's zero-size stub. The workspace `Cargo.toml` documents this next to the reth pin.
- reth's `ProviderFactory::check_consistency()` returns the pair `(rocksdb_unwind, static_file_unwind)`. That signature is why "RocksDB" appears at the call site. In our build the first element is always `None`, for two independent reasons: the feature-off stub, and reth also gates every RocksDB invariant behind the `storage_v2` settings flag, which we never set (zero hits in the repo).
- The half that does fire compares the static-file tips against the MDBX stage checkpoints (`StageId::Headers` and friends) and prunes leading static files back to the MDBX tip. That is exactly the #1238 crash window: `persist_executed_output` fsyncs static files outside the MDBX transaction, so a crash between the two leaves static files one block ahead, and every restart then dies with `UnexpectedStaticFileBlockNumber`.
- The new test exercises our config end to end: it aborts mid-commit with the headers static file at block 2 and the MDBX Headers checkpoint at 1, reopens the same datadir, and asserts that startup prunes the static file back to 1 and that block 2 then re-persists through the production path.

So it is both working and needed in our config, and MDBX-only in practice.  The "RocksDB" string in the startup error can never print in our build because its unwind target is always `None`.  I kept the label only because the call site destructures reth's tuple; I'd be happy to drop it and report "the static files" alone if you prefer.

grantkee:
> So it is both working and needed in our config, and MDBX-only in practice. The "RocksDB" string in the startup error can never print in our build because its unwind target is always `None`. I kept the label only because the call site destructures reth's tuple; I'd be happy to drop it and report "the static files" alone if you prefer.

`match static_file_unwind` would replace ~15  lines of `filter_map`/`reduce`/`map_or` but could also become a footgun - I'm okay leaving it for now. We don't really use the reth stages eith (optimization inspired by eregon to support eth reorgs) which is another reason we should consider forking reth code

grantkee (approved):
The reason this PR is relevant is that we never enter reth's launcher. We use reth-node-builder for types only; there are zero references to NodeBuilder, LaunchContext, or with_provider_factory. Reth runs the check at launch/common.rs:510 — a path our binary never reaches. The constructor we do call carries this warning in v1.11.3:

▎ The storage backends used by the produced factory MAY be inconsistent. It is recommended to call check_consistency after creation…

Why our crates/node/ startup logic can't catch it

We have 11 startup reconciliation mechanisms, but every one reconciles consensus-DB ↔ execution tip, and every one reads that tip through APIs that resolve through the static files, not MDBX:

- static files: last_block_number(), header_by_number(), sealed_header(), block_hash()
- mdbx: best_block_number() (StageId::Finish), last_finalized_block_number()

So in the torn state, every check we own sees N+1 and reports healthy — while the engine's starting parent_header (via lookup_head() → StageId::Finish) is N. Nothing in TN compares StageCheckpoints against the static-file tip.

And the closest thing we have to a heal actively makes it worse. heal_finalized_to_persisted_tip (crates/tn-reth/src/env/epoch.rs:147) reads tip from static files (N+1) and finalized from MDBX (N), sees finalized < tip, resolves sealed_header(N+1) successfully from the static file, and then durably writes N+1 as finalized and safe into MDBX. Its own doc comment already conceded this at epoch.rs:144-146. The PR's placement fixes the ordering so that heal correctly no-ops.

Haxatron:
Fix verified. If static files/RocksDB are ahead of MDBX, `check_consistency` now automatically repairs. For the other way round, startup is immediately aborted, and an error is returned to the operator to manually repair from the snapshot.

Issue #25, opened 2024-07-18
Mix Hash Randomness


grantkee:
# Problem
Mix hash is a header field that's used by the EVM at the time of execution as a source of randomness.

Workers need this value to create the next block they propose. The current implementation uses a `keccak256` hash of the parent's `parent_beacon_block_root` (digest of consensus output) and the timestamp. However, this is only effective at preventing non-node operators from predicting the mix hash. Node operators could manipulate this value by withholding block production until a favorable timestamp. Including transaction count is just as susceptible to manipulation, and gas used can't be determined until after execution. Importantly, any value that's available to just the worker is susceptible to manipulation.

Simply using the digest from consensus output is an insufficient source of randomness because batches can be built off historic parents. An attacker could theoretically know the next digest and anticipate the mix hash in the upcoming batch by using a historical parent for the proposed batch. The ability to predict upcoming mix hashes would undermine the security of on-chain programs that rely on PREVRANDAO as a source of randomness in the EVM.

Although it seems unlikely that node operators would take advantage of this opportunity to essentially attack on-chain programs, it is unacceptable to ignore the fact that this approach requires a certain level of trust. Social governance could step in to punish nodes that manipulate this value, but it seems very difficult to prove intention.

See [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399) for more background.

## Additional Context
Mix hash MUST be difficult to predict, easy to verify, and impossible to manipulate at the time of block proposing. The same mix hash can be used and well-known during execution of consensus output because no new transactions are included in the final block.

PR #1249, opened 2026-08-24
fix(engine): derive PREVRANDAO from the epoch seed chain behind a fork gate (#1247)

### LINKED ISSUE #1247: Transaction-dependent PREVRANDAO lets a batch/header proposer grind on-chain randomness
## Summary

TN exposes the EVM `PREVRANDAO` as a value derived from the consensus-output digest and the batch digest.  Both inputs commit to transaction bytes and transaction ordering.  A malicious committing leader can build many otherwise-valid transaction selections or orderings after the causal DAG is known, compute the resulting `PREVRANDAO` for each, and publish only the candidate whose value gives it a favorable application outcome.

This is materially more biasable than Ethereum's beacon-chain RANDAO.  On Ethereum, changing transaction inclusion or ordering does not change `PREVRANDAO`.  On TN it does, so one proposal opportunity yields many candidate values instead of one.

Issue #25 ("Mix Hash Randomness", closed COMPLETED 2025-01-09) already stated the requirement: the mix hash "MUST be difficult to predict, easy to verify, and impossible to manipulate at the time of block proposing."  The current implementation does not meet the "impossible to manipulate at the time of block proposing" part.  This report is filed against the current code path, not the older `keccak256(parent_beacon_block_root, timestamp)` scheme described in #25.

All references below are on `main` at commit `648b6f81ca01cc1c1399ac2b113a450e95c11b3d`.  The code is unchanged since `7891daa5`.

## Affected code

Each batch digest commits to the complete serialized batch, which includes the ordered transaction byte vectors.  [`Batch::digest`](https://github.com/Telcoin-Association/telcoin-network/blob/648b6f81ca01cc1c1399ac2b113a450e95c11b3d/crates/types/src/worker/sealed_batch.rs#L123-L129):

```rust
pub fn digest(&self) -> BlockHash {
    let mut hasher = crypto::DefaultHashFunction::new();
    hasher.update(encode(self).as_ref());
    // finalize
    BlockHash::from_slice(hasher.finalize().as_bytes())
}
```

The committed sub-DAG digest hashes every included header digest in [`CommittedSubDag::digest`](https://github.com/Telcoin-Association/telcoin-network/blob/648b6f81ca01cc1c1399ac2b113a450e95c11b3d/crates/types/src/primary/output.rs#L561-L562):

```rust
for cert in &self.inner.headers {
    hasher.update(cert.digest().as_ref());
}
```

The consensus-header digest then hashes that committed sub-DAG in [`ConsensusHeader::digest_from_parts`](https://github.com/Telcoin-Association/telcoin-network/blob/648b6f81ca01cc1c1399ac2b113a450e95c11b3d/crates/types/src/primary/block.rs#L70-L72):

```rust
hasher.update(parent_hash.as_ref());
```

The engine uses that digest as `output_digest` in [`payload_builder.rs`](https://github.com/Telcoin-Association/telcoin-network/blob/648b6f81ca01cc1c1399ac2b113a450e95c11b3d/crates/engine/src/payload_builder.rs#L45):

```rust
let output_digest: B256 = output.digest().into();
```

For every executed batch, the engine derives the mix hash by XORing that value with the current batch digest, at [`payload_builder.rs`](https://github.com/Telcoin-Association/telcoin-network/blob/648b6f81ca01cc1c1399ac2b113a450e95c11b3d/crates/engine/src/payload_builder.rs#L202):

```rust
let mix_hash = output_digest ^ batch_digest;
```

[`TNPayload::prev_randao`](https://github.com/Telcoin-Association/telcoin-network/blob/648b6f81ca01cc1c1399ac2b113a450e95c11b3d/crates/tn-reth/src/payload.rs#L131-L132) returns that value directly, and TN places it in the block environment and header:

```rust
pub(crate) fn prev_randao(&self) -> B256 {
    self.mix_hash
}
```

## Dependency chain

```text
transaction bytes and ordering
        |
     batch_digest
        |
primary-header digest -> committed-sub-DAG digest -> output_digest
        |___________________________________|
                                        XOR -> PREVRANDAO
                         batch_digest ______|
```

## Why the XOR does not remove proposer influence

`output_digest` is a hash that contains the batch digest transitively, through the header payload and the committed sub-DAG.  It is not the batch digest itself.  XORing the hash with the standalone `batch_digest` therefore does not cancel.  Changing one transaction, its position, an attacker-controlled transaction's calldata, or which prepared batch is referenced changes `batch_digest` and changes the hash that contains it.  Each construction gives the proposer another effectively independent candidate.

A proposer that searches for `PREVRANDAO % N == x` needs about `N` candidates on average.  It does not need a full 256-bit preimage.

The in-code comment at [`payload_builder.rs:200`](https://github.com/Telcoin-Association/telcoin-network/blob/648b6f81ca01cc1c1399ac2b113a450e95c11b3d/crates/engine/src/payload_builder.rs#L200) states the XOR exists to "ensure unique mixed hash per batch."  That is a uniqueness property, not an unpredictability or non-grindability property.

## Threat model

A malicious authority acting as the committing consensus leader can prepare multiple otherwise-valid worker batches or leader-header payload selections after the relevant causal DAG is known, compute the corresponding output candidates, and publish only the favorable construction.  It grinds locally and still publishes a valid output, so the usual reveal-or-withhold cost of a randomness design does not apply.

A non-leader batch producer has less control, because it may not know the final committed sub-DAG.  An ordinary transaction submitter can perturb the value by changing signed transaction bytes, but reliably selecting a favorable final value generally needs control of, or coordination with, the batch or header proposer, because the final output is not known when an unaffiliated user submits a transaction.

The issue does not create validator state divergence.  The chosen value is committed in the canonical header, so all validators execute the same biased value.  This is distinct from #1032, which concerned a state fork from the committee-shuffle seed.

## Difference from EIP-4399

[EIP-4399](https://eips.ethereum.org/EIPS/eip-4399) exposes the preceding beacon-chain RANDAO mix.  Transaction inclusion and ordering cannot generate new candidate mixes.  A malicious Ethereum proposer can withhold its reveal or censor a transaction after seeing the mix, but it pays the opportunity cost of the proposal and has only limited influence per slot.  TN instead lets transaction and payload construction generate many candidate values inside one proposal opportunity.

## Impact

Contracts that treat `PREVRANDAO` as an Ethereum-equivalent randomness source can be biased by a proposer.  Depending on the consuming application, this can influence lotteries, randomized allocations, games, auction tie-breaking, or other value-bearing decisions.

Severity is Medium, conditional on consuming contracts relying on `PREVRANDAO` for randomness.  There is no consensus safety or state-divergence risk.

## Recommendation

Derive `PREVRANDAO` only from consensus entropy that is independent of transaction bytes, batch selection, transaction ordering, beneficiary selection, and other proposer-controlled execution contents.

TN maintains a post-fork epoch seed chain based on deterministic, digest-pinned BLS seed signatures (`crates/types/src/primary/seed_chain.rs`, rolling out under #1062 and #1086).  Use that chain with explicit domain separation and fixed, non-grindable coordinates, for example:

```text
keccak256("TN_PREVRANDAO" || seed_chain_value || consensus_output_number || batch_index)
```

`consensus_output_number` and `batch_index` are fixed by the committed order, so the proposer cannot vary them to search.  The seed chain value is fixed once the epoch seed is set, so transaction construction cannot generate new candidates.

## Credit

Reported by Haxatron on the Cantina collaborative review (finding #9, 2026-08-24).  Re-verified against `main` at `648b6f81` on 2026-08-24.

MavenRain:
Closes #1247.

## Problem

TN exposes the EVM `PREVRANDAO` as `output_digest ^ batch_digest` (`crates/engine/src/payload_builder.rs`).  Both inputs commit to transaction bytes and transaction ordering: the batch digest hashes the encoded batch, and the consensus header digest covers the committed sub-DAG, whose header digests cover the batch digests.  The XOR does not cancel, because the header digest contains the batch digest through a hash, not as a term.  A committing leader can therefore build many otherwise-valid payload selections or orderings after the causal DAG is known, compute the resulting `PREVRANDAO` for each, and publish only the one whose value favors it.  A proposer that wants `PREVRANDAO % N == x` needs about `N` candidates, not a 256-bit preimage.  Issue #25 already stated the requirement: the mix hash "MUST be difficult to predict, easy to verify, and impossible to manipulate at the time of block proposing".  The XOR gives uniqueness per batch, not non-manipulability.  There is no consensus safety or state-divergence risk in the current code: every node executes the same biased value.

## Fix

Post-fork, derive `PREVRANDAO` only from consensus entropy that is independent of payload construction:

```text
keccak256("TN_PREVRANDAO_V1" || seed chain value || consensus block number || batch index)
```

- [x] `crates/types/src/primary/output.rs`: new `ConsensusOutput::prev_randao(batch_index, batch_digest)` owns the fork dispatch;  a free `seeded_prev_randao` pins the exact byte layout;  the versioned domain tag matches the seed-chain domain style.  The seed chain value (`committee_shuffle_seed`) folds only digest-pinned deterministic BLS seed signatures (#1062, #1086), so every certificate for a leader header carries identical bytes and no payload change can move it.  The block number and batch index are fixed by the committed order.  The committing leader still learns the value moments before others, because it computes its own seed signature;  that is the same position an Ethereum proposer holds with its RANDAO reveal (withhold-or-publish), and it cannot generate additional candidates from it.
- [x] `crates/types/src/forks.rs`: `PREVRANDAO_FORK_EPOCH` (dormant `Epoch::MAX` placeholder), `prevrandao_seed_active`, and a `TN_PREVRANDAO_FORK_EPOCH` test-utils override, mirroring the seed-signature trio.  The gate requires `seed_signature_active` as a fail-closed conjunct: if a schedule or override ever orders the two forks the other way, the code keeps the legacy XOR instead of promoting the forkable legacy leader-aggregate seed (#1032) into `PREVRANDAO`.  A compile-time `const` assertion additionally enforces the rollout order on adiri: a PR that sets the PREVRANDAO fork epoch below the seed fork epoch fails to compile instead of shipping a silently dormant gate.  Pre-fork epochs replay the XOR byte-identically;  the empty epoch-close block passes a zero batch digest, which reduces the XOR to the bare consensus header digest, exactly the value that path used before.  Non-adiri builds are active from genesis, the same rollout contract as the seed-signature fork.  A follow-up rollout PR sets the concrete adiri epoch under the standard two-step hard-fork rule (deploy the dormant gate fleet-wide first, then land the epoch).
- [x] `crates/engine/src/payload_builder.rs`: both call sites (per-batch, and the empty epoch-close block) now call `output.prev_randao(...)`.

The committee-shuffle path is untouched: the shuffle seed is `TNPayload::close_epoch`, never `mix_hash`, so this change only affects the `PREVRANDAO` opcode and the header field.

## Testing

- New unit tests in `crates/types/src/primary/output.rs`: exact byte-layout pin;  per-input sensitivity;  the anti-grinding differential (two outputs that differ only in their payload digests produce the same post-fork `PREVRANDAO`, while the pre-fork XOR differs, which is exactly the grinding channel);  batch-index separation;  and the empty-block reduction.  New gate tests in `crates/types/src/forks.rs` mirror the seed-signature gate tests, and the rollout-order contract (PREVRANDAO fork at or above the seed fork) is a compile-time `const` assertion rather than a runtime test.
- `crates/engine/tests/it/main.rs`: the four mix-hash expectations now assert against `ConsensusOutput::prev_randao` with the same `(batch_index, batch_digest)` pair the engine passes, and `test_simple_basefee_penalty` additionally checks the executed value against a hand-composed keccak oracle that is independent of the library derivation.
- Local static ladder (pinned toolchains): nightly rustfmt check on `tn-types` and `tn-engine`, `+1.94` check with `--all-targets`, nightly clippy with `--all-targets --no-deps`, the new `tn-types` unit tests under `+1.94`, and the four updated engine integration tests under `+1.94`.
- Mutation-confirmed: changing the domain tag fails the byte-layout pin, and forcing the legacy arm fails the anti-grinding differential.


sstanfield (approved):
LGTM

grantkee (approved):
Well done

PR #1255, opened 2026-08-24
fix(batch-builder): size batches by encoded bytes, not InMemorySize

### LINKED ISSUE #1248: Batch producer and validator measure transaction size in different units; a remote user can loop an honest worker on peer-invalid batches
## Summary

The batch producer and the batch validator apply the same 1,000,000-byte limit to two different measurements.  The producer counts each transaction with Alloy's in-memory heuristic (`InMemorySize`).  Peers count the actual EIP-2718 encoded bytes.  A remote user can send transactions where the in-memory count is below the limit but the encoded count is above it.  The honest worker then builds a batch that its own accounting accepts but every peer rejects.  A rejected batch does not remove its transactions from the pool, so the worker rebuilds the same peer-invalid batch each interval and never reaches quorum.

This is a remote, stake-free denial of service against the worker lane that the attacker's sender address selects.  The attacker does not pay the quoted fee, because the batch never executes.

Reported by Haxatron.

## Affected code

Line anchors are at commit `7891daa5c7c34f9ba848835d81101d879d9e9307`.  The code is unchanged through `origin/main` at the time of writing.

The producer counts `tx.size()` (the in-memory heuristic) but stores `encoded_2718()` bytes.  See `crates/batch-builder/src/batch.rs` L83-L113:

```rust @quoted-code
let tx = pool_tx.to_consensus();

if total_bytes_size + tx.size() > max_size {
    best_txs.max_batch_size(&pool_tx, tx.size(), max_size);
    continue;
}

total_bytes_size += tx.size();
transactions.push(tx.into_inner().encoded_2718());
```

`tx.size()` is Alloy's `InMemorySize`.  In this repo the trait is re-exported under a different name, so the call site does not make the heuristic obvious.  See `crates/tn-reth/src/txn_pool.rs`:

```rust @quoted-code
pub use reth_primitives_traits::InMemorySize as TxnSize;
```

Peers count the encoded byte vectors.  See `crates/batch-validator/src/validator.rs` L121-L140 (`validate_batch_size_bytes`):

```rust @quoted-code
let total_bytes = transactions
    .iter()
    .map(|tx| tx.len())
    .reduce(|total, size| total + size)
    .ok_or(BatchValidationError::EmptyBatch)?;

if total_bytes > max_batch_size(epoch) {
    return Err(BatchValidationError::HeaderTransactionBytesExceedsMax(total_bytes));
}
```

Both paths use the same limit.  See `crates/types/src/worker/sealed_batch.rs` L195-L204:

```rust @quoted-code
pub fn max_batch_size(_epoch: Epoch) -> usize {
    1_000_000
}
```

A rejected batch applies no pool changes, so the worker keeps rebuilding it.  See `crates/batch-builder/src/lib.rs` L177-L215 (`spawn_execution_task`):

```rust @quoted-code
BlockSealError::QuorumRejected
| BlockSealError::AntiQuorum
| BlockSealError::Timeout
| BlockSealError::NotValidator
| BlockSealError::FailedToReport
| BlockSealError::FailedQuorum => {
    // return empty vec to indicate no transactions mined
    // NOTE: this will apply no changes to transaction pool
    Ok(MinedBatchResult { mined_transactions: vec![], changed_accounts: vec![] })
}
```

## Root cause

`InMemorySize` estimates heap and struct memory.  It is not the encoded transaction size.  For most small transactions the estimate is larger than the encoded form, so the producer looks conservative and the defect stays hidden in normal testing.  A large EIP-2930 access list reverses the relationship.

The in-memory estimate charges 32 bytes for each `B256` storage key.  In `alloy-eip2930`, `AccessListItem::size()` is:

```rust @quoted-code
mem::size_of::<Address>() + self.storage_keys.capacity() * mem::size_of::<B256>()
```

RLP encodes each 32-byte key as a one-byte string prefix (`0xa0`) followed by the 32 bytes.  Each key therefore costs about one more encoded byte than the memory estimate charges.  With 1,024 keys per transaction, this repeated per-element gap grows larger than the fixed per-transaction overhead that normally makes `InMemorySize` conservative.  A power-of-two key count keeps `Vec` capacity equal to length, so the `.capacity()` term equals `keys * 32`.

## Impact

- Severity: high.  Availability of the worker lane.
- Access: remote and stake-free.  Any account can submit the transactions.  The attacker can generate sender keys until they map to a target worker.
- Cost to attacker: none of the quoted priority fee is paid, because the batch never executes.
- Effect: in a single-worker deployment, ordinary transaction batches cannot reach quorum.  In a multi-worker deployment, the selected lane stalls or degrades.
- The producer does not run the encoded-length validator on its own outbound batch before broadcast.  `validate_batch` runs on the receive path and the vote path only, so the oversized batch does reach peers.
- This does not by itself cause divergent execution state, because peers consistently reject the oversized batch.

## Proof of Concept

Reproducer measurements:

```text
Producer InMemorySize total:       996,928 bytes
Validator EIP-2718 byte total:   1,003,439 bytes
Complete ReportBatch RPC size:   1,003,533 bytes
Batch limit:                     1,000,000 bytes
Default RPC transport limit:     1,048,576 bytes
Declared gas:                   19,200,000 / 30,000,000
```

Each transaction stays valid.  Its approximate intrinsic gas is:

```text
21,000
+ 2,400                one access-list address
+ 1,024 * 1,900        access-list storage keys
+ 91,500 * 4           zero calldata
= 2,335,000 gas per transaction
```

The 2.4M gas limit covers that cost.  All eight transactions declare 19.2M gas against the 30M batch limit.  Each transaction is also below Reth's individual transaction-size limit.  The complete `ReportBatch` request stays below the default 1 MiB RPC limit, so the producer can send it to peers.  Peers reject it because its transaction vectors exceed the 1,000,000-byte batch limit.

Apply the test to commit `7891daa5c7c34f9ba848835d81101d879d9e9307`.  Add `alloy = { workspace = true }` to `crates/batch-builder/Cargo.toml` under `[dev-dependencies]`, then add this test to the `tests` module in `crates/batch-builder/src/lib.rs`:

```rust @quoted-code
use alloy::eips::eip2930::AccessListItem;
use assert_matches::assert_matches;
use std::sync::Arc;
use tempfile::TempDir;
use tn_reth::{test_utils::TransactionFactory, RethChainSpec, TxnSize as _};
use tn_types::{
    max_batch_gas, max_batch_size, test_genesis, AccessList, BatchValidation,
    BatchValidationError, Encodable2718 as _, GenesisAccount, TaskManager,
    TransactionTrait as _, Address, Bytes, B256, MIN_PROTOCOL_BASE_FEE, U160, U256,
};
use tn_worker::WorkerRequest;

#[tokio::test]
async fn poc_in_memory_size_undercharges_batch_wire_bytes() {
    const TX_COUNT: usize = 8;
    const STORAGE_KEYS_PER_TX: usize = 1_024;
    const ZERO_CALLDATA_BYTES: usize = 91_500;
    const GAS_LIMIT_PER_TX: u64 = 2_400_000;
    const EPOCH: u32 = 0;

    let mut tx_factory = TransactionFactory::new();
    let genesis = test_genesis().extend_accounts([(
        tx_factory.address(),
        GenesisAccount::default().with_balance(U256::MAX),
    )]);
    let chain: Arc<RethChainSpec> = Arc::new(genesis.into());
    let task_manager = TaskManager::new("batch-size-mismatch-poc");
    let tmp_dir = TempDir::new().unwrap();
    let reth_env = RethEnv::new_for_temp_chain(
        chain.clone(),
        tmp_dir.path(),
        &task_manager,
        None,
    )
    .unwrap();
    let txpool = reth_env.init_txn_pool().unwrap();

    let mut builder_measured_bytes = 0usize;
    let mut validator_measured_bytes = 0usize;
    let mut total_gas_limit = 0u64;

    for tx_index in 0..TX_COUNT {
        let storage_keys = (0..STORAGE_KEYS_PER_TX)
            .map(|key_index| {
                let unique = tx_index * STORAGE_KEYS_PER_TX + key_index;
                B256::from(U256::from(unique).to_be_bytes::<32>())
            })
            .collect();
        let access_list = AccessList(vec![AccessListItem {
            address: Address::from(U160::from(tx_index + 1)),
            storage_keys,
        }]);

        let tx = tx_factory.create_explicit_eip1559(
            Some(chain.chain.id()),
            None,
            Some(0),
            Some(MIN_PROTOCOL_BASE_FEE.into()),
            Some(GAS_LIMIT_PER_TX),
            Some(Address::ZERO),
            Some(U256::ZERO),
            Some(Bytes::from(vec![0u8; ZERO_CALLDATA_BYTES])),
            Some(access_list),
        );

        builder_measured_bytes += tx.size();
        validator_measured_bytes += tx.encoded_2718().len();
        total_gas_limit += tx.gas_limit();
        tx_factory.submit_tx_to_pool(tx, txpool.clone()).await;
    }

    let max_bytes = max_batch_size(EPOCH.into());
    assert!(builder_measured_bytes <= max_bytes);
    assert!(validator_measured_bytes > max_bytes);
    assert!(total_gas_limit <= max_batch_gas(EPOCH.into()));

    // The actual producer includes all eight transactions.
    let output = build_batch(
        BatchBuilderArgs::new(txpool.clone(), Address::ZERO, EPOCH.into()),
        0,
        MIN_PROTOCOL_BASE_FEE,
    );
    assert_eq!(output.batch.transactions().len(), TX_COUNT);
    let actual_wire_bytes: usize =
        output.batch.transactions().iter().map(Vec::len).sum();
    assert_eq!(actual_wire_bytes, validator_measured_bytes);

    // The complete ordinary report still fits through the default 1 MiB RPC codec.
    let sealed_batch = output.batch.seal_slow();
    let report_request = WorkerRequest::ReportBatch {
        sealed_batch: sealed_batch.clone(),
    };
    let report_request_bytes = tn_types::encode(&report_request);
    assert!(report_request_bytes.len() <= 1024 * 1024);

    // The actual peer validator rejects precisely because its byte count exceeds 1 MB.
    let validator = tn_batch_validator::BatchValidator::new(
        reth_env,
        None,
        0,
        MIN_PROTOCOL_BASE_FEE,
        EPOCH.into(),
    );
    let result = validator.validate_batch(sealed_batch);
    assert_matches!(
        result,
        Err(BatchValidationError::HeaderTransactionBytesExceedsMax(actual))
            if actual == actual_wire_bytes
    );

    eprintln!(
        "POC PASS: builder_size={builder_measured_bytes}, wire_size={actual_wire_bytes}, \
         rpc_size={}, max={max_bytes}, gas={total_gas_limit}/{}",
        report_request_bytes.len(),
        max_batch_gas(EPOCH.into()),
    );
}
```

Run:

```bash
git submodule update --init tn-contracts
cargo test -p tn-batch-builder --lib \
  poc_in_memory_size_undercharges_batch_wire_bytes \
  -- --nocapture
```

Observed output:

```text
POC PASS: builder_size=996928, wire_size=1003439, rpc_size=1003533,
max=1000000, gas=19200000/30000000
test tests::poc_in_memory_size_undercharges_batch_wire_bytes ... ok
```

## Recommendation

Make the producer measure the same encoded length that the validator measures.  Encode each transaction once in `crates/batch-builder/src/batch.rs`, then use that length for the limit check and reuse the bytes for the batch.

- Compute the encoded form before the byte-cap check: `let encoded = tx.inner().encoded_2718();` and `let encoded_len = encoded.len();`.
- Use `encoded_len` in place of `tx.size()` in the byte-cap comparison, in the `best_txs.max_batch_size(&pool_tx, ..)` report, and in the running `total_bytes_size` total.
- Store the `encoded` vector in `transactions` instead of calling `encoded_2718()` a second time.

Any batch the producer accepts then also passes `validate_batch_size_bytes` on every honest peer.  A larger fix could add an outbound self-check that runs the batch through the encoded-length validation before broadcast, so a producer bug of this class cannot cost the node a peer penalty.

MavenRain:
Closes #1248.
(Cantina #18)
## Problem

The batch producer and the batch validator apply the same 1,000,000-byte limit
(`max_batch_size(epoch)`) to two different measurements of a transaction.  The
producer caps the batch on `TxnSize`, which is an in-memory heap-and-struct
estimate.  Peers cap the batch on the encoded EIP-2718 byte length.  A remote
user can craft transactions whose in-memory estimate stays below the limit
while the encoded length goes above it.  The honest worker then builds a batch
that its own accounting accepts but every peer rejects.  A rejected batch
applies no changes to the transaction pool, so the worker rebuilds the same
peer-invalid batch every interval and never reaches quorum.

This is a remote, stake-free denial of service against the worker lane that the
attacker's sender address selects.  The attacker pays none of the quoted
priority fee, because the batch never executes.  No attacker stake and no
protocol privilege is required.  Any account can submit the transactions.

## Root cause

The producer capped and accumulated on `tx.size()` in
`crates/batch-builder/src/batch.rs`, then stored `encoded_2718()` bytes:

```rust
if total_bytes_size + tx.size() > max_size {
    best_txs.max_batch_size(&pool_tx, tx.size(), max_size);
    continue;
}
total_possible_gas += tx.gas_limit();
total_bytes_size += tx.size();
transactions.push(tx.into_inner().encoded_2718());
```

`TxnSize` is `pub use reth_primitives_traits::InMemorySize as TxnSize`
(`crates/tn-reth/src/txn_pool.rs`).  `InMemorySize` estimates heap and struct
memory, not the wire length.  The re-export name hides the heuristic at the
call site.  For most small transactions the estimate is larger than the encoded
form, so the producer looks conservative and the gap stays hidden in normal
testing.

A large EIP-2930 access list reverses the relationship.  `alloy-eip2930`'s
`AccessListItem::size()` charges `size_of::<B256>()` (32 bytes) per storage
key.  RLP encodes each 32-byte key as a one-byte string prefix (`0xa0`) plus the
32 bytes, so the encoded form costs about one more byte per key than the memory
estimate charges.  With enough keys, this repeated per-key gap exceeds the fixed
per-transaction overhead that normally makes `InMemorySize` conservative, and
the producer undercounts the batch.

The validator measures the encoded bytes.  `validate_batch_size_bytes` in
`crates/batch-validator/src/validator.rs` sums `tx.len()` over the encoded byte
vectors and compares the total against `max_batch_size(epoch)`.  The producer
and the validator shared the limit but not the unit.

## Fix

Cap and accumulate the batch on the encoded EIP-2718 byte length, the exact
measurement the validator applies.  The producer now encodes each candidate
transaction once and sizes the batch by `encoded.len()`, so the producer's
running total equals the sum of `tx.len()` over the encoded vectors that the
validator later checks.  The byte cap is now one shared constant enforced on the
same measurement on both the producing (writer) path and the validating (reader)
path, so an honest producer can no longer build a batch that a correct peer
rejects for size.

- `crates/batch-builder/src/batch.rs`: size the batch by `encoded_2718().len()`
  instead of `TxnSize`.  The encoded bytes are computed once and reused for the
  pushed transaction, so the accept path still does a single encode, as before.
  Only a candidate that fails the byte check pays an extra encode, and that
  happens at most a few times per batch at the size boundary.  The now-unused
  `TxnSize` import is dropped.

The change is minimal by design.  Aligning the producer's unit with the
validator's unit is sufficient to close the divergence.  No separate
self-validation of the outbound batch is added, because the producer's
accounting now equals the validator's measurement exactly, so the outbound batch
cannot exceed the peer limit.

## Testing

Static gate under the pinned toolchain from `rust-toolchain.toml`:

```
cargo +1.94 clippy -p tn-batch-builder --all-targets --no-deps
```

Clean, no errors or warnings.  The `--all-targets` build confirms the existing
batch-builder test targets still compile against the changed accounting.

Regression test added to `crates/batch-builder`:
`build_batch` is driven with a pool of EIP-2930 transactions that carry a large
access list, the shape that makes `InMemorySize` undercount the encoded length.
The test asserts that the produced batch passes the validator's own
`validate_batch_size_bytes` check, so the producer can no longer emit a batch
that a correct peer rejects for size.  The test was confirmed by mutation:
reverting the accounting back to `tx.size()` makes the producer overpack and the
assertion fails.

The dynamic tier (workspace `nextest` in GitHub CI, and the pinned
`cargo +1.94 test --no-run --workspace` attest on the second machine) runs on
push.

### DIFF
--- a/crates/batch-builder/src/batch.rs
+++ b/crates/batch-builder/src/batch.rs
@@ -7,7 +7,7 @@
 //! The mined transactions are returned with the built block so the worker can update the pool.
 
 use std::collections::HashMap;
-use tn_reth::{ChangedAccount, TxPool, TxnSize};
+use tn_reth::{ChangedAccount, TxPool};
 use tn_types::{
     max_batch_gas, max_batch_size, Address, Batch, BatchBuilderArgs, Encodable2718 as _,
     TransactionTrait as _, TxHash, WorkerId, U256,
@@ -38,7 +38,8 @@ pub struct BatchBuilderOutput {
 /// Returns the [`BatchBuilderOutput`] and cannot fail. The batch continues to add
 /// transactions to the proposed block until either:
 /// - accumulated transaction gas limit reached (measured by tx.gas_limit())
-/// - max byte size of transactions (measured by tx.size())
+/// - max byte size of transactions (measured by the encoded EIP-2718 byte length, the same
+///   measurement peers apply in the batch validator)
 ///
 /// NOTE: it's possible to under utilize resources if users submit transactions
 /// with very high gas limits. It's impossible to know the amount of gas a transaction
@@ -106,24 +107,37 @@ pub fn build_batch<P: TxPool>(
             continue;
         }
 
-        // ensure block has capacity (in bytes) for this transaction
-        if total_bytes_size + tx.size() > max_size {
-            // the tx could exceed max gas limit for the block
+        // encode the transaction once and measure the batch by the encoded
+        // (EIP-2718) byte length.  Peers size a batch by exactly this value in
+        // the batch validator (`validate_batch_size_bytes` sums `tx.len()` over
+        // the encoded byte vectors against `max_batch_size(epoch)`), so the
+        // producer must cap on the same measurement.  The earlier heuristic
+        // (`TxnSize`, Reth's `InMemorySize`) estimates heap and struct memory,
+        // not the wire length; for access-list-heavy transactions it undercounts
+        // the encoded length and lets the producer build a batch that its own
+        // accounting accepts but every peer rejects, stalling the worker lane
+        // (issue #1248).
+        let tx_gas_limit = tx.gas_limit();
+        let encoded = tx.into_inner().encoded_2718();
+
+        // ensure the batch has capacity (in bytes) for this transaction
+        if total_bytes_size + encoded.len() > max_size {
+            // the tx could exceed the max byte size for the batch
             // marking as invalid within the context of the `BestTransactions` pulled in this
             // current iteration  all dependents for this transaction are now considered invalid
             // before continuing loop
-            best_txs.max_batch_size(&pool_tx, tx.size(), max_size);
+            best_txs.max_batch_size(&pool_tx, encoded.len(), max_size);
             debug!(target: "worker::batch_builder", ?pool_tx, "marking tx invalid due to bytes constraint");
             continue;
         }
 
         // txs are not executed, so use the gas_limit
-        total_possible_gas += tx.gas_limit();
-        total_bytes_size += tx.size();
+        total_possible_gas += tx_gas_limit;
+        total_bytes_size += encoded.len();
 
         // append transaction to the list of executed transactions
         mined_transactions.push(*pool_tx.hash());
-        transactions.push(tx.into_inner().encoded_2718());
+        transactions.push(encoded);
 
         // track max nonce per sender for pool state updates
         let sender = pool_tx.sender();
@@ -188,9 +202,10 @@ pub fn build_batch<P: TxPool>(
 mod tests {
     use super::*;
     use crate::test_utils::TestPool;
+    use alloy::eips::eip2930::{AccessList, AccessListItem};
     use std::sync::Arc;
     use tn_reth::{test_utils::TransactionFactory, RethChainSpec};
-    use tn_types::{test_genesis, BatchBuilderArgs, Bytes, MIN_PROTOCOL_BASE_FEE, U256};
+    use tn_types::{test_genesis, BatchBuilderArgs, Bytes, B256, MIN_PROTOCOL_BASE_FEE, U256};
 
     /// The optimistic `changed_accounts` update must carry the sender's real balance, not an
     /// inflated `U256::MAX`. An inflated balance lets the pool promote a sender's parked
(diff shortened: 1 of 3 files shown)


sstanfield (approved):

crates/batch-builder/src/batch.rs
```diff
@@ -106,24 +107,37 @@ pub fn build_batch<P: TxPool>(
...
+        // (`TxnSize`, Reth's `InMemorySize`) estimates heap and struct memory,
+        // not the wire length; for access-list-heavy transactions it undercounts
+        // the encoded length and lets the producer build a batch that its own
+        // accounting accepts but every peer rejects, stalling the worker lane
+        // (issue #1248).
+        let tx_gas_limit = tx.gas_limit();
+        let encoded = tx.into_inner().encoded_2718();
+
+        // ensure the batch has capacity (in bytes) for this transaction
+        if total_bytes_size + encoded.len() > max_size {
```
grantkee, line 124: nit: this `encoded.len()` is calculated L124 & L129 instead of a single var

grantkee (approved):

Haxatron:
Fix verified: both batch build and validation now use `len()` of encoded EIP-2718 representation of the transaction

PR #1256, opened 2026-08-25
fix(network): cap a peer's observed connection IP set (#1251)

### LINKED ISSUE #1251: Peer observed_ip_addresses set is uncapped (slow OOM + ban fan-out)
**Component:** `crates/network-libp2p/src/peers/peer.rs`
**HEAD:** `648b6f81` (main)
**Class:** Defense-in-depth (low-severity OOM / ban amplification)

## What

`observed_ip_addresses` is a `HashSet<IpAddr>` with no size cap:

`crates/network-libp2p/src/peers/peer.rs:58`
```rust
observed_ip_addresses: HashSet<IpAddr>,
```

Entries are inserted from genuine connection source IPs (`peer.rs:296` and `:322`).  A peer that connects from many distinct real source IPs grows this set without bound.  The set feeds ban fan-out: on a ban, each recorded IP is added to `banned_peers_by_ip` and produces a `Ban` action, so a large set amplifies that work.

## Why it is low (and why the obvious "cap it like multiaddrs" is wrong)

This set is **deliberately separate** from `multiaddrs`, which is already capped (`MAX_MULTIADDRS_PER_PEER` = 32, `peer.rs:27`, `note_multiaddr` at `:161`).  The two must stay separate:

- `multiaddrs` includes self-advertised addresses.  It is capped so a peer cannot inflate it via advertised records (GHSA-29v6).
- `observed_ip_addresses` is fed **only** by genuine connection source IPs, never by advertised records.  That is intentional per GHSA-6qcj: it prevents an attacker from inflating an honest peer's ban count via advertised IPs.

So the fix is not "route it through the multiaddr cap".  To grow the observed set, the attacker must genuinely connect from many distinct real source IPs (an IPv6 /64 makes this feasible).  Each entry is ~16 bytes and growth is slow.  Real, but low.

## Fix

Add a standalone cap on `observed_ip_addresses` (evict oldest or refuse past a bound), keeping it independent from the multiaddr set so both hardening properties hold.  Bound the per-ban IP fan-out accordingly.

MavenRain:
Closes #1251.

## Problem

`Peer::observed_ip_addresses` is an unbounded `HashSet<IpAddr>` (`crates/network-libp2p/src/peers/peer.rs`).  It is fed only by genuine connection source IPs (`register_incoming` / `register_outgoing`), so growing it requires the peer to really present each IP on a connection, but one IPv6 /64 makes that cheap.  A single hostile peer can grow its entry without limit (slow memory growth), and every recorded IP later becomes one `Ban` action in the per-ban fan-out (`filter_new_ips_to_ban`), so the set size is also the per-ban work factor.

The obvious fix (route the set through the existing multiaddr cap) is wrong: the two sets are deliberately separate.  `multiaddrs` includes self-advertised addresses and is capped against a republish flood (GHSA-29v6-gvv5-45gx), while `observed_ip_addresses` must never admit an advertised address at all, so an attacker cannot feed the per-IP ban counter by advertising an honest peer's IP (GHSA-6qcj-p42p-779j).  The cap here is standalone and keeps that separation intact.

## Fix

- [x] `crates/network-libp2p/src/peers/peer.rs`: new standalone bound `MAX_OBSERVED_IPS_PER_PEER = 16`, enforced by a shared `note_observed_ip` helper that replaces the duplicated insert in `register_incoming` / `register_outgoing`.
- [x] **Refuse, never evict.**  Ban accounting relies on the set growing monotonically: per-IP ban counts are incremented from this set when a ban starts (`add_banned_peer`) and decremented from it again when the ban ends (`remove_banned_peer`), so the unban-time set must contain every IP the ban-time set held.  An eviction between those two reads would strand a count and keep an IP banned after its peer is unbanned, penalizing any honest peer sharing that IP behind a NAT . . . the same harm class GHSA-6qcj closed for advertised addresses.  Refusal closes that stranding direction.  The mirror direction (an IP admitted while a ban is in force is decremented at unban without ever having been incremented) is a pre-existing hazard of re-reading the live set at both moments;  the cap neither causes nor fixes it.  A ban-time snapshot held by `BannedPeers` would close both directions and would then also permit recency-biased eviction;  that is a natural follow-up, not this change.
- [x] **Retention trade-off, stated plainly.**  Once the cap is full a later connection contributes no IP, so a ban fans out only to the first addresses the peer presented.  This does not change the containment class: the peer-identity ban is the primary control and the per-IP fan-out is defense-in-depth.  An attacker holding more addresses than any cap re-enters from a fresh address under refusal and eviction alike, while an attacker holding at most the cap's worth of addresses has every address recorded.
- [x] **Per-ban fan-out is bounded by the same constant**: a ban now yields at most `MAX_OBSERVED_IPS_PER_PEER` `Ban` actions per peer, because the fan-out source (`filter_new_ips_to_ban`) reads the capped set.
- [x] **The bound is derived, not guessed, for the honest path**: sixteen leaves generous room for real churn (dual-stack v4 + v6, DHCP renewal, mobile roaming).  A peer that exhausts the cap stops accruing new IPs, but every retained IP is one it genuinely connected from, so a later ban still lands on real attacker addresses.

## Testing

- [x] `observed_ips_clamp_at_the_cap`: 200 distinct genuine source IPs across both connection directions clamp the set at the cap;  the flood tail is refused;  the ban fan-out source is bounded by the same cap.
- [x] `observed_ips_refuse_new_entries_instead_of_evicting`: every IP admitted before the cap survives an arbitrary later flood (nothing is evicted, so ban add/remove stays symmetric), and re-presenting a recorded IP at the cap is a no-op.
- [x] Both new tests confirmed by mutation: with the capacity check dropped (`has_capacity = true`) they fail;  restored, they pass.
- [x] Local battery: nightly `cargo fmt`, `cargo clippy -p tn-network-libp2p --all-targets --all-features --no-deps -- -D warnings`, `cargo test -p tn-network-libp2p --lib`.  CI runs the workspace fmt/clippy/nextest lanes on push;  the pinned `+1.94` attest runs post-push on the second machine.

### DIFF
--- a/crates/network-libp2p/src/peers/peer.rs
+++ b/crates/network-libp2p/src/peers/peer.rs
@@ -43,6 +43,47 @@ use tracing::{error, warn};
 /// the store keeps for a peer.
 pub(crate) const MAX_MULTIADDRS_PER_PEER: usize = 1;
 
+/// Maximum number of distinct observed connection IPs retained for a single peer.
+///
+/// `observed_ip_addresses` is fed only by genuine connection source IPs (`register_incoming` /
+/// `register_outgoing`), never by self-advertised records, so growing it requires the peer to
+/// really present each IP on a connection. One IPv6 /64 makes that cheap enough to matter:
+/// without a bound one peer entry grows without limit (slow memory growth), and every recorded
+/// IP later becomes one entry of the per-ban fan-out (`filter_new_ips_to_ban` yields one `Ban`
+/// action per IP), so the set size is also the per-ban work factor (issue #1251).
+///
+/// This bound is deliberately independent of [`MAX_MULTIADDRS_PER_PEER`]. The two sets must stay
+/// separate: the multiaddr set includes self-advertised addresses and is capped against a
+/// republish flood (GHSA-29v6-gvv5-45gx), while this set must never admit an advertised address
+/// at all, so an attacker cannot feed the per-IP ban counter by advertising an honest peer's IP
+/// (GHSA-6qcj-p42p-779j). Routing this set through the multiaddr cap would collapse that
+/// separation.
+///
+/// The cap refuses a new IP once the set is full; it never admits one by evicting another. Ban
+/// accounting relies on the set only ever growing: per-IP ban counts are incremented from this
+/// set when a ban starts (`add_banned_peer`) and decremented from it again when the ban ends
+/// (`remove_banned_peer`), so the unban-time set must contain every IP the ban-time set held.
+/// An eviction between those two reads would strand a count and keep the IP banned after its
+/// peer is unbanned, penalizing any honest peer that shares the IP (the same harm class
+/// GHSA-6qcj closed for advertised addresses). Refusal keeps growth monotonic, which closes
+/// that stranding direction. The mirror direction (an IP admitted while a ban is in force is
+/// decremented at unban without ever having been incremented) is a pre-existing hazard of
+/// re-reading the live set at both moments; the cap neither causes nor fixes it. A ban-time
+/// snapshot held by `BannedPeers` would close both directions and would then also permit
+/// recency-biased eviction; that is a follow-up, not this change.
+///
+/// The retention trade-off is deliberate: once the cap is full a later connection contributes
+/// no IP, so a ban fans out only to the first addresses the peer presented. This does not
+/// change the containment class. The peer-identity ban is the primary control and the per-IP
+/// fan-out is defense-in-depth: an attacker holding more addresses than any cap re-enters from
+/// a fresh address under refusal and eviction alike, while an attacker holding at most the
+/// cap's worth of addresses has every address recorded. Every retained IP is one the peer
+/// genuinely connected from.
+///
+/// Sixteen leaves generous room for honest churn (dual-stack v4 + v6, DHCP renewal, mobile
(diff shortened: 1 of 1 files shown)


sstanfield (approved):

crates/network-libp2p/src/peers/peer.rs
```diff
@@ -43,6 +43,47 @@ use tracing::{error, warn};
...
+/// no IP, so a ban fans out only to the first addresses the peer presented. This does not
+/// change the containment class. The peer-identity ban is the primary control and the per-IP
+/// fan-out is defense-in-depth: an attacker holding more addresses than any cap re-enters from
+/// a fresh address under refusal and eviction alike, while an attacker holding at most the
+/// cap's worth of addresses has every address recorded. Every retained IP is one the peer
+/// genuinely connected from.
+///
+/// Sixteen leaves generous room for honest churn (dual-stack v4 + v6, DHCP renewal, mobile
+/// roaming) while bounding both the per-peer memory and the per-ban fan-out.
+pub(crate) const MAX_OBSERVED_IPS_PER_PEER: usize = 16;
```
grantkee, line 85: This seems high since we expect nodes to have 1 static IP address, but not a blocker bc there just needs to be some cap

grantkee (approved):

PR #1258, opened 2026-08-25
fix(network): bound peer-exchange discovery sampling to the missing target (#1253)

### LINKED ISSUE #1253: process_peer_exchange allocates + shuffles full incoming map before take()
**Component:** `crates/network-libp2p/src/peers/manager.rs`
**HEAD:** `648b6f81` (main)
**Class:** Defense-in-depth (low-severity CPU amplification)

## What

`process_peer_exchange` does an O(N) allocate-plus-shuffle over the attacker-sized incoming map before it applies `take()`:

`crates/network-libp2p/src/peers/manager.rs:652`
```rust
// collect + shuffle the whole incoming set, then take(N)
```

The work is proportional to the sender-controlled entry count, not to the final `take()` bound.

## Why it is low

The incoming message size is bounded by the codec's max message size, so N has a ceiling.  The amplification is a shuffle over that bounded N per exchange, not unbounded work.  Low.

## Fix

Bound the collection before the shuffle, or sample `take()` entries without materializing and shuffling the whole map, so the CPU cost tracks the output size rather than the sender-supplied input size.

MavenRain:
Closes #1253.

## Problem

`process_peer_exchange` collects every eligible entry of the incoming `PeerExchangeMap` into a `Vec`, shuffles the whole `Vec`, and only then applies `take(peers_to_take)`.  The allocation and the shuffle scale with the sender-controlled entry count, not with the number of discovery slots this node still has open.  Any disconnecting peer can pack the exchange message up to the codec's max message size and make the receiver pay a full collect plus shuffle to insert at most a handful of peers.  The codec bound keeps the entry count finite, so this is a low-severity CPU amplification, not an unbounded DoS.  No attacker coordination is required.

## Fix

- [x] `crates/network-libp2p/src/peers/manager.rs`: reservoir-sample the missing target number of eligible peers directly from the exchange iterator with `IteratorRandom::choose_multiple`, replacing collect + shuffle + take.  The selection stays uniform over the eligible set.  The buffer never holds more than `max_discovery_peers - current_count` entries and no shuffle over the full map runs.  The bound comes from `PeerConfig::max_discovery_peers()`, the same parameter the discovery heartbeat already enforces, so the cap is shared, not guessed.  The per-entry eligibility filter still runs once per incoming entry;  that pass is inherent to reading the message and is cheap membership work, while the expensive allocate-and-shuffle step now tracks the output size.
- [x] Drop the now-unused `SliceRandom` import.

## Testing

- [x] New regression test `test_process_peer_exchange_bounds_discovery_to_missing_target`: with 3 discovery peers pre-seeded, an exchange map holding `max_discovery_peers + 8` eligible peers tops the set up to exactly `max_discovery_peers`, the pre-seeded peers survive, and exactly the missing target number of entries comes from the exchange map.  Confirmed by mutation: with the sample bound lifted, and separately with the bound not subtracting `current_count`, the test fails on the count assertions.
- [x] New regression test `test_process_peer_exchange_skips_when_discovery_full`: with the discovery set already at `max_discovery_peers`, an exchange adds nothing.
- [x] `cargo +nightly fmt -p tn-network-libp2p -- --check` green.
- [x] `cargo +1.94 clippy -p tn-network-libp2p --all-targets --no-deps` green.
- [x] `cargo +1.94 test -p tn-network-libp2p --lib process_peer_exchange` green (both peer-exchange tests).

CI runs the workspace lanes on push;  the pinned `+1.94` attest runs post-push on the second machine.

### DIFF
--- a/crates/network-libp2p/src/peers/manager.rs
+++ b/crates/network-libp2p/src/peers/manager.rs
@@ -17,7 +17,7 @@ use crate::{
     types::{NetworkInfo, NetworkResult, RpcInfo},
 };
 use libp2p::{core::ConnectedPoint, kad::PeerInfo, multiaddr::Protocol, Multiaddr, PeerId};
-use rand::seq::{IteratorRandom as _, SliceRandom as _};
+use rand::seq::IteratorRandom as _;
 use std::{
     collections::{HashMap, HashSet, VecDeque},
     net::IpAddr,
@@ -657,8 +657,12 @@ impl PeerManager {
 
         // seed discovery peers from peer exchange
         if current_count < max_discovery_peers {
-            // convert eligible peers to `PeerInfo` for processing
-            let mut peers: Vec<_> = peers
+            // reservoir-sample the missing target number of eligible peers so the
+            // collected buffer and the shuffle stay bounded by the open discovery
+            // slots; the per-entry eligibility pass still visits each codec-bounded entry
+            let peers_to_take = max_discovery_peers - current_count;
+            let mut rng = rand::rng();
+            let peers = peers
                 .into_iter()
                 .filter_map(|(_, (net_key, addrs))| {
                     let info =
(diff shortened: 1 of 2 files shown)


sstanfield (approved):

grantkee (approved):
Nice

PR #1261, opened 2026-08-25
fix(network): bound discovery inserts at max_discovery_peers (#1252)

### LINKED ISSUE #1252: process_peers_for_discovery extends set with no insert-time cap
**Component:** `crates/network-libp2p/src/peers/manager.rs`
**HEAD:** `648b6f81` (main)
**Class:** Defense-in-depth (low-severity, transient)

## What

`process_peers_for_discovery` extends its working set with no cap at insert time:

`crates/network-libp2p/src/peers/manager.rs:1103`
```rust
// extend with discovered peers, no insert-time bound
```

The sibling `process_peer_exchange` does cap at insert (`manager.rs:652-690`), so the two paths are inconsistent.

## Why it is low

The discovery set is re-bounded every `discovery_heartbeat`, so any overshoot is transient and self-correcting rather than a durable growth.  The inconsistency with the capped PEX path is the concern, not a standalone OOM.

## Fix

Apply an insert-time cap in `process_peers_for_discovery` matching the PEX path, so the bound holds within a heartbeat window and the two discovery paths are consistent.

MavenRain:
Closes #1252.

## Problem

`process_peers_for_discovery` extends `discovery_peers` with every eligible peer it receives.  There is no bound at insert time.  A kad closest-peers response is attacker-influenced input, so one burst of responses can grow the set past `max_discovery_peers` until the next `discovery_heartbeat` prunes it.  The sibling inlet `process_peer_exchange` already enforces its bound at insert time, so the two discovery inlets enforce different bounds on the same set.  The overshoot is transient and self-correcting, so this is defense-in-depth, not a standalone OOM.

## Fix

- [x] `crates/network-libp2p/src/peers/manager.rs`: `process_peers_for_discovery` merges the eligible newcomers, then randomly evicts down to `max_discovery_peers` over old and new entries alike.  The bound now holds at the end of every call, not only at the next heartbeat.
- [x] The bound is enforced by eviction, never by rejecting newcomers.  A first-come cap would let whoever fills the set first (or an attacker feeding `max_discovery_peers` eligible-but-unreachable ids while the node sits at `target_num_peers` and dials nothing) own the pool and starve fresh kad results indefinitely.  Random eviction over the union keeps the pool rotating with exactly the selection semantics of the heartbeat prune it mirrors, so no attacker-independent behavior changes.
- [x] The bound is the same shared config value (`PeerConfig::max_discovery_peers`) that the PEX inlet and the heartbeat prune already use.  No new constant is introduced.  `process_peer_exchange` stays fill-to-spare-capacity only, which keeps kad-discovered peers prioritized over exchange peers, per its doc.
- [x] Doc comment on the function states the bound, the eviction rationale, and the kad-over-PEX prioritization.

## Testing

- [x] New regression test `test_process_peers_for_discovery_caps_inserts`: a single batch of `max_discovery_peers + 10` eligible peers lands the set exactly at the cap;  a second batch at the cap leaves it exactly at the cap;  an address update for a tracked peer replaces its entry without growing the set and evicts nothing.  The test is deterministic (size assertions only, no sleeps, no wall-clock assumptions).
- [x] Mutation-confirmed: with the sample-down arm reverted to the old unconditional `extend`, the new test fails;  with the fix restored, it passes.
- [x] Local static ladder: `cargo +nightly fmt -- --check` and scoped `clippy -p tn-network-libp2p --all-targets --all-features --no-deps -- -D warnings` green;  `cargo test -p tn-network-libp2p` green.
- [x] Existing discovery tests (duplicate filtering, self-id filtering, heartbeat pruning) unchanged and passing;  the heartbeat prune still runs as the backstop.

### DIFF
--- a/crates/network-libp2p/src/peers/manager.rs
+++ b/crates/network-libp2p/src/peers/manager.rs
@@ -1111,12 +1111,31 @@ impl PeerManager {
 
     /// Process newly discovered peers for potential dial attempts.
     ///
-    /// Only eligible peers are stored for dialing during heartbeat.
+    /// Only eligible peers are stored for dialing during heartbeat. The set is bounded to
+    /// `max_discovery_peers` at insert time, so the bound holds within a heartbeat window
+    /// instead of relying on the next heartbeat to prune overshoot (issue #1252). The bound is
+    /// enforced by merging the newcomers and then randomly evicting down to the cap, never by
+    /// rejecting newcomers: a first-come cap would let whoever fills the set first (or an
+    /// attacker feeding eligible-but-unreachable ids) own the pool and starve fresh kad
+    /// results, while random eviction over the union keeps the pool rotating exactly like the
+    /// heartbeat prune it mirrors. [`Self::process_peer_exchange`] stays fill-to-spare-capacity
+    /// only, which keeps kad-discovered peers prioritized over exchange peers.
     pub(crate) fn process_peers_for_discovery(&mut self, mut peers: Vec<PeerInfo>) {
         peers.retain(|peer| self.eligible_for_discovery(peer));
-        let peers: HashSet<_> = peers.into_iter().map(|info| (info.peer_id, info.addrs)).collect();
         trace!(target: "peer-manager", ?peers, "adding eligible peers to discovery map");
-        self.discovery_peers.extend(peers);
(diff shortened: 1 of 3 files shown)


sstanfield (approved):

grantkee (approved):
LGTM!

PR #1265, opened 2026-08-25
fix(worker): bound the shed-task fan-out for denied inbound sync streams (#1254)

### LINKED ISSUE #1254: Worker spawns a shed task per inbound sync stream before the admission check
**Component:** `crates/consensus/worker/src/network/mod.rs`
**HEAD:** `648b6f81` (main)
**Class:** Defense-in-depth (low-severity task fan-out)

## What

The worker spawns a task per inbound sync stream **before** the admission check runs:

`crates/consensus/worker/src/network/mod.rs:314`
```rust
// task spawned per inbound sync stream, ahead of the admission gate
```

So the number of in-flight spawned tasks is gated only by the libp2p per-connection substream limit, not by an application-level admission bound.

## Why it is low

The libp2p per-connection substream cap bounds the fan-out per connection.  That cap lives in the libp2p configuration (out of the scanned scope), so the exposure depends on that limit rather than on an uncapped worker-side spawn.  Low, and noted for completeness.

## Fix

Move the admission check ahead of the task spawn, or gate the spawn behind the same per-peer concurrency budget the sync handlers already use, so a burst of opened substreams cannot fan out to more tasks than admission would allow.

MavenRain:
Closes #1254.

## Problem

`process_inbound_sync_stream` checks admission before it spawns, but only for the admitted path. A denied stream still gets its own spawned task, whose one job is to write `Deny(AtCapacity)` and close. The deny write is the right protocol (the requester fails fast and tries elsewhere), but the spawn itself had no application-level bound. A peer that opens a burst of substreams beyond the admission caps fans the worker out to one shed task per stream, bounded only by the libp2p per-connection substream limit, which lives outside this crate.

No attacker sophistication is required: opening substreams is cheap for the requester, and every one of them costs the worker a spawned task before any admission decision applies to it.

## Fix

- [x] `crates/consensus/worker/src/network/mod.rs`: add `MAX_CONCURRENT_SHED_TASKS` (8) and a `shed_task_semaphore` on `WorkerNetwork`. The shed write for a denied stream now runs on its own budgeted task (`shed_inbound_sync_stream`). The budget check is a `try_acquire` ahead of the spawn, so it is cheap-first and attacker-independent.
- [x] Past the budget the stream is dropped without a spawn. The requester sees a reset instead of a `Deny` frame and retries elsewhere, the same recovery it already has for a lost stream.
- [x] The admitted path now spawns only after `try_admit_sync` returns a permit, so no code path spawns ahead of an admission decision. Total sync-task fan-out is bounded by `MAX_CONCURRENT_BATCH_STREAMS` (5) admitted tasks plus `MAX_CONCURRENT_SHED_TASKS` (8) shed tasks.
- [x] The shed permit is held for the task lifetime and the deny write keeps its `SYNC_REQUEST_READ_TIMEOUT` bound, so a peer that never reads cannot pin a budget slot.

## Testing

- Nightly `fmt --check` on `tn-worker`
- `clippy -p tn-worker --no-deps --all-targets` with `-D warnings`
- `tn-worker` unit tests `network::tests` (new `shed_admit_enforces_budget_and_frees_on_drop` next to the existing admission test; confirmed by mutation: with `try_admit_shed` forced to `None` the test fails, restored green)
- CI (nightly fmt, workspace clippy, nextest) fires on push; the pinned `+1.94` attest runs on the second machine.

### DIFF
--- a/crates/consensus/worker/src/network/mod.rs
+++ b/crates/consensus/worker/src/network/mod.rs
@@ -69,6 +69,18 @@ pub const MAX_PENDING_REQUESTS_PER_PEER: usize = 2;
 /// distinct (possibly forged) digests. Load beyond the cap is shed, not queued.
 pub const MAX_CONCURRENT_GOSSIP_PREFETCHES: usize = 8;
 
+/// Maximum number of concurrent shed tasks for denied inbound sync streams.
+///
+/// A denied stream still gets a short-lived task that writes
+/// [`DenyReason::AtCapacity`] so the requester fails fast instead of waiting out
+/// a timeout. Without a budget on that spawn, a burst of opened substreams
+/// beyond the admission caps fans out to one task each, bounded only by the
+/// libp2p per-connection substream limit (#1254). Past this budget the stream
+/// is dropped without spawning (the requester sees a reset and retries
+/// elsewhere), so the worker's total sync-task fan-out stays bounded by
+/// [`MAX_CONCURRENT_BATCH_STREAMS`] admitted tasks plus this many shed tasks.
+pub(crate) const MAX_CONCURRENT_SHED_TASKS: usize = 8;
+
 /// Timeout for reading the opening request frame of an inbound sync stream.
 ///
 /// A peer that opens a sync stream but never sends its request frame trips this
@@ -122,6 +134,15 @@ fn try_admit_sync(
     })
 }
 
+/// Try to reserve a slot in the bounded shed-task budget.
+///
+/// Returns `None` when [`MAX_CONCURRENT_SHED_TASKS`] shed tasks are already in
+/// flight; the caller then drops the stream without spawning. Dropping the
+/// permit frees the slot for the next denied stream.
+fn try_admit_shed(semaphore: &Arc<Semaphore>) -> Option<OwnedSemaphorePermit> {
+    semaphore.clone().try_acquire_owned().ok()
+}
+
 /// Handle inter-node communication between primaries.
 #[derive(Debug)]
 pub struct WorkerNetwork<DB, Events> {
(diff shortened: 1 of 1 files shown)


sstanfield (approved):

crates/consensus/worker/src/network/mod.rs
```diff
@@ -459,4 +519,25 @@ mod tests {
         assert_eq!(semaphore.available_permits(), MAX_CONCURRENT_BATCH_STREAMS);
         assert!(sync.lock().get(&p).is_none());
     }
+
+    // The shed budget admits exactly MAX_CONCURRENT_SHED_TASKS concurrent shed
+    // tasks, denies the next, and frees a slot when a shed permit drops.
+    #[test]
+    fn shed_admit_enforces_budget_and_frees_on_drop() {
```
grantkee, line 526: This test is mostly pinning "non-blocking, fallible, owned-permit", but doesn't cover the deny -> shed routing, the binding to `self.shed_task_semaphore`, or the permit's hold-for-task-lifetime

Consider refactoring to use generic stream and updating test like `send_sync_batches_over_stream`:
```rust
fn shed_sync_stream<S>(shed_sem: &Arc<Semaphore>, spawner: &TaskSpawner, epoch: Epoch, peer: BlsPublicKey, stream: S) where S: futures::AsyncWrite + Unpin + Send + 'static
```

Then test with a writer that always returns `Poll::Pending`, under `#[tokio::test(start_paused = true)]`

```rust
for _ in 0..MAX_CONCURRENT_SHED_TASKS { shed_sync_stream(&sem, &spawner, 0, peer(), PendingWriter); }
assert_eq!(sem.available_permits(), 0, "each shed task holds its slot while it runs");
// over budget: no spawn, no negative accounting
shed_sync_stream(&sem, &spawner, 0, peer(), PendingWriter);
assert_eq!(sem.available_permits(), 0);
// the bounded write trips SYNC_REQUEST_READ_TIMEOUT and every slot returns
tokio::time::advance(SYNC_REQUEST_READ_TIMEOUT + Duration::from_secs(1)).await;
assert_eq!(sem.available_permits(), MAX_CONCURRENT_SHED_TASKS);
```

crates/consensus/worker/src/network/mod.rs
```diff
@@ -415,6 +433,48 @@ where
...
+    ///
+    /// Spawns a short-lived task that writes [`DenyReason::AtCapacity`] and
+    /// closes, so the requester fails fast and tries elsewhere. The spawn is
+    /// gated by [`MAX_CONCURRENT_SHED_TASKS`]: past that budget the stream is
+    /// dropped without spawning (the requester sees a reset), so a burst of
+    /// over-cap substreams cannot fan out to unbounded tasks (#1254).
+    fn shed_inbound_sync_stream(&self, peer: BlsPublicKey, stream: Stream) {
+        try_admit_shed(&self.shed_task_semaphore).map_or_else(
+            || {
+                debug!(target: "worker::network", %peer, "dropping inbound sync stream: shed budget exhausted");
```
grantkee, line 447: Follow up issue to track this as a metric #1307

grantkee (approved):
LGTM - two follow up issues and one suggestion to improve testing

crates/consensus/worker/src/network/mod.rs
MavenRain: Created a follow-up issue for this: https://github.com/Telcoin-Association/telcoin-network/issues/1314

PR #1267, opened 2026-08-25
feat(consensus,config,network-types): per-worker LocalNetwork seam (#556, PR 3)

### LINKED ISSUE #556: feat: Multi-Worker Support — PR 3: Per-Worker `LocalNetwork`
## Problem

## Goal

- RPC server (unique port)
- Transaction pool

## Why

## Design Constraints

## Current State

---

## This PR: Per-Worker `LocalNetwork`

Each worker gets its own `LocalNetwork` instance for communicating with the primary. This is the seam for future process separation — today these are in-process channels, but they can be replaced with remote RPC clients without changing the worker or primary code. With `num_workers = 1`, there is exactly one `LocalNetwork` and behavior is identical to today.

### Scope

**`crates/config/src/consensus.rs` — `ConsensusConfig`:**
- Change the `LocalNetwork` field from a single instance to `local_networks: Vec<LocalNetwork>` (indexed by `WorkerId`)
- Add `local_network(worker_id: WorkerId) -> &LocalNetwork` accessor

**`crates/network-types/src/local.rs` — `LocalNetwork`:**
- No structural changes needed — each instance is already self-contained
- Each worker creates its own `LocalNetwork` and registers its own handlers
- The primary-side handler (`WorkerReceiverHandler`) accepts `WorkerOwnBatchMessage` which already contains `worker_id`

**`crates/consensus/primary/src/` — Primary receiver:**
- `WorkerReceiverHandler` already handles `WorkerOwnBatchMessage` with `worker_id` — no handler change needed
- The primary must register as the `worker_to_primary_handler` on **each** worker's `LocalNetwork` instance

**`crates/consensus/worker/src/worker.rs` — Worker:**
- Currently receives `LocalNetwork` from `ConsensusConfig`
- Change to receive its specific `LocalNetwork` instance by `worker_id` index

blocked by #554

MavenRain:
Closes #556.

## Summary

Give every worker its own `LocalNetwork` instance for primary communication, and share the proposer's header slots across workers.  This is PR 3 of the multi-worker series (#554 - #559): the seam lands before the spawn loop (#557), and with one worker the node behaves as before, except that two silent failure modes now surface loudly (called out below).

## Changes

- [x] `crates/config/src/consensus.rs`: `ConsensusConfigInner` holds `local_networks: BTreeMap<WorkerId, LocalNetwork>`, one instance per committee worker id, built in `new_with_committee` from `Committee::worker_ids()`.  The key set derives from the chain-carried count (#554 / #1219), not from any per-operator value, so it cannot drift from the committee the config carries.  The accessor is total: `local_network(worker_id) -> Option<&LocalNetwork>`;  `local_networks()` iterates all instances in ascending id order.
- [x] `crates/types/src/committee.rs`: `Committee::worker_ids()` walks `0..number_of_workers()` over the `WorkerId` domain directly, with no numeric cast.
- [x] `crates/network-types/src/local.rs`: `report_own_batch` / `report_others_batch` return `Err` on a missing handler, symmetric with `synchronize` / `fetch_batches`.  Before, they warned and returned `Ok(())`: a missed registration made `Worker::seal` report success, the batch builder took its success branch and evicted the batch's transactions from the pool as mined, and the digest never reached the proposer.  The `FailedToReport` branch built for exactly this case was unreachable; it is now reachable.  Handler setters are one-shot: a second registration returns `Err` instead of silently overwriting.
- [x] `crates/network-types/src/notify.rs`: `WorkerOwnBatchMessage` / `WorkerOthersBatchMessage` fields are private behind constructors and read accessors, so the worker id is stamped once at the sending seam and is read-only after.
- [x] `crates/consensus/primary/src/network/mod.rs` and `primary.rs`: the primary registers a `WorkerReceiverHandler` on every worker's instance, each handler bound to that instance's worker id.  A message whose stamped id disagrees with the instance it arrived on is rejected before the digest channel or the payload store.  The seam establishes the reporting worker's identity instead of trusting the message's self-stamped id . . . the invariant the comment block in `header_validator.rs` attributes to the primary, now actually enforced.  No attacker is required today (all senders are in-process), but once workers are separate processes the id arrives over a wire, and this binding is what keeps it trustworthy.
- [x] `crates/consensus/primary/src/state_sync/header_validator.rs`: the missing-batch sync loop fetches `local_network(worker_id)` per worker.  `worker_id` is peer-supplied here, so an id outside the committee's worker set becomes `HeaderError::UnkownWorkerId` (never a skip, a fallback instance, or a panic).
- [x] `crates/consensus/worker/src/worker.rs` and `network/handler.rs`: a worker takes its own instance by id; a miss surfaces as an error (`new_worker` now returns `eyre::Result`).
- [x] `crates/consensus/executor/src/subscriber.rs`: `fetch_batches` partitions `header.payload()` per `WorkerId`, and `fetch_batches_from_peers` fans out one concurrent leg per worker over that worker's instance (`try_join_all`, merged).  Each leg is raced against a stall watchdog that warns every 300 s without cancelling: the fetch stays unbounded on purpose (the sub-dag is committed, so a quorum holds its batches and `BatchFetcher::fetch_for_primary` retries until they arrive; a hard deadline would fail-stop the node during a transient outage and, on operator restart, re-enter the identical window against the still-unreachable peers).  The issue amendment asked for a per-leg timeout; self-review flagged that restart hazard, so this lands as loud-stall-not-cancel . . . flip to a terminal deadline in review if preferred.  A payload worker id with no instance is `SubscriberError::UnexpectedWorkerId` (sub-dag payloads passed `Header::validate`, so this is a protocol violation, not a routine miss).
- [x] `crates/consensus/primary/src/proposer.rs`: the pending digest queue becomes one FIFO lane per worker (`BTreeMap<WorkerId, VecDeque<ProposerDigest>>`).  Header slots are distributed uniformly across the workers with pending digests: rounds of one digest per worker in ascending id order until `max_header_num_of_batches` fills, so a busy worker cannot head-of-line block a quiet one (design discussion recorded on #556, 2026-08-18).  A round-robin cursor persists across headers: the walk resumes after the last lane served, so when more workers have pending digests than a header holds, no lane is excluded permanently.  FIFO holds within each lane; there is no total order across lanes.  `enough_digests` counts the total across lanes; repropose returns an expired header's digests to the front of its lanes.  With one worker the drain is exactly the old FIFO.
- [x] `crates/node`: epoch entry refuses to start when the committee's worker count disagrees with the gas accumulator's slot count.  Both derive from the on-chain `WorkerConfigs` state, so a mismatch means the views disagree; refusing beats running with aliased worker lanes or feeding the accumulator ids it panics on.  `PrimaryNode::new` and the worker spawn path propagate the new fallible registrations.

## Behavior at `num_workers = 1`

Identical wiring: one instance, id 0, same registration order.  Two deliberate deltas, both converting silent failures to loud ones:

1.  A batch report that reaches an unregistered seam now fails `seal()` (transactions stay in the pool) instead of silently discarding the digest while evicting the transactions as mined.
2.  A batch fetch against an unresponsive worker now warns every 300 s while it waits, instead of stalling forever with no signal anywhere.  Liveness is unchanged: the fetch still completes whenever the batches arrive.

## Testing

New tests, each confirmed by reverting the code it guards and watching it fail:

- `tn-types`: `worker_ids_walk_the_committee_count_in_order`
- `tn-network-types`: `report_paths_surface_a_missing_handler`, `handlers_bind_exactly_once`
- `tn-primary`: `test_worker_receiver_handler_bound_to_worker_id` (identity binding), `test_drain_digests_shares_header_slots_across_workers` (fair slot sharing, starvation regression), `test_drain_digests_rotates_across_headers_when_lanes_exceed_cap` (cursor rotation when lanes exceed the header cap), `test_drain_digests_single_worker_is_fifo` (degenerate case)
- `tn-executor`: `fetch_batches_from_peers_fans_out_per_worker` (per-worker legs, merge, `UnexpectedWorkerId`)

Local static ladder under the pinned +1.94 toolchain: fmt, check with all targets, and clippy (scoped `-p` list, `--no-deps`, `-D warnings`) over the eight touched crates, plus nightly `fmt -- --check`.  Targeted test runs for the crates above under the pinned toolchain.  CI runs the full workspace ladder on push.

## Out of scope

Per-worker swarms (#555), the `EpochManager` spawn loop and collections (#557), engine initialization (#558), and the per-worker tx-pool base-fee isolation discussed on the issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11801,6 +11801,7 @@ dependencies = [
  "rustversion",
  "serde",
  "tn-types",
+ "tokio",
  "tracing",
 ]
 
--- a/crates/batch-builder/src/lib.rs
+++ b/crates/batch-builder/src/lib.rs
@@ -549,7 +549,9 @@ mod tests {
         let address = Address::from(U160::from(33));
         let client = LocalNetwork::new_with_empty_id();
         let worker_to_primary = Arc::new(MockWorkerToPrimaryHang {});
-        client.set_worker_to_primary_local_handler(worker_to_primary);
+        client
+            .set_worker_to_primary_local_handler(worker_to_primary)
+            .expect("register mock primary handler");
         let temp_dir = TempDir::new().unwrap();
         let store = open_db(temp_dir.path());
         let qw = TestMakeBlockQuorumWaiter::new_test();
@@ -679,7 +681,9 @@ mod tests {
         let address = Address::from(U160::from(33));
         let client = LocalNetwork::new_with_empty_id();
         let worker_to_primary = Arc::new(MockWorkerToPrimaryHang {});
-        client.set_worker_to_primary_local_handler(worker_to_primary);
+        client
+            .set_worker_to_primary_local_handler(worker_to_primary)
+            .expect("register mock primary handler");
         let temp_dir = TempDir::new().unwrap();
         let store = open_db(temp_dir.path());
         let seal_timeout = Duration::from_secs(5);
--- a/crates/batch-builder/tests/it/build_batches.rs
+++ b/crates/batch-builder/tests/it/build_batches.rs
@@ -41,7 +41,9 @@ async fn test_make_batch_el_to_cl() -> eyre::Result<()> {
 
     // Mock the primary client to always succeed.
     let mock_server = MockWorkerToPrimary();
-    network_client.set_worker_to_primary_local_handler(Arc::new(mock_server));
+    network_client
+        .set_worker_to_primary_local_handler(Arc::new(mock_server))
+        .expect("register mock primary handler");
 
     let qw = TestMakeBlockQuorumWaiter::new_test();
     let timeout = Duration::from_secs(5);
--- a/crates/config/src/consensus.rs
+++ b/crates/config/src/consensus.rs
@@ -3,7 +3,7 @@ use crate::{
     Config, ConfigFmt, ConfigTrait as _, KeyConfig, NetworkConfig, Parameters, TelcoinDirs,
 };
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
     sync::Arc,
 };
 use tn_network_types::local::LocalNetwork;
@@ -22,7 +22,12 @@ struct ConsensusConfigInner<DB> {
     node_storage: DB,
     key_config: KeyConfig,
     authority: Option<Authority>,
-    local_network: LocalNetwork,
+    /// One [`LocalNetwork`] per worker id, keyed by the committee's worker ids.
+    ///
+    /// Each worker communicates with the primary over its own instance: this is the seam for
+    /// future process separation. Sized from [`Committee::number_of_workers`], the chain-derived
+    /// count, so the key set cannot drift from the committee the config carries.
+    local_networks: BTreeMap<WorkerId, LocalNetwork>,
     network_config: NetworkConfig,
     genesis: HashMap<HeaderDigest, Certificate>,
     /// Digest of the previous epoch's `EpochRecord` ([`EpochDigest::default`] for epoch 0).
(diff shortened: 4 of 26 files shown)


sstanfield (approved):
LGTM

grantkee (commented):
LGTM!

Issue #1259, opened 2026-08-25
batch gossip exposes transactions to priority-fee poaching by re-packing


MavenRain:
**Component:** worker batch gossip + `crates/engine/src/payload_builder.rs` fee crediting
**Class:** economic (public-mempool MEV);  low severity, known-hard

## What

PR #1224 (#1222) credits each batch's priority fees to the digest-bound `Batch::beneficiary`, which closes fee redirection by digest copying.  A transaction-level residual remains, flagged by @sstanfield in the PR review:

> This seems good.  I think there could still be a related attack where the validator poaches individual transactions.  That may be hard to close...

A validator receives another validator's batch during gossip, before certification.  Batch contents are plaintext, so it can extract the transactions, build its own batch containing them, and set itself as `Batch::beneficiary`.  If its batch is ordered at or before the victim's, it collects those transactions' priority fees.

## Why it is low

- The poacher does real work: it must build, gossip, and certify its own batch through the normal path.  Nothing is redirected from another validator's certified work.
- Only the first-ordered copy collects.  The later duplicates fail the nonce check at execution, so the poach is an inclusion race on publicly visible transactions.
- This is the standard economics of every transparent-gossip design (public-mempool MEV), not a defect introduced by #1224.

## Fix directions

Closing it requires batch contents to stay sealed until ordering:

1. Commit-reveal: gossip and certify a commitment, reveal transactions after ordering.  Adds a round trip and a data-availability problem for unrevealed batches.
2. Threshold encryption: encrypt batch contents to the committee, decrypt after commit.  Heavier cryptography, same availability question.
3. Accept and monitor: keep the residual documented (engine README threat-model table, added in #1224) and watch for poaching patterns in practice.

Filing so the residual is tracked rather than lost in PR review history.  Option 3 is the status quo after #1224 merges.

PR #1268, opened 2026-08-25
feat(engine): flag cross-producer transaction re-packing during execution (#1259)

MavenRain:
Implements the accept-and-monitor direction (option 3) of #1259. The residual itself stays open in that issue. (Cantina #5)

## Summary

Batch contents travel in plaintext gossip before certification, so a validator can extract another validator's transactions, pack them into its own batch, and race for the priority fees. Sealing batch contents through ordering (commit-reveal or threshold encryption) is the only full fix; until then this PR makes the pattern observable in production instead of visible only as an anonymous duplicate.

The engine now attributes every transaction hash to the producer of the first-ordered batch that carried it, over a bounded window of recent consensus outputs. A transaction that reappears in a batch from a different producer increments `tn_engine.cross_producer_repacked_txs_total` and warn-logs the duplicate with both producers. The first-ordered copy is the copy execution credits with the fees, so a producer that repeatedly wins duplicate races against transactions first gossiped by others is the poaching pattern operators watch for.

## What this adds

- [x] `crates/types/src/repack_monitor.rs` (new): `RepackMonitor`, a shared cheap-clone handle following the `GasAccumulator` pattern. Rolling window of the last 16 consensus outputs, hard-capped at 65,536 retained transaction hashes; first-seen attribution per hash; cross-producer duplicates reported to the caller and counted monotonically.
- [x] `crates/engine/src/payload_builder.rs`: `execute_consensus_output` takes the monitor and observes each batch in ordering position. Telemetry only . . . execution is unchanged on every path.
- [x] `crates/engine/src/metrics.rs`: new `cross_producer_repacked_txs_total` counter.
- [x] `crates/engine/src/lib.rs`: `ExecutorEngine` owns the monitor and clones it into each execution task, so the window spans the engine's lifetime (one epoch). A fresh engine at an epoch boundary starts a fresh window, which at worst misses a poach racing across the boundary.

## Threat model

- Observation is attacker-independent and cheap-first: one keccak per raw transaction, no decoding, no committee state, before any EVM work.
- Both bounds are derived from real parameters, not guessed: 16 outputs is twice the engine's `MAX_QUEUED_OUTPUTS` (8) backlog, and a poached copy races the victim's copy through certification, so the two copies land in the same output or a nearby one; 65,536 retained hashes cost a few megabytes.
- Saturation fails soft. A producer that floods tiny transactions evicts the oldest outputs and shrinks the effective window. That blinds telemetry only; the monitor never rejects a batch and never changes execution, so it cannot be griefed into a liveness or consensus effect. The flood itself is visible in batch metrics.
- A flagged transaction is an indicator, not proof: an honest sender that submits the same transaction to two validators' RPC endpoints produces the same signature. The monitor observes; operators judge the pattern.
- The skipped duplicate was already counted anonymously (`tn_reth.invalid_txs_skipped_total`). This adds the attribution that turns "a duplicate happened" into "whose batch carried it first".

## Testing

- Unit (`repack_monitor.rs`): cross-producer duplicate flagged with stable first attribution; same-producer duplicates ignored (re-gossip is not poaching); same-output cross-producer flagged; window eviction forgets first-seen and re-attributes; tx-cap eviction drops the oldest output whole.
- Integration (`crates/engine/tests/it/main.rs`): one output with two certified batches from two different producer addresses, the second re-packing a transaction from the first. Execution succeeds (the duplicate is skipped at transaction-level validation), both blocks build, and the monitor reports exactly one cross-producer duplicate.
- Metrics registration test extended for the new counter.
- Local static ladder (gateledger, tnquiet, -j 2): nightly `fmt --check`; `check` and `clippy --no-deps -- -D warnings` on `-p tn-types -p tn-engine -p tn-batch-builder --all-targets`; testledger'd unit and integration tests above.
- CI (nightly workspace clippy, workspace nextest) fires on push; the pinned `+1.94` attest runs on the second machine.

## Out of scope

Sealing batch contents (options 1 and 2 in #1259) is a protocol change with a data-availability question and stays tracked in the issue. PR #1224 documents the residual in the engine README threat-model table; this PR adds the watch that posture calls for.


sstanfield (commented):
How will this interact with a user submitting a transaction to more than one RPC/validator?  It seems like will will flag those as malicious when they are not and in that case the first executed batch might be the right one?  It might not be an edge case to worry to much about but worth a discussion.

MavenRain:
That's a good question, and the case is real: a sender that submits the same signed transaction to two validators' RPC endpoints puts byte-identical copies into two producers' batches.  At the single-event level the monitor cannot tell that apart from a poach.  The threat-model section calls this out ("an indicator, not proof"), but it deserves the discussion, so here is the full reasoning.

Three properties keep it from being a problem:

1. Nothing is flagged as malicious in-protocol. The monitor never rejects a batch, never penalizes a producer, and never changes execution. It increments a counter and emits a warn log. An honest dual-submission costs both producers, and the sender, nothing.

2. The first executed copy still wins, exactly as today. Attribution follows the first-ordered batch, which is the copy that executes and collects the fees. The later copy is skipped by the existing duplicate check, unchanged by this PR. So yes, the first executed batch is the right one, and this PR does not alter which copy that is.

3. The signal operators act on is the rate asymmetry, not any single event. Honest multi-RPC submission is symmetric: across many transactions, the first-ordered copy lands with whichever submitted-to validator proposes sooner, so the "wins" spread across producers in rough proportion to proposal luck. Poaching is asymmetric: one producer repeatedly wins cross-producer duplicates against transactions first gossiped by many different peers. Dual-submission raises the background level of the counter for everyone; poaching moves one producer's rate away from the pack.

If production traffic shows enough dual-submission noise to obscure that asymmetry, a follow-up can sharpen the metric (for example a per-producer-pair breakdown, or alerting only on a sustained one-sided rate) without touching the observation path.  I'd be happy to file that as a tracked follow-up if you think the noise floor warrants it.

sstanfield (commented):
I want to discuss this further, I don't love this approach.  I think it makes more sense to have dedicated "watchers" (think we have discussed those) to watch for malicious activity vs have every validator do this.  If it was something that could result in action that would make sense but it does not.  Maybe it is cheap enough to not matter but not sure that is the case (looks like a lot of hashing if nothing else) and may become a burden on busy nodes.  At any rate not sure this is work every node should just do if not requested.

MavenRain:
Thanks Steve.  I'll concede the core point: this need not be mandatory work for every validator.

Proposal: gate the monitor behind a node config flag, default off.  A node that does not opt in constructs no window, hashes nothing, and skips the check.  That makes any node an opt-in "watcher": operators who want the signal enable it, everyone else pays zero.  If we later add dedicated watcher processes, the same code moves there.

On cost: the work is one keccak256 per raw transaction plus a bounded map lookup, in the same loop that then signature-recovers and executes those transactions, so it is a small fraction of per-tx work the node already does.  I can also drop the extra hashing entirely by reusing the tx hashes the engine computes when it decodes the batch.  Memory is hard-capped at 16 outputs / 65,536 hashes.

On "could result in action":  I agree that it is telemetry only (a counter and a rate-bounded warn).  The intent is to gather evidence that #1259 poaching happens in the wild before designing any response.  Default-off opt-in still gets that evidence from operators who care.

Question: does a default-off flag on the node address your concern, or would you rather this live outside the validator entirely as a separate watcher service?  The flag is a small delta on this PR; a standalone watcher needs its own feed of ordered batches, which is a larger design.

sstanfield:
> Thanks Steve. I'll concede the core point: this need not be mandatory work for every validator.
> 
> Proposal: gate the monitor behind a node config flag, default off. A node that does not opt in constructs no window, hashes nothing, and skips the check. That makes any node an opt-in "watcher": operators who want the signal enable it, everyone else pays zero. If we later add dedicated watcher processes, the same code moves there.
> 
> On cost: the work is one keccak256 per raw transaction plus a bounded map lookup, in the same loop that then signature-recovers and executes those transactions, so it is a small fraction of per-tx work the node already does. I can also drop the extra hashing entirely by reusing the tx hashes the engine computes when it decodes the batch. Memory is hard-capped at 16 outputs / 65,536 hashes.
> 
> On "could result in action": I agree that it is telemetry only (a counter and a rate-bounded warn). The intent is to gather evidence that #1259 poaching happens in the wild before designing any response. Default-off opt-in still gets that evidence from operators who care.
> 
> Question: does a default-off flag on the node address your concern, or would you rather this live outside the validator entirely as a separate watcher service? The flag is a small delta on this PR; a standalone watcher needs its own feed of ordered batches, which is a larger design.

I think default off behind a cli flag is good for now.  I think further down the line we may wind up with a "watcher" node type, basically an observer that digs into everything for bad behavior (outside this PRs scope).  As far as intent of detecting if/when this is an issue this is fine.

MavenRain:
Implemented.  The monitor now defaults off behind `--enable-repack-monitor` (same shape as `--enable-state-export`).  Default-off is encoded in the type: `RepackMonitor::default()` holds no window and hashes nothing, so a node that does not opt in does no extra work per batch. `RepackMonitor::enabled()` is the only way to turn it on, and the node logs the monitor state at engine startup.  Tests cover both the disabled default and the enabled path through the engine.

sstanfield (approved):
Looks good, thanks.

PR #1270, opened 2026-08-25
fix(primary): key intra-round commit order on the leader digest behind a fork gate (#1260)

### LINKED ISSUE #1260: order_dag intra-round linearization is grindable: a proposer can bias its own certificate's execution position by grinding its header digest
## Summary

The committed sub-DAG is linearized by `order_dag` (`crates/consensus/primary/src/consensus/utils.rs`).  Within a single round, the relative order of certificates is fixed by `HeaderDigest`.  A header digest is a BLAKE3 hash over proposer-controlled fields (the `payload` `IndexMap` insertion order and `created_at`), so a Byzantine proposer can cheaply grind its own header digest to move its certificate toward the front or back of its round.  That order is preserved with no re-sort all the way into execution, so the grindable consensus order is the final transaction-execution order at certificate and batch granularity.

This is not a safety or liveness bug.  It is a transaction-ordering-fairness / MEV concern, and it is the un-mitigated form of a known limitation of DAG-BFT linearization.

Reported by shotes (Researcher) in the Cantina review;  this issue records the verified mechanism, the end-to-end reach into execution, and a suggested fix.

Commit: `648b6f81`.

## Mechanism

`order_dag` is a stack-based depth-first pre-order traversal:

- For each certificate `x`, it iterates `x.header().parents()` (`utils.rs:26`).  `parents()` returns `&BTreeSet<HeaderDigest>` (`crates/types/src/primary/header.rs:192`), so the iteration order is the ascending `HeaderDigest` order.
- The last step is `ordered.sort_by_key(|x| x.round())` (`utils.rs:55`).  Rust's `slice::sort_by_key` is a stable sort, so it preserves the DFS discovery order inside each round.

Net effect: within a round, the certificate order is the DFS discovery order, and that order is derived from the digest-ordered parent sets.  A refinement of the original report: it is not a literal single "sort within a round by digest," but the stable round-sort preserves a DFS order that is itself keyed on parent digests, so the conclusion (intra-round order is deterministic and digest-derived) holds.

## The grind surface

`HeaderDigest` is `blake3(bcs(HeaderRef))` (`header.rs:526-536`; `DefaultHashFunction = blake3::Hasher`, `DIGEST_LENGTH = 32`).  The preimage includes fields the proposer controls:

- **`payload` order.**  `payload` is an `IndexMap<BlockHash, WorkerId>` and is serialized in insertion order into the digest preimage.  `Header::validate` (`header.rs:123-169`) checks only `payload.len() <= MAX_HEADER_NUM_OF_BATCHES` (= 10, `crates/types/src/primary/mod.rs:71`) and that worker ids are valid.  It does not check payload order.  So the proposer can freely permute up to 10 entries, which gives up to `10! ≈ 3.6M` distinct legal digests with identical content.
- **`created_at`.**  Weakly constrained.  It must be monotonic against parents (`crates/consensus/primary/src/network/handler.rs:844-851`) and within `max_header_time_drift_tolerance` of the voter's clock (`handler.rs:872-898`;  default 1 second, `crates/config/src/network.rs`).  At 1-second granularity this yields only about 1 to 2 usable values on its own.

Cost per candidate is one BLAKE3 hash over a few hundred bytes, so millions of candidates per second on ordinary hardware.  No re-signing is needed: the only signature in the preimage, `seed_signature`, binds `(epoch, round, prior_epoch_record)` only (`crates/consensus/primary/src/proposer.rs:283-309`) and is reused unchanged across every candidate.  The proposer therefore has far more grinding entropy than the roughly `log2(committee_size)` bits needed to force a near-extremal (first or last) position in a round.

## It reaches execution with no re-sort

The order from `order_dag` is carried into block execution with no intervening re-sort by author, stake, gas price, or a re-hash:

1. `bullshark.rs:233` binds `sequence = utils::order_dag(&leader, state)`.
2. `bullshark.rs:253-260` moves `sequence` into `CommittedSubDag::new` unmodified.
3. `crates/types/src/primary/output.rs:440` builds `headers` with an order-preserving `into_iter().map().collect()`.  `headers()` returns that same-order slice (`output.rs:524-526`).
4. `crates/consensus/executor/src/subscriber.rs:450-526` builds `batch_digests` and `batches: Vec<CertifiedBatch>` by iterating `sub_dag.headers()` in order.  The one downstream `BTreeSet` (`batch_set`) is a dedup set for the peer fetch RPC only;  it does not set the final order.
5. `output.rs:194-201` `flatten_batches()` enumerates `cert_idx` over that vector in place.
6. `crates/engine/src/payload_builder.rs:189-239` consumes the flattened list sequentially and calls `execute_payload` once per `(cert_idx, batch_idx)`, so execution blocks extend the chain in exactly that order.

So the certificate whose batches execute first in a round, relative to other certificates in the same round, is decided by the grindable digest order.

## Why this matters, and the honest limits

The extractable effect is a same-round positioning advantage: a proposer can push its whole header (its set of up to 10 batches) ahead of or behind a rival's whole header.  This is the exact analog measured for Sui's Mysticeti in "Fair on the Surface: Transaction-Ordering Bias and MEV in Mysticeti DAG-based BFT Protocol" (arXiv 2607.13378), where the favored side won same-round ordering about 89% of the time on a 13-validator network with no attacker, rising above 94% with a strategic proposal delay.  Telcoin's ~1s drift tolerance leaves a similar delay-then-react window open.

Scope and severity, stated plainly:

- Not a safety or liveness bug.  Consensus agreement is unaffected;  any deterministic tie-break is valid for agreement.
- Requires the attacker to be a committee primary / proposer.
- Coarse granularity.  It reorders whole certificates (and their batches), not individual transactions across authors, so it does not by itself enable per-transaction sandwiching.
- Probabilistic, not targeted.  Same-round headers are normally produced concurrently, so a proposer generally cannot grind to beat one specific known rival digest on demand.  The realistic strategy is an always-available self-favoring bias toward an extremal position, not a targeted front-run of a named transaction.

We rate this low severity, consistent with how the adjacent fairness / MEV issues in this repo are triaged (for example #1259).  It is closely related to #1247 (a proposer grinding on-chain randomness) and #1222 / #1259 (proposer / batch-level fee fairness).

## Telcoin sits at the un-mitigated end

Deterministic intra-round tie-break is inherited from the Narwhal / Bullshark / Tusk DFS linearization, so this is a known DAG-BFT limitation rather than a telcoin-specific defect in isolation.  What is telcoin-specific is that both mitigations the reference implementation now ships are absent here:

- `order_dag` applies no per-commit, leader-digest-seeded shuffle to the intra-round order.
- The execution layer applies no gas-price-aware re-sort.

So the raw grindable consensus order is the final execution order end to end.  The in-repo comment on the round sort only calls it cosmetic ("prettier," `utils.rs:51-54`);  no comment or doc flags the ordering as grindable or unfair.

## Suggested fix

Adopt the fix Sui / Mysten shipped for the same class: replace the deterministic, ahead-of-time-predictable intra-round tie-break with an unpredictable per-commit key.  Concretely, order same-round certificates by `hash(committed_leader_digest, certificate_digest)` inside `order_dag` (or at the point the sub-DAG is linearized).  Because the committed leader's digest is not known when a proposer builds its own round-r header, seeding the tie-break on it removes the ability to grind a position in advance, while keeping the order fully deterministic for every honest node.

## Reproduction sketch

1. Build a valid round-r header, then enumerate permutations of its `payload` `IndexMap` (and the 1 to 2 legal `created_at` values), recomputing the BLAKE3 header digest for each.
2. Select the permutation whose digest lands the certificate at the desired extremal position under the `order_dag` DFS + stable round-sort, given the observed round-(r-1) parent set.
3. Confirm via the trace above that the chosen certificate's batches are built into an earlier (or later) execution block than a rival same-round certificate's batches.

MavenRain:
Closes #1260.

## Problem

`order_dag` linearizes a committed sub-DAG with a stable `sort_by_key(round)`, so the relative order of certificates inside a round is the DFS discovery order, and that order is derived from the digest-ordered parent sets (`parents()` is a `BTreeSet<HeaderDigest>`).  A header digest is a BLAKE3 hash over proposer-controlled fields: `Header::validate` does not constrain the `payload` `IndexMap` insertion order (up to 10! permutations of identical content) and only weakly constrains `created_at`.  A Byzantine proposer can therefore grind its own header digest (millions of candidates per second, no re-signing needed) to steer its certificate toward either end of its round, and that order reaches execution with no re-sort at any layer (#1260 traces the full path from `bullshark.rs` through `CommittedSubDag` and the executor subscriber into `payload_builder.rs`).

This is not a safety or liveness defect.  It is the un-mitigated transaction-ordering-fairness / MEV concern of DAG-BFT linearization: the attack needs no cooperation, is invisible on chain, and pays off at certificate and batch granularity.

## Fix

Adopt the same class of fix Sui / Mysten shipped for this: key the intra-round tie-break on the committed leader's digest.  Post-fork, `order_dag` sorts by `(round, blake3(leader_digest || certificate_digest))` via `sort_by_cached_key`.  No proposer other than the leader knows the committed leader's digest while building its own header (same-round headers are produced concurrently, and the leader that commits a round is only decided one round later), so a non-leader position can no longer be ground in advance.  The order stays a pure function of committed data, so every honest node derives the identical sequence, and the leader still sorts last (it is the only certificate at the highest round).

Honest residual, stated in the doc comments rather than papered over: the schedule-determined leader still controls its own digest, so on its own turn it keeps a grind over its sub-DAG's intra-round permutations.  That residual is accepted deliberately: it shrinks the attacker set from every proposer on every commit to the one leader on its own turn;  any tie-break that is a pure function of committed data keeps it (the leader's header is the last input to be fixed);  and removing it requires unpredictable post-commit randomness the protocol does not have at linearization time.

Because the ordering change alters execution order, it is consensus-breaking across versions and ships behind a new epoch fork gate following the exact `seed_signature` / `multi_workers` pattern:

- `LEADER_SEEDED_ORDERING_FORK_EPOCH` is a `u32::MAX` placeholder under `adiri` (dormant until a dedicated epoch-setting PR arms it, after the fleet runs a gate-capable build);  every other build is active from genesis (no non-adiri chain carries legacy-ordered commits, the same premise the existing gates document).
- The gate reads `leader.epoch()`, the epoch carried inside the committed value, never node-local committee state, so replay of a pre-fork commit reproduces the historical sequence.  A sub-DAG cannot mix epochs (`new_certificate` drops cross-epoch certificates and the DAG store is cleared per epoch), so one leader epoch governs one commit.
- One rollout difference from the wire-format forks is called out in the constant's docs: a straggler past an armed boundary does not fail to decode, it forks away loudly at its next commit, so the fleet must be fully upgraded before the epoch-setting PR lands.

Replay and downstream order-dependence were audited before the change: `order_dag` has exactly one caller (the live commit path in `bullshark.rs`);  every resync / pack-load / state-sync path consumes the stored `CommittedSubDag` rather than re-deriving an order;  `CommittedSubDag` digests are only ever compared stored-vs-stored;  reputation scoring, `state.update`, the executor subscriber, and pack batch records are order-insensitive or permutation-invariant within a round.

## Changes

- [x] `crates/types/src/forks.rs`: new `LEADER_SEEDED_ORDERING_FORK_EPOCH` (`u32::MAX` placeholder, `adiri` only), `leader_seeded_ordering_active(epoch)`, override-free `leader_seeded_ordering_build_fork_active`, and the `TN_LEADER_SEEDED_ORDERING_FORK_EPOCH` test-utils override, each mirroring the existing gates' docs and rollout contract;  plus the two standard gate tests (rollout-contract grid derived from the constant with a saturating pre-fork row, and override-inert-when-unset).
- [x] `crates/consensus/primary/src/consensus/utils.rs`: post-fork `order_dag` sorts by `(round, leader_seeded_key(leader_digest, cert_digest))`;  pre-fork keeps the legacy stable round-only sort byte for byte.  `leader_seeded_key` is `blake3(leader_digest || certificate_digest)` via `DefaultHashFunction`, with the residual-leader-grind trade-off documented on the function.
- [x] `crates/consensus/primary/src/consensus/tests/bullshark_tests.rs`: four behavioral tests over a rounds-1..=6 optimal DAG: seeded sort only permutes within rounds (membership, round monotonicity, leader last), is deterministic, depends on the leader digest (two distinct leaders disagree on some round's permutation, failure probability under 2e-7 per run), and does not reduce to ascending digest order.  On a build where the fork is dormant (`adiri` feature unification, or a high env override latched) the tests skip with a loud note instead of failing.
- [x] `crates/telcoin-network-cli/src/node.rs`: the new constant joins the startup fork-schedule log in both the `adiri` and from-genesis branches.

## Testing

Local static ladder (all under `gateledger`, scoped to the three changed crates, shared target, `-j 2`):

- `cargo +nightly fmt -p tn-types -p tn-primary -p telcoin-network-cli -- --check` green.
- `cargo check -q -p tn-types -p tn-primary -p telcoin-network-cli` green.
- `cargocho clippy -- -p tn-types -p tn-primary -p telcoin-network-cli --all-targets --no-deps -j 2 -- -D warnings` green.
- `cargo test -p tn-types leader_seeded_ordering` and `cargo test -p tn-primary bullshark` (full bullshark suite plus the four new tests) green.
- ctxcat-review over the staged diff: the surviving findings (leader residual grind must be documented, tests must skip rather than fail on dormant builds, gate-test grid underflow) are all addressed in this diff.
- Mutation-confirmed, in-diff mutants only: dropping the leader digest from the key kills `order_dag_seeded_sort_depends_on_the_leader`;  dropping the round from the key kills `order_dag_seeded_sort_preserves_rounds_and_membership`;  inverting the fork gate kills `order_dag_seeded_sort_depends_on_the_leader`;  forcing the non-adiri build arm to `false` kills the forks rollout-contract test;  making the key the raw certificate digest kills `order_dag_seeded_sort_is_not_digest_ascending`.  Each mutant was seen to fail and the tree byte-restored (worktree equals index).

Dynamic tier: GitHub CI fires on this push (nightly fmt/clippy, workspace nextest, adiri lane);  the pinned `cargo +1.94 test --no-run --workspace` attest runs on the second machine post-push.


sstanfield (dismissed):

crates/consensus/primary/src/consensus/utils.rs
```diff
@@ -52,10 +54,55 @@ pub(crate) fn order_dag(leader: &Certificate, state: &ConsensusState) -> Vec<Cer
...
+    // leader's digest. No other proposer knows that digest while it builds its own header, so
+    // for every validator except the leader a position can no longer be ground in advance. The
+    // schedule-determined leader still controls its own digest and so keeps a grind over its
+    // sub-DAG's intra-round permutations on its own turn; `leader_seeded_key` documents why
+    // that residual is accepted. The gate reads the epoch carried inside the leader
+    // itself, never node-local state, so replay of a pre-fork commit reproduces the historical
+    // sequence.
+    if forks::leader_seeded_ordering_active(leader.epoch()) {
+        let leader_digest = leader.digest();
+        ordered.sort_by_cached_key(|x| (x.round(), leader_seeded_key(&leader_digest, &x.digest())));
```
grantkee: Keying the tie-break on `leader.digest()` moves the grind rather than removing it, and it moves it to the strongest adversary at the worst moment. The leader's header is the *last* input fixed before linearization, so the leader can enumerate candidate digests **after** observing every certificate and batch in the sub-DAG it is about to commit — and post-fork that digest is the sole free variable, re-keying every round of the sub-DAG at once. Under the legacy sort the leader's own digest never entered the ordering of any lower round at all.

Working through the mechanics on this diff:

- Grinding is cheap and offline (not equivocation). `Header::validate` (`crates/types/src/primary/header.rs:123-170`) checks epoch, parent count, batch count, authority membership and worker-id range — and nothing else. It constrains neither the `payload` insertion order (which is preimage-relevant, `header.rs:271-289`, `:526-537`) nor `created_at` (bounded only by the parent maximum below and a drift tolerance above, `crates/consensus/primary/src/network/handler.rs:862-866`, `:891-906`). With `MAX_HEADER_NUM_OF_BATCHES = 10` that is up to 10! ≈ 3.6M valid payload permutations per header, plus free parent-subset choice.
- The legacy arm gave the leader **no** such control. Hand-tracing the DFS above (stack buffer, `Header::parents()` a `BTreeSet<HeaderDigest>`), round-`L-1` order is the leader's parent set *reversed* — a function of the **parents'** digests. `leader.digest()` is evaluated only inside the new `if` on line 70 and never on the legacy path. So the leader's control over lower-round order goes from none to effectively total: the seeded sort discards DFS order entirely. Cost is ~n candidate digests for a targeted slot, ~n! for an arbitrary permutation of one round.
- The position is load-bearing. `order_dag`'s output *is* the execution order — `crates/consensus/executor/src/subscriber.rs:457-461` walks `sub_dag.headers()` in sequence to build `batch_digests`, with no re-sort downstream.

The sibling PREVRANDAO fork (#1247) documents this exact attack and rejected exactly this shape of seed (`crates/types/src/forks.rs:285-294`): "a committing leader can enumerate otherwise-valid payload constructions **after the causal DAG is known** and publish the one whose `PREVRANDAO` favors it." It replaced the leader-influenced seed with the epoch seed chain.

**Two dead ends worth naming before the fix, because both look reasonable.** First: do *not* concatenate `leader_digest` alongside a non-grindable seed. A hash is grindable if the adversary controls any part of the preimage, so `blake3(leader_digest ‖ seed_sig ‖ cert_digest)` is *worse* than either alternative — the signature degenerates into a domain tag while the full late grind survives. Second: bare `seed_signature` has a residual. The schedule is round-robin over the committee (`crates/consensus/primary/src/consensus/leader_schedule.rs:301`, modulo the reputation swap table at `:309-311`) and the seed message is fixed at epoch start, so a future leader can pre-sign its round and hand the signature to an accomplice, who then grinds its own certificate digest for a slot in its own round. Strictly dominated by the current late grind (no interior third-party reordering, no visibility into what it jumps) but not zero.

**The epoch seed chain closes both, and this commit path already computes it three statements later.** `CommittedSubDag::new` derives `randomness = seed_chain.fold(leader.round(), sig)` (`crates/types/src/primary/output.rs:483`), and `order_dag` runs on the same `&ConsensusState` immediately before it: `bullshark.rs:233` calls `order_dag`, `:259` passes `state.seed_chain()` into `CommittedSubDag::new`, and `:276` only advances it afterwards. So `state.seed_chain()` inside `order_dag` holds *exactly* the value this commit is about to fold — same value, one statement earlier, no new input and no new fork surface. `crates/types/src/primary/seed_chain.rs:1-27` states the properties: every fold input is digest-pinned or derived state (canonicality), and "a candidate seed cannot be evaluated until the immediately preceding commit is published… **pre-signing future rounds offline is useless without the previous value**" — precisely the collusion gap the bare signature leaves open. The only accepted residual is propose-or-withhold (last-actor bias), a one-shot abort rather than a selection over millions of candidates.

```rust
// utils.rs — also folds in the domain separator from the `leader_seeded_key` comment below
const INTRA_ROUND_ORDER_DOMAIN: &[u8] = b"TN_INTRA_ROUND_ORDER_V1";

if forks::leader_seeded_ordering_active(leader.epoch()) {
    match leader.header().seed_signature() {
        // The same value `CommittedSubDag::new` folds into `randomness` for this commit:
        // non-grindable (the BLS signature is unique per `(author, epoch, round)`) and not
        // evaluable by anyone until the preceding commit is published.
        Some(sig) => {
            let seed = state.seed_chain().fold(leader.round(), sig).into_inner();
            ordered.sort_by_cached_key(|x| (x.round(), intra_round_key(&seed, &x.digest())));
        }
        // Unreachable while the gate conjoins `seed_signature_active`. Degrade to the legacy
        // rule rather than panic: a mis-set schedule then holds the whole fleet on the pre-fork
        // order uniformly instead of halting it.
        None => ordered.sort_by_key(|x| x.round()),
    }
} else {
    ordered.sort_by_key(|x| x.round());
}
```

That `None` arm wants the same fail-closed conjunct `prevrandao_seed_active` already uses (`forks.rs:337-339`) — rename this PR's gate body to a private `leader_seeded_ordering_fork_point_active` and wrap it:

```rust
pub fn leader_seeded_ordering_active(epoch: Epoch) -> bool {
    seed_signature_active(epoch) && leader_seeded_ordering_fork_point_active(epoch)
}

#[cfg(feature = "adiri")]
const _: () = assert!(LEADER_SEEDED_ORDERING_FORK_EPOCH >= SEED_SIGNATURE_FORK_EPOCH);
```

Two doc claims should go with the seed, since they document an accepted residual this construction removes and are false as written. `utils.rs:90-92` says "any seed that is a pure function of committed data keeps it" and "removing it needs unpredictable post-commit randomness, which the protocol does not have at linearization time" — both `seed_signature` and the seed chain are in-scope counterexamples, and the seed chain *is* available at linearization time. And `forks.rs:530` says "no proposer other than the leader knows that digest", which is an honest-leader assumption stated as a fact about the adversary.

**In fairness on timing:** this commit (`85549f74`) is dated 2026-08-25; #1247 landed in `main` as `1e8ee97c` on 2026-08-28 and merged here on 08-29. The PREVRANDAO precedent was in flight, not yet in `main`, when this seed was chosen. `seed_signature` and the seed chain did already exist.

The call between the seed-chain fold (recommended — closes the pre-signing gap) and bare `seed_signature` (simpler, leaves the two-party collusion residual) is yours; both are strictly better than the leader digest.

crates/types/src/forks.rs
```diff
@@ -517,6 +517,119 @@ pub fn multi_workers_fork_epoch_override() -> Option<Epoch> {
...
+/// below the constant), then land the epoch-setting PR fleet-wide before the fork epoch
+/// begins. A straggler still on an old build past the boundary orders the same certificates
+/// differently, executes them in a different sequence, and forks away from the upgraded fleet
+/// at its next commit. That divergence is loud (its executed state stops matching the fleet's)
+/// but it is a fork, not a decode error, so the fleet must be fully upgraded before the epoch
+/// is armed.
+///
+/// Non-adiri builds (mainnet) have no dormant period: the seeded order is active from genesis
+/// and this constant does not exist there.
+pub const LEADER_SEEDED_ORDERING_FORK_EPOCH: Epoch = u32::MAX;
```
grantkee, line 690: The gate's shape is correct field-for-field against its three siblings — the new boundary test even uses `saturating_sub(1)` where the siblings would overflow-panic if armed at 0 — but there's no arming constraint recorded against `ADIRI_DUP_BATCH_EPOCH`, and this fork interacts with it.

Duplicate-batch attribution **is** intra-round-position-sensitive for adiri epochs at or below `ADIRI_DUP_BATCH_EPOCH` (160, `forks.rs:71`). A batch referenced by two headers is credited to whichever header comes **first in sequence** and dropped from the second: `crates/consensus/executor/src/subscriber.rs:512` skips the `cert_batches.push` for `leader_epoch() <= ADIRI_DUP_BATCH_EPOCH`, mirrored in the pack reader at `crates/storage/src/consensus_pack.rs:1809` and `:1948`. Since this PR's tie-break permutes exactly that sequence within a round, arming the fork at or below epoch 160 would change replayed attribution on adiri — a resync divergence, not just a reordering.

Both siblings record their arming constraint: PREVRANDAO by a `const` assert (`forks.rs:316-323`), MULTI_WORKERS by a documented paragraph (`:426`). This one has neither, and there's already precedent for this exact assert against this exact constant at `crates/storage/src/consensus_pack.rs:2401`.

```rust
#[cfg(feature = "adiri")]
const _: () = assert!(LEADER_SEEDED_ORDERING_FORK_EPOCH > ADIRI_DUP_BATCH_EPOCH);
```

Worth an "Arming constraint" paragraph on the constant alongside it, since the assert alone doesn't say *why* to a future reader setting the epoch. Practically the placeholder makes this unreachable today; the point is that the epoch-setting PR is where it would bite, and that PR won't be looking at this interaction.

crates/types/src/forks.rs
```diff
@@ -517,6 +517,119 @@ pub fn multi_workers_fork_epoch_override() -> Option<Epoch> {
...
+///
+/// An environment variable rather than a process-global setter because e2e tests drive real
+/// node processes spawned via `TN_BIN_PATH`, which share no memory with the harness: a static
+/// would silently reach only the in-process tests.
+///
+/// Compiled out entirely without `test-utils`, so a production binary keeps the compile-time
+/// constant and cannot be repointed at runtime by its environment. An unparseable value is
+/// ignored rather than defaulted, leaving the build's own fork point in force.
+#[cfg(feature = "test-utils")]
+pub fn leader_seeded_ordering_fork_epoch_override() -> Option<Epoch> {
```
grantkee, line 792: The paragraph here justifies the environment-variable form by an e2e need, but no e2e test actually consumes it — `TN_LEADER_SEEDED_ORDERING_FORK_EPOCH` is never set anywhere. Grepping the whole tree gives six hits: this doc, the reader itself, the gate doc at `:568`, and two test assertion messages. Zero `set_var`, zero `command.env`, zero Makefile, zero CI.

The wired siblings show what the difference looks like: `TN_SEED_SIGNATURE_FORK_EPOCH` and `TN_MULTI_WORKERS_FORK_EPOCH` are named as constants in `crates/e2e-tests/tests/it/epochs.rs:45,49`, defaulted in `Makefile:152,162`, exported on every e2e target (`Makefile:166,172,176,252,253`), and forwarded to spawned children by the loop at `crates/e2e-tests/src/lib.rs:56`. The new one is in none of those.

Two things surfaced while checking this. `TN_PREVRANDAO_FORK_EPOCH` is identically dead (pre-existing on `main`, not this PR's doing) — so it's two of four overrides unwired, not one. More interesting, there's an **inverted asymmetry** here: the e2e binary is non-adiri, so `leader_seeded_ordering_build_fork_active` returns `true` and every spawned e2e node runs the seeded order *active from genesis*, while the two wired siblings default to a dormant `u32::MAX`. No e2e lane can currently exercise the legacy ordering path at all — the reverse of the usual dormant-by-default posture.

That also makes two doc sentences stale as of this PR: `crates/e2e-tests/src/lib.rs:39-48` ("Pins every fork-epoch override on the child… dormant for both forks") and `crates/e2e-tests/tests/it/epochs.rs:536` ("Both forks are pinned, not just the one a given test is about"). Both now describe two of three.

Adding the var to the `for var in [...]` loop at `crates/e2e-tests/src/lib.rs:56`, the `Makefile` default and the e2e targets would make the override real, and would give the cross-fork e2e test somewhere to stand.

crates/telcoin-network-cli/src/node.rs
```diff
@@ -127,12 +127,14 @@ impl<Ext: clap::Args + fmt::Debug> NodeCommand<Ext> {
...
             multi_workers_fork_epoch = tn_types::forks::MULTI_WORKERS_FORK_EPOCH,
+            leader_seeded_ordering_fork_epoch = tn_types::forks::LEADER_SEEDED_ORDERING_FORK_EPOCH,
             "fork schedule (adiri)"
         );
         #[cfg(not(feature = "adiri"))]
         info!(
             target: "cli",
-            "fork schedule: seed_signature and multi_workers active from genesis"
+            "fork schedule: seed_signature, multi_workers, and leader_seeded_ordering active \
+             from genesis"
```
grantkee: This PR adds its own entry to the "full fork schedule" log but the schedule logged is still incomplete: `PREVRANDAO_FORK_EPOCH` appears in no log statement repo-wide (grepping the constant gives hits only inside `crates/types/src/forks.rs`, and "prevrandao" appears nowhere in this crate at all).

That matters because the new constant's own doc leans on this log as the sole detection mechanism — `forks.rs:540-543`: "The full fork schedule is logged at startup so operators can diff it across the fleet; a compile-time constant that differs between binaries has no other in-protocol detection." I checked that premise independently and it holds: there's no libp2p `identify`, protocol IDs carry hardcoded literal versions, `ConsensusHeader.extra` is hashed as `B256::default()` unconditionally, and the only mismatch detector (`crates/state-sync/src/lib.rs:414-421`) fires *after* divergence. So a fork missing from this log has no operator-visible signal until nodes have already forked.

One framing correction on my own claim: there is no non-adiri prevrandao *constant* — the non-adiri path is a hardcoded `true` (`forks.rs:373-377`) — so the non-adiri arm can't log a value for it. It should still name prevrandao in the sentence, since it's genesis-active there.

```rust
#[cfg(feature = "adiri")]
info!(
    target: "cli",
    // ...
    leader_seeded_ordering_fork_epoch = tn_types::forks::LEADER_SEEDED_ORDERING_FORK_EPOCH,
    prevrandao_fork_epoch = tn_types::forks::PREVRANDAO_FORK_EPOCH,
    "fork schedule (adiri)"
);
info!(
    target: "cli",
    "fork schedule: seed_signature, multi_workers, prevrandao, and leader_seeded_ordering \
     active from genesis"
);
```

Also noted while looking: `PREVRANDAO_FORK_EPOCH` is the only rollout constant whose doc omits the "logged at startup" sentence. #1247 skipped the sentence and the log line together, which is presumably how it slipped. Not this PR's bug, but this PR is the one editing the log.

crates/consensus/primary/src/consensus/tests/bullshark_tests.rs
```diff
@@ -1529,3 +1529,191 @@ async fn not_enough_support_and_missing_leaders_and_gc() {
...
+        acc.entry(certificate.round()).or_default().push(certificate.digest());
+        acc
+    })
+}
+
+/// Post-fork `order_dag` invariants (#1260): rounds never decrease, the leader is the final
+/// certificate, and the leader-seeded tie-break only permutes certificates within their round,
+/// so per-round membership is exactly what the legacy round-only sort would emit.
+#[tokio::test]
+async fn order_dag_seeded_sort_preserves_rounds_and_membership() {
```
grantkee: The four new tests here assert structural and statistical properties — round monotonicity, per-round membership, leader-sorts-last, distribution — but nothing anywhere *asserts* an intra-round ordering property on either arm, and nothing exercises the gate flipping across an epoch boundary.

**On the legacy arm, first correcting something I initially got wrong:** the legacy branch *is* executed in CI. `lane_test_adiri` runs `--features tn-node/adiri,...` (`etc/ci-lanes.sh:113`), which forwards `tn-types/adiri`, and Cargo unifies that onto `tn-primary` as a plain dependency — so seven `tn-node` integration tests reach `bullshark.rs:233` through `spawn_consensus` and hit the legacy arm on every adiri run (`test_catchup_accumulator`, `..._with_empty_outputs`, `..._partial_execution`, `test_sync_then_catchup_recovers_two_worker_accumulator`, `test_entry_reads_static_fee_at_boundary`, `epoch_block_height_is_closing_block_plus_one`, `test_entry_read_matches_close_written_eip1559_fee`). The same seven exercise the seeded arm under `lane_test_default`. What is never built is `tn-primary`'s *own* test binaries under adiri.

**What survives:** none of those seven asserts anything about order — they assert gas accumulators, base fees and block heights. So both arms execute, neither is pinned. And the boundary flip is untested: `forks.rs:862-906` sweeps the epoch boundary but asserts the gate predicate only, never `order_dag`. That flip is what changes the `ConsensusHeader` digest, so it's the case most worth having a test for.

The blocker for testing both arms in one process is that the branch is only reachable through the gate, and the gate can't be flipped at runtime — nextest has no per-test env override and the `OnceLock` latch rules out setting the variable from inside a test. Making the branch injectable solves that and also removes the need for the `seeded_branch_active()` guard entirely:

```rust
pub(crate) fn order_dag(leader: &Certificate, state: &ConsensusState) -> Vec<Certificate> {
    order_dag_inner(leader, state, forks::leader_seeded_ordering_active(leader.epoch()))
}

pub(crate) fn order_dag_inner(
    leader: &Certificate,
    state: &ConsensusState,
    leader_seeded: bool,
) -> Vec<Certificate> {
    // existing body, branching on `leader_seeded` instead of calling the gate
}
```

With that in place three tests become straightforward: a legacy-arm pin against the DFS discovery order for a fixed DAG, a two-arms-disagree test over the same state, and a cross-fork e2e mirroring `test_epoch_sync_across_multi_workers_fork` (`crates/e2e-tests/tests/it/epochs.rs:45-58`) once `TN_LEADER_SEEDED_ORDERING_FORK_EPOCH` is actually wired.

crates/consensus/primary/src/consensus/tests/bullshark_tests.rs
```diff
@@ -1529,3 +1529,191 @@ async fn not_enough_support_and_missing_leaders_and_gc() {
...
+fn seeded_branch_active() -> bool {
+    let active = forks::leader_seeded_ordering_active(0);
+    if !active {
+        eprintln!(
+            "skipping leader-seeded ordering assertions: the fork is dormant in this build \
+             (adiri feature unified, or TN_LEADER_SEEDED_ORDERING_FORK_EPOCH latched high)"
+        );
+    }
+    active
+}
```
grantkee: This guard makes the four tests below silently self-neutering rather than failing loudly, which is the opposite of the convention this same PR applies in `forks.rs`. All four bodies sit entirely inside `if seeded_branch_active() { .. }` (`:1601-1635`, `:1642-1652`, `:1666-1690`, `:1703-1718`), so on a build where the fork is dormant you get four tests that pass having asserted nothing.

`forks.rs:912-919`, added in this PR, hits the identical situation — a `OnceLock`-latched override that a given process may not be able to observe — and handles it the other way, with the reasoning spelled out: "Fail loudly rather than assert a property this process cannot hold; a silent skip here would read as a pass."

**Impact is nil today, though**, which is why I'd call this style rather than coverage: `seeded_branch_active()` cannot return `false` in any CI lane or in a plain `cargo test -p tn-primary`. `tn-primary` declares no `adiri` feature, so even `--all-features` can't enable it; no crate defaults it on; the resolver is v2; and `lane_test_adiri` doesn't build `tn-primary`'s test targets. So it's a latent trap that arms itself the day someone adds an `adiri` feature to `tn-primary` or runs a lane with the env var latched high — at which point four tests go green while testing nothing.

The `order_dag_inner` refactor removes the guard entirely and is the better fix. Failing that, swapping the `if` for an `assert!` carrying the sibling's "run it in its own process" message keeps the tests honest.

Two incidentals in the same block while I'm here:

- `two_round_6_leaders` (`:1575-1584`) selects its two certificates via `HashMap::values()`, and `Dag`'s inner map is `HashMap<AuthorityIdentifier, _>` (`state.rs:29`) with the default `RandomState` — so which pair of leaders a run picks varies per process, and a failure is not reproducible.
- All four tests are `#[tokio::test]` with no `.await` in the body; plain `#[test]` would do.

crates/consensus/primary/src/consensus/utils.rs
```diff
@@ -52,10 +54,55 @@ pub(crate) fn order_dag(leader: &Certificate, state: &ConsensusState) -> Vec<Cer
...
+/// a pure function of committed data, so every honest node derives the identical order. The
+/// leader still sorts last: it is the only certificate at the highest round, so no tie-break
+/// applies to it.
+fn leader_seeded_key(
+    leader_digest: &HeaderDigest,
+    certificate_digest: &HeaderDigest,
+) -> [u8; DIGEST_LENGTH] {
+    let mut hasher = DefaultHashFunction::new();
+    hasher.update(leader_digest.as_ref());
+    hasher.update(certificate_digest.as_ref());
```
grantkee, line 133: `blake3(leader_digest ‖ certificate_digest)` with no domain tag. Nothing rests on it today — both operands are fixed 32-byte digests, so there's no length ambiguity within this use — but the 64-byte preimage is a **proper prefix of `CommittedSubDag::digest`'s own preimage** for a two-header sub-DAG (`crates/types/src/primary/output.rs:602-613`), which is close enough to be worth not relying on.

It's also the only derivation in the repo without one: `Intent::consensus(IntentScope::EpochCloseSeed)` (`epoch.rs:277-281`), `ROOT_DOMAIN` / `FOLD_DOMAIN` (`seed_chain.rs:34-37`), and `keccak256(domain ‖ …)` in `prev_randao` all tag their inputs.

Cheapest as the first hasher update, folded into whatever the seed ends up being:

```rust
const INTRA_ROUND_ORDER_DOMAIN: &[u8] = b"TN_INTRA_ROUND_ORDER_V1";

// ...
let mut hasher = DefaultHashFunction::new();
hasher.update(INTRA_ROUND_ORDER_DOMAIN);
hasher.update(seed.as_ref());
hasher.update(certificate_digest.as_ref());
hasher.finalize().into()
```

Since the key is consensus-critical, a byte-layout pin test over a fixed input pair is worth adding next to it — that's the thing that would catch an accidental operand swap or a dropped update in a later refactor.

grantkee (commented):
This is well done - there's one subtle gotcha I flagged that still allows a validator to grind. I think we should use the seed-chain fold instead of leader digest - the rest are low/test focused

grantkee (changes requested):

MavenRain:
Thanks for the review.  Every item landed in 73ae64cc, which is already on this branch;  the only later change is the merge of main (#1294). Per comment:

1. [utils.rs, leader-digest grind](https://github.com/Telcoin-Association/telcoin-network/pull/1270#discussion_r3919735176): the tie-break now keys on the epoch seed chain fold, `state.seed_chain().fold(leader.round(), sig)`, the same value `CommittedSubDag::new` folds into `randomness`. A leader header without a seed signature falls back to the legacy round-only sort. `leader_seeded_ordering_active` conjoins `seed_signature_active` through a private fork-point fn, with a compile-time assert `LEADER_SEEDED_ORDERING_FORK_EPOCH >= SEED_SIGNATURE_FORK_EPOCH` under adiri. The three stale doc sentences are gone.
2. [forks.rs, duplicate-batch attribution](https://github.com/Telcoin-Association/telcoin-network/pull/1270#discussion_r3919735185): added the compile-time assert `LEADER_SEEDED_ORDERING_FORK_EPOCH > ADIRI_DUP_BATCH_EPOCH` under adiri and an "Arming constraints" section on the constant's doc.
3. [forks.rs, unused env override](https://github.com/Telcoin-Association/telcoin-network/pull/1270#discussion_r3919735189): `TN_LEADER_SEEDED_ORDERING_FORK_EPOCH` is now wired: the child-spawn loop in `crates/e2e-tests/src/lib.rs`, a Makefile default plus an export on every e2e target next to its two siblings, the constant in `epochs.rs`, and a cross-fork test `test_epoch_sync_across_leader_seeded_ordering_fork` mirroring the multi-workers one. The two stale sentences now describe all three forks.
4. [node.rs, incomplete fork-schedule log](https://github.com/Telcoin-Association/telcoin-network/pull/1270#discussion_r3919735195): the adiri log carries `prevrandao_fork_epoch` and `leader_seeded_ordering_fork_epoch`, and the non-adiri sentence names both. The `PREVRANDAO_FORK_EPOCH` doc gained the "logged at startup" sentence.
5. [bullshark_tests.rs, structural-only tests](https://github.com/Telcoin-Association/telcoin-network/pull/1270#discussion_r3919735202): added the `order_dag_inner(leader, state, leader_seeded)` seam with `order_dag` delegating to it. New tests pin the legacy arm against an independent DFS-discovery replay of a fixed DAG and show the two arms keep membership but disagree on order over the same state.
6. [bullshark_tests.rs, silent-skip guard](https://github.com/Telcoin-Association/telcoin-network/pull/1270#discussion_r3919735208): `seeded_branch_active` is removed. The fixture stamps deterministic per-author seed signatures, `two_round_6_leaders` selects leaders through a `BTreeMap`, and the tests are plain `#[test]`.
7. [utils.rs, no domain tag](https://github.com/Telcoin-Association/telcoin-network/pull/1270#discussion_r3919735214): `intra_round_key` hashes `TN_INTRA_ROUND_ORDER_V1 || seed || certificate_digest`, and `intra_round_key_byte_layout_is_pinned` pins the 32-byte output for a fixed input pair.

This should be ready for another look.

grantkee (approved):
LGTM!

PR #1273, opened 2026-08-26
fix: source worker pool base fee from the per-worker container

### LINKED ISSUE #1262: Worker pool pending base fee reverts to the previous epoch's fee at the epoch boundary (residual of #791)
Title: Worker pool pending base fee reverts to the previous epoch's fee at the epoch boundary (residual of #791)

## Problem

At epoch entry, `set_worker_base_fee` installs the recalculated per-worker base fee into the worker transaction pool (`crates/node/src/engine/inner.rs:187`, called from `crates/node/src/manager/node/start_epoch.rs:482`). This is the fix for #791.

Two later paths overwrite that value with the base fee of the canonical tip header:

1. The pool maintenance task: `crates/tn-reth/src/txn_pool.rs:240` sets `pending_block_base_fee` from `tip.base_fee_per_gas`.
2. The batch builder's post-batch update: `crates/batch-builder/src/lib.rs:320-323` sets it from `last_canonical_update.base_fee_per_gas`.

At the boundary, the tip is the previous epoch's closing block. Its header carries the old epoch's fee: the block is priced from the pre-close EVM env (`crates/tn-reth/src/evm/block.rs:1519`), and `record_next_epoch_base_fees` writes the new fee only to `WorkerConfigs` storage (`block.rs:483-553`), not to the header.

So path 2 reverts the pool to the old fee as soon as the first batch of the new epoch seals, and path 1 does the same when the closing-block notification is applied after the entry setter (see #1236 for how canonical updates can lag). The stale value persists until the first new-epoch block executes. A consensus output with zero batches that does not close an epoch produces no block (`crates/engine/src/payload_builder.rs:107-121`), so the window can span many rounds.

## Impact

The pool partitions pending and queued transactions by `pending_block_base_fee`. During the window:

- Fee rose at the boundary: transactions with `max_fee_per_gas` below the new fee remain in the pending sub-pool. `build_batch` does not filter by fee (`crates/batch-builder/src/batch.rs:56`, selection loop at 72-128; the epoch fee only stamps the batch header at lines 52 and 132). The batch validator checks only the batch-level declared fee, never a per-transaction fee (`crates/batch-validator/src/validator.rs:40-83`, 188-196). Underpriced transactions therefore land in accepted batches and waste batch space.
- Fee fell at the boundary: transactions priced between the new and old fee stay demoted and are not proposed. If they are the only waiting transactions, the builder does not build (the build loop gates on a non-empty pending set), no block executes, and the stale value does not refresh until some transaction pays the old fee.

The per-boundary fee delta is bounded (12.5% per epoch step, `crates/types/src/gas_accumulator.rs`). Batch headers always carry the correct epoch fee and validators enforce equality, so consensus is unaffected. Severity: low. This is a bounded, mostly self-healing liveness and fairness defect.

## Suggested fix

Derive `pending_block_base_fee` in both update paths from the gas accumulator's per-worker fee for the current epoch instead of the tip header. Alternatively, clamp the canonical-update fee to the accumulator value when the tip belongs to the previous epoch. Also confirm that `set_block_info` re-partitions transactions already in the pool; if it does not, the #791 entry setter never re-sorts them.

## Origin

Cantina collaborative review: Haxatron's note on batch-builder gas tracking and the follow-up on `txn_pool.rs` (the epoch-boundary case), with the batch-builder instance tracked as review finding 17. This issue records the residual mechanism that survives the #791 fix. Line references are against main `648b6f81`.

MavenRain:
Closes #1262.  (Cantina #25)

## Problem

At an epoch boundary the gas accumulator recalculates each worker's base fee and `set_worker_base_fee` installs it in the worker's transaction pool.  The pool does not keep it.  Two paths overwrite the pool's `pending_block_base_fee` with the fee from the canonical tip header:

1. The pool maintenance task (`process_canon_state_update`) on every canonical notification.
2. The batch builder's post-batch pool update, which reads its stored `last_canonical_update`.

The closing block of an epoch carries the OLD epoch's base fee in its header (the new fee goes to `WorkerConfigs` storage, not to a header).  So the first canonical update after a boundary reverts the pool to the previous epoch's fee, and it stays wrong until an execution block from the new epoch becomes canonical.  Consensus output with zero batches for a non-closing round produces no block, so the window can span many rounds.

No attacker is required;  the drift happens on every boundary where the fee moves.  The divergence is bounded (one EIP-1559 step, at most 12.5%, per boundary) and consensus is unaffected.  When the fee rises, the pool marks underpriced transactions pending and the builder puts them in batches;  peers ACCEPT those batches because the batch validator checks only the batch-level declared fee, never per-transaction fees, so they waste batch space and the transactions revert at execution.  When the fee falls, the pool over-prices admission and can park correctly priced transactions, which wedges batch building until the pool catches up.

## Root cause

Residual of #791.  That fix seeded the pool's fee at worker initialization and at the boundary setter, but left both canonical-update paths sourcing the fee from tip headers.  TN headers are not a valid fee source across an epoch boundary;  the accumulator's per-worker `BaseFeeContainer` is the single source of truth.

## Fix

Thread the shared per-worker `BaseFeeContainer` into `WorkerTxPool` and make `update_canonical_state` read it directly.  The `pending_block_base_fee` parameter is gone, so no caller can inject a header fee:  both overwrite paths flow through this one choke point.  `run_epoch` writes the boundary fee into the same container before the new epoch starts, so the pool always reprices with the current epoch's fee.

`set_worker_base_fee` stays:  it is the immediate re-partition trigger at the boundary.  Reth's `set_block_info` (v1.11.3) runs the same `update_basefee` machinery as `on_canonical_state_change` and physically moves transactions between the pending, basefee, and queued subpools, which closes the open verification item on the issue:  the #791 entry setter does re-sort the pool.

## Changes

- [x] `crates/tn-reth/src/txn_pool.rs`: `WorkerTxPool` gains the `BaseFeeContainer` field;  `update_canonical_state` drops its `pending_block_base_fee` parameter and reads the container;  `process_canon_state_update` no longer computes a fee from the tip header;  `get_pending_base_fee` (hardcoded `MIN_PROTOCOL_BASE_FEE`, TODO issue 114) is removed from the `TxPool` trait and both impls;  module docs updated.
- [x] `crates/tn-reth/src/env/mod.rs`: `init_txn_pool` takes the container.
- [x] `crates/batch-builder/src/lib.rs`: the post-batch update no longer computes a fee from the stale `last_canonical_update`.
- [x] `crates/node/src/engine/inner.rs` + `mod.rs`: `initialize_worker_components` takes `BaseFeeContainer` instead of `u64`;  seeds `pending_basefee` from it;  corrects the `set_worker_base_fee` doc claim (underpriced transactions land in ACCEPTED batches and waste space;  peers do not reject them).
- [x] `crates/node/src/manager/node/start_epoch.rs`: passes the live container to worker initialization;  keeps the u64 snapshot for the batch validator and the epoch setter (the fee is constant within an epoch).
- [x] Tests: call sites adapted;  two new regression tests (below).

## Testing

- `test_canonical_update_cannot_clobber_epoch_base_fee` (tn-reth): pool built with `BaseFeeContainer::new(MIN_PROTOCOL_BASE_FEE + 1234)`, then `update_canonical_state` with the genesis sealed block, whose header fee differs by construction.  The pool's `pending_basefee` must stay at the epoch fee.
- `test_post_batch_update_installs_epoch_base_fee` (batch-builder): drives one full mined-batch cycle (build, ack, post-batch pool update, `wait_until` the mined transactions are pruned) and asserts the pool's fee is the epoch fee, not the genesis header's.
- Both tests confirmed by mutation:  reintroducing the tip-header fee source at the choke point makes both fail;  restoring the fix turns them green again.
- `cargo +1.94 check -p tn-reth`;  `cargo +1.94 clippy -p tn-reth -p tn-batch-builder -p tn-batch-validator -p tn-node --all-targets --no-deps`;  `cargo +1.94 test -p tn-node test_worker_pool_base_fee_sourced_from_accumulator` (the #791 regression test, adapted);  `cargo +nightly fmt --all -- --check`.


sstanfield (approved):

grantkee (approved):
LGTM

PR #1274, opened 2026-08-26
fix(engine): log skipped duplicate transactions at debug with the tx hash

### LINKED ISSUE #1263: engine: skipped-duplicate transactions are logged at warn! with a full {:#?} dump of the recovered transaction
## Summary

The block builder logs every skipped `InvalidTx` at `warn!` level and pretty-prints the full recovered transaction, including its calldata bytes. This path is expected traffic in normal multi-worker operation, so the default operator log level fills with large debug dumps of transactions that are not a problem.

## Detail

`crates/tn-reth/src/env/execution.rs` (execution loop of the block builder):

```rust
Err(BlockExecutionError::Validation(BlockValidationError::InvalidTx {
    error,
    ..
})) => {
    // allow transaction errors (ie - duplicates)
    //
    // it's possible that another worker's batch included this transaction
    warn!(target: "engine", %error,  "skipping invalid transaction: {:#?}", recovered);
    // expected in normal operation - see the field docs, this is not an alert
    RETH_METRICS.invalid_txs_skipped_total.increment(1);
    continue;
}
```

Three parts of the same block disagree with the log level:

1. The code comment says the case is "expected in normal operation" and "not an alert".
2. The metric docs (`crates/tn-reth/src/metrics.rs`) describe `tn_reth.invalid_txs_skipped_total` as "expected - duplicates/invalid transactions across" workers.
3. The log level is `warn!`, and the default filter is `info,evm=debug` (`crates/telcoin-network-cli/src/cli.rs`), so every occurrence reaches operator logs.

`{:#?}` on `recovered` prints the full `Recovered<TransactionSigned>`, calldata included. A transaction with large calldata produces a multi-kilobyte log entry each time a duplicate of it is skipped. With multiple workers, duplicates across batches are routine, so this fires under normal load. A sender can also inflate it on purpose: one large-calldata transaction that lands in more than one worker's batch is paid for once and dumped at `warn!` on every skip.

Disk exhaustion is bounded because the node uses reth's `LogArgs`, which has file-log rotation. The remaining cost is signal quality: operators who alert on `warn` get a channel flooded with expected traffic, and real warnings are harder to see.

The `error!` a few lines above (signer-recovery failure with `?tx_bytes`) is a different case: that path only fires when a certified batch carries undecodable bytes, which is Byzantine behavior and intentionally alertable. This issue is only about the expected-duplicate path.

## Suggested fix

Downgrade the log to `debug!` (or `trace!`) and log the transaction hash instead of the full debug representation, for example:

```rust
debug!(target: "engine", %error, tx_hash = ?recovered.hash(), "skipping invalid transaction");
```

The `invalid_txs_skipped_total` metric already provides the operator-facing signal for this case.

## Credit

Raised by researchers **shotes** and **Haxatron** on the Cantina collaborative review.

MavenRain:
Closes #1263.  (Cantina #24)

## Problem

The block builder logs every skipped `InvalidTx` at `warn!` and pretty-prints the full recovered transaction with `{:#?}`, calldata included (`crates/tn-reth/src/env/execution.rs`).

Three parts of the same block disagree with that level.  The code comment calls the case "expected in normal operation".  The `invalid_txs_skipped_total` metric docs describe it as expected duplicate traffic across workers.  The default filter (`info,evm=debug`) still delivers every occurrence to operator logs.

With multiple workers, duplicate skips are routine, so the default log fills with multi-kilobyte transaction dumps under normal load.  The noise is also attacker-inflatable at no marginal cost: a sender pays once for one large-calldata transaction, and the node dumps it at `warn!` on every skip when it lands in more than one worker's batch.  Disk use stays bounded by reth `LogArgs` file rotation;  the real damage is signal quality for operators who alert on `warn`.

The `error!` above this path (signer-recovery failure) is unchanged.  That path fires only when a certified batch carries undecodable bytes, which is Byzantine behavior and intentionally alertable.

## Fix

- [x] `crates/tn-reth/src/env/execution.rs`: downgrade the skip log to `debug!` and log the transaction hash instead of the full `{:#?}` dump.  The `invalid_txs_skipped_total` metric remains the operator-facing signal for this case.

## Testing

Log-level-only change;  no behavior change and no new items to document.  Local static ladder: `cargo +nightly fmt --all -- --check`, `cargo +1.94 check -p tn-reth`, and `cargo +1.94 clippy -p tn-reth --all-targets --no-deps -- -D warnings`, all green.  CI runs the full workspace ladder (nightly fmt, workspace clippy `-D warnings`, nextest) on push, and the pinned `+1.94` attest runs on the second machine post-push.

### DIFF
--- a/crates/tn-reth/src/env/execution.rs
+++ b/crates/tn-reth/src/env/execution.rs
@@ -169,7 +169,7 @@ impl RethEnv {
                     // allow transaction errors (ie - duplicates)
                     //
                     // it's possible that another worker's batch included this transaction
-                    warn!(target: "engine", %error,  "skipping invalid transaction: {:#?}", recovered);
+                    debug!(target: "engine", %error, tx_hash = ?recovered.hash(), "skipping invalid transaction");
                     // expected in normal operation - see the field docs, this is not an alert
                     RETH_METRICS.invalid_txs_skipped_total.increment(1);
                     continue;


grantkee (approved):
Let's add this as a metric as well

MavenRain:
> Let's add this as a metric as well

Filed a follow-up issue for this: https://github.com/Telcoin-Association/telcoin-network/issues/1284

PR #1276, opened 2026-08-26
fix(engine): drop the cumulative trie overlay from built blocks (#1266)

### LINKED ISSUE #1266: Quadratic O(N^2*M) trie-overlay copy when building blocks of one ConsensusOutput
## Summary

Building the blocks of one `ConsensusOutput` does `O(N^2 * M)` copy work and holds
`O(N^2 * M)` peak memory, where `N` is the number of blocks (batches) in the output
and `M` is the state updates per block.  The underlying state change is only `O(N * M)`.
The cause is that TN retains every ancestor's cumulative trie overlay for the whole
output loop, which defeats the copy-on-write fast path in reth's
`DeferredTrieData::sort_and_build_trie_input`.  Credit: Haxatron (Cantina collaborative
review).

## Where

- `crates/engine/src/payload_builder.rs` — the ancestor accumulation.
- `crates/tn-reth/src/env/execution.rs:188` — the call into reth.
- reth `crates/chain-state/src/deferred_trie.rs` (v1.11.3, pinned commit
  `d6324d63e27ef6b7c49cdc9b1977c1b808234c7b`) — `sort_and_build_trie_input`, Case 1.

## Mechanism

1. `payload_builder.rs` builds one block per batch in a loop.  `ancestors:
   Vec<DeferredTrieData>` starts empty and grows by one handle after each block
   (`ancestors.push(last_block.trie_data_handle())`).  Every block is executed with the
   whole accumulated slice `&ancestors`.  `anchor_hash` is computed once before the loop
   and passed unchanged to every block.

2. reth's `sort_and_build_trie_input` takes its "reuse parent overlay" fast path (Case 1)
   on every block after the first, because `ancestors.last()` exists and the parent's
   anchor equals the constant `anchor_hash`.  Case 1 clones the parent's CUMULATIVE
   overlay Arcs:

   ```rust
   let mut overlay = TrieInputSorted::new(
       Arc::clone(&trie_input.nodes),
       Arc::clone(&trie_input.state),
       Default::default(),
   );
   ...
   Arc::make_mut(&mut overlay.state).extend_ref_and_sort(&sorted_hashed_state);
   Arc::make_mut(&mut overlay.nodes).extend_ref_and_sort(&sorted_trie_updates);
   ```

   `AnchoredTrieInput.trie_input` is, by its own doc, the "Cumulative trie input overlay
   from all in-memory ancestors" — so its size is proportional to the number of ancestors.

3. `Arc::make_mut` only skips the clone when it holds the unique strong reference.  It
   never does here: the parent block's `ComputedTrieData` is still alive (the handle sits
   in `ancestors` for the whole loop) and `parent.wait_cloned()` takes another reference,
   so `strong_count >= 2`.  `make_mut` therefore deep-copies the full cumulative overlay
   on every block, then extends the copy.

4. Block `i` copies `~O(i * M)` and stores a cumulative overlay of `~O(i * M)`.  Summed
   over `N` blocks that is `O(N^2 * M)` copy work and `O(N^2 * M)` peak memory retained
   during one output build, from only `O(N * M)` of actual state change.

reth's Case 1 comment marks the branch "O(1): Reuse parent's overlay".  That is only true
when the caller hands over a uniquely-owned parent overlay so `make_mut` can mutate in
place.  TN's retain-all-ancestors pattern breaks that assumption, so the intended O(1)
reuse silently degrades to a per-block deep copy.

## Impact

Performance / resource exhaustion on the single-threaded engine execution path.  `N` is
the number of blocks in one committed `ConsensusOutput`;  it is consensus-bounded, not an
attacker free variable, but grows with committee size and backlog (a sub-dag committing
many certificates times worker batches).  `M` is gas-bounded per block but is maximized by
transactions that write many disjoint storage slots.  For small outputs the overhead is
negligible;  under load, the quadratic copy and quadratic transient memory are a
block-production-stall and transient-OOM vector.  No correctness impact: every node
computes the same result, only more slowly and with more memory.

## Suggested direction

Restore the intended O(1) reuse by giving reth's fast path a uniquely-owned parent overlay
to mutate in place, or stop retaining the cumulative overlay on ancestor handles that are
kept only for persistence.  One option: pass only the immediate parent handle (not the
whole growing slice) and arrange for its overlay Arc to be uniquely owned at the
`make_mut` call, so the extend is in place.  Any fix should be validated with a benchmark
over a large-`N` output (many storage-writing batches) confirming linear, not quadratic,
scaling of build time and peak memory.

## Reproduction sketch

Build a `ConsensusOutput` with many batches (large `N`), each batch full of transactions
that `SSTORE` disjoint slots (large `M`).  Instrument `sort_and_build_trie_input` (or
measure engine block-build wall time and RSS) and confirm total copy work grows with
`N^2`, not `N`.

MavenRain:
Closes #1266.

## Problem

Building the blocks of one `ConsensusOutput` did `O(N^2 * M)` copy work and held `O(N^2 * M)` peak memory, for `N` blocks of `M` state updates each, from only `O(N * M)` of actual state change.  `payload_builder.rs` grew an `ancestors: Vec<DeferredTrieData>` by one handle per block and passed the whole slice into reth's `DeferredTrieData::sort_and_build_trie_input`.  reth's parent-reuse fast path is only O(1) when the caller hands it a uniquely owned parent overlay.  Our retained handles keep the overlay `Arc`s alive, so `Arc::make_mut` deep-copied the full cumulative overlay for every block, and every block's bundle then retained its own cumulative snapshot until the output persisted.

`N` is consensus-bounded, not an attacker free variable, but it grows with committee size and backlog.  `M` is gas-bounded per block and is maximized by transactions that write many disjoint storage slots.  Under load the quadratic copy and the quadratic transient memory are a block-production stall and a transient-OOM vector on the single-threaded engine execution path.  No correctness impact;  every node computed the same result, only more slowly.

## Fix

Stop building the cumulative overlay at all.  Nothing on our node's path reads it:

- Persistence (`save_blocks`) and canonical-chain notifications (`blocks_to_chain`) consume only the per-block sorted `hashed_state` / `trie_updates`.
- State roots for the next block in the loop, and RPC state and proof queries over unpersisted blocks, go through reth's `MemoryOverlayStateProvider`, which assembles its trie input from the per-block sorted deltas on every call and never reads `anchored_trie_input`.
- The only readers of the cached overlay in reth (`LazyOverlay`, `OverlayStateProviderFactory`, one metrics pair in `payload_validator.rs`) are engine-tree code TN never runs, and each has a documented fallback when the overlay is absent.

Changes:

- [x] `tn-reth/src/env/execution.rs`: `build_block_from_batch_payload` loses its `anchor_hash` / `ancestors` parameters, sorts this block's deltas in parallel (`rayon::join`, mirroring reth's own sort), and constructs `ComputedTrieData::without_trie_input`.  reth provides that constructor exactly for "block builders or sequencers" with this shape.
- [x] `engine/src/payload_builder.rs`: the `ancestors` accumulation, the `anchor_hash` plumbing, and the per-block `trie_data_handle()` pushes are removed from both the empty-close-epoch arm and the batch loop.
- [x] `tn-reth/src/lib.rs`: drop the now-unused `DeferredTrieData` re-export.
- [x] Test callers in `test_utils.rs`, `tests/it/pipeline_helpers.rs`, and `e2e-tests/tests/it/staking.rs` updated to the new signature (they all passed empty ancestor slices already).

Total cost per output is now `O(N * M)` sort-and-persist work with no cumulative snapshots retained.  The change is consensus-invisible: block hashes, state roots, and persisted bytes are computed from the same inputs as before.

## Testing

- New regression test `test_built_block_retains_no_cumulative_trie_overlay` (tn-reth `it` suite) executes two chained blocks through the full pipeline and asserts each block's trie bundle carries no `anchored_trie_input`.  Mutation-confirmed: reverting the constructor to `with_trie_input` makes the test fail;  the restored fix passes.
- `cargo +1.94 check -p tn-reth -p tn-engine -p e2e-tests --all-targets`: green.
- `cargo +1.94 clippy -p tn-reth --all-targets --no-deps -- -D warnings` and `-p tn-engine --lib`: green.  The `tn-engine` `it` and `e2e-tests` `it` targets carry pre-existing 1.94 clippy findings in files this PR does not touch (CI's nightly clippy lane is the gate).
- `cargo +1.94 test -p tn-engine --lib`: 3 passed (includes the `payload_builder` persist-fault unit tests).
- `cargo +nightly fmt -- --check`: green.
- A large-`N` benchmark (many storage-writing batches) confirming linear scaling is a good follow-up;  the removed work is strictly a superset of the remaining work, so the direction cannot regress.


crates/tn-reth/src/env/execution.rs
```diff
@@ -81,12 +81,21 @@ impl RethEnv {
...
+    /// The returned block's trie bundle holds only this block's own sorted state and
+    /// trie-update deltas. It never carries the cumulative ancestor overlay, so
+    /// `anchored_trie_input` is `None`. No consumer on the node's path reads that
+    /// overlay: persistence and canonical-chain notifications use the per-block deltas,
+    /// and state roots and RPC proofs over unpersisted blocks assemble their trie input
+    /// from the same per-block deltas in reth's `MemoryOverlayStateProvider`. Building
+    /// the overlay was also not O(1) here: the parent's bundle keeps the overlay `Arc`s
+    /// alive, so reth's parent-reuse fast path (`Arc::make_mut`) deep-copies the whole
+    /// cumulative overlay for every block. One `ConsensusOutput` of `N` blocks with `M`
+    /// state updates each paid `O(N^2 * M)` copy work and transient memory (#1266).
```
grantkee: Residual `O(N^2 * M)` ancestor-delta merge remains on the same hot path. Consider updating comment: 

```rust
/// the overlay was also not O(1) here: the parent's bundle keeps the overlay `Arc`s
/// alive, so reth's parent-reuse fast path (`Arc::make_mut`) deep-copies the whole
/// cumulative overlay for every block. Dropping it removes that copy and the
/// `O(N * M)` transient memory it held (#1266).
///
/// NOT fixed by this: a separate `O(N^2 * M)` term remains in the state-root path and
/// is upstream behaviour, unchanged here. Because the whole output persists only after
/// the loop (`execute_consensus_output`), block `i` resolves its parent through a fresh
/// `MemoryOverlayStateProvider` over `i - 1` unpersisted ancestors, and
/// `MemoryOverlayStateProviderRef::trie_input()` re-merges every ancestor's deltas from
/// scratch on each call (its `OnceLock` is cloned empty by `as_ref()`, so the cache never
/// survives a block). `state_root_with_updates` then clones that merged input and
/// `TrieInputSorted::from_unsorted` re-sorts it. Total per output: `O(N^2 * M)` merge,
/// copy and sort. `N` is bounded by
/// `committee.size() * MAX_GC_DEPTH * MAX_HEADER_NUM_OF_BATCHES`.
```

follow up optimization filed in #1301 - not a blocker

grantkee (dismissed):
LGTM - one follow up issue to further optimize, but low priority

sstanfield (approved):

Haxatron:
LGTM.

PR #1280, opened 2026-08-26
fix(tn-reth): cap forwarder RPC response reads and clamp remote reasons (#1275)

### LINKED ISSUE #1275: worker forward: validator RPC responses are read and logged without a size bound
## Summary

The worker transaction forwarder reads validator RPC responses with no size bound, and it logs remote-controlled response strings with no length clamp. A byzantine committee member can advertise an endpoint that returns very large responses. Each forward attempt against that endpoint buffers the full response in memory, and the confirmed-rejection path writes the remote string to the log at `warn!` under the default filter.

Credit: surfaced by researcher shotes in the Cantina collaborative review (comment of 2026-08-26): "another `warn!` at crates/tn-reth/src/forward.rs [...] is reached by an unbound response from a validator endpoint [...]. The warn prints the rejection reason which can include the entire response bytes. Though the `warn!` should probably stay there for this one, and instead the observer forwarding the transactions should limit the response size via configuring the http provider." Line numbers below are against main `28b84e49`; the review UI cites the older review pin.

## Details

Threat model, from the module's own docs (`crates/tn-reth/src/forward.rs:188`): the endpoint is chosen by a committee member and dialed by an unattended observer process. `ForwardTargetPolicy` already treats the advertised endpoint as hostile for SSRF; the response is not treated as hostile yet.

Two gaps:

1. **Unbounded response read.** The forwarder builds `RootProvider::new_http` (`forward.rs:540`) on the default alloy reqwest transport. That transport buffers the entire response body with `resp.bytes().await` (alloy-transport-http 1.8.3, `reqwest_transport.rs:55`) before any parse, on both success and error statuses. Nothing caps the size. Only `FORWARD_SEND_TIMEOUT` (5 s) bounds the transfer, and it bounds time, not bytes. With `MAX_CONCURRENT_FORWARDS = 64`, up to 64 tasks can each buffer one such response at once.

2. **Unclamped remote strings in dispositions and logs.** `classify_error` (`forward.rs:1158`) moves remote-controlled strings into `Disposition` values verbatim:
   - `RpcError::ErrorResp(payload)`: `payload.message` is an arbitrary string from the remote server. `classify_server_error` folds it into `Disposition::Rejected(format!("code {code}: {message}"))` or `Disposition::TryNext(message)`. A held rejection at the end of the chain becomes `ForwardOutcome::Rejected` (`forward.rs:1090`), which logs the message at `warn!` (`forward.rs:655`). The rejection-overridden path also logs it at `warn!` (`forward.rs:1018`). The default filter passes `warn!`, so one malicious endpoint turns each forwarded transaction into a log write of attacker-chosen size.
   - `RpcError::Transport(err)`: the `Display` of `TransportErrorKind::HttpError` embeds the full HTTP body ("HTTP error {status} with body: {body}", alloy-transport 1.8.3 `error.rs`). Today this reason only reaches `debug!` (`forward.rs:1078`), below the default filter, but the full-body string is still allocated, and the held-rejection mirror (`forward.rs:1045`) retains rejection strings in memory for the life of the walk.

Effect: disk-write amplification through the log on the `warn!` paths, and a memory spike from the unbounded buffer, both at a rate the forward cadence repeats per transaction. No consensus impact; the observer node degrades locally.

## Fix sketch

Two independent bounds; apply both:

1. Clamp at classification. Truncate the message in `classify_error` / `classify_server_error` to a small cap (for example 256 bytes) before it enters a `Disposition`. This bounds every log site and the held-rejection retention at one choke point, and keeps the `warn!` sites (per shotes, the `warn!` should stay).
2. Bound the read itself. Replace the default transport with a size-capped one (an alloy `ClientBuilder` tower layer, or a custom `Http` transport that reads the body as a stream and stops at a cap). reqwest's `Client` has no max-response-size knob, so the cap needs a layer or a custom transport. This bounds the allocation the clamp alone cannot.

## Severity

Low. Requires a byzantine committee member (or a compromised advertised endpoint). Damage is local to observer/worker nodes: log growth and transient memory, no loss of funds or consensus effect. Same class as #1263, plus the allocation angle.

MavenRain:
Closes #1275.  (Cantina #24)

## Problem

The worker transaction forwarder dials an endpoint a committee member chose.  `ForwardTargetPolicy` already treats that endpoint as hostile at the dial site (issue #1092), but the response was still trusted:

- The provider came from `RootProvider::new_http`, whose stock reqwest transport buffers the entire response body before any parse, on success and error statuses alike.  Nothing capped the size;  `FORWARD_SEND_TIMEOUT` (5 s) bounds time, not bytes.  With `MAX_CONCURRENT_FORWARDS = 64`, up to 64 tasks could each hold one such body at once.
- `classify_error` / `classify_server_error` moved remote-controlled strings into `Disposition` reasons verbatim.  A confirmed rejection prints its reason at `warn!` (as does a rejection override), which the default filter passes, and the held-rejection mirror retains the string for the life of the walk.  One byzantine endpoint turned each forwarded transaction into a log write and a retained allocation of the server's choosing.

No attacker beyond a byzantine committee member (or a compromised advertised endpoint) is required.  The damage is local to observer/worker nodes: log growth and transient memory, no consensus effect.

## Fix

Two independent bounds, both applied (the `warn!` sites stay, per the report):

- [x] `crates/tn-reth/src/forward.rs`: a `CappedHttp` transport mirrors alloy's stock reqwest transport step for step, but pulls the body chunk by chunk and refuses it the moment it exceeds `MAX_FORWARD_RESPONSE_BYTES` (64 KiB).  reqwest's `Client` has no response-size knob of its own, so the bound lives in the transport.  An oversized body surfaces as a transport error, which classifies as `Disposition::Unreachable`: the endpoint is demoted exactly like any endpoint that produced no verdict, and a later delivery lifts the demotion as before.  Worst case buffered across all permits is now 64 x 64 KiB = 4 MiB.  The HTTP client is alloy's own re-exported `reqwest::Client`, so the TLS backend is unchanged, and `guess_local_url` keeps the stock is-local behavior.
- [x] `classify_error` / `classify_server_error` clamp every reason to `MAX_REASON_BYTES` (256 bytes, cut on a `char` boundary, with a `[truncated N of M bytes]` marker) before it enters a `Disposition`.  Every log site and the held-rejection mirror sit downstream of this one choke point.  256 bytes keeps every legitimate reth rejection message intact, and the caps are sized from the real payloads (a `send_raw_transaction` verdict is a few hundred bytes), not guessed.
- [x] `Cargo.toml` / `crates/tn-reth/Cargo.toml`: add `tower` (the `Service` trait only, `default-features = false`) so `CappedHttp` can stand where the stock transport stands.  tower 0.5 was already in the dependency graph through alloy.

## Testing

- [x] `clamp_reason` unit tests: pass-through at the cap;  `char`-boundary cut with the truncation marker.
- [x] `classify_server_error` / `classify_error` bound tests: a 1 MiB remote message (JSON-RPC error message;  HTTP-error body, whose `Display` embeds the response body) reaches the disposition reason bounded.
- [x] `collect_capped` bound test: chunks up to the cap accumulate;  the first chunk past it aborts the read.
- [x] Loopback end-to-end tests: the capped provider round-trips a small response;  a JSON-RPC error verdict still classifies ("already known" maps to `Delivered`), so the cap changes the outcome only for oversized bodies;  a body one byte past the cap is refused as a transport error.
- [x] Mutation-confirmed: with the cap check removed, the oversized-response tests fail;  with `clamp_reason` made the identity, the bound tests fail;  both restored.
- [x] Local static ladder under the pinned toolchains: nightly `cargo fmt -- --check`, `cargo +1.94 check -p tn-reth --all-targets`, nightly `clippy -p tn-reth --all-targets --no-deps`, `cargo +1.94 test -p tn-reth --lib forward::tests`.  CI runs the workspace lanes on push;  the pinned attest runs post-push.


sstanfield (approved):

grantkee (approved):
LGTM!

PR #1288, opened 2026-08-27
fix(tn-reth): splice tmp-chain deployed code into genesis to keep EIP712 immutables (#1278)

### LINKED ISSUE #1278: Genesis ceremony splices compile-time deployedBytecode, leaving ConsensusRegistry EIP712 immutables zero
## Summary

The genesis ceremony splices the compile-time `deployedBytecode.object` from the forge artifact into the real genesis.  Forge zeroes every immutable slot in `deployedBytecode.object`.  The constructor writes the real immutable values only into the code that the tmp chain deploys, and the ceremony discards that code.  So on the real chain, all five EIP712 immutables in `ConsensusRegistry` are zero.

Found by shotes on the Cantina review (confirmed there by Haxatron).  Filed here so it can be tracked and fixed upstream.

## Details

The ceremony executes the constructors on a throwaway chain, harvests the constructor-written storage, and then splices the artifact bytecode plus that storage into the real `Genesis`:

- Code comes from the artifact, not from the tmp chain: [`crates/tn-reth/src/env/genesis.rs#L329-L335`](https://github.com/Telcoin-Association/telcoin-network/blob/28b84e49274755cc6aa381421a565b1b963b2945/crates/tn-reth/src/env/genesis.rs#L329-L335) fetches `CONSENSUS_REGISTRY_JSON` -> `deployedBytecode.object`.
- The splice: [`crates/tn-reth/src/env/genesis.rs#L375`](https://github.com/Telcoin-Association/telcoin-network/blob/28b84e49274755cc6aa381421a565b1b963b2945/crates/tn-reth/src/env/genesis.rs#L375) sets that bytecode as the account code, with only the harvested storage.

Immutables live in code, not in storage, so the storage harvest cannot carry them over.

The pinned tn-contracts artifact (`0fb6b01f`) confirms the zeroing.  `artifacts/ConsensusRegistry.json` -> `deployedBytecode.immutableReferences` lists 5 immutable sites, and every referenced 32-byte segment in `deployedBytecode.object` is all zero:

```python
import json, re
a = json.load(open("artifacts/ConsensusRegistry.json"))
db = a["deployedBytecode"]
code = bytes.fromhex(db["object"].removeprefix("0x"))
for astid, refs in db["immutableReferences"].items():
    for r in refs:
        seg = code[r["start"]:r["start"] + r["length"]]
        print(astid, r["start"], "ALL-ZERO" if not any(seg) else seg.hex())
# all 5 sites print ALL-ZERO
```

The 5 immutables are the Solady `EIP712` cache (`lib/solady/src/utils/EIP712.sol#L39-L43`):

```solidity
uint256 private immutable _cachedThis;
uint256 private immutable _cachedChainId;
bytes32 private immutable _cachedNameHash;
bytes32 private immutable _cachedVersionHash;
bytes32 private immutable _cachedDomainSeparator;
```

Runtime effect with all five at zero:

1.  `_cachedDomainSeparatorInvalidated()` is always true (`_cachedChainId == 0` and `_cachedThis == 0` never match the live chain id and address).  So the zero `_cachedDomainSeparator` is never used directly, and every call rebuilds the separator.  That part is harmless.
2.  The rebuild is not harmless.  `StakeManager` does not override `_domainNameAndVersionMayChange()`, so `_buildDomainSeparator()` takes the cached branch and reads `_cachedNameHash` and `_cachedVersionHash`.  Both are zero.  The live domain separator is therefore built with `nameHash = 0` and `versionHash = 0` instead of `keccak256("Telcoin StakeManager")` and `keccak256("1")` (`src/consensus/StakeManager.sol#L334-L336` on the pin).

Affected digest paths in `ConsensusRegistry` (pin `0fb6b01f`):

- `delegationDigest` view: `src/consensus/ConsensusRegistry.sol#L574`
- `delegateStake` verification: `src/consensus/ConsensusRegistry.sol#L640` and the `SignatureCheckerLib.isValidSignatureNowCalldata` check at `#L1124`

## Impact

- Any signer that follows the EIP-712 spec produces a digest the contract rejects.  `eth_signTypedData_v4` with domain `{ name: "Telcoin StakeManager", version: "1", chainId, verifyingContract }`, ethers `TypedDataEncoder`, and forge test signing all disagree with the on-chain separator.  `delegateStake` then reverts on a valid delegation.  Standard wallet and tooling flows for delegated staking do not work.
- The on-chain `delegationDigest` view uses the same wrong separator, so a flow where the validator signs the raw digest returned by that view still verifies.  The two paths are consistent with each other, and both differ from the spec.
- The domain loses its name and version separation (`nameHash = 0`, `versionHash = 0`).  Chain id and verifying contract address remain in the separator, so cross-contract replay is still blocked.
- Off-chain verifiers that recompute the digest per spec cannot validate signatures the contract accepts, and vice versa.

Severity: Low.  No funds are at direct risk.  The documented signing flow is broken and the contract is not EIP-712 conformant.

## Suggested fix

Options, in order of preference:

1.  Harvest the deployed runtime code from the tmp chain account (the code the CREATE actually installed) and splice that, instead of the artifact `deployedBytecode.object`.  This preserves constructor-patched immutables for any current or future contract.
2.  Patch the immutable values into `deployedBytecode.object` at the offsets given by `deployedBytecode.immutableReferences` before the splice.
3.  Remove the immutable dependence in the contract: override `_domainNameAndVersionMayChange()` to return true in `StakeManager`.  The separator is then rebuilt from the string literals on every call and the zeroed immutables have no effect.  This costs extra gas per digest.

Independent of the option chosen, add a fail-loud gate next to the existing `__$..$__` link-placeholder check: bail during the ceremony when an artifact's `deployedBytecode.immutableReferences` is non-empty and the ceremony is about to splice compile-time code.  That turns any future immutable in a genesis contract into a loud failure instead of silent zeros.

## Credit

Found by shotes (Cantina review, comment on `crates/batch-builder`), confirmed by Haxatron.  Verified against telcoin-network `28b84e49` with tn-contracts pin `0fb6b01f`.

MavenRain:
Fixes #1278. (Cantina #26)

## Problem

The genesis ceremony executes the system-contract constructors on a throwaway chain, harvests the constructor-written storage, and then splices the artifact's compile-time `deployedBytecode.object` into the real genesis.  forge zeroes every immutable site in that compile-time code;  the constructor patches the real values only into the code the tmp-chain CREATE installs, and the ceremony discarded that code.

On the real chain all five Solady EIP712 immutables in `ConsensusRegistry` are zero.  The cached-separator invalidation check always fires (cached chain id and address are zero), so every digest rebuild runs `_buildDomainSeparator()` with `_cachedNameHash = _cachedVersionHash = 0`.  Any spec-conformant signer (`eth_signTypedData_v4`, ethers `TypedDataEncoder`, forge signing) produces a digest `delegateStake` rejects, and off-chain verifiers cannot validate signatures the contract accepts.  No attacker required:  the documented delegated-staking flow is broken on every deployment.

## Fix

- [x] `crates/tn-reth/src/env/genesis.rs`: splice the deployed runtime code harvested from the tmp chain's `BundleState` (account code, with a fallback to the bundle's contract map by code hash) for `ConsensusRegistry` and `WorkerConfigs`, instead of the artifact's compile-time `deployedBytecode.object`.  The deployed code carries the constructor-patched immutables.  The tmp-chain values stay correct on the real chain:  `_cachedThis` is the nonce-derived tmp address, so Solady's invalidation check always forces the rebuild path, and the rebuild reads only the chain-independent `_cachedNameHash` / `_cachedVersionHash`.
- [x] New fail-loud gate `ensure_no_immutable_references`: a contract whose compile-time artifact code is still spliced (currently only `Issuance`) fails the ceremony when its `deployedBytecode.immutableReferences` is non-empty.  A future immutable in a genesis contract becomes a loud error instead of silent zeros.  A failed code harvest (no code at the nonce-derived tmp address) also bails, mirroring the existing storage-harvest gate.
- [x] `crates/config/src/node.rs`: the mainnet code pin now compares the committed registry code byte-for-byte outside the artifact's `immutableReferences` sites and requires a non-zero value inside every site.
- [x] `chain-configs/mainnet`: regenerated via `cargo test -p telcoin-network-cli regenerate_mainnet_chain_configs -- --ignored`.  The registry account's code now carries the patched immutable segments.  The lever also refreshes the placeholder committee keys, the genesis timestamp, and the validator-key-derived registry storage slots each run;  these are throwaway placeholders, replaced before launch.

## Testing

- New `genesis_registry_code_carries_patched_immutables`: the spliced registry code must equal the artifact outside `immutableReferences`, carry a non-zero value at every one of the 5 sites, and contain `keccak256("Telcoin StakeManager")` and `keccak256("1")` among the patched segments.  Seen to fail against the pre-fix artifact splice (mutation-confirmed).
- New `immutable_reference_gate_discriminates_artifacts`: the gate rejects the registry artifact (5 immutable sites) and passes `Issuance` (none).
- Reworked `mainnet_genesis_registry_code_matches_current_artifact` is green against the regenerated chain-configs and fails on a genesis that ships an all-zero immutable segment.
- Reworked e2e `test_genesis_with_consensus_registry_accounts`:  the on-chain registry code check was a byte-for-byte compare against the compile-time artifact, which pinned the zeroed-immutable bug.  It now compares modulo `immutableReferences` (byte-equal outside the sites, non-zero inside every site), the same shape as the mainnet pin test.
- Local static ladder under the pinned toolchain:  nightly `fmt -- --check`, `cargo +1.94 check -p tn-reth -p tn-config --all-targets`, nightly `clippy --all-targets --no-deps`, plus `cargo +1.94 test -p tn-reth --lib env::genesis`.  CI runs the workspace lanes on push;  the pinned `+1.94` attest runs on the second machine.

Out of scope: the adiri `ConsensusRegistry` fork swap (`consensus_registry_runtime_code` in `evm/block.rs`) installs the same compile-time artifact bytes at the fork boundary.  A live-chain code swap cannot run a constructor, so preserving immutables there needs the `immutableReferences` patch approach (option 2 in #1278) and is left to a follow-up.


sstanfield (approved):

grantkee (approved):
Good catch

PR #1291, opened 2026-08-27
fix(tn-reth): shuffle the forward fallback dial order per forward (#1277)

### LINKED ISSUE #1277: Grindable pubkey-sorted fallback order gives one validator permanent first position on forwarded transactions
## Summary

The worker transaction forwarder tries fallback endpoints in a fixed order: ascending BLS public key.  A committee validator can grind its BLS key at registration so it sorts first.  That validator then gets the first attempt at every fallback-routed transaction from every observer worker, for the whole epoch.  Because the fallback chain stops at the first `Delivered` or `Rejected` disposition, first position is not only a receipt bias: it is a first-say position over the fate of fallback-routed transactions.

Credit: surfaced by researcher shotes in the Cantina collaborative review (comment of 2026-08-26): "The fallback list is ordered by validator bls pubkey.  A validator that grinds a favorable pubkey can always be at the top of the fallback list, creating a small bias towards receiving forwarded transactions."  Researcher Haxatron added: "related to the censorship vector discussed".  Line numbers below are against main `28b84e49`.

## Details

The ordering is deterministic at every layer:

- `Committee.authorities` is a `BTreeMap<BlsPublicKey, Authority>` (`crates/types/src/committee.rs:203`).
- `cached_providers` returns a `BTreeMap<BlsPublicKey, RootProvider>` (`crates/tn-reth/src/forward.rs:377`).
- `spawn_forward` builds the fallback list from that map's keys, so it is ascending by BLS public key (`forward.rs:451`):

```rust
let fallbacks: Vec<BlsPublicKey> = providers.keys().cloned().collect();
```

- Each transaction tries its owning validator first, then the fallbacks in that fixed order (`forward.rs:479`).

The protocol does not constrain a validator's choice of BLS key.  Grinding a key that sorts below the other committee members' keys costs a handful of key generations.  The ground key holds first fallback position for every observer worker on the network, and keeps it until a lower key joins the committee.

What first position buys, for every transaction whose owner path fails (owner did not advertise an RPC endpoint, or its send timed out or returned a transport error):

1. **Receipt bias.** The ground validator receives the transaction first, with the pool-inclusion and fee upside that carries.  This is the "small bias" in shotes' comment.
2. **First-say censorship.** The chain stops at the first `Delivered` or `Rejected` (`forward.rs:509`-`forward.rs:516`).  A malicious first fallback can accept the transaction and drop it locally;  no other validator receives it from that forward.  It can also return a considered rejection, which the forwarder trusts as universal ("No other validator will accept it either", `forward.rs:726`) and gives up.  This is the censorship vector Haxatron references: the deterministic order lets one ground key hold that position permanently instead of occasionally.

The slot routing itself (`crates/consensus/worker/src/worker.rs:100`, `owning_validator` at `forward.rs:707`) is not affected: each validator owns one slot regardless of its key, and sender addresses spread uniformly across slots.  Only the fallback order is grindable.

## Fix sketch

Shuffle the fallback list per forwarded transaction (or per batch) with a local RNG before walking it.  Keep owner-first routing.  Forwarding is a local networking decision on the observer, not a consensus input, so nondeterminism is safe here.  A shuffle removes the stable first position;  every usable endpoint then receives an equal share of fallback traffic in expectation, and a censoring validator only intercepts its random share instead of all of it.

## Severity

Low.  Requires committee membership with a ground key.  The bias covers only fallback-routed traffic (owner unadvertised or unreachable), and the censorship primitive is bounded the same way: senders can still reach validators directly, and the owner path is untouched.  But the fix is small and removes a standing incentive to grind registration keys.

MavenRain:
Closes #1277.

## Problem

#1180 (issue #1173) replaced the raw BLS-pubkey byte sort of the forwarder's fallback dial order with a per-forward rotation, so a validator can no longer grind a key that sorts first and hold the first fallback position permanently.  One grindable structure survives the rotation: the relative order is still the key sort.  A rotation is a cyclic shift, so the validator dialed after a given validator is fixed for the whole epoch.  A committee member can grind its BLS key to sort directly behind a chosen predecessor, and it then inherits that predecessor's redirect traffic whenever the predecessor is down (crashed, or attacked by the same member).  The inherited position carries the same first-say weight over those transactions that #1180 removed for the head of the list.

The rejection half of #1277's censorship concern is already closed by #1175 (a considered rejection is held until a second validator confirms it).  This PR removes the remaining order structure.

## Fix

- [x] `crates/tn-reth/src/forward.rs`: `spawn_forward` draws a fresh uniform shuffle of the fallback list per spawned forward.  `shuffled_fallbacks` is rand's in-place Fisher-Yates (`SliceRandom::shuffle`, per review), so no choice of committee key buys a position anywhere in the walk: first position and every successor are uniform per forward.
- [x] Per forward, not per transaction: every transaction of one forward walks the same order, so one account's transactions stay on one fallback validator within a batch (nonce-gap hygiene), matching the owner-first routing, which is unchanged.
- [x] The `fallback_rotation` counter (field, random seed, `rotated_fallbacks`, `rotation_start`) is retired;  the shuffle subsumes the rotation.
- [x] Totality: the shuffle is an in-place permutation of the owned list, so no fallback entry can be dropped or duplicated.

The shuffle is attacker-independent: the dial-order distribution is the same whatever key a member registers.  Forwarding stays a local networking decision on the observer, not a consensus input, so the nondeterminism crosses no trust boundary.

## Testing

- `shuffled_fallbacks_permutes_without_losing_entries` (replaces the rank-stream test, per review): 64 seeded shuffles of eight keys;  every shuffle is a permutation (no entry dropped or duplicated), at least one leaves input order, and the empty list stays empty.  Deliberately not an exact-order assertion: rand documents `StdRng` and the shuffle algorithm as free to change in any release.
- `spawned_forwards_spread_first_dials_across_fallbacks` (replaces the rotation test): 64 single-transaction forwards of an ownerless transaction against three counting endpoints;  every endpoint must receive at least one first dial.  A uniform shuffle fails this with chance 3 * (2/3)^64, under 2e-11 (the pinned-counter determinism died with the counter;  the bound is far below any infrastructure failure rate).  Each forward holds one instant local dial, so no batch-budget timing couples in.
- Both tests were confirmed by mutation: with the shuffle degraded to the identity order (empty-prefix shuffle), both fail;  restored, both pass.
- Local static ladder under the pinned toolchains: nightly `cargo fmt -p tn-reth -- --check`, `cargo +1.94 check -p tn-reth --all-targets`, nightly `cargo clippy -p tn-reth --all-targets --no-deps`, `cargo +1.94 test -p tn-reth --lib forward::tests`.


crates/tn-reth/src/forward.rs
```diff
@@ -883,33 +871,33 @@ fn next_txn_budget(deadline: Instant, now: Instant) -> Option<Duration> {
...
+    // The index in the sort key is what keeps this total: equal ranks (and a rank stream
+    // shorter than the list, padded here with `u64::MAX`) collapse to input order instead of
+    // dropping entries behind a duplicate map key.
+    let ranked: BTreeMap<(u64, usize), BlsPublicKey> = fallbacks
+        .into_iter()
+        .zip(ranks.into_iter().chain(std::iter::repeat(u64::MAX)))
+        .enumerate()
+        .map(|(index, (key, rank))| ((rank, index), key))
+        .collect();
+    ranked.into_values().collect()
```
grantkee: Instead of hand-rolling, consider using `rand::shuffle`:

Add to the imports at the top of `forward.rs`: `use rand::{seq::SliceRandom as _, Rng};`

```diff
-fn shuffled_fallbacks(
-    fallbacks: Vec<BlsPublicKey>,
-    ranks: impl IntoIterator<Item = u64>,
-) -> Vec<BlsPublicKey> {
-    // The index in the sort key is what keeps this total: equal ranks (and a rank stream
-    // shorter than the list, padded here with `u64::MAX`) collapse to input order instead of
-    // dropping entries behind a duplicate map key.
-    let ranked: BTreeMap<(u64, usize), BlsPublicKey> = fallbacks
-        .into_iter()
-        .zip(ranks.into_iter().chain(std::iter::repeat(u64::MAX)))
-        .enumerate()
-        .map(|(index, (key, rank))| ((rank, index), key))
-        .collect();
-    ranked.into_values().collect()
-}
+fn shuffled_fallbacks(mut fallbacks: Vec<BlsPublicKey>, rng: &mut impl Rng) -> Vec<BlsPublicKey> {
+    // Fisher-Yates, in place: uniform over all permutations, O(n), and it reuses the caller's
+    // allocation rather than staging the list through a map.
+    fallbacks.shuffle(rng);
+    fallbacks
+}
```

Keep the whole #1173 / #1277 rationale paragraph verbatim; only the summary line and the tiebreak sentence change (`sorted by \`ranks\`, ties ... staying in input order` → `shuffled uniformly with \`rng\``).

Call site (`forward.rs:558-561`):

```diff
-        let fallbacks = shuffled_fallbacks(
-            providers.keys().cloned().collect(),
-            std::iter::repeat_with(rand::random::<u64>),
-        );
+        let fallbacks = shuffled_fallbacks(providers.keys().cloned().collect(), &mut rand::rng());
```

Replacement for the unit test at `forward.rs:2166-2181` — note it must **not** assert an exact seeded permutation:

```rust
    /// The shuffle is a permutation: no entry is ever dropped or duplicated, and the order
    /// actually moves. Deliberately not an exact-order assertion - rand documents `StdRng`'s
    /// and the shuffle's algorithms as free to change in any release.
    #[test]
    fn shuffled_fallbacks_permutes_without_losing_entries() {
        let keys: Vec<_> = (1..=8).map(test_key).collect();
        let expected: BTreeSet<_> = keys.iter().copied().collect();
        let mut rng = StdRng::seed_from_u64(1277);
        let mut moved = false;
        for _ in 0..64 {
            let shuffled = shuffled_fallbacks(keys.clone(), &mut rng);
            assert_eq!(shuffled.len(), keys.len(), "a shuffle changed the fallback count");
            assert_eq!(
                shuffled.iter().copied().collect::<BTreeSet<_>>(),
                expected,
                "a shuffle dropped or duplicated a fallback"
            );
            moved |= shuffled != keys;
        }
        assert!(moved, "64 shuffles of 8 keys never left input order");
        assert!(shuffled_fallbacks(Vec::new(), &mut rng).is_empty());
    }
```

grantkee (dismissed):
LGTM - one suggestion to use `rand` lib's `shuffle` to reduce the amount of code in `shuffle_fallbacks`

crates/tn-reth/src/forward.rs
MavenRain: Adopted . . .  `shuffled_fallbacks` is now rand's in-place `SliceRandom::shuffle`, the call site passes `&mut rand::rng()`, and the unit test asserts permutation plus movement instead of an exact seeded order.

sstanfield (approved):

grantkee (approved):

PR #1293, opened 2026-08-27
ci: run the PR checks on merge_group so main can use a merge queue

### LINKED ISSUE #1292: rapid development causes semantic merge conflicts
## Problem

The repo is very active and it's cumbersome to constantly rebase `main` and reattest. Semantic merge conflicts occassionally occur (see #1283 for example)

## Solution

Try merge queue and see if that reduces friction.

grantkee:
The attestation gate answers "did the full suite pass on this PR's head commit". It cannot answer "does main still work once this PR lands", which is the question that produces the semantic merge conflicts maintainers keep cleaning up: two PRs each green on their own, broken together. A merge queue answers exactly that, so the two gates are complementary and this keeps both.

pr.yaml now triggers on merge_group as well as pull_request. That is not optional plumbing: GitHub evaluates the branch's required checks a second time against the queue's temporary commit, and a required check that never reports on merge_group does not fail the queue, it hangs it until the timeout ejects the PR.

- maintainer-verify.yaml is folded into pr.yaml as the `verify-on-chain` job, so `CI Success` is the only required check name and only one workflow has to know about merge_group. The job drops `environment: merge-into-main`; it does a read-only cast call with no secrets, and a protection rule on that environment would park the queue on a manual approval until it timed out.
- The merge group's own commit hash is created by GitHub seconds before CI starts and can never be attested. verify_attestation.sh walks the group's first-parent chain instead and verifies the PR heads folded into it. Derivation trouble or an adiri outage warns and passes there rather than ejecting a whole batch over a problem that is not the code's -- those heads were already verified on the pull_request event, which has to pass before a PR may enter the queue. A head the registry positively reports as unattested still fails hard.
- verify_commit_hash.sh retries the RPC and now separates "not attested" (exit 1) from "adiri never answered" (exit 2). Previously set -e turned a single flaky read into an indistinguishable "not attested".
- The maintainer and draft skips stay on pull_request and are disabled in the queue: whoever wrote it, the merged commit has never been compiled anywhere. This held before only by accident, because the PR-author login is empty on merge_group. CI Success now treats a skipped lane as a failure there for the same reason.
- CI Success allowlists success/skipped instead of blocklisting failure. The old form greened the gate on a cancelled job.
- Never cancel a merge_group run: a cancelled run reports a non-success conclusion and ejects the PR. cancel-in-progress is now scoped to pull_request, and the group's ref keys the concurrency group.
- Added timeout-minutes to every job and contents: read at the workflow level.

test-and-attest.sh refuses to attest a branch that is behind origin/main (ALLOW_STALE_BASE=1 overrides). The queue re-runs fmt, clippy and both test lanes on the merged commit, but it cannot afford e2e -- that only ever runs locally, so attesting a stale base yields an e2e result for a combination that is not the one being merged. Its `cd` to the repo root also moves above `source .env` so the script works from any cwd.

.github/ACTIONS.md documents the split and the ruleset changes this needs: enable the merge queue, make `CI Success` the sole required check, and drop `verify-on-chain` from the required list.

closes #1292


MavenRain (dismissed):
Yes, plz and thx

sstanfield (dismissed):

MavenRain (approved):

PR #1294, opened 2026-08-27
feat(tn-reth): draw  epoch-close committees region-aware behind a fork gate (#1279)

### LINKED ISSUE #1279: tn-reth: committee shuffle ignores validator region, contrary to tn-contracts design.md
## Summary

The ConsensusRegistry contract stores a GSMA `region` for each validator, and governance can set it with `setValidatorRegion`. The tn-contracts design doc describes a region-aware committee shuffle and states the algorithm "is implemented in the Rust protocol client, not in Solidity". The client's committee selection is region-blind: `assemble_new_committee` reads only `currentStatus` and `validatorAddress`, runs one flat Fisher-Yates shuffle, and discards `region`. No other client code reads the field. Either the shuffle needs the documented algorithm (behind a fork gate), or the docs need to say the algorithm is not implemented yet.

Credit: surfaced by researcher Haxatron in the Cantina collaborative review (comment of 2026-08-26): "Committee selection is still not region aware contrary to the docs stated here: https://github.com/Telcoin-Association/tn-contracts/blob/master/src/consensus/design.md#region-aware-committee-shuffle" and "unsure if this is intentional or if the docs are outdated". Line numbers below are against main `28b84e49`.

## Details

The code path:

- `read_committee_eligible_pool` (`crates/tn-reth/src/evm/block.rs:992`) decodes the full `ConsensusRegistry::ValidatorInfo` struct for every eligible validator via `getValidatorsInfo` (`block.rs:1012-1019`). The struct includes `uint8 region` (`crates/tn-reth/src/system_calls.rs:92-94`), so region data is already in memory at the shuffle site.
- The pool flows intact into `assemble_new_committee` (`block.rs:981`, body at `block.rs:1557-1614`). That function partitions by `currentStatus == PendingExit` (`block.rs:1574-1575`), runs a flat Fisher-Yates over the merged pool (`block.rs:1594-1597`), and projects out only `validatorAddress` (`block.rs:1601-1602`). `region` is never read.
- A repo-wide sweep at `28b84e49` finds 11 occurrences of `region` in the client: the ABI declarations, hardcoded `region: 0` defaults in genesis and test fixtures, and unrelated prose. No selection path, epoch code, or feature-gated variant consults the field. `setValidatorRegion` exists only as an ABI binding with no call site.

The history shows an intentional deferral that lost its tracking issue:

- telcoin-network PR #653 (merged 2026-05-05) added the `region` field and the setter binding, and its body says so directly: "The shuffle does not use this information yet. The shuffle needs to be updated before assigning regions to validators to prevent forks during syncing." The squashed history contains a sub-commit "restore original shuffle - do not use validator region yet", so a shuffle change was drafted inside that PR and backed out before merge.
- tn-contracts #92 scoped the on-chain half (storage plus views) and deferred the client shuffle. It closed 2026-06-05 with only the storage half shipped. Neither repo has a follow-on issue for the client half.
- The docs claim the client half exists, in the present tense. design.md (line 93): "The committee shuffle algorithm above is implemented in the Rust protocol client, not in Solidity." invariants.md (line 68): "region is used by the protocol's Rust-side committee shuffle algorithm, not by Solidity logic."

So the answer to the reviewer's question is both: the omission was intentional at merge time, and the docs overstate what shipped.

Effect today: `setValidatorRegion` succeeds, emits its event, and changes nothing. An operator who assigns regions and trusts the documented guarantee gets no geographic diversity in committee selection. Setting the field causes no consensus divergence, because no client code reads it.

## Fix sketch

Two options; either closes the divergence:

1. Implement the documented algorithm in `assemble_new_committee`. The RNG draw order is consensus-critical (`block.rs:1591`), so the new draw sequence must sit behind a fork epoch, with the flat shuffle preserved for historical replay. This is the same concern PR #653 named ("prevent forks during syncing"). No new ABI plumbing is needed: `eligible_pool: Vec<ConsensusRegistry::ValidatorInfo>` already carries `region`.
2. Correct tn-contracts `design.md` and `invariants.md` to present the region-aware shuffle as planned, not implemented, and open a tracking issue for the client work so the deferral stays visible.

Until one of these lands, a doc note that `setValidatorRegion` has no protocol effect would prevent false assurance.

## Severity

Informational. Docs and code diverge; no attacker profit and no consensus impact. The risk is operational: a governance action that appears to add geographic fault tolerance does nothing.

MavenRain:
Closes #1279. (Cantina #29)

## Summary

The ConsensusRegistry stores a GSMA `region` for each validator, and tn-contracts `design.md` documents a region-aware committee shuffle as "implemented in the Rust protocol client".  The client is region-blind: `assemble_new_committee` runs one flat Fisher-Yates over the merged pool and never reads `region`.  PR #653 drafted the region-aware shuffle and backed it out before merge ("restore original shuffle - do not use validator region yet") because the draw order is consensus-critical and needs a fork boundary.  This PR lands the documented algorithm behind one.

## Changes

- [x] `crates/types/src/forks.rs`: `REGION_SHUFFLE_FORK_EPOCH` (adiri, `u32::MAX` placeholder) plus `region_shuffle_active(epoch)` and a `TN_REGION_SHUFFLE_FORK_EPOCH` test override, mirroring the seed-signature (#1032) and multi-workers (#554) gates.  The gate reads the epoch the assembled committee will serve (the concluding epoch plus one, from the block's own nonce), never node-local state, so production and replay derive the same verdict.
- [x] `crates/tn-reth/src/evm/block.rs`: `assemble_new_committee` takes the gate verdict and branches only the ordering step.  The legacy arm keeps the historical flat draws verbatim (the golden-order unit test pins them).  The region arm implements the design.md sequence: Fisher-Yates inside each region group in ascending region order, Fisher-Yates over the unassigned (region 0) pool, Fisher-Yates over the region visit order, round-robin one seat per region per round, then the unassigned pool fills the remaining seats.  Both arms share one `fisher_yates` helper, the pending-exit pool assembly, the trim, and the undersized check.
- [x] `crates/telcoin-network-cli/src/node.rs`: `region_shuffle_fork_epoch` joins the startup fork-schedule log.

## Fork safety

- The draw order decides which validators survive the truncation to committee size, so the switch is a consensus change and follows the standard rollout: this PR ships the gate dormant on adiri;  a follow-up epoch-setting PR arms it once the fleet runs a gate-capable build.  Non-adiri builds are active from genesis, on the same assumption the other gates state: no non-adiri network carries flat-shuffle history.
- While every validator's `region` is 0 (the storage default; `setValidatorRegion` is the only writer and has no protocol effect today), the region-aware sequence draws identically to the flat shuffle, element for element: the partition draws nothing, the per-group and visit-order shuffles are empty, and the unassigned shuffle covers the whole pool in pool order.  An armed fork is therefore output-identical until governance assigns a region;  a unit test asserts the identity.  The operational constraint runs the other way and is documented on the constant: governance must not assign a nonzero region until the fork epoch has begun fleet-wide.
- The pre-registry-fork legacy ABI read decodes the byte-identical struct with `region` 0 absent any governance write, so replayed pre-fork epochs stay on the flat draws twice over (dormant gate, zero regions).
- design.md's final fill step ("if still not full, additional assigned validators fill remaining seats") is vacuously satisfied: the round-robin only ends short of the target when every assigned group is exhausted.  A pool below the target still fails closed with `UndersizedCommittee`.

## Testing

- `cargo +nightly fmt -- --check` clean.
- `cargo check --workspace --all-targets` green on the pinned 1.94 toolchain.
- clippy over the buildplan-emitted crate list with `--no-deps` green.
- Unit tests (`cargo test -p tn-reth --lib`, `cargo test -p tn-types --lib`):
  - `assemble_region_aware_degrades_to_flat_when_no_regions_assigned` (the identity above),
  - `assemble_region_aware_seats_one_validator_per_region_in_round_one` (diversity plus same-seed determinism),
  - `assemble_region_aware_second_round_tops_up_regions_evenly` (per-region seat counts {2, 2, 1, 1} for six seats over four regions),
  - `assemble_region_aware_fills_from_unassigned_after_round_robin`,
  - `assemble_region_aware_rejects_undersized_pool`,
  - `region_shuffle_build_fork_gate_matches_this_builds_rollout_contract` and `region_shuffle_override_is_inert_when_unset` (fork gate rollout contract),
  - the existing golden-order test pins the legacy draw sequence unchanged.
- Each new load-bearing test was confirmed by mutation: breaking the round-robin interleave, skipping the unassigned shuffle, and dropping the unassigned fill each failed the matching test;  the tree was restored and the suite re-run green.
- CI's nextest lane runs on push;  the pinned 1.94 attest runs post-push on the second machine.


sstanfield (approved):
I thought there was a general question (beyond a fork) about when to activate this. But if we are ready this looks ok to me.

Haxatron:
Note: The downside with a region aware shuffle design to consider is that the validators can submit the false region with lowest number of validators to increase their chances of getting into a committee, even if the validator does not live in the region

Perhaps this should be monitored and then slashed via the smart contract

MavenRain:
Thanks, that's a real gap.  I've opened #1327 to track it: https://github.com/Telcoin-Association/telcoin-network/issues/1327 


One correction to the premise worth noting there: validators can't write their
 own region, `setValidatorRegion` is `onlyOwner`, so the vector is a false location claim made to governance rather than a direct self-report.  The incentive still holds because the round-robin seats one validator per assigned second paragraph and the first sentence alone still stands. 

I have updated the memory topic file to record that the draft was filed as #1327.

MavenRain:
@sstanfield The fork is set at `u32::MAX`, allowing us to separate the question of when this fork lands from the plumbing to support it, which, itself, is justified by the Cantina finding.

PR #1295, opened 2026-08-27
fix: resolve eth_feeHistory base fee per query instead of pinning a container clone

### LINKED ISSUE #1282: eth_feeHistory's long-held BaseFeeContainer clone goes stale for worker ids >= 1 after a worker-count shrink and regrow (multi-worker follow-up to #1235)
Follow-up to @grantkee's review note on #1235 (approved, "not a blocker"):

> `initialize_worker_components` runs once per node lifetime and moves a `BaseFeeContainer` clone into the RPC server, which holds it forever.  `GasAccumulator::set_num_workers` truncates slots on shrink and creates fresh default containers on regrow, so for any worker id ≥ 1 a shrink→regrow across epoch boundaries leaves the still-running RPC server holding a container that `EpochBaseFees::apply` no longer writes to — `eth_feeHistory` would quote a permanently stale fee for that worker.  Unreachable today: only `DEFAULT_WORKER_ID` (0) is initialized, the count is clamped to ≥ 1, and grows are index-stable, so slot 0's container is stable for the node lifetime.  The multi-worker foundation ( #554 / #1219 ) is actively landing, and the comment at `start_epoch.rs:439` ("otherwise, loop here") marks this exact call site as the future multi-worker loop.

## Problem

#1235 hands the RPC server a live `BaseFeeContainer` so `eth_feeHistory` serves its next-block entry from the epoch schedule.  The handoff is a one-time move:

- `ExecutionNodeInner::initialize_worker_components` (`crates/node/src/engine/inner.rs`) runs once per node lifetime.  It moves a `BaseFeeContainer` clone into the RPC server.  The server holds that clone until shutdown.
- `GasAccumulator::set_num_workers` (`crates/types/src/gas_accumulator.rs`) resizes with `inner.resize_with(num_workers, Accumulated::default)`.  A shrink truncates the high slots.  A regrow creates fresh default containers in their place.

For a worker id >= 1, a shrink followed by a regrow across epoch boundaries replaces that slot's container.  `EpochBaseFees::apply` then writes each epoch's fee into the new container.  The RPC server still holds the old one.  `eth_feeHistory` quotes a permanently stale fee for that worker, silently, on every request.

## Why this is unreachable today

- Only `DEFAULT_WORKER_ID` (0) is initialized (`crates/node/src/manager/node/start_epoch.rs`, "only support one worker for now (with id 0) - otherwise, loop here").
- `set_num_workers` clamps the count to >= 1, so slot 0 is never truncated.
- Grows are index-stable: existing slots keep their index and their container.

Slot 0's container is therefore stable for the node lifetime, and worker 0 is the only worker with an RPC server.

## When it becomes real

The multi-worker series (#554 through #559;  foundation merged as #1219) replaces the single-worker call site with a per-worker loop.  Once a worker with id >= 1 gets an RPC server, the shrink-and-regrow path is reachable through ordinary committee-driven worker-count changes, and the stale quote follows.

## Possible fixes

1. **Hand the RPC server a lookup, not a clone (recommended).**  Pass the `GasAccumulator` (or a narrow handle over it) into the RPC layer and resolve `base_fee(worker_id)` per query.  The RPC then always reads the slot the accumulator currently owns, and no lifecycle coordination is needed.
2. **Make `set_num_workers` container-preserving on regrow.**  Keep truncated slots' containers and reattach them when the count grows back.  This cuts against the current design: `inc_block` panics on truncated ids by design, and per-epoch counters must reset.
3. **Fold RPC re-pointing into the multi-worker loop.**  Re-run (or extend) worker-component initialization when the count changes.  This changes the run-once lifecycle of `initialize_worker_components`.

## Verification

- Unit test that pins the hazard: hold a clone of slot 1's container, `set_num_workers(1)` then `set_num_workers(2)`, run `EpochBaseFees::apply`, and assert the held clone does not see the new fee.  After the fix, assert the RPC read path returns the live value instead.  (The shrink direction through `apply` already has coverage in `crates/node/src/manager/node/run_epoch.rs`.)
- E2E, once the multi-worker loop lands: change the worker count across epoch boundaries, query `eth_feeHistory` on a worker id >= 1, and assert the next-block entry matches the epoch schedule on every validator.

## Dependencies

- #1235 introduces the held clone (this issue describes code on that PR).
- #557 / #558 make the path reachable;  the fix belongs with (or before) that work.

MavenRain:
Closes #1282.

## Problem

The RPC server received a one-time `BaseFeeContainer` clone at worker startup and held it for its whole life.  `GasAccumulator::set_num_workers` truncates the per-worker slots on a shrink and creates fresh default containers on a regrow, so for worker ids >= 1 the held clone detached from the slot the accumulator owns.  From then on the epoch schedule (`EpochBaseFees::apply`) wrote the new container while `eth_feeHistory` and `eth_blobBaseFee` kept quoting the orphaned one: a permanently stale base fee for the rest of the node's life.  No attacker is required;  a routine committee-driven worker-count shrink and regrow is enough.

## Fix

Hand the RPC server a lookup, not a clone (option 1 from the issue).  A new `WorkerBaseFee` handle wraps the `GasAccumulator` plus a `WorkerId` and resolves the worker's slot on every read, so a long-lived holder always sees the container the accumulator currently owns.  The handle is total: a read while the slot is truncated answers `MIN_PROTOCOL_BASE_FEE`, which is exactly the value the regrown slot is reborn with, so a fee quote never panics a running node (in contrast to `inc_block`, whose out-of-range panic is by design on the consensus path).

`initialize_worker_components` and `get_rpc_server` now take the handle instead of a pinned container clone;  the read site in `FeeHistoryWithEpochBaseFee` is unchanged (`self.base_fee.base_fee()` at query time).

## Changes

- [x] `crates/types/src/gas_accumulator.rs`: add `WorkerBaseFee` handle and `GasAccumulator::worker_base_fee`;  document the staleness hazard on `GasAccumulator::base_fee`
- [x] `crates/tn-reth/src/rpc_fee_history.rs`: `FeeHistoryWithEpochBaseFee` stores a `WorkerBaseFee` instead of a `BaseFeeContainer` clone
- [x] `crates/tn-reth/src/env/rpc.rs`: `get_rpc_server` takes the handle;  add a shrink-and-regrow RPC regression test
- [x] `crates/node/src/engine/inner.rs`, `crates/node/src/engine/mod.rs`: `initialize_worker_components` takes the handle
- [x] `crates/node/src/manager/node/start_epoch.rs`: build the handle from the accumulator and pass it through
- [x] `crates/node/tests/it/main.rs`: source the test's base fee through the accumulator and pass a handle

## Testing

- Three new `tn-types` unit tests: `held_container_clone_goes_stale_after_shrink_and_regrow` (reproduces the bug and shows the handle survives it), `worker_base_fee_answers_rebirth_default_while_truncated`, and `worker_base_fee_and_clone_agree_across_a_pure_grow`.
- New `tn-reth` regression test `test_fee_history_quote_tracks_accumulator_across_shrink_and_regrow`: builds the production RPC server for worker 1 of 2, shrinks to 1 and regrows to 2, writes a fresh epoch fee, and asserts `eth_feeHistory` quotes the live fee.
- Each new test was confirmed by in-diff mutation (reverting the per-read resolution made the stale-clone and RPC tests fail;  changing the truncated-slot default made the rebirth test fail), then the fix was restored.
- Local static ladder under the house gates: nightly `cargo fmt -- --check`, `cargo +1.94 check --workspace --all-targets`, scoped `clippy --no-deps` on the touched crates.


sstanfield (dismissed):

grantkee (approved):
LGTM!

PR #1296, opened 2026-08-27
fix(tn-reth): label the skipped-invalid-transaction metric by error kind (#1284)

### LINKED ISSUE #1284: Label the skipped-invalid-transaction metric by error kind (follow-up to PR #1274)
## Problem

PR #1274 changed the skipped-invalid-transaction log in `build_block_from_batch_payload` (`crates/tn-reth/src/env/execution.rs:172`).  The log level went from `warn!` to `debug!`.  The message now carries the transaction hash instead of the full transaction dump.

@grantkee approved the PR with this review comment:

> Let's add this as a metric as well

The skip count is already a metric.  The same match arm increments `tn_reth.invalid_txs_skipped_total` on the line after the log (`crates/tn-reth/src/env/execution.rs:174`, field at `crates/tn-reth/src/metrics.rs:112`).

One signal is not covered.  The counter is unlabeled.  The old `warn!` line put the validation error in the default operator log.  After #1274, that cause is only visible with `debug`-level logging enabled.  An operator who sees the counter climb cannot tell expected duplicates (the same transaction landed in more than one worker's batch) from other `InvalidTx` causes.

## Proposal

Label the skip metric by error kind, so the cause the old `warn!` carried is visible in metrics:

- Add a small closed reason enum that maps `BlockValidationError::InvalidTx` causes to a bounded label set, with a catch-all `other` variant to keep cardinality fixed.
- Follow the existing labeled-counter pattern in the crate: `tn_reth.forwarded_txns_dropped_total` with `ForwardDropReason` (`crates/tn-reth/src/metrics.rs:139`).  That pattern pre-registers every label value at zero, so a healthy node exports the full series.
- Implementation note from the module docs: the `reth_metrics::Metrics` derive cannot express labels.  A labeled series uses the `metrics::counter!` macro path, the same way the forwarder drop series does.

The unlabeled `tn_reth.invalid_txs_skipped_total` can stay as the total, or the labeled series can replace it.  Maintainer preference decides.

## References

- PR #1274 (the log change and the review comment)
- Issue #1263 (the log-noise report #1274 closed)
- `crates/tn-reth/src/env/execution.rs:161-180` (the skip path)
- `crates/tn-reth/src/metrics.rs` (existing counters and the labeled-series pattern)

MavenRain:
Closes #1284.

## Problem

PR #1274 moved the skipped-invalid-transaction log in `build_block_from_batch_payload` from `warn!` to `debug!`.  The validation error the old `warn!` carried left the default operator log with it.  The skip count is a metric (`tn_reth.invalid_txs_skipped_total`), but the counter is unlabeled: an operator who sees it climb cannot tell expected duplicates (the same transaction landed in more than one worker's batch) from other `InvalidTx` causes.  Grant asked for the metric half in the #1274 review: "Let's add this as a metric as well".

## Fix

Label `tn_reth.invalid_txs_skipped_total` by error kind, following the crate's existing labeled-counter pattern (`tn_reth.forwarded_txns_dropped_total` with `ForwardDropReason`).

- [x] `crates/tn-reth/src/metrics.rs`: new `InvalidTxSkipReason` enum maps `BlockValidationError::InvalidTx` causes onto a closed label set: `nonce_too_low` (the expected-duplicate signal), `nonce_too_high`, `insufficient_funds`, `fee_cap_below_base_fee`, and a catch-all `other` that keeps cardinality fixed across revm upgrades.  The match on revm's `InvalidTransaction` is exhaustive, so a dependency bump that adds a variant fails to compile here instead of landing in a bucket nobody chose.  Classification goes through `alloy_evm::InvalidTxError::as_invalid_tx_err`, and an error that exposes no underlying `InvalidTransaction` counts as `other`.
- [x] The labeled series replaces the unlabeled field on `RethEnvMetrics` under the same name, so the total is the sum over the `reason` label and no second series double-counts the skips.  The `reth_metrics::Metrics` derive cannot express labels, so the series moves to the `metrics::counter!` macro path the forwarder drop series already uses.
- [x] `init` pre-registers every label value at zero, same rationale as the existing counters: a healthy node exports the full series, so an absent series stays distinguishable from a broken exporter, and the first skip renders as a `rate()` step from zero.
- [x] `crates/tn-reth/src/env/execution.rs`: the skip arm classifies the error and records the labeled skip.  The `debug!` line with the exact error is unchanged;  the label carries the cause class into metrics.

The bound is fixed and attacker-independent: five label values, chosen at compile time.  Batch content cannot grow the label set, so a transaction crafted to fail with an exotic cause lands in `other` instead of minting a new series.

## Testing

- `test_invalid_tx_skip_reasons_register_at_zero_and_count_separately`: registration exports every reason at zero;  a recorded skip moves only its own reason's series.
- `test_invalid_tx_skip_reason_classification`: each named cause maps to its own label;  an unnamed cause and an error with no underlying `InvalidTransaction` both map to `other`.
- The two pre-existing `RethEnvMetrics` tests drop their skip-counter halves;  the register-at-zero property for the skip series now lives in the labeled test above.

Local static ladder under the pinned toolchain: `cargo +nightly fmt --all -- --check`, `cargo +1.94 check -p tn-reth --all-targets`, `cargo +1.94 clippy -p tn-reth --all-targets --no-deps -- -D warnings`, `cargo +1.94 test -p tn-reth --lib -- metrics::tests`.  Both new tests were mutation-confirmed: reverting the classify arm and the increment made them fail, restoring the code made them pass.


grantkee (dismissed):
Merge conflict - but otherwise LGTM - flagging #1302 for follow up

grantkee (approved):

Issue #1283, opened 2026-08-27
main fails cargo check --workspace --all-targets: eth_syncing test calls get_rpc_server without the #1235 BaseFeeContainer argument


MavenRain:
## What happens

`origin/main` (c0bbcb3a) does not compile when test targets build:

```
error[E0061]: this method takes 4 arguments but 3 arguments were supplied
   --> crates/tn-reth/src/env/rpc.rs:411:31
    |
411 |         let server = reth_env.get_rpc_server(pool, network.clone(), RpcModule::new(()))?;
    |                               ^^^^^^^^^^^^^^                        ------------------ argument #3 of type `tn_types::gas_accumulator::BaseFeeContainer` is missing
```

Any of these reproduce it on main:

```sh
cargo check --workspace --all-targets
cargo test -p tn-reth --lib
```

## Why

Two PRs crossed:

- #1235 (34f0f83b) added the `base_fee: BaseFeeContainer` parameter to
  `RethEnv::get_rpc_server` (crates/tn-reth/src/env/rpc.rs:85-89) and updated the
  call sites that existed at the time.
- #1232 (1e4fc55f) merged after #1235, but its branch was based on 28b84e49,
  before the signature change.  Its new test
  `test_eth_syncing_answers_from_worker_shim_flag` calls the 3-argument form.
  The merge was textual, so the stale call landed on main.

Pre-merge CI ran on the #1232 branch, where the 3-argument call was still
correct.  Nothing rebuilt the merged result with `--all-targets`.

## Fix

Pass a default container in the test, the same value the
`new_for_temp_chain` path uses (crates/tn-reth/src/env/rpc.rs:200):

```rust
let server = reth_env.get_rpc_server(
    pool,
    network.clone(),
    BaseFeeContainer::default(),
    RpcModule::new(()),
)?;
```

The base-fee value does not matter for this test;  it asserts `eth_syncing`
answers only.

Note: the pending merge of main into PR #1237 carries this same one-line
resolution inside its merge commit, so that branch compiles.  Main needs the
fix directly for every other consumer.

PR #1297, opened 2026-08-27
fix(tn-reth): reject value-bearing calls to the BLS precompile (#1286)

### LINKED ISSUE #1286: bls precompile: no payability guard, so value attached to `blsVerify` is stranded at 0xb151 with no outflow path
## Summary

Issue #1190 established that a custom precompile must enforce payability itself, because registering the address short-circuits the interpreter and no compiled `CALLVALUE` guard ever runs.  PR #1201 (merged 2026-08-21) added that guard to the TEL dispatcher at `0x7e1`.  The BLS precompile at `0xb151` has the same exposure and no guard.

The entrypoint forwards only calldata and gas (`crates/tn-reth/src/evm/bls_precompile/mod.rs:125-127`):

```rust
fn bls_precompile(input: PrecompileInput<'_>) -> PrecompileResult {
    dispatch(input.data, input.gas)
}
```

No code in the module reads `PrecompileInput::value`.  `blsVerify(..)`, the only selector (`mod.rs:139-142`), accepts a nonzero call value.

The value-transfer semantics are the ones #1190 documented for `0x7e1`, because both precompiles register through the same machinery: `add_bls_precompile` (`mod.rs:115-121`) and `add_telcoin_precompile` (`tel_precompile/mod.rs:126-132`) each call `PrecompilesMap::apply_precompile`, and one `PrecompileProvider::run` impl serves both addresses.  On the pinned dependency versions:

- `revm-handler` 15.0.0 journals the caller-to-target `BalanceTransfer` in `make_call_frame` (`frame.rs:170-180`) before it invokes `precompiles.run` (`frame.rs:192`).  On precompile success, `checkpoint_commit` keeps the transfer (`frame.rs:195`).  Only a precompile error reverts it (`frame.rs:200`).
- `alloy-evm` 0.27.3 populates the field the dispatcher would need: `PrecompileInput { .., value: inputs.call_value(), .. }` (`src/precompiles.rs:472`).

So a successful `blsVerify` call with attached value both returns the verification result and permanently credits `0xb151`'s native balance.

## Why the stranded value is unrecoverable

`0xb151` has no outflow path at all.  A sweep of `crates/tn-reth` for `BLS_G1_PRECOMPILE_ADDRESS` finds only registration, docs, and value-less test calls;  nothing reads or moves the address's native balance.  The genesis account carries only the `INVALID`-opcode marker bytecode and no balance (`crates/tn-reth/src/env/genesis.rs:363-369`).  `0x7e1` at least has governance `burn` as a destruction path for stray wei;  `0xb151` has nothing, so value that lands there is inert forever.

The same asymmetry as #1190 applies: a bare value transfer with empty calldata does revert, because `dispatch` errors on short calldata (`mod.rs:134-137`, `"Invalid input: too short"`) and the journal revert returns the value.  The naive send path is already rejected.  A value-bearing `blsVerify` call is the inconsistent survivor.  (`SELFDESTRUCT` force-sends remain possible independently of this issue.)

## Impact

Low / informational.  There is no attacker profit: the sender loses funds and no party can extract them.  The realistic harm is an integration footgun.  A contract or wallet that attaches `msg.value` to a `blsVerify` call destroys those funds irrecoverably, where a Solidity contract with the same non-payable ABI would revert.  It also leaves the two custom precompiles inconsistent: after #1201, `0x7e1` enforces the payability half of Solidity call semantics and `0xb151` does not.

Unlike the TEL case there is no second-order accounting concern: `0xb151` tracks no supply or ledger, so a stray balance confuses nothing on-chain.  The cost is only the stranded wei and the semantic gap.

## Suggested fix

One guard at the top of `bls_precompile`, mirroring the payability gate #1201 added to the TEL dispatcher.  `blsVerify` is pure, so the payable list is empty:

```rust
use tn_types::U256;

const NON_PAYABLE_CALL_VALUE: &str = "call value: selector is not payable";

fn bls_precompile(input: PrecompileInput<'_>) -> PrecompileResult {
    (input.value == U256::ZERO)
        .then_some(())
        .ok_or_else(|| PrecompileError::Other(NON_PAYABLE_CALL_VALUE.into()))?;
    dispatch(input.data, input.gas)
}
```

This changes no existing behavior: no test sends nonzero value to `0xb151`.  The property harness calls in `tests/it/bls_precompile_props.rs` all go through `exec_to`, which forwards `U256::ZERO` to the value-bearing `exec_value_to` (`precompile_test_utils.rs:231-255`).  (`U256` is not yet imported in `bls_precompile/mod.rs`, hence the `use` line above.)

Scope note: the TEL dispatcher's other two gates do not transfer here.  `is_static` needs no check because `blsVerify` touches no state, and the direct-call guard from #1210 protects caller-based authorization, which this precompile does not perform.  Payability is the one missing gate.

The harness already exposes the entrypoint a regression test needs: call `exec_value_to` directly with a nonzero value.  Assert the value-bearing `blsVerify` call reverts and `0xb151`'s balance is unchanged, and keep a zero-value positive control that succeeds.

## Environment

- main `08011edf`
- `alloy-evm` 0.27.3, `revm` 34.0.0, `revm-handler` 15.0.0 (Cargo.lock)

## Credit

Reported by researcher shotes in the Cantina collaborative review, finding #21, "Precompiles do not revert value bearing calls on non-payable functions" (2026-08-26).  The finding names both precompiles against the review's pinned commit;  the TEL half was already tracked as #1190 and fixed by #1201.  This issue tracks the remaining BLS half.

MavenRain:
Closes #1286.  (Cantina #21)

## Problem

Registering an address as a precompile short-circuits the interpreter, so the `CALLVALUE` guard a compiled non-payable dispatcher would run never executes (#1190).  On the pinned `revm-handler` 15.0.0 the caller-to-target balance transfer is journaled before the precompile runs and commits when it succeeds.  The BLS precompile entrypoint at `0xb151` forwarded only `input.data` and `input.gas`, so a successful `blsVerify` call with attached value both returned the verification result and permanently credited `0xb151`.

Nothing reads or pays out that balance: no selector draws from it, and the genesis account carries only the `INVALID`-opcode marker bytecode.  Unlike `0x7e1` there is not even a governance `burn` to destroy stray wei, so value that lands there is inert forever.  A bare value send already reverts on the short-calldata check;  a value-bearing `blsVerify` call was the inconsistent survivor.  PR #1201 closed the same gap for the TEL dispatcher at `0x7e1`;  this closes the remaining BLS half of the Cantina finding.

## Fix

- [x] `crates/tn-reth/src/evm/bls_precompile/mod.rs`: reject nonzero `PrecompileInput::value` at the top of `bls_precompile`, before dispatch.  `blsVerify` is pure, so the payable list is empty.  The gate keeps #1201's fail-safe classification: a selector added later rejects value unless it is explicitly made payable.  Indirect frames cannot land wei here either way (`DELEGATECALL` value is apparent only;  `CALLCODE` transfers its explicit operand from the executing contract to itself), and rejecting them mirrors a compiled non-payable dispatcher, whose `CALLVALUE` check fires in those frames too.  The rejection surfaces as a precompile-error halt, the same mechanism as every other error from the dispatcher and identical to the TEL gate's.
- [x] Doc comment on `add_bls_precompile` pinning `DynPrecompile::new_stateful` as load-bearing for this gate: reth's engine-tree precompile cache keys entries on calldata alone and caches only pure precompiles, so `new_stateful` (`is_pure() == false`) is what guarantees a cached zero-value `Ok` can never be replayed for a value-bearing call with identical calldata.
- [x] Doc comments: a module-level payability note and a `# Payability` section on the entrypoint.

Scope note: the TEL dispatcher's other two gates do not transfer here.  `blsVerify` touches no state, so `is_static` needs no check, and the direct-call guard (#1210) protects caller-based authorization this precompile does not perform.  Payability was the one missing gate.

## Testing

- [x] New regression test `test_value_bearing_verify_rejected` (`tests/it/bls_precompile_props.rs`): a value-bearing `blsVerify` call through the real EVM path must halt with `HaltReason::PrecompileError` specifically (so a harness-level nonce/balance/gas failure cannot satisfy the test by accident), `0xb151`'s balance stays zero, and the caller keeps the wei;  a zero-value control with identical calldata verifies.  Confirmed by mutation: with the guard reverted the test fails (the call succeeds), with the guard restored it passes.
- [x] `cargo test -p tn-reth --test it bls_precompile_props` green under the pinned toolchain.  No existing test sends nonzero value to `0xb151`, so no behavior change for the existing suite.
- [x] Nightly `cargo fmt --all -- --check` green;  scoped `cargo clippy --no-deps` over the buildplan-emitted crate list.  The one pre-existing failure on main (the #1283 E0061 in the `env/rpc.rs` unit tests, fixed by open PR #1285) is untouched by this diff;  the `tn-reth` lib and `it` test targets were linted as separate invocations so that failure could not mask them.
- [x] Multi-agent review over the staged diff plus a deep semantic pass against the pinned `revm`/`alloy-evm` sources (journal/commit path, `STATICCALL`/`DELEGATECALL`/`CALLCODE` interplay, cache coupling, gas semantics);  its findings are folded into the doc comments and the hardened test assertion above.


grantkee (approved):
LGTM

PR #1298, opened 2026-08-27
fix(tn-reth): derive per-worker RPC/IPC endpoints in start_rpc (#1287)

### LINKED ISSUE #1287: multi-worker: per-worker RPC/IPC endpoints are never derived; start_rpc reuses one NodeConfig.rpc, so N workers collide (IPC silently, http/ws fatally); #558 premise is wrong
## Summary

`RethEnv::start_rpc` reads one process-wide `NodeConfig.rpc` on every call.  `ExecutionNodeInner::initialize_worker_components` calls it once per worker.  With `num_workers > 1` (the #554 to #559 track) every worker's RPC server therefore targets the same http/ws socket and the same IPC path.  Nothing in the tree derives an RPC port or an IPC path from a `worker_id`.  Issue #558 assumes the opposite ("The existing `reth_env.get_rpc_server()` allocates ports"), so the gap has no owner in the stack yet.

This is not a bug on `main` today (only `DEFAULT_WORKER_ID` is ever initialized).  It is a design gap that PR 5 (#558) has to close, and the default configuration fails silently rather than loudly, so it is worth pinning down before that PR.

## Problem

All references are `origin/main` @ 4c08db5f and reth v1.11.3 (the lockfile pin).

- `crates/tn-reth/src/env/rpc.rs:147-149`: `start_rpc` builds `self.node_config().rpc.rpc_server_config()` and starts it.  No per-call input.  `get_rpc_server` (lines 85-144) allocates no sockets.
- `crates/tn-reth/src/env/mod.rs:177`: `RethEnv` stores one clone of the `NodeConfig` at construction;  the only accessor returns `&NodeConfig`.  Nothing mutates `rpc` afterwards.
- `crates/node/src/engine/inner.rs:136-181`: `initialize_worker_components(worker_id, ..)` calls `self.reth_env.get_rpc_server(..)` (163) and `self.reth_env.start_rpc(&server).await?` (173) on the single shared `reth_env` and stores the handle in that worker's `WorkerComponents` (`workers` vec, field at line 44).  The guard at 178-180 already expects ascending `worker_id` calls, so this is the loop #557 will drive.
- `crates/node/src/manager/node/start_epoch.rs:461-462`: today `let worker_id = DEFAULT_WORKER_ID;` with the comment "only support one worker for now (with id 0) - otherwise, loop here".  The count itself is no longer hardcoded: `Committee::number_of_workers` (`crates/types/src/committee.rs:978`) returns the worker count synced from the on-chain `WorkerConfigs` contract, and `check_committee_worker_count` (`start_epoch.rs:1118-1138`) only warns when that count exceeds the one worker this version spawns.
- `crates/tn-reth/src/rpc_server_args.rs:76,100,120,124,222-234`: the defaults are `http: false`, `ws: false`, `ipcdisable: false`, `ipcpath: /tmp/tn.ipc`.  A default node serves RPC over IPC only.

What happens on the second `start_rpc` call in one process:

1. IPC first.  reth starts the IPC server before http/ws (`reth/crates/rpc/rpc-builder/src/lib.rs:1330-1335`).  On unix reth-ipc runs `std::fs::remove_file(&self.endpoint)` before it binds (`reth/crates/rpc/ipc/src/server/mod.rs:129-141`).  The second worker unlinks worker 0's socket file and binds its own.  Worker 0's listener keeps its fd (open connections survive) but no new client can reach it by path.  No error, no log above `debug!`.  With the default configuration (IPC only) the node starts cleanly and the IPC endpoint serves only the last worker's pool.  `interprocess` 2.4.0 also unlinks the path again when the newest listener drops (`ReclaimGuard`, `src/os/unix/uds_local_socket.rs:40-74`).
2. http/ws second.  With a fixed non-zero port the second bind returns `AddrInUse` (`rpc-builder/src/lib.rs:1360-1387` and `1413-1461`, each `.build(addr).await.map_err(..)?`).  `start_rpc` has no retry or fallback, `initialize_worker_components` propagates with `?`, and `run_epochs` treats the error as fatal (`crates/node/src/manager/node.rs:1124-1126`).  The node fails to start, after step 1 already unlinked worker 0's IPC socket.

`--with-unused-ports` does not close the gap.  reth's `RpcServerArgs::with_unused_ports` sets http/ws/auth ports to 0 and appends ONE random 8-char suffix to `ipcpath` at config time (`reth/crates/node/core/src/args/rpc_server.rs:748-763`).  TN calls it once per node in `RethConfig::new` (`crates/tn-reth/src/cli.rs:402-406`).  So with the flag each worker gets a fresh OS-assigned http/ws port, but all workers still share the one random IPC path and step 1 still happens.  The flag is also a test lever: it `conflicts_with = "instance"` (`crates/telcoin-network-cli/src/node.rs:72-73`) and leaves operators without predictable ports.

reth's `--instance` mechanism is per process, not per worker: `adjust_instance_ports` shifts `http_port -= instance - 1`, `ws_port += instance * 2 - 2`, `auth_port += instance * 100 - 100`, `ipcpath += "-{instance}"` (`rpc_server.rs:711-723`), applied once to the node's `NodeConfig` (`node_config.rs:486-490`).

Tests do not exercise this path: e2e tests spawn one OS process per validator with caller-chosen `--ipcpath` and probed ports (`crates/e2e-tests/tests/it/common.rs:428-450`, `crates/e2e-tests/src/lib.rs:311`), and temp chains disable IPC since the #1165 fix (`crates/tn-reth/src/env/genesis.rs:109`).

## Impact

- `main` today: none (single worker).
- Multi-worker (#557/#558 as specified): default configuration starts and silently serves IPC for one worker only;  any fixed http/ws port makes the node fail to start.  Both are node-local;  no consensus or attacker-profit angle.
- #558's text "ensure it assigns distinct ports for each call" has no mechanism to attach to: `get_rpc_server` assigns nothing.

## Fix

Give `start_rpc` a per-worker `RpcServerArgs` instead of the shared one.  Suggested shape, for the #558 PR or a small precursor:

1. `RethEnv::start_rpc(&self, server, worker_id)` (or a `rpc_server_config_for(worker_id)` helper) derives the worker's endpoints from the operator's `NodeConfig.rpc`:
   - worker 0 keeps the operator's values unchanged (backward compatible, faucet and existing tooling stay on worker 0);
   - worker `w > 0`: reuse reth's instance arithmetic keyed on `w` (`http_port - w`, `ws_port + 2w`, `ipcpath + "-w{w}"`), or an explicit per-worker list in config.  If `--instance` stays supported the two offsets must not overlap (instance `i` worker `w` vs instance `i + w` worker 0), so stride the worker offset past the instance range or keep workers on a separate band.
   - under `--with-unused-ports`: http/ws already get port 0;  re-randomize (or disable) the IPC path per worker.
2. Log the resolved endpoints per worker at `info!` at start-up;  `RpcNodeInfo` carries no endpoints (`crates/execution/tn-rpc/src/lib.rs:14-41`), so no wire change is needed.
3. Regression test: build two `RpcServerConfig`s for workers 0 and 1 from one `RpcServerArgs` and assert distinct http/ws `SocketAddr`s and distinct IPC paths.  A second test can start two servers on one `RethEnv` and assert both IPC sockets exist afterwards.
4. Correct the #558 text: the port allocation is TN's to do, reth does not do it.

Operator docs should state the per-worker port map once the scheme is fixed.

## Credit

Surfaced by researcher shotes in the Cantina collaborative review (comment 2026-08-19, filed as Informational finding #23 on 2026-08-26) on `RethConfig::new`;  the IPC unlink behavior and the `--with-unused-ports` single-random-path detail were confirmed against reth v1.11.3 while triaging it.

Related: #554, #557, #558, #1165, #1172.

MavenRain:
Closes #1287.  (Cantina #23)

## Problem

`RethEnv::start_rpc` read the one process-wide `NodeConfig.rpc` on every call, and `ExecutionNodeInner::initialize_worker_components` calls it once per worker.  With more than one worker (the #554 to #559 track) every worker's RPC server targeted the same http/ws socket and the same IPC path:

- IPC first: reth unlinks the endpoint path before it binds, so the second worker silently stole worker 0's socket file.  The default configuration (IPC only) started cleanly and served only the last worker's pool, with nothing above `debug!`.
- http/ws second: a fixed non-zero port made the second bind fail with `AddrInUse`, which `initialize_worker_components` propagates as a fatal startup error, after the IPC steal already happened.

`--with-unused-ports` did not close the gap: it randomizes the IPC path once per process, so every worker still shared it.  No attacker is required for either failure;  the default configuration failed silently rather than loudly.

Not a bug on `main` today (only `DEFAULT_WORKER_ID` is ever initialized).  This is the design gap #558 has to close, landed as its precursor.

## Fix

- [x] `crates/tn-reth/src/env/rpc.rs`: `start_rpc` takes the `worker_id` and derives that worker's `RpcServerArgs` from the operator's config (`worker_rpc_server_args`) instead of reusing the shared one:
  - worker 0 keeps the operator's values unchanged, so single-worker nodes and existing tooling see exactly the configured endpoints;
  - worker `w > 0` shifts an enabled http port down by `200 * w`, shifts an enabled ws port up by `400 * w`, and suffixes an enabled IPC path with `-w{w}`;
  - the stride is derived from reth's real `--instance` bound, not guessed: instance offsets span at most 199 ports (`instance <= 200` in reth v1.11.3), so a 200-port worker stride keeps every `(instance, worker)` pair on a distinct port, and the `-w{w}` IPC suffix is disjoint from instance's `-{i}` by construction;
  - port 0 (the `--with-unused-ports` sentinel) passes through, since the OS assigns a distinct port per bind, and the flag's one random IPC path still gets the per-worker suffix;
  - a shift that leaves the valid range fails startup loudly (`TnRethError::WorkerRpcPort`) instead of wrapping into a port another worker or instance owns;  a derived http port additionally stops above the privileged range (below 1024 a non-root process cannot bind), so that failure also names the worker, base, and offset instead of dying later in the bind with a bare permission error;
  - when both transports are enabled on fixed ports the derivation requires `ws_port >= http_port` (reth's default layout;  equality is the shared http+ws server):  http bands stride down and ws bands stride up, so inverted bases would let one worker's http band land on another worker's ws band;  the first derived worker fails loudly (`TnRethError::WorkerRpcPortOrder`) and worker 0 still binds inverted bases as configured, keeping single-worker nodes working;
  - a disabled transport keeps its configured value and can never error.
- [x] `crates/tn-reth/src/env/rpc.rs`: the resolved endpoints (http/ws/ipc) are logged per worker at `info!` once the server starts.
- [x] `crates/node/src/engine/inner.rs`: `initialize_worker_components` passes its `worker_id` to `start_rpc`.
- [x] `crates/tn-reth/src/error.rs`: new `WorkerRpcPort` and `WorkerRpcPortOrder` variants carrying the worker id (`WorkerId`), the transport, and the offending ports and offset.

No wire change: `RpcNodeInfo` carries no endpoints.

## Testing

- `test_worker_zero_keeps_operator_rpc_endpoints`: worker 0's derived args equal the operator's.
- `test_worker_endpoints_are_distinct_per_worker`: workers 1 and 2 derive distinct http/ws ports and IPC paths with the exact band arithmetic.
- `test_worker_bands_clear_the_instance_range`: the closest approach between the worker and instance schemes (worker 1 of instance 1 against worker 0 of instance 200) stays separated on both transports.
- `test_out_of_range_worker_port_is_a_loud_error`: an enabled http port below the band and an enabled ws port at `u16::MAX` both fail with `WorkerRpcPort`, not a wrapped port.
- `test_zero_ports_stay_os_assigned_and_ipc_still_suffixes`: `--with-unused-ports` semantics survive derivation.
- `test_inverted_transport_bases_are_a_loud_error`: fixed ws below fixed http fails the first derived worker with `WorkerRpcPortOrder`;  worker 0 keeps the inverted config as given.
- `test_equal_transport_bases_stay_valid`: reth's shared http+ws server (equal ports) derives non-colliding split bands.
- `test_http_band_below_the_privileged_floor_is_a_loud_error`: a derived http port under 1024 fails at derivation time, not at bind time.
- `test_disabled_transports_keep_their_configured_values`: a disabled transport never shifts and never errors;  disabled IPC keeps its path.
- `test_two_workers_bind_distinct_live_ipc_sockets` (unix): two workers started on one `RethEnv` leave both IPC socket files on disk;  before this change the second start unlinked worker 0's.

Ran nightly `cargo fmt -- --check`, `cargo check --workspace --all-targets`, and `cargo nextest run -p tn-reth` on the rpc tests locally.  Each new load-bearing test was seen to fail with the derivation reverted to the shared config and pass with it restored.


grantkee (dismissed):
LGTM!

grantkee (approved):

PR #1299, opened 2026-08-27
fix(network): split kad PutRecord shed threshold from the penalty threshold (#1289)

### LINKED ISSUE #1289: Harden kad PutRecord rate limiter: split shed threshold from penalty threshold, correct rationale, add a metric
## Context

The per-source inbound kad `PutRecord` rate limiter (`peers/manager.rs` `put_record_rate_limited`, checked at `consensus.rs:2019`) has three related weaknesses in how it sizes the drop threshold and applies penalties.  One change resolves all three, plus the requested metric.

None of this is exploitable adversarially.  The source is the authenticated libp2p sender, so a flooder bans only itself and evades with a fresh key.  The risk is the honest case: a non-committee peer doing its DHT duty can be shed and then temporarily banned once the validator set grows.

### Finding 1: the threshold is modeled on the wrong traffic

`MAX_PUT_RECORDS_PER_WINDOW = 30` was sized against a source republishing its own record.  The dominant inbound traffic is kad replication: on every `record_replication_interval` tick (libp2p default 1h) the source re-puts every stored record whose publisher is not the source, one iterative query per record, each fanning out to `replication_factor` (20) closest peers.  The count one target sees from one source per tick is:

```
count(N, R, k)  =  N * min(1, k / R)        where k = replication_factor = 20
```

`N` is the source's stored-record count, `R` is the peer set the source reaches.  With healthy routing this is ~20, below 30.  When reach is pinned near 30 (a common kad state), the crossing point is `0.67 * N > 30`, so N ~= 45 records.  With reach <= k (bootstrap, partition heal, churn), `count = N` and it crosses at N = 31.

This does not fire on today's 5-authority testnet (~5 messages per tick, 6x margin).  It is a latent defect that activates as the validator set grows toward the stated ~100 target.  It should be fixed before the set grows past ~40.

The doc comment at `manager.rs:39-47` states the wrong model ("small handful of records per minute" counts only the one record that replication excludes) and will mislead the next person to tune the constant.

### Finding 2: one honest burst becomes a temporary ban

`PutRecordWindow` has no "already penalized this window" state, so `process_penalty` fires on every message past the threshold.  Escalation for a score-0 peer:

| Quantity | Value |
|---|---|
| First penalized message | #31 (`count > 30`) |
| `Penalty::Medium` | `add(-5.0)` |
| `min_score_before_disconnect` | -20.0 |
| `min_score_before_ban` | -50.0 |

- msg 34 (-20) -> disconnect plus a 10 minute temporary ban;  reconnects are refused.
- msg 40 (-50) -> full ban path.

Decay does not offset a burst: the penalty path does not decay, and heartbeat decay has a 300s half-life.  So a single benign replication burst above the threshold walks an honest peer to a ~30 minute ban, potentially every hour once Finding 1 puts the burst over the line.

Two consequences stated in the original scan are wrong and are recorded here for accuracy: the ban is peer-id-scoped not IP-scoped (`BANNED_PEERS_PER_IP_THRESHOLD = 1`, a single banned peer's IP is not blocked), and the ban is temporary (~30 min) not outright.

### No metric on the rate-limit trip (consensus.rs:2020)

Requested in review.  The shed/penalty path emits no counter, so operators cannot see honest shedding building up before it becomes a ban.

### (a) Correct the rationale comment (land regardless)

```rust
/// Maximum inbound kad `PutRecord` messages accepted from a single source per
/// [`PUT_RECORD_RATE_WINDOW`] before the source is rate limited.
///
/// Sized against kad *replication* fan-in, not against a source republishing its own
/// record. On every `record_replication_interval` tick (libp2p default 1h) libp2p-kad
/// re-puts every stored record whose publisher is **not** the local node
/// (`libp2p-kad::jobs::PutRecordJob::poll`), one iterative query per record, each fanning
/// out to the `replication_factor` (20) closest peers. The count one target sees from one
/// source per tick is therefore
///
///     N * min(1, replication_factor / R)
///
/// where `N` is the source's stored-record count (~one row per node seen, capped by
/// `MemoryStoreConfig::max_records` = 1024) and `R` is the peer set the source's lookup
/// reaches. With healthy routing (`R >= replication_factor`) this is ~20 and Poisson
/// distributed; when a source's reach is small (bootstrap, partition heal, post-restart)
/// it degrades toward `N`.
```

### (b) Split the drop threshold from the penalty threshold (resolves Findings 1 and 2)

Dropping a record is nearly free: kad re-replicates hourly and the same record arrives from up to `replication_factor` other peers, so a shed record is redundant.  Penalizing an honest source costs a ban.  So shed aggressively, score conservatively.

```rust
// crates/network-libp2p/src/peers/manager.rs

/// Inbound kad `PutRecord` messages accepted from one source per window before records
/// are shed. Bounds the BLS verify + store write on the network task.
const MAX_PUT_RECORDS_PER_WINDOW: usize = 30;

/// Sustained inbound rate above which a source is *scored*, not merely shed.
///
/// Deliberately far above [`MAX_PUT_RECORDS_PER_WINDOW`]: shedding a redundant record
/// costs nothing, whereas penalizing an honest source costs a ban, and honest kad
/// replication fan-in scales with network size (see above). A source exceeding this is
/// treated as a flood, and then at most once per window (GHSA-f6rq-62rr-4h9g).
const PUT_RECORD_PENALTY_THRESHOLD: usize = 512;

struct PutRecordWindow {
    count: usize,
    started: Instant,
    /// Whether a penalty was already assessed for this window (one per window, not per message).
    penalized: bool,
}

/// Outcome of recording an inbound `PutRecord` against the per-source window.
pub(crate) enum PutRecordRate {
    /// Under the drop threshold - process the record.
    Allowed,
    /// Over the drop threshold - shed the record, no penalty.
    Shed,
    /// Over the penalty threshold and not yet penalized this window - shed and penalize once.
    Flooding,
}

pub(crate) fn put_record_rate_limited(&mut self, source: PeerId) -> PutRecordRate {
    if self.is_local_peer(&source) {
        return PutRecordRate::Allowed;
    }
    let now = Instant::now();
    let window = self
        .put_record_windows
        .entry(source)
        .or_insert(PutRecordWindow { count: 0, started: now, penalized: false });
    if now.duration_since(window.started) >= PUT_RECORD_RATE_WINDOW {
        window.count = 1;
        window.started = now;
        window.penalized = false;
    } else {
        window.count += 1;
    }

    if window.count <= MAX_PUT_RECORDS_PER_WINDOW {
        PutRecordRate::Allowed
    } else if window.count > PUT_RECORD_PENALTY_THRESHOLD && !window.penalized {
        window.penalized = true;
        PutRecordRate::Flooding
    } else {
        PutRecordRate::Shed
    }
}
```

```rust
// crates/network-libp2p/src/consensus.rs:2019
match self.swarm.behaviour_mut().peer_manager.put_record_rate_limited(source) {
    PutRecordRate::Flooding => {
        debug!(target: "network-kad", ?source, "put record flood: penalizing source");
        self.swarm.behaviour_mut().peer_manager.record_put_record_rate_limited();
        self.swarm.behaviour_mut().peer_manager.process_penalty(source, Penalty::Medium);
    }
    PutRecordRate::Shed => {
        trace!(target: "network-kad", ?source, "shedding rate limited put request");
        self.swarm.behaviour_mut().peer_manager.record_put_record_rate_limited();
    }
    PutRecordRate::Allowed => {
        if let Some((key, value)) = self.peer_record_valid(&record) {
            // ... existing validate/store path unchanged ...
        }
    }
}
```

This keeps the starvation bound where the PR put it (30 verifies per source-minute, so ~39 peers x 30 x ~1ms ~= 1.2s/min ~= 2% duty cycle) while removing the honest-peer ban.  A genuine flooder is still shed from message 31 and still accrues penalties, on a ~10 minute-to-ban timescale appropriate to a peer whose traffic is already discarded.

Note on the alternative of simply raising the single threshold: the drop threshold is the starvation bound.  Raising it to 256 puts the worst case at ~39 x 256 x ~1ms ~= 10s/min ~= 17% duty cycle.  That is why the two thresholds are split rather than the one raised.

### (c) Pin the replication interval explicitly

The cadence driving all of this is an inherited libp2p default that no one chose.  Make it a deliberate config value alongside the publication interval set at `consensus.rs:443`:

```rust
kad_config.set_replication_interval(Some(libp2p.kad_replication_interval));
```

### Metric

`record_put_record_rate_limited()` in the sketch above is the requested counter;  wire it into the network metrics registry so operators see shed and flood events.  Emit it on both the `Shed` and `Flooding` arms.

### Optional refinement (needs a human decision)

A derived threshold auto-scales with the network: since all nodes converge to approximately the same record set, the local kad store length is a good proxy for the source's.  `limit = max(30, kad_store_len * 2)`, refreshed on the existing 30s heartbeat.  Cleaner in principle but adds plumbing between `KadStore` and `PeerManager`;  the drop/penalty split in (b) already removes the harmful consequence.  Recommend (b) now, consider the derived form if operational data later shows meaningful honest shedding.

## Testing

- Extend `crates/network-libp2p/src/tests/peer_manager.rs`: assert `Allowed` up to 30, `Shed` from 31 to 512, `Flooding` exactly once past 512 then `Shed` again in the same window, and window reset clears `penalized`.
- Metric assertion: shed and flood arms each bump the counter once.

MavenRain:
Closes #1289.

## Problem

Inbound Kademlia replication can legitimately exceed the existing 30-record per-source minute limit. A replication tick can send every record in a source's store to a reachable target, up to the current 1,024-record store cap. Penalizing every dropped record can therefore disconnect honest peers during bootstrap, partition recovery, or network growth.

The expensive-work bound must remain cheap-first: count the authenticated source's messages before signature verification and storage. Raising that bound would increase the work an unbanned source can impose on the consensus network task.

## Changes

- [x] **Separate shedding from scoring.** `PutRecordRate` distinguishes accepted records, unscored shedding, and flooding. The first 30 records per source in a 60-second tumbling window remain eligible for verification and storage.
- [x] **Derive the penalty threshold from honest fan-in.** The threshold is twice the 1,024-record store ceiling, with a test pinning `MemoryStoreConfig::default().max_records`. Messages 31 through 2,048 are shed without scoring; message 2,049 applies one `Penalty::Severe` (-10). Further messages are shed without additional penalties through message 16,384.
- [x] **Escalate sustained floods.** Above eight times the penalty threshold (16,384), every subsequent classified message returns `Flooding`, driving non-exempt sources through the existing severe-penalty disconnect and ban handling. This preserves the original 30-record expensive-work bound and does not depend on successfully verifying attacker-provided records.
- [x] **Preserve active over-limit windows across reconnects.** Disconnects retain windows above the shed threshold, including their penalty latch, until the original 60-second window expires. Heartbeat maintenance reclaims expired entries even if the source never returns. Below-limit disconnects still discard their windows.
- [x] **Make Kademlia limits explicit and validate them at startup.** Pin the replication interval (default 1 hour) and encoded packet size (16 KiB). Both configuration funnels validate nonzero replication/publication intervals strictly below the record TTL, returning field-named errors before constructing either critical swarm. Documentation distinguishes the zero-replication panic from record-expiration problems.
- [x] **Expose shedding and escalation.** `tn_network.put_records_rate_limited_total` records `shed` and `flood` outcomes inside the classifier; `flood` includes repeated escalation above the higher ceiling.
- [x] **Test the actual scoring call site.** A connected, tracked peer with no committee or allowlist exemption relays a valid record. Tests assert unchanged scores throughout shedding, one exact severe penalty, the inclusive higher boundary, and a real disconnect request after crossing it. Existing short-TTL fixtures now supply a valid replication cadence.

The earlier description's ten-minute ban estimate was incorrect: with default decay, one Medium penalty per minute approaches approximately -38.6, crossing the disconnect threshold but never reaching the -50 score-based ban threshold. This revision uses Severe penalties plus repeated escalation above the higher ceiling. Peer exemptions and the existing reputation/decay rules still apply; it is not a global bandwidth bound or protection against fresh identities.

## Testing

Under the pinned Rust 1.94 toolchain, the following selection passed twice, including after exact
restoration of every mutant: 7 config tests and 16 network tests, 23 total, with no replayed results.

```sh
testledger --force --ledger .validation/testledger -- cargo +1.94 test --locked \
  -p tn-config -p tn-network-libp2p --lib -j 2 --target-dir .validation/target -- \
  libp2p_config test_put_record_ test_kad_put_ \
  test_killed_peer_record_expires_local_record_survives
```

- Mutation-confirmed: disabling cadence guards, accepting TTL equality, lowering the penalty
  threshold below the honest ceiling, swapping the call-site arms, removing its penalty, resetting
  windows on disconnect, suppressing repeated flood penalties, and disabling expiry cleanup all
  produced the expected test assertion failures. Every mutant compiled and was restored exactly.
- `cargo +nightly-2026-03-20 fmt --all -- --check` passed through `gateledger`.
- Independent strict review found no production defects or CI weakening. `ctxcat-review` was
  unavailable, so the documented strict-review fallback was used.
- Pinned nightly Clippy passed with default and all features for all 23 packages selected by
  `buildplan`. The broader `--all-targets --keep-going` pass reported 130 pre-existing test lint
  diagnostics; an independent source-to-diff audit confirmed that none intersects this change.
  No lint allowances or CI settings were changed. CI's Clippy commands use default target settings.

The default-feature Clippy command below passed through `gateledger`; the same command with
`--all-features` also passed:

```sh
cargo +nightly-2026-03-20 clippy --locked \
  -p telcoin-network -p telcoin-network-cli -p tn-batch-builder -p tn-batch-validator \
  -p tn-config -p tn-engine -p tn-executor -p tn-exex -p tn-network-libp2p -p tn-node \
  -p tn-primary -p tn-reth -p tn-rpc -p tn-storage -p tn-test-utils \
  -p tn-test-utils-committee -p tn-types -p tn-worker -p tn-worker-gateway \
  -j 2 --target-dir .validation/target --no-deps -- -D warnings
```

The repository skips compile/test CI lanes for maintainer-authored PRs. The new commit requires the separate pinned `make attest` workflow before merge; local regression checks do not submit an on-chain attestation.


crates/config/src/network.rs
```diff
@@ -153,6 +153,17 @@ pub struct LibP2pConfig {
...
+    /// `replication_factor` closest peers.
+    ///
+    /// This cadence drives the dominant inbound `PutRecord` fan-in each node sees from
+    /// each peer (see `MAX_PUT_RECORDS_PER_WINDOW` in network-libp2p). Pinned explicitly
+    /// so the value is a deliberate choice rather than an inherited libp2p default; the
+    /// default matches the libp2p default (1h).
+    ///
+    /// Must be nonzero and < `kad_record_ttl`, otherwise replication either spins the
+    /// event loop or lets other publishers' records expire before they are re-replicated.
+    pub kad_replication_interval: Duration,
```
grantkee: A kad_replication_interval of exactly zero panics the network task inside libp2p-kad on the first swarm poll, and nothing validates the field.

This is the first time the knob is operator-reachable — `.set_replication_interval(...)` at `consensus.rs:456` did not exist on `main`, and the field deserializes straight from `<datadir>/network-config` through `NetworkConfig::read_config` -> `load_from_path_or_default` (`traits.rs:56-74`, called at `crates/node/src/manager/node.rs:744`). `Some(ZERO)` is materially different from `None` here: at `libp2p-kad-0.48.0/src/behaviour.rs:479-489`, `None` falls back to the 12h publication interval, while `Some(ZERO)` is taken verbatim as `PutRecordJob.inner.interval`.

The failure is a hard `assert!`, not a `debug_assert!` — `libp2p-kad-0.48.0/src/jobs.rs:251`:

```rust
assert!(!self.inner.check_ready(cx, now));
```

`PutRecordJob::new` (`jobs.rs:152-166`) sets `deadline = now + ZERO == now`. The first `Behaviour::poll` reaches `check_ready` (`jobs.rs:113-120`), which returns `true` because `now >= *deadline` holds for equality; the snapshot drains; `jobs.rs:248` re-arms with `deadline == now`; and `jobs.rs:251` asserts on a `check_ready` that returns `true` again. Both arms of `jobs.rs:115` fire — a `Delay::new(ZERO)` also resolves `Ready`. With an empty store it panics on the very first poll, because the loop body never runs.

Blast radius is a fail-fast whole-node shutdown rather than a silent swarm death: `node.rs:1019` spawns `primary_network.run()` via `spawn_critical_task`, `TaskHandle`'s `Future` impl (`types/src/task_manager.rs:117-121`) maps the `JoinError` to `CriticalJoinError`, and `until_exit_internal` (`task_manager.rs:472-489`) calls `shutdown.notify()`. Same for the worker network at `node.rs:1068`. So the cost is a libp2p-internal backtrace instead of a field-named config error.

Two notes on the doc comment directly above this field. "Spins the event loop" isn't the failure mode — it panics. And the two clauses have different severities: only *exactly* zero panics, while `>= kad_record_ttl` merely degrades replication (sub-millisecond non-zero values are racy rather than deterministic, since `deadline = now + 1ns > now` and the `futures_timer` delay hasn't fired). Worth splitting those apart in the wording.

There is no validation anywhere: `impl LibP2pConfig` (`network.rs:179-201`) has only the gossip-topic helpers, and `grep "is_zero()\|Duration::ZERO" crates/config/src crates/network-libp2p/src` returns nothing. The sibling precedent is exact — `ScoreConfig::validate()` is called at `crates/config/src/consensus.rs:195` and again on the boot path at `crates/node/src/manager/node.rs:1000`, eight lines above the `ConsensusNetwork::new_for_primary` call that would install the bad kad config. Its own doc comment (`network.rs:539-541`) states the rationale verbatim: "fails fast with a field-named error rather than halting the running swarm."

Suggested fix, in `impl LibP2pConfig`:

```rust
/// Reject a libp2p config whose kad cadences would panic or silently break record persistence.
///
/// `kad_replication_interval` is handed to libp2p's `PutRecordJob` as its tick interval. A zero
/// interval makes the job re-arm with `deadline == now`, and the unconditional
/// `assert!(!self.inner.check_ready(cx, now))` in libp2p-kad's `jobs.rs` fires on the first swarm
/// poll, panicking the critical network task. Validate at startup, on the same footing as
/// `ScoreConfig::validate`, so a bad cadence fails with a field-named error instead of a
/// libp2p-internal backtrace.
pub fn validate(&self) -> eyre::Result<()> {
    eyre::ensure!(
        !self.kad_replication_interval.is_zero(),
        "LibP2pConfig.kad_replication_interval must be non-zero",
    );
    eyre::ensure!(
        self.kad_replication_interval < self.kad_record_ttl,
        "LibP2pConfig.kad_replication_interval ({:?}) must be < kad_record_ttl ({:?}) \
         or records expire before they are re-replicated",
        self.kad_replication_interval,
        self.kad_record_ttl,
    );
    eyre::ensure!(
        !self.kad_publication_interval.is_zero()
            && self.kad_publication_interval < self.kad_record_ttl,
        "LibP2pConfig.kad_publication_interval ({:?}) must be non-zero and < kad_record_ttl ({:?})",
        self.kad_publication_interval,
        self.kad_record_ttl,
    );
    Ok(())
}
```

Called at both existing funnels beside the score-config guard (`config/src/consensus.rs:195` and `node/src/manager/node.rs:1000`), and mirroring the `ScoreConfig` test pattern at `network.rs:735-737` with a default-validates case plus a zero-interval rejection.

One repro note if you want to reproduce it: `kad_replication_interval: 0s` is not valid YAML for this field, because stock `serde` `Duration` is used. The repro value is `kad_replication_interval: {secs: 0, nanos: 0}`.

crates/network-libp2p/src/consensus.rs
```diff
@@ -2040,43 +2041,53 @@ where
...
-                // version. Log only; no penalty.
-                trace!(target: "network-kad", ?source, "ignoring stale but valid kad record");
-            }
-        } else {
-            warn!(target: "network-kad", "Received invalid peer record!");
+        // returned above, so this bounds the unbanned population. Honest kad replication
+        // fan-in can cross the shed threshold as the network grows, so shedding carries no
+        // penalty (a shed record is redundant: up to `replication_factor` other peers re-put
+        // it hourly); only a source past the flood threshold is scored, once per window.
+        match self.swarm.behaviour_mut().peer_manager.put_record_rate_limited(source) {
```
grantkee, line 2057: No test covers this match, so transposing the Shed and Flooding arm bodies leaves the whole suite green.

The mutation-confirmation list in the description covers three classifier-level mutations — reverting the penalty latch, conflating the two thresholds, dropping the reset of `penalized` — and each of those does fail a test in `tests/peer_manager.rs`. The one that isn't in that set is the call-site mutation: swapping the bodies of the `Shed` and `Flooding` arms right here. It type-checks (both arms are `()`, both bind `source`) and nothing fails.

The reason is that all eight tests touching this path assert only the returned enum, the window struct, or the counter — never a score or a penalty. `peer_manager.rs:2182` checks `Allowed` x30 then `Shed`; `:2196` is named `test_put_record_rate_limit_penalizes_flood_once_per_window` but checks no penalty; `:2219`, `:2235`, `:2254`, `:2278` (`shed == 482`, `flood == 1`) and `metrics.rs:217` likewise. And `git diff origin/main...HEAD --stat` shows `tests/network_tests.rs` untouched on this branch, so the four existing `process_kad_put_request` call sites are unchanged: `network_tests.rs:2801`, `:3000`, `:3044`, `:3171` each invoke it exactly once with no loop, so three reach this match with `window.count == 1` -> `Allowed` and the fourth (`:3044`, `publisher: None`) returns at `consensus.rs:2036` before the limiter. Full-swarm tests can't cover it incidentally either, since kad only re-puts on the 1h replication interval.

Worth closing because the undetected regression wouldn't be cosmetic. The once-per-window guard lives only in the `Flooding` classifier arm (`manager.rs:977-980` is what sets `penalized`), so a swapped `Shed` arm penalizes on *every* shed message — an honest source at 41 puts/window would walk to `min_score_before_ban = -50.0` (`config/src/network.rs:507`) at `Penalty::Medium = -5.0` (`score.rs:91`). That's the exact outcome this branch exists to prevent, reintroduced by a transposition that no test sees.

One wiring test in `network_tests.rs`, beside the existing `process_kad_put_request` tests, closes it. It needs the two constants visible to the consensus test module — they're module-private at `manager.rs:60` and `manager.rs:70`, so both would become `pub(crate) const`:

```rust
/// Wiring regression (#1289): the shed arm of `process_kad_put_request` must not score the
/// source, and the flood arm must score it exactly once. Guards against the two arm bodies being
/// transposed, which the classifier-level tests in `tests/peer_manager.rs` cannot see.
#[tokio::test]
async fn test_kad_put_shed_is_unscored_and_flood_is_penalized() -> eyre::Result<()> {
    use crate::peers::manager::{MAX_PUT_RECORDS_PER_WINDOW, PUT_RECORD_PENALTY_THRESHOLD};

    let TestTypes { peer1, peer2, .. } =
        create_test_types::<TestWorkerRequest, TestWorkerResponse>();
    let mut network = peer1.network;
    let source = *peer2.network.swarm.local_peer_id();
    let record = peer2.network.get_peer_record();

    // shed range: redundant records are dropped, the source is NOT scored
    for _ in 0..=MAX_PUT_RECORDS_PER_WINDOW + 10 {
        network.process_kad_put_request(source, record.clone())?;
    }
    let shed_score = network.swarm.behaviour().peer_manager.peer_score(&source);
    assert!(
        shed_score.is_none_or(|s| s >= 0.0),
        "shedding must never penalize an honest source (#1289), got {shed_score:?}"
    );
    assert!(!network.swarm.behaviour().peer_manager.peer_banned(&source));

    // cross the penalty threshold: the flood arm scores the source, once per window
    for _ in 0..PUT_RECORD_PENALTY_THRESHOLD {
        network.process_kad_put_request(source, record.clone())?;
    }
    let flood_score =
        network.swarm.behaviour().peer_manager.peer_score(&source).expect("source now tracked");
    assert!(flood_score < 0.0, "a source past the penalty threshold must be scored");

    // further messages in the same window do not re-penalize
    for _ in 0..50 {
        network.process_kad_put_request(source, record.clone())?;
    }
    assert_eq!(
        network.swarm.behaviour().peer_manager.peer_score(&source),
        Some(flood_score),
        "the flood penalty is assessed at most once per window"
    );

    Ok(())
}
```

Transposing the arms fails the first assertion; dropping the penalty from `Flooding` fails the second. Note this touches a file the PR doesn't currently change.

crates/network-libp2p/src/peers/manager.rs
```diff
@@ -37,21 +37,59 @@ mod peer_manager;
...
 
+/// Sustained inbound rate above which a source is scored, not merely shed.
+///
+/// Deliberately far above [`MAX_PUT_RECORDS_PER_WINDOW`]: shedding a redundant record
+/// costs nothing (kad re-replicates hourly, so the same record arrives from up to
+/// `replication_factor` other peers), whereas penalizing an honest source walks it to a
+/// temporary ban. Honest replication fan-in also scales with network size (see above). A
+/// source exceeding this is treated as a flood and penalized, at most once per window
+/// (GHSA-f6rq-62rr-4h9g).
+const PUT_RECORD_PENALTY_THRESHOLD: usize = 512;
```
grantkee: This threshold sits below the honest ceiling the store cap allows, so one replication tick from a source with a full store can cross it.

The description sizes the fan-in against the ~100-validator target, where `N` is around 100 and 512 has comfortable headroom — that reasoning holds. The gap is that 512 is chosen independently of the thing that actually bounds `N`, so the two can drift: `KadStore::new` adopts `MemoryStoreConfig::default()` verbatim (`kad.rs:223`) and enforces the cap at `kad.rs:462-467`, and that default is `max_records = 1024` (`libp2p-kad-0.48.0/src/record/store/memory.rs:66`). The provable honest ceiling is therefore double this constant.

Two things make the `R <= replication_factor` branch of the fan-in model the normal case here rather than an edge case. `replication_factor` is never set by the repo, so it stays at `K_VALUE = 20`. And the kad routing table is populated only from live connections — `add_address` at `consensus.rs:1727` under `PeerConnected`, with `remove_peer` at `:1647`, `:1703`, `:1749` — under `BucketInserts::Manual` (`consensus.rs:449`), while connected peers are capped at `max_peers() = 39` (`config/src/network.rs:406`, `:412`, `:427`). So `R <= 39` always and `R <= 20` at or below target, which puts the `min(1, 20/R)` multiplier at 1 and sends every stored record to every connected peer.

Upstream has no pacing that would spread that across windows. `jobs.rs:238-245` returns records back-to-back from a snapshot `vec::IntoIter` with no timer and no per-tick cap. `JOBS_MAX_QUERIES = 100` and `JOBS_MAX_NEW_QUERIES = 10` (`jobs.rs:80-83`) are applied per `Behaviour::poll` invocation (`behaviour.rs:2553`, `:2570-2575`), and `libp2p-swarm-0.47.1` re-enters `behaviour.poll` in an unbudgeted loop while it returns `Ready` (`libp2p-swarm/src/lib.rs:1201-1240`). The only real limit is 100 concurrent queries, so at a pessimistic 5s mean query time that's `100 * 60 / 5 = 1200 > 1024` — the whole store drains inside one window unless queries average over 6s.

Worth being clear about how much this actually costs today, because it's less than it first looks: one tick delivers at most `N` messages to a given target in total, and `1024 < 2 * 513`, so a single tick can produce at most *one* `Penalty::Medium`. The next tick is an hour later, by which point `-5` has decayed by `2^(-3600/300)` to `-0.0012`. Committee members and operator-allowlisted peers bypass the score model entirely (`all_peers.rs:856-876`, `peer.rs:292-312`), so only observers and nvv are exposed at all. Net effect is a spurious `-5.0` at most hourly on a non-exempt honest peer, plus a misleading `Flooding` metric and `debug!` line — not the ban walk. The reason to fix it is that the constant silently stops covering the ceiling as the network grows, and nothing fails when it does.

Deriving it from the cap keeps the two from drifting:

```rust
/// Upper bound on records a source can hold, and therefore on the `PutRecord` messages one honest
/// replication tick can deliver to a single target.
///
/// Mirrors `MemoryStoreConfig::default().max_records`, which `KadStore::new` adopts verbatim (see
/// `crates/network-libp2p/src/kad.rs`). Pinned by `test_penalty_threshold_above_honest_ceiling`.
const KAD_MAX_STORED_RECORDS: usize = 1024;

/// Sustained inbound rate above which a source is scored, not merely shed.
///
/// Must stay strictly ABOVE the honest ceiling. One replication tick drains the source's whole
/// store with no pacing: `libp2p-kad::jobs::PutRecordJob::poll` yields records back-to-back from a
/// snapshot, `Behaviour::poll` starts up to `JOBS_MAX_NEW_QUERIES` (10) per invocation bounded only
/// by `JOBS_MAX_QUERIES` (100) concurrent queries, and `libp2p-swarm` re-polls the behaviour
/// without a budget. A source whose routing reach is <= `replication_factor` therefore replicates
/// every stored record to every reachable peer and can legitimately deliver
/// `KAD_MAX_STORED_RECORDS` messages to one target inside a single window.
const PUT_RECORD_PENALTY_THRESHOLD: usize = 2 * KAD_MAX_STORED_RECORDS;
```

Plus a test pinning `PUT_RECORD_PENALTY_THRESHOLD > KAD_MAX_STORED_RECORDS` and one asserting `MemoryStoreConfig::default().max_records == KAD_MAX_STORED_RECORDS`, so a future libp2p bump that raises the store cap fails loudly instead of quietly reopening the gap. This leaves the shed behaviour (30/window, unpenalized) untouched, so the GHSA-f6rq-62rr-4h9g starvation bound is unaffected — only the scoring trigger moves out of honest reach.

Raising it does widen the unpenalized band, so it pairs with the separate comment on the classifier below about the free band and the missing hard cutoff — the two are opposite-direction and are best decided together.

crates/network-libp2p/src/peers/manager.rs
```diff
@@ -907,31 +945,43 @@ impl PeerManager {
...
             }
-            window.count > MAX_PUT_RECORDS_PER_WINDOW
+            let rate = match () {
+                _ if window.count <= MAX_PUT_RECORDS_PER_WINDOW => PutRecordRate::Allowed,
+                _ if window.count > PUT_RECORD_PENALTY_THRESHOLD && !window.penalized => {
+                    window.penalized = true;
+                    PutRecordRate::Flooding
+                }
+                _ => PutRecordRate::Shed,
+            };
```
grantkee, line 1089: A source pinned at exactly the penalty threshold accrues no penalty ever, and nothing caps shed volume itself.

The boundaries here are exactly as intended, with no off-by-one: count 1..=30 gives `Allowed` (BLS verify plus store write), 31..=512 gives `Shed` with no penalty ever, 513 gives one `Penalty::Medium`, and 514+ in the same window gives `Shed` again. The window is tumbling rather than sliding (reset when `now - started >= 60s` at `manager.rs:967-972`), and `penalized` clears on reset. The new tests pin these ranges precisely. And the cheap-first ordering does hold — per shed message the swarm task does two `peer_banned` lookups, one `put_record_windows` entry plus increment, one metric counter and a `trace!` guard, on the order of microseconds against the ~1ms BLS verify and store write it skips. The GHSA-f6rq-62rr-4h9g bound is unchanged from `main`. None of that is in question.

One correction to the description, though. "Still accrues penalties on a ~10-minute-to-ban timescale" overstates where it lands. With `d = exp(-ln2 * 60/300) = 0.870551` per window, the recurrence `s_{n+1} = 0.87055 * s_n - 5` runs `-5.00, -9.35, -13.14, -16.44, -19.31, -21.82`, crossing `-20` at about the 6th window and disconnecting via `Reputation::Disconnected` -> `Disconnecting { banned: true }` -> `ConnectionStatus::Banned` (`all_peers.rs:283-292`, `:177-180`). But steady state is `-5/0.129449 = -38.6`, which is below `-20` so disconnect is reached, and above `min_score_before_ban = -50` so a score-based ban never is — no 30-minute `banned_before_decay` lockout. Decay back above `-20` takes `300 * log2(38.6/20) ~= 285s`, after which the heartbeat emits `Unbanned`. So it's a ~6-minute first strike and then a ~5-minute duty cycle, not a walk to a ban. Containment exists; it just tops out at disconnect.

Two residual gaps that are genuinely new with this branch:

- **No hard cutoff.** The penalty ladder is the only brake, and one Medium per 60s makes it slow. Nothing bounds shed volume itself, so a source can hold ~512 x 16 KiB ~= 8 MiB per source-minute inbound indefinitely. (On the volume figure: `max_value_bytes = 65 KiB` is only checked in the local store on the accept path at `kad.rs:444` and is unreachable over the wire — kad's codec caps a whole message at `DEFAULT_MAX_PACKET_SIZE = 16 KiB`, `libp2p-kad-0.48.0/src/protocol.rs:51` and `:159`, and the repo never calls `set_max_packet_size`.)
- **Churn resets the counter.** `register_disconnected` does `put_record_windows.remove(peer_id)` (`manager.rs:648`), so a reconnecting source gets a fresh count while its score decays. A reconnecting Sybil set can hold 512/window/connection at permanently zero penalty, and each fresh peer id also gets a clean-score 6-minute grace window.

A hard ceiling above the penalty threshold closes the first one:

```rust
/// Records per window above which the source is disconnected outright, not merely scored.
/// Two orders of magnitude above honest replication fan-in; no honest cadence reaches it.
const PUT_RECORD_DISCONNECT_THRESHOLD: usize = 8 * PUT_RECORD_PENALTY_THRESHOLD;

// in put_record_rate_limited:
let rate = match () {
    _ if window.count <= MAX_PUT_RECORDS_PER_WINDOW => PutRecordRate::Allowed,
    // hard cutoff: report every message past the ceiling so the caller keeps scoring
    _ if window.count > PUT_RECORD_DISCONNECT_THRESHOLD => PutRecordRate::Flooding,
    _ if window.count > PUT_RECORD_PENALTY_THRESHOLD && !window.penalized => {
        window.penalized = true;
        PutRecordRate::Flooding
    }
    _ => PutRecordRate::Shed,
};
```

Escalating the once-per-window penalty from `Penalty::Medium` to `Penalty::Severe` (`-10`) would go with it: steady state becomes `-77`, disconnect lands at window 3 instead of 6, and `-50` becomes reachable for a sustained flood. Paired with raising the penalty threshold above the honest ceiling (the other comment on this file), that widens the honest safe band and narrows the attacker's free band at the same time, instead of trading one against the other.

The churn reset is more of a judgment call than a fix — retaining the window in `register_disconnected` when `window.count > MAX_PUT_RECORDS_PER_WINDOW` and letting it age out on its own 60s timer would close it, but it changes reconnect semantics for a Sybil case that may be out of scope here. Happy to leave it. Separately, pinning the wire bound explicitly with `.set_max_packet_size(..)` on the kad `Config` at `consensus.rs:453` would make the 16 KiB cap a deliberate choice rather than an inherited default, in the same spirit as pinning the replication interval in this PR.

grantkee (dismissed):
LGTM - I left some suggestions to harden this PR, but nothing blocking

crates/config/src/network.rs
MavenRain: Addressed in [ab4b45b9](https://github.com/Telcoin-Association/telcoin-network/commit/ab4b45b9d5fe3568a20a0380455621de82f9ab2f). Added `LibP2pConfig::validate()` at both startup configuration funnels beside the score guard. Replication and publication intervals must be nonzero and strictly below the record TTL, with field-named errors before either swarm starts. The docs now distinguish the zero-replication panic from record-expiration problems. Added default, boundary, zero-TTL, and YAML-zero regressions, and updated the short-TTL network fixture to use a valid replication interval.

crates/network-libp2p/src/consensus.rs
MavenRain: Addressed in [ab4b45b9](https://github.com/Telcoin-Association/telcoin-network/commit/ab4b45b9d5fe3568a20a0380455621de82f9ab2f) with two `process_kad_put_request` wiring regressions. The fixture registers a connected source, requires a real `Some(0.0)` score, and verifies it has neither committee nor allowlist exemption. It relays a valid signed record and checks unscored shedding, exactly one Severe penalty (-10), unchanged scoring below the hard cutoff, and a disconnect request plus temporary ban after crossing it. These assertions cover the arm-swap and missing-penalty cases the classifier tests could not observe. Mutation checks confirmed that transposing the arms fails the shedding assertion and removing the penalty fails the flood-score assertion.

crates/network-libp2p/src/peers/manager.rs
MavenRain: Addressed in [ab4b45b9](https://github.com/Telcoin-Association/telcoin-network/commit/ab4b45b9d5fe3568a20a0380455621de82f9ab2f). Added `KAD_MAX_STORED_RECORDS = 1024` and derived the penalty threshold as twice that ceiling (2,048). A regression pins the ceiling to `MemoryStoreConfig::default().max_records` and requires the penalty threshold to stay strictly above it. Updated the fan-in documentation to cover a whole unpaced store replication tick. The expensive-work limit remains 30 records per source-window, with unscored shedding through message 2,048.

crates/network-libp2p/src/peers/manager.rs
MavenRain: Addressed in [ab4b45b9](https://github.com/Telcoin-Association/telcoin-network/commit/ab4b45b9d5fe3568a20a0380455621de82f9ab2f). Flooding now applies `Penalty::Severe`; the hard cutoff is eight times the penalty threshold (16,384), and every classified message above it escalates again. Added boundary and scoring/disconnect regressions. Also retained active over-limit windows and their penalty latch across reconnects, with heartbeat expiry at the original 60-second boundary and regressions for both paths. The kad codec now explicitly pins the 16 KiB packet limit.

Corrected the PR's decay explanation: one Medium penalty per minute approaches -38.6, reaching disconnection without the -50 score-based ban. The revised description reflects Severe scoring and hard-cutoff escalation without promising a fixed ban time.

grantkee (dismissed):

grantkee (dismissed):

grantkee:
@MavenRain I just noticed we've both tried to add this to merge queue twice haha "View details" shows there was a semantic merge conflict that needs to be resolved (`self.network` is moving)

sstanfield (approved):

PR #1300, opened 2026-08-27
fix(network): sweep expired per-source rate windows at heartbeat (#1290)

### LINKED ISSUE #1290: Bound the per-source kad rate-limit maps against the disconnect back-pressure leak
## Severity

Low, theoretical-leaning.  Concurrency / resource exhaustion.  Reachable, but not by ordinary connection churn.

## Problem

`put_record_windows` is unbounded with exactly one eviction site, `register_disconnected` (`manager.rs:610`).  `heartbeat` does not touch it, and `process_kad_put_request` has no connectivity guard before calling `put_record_rate_limited`.

In the normal case the map is bounded by the connected-peer set, exactly as the doc comment claims.  libp2p-swarm runs `behaviour.poll()` before `pool.poll()`, and for a given connection the handler delivers `PutRecord` to kad before `ConnectionClosed`, with kad's `queued_events` drained before the pool is polled again.

The one real hole is the `pending_handler_event` back-pressure branch.  Kad pushes a `NotifyHandler { PutRecordRes }` for every inbound PutRecord, unconditionally and before the app-level limiter runs.  The per-connection command channel is bounded at `task_command_buffer_size: 32`.  An attacker who floods PutRecords and does not read the responses stalls that channel, pins `pending_handler_event = Some(..)`, and from then on every pool event, including `ConnectionClosed`, is processed while kad still holds queued PutRecord events.  On close, the stale event is drained and `put_record_rate_limited` re-inserts a window for a peer that will never disconnect again.

Practical impact: ~120 B per entry with hashbrown overhead.  Reaching 1 GB needs ~9M leaked entries, each requiring a fresh identity to complete a full transport + noise + yamux handshake, send a PutRecord, stall its response channel, and abort.  There is no global inbound-connection cap, so churn is not throttled, but the per-entry cost makes this a slow-burn leak rather than a practical OOM.

The `MAX_PUBLISHED_TO_PEERS` precedent (`consensus.rs:58-72`) documents this exact hazard for the same key type, but `published_to_peers` leaks on every connect unconditionally, whereas `put_record_windows` leaks only on the back-pressure race.

## Proposed fix

Option 1 is preferred: it is smaller, and it makes eviction correctness independent of the swarm's poll ordering rather than relying on it.

```rust
// Option 1 - sweep stale windows in the existing heartbeat (manager.rs:346),
// alongside unban_temp_banned_peers()
let now = Instant::now();
self.put_record_windows
    .retain(|_, w| now.duration_since(w.started) < PUT_RECORD_RATE_WINDOW);
```

O(n) per 30s heartbeat over a map that is normally about the connected-peer count (~30).  It also means `register_disconnected`'s eviction no longer has to be race-free to be correct.

Option 2 mirrors the `published_to_peers` precedent with an `LruCache<PeerId, PutRecordWindow>` capped at 10 000.  Trade-off: an attacker could then evict an honest peer's window, so the cap must stay well above `PeerConfig::max_peers()` (~39), which 10 000 is.

## Same shape elsewhere

`stream/behavior.rs:125` (`inbound: HashMap<PeerId, InboundWindow>`, evicted only in `on_disconnected`) has the identical structure.  Whichever fix is chosen should be applied to both.  Any future per-source window map on the kad path needs the same heartbeat sweep.

## Testing

- Unit test in `crates/network-libp2p/src/tests/peer_manager.rs`: seed `put_record_windows` with a back-dated `started`, run the heartbeat sweep, assert the stale entry is gone and a fresh one is kept.

MavenRain:
Closes #1290.

## Problem

`put_record_windows` has exactly one eviction site, `register_disconnected`.  In the normal case the map stays bounded by the connected-peer set.  The back-pressure branch breaks that bound: kad pushes a `PutRecordRes` toward the per-connection command channel (bounded at `task_command_buffer_size: 32`) before the app-level limiter runs.  An attacker who floods `PutRecord`s and never reads the responses stalls that channel and pins `pending_handler_event`.  Every pool event, including `ConnectionClosed`, is then processed while kad still holds queued `PutRecord` events.  When the stale event drains, `put_record_rate_limited` re-inserts a window for a peer that never disconnects again.  The entry leaks (about 120 B each with map overhead).  The same shape exists in the stream behaviour: `inbound: HashMap<PeerId, InboundWindow>` is evicted only in `on_disconnected`.

The invariant is fragile even without the leak: eviction correctness rests on the swarm's poll ordering (`behaviour.poll()` before `pool.poll()`, handler-delivery before close), not on anything the map enforces itself.

## Fix

Sweep expired windows by age, so the bound no longer depends on poll ordering or on observing the disconnect (option 1 from the issue, applied to both maps):

- [x] `peers/manager.rs`: `sweep_put_record_windows` retains only entries younger than `PUT_RECORD_RATE_WINDOW`;  called from the existing heartbeat, next to `unban_temp_banned_peers`.  O(n) per heartbeat over a map normally sized at the connected-peer count (~30).
- [x] `stream/behavior.rs`: `sweep_stale_windows` retains only entries younger than `INBOUND_RATE_WINDOW`;  called from the existing 1s `sweep` tick that already expires stale pending opens.

The sweep is invisible to both limiters.  An expired window and a missing window act the same: the next message starts a fresh window.  An active flooder refreshes its window on every reset, so its entry is never older than the rate window and the sweep never frees it early.  Each retain predicate is the exact complement of that limiter's expiry check, so the limiter's reset boundary and the sweeper's drop boundary are one shared constant, not two guesses.  Disconnect-time eviction stays in place;  the sweep only removes what a disconnect misses.

## Testing

- [x] `test_put_record_window_swept_at_heartbeat` (`tests/peer_manager.rs`, the test the issue asks for): seeds one back-dated window and one fresh window, runs `heartbeat()`, asserts the stale entry is gone and the fresh one is kept.
- [x] `stale_inbound_window_swept` (`stream/behavior.rs`): the same shape against `sweep_stale_windows`.
- [x] `poll_tick_sweeps_stale_inbound_window` (`stream/behavior.rs`): pins the `poll` wiring itself;  the sweep interval's first tick is immediately ready, so one poll with a noop waker runs the sweep and drops the back-dated window.
- [x] All three new tests mutation-confirmed (4 in-diff mutants): heartbeat sweep call removed, manager retain predicate inverted, poll-tick sweep wiring removed, behavior retain predicate inverted;  each mutant fails its test, and the restored tree re-runs green.
- [x] `network-libp2p/Cargo.toml`: dev-dependency `tokio` gains the `test-util` feature for the paused-clock poll test.
- [x] Local static ladder green: nightly `cargo fmt --all -- --check`;  nightly clippy `--all-features --no-deps -- -D warnings` over the buildplan-emitted 21-crate `-p` list;  scoped `cargo test -p tn-network-libp2p` for the touched suites.  CI runs the full workspace ladder on push;  the pinned `+1.94` attest runs post-push.


sstanfield (approved):

grantkee (approved):
LGTM!

PR #1306, opened 2026-08-28
perf(batch-builder): read sender balances through one state provider (#1303)

### LINKED ISSUE #1303: optimization: batch-builder mdbx lookups
## Problem

`build_batch` constructs the optimistic `changed_accounts` update by reading each sender's canonical balance one address at a time:

```rust
// crates/batch-builder/src/batch.rs:172-181
let changed_accounts: Vec<ChangedAccount> = sender_nonces
    .into_iter()
    .map(|(address, max_nonce)| ChangedAccount {
        address,
        nonce: max_nonce + 1,
        balance: pool
            .get_account_balance(address)
            .saturating_sub(sender_costs.get(&address).copied().unwrap_or(U256::ZERO)),
    })
    .collect();
```

`WorkerTxPool::get_account_balance` (`crates/tn-reth/src/txn_pool.rs:512-519`) calls `BlockchainProvider::basic_account`. In reth v1.11.3 that is not a cheap read against a held provider — it constructs a whole new one on every call:

- `AccountReader for BlockchainProvider` is `self.consistent_provider()?.basic_account(address)` (`crates/storage/provider/src/providers/blockchain_provider.rs:773-775`)
- `ConsistentProvider::new` opens a **new MDBX read transaction** via `database_provider_ro()` and snapshots the in-memory canonical head (`crates/storage/provider/src/providers/consistent.rs:66-80`)
- `basic_account` then calls `latest_ref()`, which builds a **new `MemoryOverlayStateProvider`** over the in-memory canonical blocks (`consistent.rs:107-118`)

So a batch with N distinct senders opens N read transactions and builds N memory overlays, plus the incidental hash-to-number and prune-checkpoint lookups each one needs before the account read itself.

## Impact

This is on the hot path, not a recovery path:

- It runs on **every batch build**, inside the spawned async `next-batch` task (`crates/batch-builder/src/lib.rs:206`), synchronously, with no yield point between reads.
- Batch cadence is one build per round per worker (`min_header_delay` 1000 ms / `max_header_delay` 2500 ms, `crates/config/src/node.rs:267`).
- N is bounded only by the batch limits — 30M gas and 1MB (`crates/types/src/worker/sealed_batch.rs:197-205`). At 21,000 gas for a minimal transfer that is **up to ~1,400 distinct senders, i.e. ~1,400 read transactions, per batch**. There is no per-iteration cap of the kind the pool resync path uses.

No correctness issue: the pool has no cross-address invariant, so reading each sender at a slightly different canonical tip is fine. This is purely cost — runtime-thread occupancy and allocation churn on the path that gates batch sealing.

## Proposed solution

Acquire one state provider for the whole set, the way reth's own `load_accounts` does (`reth crates/transaction-pool/src/maintain.rs:682-687` — one provider at 682, addresses looped against it at 686-687).

Add a batched accessor to the `TxPool` trait (`crates/tn-reth/src/txn_pool.rs:108`):

```rust
/// Return the canonical balances of `addresses` as of the latest committed block.
///
/// Acquires ONE state provider for the whole set: `BlockchainProvider::basic_account` builds a
/// fresh `ConsistentProvider` per call -- an MDBX read transaction plus a
/// `MemoryOverlayStateProvider` over the in-memory canonical blocks -- so a per-address read
/// repeats that setup once per sender in the batch.
fn get_account_balances(&self, addresses: &[Address]) -> HashMap<Address, U256>;
```

with `WorkerTxPool`'s impl acquiring `self.1.latest()` once and looping. Missing accounts and read errors keep the existing `U256::ZERO` fallback, which is already documented as the conservative choice at `txn_pool.rs:104-107`.

Touch points: the single production caller at `batch.rs:178`, and both `TxPool` impls — `WorkerTxPool` (`txn_pool.rs:490`) and `TestPool` (`crates/batch-builder/src/test_utils.rs:41`). `get_account_balance` can either stay as a single-address convenience or be replaced outright.

## How this was found

Surfaced while reviewing the #1243 . That branch adds the same per-address pattern in `WorkerTxPool::reload_dirty_accounts`, but there it is capped at 100 addresses per iteration and only reachable after a canonical-state broadcast lag. This call site is the pre-existing, uncapped, every-round instance, and is the higher-value target of the two.

Related: the pre-existing TODO at `crates/tn-reth/src/txn_pool.rs:365-368` asks whether `on_canonical_state_change` should be offloaded to a blocking task; it runs synchronously in the pool's critical task for every `Commit` and is O(pool size) under the pool write lock.

MavenRain:
Closes #1303.

## Problem

`build_batch` constructs the optimistic `changed_accounts` update by reading each sender's canonical balance one address at a time through `TxPool::get_account_balance`.  In reth v1.11.3 `BlockchainProvider::basic_account` is not a cheap read against a held provider: every call builds a fresh `ConsistentProvider`, which opens a new MDBX read transaction and a new `MemoryOverlayStateProvider` over the in-memory canonical blocks, plus the incidental hash-to-number and prune-checkpoint lookups each one needs.

This is the hot path, not a recovery path.  It runs on every batch build, synchronously inside the spawned `next-batch` task, once per round per worker.  The distinct-sender count is bounded only by the 30M gas / 1MB batch limits, so a batch of minimal transfers can hold ~1,400 distinct senders, i.e. ~1,400 read transactions and memory overlays per build.  No attacker is required for the steady-state cost;  an adversarial many-sender transaction mix pins it at the worst case every round.  The pool resync path caps the same pattern at 100 addresses per iteration;  this call site had no cap at all.

There is no correctness issue: the pool has no cross-address invariant, so reading each sender at a slightly different canonical tip was fine.  This is purely cost (runtime-thread occupancy and allocation churn on the path that gates batch sealing).

## Fix

Acquire one state provider for the whole sender set, the way reth's own `load_accounts` does.

- [x] `crates/tn-reth/src/txn_pool.rs`: replace `TxPool::get_account_balance` with a batched `TxPool::get_account_balances(&self, addresses: &[Address]) -> HashMap<Address, U256>`.  The doc contract requires an entry for every requested address and keeps the documented conservative fallback: a missing account or a read error yields `U256::ZERO`, which can only keep a sender's remaining transactions parked, never promote an unfunded one, and the engine's authoritative canonical update corrects it within the same consensus round.
- [x] `WorkerTxPool` impl acquires ONE state provider (`self.1.latest()`) and loops the addresses against it.  A provider-acquisition failure now degrades the WHOLE set to the conservative zero instead of one address, so it is logged with `warn!` (error, address count) rather than silently, and the trait doc states the all-or-nothing degradation explicitly.
- [x] `crates/batch-builder/src/batch.rs`: the single production caller collects the sender set and issues one `get_account_balances` call.  The per-sender mined-cost debit (`saturating_sub`, issue #1158) is unchanged.
- [x] `crates/batch-builder/src/test_utils.rs`: `TestPool` maps the requested set over its balance table, preserving the `U256::MAX` default for tests that do not exercise balance.

`get_account_balance` had exactly one production caller, so it is replaced outright rather than kept as a single-address convenience.

## Testing

- `cargo +nightly fmt -- --check` green (nightly, per `rustfmt.toml` comment wrapping).
- `cargo +1.94 check --workspace --all-targets` green under the pinned toolchain (trait signature change, so the full-workspace caller sweep is the gate).
- `cargo +1.94 test -p tn-batch-builder` green: the existing #1158 regression tests drive `build_batch` through the batched accessor and assert the exact optimistic balances (real balance minus mined cost, nonce advanced past mined transactions), so a broken batched read fails them.
- GitHub CI (nightly fmt, nightly clippy `-D warnings`, nextest) fires on push;  the pinned `+1.94` attest runs post-push on the second machine.


crates/tn-reth/src/txn_pool.rs
```diff
@@ -577,13 +586,33 @@ impl TxPool for WorkerTxPool {
...
     }
 
-    fn get_account_balance(&self, address: Address) -> U256 {
-        self.1
-            .basic_account(&address)
-            .ok()
-            .flatten()
-            .map(|account| account.balance)
-            .unwrap_or(U256::ZERO)
+    fn get_account_balances(&self, addresses: &[Address]) -> HashMap<Address, U256> {
```
grantkee, line 589: nit: an empty sender set still acquires a state provider, so a build that reaches this call with no senders opens an MDBX read transaction to answer no queries.

`self.1.latest()` runs at `:593`, before `addresses` is touched at `:605` — it appears earlier only as `addresses.len()` inside the log line. Acquisition does address-count-independent work: `latest()` (`blockchain_provider.rs:513-523`) → `block_state_provider` (`:130-137`) → `database.history_by_block_hash(anchor)` → `self.db.tx()?` (`database/mod.rs:238-240`). That is one MDBX read transaction plus a boxed provider allocation. The `head_state() == None` branch is not free either, since `LatestStateProvider::new(self.database_provider_ro()?)` opens the same transaction. The predecessor did zero provider work in this case, because the per-address `.map()` closure never ran.

To be clear about magnitude, the cost is the read transaction and not much more. The memory overlay is usually empty in TN: `finalize_block` calls `remove_persisted_blocks` every round (`env/execution.rs:271-286`), so the in-memory chain only holds blocks for the round currently executing.

On whether the empty set is actually reachable, the obvious mechanism does not work. Every transaction in the selection loop hitting a `continue` is impossible in production, because none of the three skips can fire on the *first* transaction: over-gas needs `gas_limit > 30_000_000` but the pool validator already rejects `gas_limit > block_gas_limit` at admission and TN's block gas limit is exactly 30,000,000 with a strict `>` comparison; the non-allowlisted-type skip is dead code because types 3 and 4 are refused at admission by `.no_eip4844()` / `.no_eip7702()`; and over-byte-size needs `encoded.len() > 1_000_000` against a validator cap of `DEFAULT_MAX_TX_INPUT_BYTES` = 131,072.

It is reachable by a different route. `sender_nonces` is empty whenever `best_transactions()` yields nothing, and that races: the build gate reads the pool in the run-loop task (`crates/batch-builder/src/lib.rs:391`), while `build_batch` runs in a separately spawned tokio task (`lib.rs:207`) and calls `pool.best_transactions()` as its first action (`batch.rs:60`). Concurrently `spawn_maintenance_task` → `update_canonical_state` removes mined transactions and re-prices the pending sub-pool. Between the gate and selection the pending set can empty, which reaches this method with `&[]`. Short window, but it is the steady-state interleaving rather than an exotic one, since both consumers observe the same canonical notification independently.

Nothing downstream misbehaves, which is why this is a waste rather than a bug: `changed_accounts` and `mined_transactions` populate in the same loop body, so empty implies empty, and the run loop short-circuits at `lib.rs:454-459` before it would submit a pool update at `:472`.

The reason I would still take the guard is forward-looking. `max_batch_gas` and `max_batch_size` are epoch-parameterized and documented as fork-changeable (`sealed_batch.rs:196`, `:202`), and they only coincidentally sit at or above the pool's admission caps today. A fork lowering `max_batch_gas` below the block gas limit, or `max_batch_size` below 131,072, makes the first-transaction skip possible and the whole selection loop can then drain to nothing on its own.

The guard belongs here rather than at the call site, since the trait's "an entry for every requested address" contract is satisfied trivially by an empty map:

```suggestion
    fn get_account_balances(&self, addresses: &[Address]) -> HashMap<Address, U256> {
        // An empty set needs no state: acquiring the provider opens an MDBX read
        // transaction (and, mid-round, collects the in-memory canonical blocks into a
        // memory overlay). `build_batch` reaches this call with no senders whenever the
        // pending set drained between the build gate and `best_transactions()`.
        if addresses.is_empty() {
            return HashMap::new();
        }
```

grantkee (changes requested):
One small nit that I think is within scope for this PR. Everything else is solid

grantkee (approved):
LGTM!

PR #1309, opened 2026-08-29
Refactor the epoch vote certifier code.

sstanfield:
This PR is some follow up PR feedback from the just merged PR on this branch.


crates/node/src/manager/epoch_votes.rs
```diff
@@ -101,6 +101,10 @@ async fn manage_epoch_votes(
                         if vote.epoch != epoch_rec.epoch {
                             continue;
                         }
+                        if vote.public_key == me && vote.epoch_hash == epoch_hash {
```
MavenRain, line 114: **Blocker: the merge of main silently undid this fix. Two copies of the `my_vote` capture are now live, and the old one wins.**

Commit 679473d4 did the right thing: it deleted the unguarded capture and re-added it here, after the epoch check and with the `epoch_hash` guard. But main's 54ab4535 is a squash of this branch's pre-feedback history, so it still carried the old form. When d81a6259 merged main back in, git's line-based 3-way merge saw the branch's delete-and-move and main's untouched old lines as two independent edits and kept both. `git show --cc d81a6259 -- crates/node/src/manager/epoch_votes.rs` prints zero hunks, so no conflict ever surfaced.

The file at this head now reads (lines 97-107):

```rust
if vote.public_key == me {
    // Record our vote if we see it for this epoch so we can revote if/when needed.
    my_vote = Some(vote);
}
if vote.epoch != epoch_rec.epoch {
    continue;
}
if vote.public_key == me && vote.epoch_hash == epoch_hash {
    // Record our vote if we see it for this epoch so we can revote if/when needed.
    my_vote = Some(vote);
}
```

The first block runs before the second, so the guarded capture is dead code. Worse, the guard is bypassed in exactly the case it was added for: a self-vote with our key but a different `epoch_hash` (same epoch) is captured by the first block, the `continue` never fires for it, and the third block skips it, so the stale value is never corrected. The per-timeout republish and the new final republish then gossip a vote for the wrong record hash.

Fix: delete lines 97-100 (the unguarded block). That restores what 679473d4 intended. Same story for the duplicated final republish below (see the comment at the post-persist republish).

crates/node/src/manager/epoch_votes.rs
```diff
@@ -195,6 +199,12 @@ async fn manage_epoch_votes(
                 );
             }
         }
+        // Republish our vote one final time.  This is not strictly needed but if we were the first
```
MavenRain, line 206: **Blocker: same evil-merge artifact as the capture above. The final republish now exists twice, and `test_publishes_delayed_initial_and_at_quorum` cannot pass at this head.**

679473d4 moved this block from before `BlsAggregateSignature::aggregate` to here, after `save_certificate` and `persist`, which is what the #1230 review asked for. The d81a6259 merge resurrected the old pre-aggregate copy from main's squash 54ab4535. At this head both are present:

- lines 160-165: republish before `aggregate` (the old copy)
- lines 202-207: republish after `persist` (this one)

Three consequences:

1. The #1230 concern is not actually fixed: the pre-aggregate copy still puts a network round trip between reaching quorum and the durable `save_certificate`/`persist` step, and a crash in that widened window still loses the assembled quorum.
2. The quorum path publishes our vote twice back to back.
3. The new `test_publishes_delayed_initial_and_at_quorum` asserts exactly 2 publishes (delayed initial plus one final republish). The quorum path now emits 3. Publish 2 (pre-aggregate) lands before the cert is stored and publish 3 (post-persist) lands milliseconds after, so the `wait_until` predicate `cert stored && publishes == 2` times out; even if a poll hit the tiny window, the `assert_eq!(publishes, 2)` after the 200 ms sleep observes 3. The test fails deterministically. CI never caught it because all build/test jobs skip on maintainer PRs and no attestation has run for d56b59b6.

Fix: delete lines 160-165 (the pre-aggregate copy). Keeping this post-persist one preserves the intent of the move.

crates/node/src/manager/epoch_votes.rs
```diff
@@ -205,6 +215,10 @@ async fn manage_epoch_votes(
         // Try to recover by downloading the epoch record and cert from a peer
         let mut got_epoch_record = false;
         for _ in 0..EPOCH_CERT_RECOVERY_ATTEMPTS {
+            if vote_rx.is_closed() {
```
MavenRain: **The `is_closed()` bail is over-broad for recovery and under-effective as a task bound.**

Two separate problems with the placement:

1. Over-broad. The peer fetch needs only the network handle and the db; whether more votes can arrive is irrelevant to it. And `is_closed()` becomes true on the benign path too: `vote_queues.pop_front()` at capacity 5 drops a live collector's sender. An evicted collector that still holds an uncertified record now runs zero `request_epoch_cert` attempts, falls through to `!got_epoch_record`, and logs "We have a missing epoch certificate ... sync will be compromised!" for a condition it never tested. Before this change it would at least have tried the five fetches. That collector was the epoch's only self-heal path until restart.

2. Under-effective. The recovery loop has no shutdown arm (the `Noticer` is only selected in the vote loop above), and the long pole is inside `request_epoch_cert` itself, three sequential peer request timeouts, which a between-attempts `is_closed()` cannot cancel.

Suggestion: make the exit condition the real shutdown signal, `tokio::select!` on `&shutdown` around `request_epoch_cert`, and decide the eviction semantics explicitly. If an evicted round should die without recovery, a one-line comment saying so plus a quieter log would close the question; if not, let it run at least one fetch before bailing.

One adjacent note while this block is open: the ladder has no backoff. With no connected peers the network task answers `SendRequestAny` immediately with `NoPeers` (network-libp2p consensus.rs:974), so all 5 attempts (15 peer requests) complete in microseconds and the node never retries for that epoch. A short sleep between attempts would make `EPOCH_CERT_RECOVERY_ATTEMPTS` mean what it says. Pre-existing, not introduced here.

crates/node/src/manager/epoch_votes.rs
```diff
@@ -291,12 +305,16 @@ async fn handle_new_vote(
...
         // We do not have a collector for this epoch so start one and send it this vote.
         let (epoch_vote_tx, epoch_vote_rx) = mpsc::channel(10_000);
-        // If we recieve a valid vote and aren't collecting votes to certify then start.
+        // If we receive a valid vote and aren't collecting votes to certify then start.
         let Some((epoch_rec, None)) =
             consensus_chain.epochs().get_epoch_by_hash(vote.epoch_hash).await
         else {
             // Missing the record or it is certified.  These were pre-checked when the gossip came
             // in so this should not happen.
+            error!(
```
MavenRain: **This error also fires for the already-certified case, which is a benign race that this same PR makes more frequent.**

The `let ... else` arm is taken both when the record is missing and when it is `Some((rec, Some(cert)))`, already certified. The message asserts "missing epoch record" and "should not happen" for both. A vote that passes the gossip pre-check while the record is uncertified, and arrives here after a cert lands (peer recovery, state-sync, or our own quorum), hits the certified case on a normal path. The new final republish aims one more vote at laggy validators right around the moment certs land, so this race gets more common, and operators see a critical-sounding error for expected behavior.

Suggestion: split the two cases. Keep `error!` for the truly-missing record, and log the certified case at `debug!` (or drop it), with a message that says what actually happened, for example "vote arrived after the epoch was certified, ignoring".

crates/node/src/manager/epoch_votes.rs
```diff
@@ -391,9 +409,14 @@ pub(crate) fn spawn_epoch_vote_collector(
                             "Failed to certify last epoch, re-publishing epoch record vote for epoch {} {}", last_epoch_rec.epoch, last_epoch_rec.digest()
                         );
                         // Sending our vote to this channel will trigger us to start the vote collector when we get it.
-                        // The other committe members of last epoch should also do the same.
+                        // The other committee members of last epoch should also do the same.
                         // This should not happen and if it does certification may fail again but retry after an epoch anyway.
-                        let _ = consensus_bus.new_epoch_votes().send(epoch_vote).await;
+                        if consensus_bus.new_epoch_votes().send(epoch_vote).await.is_err() {
```
MavenRain: **These new `is_err()` branches cannot fire. The failure modes the bus actually has are a silent drop and a park, and neither is logged.**

`new_epoch_votes()` is a `QueChannel`; its `send` (consensus_bus.rs:150-153) returns `Ok(())` without sending when nothing has subscribed, and otherwise awaits the bounded mpsc send, which errors only when the receiver is dropped. The receiver here is `vote_rx`, owned by this same "Epoch Vote Collector" task (subscribed at line 359, before the spawn). So while this code runs, `send` can succeed or park, never `Err`, and both new error branches are dead code.

The reachable bad cases are the opposite shape:

- If the channel is full (bounded capacity) while the collector is executing this outer body, the body awaits a send to a queue only it drains: it parks until gossip stops filling the channel, and in the worst case wedges the critical collector task with no log.
- If the bus were ever unsubscribed, `send` silently drops the vote and reports success.

`try_send` (consensus_bus.rs:156-160) is the call shape whose `Err` can actually fire here, and it also converts the park into a logged drop: `if let Err(e) = consensus_bus.new_epoch_votes().try_send(epoch_vote) { error!(...) }` at both sites. That would make the log the #1230 comment asked for real. Minor: whatever shape you pick, the identical block appears at both publish points; a small helper would keep them from drifting.

crates/node/src/manager/epoch_votes.rs
```diff
@@ -1286,4 +1315,434 @@ mod epoch_vote_collector_tests {
...
+            kc1,
+            primary_network,
+            vote_rx,
+            consensus_chain,
+            node_shutdown,
+        )
+        .await;
+
+        assert!(
+            publishes.load(Ordering::SeqCst) >= 1,
```
MavenRain: **This test cannot distinguish the guarded capture from the unguarded one, which is why the duplicate-capture regression above survives it.**

The only seeded vote is `epoch_rec.sign_vote(&kc1)`, whose `epoch` and `epoch_hash` both match the record under collection, so the pre-existing unguarded capture and the new guarded capture assign the identical value. And the assertion is `publishes >= 1` with no payload inspection, so it cannot tell which vote was republished. Delete lines 104-107, or delete 97-100, or keep both: the test passes in all three variants.

To pin the property: first send a self-signed vote for a different epoch or hash, then the matching one, and assert on the `NetworkCommand::Publish` payload the mock already receives that the republished vote carries `epoch_hash == epoch_rec.digest()`.

Same pattern in `test_fork_digest_mismatch_triggers_shutdown` (line 1342): the docstring says the collector must fail-stop "rather than silently adopt the peer's record", but the only assertion is `noticer.noticed()`. A regression that notifies and also saves the peer's record passes. Adding `assert!(consensus_chain.epochs().cert_by_digest(alt_rec.digest()).await.is_none())` and a latest-record check would pin the non-adoption half. (Also, the "Returns once shutdown is signalled" comment is not quite right: the function returns because the recovery loop breaks after the download, nothing on that path polls the notifier.)

crates/node/src/manager/epoch_votes.rs
```diff
@@ -421,21 +449,18 @@ pub(crate) fn spawn_epoch_vote_collector(
...
+            // See spawn_epoch_record_collector() for how syncing nodes can keep up their epoch certs if not following the tip.
         }
     });
 }
 
 /// Re-arm the epoch-vote round for a stored-but-uncertified latest epoch record.
 ///
-/// The vote collector is armed only by `epoch_record_watch`, and only the epoch-close paths
-/// write that channel. The channel is in-memory. A node that persists its epoch record and
+/// The vote collector is armed by a vote for an uncertified epoch record, and only the epoch-close
```
MavenRain: **The rewritten docstring now contradicts the function it documents.**

Two spots:

1. "The vote collector is armed by a vote for an uncertified epoch record, and only the epoch-close paths write that channel." The "that channel" is left dangling from the deleted `epoch_record_watch` sentence, and it is the wrong arming story for this function: `revote_uncertified_epoch_record_on_startup` arms the collector by writing the watch (`epoch_record_watch.send_replace(...)`), which is also the reason the surviving "Call this AFTER `spawn_epoch_vote_collector`" instruction matters.

2. "this heals the latest stored record or previous record only". The body only inspects `db.latest_record()`. N-1 is healed by the collector's failsafe, and that fires only when a watch value is delivered, which on the startup path happens only when the latest record is itself uncertified. In the common gap shape (epoch N certified, N-1 uncertified because its round failed and the next epoch then succeeded), startup heals nothing. Saying "re-arms the latest stored record; if that record is uncertified, the collector's failsafe also retries the record before it" would match the code.

MavenRain (changes requested):
Reviewed at head d56b59b6 (diff vs merge-base db465a06, +477/-15 in `crates/node/src/manager/epoch_votes.rs` plus a dev-only `tokio/test-util` addition).

The intent of the feedback commits is right, and several items land cleanly: the dead-collector respawn, the doc and typo corrections, and five tests that target exactly the #1230 test-gap list. But the branch as pushed does not implement the feedback it set out to implement, for one structural reason and a few local ones.

**Blocker: the d81a6259 merge of main silently reverted half of 679473d4.** Main's 54ab4535 is a squash of this branch's pre-feedback history, so it still carried the old unguarded `my_vote` capture and the old pre-aggregate republish. The 3-way merge kept the branch's moved copies and main's old copies side by side, with no conflict; `git show --cc d81a6259 -- crates/node/src/manager/epoch_votes.rs` prints zero hunks, so nothing ever surfaced. At this head:

- Lines 97-100 and 104-107: both captures. The unguarded one runs first, so the new guard is dead code, and a self-vote for the same epoch with a different hash is still captured and republished.
- Lines 160-165 and 202-207: both republishes. The durable step is still delayed by the pre-aggregate copy, the quorum path publishes twice, and `test_publishes_delayed_initial_and_at_quorum` asserts exactly 2 publishes while this path emits 3, so the new test fails at this head. CI cannot catch it: all build/test jobs skip on maintainer PRs, and no attestation has run for d56b59b6.

The fix is two deletions, lines 97-100 and 160-165. Since this branch continues past its own squash-merge, every future merge of main into it carries the same silent-duplication hazard; rebasing onto main, or diffing the file against main's copy after each merge, would surface it.

**The other findings, each in an inline comment:**

- The new bus-send `is_err()` branches are unreachable (`QueChannel::send` silently succeeds when unsubscribed and parks when full; the receiver is this same task), so the logging the #1230 comment asked for is dead code, and the reachable failure modes, a silent drop and a park of the critical collector task, stay unlogged. `try_send` gives a branch that can actually fire.
- The `is_closed()` bail runs before any recovery attempt, so an evicted collector skips cert recovery entirely and logs "sync will be compromised" for a condition it never tested; the loop also has no shutdown arm, and the check cannot cancel a fetch in flight.
- `test_my_vote_republished_on_timeout` cannot distinguish the guarded capture from the unguarded one (it passes with either or both present), and `test_fork_digest_mismatch_triggers_shutdown` asserts the shutdown but not the non-adoption of the peer's record.
- The new "missing epoch record" error also fires for the benign already-certified race, which the final republish makes more common.
- The rewritten `revote_uncertified_epoch_record_on_startup` docstring contradicts the code (this function arms via the watch, and the N-1 healing claim does not hold on the startup path).

**Smaller residuals**, no diff line to anchor on: `handle_new_vote` still keys the queue entry on the unsigned `vote.epoch` (line 341) rather than `epoch_rec.epoch` from the record found by hash, the un-applied half of the #1230 suggestion. The recovery ladder has no backoff, so with no connected peers all five attempts complete in microseconds (pre-existing). And a shutdown during the 1 s post-quorum grace window discards a fully-collected super-quorum that was one `aggregate` call away from durable (`_ = &shutdown => return` skips the save path).

MavenRain (approved):
Comments having been acknowledged and/or addressed, I'm good with this

grantkee (approved):
LGTM! Good work @sstanfield

PR #1311, opened 2026-08-29
follow-up: update docs, tests, and tools for prevrandao adiri fork

### LINKED ISSUE #1247: Transaction-dependent PREVRANDAO lets a batch/header proposer grind on-chain randomness

grantkee:
Follow-up to #1249 (fix #1247). The derivation is untouched; this land seven gaps a post-merge verification pass found, one commit per finding. (Cantina #9)

- The engine's hand-composed mix-hash oracle now dispatches on `prevrandao_seed_active` like production, so it holds under both feature sets; it was red under `--features adiri`. tn-engine joins the adiri CI lane (whose comment is reworded to match), and the widened lane is what would have caught the failure.
- `TN_PREVRANDAO_FORK_EPOCH` is declared in the Makefile as a third sibling defaulting to dormant, exported on the five fork-pinned test lanes, and pinned by the e2e harness loop. The pin keeps a `TN_SEED_SIGNATURE_FORK_EPOCH=0` lane from silently arming PREVRANDAO on non-adiri builds; reaching the seeded derivation needs both variables armed, and the comments now say so. Bare `--workspace` lines stay variable-free so the override inertness test keeps a clean process.
- The startup fork-schedule log gains `prevrandao_fork_epoch` in the adiri arm and names prevrandao in the non-adiri genesis-active message; the constant's docs carry the siblings' fleet-diff rationale.
- New boundary keepers derive both grid points from the constant, so an arming PR retargets them with no edit: under adiri the legacy XOR must replay byte-identically at the epoch below the fork and the seeded arm must fire at the fork epoch, with anti-vacuity tripwires; without adiri every epoch takes the seeded arm. A third test pins the fail-closed conjunction: seed-dormant epochs stay legacy regardless of the fork point.
- `docs/src/evm-compatibility.md` and the tn-reth README now document the fork-gated derivation, with indexer guidance: branch on the block's epoch (upper 32 bits of `nonce`); the seed chain value equals the epoch-close `extra_data`.
- The residual last-actor bias is recorded at `prev_randao`, the fork constant, the payload attribute, and the EVM block module docs: the committing leader can precompute every value and withhold the commit, so the fork narrows grinding to that one choice rather than making PREVRANDAO unbiasable.
- Fixed a stale test comment claiming `mix_hash` seeds the committee shuffle; the shuffle seed rides `close_epoch`, derived from `committee_shuffle_seed`, never `mix_hash`.

All four CI lanes pass locally: fmt, clippy (default and all-features), the adiri lane at 817/817 (up from 777 with tn-engine and the new keepers), and the default workspace at 1594/1594.


MavenRain (dismissed):
Reviewed at 428a3735 line by line against db465a06, with the fork-gate semantics re-derived from the code.  All seven commit-level claims hold: the engine oracle's legacy arm is byte-identical to production (`ConsensusOutput::digest()` is `consensus_header_hash()`), the keeper tripwires hold at the current constants, `tn-engine` forwards `adiri` so the widened lane line is well-formed, and the `epoch.rs` comment fix is correct (the shuffle seed rides `close_epoch` via `TNPayload::new` from `committee_shuffle_seed()`; `mix_hash` is not among the shuffle inputs).

I'll approve.  Seven inline comments follow (3 MED, 3 LOW, 1 INFO), none blocking.  The MEDs: the indexer-facing `extra_data` sentence only holds for the closing block;  after the pin, nothing anywhere executes the seeded mix_hash on spawned nodes;  and the new seed-conjunct test cannot fail through the conjunct it pins.

docs/src/evm-compatibility.md
```diff
@@ -112,13 +112,19 @@ TN repurposes several Ethereum block header fields to carry consensus-layer meta
...
 | `parent_beacon_block_root` | Beacon chain parent root     | Digest of the `ConsensusHeader` that committed the transactions                            |
 | `extra_data`               | Arbitrary miner data         | Committee-shuffle seed (the epoch seed chain value) at epoch boundaries, empty bytes otherwise |
 | `base_fee_per_gas`         | Adjusts per block (EIP-1559) | Fixed for the entire epoch, adjusts at epoch boundaries. See [basefees](basefees.md)       |
 | `withdrawals`              | Beacon chain withdrawals     | Validator reward records at epoch boundaries, empty otherwise                              |
 
+The `mix_hash` derivation changed at a hard fork. Blocks from epochs before the PREVRANDAO fork epoch use
+the legacy `output_digest ^ batch_digest` form; blocks from that epoch onward use the seed-chain
+derivation. Indexers that reproduce `mix_hash` from consensus data must branch on the block's epoch (the
+upper 32 bits of `nonce`). The seed chain value is the same one published as `extra_data` on the epoch's
```
MavenRain: **MED (docs):** this sentence equates the per-commit seed chain value with the epoch-close `extra_data`, which only holds for the closing block itself.

`committee_shuffle_seed()` returns the commit's own fold state (`seed_chain.fold(leader.round(), sig)` in `CommittedSubDag::new`, `output.rs:489`; the cursor advances every commit, `bullshark.rs:276`), while `extra_data` publishes only the closing commit's value (`payload.rs:104-107`). Mid-epoch, the seed hashed into `mix_hash` appears on no executed header at all.  An indexer that follows the sentence as written reproduces the wrong `mix_hash` for every non-closing block in the epoch.

Suggested wording: the seed is that block's own commit-level chain value; it equals the closing block's `extra_data` only for the closing block itself;  for any other block, fold the epoch's seed chain forward from consensus data (epoch root, then each prior commit's leader round and seed signature) up to that block's commit.

Same sentence appears in the `crates/tn-reth/README.md` mix_hash row and the PR body.

Makefile
```diff
@@ -161,19 +161,33 @@ TN_SEED_SIGNATURE_FORK_EPOCH ?= 4294967295
 # multi_workers_override_is_inert_when_unset requires a process WITHOUT the variable.
 TN_MULTI_WORKERS_FORK_EPOCH ?= 4294967295
 
+# PREVRANDAO fork epoch for the e2e lanes (#1247). Same shape as the two variables above.
+# Defaults to u32::MAX so the default lanes run the fork DORMANT (the legacy
+# `output_digest ^ batch_digest` mix hash, what adiri has already executed); non-adiri builds
+# are otherwise active from genesis. NOT independent of the seed-signature variable in the
+# arming direction: `prevrandao_seed_active` requires `seed_signature_active`, so a lane needs
+# BOTH armed to reach the seeded derivation:
+#   TN_SEED_SIGNATURE_FORK_EPOCH=0 TN_PREVRANDAO_FORK_EPOCH=0 make test-e2e
```
MavenRain: **MED (coverage):** after this pin, no lane, target, or documented command executes the seeded mix_hash on spawned nodes.

Pre-PR, the documented `TN_SEED_SIGNATURE_FORK_EPOCH=0 make test-e2e` (old line 148) armed PREVRANDAO implicitly on the non-adiri e2e binary, so it exercised the seeded derivation end to end.  The pin correctly stops that (fail-closed default, and this comment now documents the two-var form), but the two-var invocation exists only as a comment: nothing runs it.  Until the arming PR, the post-fork wire format has zero process-level coverage on spawned nodes.

A `test-e2e-forked` target (or one e2e lane exporting both vars at 0) would close that gap before arming.  Also, worth a line in the PR body: any script still using the one-var form now tests the legacy arm.

crates/types/src/forks.rs
```diff
@@ -684,6 +690,25 @@ mod tests {
         });
     }
 
+    /// The fail-closed conjunct documented on [`prevrandao_seed_active`]: a seed-inactive epoch
+    /// must stay on the legacy XOR even where the PREVRANDAO fork point itself is active, so the
+    /// hash never folds the forkable legacy leader-aggregate seed.
+    #[cfg(feature = "adiri")]
+    #[test]
+    fn prevrandao_seed_active_is_closed_by_the_seed_conjunct() {
```
MavenRain: **MED (test vacuity):** this test cannot fail through the conjunct it pins.

At all three probed epochs the fork-point conjunct is already false (`PREVRANDAO_FORK_EPOCH` is `Epoch::MAX`), so deleting `&& seed_signature_active(epoch)` from `prevrandao_seed_active` leaves the test green: it cannot catch the exact regression its doc comment names (hashing the forkable legacy aggregate seed).  The compile-time `PREVRANDAO >= SEED` assert makes "fork point active, seed dormant" unreachable through the constants, so the conjunct is only testable via a helper that takes both fork points as parameters (or a single-process test that sets both env overrides).

Either restructure so the conjunct is reachable, or reword the doc comment to claim only what the probes cover.

Related nit on the grid line: `SEED_SIGNATURE_FORK_EPOCH - 1` underflow-panics if a rollout ever sets the seed fork to 0;  `saturating_sub(1)` preserves the intended failure message.

crates/types/src/primary/output.rs
```diff
@@ -720,4 +749,67 @@ mod tests {
...
+    /// THE boundary keeper for #1247: the arm switch must happen AT `PREVRANDAO_FORK_EPOCH`.
+    /// Both epochs derive from the constant, so an arming PR retargets this with no edit here.
+    #[cfg(feature = "adiri")]
+    #[test]
+    fn prev_randao_switches_arms_at_the_prevrandao_fork_epoch() {
+        let post_fork = crate::forks::PREVRANDAO_FORK_EPOCH;
+        let pre_fork = post_fork - 1;
+        // anti-vacuity tripwire, mirroring `committee_sweep_tests.rs`: an ambient override or an
+        // armed fork would make both sides land on the same arm and pass for the wrong reason
+        assert!(
```
MavenRain, line 781: **LOW:** the keeper goes vacuous if the arming PR sets `PREVRANDAO_FORK_EPOCH == SEED_SIGNATURE_FORK_EPOCH` (the const assert at `forks.rs:329` permits equality).

At equality, `pre_fork = SEED - 1` is seed-dormant, so this tripwire and both arm asserts are satisfied through the seed conjunct alone; deleting the PREVRANDAO fork-point gate entirely keeps the keeper green.  In practice the arming PR must target a future epoch (the seed fork shipped at 383), so equality would be an invalid rollout anyway;  still, one extra tripwire pair (`!prevrandao_fork_point_active(pre_fork)` and `prevrandao_fork_point_active(post_fork)`) or a strict `>` in the const assert makes the keeper hold under every constant the compile gate accepts.

crates/types/src/primary/output.rs
```diff
@@ -720,4 +749,67 @@ mod tests {
...
+            legacy.prev_randao(0, digest),
+            legacy_header ^ digest,
+            "PREVRANDAO_FORK_EPOCH - 1 must replay the legacy XOR byte-identically",
+        );
+        assert_eq!(
+            seeded.prev_randao(0, digest),
+            seeded_prev_randao(seeded.committee_shuffle_seed(), 11, 0),
+            "the gate must fire from PREVRANDAO_FORK_EPOCH onward (`>=`, not `>`)",
+        );
+        assert_ne!(
```
MavenRain, line 811: **LOW:** this `assert_ne!` is satisfied by the fixtures, not by the arm split.

`legacy` and `seeded` carry different leader headers, so their digests differ and an unconditionally-legacy `prev_randao` also passes it;  the message claims a discrimination the assert does not perform.  Comparing `seeded.prev_randao(0, digest)` against `seeded`'s own legacy recomposition (`B256::from(seeded.digest()) ^ digest`) would make it discriminate the arms.

Related nit: the non-adiri counterpart's grid hardcodes `383`, which cannot track the adiri constant it mirrors.

crates/engine/tests/it/main.rs
```diff
@@ -2376,18 +2376,26 @@ async fn test_simple_basefee_penalty() -> eyre::Result<()> {
...
-            ]
-            .concat(),
-        );
+        // Independent oracle for the mix hash: recompose the exact bytes by hand for whichever
+        // derivation this build's fork schedule selects, rather than trusting
+        // `ConsensusOutput::prev_randao` to check itself. Gated on the same predicate the engine
+        // dispatches on, so the oracle stays correct under `--features adiri` (dormant: legacy
+        // XOR) and default features (active: seeded).
+        let expected_mix_hash =
+            if tn_types::forks::prevrandao_seed_active(consensus_output.leader().epoch()) {
```
MavenRain, line 2406: **LOW:** the oracle is independent on byte layout only; its dispatch is the production predicate.

This `if` is the same `prevrandao_seed_active(leader.epoch())` call production makes, so a dispatch regression moves both sides together; for example, `prevrandao_seed_active` weakening to plain `seed_signature_active` stays green in both feature lanes at this test (the adiri boundary keeper is what catches that one at current constants).

An unconditional `assert!(tn_types::forks::prevrandao_seed_active(...))` under `cfg(not(feature = "adiri"))` next to the oracle restores the "default builds are seeded from genesis" pin this assertion used to carry.

crates/telcoin-network-cli/src/node.rs
```diff
@@ -119,20 +119,21 @@ impl<Ext: clap::Args + fmt::Debug> NodeCommand<Ext> {
...
             seed_signature_fork_epoch = tn_types::forks::SEED_SIGNATURE_FORK_EPOCH,
+            prevrandao_fork_epoch = tn_types::forks::PREVRANDAO_FORK_EPOCH,
             multi_workers_fork_epoch = tn_types::forks::MULTI_WORKERS_FORK_EPOCH,
             "fork schedule (adiri)"
         );
         #[cfg(not(feature = "adiri"))]
         info!(
             target: "cli",
-            "fork schedule: seed_signature and multi_workers active from genesis"
+            "fork schedule: seed_signature, prevrandao and multi_workers active from genesis"
```
MavenRain: **INFO:** this line can state "active from genesis" while a test-utils binary runs all three forks dormant.

The e2e harness now spawns exactly that configuration (all three overrides pinned to `u32::MAX`).  Production non-adiri binaries do not consult overrides, so the line is true where it matters;  logging the override values when `test-utils` is compiled in would make the startup line trustworthy while debugging an e2e mix_hash mismatch.  The adiri arm shares the blind spot (it prints constants, not override values), and no test covers the `prevrandao_fork_epoch` tracing field name.

docs/src/evm-compatibility.md
grantkee: Good catch - updated

Makefile
grantkee: Good catch - updated

crates/types/src/forks.rs
grantkee: Good catch - updated

crates/types/src/primary/output.rs
grantkee: Good catch - updated

crates/types/src/primary/output.rs
grantkee: Good catch - updated

crates/engine/tests/it/main.rs
grantkee: Good catch - updated

crates/telcoin-network-cli/src/node.rs
grantkee: Good catch - updated

MavenRain (dismissed):
All comments having been acknowledged/addressed

MavenRain (approved):

PR #1312, opened 2026-08-29
fix(tn-reth): quote and fill RPC fee fields from the epoch base fee (#1305, #1313)

### LINKED ISSUE #1305: `eth_gasPrice` quotes 500 gwei on a chain that charges 7 wei
## Problem

Adiri's RPC answers `eth_gasPrice` with 500,000,000,007 wei and `eth_maxPriorityFeePerGas` with 500,000,000,000 wei. The chain's actual base fee is 7 wei: the latest header carries `baseFeePerGas: 0x7` and every `eth_feeHistory` entry is `0x7`. Wallets that price a transaction off `eth_gasPrice` overpay by roughly 7x10^10.

The 7 wei is correct. It is the worker's on-chain fee from `WorkerConfigs`, currently sitting at `MIN_PROTOCOL_BASE_FEE` (`crates/types/src/gas_accumulator.rs:252`, `crates/tn-reth/src/txn_pool.rs:357`). The 500 gwei comes from reth's gas price oracle, which TN has never overridden. `eth_gasPrice` is `suggest_tip_cap() + latest_header.base_fee` (reth v1.11.3 `crates/rpc/rpc-eth-api/src/helpers/fee.rs:364`), and 500,000,000,000 is `DEFAULT_MAX_GAS_PRICE` verbatim (reth `crates/rpc/rpc-server-types/src/constants.rs:92`). The RPC is not returning an estimate, it is returning the oracle's clamp ceiling, which means the raw estimate is higher than 500 gwei.

It is higher because the quote feeds itself. Every recent adiri block holds one transaction from the same sender (`0xf5e8a439...` to `0x8b14d287...`), legacy type `0x0`, priced at `gasPrice = 600,000,000,008`. That number is not arbitrary: `500,000,000,007 x 1.2 = 600,000,000,008.4`, truncated. The client reads `eth_gasPrice` and applies a 1.2x bump. Its effective tip is `600,000,000,008 - 7 = 600,000,000,001`, which is exactly the `reward` column `eth_feeHistory` returns at the 10th, 50th and 90th percentiles. The oracle samples those tips, takes the 60th percentile, clamps to `max_price`, adds the 7 wei base fee, and quotes 500,000,000,007 again. Nothing breaks the cycle, and reth's ceiling is the only thing keeping it from diverging (600 to 720 to 864 and upward).

The oracle cannot quote TN correctly even with honest traffic, so raising or lowering the ceiling is not a fix. `ignore_price` defaults to 2 wei (reth `constants.rs:95`). Against a 7 wei base fee, a transaction priced honestly at 7 or 8 wei has an effective tip of 0 or 1 wei, below that threshold, so `get_block_values` discards every sample. With no samples, `suggest_tip_cap` falls back to a cached `last_price` that never decays and only resets on process restart, to `GWEI_TO_WEI` (reth `gas_oracle.rs:414`). The honest-traffic answer is 1 gwei plus the base fee, not 7 wei. There is no sampling regime in which this oracle reaches TN's fee level.

The overpayment is not burned. `reward_beneficiary` credits the priority fee to the block beneficiary (`crates/tn-reth/src/evm/handler.rs:171`), so the excess has been flowing to validators on every transaction since the loop started.

There is no configuration-only fix available today. reth exposes `--gpo.maxprice`, `--gpo.ignoreprice` and `--gpo.default-suggested-fee`, but TN's CLI is a deliberate allowlist: `RethCommand` (`crates/tn-reth/src/cli.rs:69`) flattens TN's own `RpcServerArgs` (`crates/tn-reth/src/rpc_server_args.rs:74`), which has no `gas_price_oracle` field, and the conversion to reth's args fills the remainder with `..Default::default()` (`rpc_server_args.rs:299`). Passing `--gpo.*` is rejected as an unknown argument.

Observed on adiri at block `0x4ea60`. Note that `web3_clientVersion` reports `0.1.0 (5f1c0b49)`, which predates the `eth_feeHistory` correction in 34f0f83b, so `eth_blobBaseFee` there still answers `0x1` rather than the `0x0` main now serves. That lag is unrelated to this bug, which is present on main.

## Solution

Replace `eth_gasPrice` and `eth_maxPriorityFeePerGas` with TN handlers, following the pattern already established for `eth_feeHistory`. `crates/tn-reth/src/rpc_fee_history.rs` declares an override trait whose method names match reth's `EthApiServer`, and `crates/tn-reth/src/env/rpc.rs:137` swaps the handlers in with `add_or_replace_if_module_configured`. The `BaseFeeContainer` those handlers need is already threaded into `get_rpc_server`, so a sibling module needs no new wiring beyond one more install call.

`eth_maxPriorityFeePerGas` should answer 0 with no delegation. TN has no congestion fee market: batches are ordered by consensus and blocks run at roughly 0.17% of the gas limit, so there is no scarcity for a tip to bid on. The honest answer is a constant, and like the existing `eth_blobBaseFee` override it cannot be perturbed by a provider fault.

`eth_gasPrice` should answer `base_fee.base_fee()`, the worker's current epoch fee from the container. That is 7 wei today and whatever governance writes into the worker's `WorkerConfigs` row tomorrow, which is the behaviour asked for: the quote tracks the on-chain fee rather than a sampled history.

Read the fee from the container rather than the latest header, for the reason `rpc_fee_history.rs` already documents. An RPC server belongs to one worker, the latest header may have been sealed by a different worker at a different fee, and across an epoch boundary the header carries the closing epoch's fee. The container holds the fee this node's next batch will actually enforce, which is what a client pricing a new transaction needs.

One behaviour changes. reth's pool orders pending transactions by coinbase tip, so a suggested tip of 0 sorts a compliant transaction last within a batch. At current utilization that is immaterial, but if preserving expressive ordering matters more than a clean zero, 1 wei costs the sender nothing and keeps the ordering signal.

The alternative is to add reth's `--gpo.*` flags to TN's allowlist and have operators set them. That is rejected for three reasons: it is the same amount of code as the override, it leaves correctness dependent on every operator configuring the flags identically, and as shown above no setting of `ignore_price` or `max_price` makes a history-sampling oracle produce TN's fee, because TN's real tips are structurally below the sampling floor.

The fix does not retroactively help the existing bot, which will keep quoting itself 500 gwei until the fleet is rolled. Once rolled, `eth_gasPrice` returns 7, its 1.2x bump produces 8 wei, and the loop dies.

### LINKED ISSUE #1313: eth_fillTransaction still prices from the gas price oracle removed from eth_gasPrice
## Problem

`eth_fillTransaction` still prices requests from the gas price oracle that #1305 removed from `eth_gasPrice` / `eth_maxPriorityFeePerGas`.

reth's `EthApiServer` serves `eth_fillTransaction` (reth v1.11.3 `crates/rpc/rpc-eth-api/src/core.rs:244`) on the same eth namespace TN exposes.  Its default handler fills a request's missing fee fields from `suggested_priority_fee()` plus the latest header's base fee, which is the tip-sampling oracle path.  Unlike `eth_sendTransaction`, it needs no signer, only a `from` address, so any RPC client can reach it.

After #1305, the same server answers `eth_maxPriorityFeePerGas` with 0 and `eth_gasPrice` with the worker's epoch base fee, while `eth_fillTransaction` fills `maxPriorityFeePerGas` with the oracle's answer: 1 gwei on a fresh process, and up to reth's 500 gwei `DEFAULT_MAX_GAS_PRICE` clamp once a client bump loop has run.  A client that fills, signs, and submits overpays by the same ratio #1305 describes, and `reward_beneficiary` credits the excess to the block beneficiary rather than burning it.

## Solution

Extend the `crates/tn-reth/src/rpc_gas_price.rs` override (or add a sibling) with a `fillTransaction` handler that delegates to the unchanged reth implementation and then overwrites the filled fee fields from the shared `BaseFeeContainer`: priority fee 0, fee cap the floored epoch base fee.  That is the delegate-then-correct pattern `rpc_fee_history.rs` already uses, installed through the same `add_or_replace_if_module_configured` call in `get_rpc_server`.

MavenRain:
Closes #1305. Closes #1313.

## Problem

Reth's tip-sampling oracle quotes fees far above TN's epoch base fee and feeds those defaults into `eth_fillTransaction`. A client that fills, signs, and submits can overpay without any attacker involvement, with the excess credited to the block beneficiary. The latest header can also carry a different fee from this worker's current epoch.

## Changes

- **Epoch fee quotes:** override `eth_gasPrice` with the live worker fee, floored at `MIN_PROTOCOL_BASE_FEE`, and `eth_maxPriorityFeePerGas` with zero. Both handlers use `WorkerBaseFee` to follow worker-slot replacements after worker-count changes. The legacy quote has no headroom because a legacy transaction pays the entire quote.
- **Estimate before pricing:** estimate a missing gas limit before adding EIP-1559 fee defaults. This preserves unpriced requests from unfunded senders and avoids validating the filled cap against a higher pending-header fee after an epoch fee decrease. Preserve reth's chain-ID and blob-field preparation and its missing-sender validation.
- **EIP-1559 defaults:** fill a missing tip with zero and a missing cap with `2 * max(epoch_fee, MIN_PROTOCOL_BASE_FEE) + tip`, using saturating arithmetic. Preserve explicit fee fields. Headroom accommodates epoch fee increases without raising the effective price, although the pool requires balance against the higher cap and arbitrary governance increases can exceed it.
- **Coherent responses:** delegate transaction construction and encoding to reth after correcting the request, so `tx` and decoded `raw` agree on fees. Legacy `gasPrice` requests receive no EIP-1559 defaults; blob defaults retain reth's source. TN rejects blob transactions at admission.
- **Registration and floor contract:** install both handlers on every configured eth transport. The protocol floor equals the pool's default admission minimum; an operator raising `--txpool.minimal-protocol-fee` can reject correctly quoted transactions, so the quote does not guarantee admission under that override.

## Testing

Passed the pinned nightly formatting check and 60 RPC tests under Rust 1.94. The new tests were mutation-confirmed: disabling early estimation reproduces both reported RPC errors; omitting only the HTTP fill override restores the oracle's 1-gwei tip; removing headroom fails the three fee-cap unit assertions while explicit-fee and legacy cases remain green.

```sh
cargo +1.94 test --locked -p tn-reth --lib -j 2 -- rpc
```

Scoped Clippy remains red on existing test-target lints. Running the same pinned-nightly command over all 21 packages emitted by `buildplan`, with `--keep-going --all-targets --no-deps -j 2 -- -D warnings`, reports the same 131 lint diagnostics on main `0d3192a` and this branch. Comparing error text, primary location, and lint name found no new diagnostics and none in `tn-reth`. The seven failing targets are in `e2e-tests`, `tn-storage`, `tn-executor`, `tn-primary`, `tn-engine`, and `tn-network-libp2p`.

Coverage includes live quote updates, the protocol floor, explicit and missing fee fields, fee-cap headroom, unfunded unpriced requests, a pending-header fee above the epoch fee, chain-ID normalization, HTTP-only registration, and agreement between returned fields and raw encoding.

The maintainer PR workflow skips compile/test lanes and requires a separate attestation of the final commit before merge.


crates/tn-reth/src/rpc_fill_transaction.rs
```diff
@@ -0,0 +1,197 @@
...
+{
+    async fn fill_transaction(
+        &self,
+        request: TransactionRequest,
+    ) -> RpcResult<FillTransaction<TransactionSigned>> {
+        // Keep reth's request-trace parity: operators grep this target.
+        tracing::trace!(target: "rpc::eth", ?request, "Serving eth_fillTransaction");
+        let corrected = fill_fee_defaults(request, self.base_fee.base_fee());
+        Ok(EthTransactions::fill_transaction(&self.eth_api, corrected).await?)
+    }
```
grantkee: Filling the fee fields before the delegate makes eth_fillTransaction return an error instead of a filled transaction whenever the canonical tip header's base fee exceeds this worker's epoch base fee.

Reth's `fill_transaction` estimates the gas limit *before* it fills any fee field. `transaction.rs:507-511` runs the estimate under `if request.as_ref().gas_limit().is_none()`, and only then does `transaction.rs:513-527` reach the fee block. Lines 130-131 here invert that: `fill_fee_defaults` sets `max_fee_per_gas` and `max_priority_fee_per_gas` but never touches `request.gas`, so the delegate's estimate still fires — now against a priced request.

That matters because the estimator's fee-cap check is not the one you'd expect to be disabled. The path is `estimate.rs:97` → `call.rs:798-813` → `rpc-convert/src/transaction.rs:444-458` → `alloy-evm/src/rpc/transaction.rs:66-73` → `alloy-evm/src/rpc/fees.rs:63-68`:

```rust
if !(max_fee.is_zero() && max_priority_fee_per_gas.is_zero())
    && max_fee < block_base_fee
{
    return Err(CallFeesError::FeeCapTooLow);
}
```

`estimate.rs:63` does set `evm_env.cfg_env.disable_base_fee = true`, but `ensure_fees` is cfg-blind — it reads only `evm_env.block_env().basefee()` — and it runs at `estimate.rs:97`, before any `evm.transact()`. The flag's only consumer is inside `transact` (`revm-handler/src/validation.rs:128-132`). And `estimate_gas_with` never zeroes `block_env.basefee`, unlike `prepare_call_env` (`call.rs:891-893`) and `eth_simulateV1` (`call.rs:143-144`). That asymmetry is the crux.

The base fee it compares against is the tip header's, not the container's: `BlockId::pending()` → `state.rs:285-287` → `pending_block.rs:94-98` → `payload.rs:192-202`, where `BuildPendingEnv for TNPayload` sets `base_fee_per_gas: parent.base_fee_per_gas.unwrap_or(MIN_PROTOCOL_BASE_FEE)`. So the estimate is validated against exactly the source this handler exists to stop trusting. `--rpc.pending-block none` does not help — `pending_block_kind()` is never consulted in `pending_block_env_and_cfg`.

Reproduced with a genesis header fee of 1000 against a container fee of 7, calling `eth_fillTransaction` through `get_rpc_server` with a bare request from a funded `from` (`cargo test -p tn-reth --lib -- probe_fill --nocapture`):

| probe | tip header fee | container fee | `request.gas` | this branch | pre-fill removed (= `main`) | with the fix below |
|---|---|---|---|---|---|---|
| A | 1000 | 7 | none | **`Err -32000 "max fee per gas less than block base fee"`** | `Ok` cap `0x3b9acde8` tip `0x3b9aca00` | `Ok` cap `0x7` tip `0x0` gas `0x5208` |
| C | 7 | 12345 | none | `Ok` cap `0x3039` tip `0x0` | `Ok` cap `0x3b9aca07` tip `0x3b9aca00` | `Ok` cap `0x3039` tip `0x0` |
| D | 1000 | 7 | 21000 | `Ok` cap `0x7` tip `0x0` | `Ok` cap `0x3b9acde8` tip `0x3b9aca00` | `Ok` cap `0x7` tip `0x0` |

Probe D is the isolating one: pre-setting `request.gas` on this branch unchanged makes the failure disappear *and* lets the epoch-fee cap survive, which localizes the fault to the delegate's internal estimate and nothing else. The `main` column also reproduces #1313 itself — tip `0x3b9aca00` is the oracle's 1-gwei fallback. (Re-run with `init_reth_defaults()` seeded first: identical, so this is independent of the process-global RPC default ordering.)

On when `latest_header.base_fee_per_gas > container.base_fee()` actually happens — one of the triggers I first wrote down does not hold, and one is worse than I thought:

- **Not per-worker fee divergence, in this build.** `start_epoch.rs:458-459` spawns only worker 0, and only worker 0 gets an RPC server (`start_epoch.rs:465`, `engine/inner.rs:166-171`). Fees *are* per-worker (`gas_accumulator.rs:269-285`), so this returns the moment a second worker is spawned — which is what line 17 of this module already flags as forward-looking. Just not reachable today.
- **The epoch boundary, confirmed.** `close_epoch` waits for the closing block to be canonical tip (`run_epoch.rs:719`) and *then* calls `adjust_base_fees` (`:722`). Both container writers (`run_epoch.rs:866-875`, `node.rs:405-410`) strictly precede any epoch-N+1 block, so the tip carries the closing epoch's fee while the container already holds the new one.
- **On an idle chain this is a widening steady state, not a boundary blip.** An empty non-closing output produces no block at all (`payload_builder.rs:112-127` returns `canonical_header` unchanged), and an empty closing block copies the parent's fee verbatim (`:131`). Meanwhile the container ratchets down up to 12.5% per under-target epoch (`gas_accumulator.rs:487-498`). So the tip fee is sticky while the container falls, and the gap widens monotonically each epoch until the container floors — unbounded duration.
- **Present from RPC start-up after a restart.** The entry read (`run_epoch.rs:216-218`) writes the new fee before `initialize_worker_components` builds the RPC server (`start_epoch.rs:484`).
- **Downward fee movement is a first-class production surface.** `Eip1559 { target_gas }` drops up to 12.5% when `gas_used < target_gas`; `Static { fee }` is returned unclamped with no floor (`gas_accumulator.rs:521`, `:567`), and `WorkerConfigs.sol:38` explicitly legalizes `Static { fee: 0 }`. Both are settable via `--worker-fee-config WORKER_ID:STRATEGY:VALUE` at genesis and the `onlyOwner` setters at `WorkerConfigs.sol:135-165`.

I'd call this worth fixing but not urgent, because it is latent on adiri as configured: the shipped default `Eip1559 { target_gas: u64::MAX }` pins every worker at `MIN_PROTOCOL_BASE_FEE` forever (`gas_accumulator.rs:749-761`), so with every fee and every header at 7, `max_fee == basefee` and the check passes. `gas_accumulator.rs:551-554` already records that as an empirical fact about adiri rather than a structural identity. Blast radius is one public RPC method — both new modules are read-only, and `eth_fillTransaction` has no in-repo consumers, so only third-party clients are affected.

One secondary effect from the same reorder: because the pre-fill makes `tx_env.gas_price() = min(max_fee, basefee) > 0`, `estimate.rs:113-120` now runs `caller_gas_allowance` where the bare-request path skipped it, capping the gas limit at `(balance - value)/gas_price` and erroring `InsufficientFundsError` when `balance < value` (`alloy-evm/src/call.rs:46-57`). An under-funded `from` can get `GasRequiredExceedsAllowance` where reth returned a filled transaction — on a method reth deliberately supports without a signer.

Suggested fix: restore reth's ordering — estimate from the unpriced request, then fill the fees, then delegate. The delegate skips its own estimate once `gas` is set, so the priced request never reaches the estimator. This compiles as-is (the call is argument-identical to reth's own `transaction.rs:509`, and `EstimateCall` is already imported at line 43); with it applied, `cargo test -p tn-reth --lib` is 245 passed / 0 failed including this PR's own `test_fill_transaction_prices_from_the_epoch_base_fee`, clippy and fmt are clean, and `raw` still decodes to the same tip/cap as the `tx` fields.

```suggestion
    async fn fill_transaction(
        &self,
        mut request: TransactionRequest,
    ) -> RpcResult<FillTransaction<TransactionSigned>> {
        // Keep reth's request-trace parity: operators grep this target.
        tracing::trace!(target: "rpc::eth", ?request, "Serving eth_fillTransaction");
        // Estimate the gas limit against the UNPRICED request, reth's own order (reth
        // `transaction.rs` estimates before it fills any fee field). The estimator builds its
        // tx env against the pending block env, whose base fee is the canonical tip header's
        // (`crate::payload`'s `BuildPendingEnv`), NOT this worker's epoch fee, and
        // `CallFees::ensure_fees` rejects `max_fee < block_base_fee` before any EVM flag
        // applies: `estimate_gas_with` sets `cfg_env.disable_base_fee` but never zeroes
        // `block_env.basefee`, unlike `prepare_call_env`. Pricing first would therefore fail
        // every bare request whenever the tip's fee exceeds the epoch fee -- a boundary fee
        // cut, or a sticky tip on an idle chain. A priced request also turns on the
        // `caller_gas_allowance` cap, which reth's bare path skips.
        //
        // Filling `gas` here also makes the delegate skip its own estimate, so the priced
        // request is never handed to the estimator.
        if request.gas.is_none() {
            let gas = EstimateCall::estimate_gas_at(
                &self.eth_api,
                request.clone(),
                alloy::eips::BlockId::pending(),
                None,
            )
            .await?;
            request.gas = Some(gas.saturating_to());
        }
        let corrected = fill_fee_defaults(request, self.base_fee.base_fee());
        Ok(EthTransactions::fill_transaction(&self.eth_api, corrected).await?)
    }
```

Worth adding the probe above as a regression test asserting `Ok` with `maxFeePerGas == 7` — it fails on HEAD and passes with the reorder. The existing `test_fill_transaction_prices_from_the_epoch_base_fee` can't catch this, since container 12345 against a genesis fee of 7 only exercises the direction where the cap is *above* the pending base fee.

One alternative I looked at and would avoid: raising the filled cap to `max(epoch_fee, tip_header_base_fee) + tip`. The delegate 2718-encodes the returned transaction from the request (`transaction.rs:529-531`), so a cap lifted off the epoch fee reintroduces exactly the overpayment #1305 and #1313 set out to remove.

The module doc at lines 28-31 needs the companion update — separate comment on that.

crates/tn-reth/src/rpc_fill_transaction.rs
```diff
@@ -0,0 +1,197 @@
...
+//! `eth_maxPriorityFeePerGas` answer), and a missing fee cap becomes the floored epoch
+//! base fee plus the tip (the `eth_gasPrice` answer, preserving reth's `cap >= tip`
+//! construction for an explicit client tip). Correcting the request *before* the
+//! delegate, rather than the response after it like `crate::rpc_fee_history`, keeps the
+//! response coherent: the delegate builds and 2718-encodes the returned transaction
+//! from the request, so the `raw` bytes and the `tx` fields both carry the corrected
+//! fees, where patching the response would desynchronize them. Every other default the
+//! delegate fills (nonce, chain id, gas estimate, value) is fee-independent and passes
+//! through unchanged, as do explicit client fee fields and legacy `gasPrice` requests,
+//! which reth's own handler already leaves untouched.
```
grantkee: The module doc lists the gas estimate among the delegate's fee-independent defaults, but reth's estimator branches on the request's fee fields in two separate places.

The claim is "Every other default the delegate fills (nonce, chain id, gas estimate, value) is fee-independent and passes through unchanged". Nonce, chain id and value hold — `alloy-evm/src/rpc/transaction.rs:76-88` fills `value.unwrap_or_default()` and `chain_id.unwrap_or(cfg.chain_id)`, and `estimate.rs:66` discards the nonce outright. The gas estimate does not.

Two branches gate on the priced request:

1. `alloy-evm/src/rpc/fees.rs:58-68` — the `max_fee >= base_fee` comparison sits inside the `Some(max_fee)` arm, under two conjunctive gates: a cap is present, *and* not both cap and tip are zero. The `None` arm (`fees.rs:83-85`) returns `block_base_fee + tip` with no comparison at all, and a request with no fee fields never reaches that function — it matches `fees.rs:93-101`, which sets `gas_price = U256::ZERO`.
2. `reth/crates/rpc/rpc-eth-api/src/helpers/estimate.rs:113-120` — `if tx_env.gas_price() > 0 { highest_gas_limit = highest_gas_limit.min(self.caller_gas_allowance(...)?) }`.

The regimes, enumerated:

| request | base-fee check | `TxEnv.gas_price` |
|---|---|---|
| no fee fields | not run (`fees.rs:93-101`) | `0` |
| tip only, no cap | not run (`fees.rs:83-85`) | `B + tip` |
| `max_fee = C`, `max_priority = 0` | **run** (`fees.rs:63-68`) → `FeeCapTooLow` if `C < B` | `min(C, B)` |
| `max_fee = 0`, tip `0` | skipped by the zero exception | `0` |

`fill_fee_defaults` always produces `max_fee_per_gas >= 7` and `max_priority_fee_per_gas = Some(tip)`, and the pending base fee is always `>= 7` (`payload.rs:202`), so a TN-filled request lands on row three every time: the base-fee check always runs, and the balance-derived gas cap always runs. Neither runs for the unpriced request reth's own ordering hands to the estimator.

Two things worth knowing that cut the other way, so they don't get re-litigated later. `fees.rs:69-74`'s `TipAboveFeeCap` is unreachable here because `fill_fee_defaults` preserves `cap >= tip`. And `estimate.rs:70`/`:154-159`/`:184-189` gate the `map_out_of_gas_err` retry on `tx_request_gas_price`, which reads the **legacy** `gasPrice` field only (`alloy-network/src/ethereum/builder.rs:78-80`), so filling `max_fee_per_gas` does not flip it. Also worth stating in the doc: the estimate is only reached at all when the request carries no `gas` (`transaction.rs:507`).

If the reorder in the other comment lands, the honest wording is that the estimate runs against the unpriced request precisely *because* reth's estimator is fee-sensitive — the inline comment in that suggestion carries it, and lines 28-31 just need the "gas estimate" item dropped from the fee-independent list:

```
//! fees, where patching the response would desynchronize them. The delegate's remaining
//! defaults (nonce, chain id, value) are fee-independent and pass through unchanged, as do
//! explicit client fee fields and legacy `gasPrice` requests, which reth's own handler
//! already leaves untouched. The gas estimate is not fee-independent, which is why it runs
//! before the fee fill rather than after it.
```

If the ordering is kept instead, the doc has to carry the consequences rather than deny them:

```
//! fees, where patching the response would desynchronize them. The delegate's remaining
//! defaults (nonce, chain id, value) are fee-independent, as are explicit client fee fields
//! and legacy `gasPrice` requests, which reth's own handler already leaves untouched. The gas
//! estimate is not. Reth estimates *before* it fills fees, so upstream a request with no fee
//! fields reaches `estimate_gas_with` with an effective gas price of zero, which skips both
//! `CallFees::ensure_fees`'s `maxFeePerGas >= baseFee` check (enforced only when a cap is
//! present and not both cap and tip are zero) and the caller-balance gas cap reth applies
//! under `if tx_env.gas_price() > 0`. Correcting the request first means the estimate runs
//! against the filled cap instead: the effective gas price becomes the pending block's base
//! fee, so `from` must hold the intrinsic gas at that fee plus `value` or the estimate
//! returns `GasRequiredExceedsAllowance`, and a container fee below the pending base fee --
//! the latest header's -- makes `ensure_fees` reject the request with `FeeCapTooLow` rather
//! than returning a filled transaction. Only requests that omit `gas` reach the estimate.
```

I've left this as a plain block rather than a suggestion, since which wording is right depends on whether the reorder goes in.

crates/tn-reth/src/rpc_fill_transaction.rs
grantkee, line 173: Filling the fee fields before the delegate puts reth's gas estimate on a priced request, which breaks eth_fillTransaction in two ways, one of them on the chain as deployed today.

Reth's `fill_transaction` estimates the gas limit *before* it fills any fee field. `transaction.rs:507-511` runs the estimate under `if request.as_ref().gas_limit().is_none()`, and only then does `transaction.rs:513-527` reach the fee block. Lines 130-131 here invert that: `fill_fee_defaults` sets `max_fee_per_gas` and `max_priority_fee_per_gas` but never touches `request.gas`, so the delegate's estimate still fires — now against a priced request. Reth's estimator branches on the request's fee fields in two places, and the reorder turns both on.

**The live one: the caller-balance gas cap.** `estimate.rs:113-120` gates a balance-derived cap on the effective gas price being non-zero:

```rust
// Check funds of the sender (only useful to check if transaction gas price is more than 0).
if tx_env.gas_price() > 0 {
    highest_gas_limit = highest_gas_limit.min(self.caller_gas_allowance(&mut db, &evm_env, &tx_env)?);
}
```

An unpriced request takes alloy's `(gas_price, None, None, None)` arm and gets `gas_price = 0` (`alloy-evm/src/rpc/fees.rs:93-101`), so upstream this branch never opens. A TN-filled request always carries a cap of at least 7 and a tip, so the effective price is always non-zero and the branch always opens. `caller_gas_allowance` then caps the search at `(balance - value) / gas_price` (`alloy-evm/src/call.rs:46-57`), and a cap below intrinsic gas surfaces as `GasRequiredExceedsAllowance` (`estimate.rs:160-168`). For a zero-balance `from` with no `value` the allowance is `0 / 7 = 0`, and the call fails outright.

This needs no fee divergence at all — it reproduces with container fee and header fee both at 7, which is adiri as configured. `eth_fillTransaction` needs no signer (reth checks only `request.from` at `transaction.rs:472-476`, unlike `send_transaction_request` which calls `find_signer`), so constructing a transaction for an unfunded address is a normal way to call it, and it is registered on every eth-namespace transport.

**The latent one: `FeeCapTooLow`.** `alloy-evm/src/rpc/fees.rs:63-68`:

```rust
if !(max_fee.is_zero() && max_priority_fee_per_gas.is_zero())
    && max_fee < block_base_fee
{
    return Err(CallFeesError::FeeCapTooLow);
}
```

`estimate.rs:63` does set `evm_env.cfg_env.disable_base_fee = true`, but `ensure_fees` is cfg-blind — it reads only `evm_env.block_env().basefee()` — and it runs at `estimate.rs:97`, before any `evm.transact()`. The flag's only consumer is inside `transact` (`revm-handler/src/validation.rs:128-132`). And `estimate_gas_with` never zeroes `block_env.basefee`, unlike `prepare_call_env` (`call.rs:891-893`) and `eth_simulateV1` (`call.rs:143-144`). That asymmetry is the crux.

The base fee it compares against is the tip header's, not the container's: `BlockId::pending()` → `state.rs:285-287` → `pending_block.rs:94-98` → `payload.rs:192-202`, where `BuildPendingEnv for TNPayload` sets `base_fee_per_gas: parent.base_fee_per_gas.unwrap_or(MIN_PROTOCOL_BASE_FEE)`. So the estimate is validated against exactly the source this handler exists to stop trusting. `--rpc.pending-block none` does not help — `pending_block_kind()` is never consulted in `pending_block_env_and_cfg`.

Both reproduce through `get_rpc_server`, and both are fixed by the same reorder (`cargo test -p tn-reth --lib -- probe_fill --nocapture`):

| probe | header fee | container fee | `from` | `value` | `gas` | this branch | pre-fill removed (= `main`) | with the fix below |
|---|---|---|---|---|---|---|---|---|
| **E** | 7 | 7 | unfunded | none | none | **`Err -32000 "gas required exceeds allowance (0)"`** | `Ok` gas `0x5208` cap `0x3b9aca07` | `Ok` gas `0x5208` cap `0x7` tip `0x0` |
| A | 1000 | 7 | funded | 100 | none | **`Err -32000 "max fee per gas less than block base fee"`** | `Ok` cap `0x3b9acde8` tip `0x3b9aca00` | `Ok` cap `0x7` tip `0x0` gas `0x5208` |
| C | 7 | 12345 | funded | 100 | none | `Ok` cap `0x3039` tip `0x0` | `Ok` cap `0x3b9aca07` tip `0x3b9aca00` | `Ok` cap `0x3039` tip `0x0` |
| D | 1000 | 7 | funded | 100 | 21000 | `Ok` cap `0x7` tip `0x0` | `Ok` cap `0x3b9acde8` tip `0x3b9aca00` | `Ok` cap `0x7` tip `0x0` |

Probe E is the live regression — note `header == container == 7`, no divergence required. Probe D is the isolating experiment: pre-setting `request.gas` on this branch unchanged makes the failure disappear *and* lets the epoch-fee cap survive, which localizes both faults to the delegate's internal estimate and nothing else. The `main` column also reproduces #1313 itself — tip `0x3b9aca00` is the oracle's 1-gwei fallback.

On when the `FeeCapTooLow` half becomes reachable, since `latest_header.base_fee_per_gas > container.base_fee()` is a documented design property here (#1262, `txn_pool.rs:14-17`, `rpc_fee_history.rs:22-28`):

- **Not per-worker fee divergence, in this build.** `start_epoch.rs:458-459` spawns only worker 0, and only worker 0 gets an RPC server. Fees *are* per-worker (`gas_accumulator.rs:269-285`), so this returns the moment a second worker is spawned — which is what line 17 of this module already flags as forward-looking. Just not reachable today.
- **The epoch boundary.** `close_epoch` waits for the closing block to be canonical tip (`run_epoch.rs:719`) and *then* calls `adjust_base_fees` (`:722`). Both container writers (`run_epoch.rs:866-875`, `node.rs:405-410`) strictly precede any epoch-N+1 block.
- **On an idle chain this is a widening steady state, not a blip.** An empty non-closing output produces no block at all (`payload_builder.rs:112-127` returns `canonical_header` unchanged) and an empty closing block copies the parent's fee verbatim (`:131`), while the container ratchets down up to 12.5% per under-target epoch (`gas_accumulator.rs:487-498`). The gap widens monotonically each epoch until the container floors.
- **Present from RPC start-up after a restart** — the entry read (`run_epoch.rs:216-218`) writes the new fee before `initialize_worker_components` builds the RPC server (`start_epoch.rs:484`).
- **Downward movement is a first-class surface.** `Eip1559 { target_gas }` drops up to 12.5% when `gas_used < target_gas`; `Static { fee }` is returned unclamped with no floor (`gas_accumulator.rs:521`, `:567`), and `WorkerConfigs.sol:38` legalizes `Static { fee: 0 }`. Both settable via `--worker-fee-config` at genesis and the `onlyOwner` setters at `WorkerConfigs.sol:135-165`. Today the shipped `Eip1559 { target_gas: u64::MAX }` pins every worker at 7 forever (`gas_accumulator.rs:749-761`), which is why this half is latent while probe E's half is not.

Suggested fix: restore reth's ordering — estimate from the unpriced request, then fill the fees, then delegate. The delegate skips its own estimate once `gas` is set, so the priced request never reaches the estimator, and both failures go away together. This compiles as-is (the call is argument-identical to reth's own `transaction.rs:509`, and `EstimateCall` is already imported at line 43); with it applied, `cargo test -p tn-reth --lib` is 245 passed / 0 failed including this PR's own `test_fill_transaction_prices_from_the_epoch_base_fee`, clippy and fmt are clean, and `raw` still decodes to the same tip/cap as the `tx` fields.

```suggestion
    async fn fill_transaction(
        &self,
        mut request: TransactionRequest,
    ) -> RpcResult<FillTransaction<TransactionSigned>> {
        // Keep reth's request-trace parity: operators grep this target.
        tracing::trace!(target: "rpc::eth", ?request, "Serving eth_fillTransaction");
        // Estimate the gas limit against the UNPRICED request, reth's own order (reth
        // `transaction.rs` estimates before it fills any fee field). Reth's estimator
        // branches on the request's fee fields twice, and pricing first turns both on:
        //  - a non-zero effective gas price opens the caller-balance gas cap
        //    (`helpers/estimate.rs`), so an unfunded `from` gets
        //    `GasRequiredExceedsAllowance` on a request that named no fees;
        //  - the cap is validated against the pending block env's base fee, which is the
        //    canonical tip header's (`crate::payload`'s `BuildPendingEnv`), NOT this
        //    worker's epoch fee, so a cap below it trips `CallFees::ensure_fees`'
        //    `FeeCapTooLow` (alloy-evm `rpc/fees.rs`). `estimate_gas_with` sets
        //    `cfg_env.disable_base_fee` but never zeroes `block_env.basefee`, unlike
        //    `prepare_call_env`, so no EVM flag suppresses that check.
        //
        // Estimating on the untouched request reproduces reth's bare-request branch
        // (gas price zero); filling `gas` here then makes the delegate skip its own
        // estimate, so the priced request is never handed to the estimator.
        if request.gas.is_none() {
            let gas = EstimateCall::estimate_gas_at(
                &self.eth_api,
                request.clone(),
                alloy::eips::BlockId::pending(),
                None,
            )
            .await?;
            request.gas = Some(gas.saturating_to());
        }
        let corrected = fill_fee_defaults(request, self.base_fee.base_fee());
        Ok(EthTransactions::fill_transaction(&self.eth_api, corrected).await?)
    }
```

Two regression tests worth adding beside `test_fill_transaction_prices_from_the_epoch_base_fee`, both of which fail on HEAD and pass with the reorder:

1. a `from` that `test_genesis()` does not fund (e.g. `Address::repeat_byte(0x42)`) with `value: None`, asserting `gas == 0x5208` rather than an error;
2. `BaseFeeContainer::new(7)` against a genesis header base fee of 1000, asserting the bare request fills rather than returning `FeeCapTooLow`.

The existing test can't catch either: it funds `from` with `U256::MAX` (`genesis.rs:29-30`), so the allowance `min()` is a no-op, and container 12345 against header 7 only exercises the direction where the cap is *above* the pending base fee.

One alternative I'd avoid: raising the filled cap to `max(epoch_fee, tip_header_base_fee) + tip`. The delegate 2718-encodes the returned transaction from the request (`transaction.rs:529-531`), so a cap lifted off the epoch fee reintroduces exactly the overpayment #1305 and #1313 set out to remove.

crates/tn-reth/src/rpc_fill_transaction.rs
grantkee: The module doc lists the gas estimate among the delegate's fee-independent defaults, but reth's estimator branches on the request's fee fields in two separate places.

The claim is "Every other default the delegate fills (nonce, chain id, gas estimate, value) is fee-independent and passes through unchanged". Nonce, chain id and value hold — `alloy-evm/src/rpc/transaction.rs:76-88` fills `value.unwrap_or_default()` and `chain_id.unwrap_or(cfg.chain_id)`, and `estimate.rs:66` discards the nonce outright. The gas estimate does not, and that is the invariant the other comment on this file turns out to violate — worth correcting here even if the reorder lands, because this sentence is what makes the hazard easy to miss on the next change.

The two branches:

1. `alloy-evm/src/rpc/fees.rs:58-68` — the `max_fee >= base_fee` comparison sits inside the `Some(max_fee)` arm, under two conjunctive gates: a cap is present, *and* not both cap and tip are zero. The `None` arm (`fees.rs:83-85`) returns `block_base_fee + tip` with no comparison, and a request with no fee fields never reaches that function at all — it matches `fees.rs:93-101`, which sets `gas_price = U256::ZERO`.
2. `reth/crates/rpc/rpc-eth-api/src/helpers/estimate.rs:113-120` — `if tx_env.gas_price() > 0 { highest_gas_limit = highest_gas_limit.min(self.caller_gas_allowance(...)?) }`.

The regimes, enumerated:

| request | base-fee check | `TxEnv.gas_price` |
|---|---|---|
| no fee fields | not run (`fees.rs:93-101`) | `0` |
| tip only, no cap | not run (`fees.rs:83-85`) | `B + tip` |
| `max_fee = C`, `max_priority = 0` | **run** (`fees.rs:63-68`) → `FeeCapTooLow` if `C < B` | `min(C, B)` |
| `max_fee = 0`, tip `0` | skipped by the zero exception | `0` |

`fill_fee_defaults` always produces `max_fee_per_gas >= 7` and `max_priority_fee_per_gas = Some(tip)`, and the pending base fee is always `>= 7` (`payload.rs:202`), so a TN-filled request lands on row three every time: the base-fee check always runs, and the balance-derived gas cap always runs. Neither runs for the unpriced request reth's own ordering hands to the estimator.

Two things that cut the other way, so they don't get re-litigated later. `fees.rs:69-74`'s `TipAboveFeeCap` is unreachable here because `fill_fee_defaults` preserves `cap >= tip`. And `estimate.rs:70`/`:154-159`/`:184-189` gate the `map_out_of_gas_err` retry on `tx_request_gas_price`, which reads the **legacy** `gasPrice` field only (`alloy-network/src/ethereum/builder.rs:78-80`), so filling `max_fee_per_gas` does not flip it. Also worth stating in the doc: the estimate is only reached at all when the request carries no `gas` (`transaction.rs:507`).

If the reorder lands, lines 28-31 just need the "gas estimate" item dropped from the list, with a pointer to why the ordering matters:

```
//! fees, where patching the response would desynchronize them. The delegate's remaining
//! defaults (nonce, chain id, value) are fee-independent and pass through unchanged, as do
//! explicit client fee fields and legacy `gasPrice` requests, which reth's own handler
//! already leaves untouched. The gas estimate is NOT fee-independent, which is why this
//! handler estimates before it prices rather than after -- see `fill_transaction`.
```

If the ordering is kept instead, the doc has to carry the consequences rather than deny them:

```
//! fees, where patching the response would desynchronize them. The delegate's remaining
//! defaults (nonce, chain id, value) are fee-independent, as are explicit client fee fields
//! and legacy `gasPrice` requests, which reth's own handler already leaves untouched. The gas
//! estimate is not. Reth estimates *before* it fills fees, so upstream a request with no fee
//! fields reaches `estimate_gas_with` with an effective gas price of zero, which skips both
//! `CallFees::ensure_fees`'s `maxFeePerGas >= baseFee` check (enforced only when a cap is
//! present and not both cap and tip are zero) and the caller-balance gas cap reth applies
//! under `if tx_env.gas_price() > 0`. Correcting the request first means the estimate runs
//! against the filled cap instead: the effective gas price becomes the pending block's base
//! fee, so `from` must hold the intrinsic gas at that fee plus `value` or the estimate
//! returns `GasRequiredExceedsAllowance`, and a container fee below the pending base fee --
//! the latest header's -- makes `ensure_fees` reject the request with `FeeCapTooLow` rather
//! than returning a filled transaction. Only requests that omit `gas` reach the estimate.
```

Left as a plain block rather than a suggestion, since which wording is right depends on whether the reorder goes in.

crates/tn-reth/src/rpc_gas_price.rs
```diff
@@ -0,0 +1,128 @@
...
+#[async_trait]
+impl EpochGasPriceServer for GasPriceWithEpochBaseFee {
+    async fn gas_price(&self) -> RpcResult<U256> {
+        // Keep reth's request-trace parity: operators grep this target.
+        tracing::trace!(target: "rpc::eth", "Serving eth_gasPrice");
+        // Floor at the pool's admission minimum: reth rejects any fee cap below
+        // `minimal_protocol_basefee` (7 wei, a config TN keeps), and a
+        // `WorkerFeeConfig::Static` row is not clamped, so a sub-floor governance fee
+        // would otherwise quote a price this node's own pool refuses.
+        Ok(U256::from(self.base_fee.base_fee().max(MIN_PROTOCOL_BASE_FEE)))
```
grantkee: This floor is the MIN_PROTOCOL_BASE_FEE constant, but the pool enforces the operator-settable txpool.minimal-protocol-fee, and the comment asserts the two are the same thing.

The comment says the floor is "the pool's admission minimum: reth rejects any fee cap below `minimal_protocol_basefee` (7 wei, a config TN keeps)". The enforcement side is right — `txpool.rs:1997-1999` is strict and origin-independent, sitting in `AllTransactions::insert_tx` with no `is_local` branch on the path:

```rust
if fee_cap < self.minimal_protocol_basefee as u128 {
    return Err(InsertErr::FeeCapBelowMinimumProtocolFeeCap { transaction, fee_cap })
}
```

But `minimal_protocol_basefee` is not a constant TN keeps — it is a live CLI flag that reaches the pool through an unbroken chain with no clamp. `--txpool.minimal-protocol-fee` is declared at reth `args/txpool.rs:338-339` as a public field on `TxPoolArgs`, TN flattens `TxPoolArgs` wholesale (`cli.rs:80-81`), destructures and re-inserts it verbatim (`cli.rs:259`, whose comment says "Nothing to override here"), and `txn_pool.rs:160` hands `node_config.txpool.pool_config()` to `Pool::eth_pool`, carrying the value through `args/txpool.rs:535` → `txpool.rs:1386`. The flag has no `value_parser` range, and a repo-wide grep for `minimal_protocol` across `crates/` returns exactly one hit: this comment. Nothing pins it.

At the default the two values are equal, since reth defaults the flag to `MIN_PROTOCOL_BASE_FEE` (`args/txpool.rs:267`) — so this is latent, not live. Raise the flag above the epoch fee and the quote reports a price this node's own pool refuses at insertion.

I'd hold this at Low and I'm not sure it needs a code change, for two reasons. An operator who raises this flag above the epoch fee has already broken the node comprehensively: the pool then refuses *every* honestly-priced transaction at insertion — from RPC, from the observer forwarder (`forward.rs:171` already documents that error), and from gossip ingress (`batch-validator/src/validator.rs:101`) — and the worker builds empty batches. The quote is a symptom, not the failure. And there's a defensible argument the current behaviour is more correct anyway: `eth_gasPrice` should report the chain's fee level, not one node's local admission policy, since following a raised local floor would make clients overpay against every other node.

So the narrow real defect is the comment: two independent values are asserted equal with nothing enforcing it. Three ways to resolve it, in increasing cost:

1. **Fix the wording only** — say the floor is the protocol constant, which equals the pool's default admission minimum, and that a node overriding `--txpool.minimal-protocol-fee` upward is out of contract.
2. **Add a startup guard** beside `RethConfig::ensure_archive_mode` (`cli.rs:510`) that refuses or warns when `txpool.minimal_protocol_basefee != MIN_PROTOCOL_BASE_FEE`. TN's fee model has no use for a raised floor and setting one bricks the pool, so rejecting it outright is defensible and makes the comment true by construction.
3. **Plumb the config.** It is already reachable at the construction site — `RethEnv::node_config()` is `pub(crate)` (`env/mod.rs:400`) and is already called three times inside `get_rpc_server`, and both `NodeConfig::txpool` and `TxPoolArgs::minimal_protocol_basefee` are public:

```rust
let min_pool_fee = self.node_config().txpool.minimal_protocol_basefee;
let fill_transaction =
    FillTransactionWithEpochBaseFee::new(eth_api.clone(), base_fee.clone(), min_pool_fee);
let gas_price = GasPriceWithEpochBaseFee::new(base_fee, min_pool_fee);
```

carrying `min_pool_fee: u64` on both structs and using `.max(self.min_pool_fee)` here and at `rpc_fill_transaction.rs:98`.

My preference is 2, since it makes the invariant real rather than tracking a value that should never move. 1 is fine if you'd rather not add a guard.

crates/tn-reth/src/rpc_fill_transaction.rs
```diff
@@ -0,0 +1,197 @@
...
+///
+/// Filling the tip also covers the request that carries an explicit `maxFeePerGas`
+/// but no tip: reth would still ask the oracle for the tip it writes into the built
+/// transaction, possibly above the client's own cap.
+fn fill_fee_defaults(request: TransactionRequest, epoch_fee: u64) -> TransactionRequest {
+    match () {
+        () if request.gas_price.is_some() => request,
+        () => {
+            let tip = request.max_priority_fee_per_gas.unwrap_or_default();
+            let fee_cap = u128::from(epoch_fee.max(MIN_PROTOCOL_BASE_FEE)).saturating_add(tip);
```
grantkee: The filled fee cap has no headroom over the epoch fee, and unlike the legacy eth_gasPrice quote, headroom here would cost the sender nothing.

Worth separating the two quote paths, because they have opposite economics and it's easy to treat them as one decision. `TNEvmHandler::reward_beneficiary` (`evm/handler.rs:167-177`) credits `effective_gas_price - basefee` to the beneficiary and burns nothing, and revm splits `effective_gas_price` by transaction type:

- **legacy** (`eth_gasPrice`'s consumer): `effective_gas_price == gas_price`, no cap-and-refund. Headroom there is money the sender actually pays out to the beneficiary — exactly the harm #1305 exists to remove, so `rpc_gas_price.rs` is right to quote the bare epoch fee and should stay that way.
- **EIP-1559** (this cap, with the tip filled to zero): `effective_gas_price == min(max_fee, base_fee + tip)`. With `tip == 0` that is `base_fee`, so raising the cap changes nothing the sender pays. It only widens the window in which the transaction is still admissible.

That asymmetry means the headroom question has a free answer on this line and an expensive one on the other, and only this one is worth taking.

Why it matters: the pool admits at the container fee (`txn_pool.rs:436`), and every sub-pool seam uses a non-strict comparison, so a cap exactly equal to the base fee is fine — `txpool.rs:2000` (`fee_cap >= pending_fees.base_fee`), `txpool.rs:1639-1646` (`Greater | Equal`), `pending.rs:230` and `parked.rs:278` (both strict `<` for removal). So this only bites on a strict fee *increase*. But when it does, the transaction is accepted and then parked in the basefee sub-pool rather than rejected, it is not batched while parked (`batch-builder/src/batch.rs:60` iterates the pending sub-pool only), and the resulting nonce gap parks that sender's later transactions too (`pool/state.rs:17-21`). Since TN's fee is flat within an epoch and steps at the boundary, the exposure is one step per epoch rather than per block — but a full 12.5% step for `Eip1559`, or an arbitrary jump for `Static`.

Suggested, if you want it:

```rust
/// Headroom multiplier for the filled EIP-1559 fee cap.
///
/// An `Eip1559` worker's fee moves by at most +12.5% per epoch
/// (`compute_next_base_fee_eip1559`), so 2x survives five consecutive up-steps
/// (1.125^5 = 1.80, 1.125^6 = 2.03). The headroom costs the sender nothing: revm's
/// `effective_gas_price` for an EIP-1559 transaction is `min(max_fee, base_fee + tip)`,
/// so with the tip filled to zero the sender still pays exactly the epoch base fee.
/// The legacy `eth_gasPrice` quote deliberately carries no equivalent headroom, because
/// there `effective_gas_price == gas_price` and the surplus is really paid out.
const FILL_FEE_CAP_HEADROOM: u128 = 2;

let fee_cap = u128::from(epoch_fee.max(MIN_PROTOCOL_BASE_FEE))
    .saturating_mul(FILL_FEE_CAP_HEADROOM)
    .saturating_add(tip);
```

This also aligns the fill path with what viem and ethers already derive from `eth_feeHistory` (`2 * baseFee + tip`), which this repo already corrects in `rpc_fee_history`.

One cost to note rather than hide: `ValidPoolTransaction::cost()` is `gas_limit * max_fee + value`, so doubling the cap doubles the balance needed for the pool's `ENOUGH_BALANCE` bit. At 7 wei against a 30M gas limit that moves 0.21 gwei to 0.42 gwei, which is negligible, but it scales with the epoch fee and is worth a line in the doc.

Entirely reasonable to decline this and instead just document why the legacy quote carries no headroom — the current behaviour is not wrong, and it is strictly better than upstream's, where `eth_gasPrice` returns `bf(N) + tip` while the pool admits at `bf(N+1)`, so upstream headroom goes negative whenever the base fee exceeds eight times the tip. I'd rather the reasoning be written down than the constant be added blindly.

crates/tn-reth/src/rpc_fill_transaction.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+/// Fill the request's missing EIP-1559 fee fields from the worker's epoch base fee.
+///
+/// Mirrors the fee block of reth's `fill_transaction` with TN's sources: a missing tip
+/// becomes zero instead of the oracle's suggestion, and a missing fee cap becomes the
+/// floored epoch base fee plus the tip instead of the latest header's base fee plus
+/// the tip. The floor is the pool's admission minimum, the same clamp
+/// `crate::rpc_gas_price` documents. Explicit client fields survive, and a legacy
+/// `gasPrice` request passes through whole: reth's handler fills no fee field for it.
```
grantkee: This says reth's handler fills no fee field for a legacy gasPrice request, but it does fill the blob fee for one.

Reth's blob-fee default is gated only on the request carrying EIP-4844 fields, and it sits *before* the `gas_price().is_none()` gate that guards the EIP-1559 fee block (`transaction.rs:492-497`):

```rust
if request.as_ref().has_eip4844_fields() &&
    request.as_ref().max_fee_per_blob_gas().is_none()
{
    let blob_fee = self.blob_base_fee().await?;
    request.as_mut().set_max_fee_per_blob_gas(blob_fee.to());
}
```

So a request carrying both `gasPrice` and `blob_versioned_hashes` does get `max_fee_per_blob_gas` filled, from `LoadFee::blob_base_fee` rather than from anything this module controls. "Passes through whole" isn't quite true for that shape.

Purely a wording thing — the module docs at lines 33-36 already record the blob non-intercept as deliberate, and #1159 refuses EIP-4844 at pool admission, so nothing behaves wrongly. It's an internal inconsistency between two comments in the same file. Narrowing both this line and the test doc at lines 181-182 to "fills no EIP-1559 fee field for it" would make them agree with lines 33-36.

crates/tn-reth/src/env/rpc.rs
```diff
@@ -372,6 +399,64 @@ mod tests {
         assert_eq!(pool.pool_size().pending, 0);
     }
 
+    /// `eth_fillTransaction` prices a bare request from the worker's epoch base fee:
+    /// tip zero and fee cap the container value, not the oracle's suggestion, in both
+    /// the returned `tx` fields and the `raw` encoding. An explicit client tip
+    /// survives and raises the filled cap by itself.
+    #[tokio::test]
+    async fn test_fill_transaction_prices_from_the_epoch_base_fee() {
```
grantkee: This test only exercises the default IPC-only registration, so it cannot catch an HTTP registration that missed the fill override.

`methods_by` unions the transports and keeps the first occurrence, which is exactly why the sibling override added `test_http_transport_gets_the_gas_price_quotes` (line 798) alongside its default-registration test, and why the fee-cap suite has `test_http_transport_gets_the_fee_cap_guard`. Its doc comment states the reason directly: the IPC-only default "could mask an HTTP registration that missed the override, and HTTP is the transport public clients actually price against". `eth_fillTransaction` is reachable by any RPC client with no signer, so HTTP is the transport that matters for it too, and it's the one override of the four with no HTTP assertion.

The install is structural, so I don't think there's a live bug here — the same `modules_config` drives the module build and all four replacements. It's the regression-coverage asymmetry that's worth closing, and it's cheap: the existing `gas_price_methods` helper already takes `RpcServerArgs`, so a fill equivalent is a few lines.

```rust
let rpc_args = reth::args::RpcServerArgs { http: true, ipcdisable: true, ..Default::default() };
```

Same test also constructs `reth::args::RpcServerArgs::default()` without first calling `init_reth_defaults()`. Both helpers that build args this way call it first (lines 608 and 718) and `gas_price_methods` documents why — a `Default`-constructed `RpcServerArgs` reads reth's process-global defaults, and the first builder in the test process locks them (#1165). I checked and it changes nothing about this test's outcome either way, so it's your call whether to add the call for consistency or drop the comment from the two helpers; right now the three sites disagree about whether the seeding is necessary.

grantkee (changes requested):
One blocker regarding `fill_transaction` filling fee fields before the delegate and a handful of suggestions

crates/tn-reth/src/rpc_fill_transaction.rs
MavenRain: Restored estimate-before-fees ordering while preserving reth's chain-ID/blob preparation and missing-sender validation. Added regressions for unfunded senders and a 1000-wei header against a 7-wei epoch fee, with tx/raw assertions. Both pass and reproduce the original errors when the reorder is removed.

crates/tn-reth/src/rpc_fill_transaction.rs
MavenRain: Corrected the module docs to explain that gas estimation depends on fee fields and must precede pricing. They now describe both the caller-balance cap and pending-header fee check, and note that explicit gas skips estimation.

crates/tn-reth/src/rpc_gas_price.rs
MavenRain: Took the wording-only option: the quote and fill docs now distinguish MIN_PROTOCOL_BASE_FEE from the operator-configurable pool minimum. They state that the values match by default and a raised --txpool.minimal-protocol-fee is outside the quote's admission guarantee. The quote continues to reflect the chain fee.

crates/tn-reth/src/rpc_fill_transaction.rs
MavenRain: Added 2x headroom to missing EIP-1559 caps, plus the client's tip, with saturating arithmetic. Explicit caps remain unchanged. Documented the higher balance requirement and why eth_gasPrice retains the bare epoch fee; updated unit and registration assertions for the new cap.

crates/tn-reth/src/rpc_fill_transaction.rs
MavenRain: Narrowed the helper and test docs to say legacy gasPrice requests receive no EIP-1559 defaults. The module also clarifies that blob defaults retain reth's helper and may still be filled.

crates/tn-reth/src/env/rpc.rs
MavenRain: Added HTTP-only fill coverage with http enabled and IPC disabled, asserting gas, chain ID, tip, and cap in both tx fields and decoded raw bytes. Refactored the shared setup and seeded reth defaults before constructing RpcServerArgs in each fill test.

grantkee (approved):
LGTM!

PR #1316, opened 2026-08-30
fix(tn-reth): re-arm the residual dirty-sender reload between canonical notifications

### LINKED ISSUE #1304: pool resync: residual dirty senders drain only at notification rate after a lag event
## Problem

`WorkerTxPool::reload_dirty_accounts` has one production call site: the maintenance loop in `spawn_maintenance_task` (`crates/tn-reth/src/txn_pool.rs`).  The loop reloads one chunk of `MAX_RELOAD_ACCOUNTS` (100) dirty senders per incoming canonical-state notification.  There is no interval, tick, or backup branch.

After a `Lagged` event the broadcast ring still holds 256 buffered `Commit`s, so the first ~25,700 senders drain back-to-back with no waiting.  The residual beyond that advances only at consensus-round cadence (`min_header_delay` 1000 ms / `max_header_delay` 2500 ms).  At the theoretical maximum of ~30,000 dirty senders (3 reachable subpools x 10,000 default subpool cap;  the blob pool is unreachable) the tail takes roughly 45-110 s.  If transaction volume spikes (which is what builds the large dirty set) and then goes very quiet, notifications slow down and the residual drain stalls with them.

Raised in review of #1243: "consider re-arming reload dirty accounts for residual drains in the event tx volume spikes then is very low."

## Why this is a follow-up rather than part of #1243

- The harm is bounded and self-healing.  Drifted senders self-clear through the batch builder's post-quorum `update_canonical_state`;  mined-but-still-pending transactions are executed and skipped, not fatal.  Worst case is tens of seconds of stale pool view and wasted batch space.
- The fix restructures the maintenance loop from `while let` over the stream into a `tokio::select!` over the stream plus a retry interval.  That changes shutdown and stream-end semantics, needs its own deterministic tests (paused-time interval tests), and interacts with the blocking-pool offload added in #1243's review round (the interval arm must not re-chunk while a reload is already in flight).

## Proposed fix

Add a `RELOAD_RETRY_INTERVAL` (around 100 ms) and re-arm the reload from a `select!` branch whenever `dirty_addresses` is nonempty, so the residual drains at reload speed instead of round speed.  This mirrors reth's `maintain_transaction_pool`, which re-arms `reload_accounts_fut` at the top of every loop iteration and keeps a `stale_eviction_interval` tick as an independent liveness path (`maintain.rs`).

MavenRain:
Closes #1304.

Follow-up to #1243 (merged): the fix restructures the maintenance loop that #1243 added, including the blocking-pool offload from its review round.  The branch has `origin/main` merged in, so this PR targets `main`.

## Problem

`WorkerTxPool::reload_dirty_accounts` has one production call site: the maintenance loop in `spawn_maintenance_task` (`crates/tn-reth/src/txn_pool.rs`).  The loop reloaded one chunk of `MAX_RELOAD_ACCOUNTS` (100) dirty senders per incoming canonical-state notification, with no interval, tick, or backup branch.

After a `Lagged` event the broadcast ring still holds up to 256 buffered `Commit`s, so the first chunks drain back-to-back.  The residual beyond that advanced only at consensus-round cadence (`min_header_delay` 1000 ms / `max_header_delay` 2500 ms).  At the theoretical ~30,000 dirty-sender maximum the tail took roughly 45-110 s, and if transaction volume spiked (which is what builds the large dirty set) and then went very quiet, notifications slowed down and the residual drain stalled with them.

No attacker is required: this is a liveness gap in recovery, and the harm is bounded and self-healing (drifted senders self-clear through the batch builder's post-quorum `update_canonical_state`;  mined-but-still-pending transactions are executed and skipped).  The cost is tens of seconds of stale pool view and wasted batch space at the worst case.  Raised in review of #1243.

## Fix

- [x] `crates/tn-reth/src/txn_pool.rs`: add `RELOAD_RETRY_INTERVAL` (100 ms) and merge the canonical-state stream with a tick stream (`futures::stream::select` over a `MaintenanceEvent` enum: `Update` / `RetryTick` / `Closed`).  A tick re-arms the chunk reload whenever dirty senders remain, so the residual drains at one chunk per interval (about 30 s at the ~30,000-sender theoretical maximum, ~4 s for the realistic post-ring residual) independent of notification traffic:  a 10-25x improvement over consensus-round cadence, with the interval deliberately throttling recovery DB pressure.  This adapts reth's `maintain_transaction_pool`, which re-arms its `reload_accounts_fut` at the top of every loop iteration;  here the re-arm is paced by an explicit interval.  A tick with no dirty senders is a no-op.
- [x] **No overlapping reloads**: the chunk reload stays awaited inline in the loop, so a tick can never start a second reload while one is in flight (the `spawn_blocking` offload from the #1243 review round is preserved).  Ticks that would fire during a reload are pushed back a full interval (`MissedTickBehavior::Delay`).
- [x] **Shutdown semantics preserved**: a chained `Closed` sentinel plus `take_while` ends the merged stream exactly when the notification stream closes, so the critical task still exits with the same `TaskError` as before.
- [x] **Testable seam**: the loop body moves to `WorkerTxPool::maintain_pool(state_stream, retry_interval, max_reload)`;  `spawn_maintenance_task` delegates to it with the production constants.

## Testing

Three new tests drive `maintain_pool` directly under `#[tokio::test]` on real time, polling to convergence with a bounded budget (600 polls at 50 ms under a 60 s wall-clock timeout;  no exact-timing assertions;  aborted maintenance tasks are joined before teardown).  Paused time is not usable here:  the `RethEnv` harness keeps `spawn_blocking` tasks alive for its whole life (`TaskSpawner::spawn_reth_task` wraps `Handle::block_on`), and tokio inhibits paused-clock auto-advance while a blocking task is in flight, so every timer waits forever (caught locally as a multi-hour test hang;  a comment in the test module records the constraint):

- [x] `test_residual_dirty_senders_drain_between_notifications`: three funded senders mined in one block while the pool is not subscribed (the drift state), then a synthetic stream of one `Lagged` marker followed by silence, chunk size 1.  The negative control asserts the drift is real;  the residual two senders can then drain only via retry ticks.  Also asserts the silent stream keeps the loop alive.
- [x] `test_maintenance_loop_ends_when_state_stream_closes`: an empty stream ends the loop with the `state_stream closed` task error despite the never-ending tick stream (kills the sentinel / `take_while` wiring).
- [x] `test_maintenance_loop_applies_canonical_notifications`: a loop subscribed to the real canonical-state broadcast applies a live `Commit` and removes the mined transaction (the notification arm survived the restructuring).

Pre-existing lag-resync tests (`test_lag_resync_discards_transactions_mined_in_lost_rounds`, `test_reload_dirty_accounts_is_bounded`) still pass.  Local gates: nightly `fmt --check`, `check -p tn-reth --all-targets`, `clippy -p tn-reth --no-deps --all-targets`, targeted test run.  Each new test was mutation-confirmed (tick stream silenced, `take_while` removed, notification apply dropped: each mutation fails exactly its test;  fix restored, all green).  CI and the pinned 1.94 attest run on push.


crates/tn-reth/src/txn_pool.rs
```diff
@@ -227,57 +271,99 @@ impl WorkerTxPool {
     /// `Stream` shape but surfaces `Lagged` as an error item, so the task can mark every
     /// pool sender dirty and reload canonical account state in bounded chunks, mirroring
     /// reth's `maintain_transaction_pool` drift recovery.
+    ///
+    /// The loop body lives in [`Self::maintain_pool`] so tests can drive it with a
+    /// synthetic stream and paused time.
```
grantkee: The doc comment here says tests can drive maintain_pool with paused time, but no test does and none can.

The three issue #1304 tests are plain `#[tokio::test]` on real time — `test_residual_dirty_senders_drain_between_notifications` (`txn_pool.rs:1029`), `test_maintenance_loop_ends_when_state_stream_closes` (`:1110`), and `test_maintenance_loop_applies_canonical_notifications` (`:1137`) — and the test-module comment at `:1023-1028` explains why: the `RethEnv` harness holds a `spawn_blocking` task for its whole life (`crates/types/src/task_manager.rs:253-256` wraps it as `spawn_blocking(move || handle.block_on(f))`), and tokio inhibits paused-clock auto-advance while a blocking task is in flight. Grepping the crate, the only `start_paused` mention in this file is inside that very comment saying paused time cannot be used, and there is no `tokio::time::pause()` anywhere in it.

The "synthetic stream" half is accurate — two of the three tests use `stream::iter` (`:1087`, `:1120`) — so it is just the "and paused time" that is wrong. It seems worth more than the usual doc nit because line 276 points the next author straight at `#[tokio::test(start_paused = true)]`, which is the multi-hour hang the PR description records hitting locally. The explanation already exists 748 lines further down; it just is not where someone reading the seam's own doc will look.

```suggestion
    /// The loop body lives in [`Self::maintain_pool`] so tests can drive it with a
    /// synthetic stream. Tests run on real time: the `RethEnv` harness holds a
    /// `spawn_blocking` task for its whole life, which inhibits tokio's paused-clock
    /// auto-advance (see the test module comment).
```

crates/tn-reth/src/txn_pool.rs
```diff
@@ -70,9 +77,46 @@ pub use reth_primitives_traits::InMemorySize as TxnSize;
...
+/// After a `Lagged` event the broadcast ring still holds up to its capacity in buffered
+/// `Commit`s, so the first reload chunks drain back-to-back. The residual beyond that used
+/// to advance only when the next notification arrived: consensus-round cadence at best, and
+/// a post-spike lull (exactly what follows the volume spike that builds a large dirty set)
+/// stalls it entirely. Re-arming on this interval drains the residual at one chunk per
+/// interval instead, about 30 s at the theoretical ~30,000-dirty-sender maximum, independent
+/// of notification traffic; the interval is a deliberate throttle on recovery DB pressure.
+/// This adapts reth's `maintain_transaction_pool`, which re-arms its `reload_accounts_fut`
+/// on every loop iteration; here the re-arm is paced by an explicit interval (issue #1304).
+const RELOAD_RETRY_INTERVAL: Duration = Duration::from_millis(100);
```
grantkee, line 97: The new 100 ms retry interval multiplies an existing unconditional warn! in the reload failure path from roughly 1 to 10 lines per second.

The line I am actually asking about is `crates/tn-reth/src/txn_pool.rs:461-467`, the aggregated `warn!` in `reload_dirty_accounts`, which this PR does not touch — commenting here because `RELOAD_RETRY_INTERVAL` is what changes its rate.

When `self.1.latest()` fails, the dirty set does not shrink: `:439-448` maps the whole chunk to `Err(*address)` and `:469` returns `pending.chain(failures).collect()`, so input size equals output size. The tick re-arms every 100 ms regardless, and the `warn!` at `:462-467` fires once per failing iteration with no dedup, rate limit, or sampling. Checking tokio 1.50's `interval.rs:462-494`, the deadline resets to `now + period` at fire time, and on this path the iteration is near-instant (`latest()` fails fast, and `update_accounts` is skipped because `accounts` is empty at `:451`), so the full 10 Hz is reached. Before this PR the same failure retried at notification cadence — one `Commit` per consensus round, 0.4-1.0 Hz given `min_header_delay` 1000 ms / `max_header_delay` 2500 ms (`crates/config/src/node.rs:336-342`).

Both the `warn!` at `:462` and the `debug!` at `:440` reach the log file, since the file filter defaults to `debug`:

| | Before | After |
|---|---|---|
| Log lines/sec (file) | ~2 | ~20 |
| WARN lines/sec (stdout) | ~1 | ~10 |
| Log bytes/day | ~35 MB | ~350 MB |

Rotation caps this at 200 MB x 5 files, so it cannot fill a disk. The cost is churning the ~1 GB log ring in about 3 days instead of 30, which is where post-mortem history lives.

Worth being clear about what this is not, since I went looking for worse and did not find it. The precondition is narrow: it needs a prior broadcast `Lagged` *and* a persistent `latest()` failure, and in steady state the tick is a pure no-op skipped at `:341-342`. A persistent `latest()` failure means MDBX is broken, which every other consumer already treats as fatal. The metric cost is nil — `:456-459` is one `fetch_add(failures.len())`, not one call per address. And the interval is a throttle relative to upstream: reth's `maintain.rs:205-233` re-arms the same reload with no interval at all. The only genuine delta is the log level.

So I would match upstream and let the counter carry the alert, which `crates/tn-reth/src/metrics.rs:148-154` already designates it to do ("Alert on a sustained rate, not on any nonzero value"). Reth logs the equivalent line at `debug!` (`maintain.rs:310`):

```rust
// One aggregated line per iteration. Emitted at DEBUG, not WARN: a persistent
// provider failure re-emits this every RELOAD_RETRY_INTERVAL (10 Hz), and the
// alertable signal is the sustained rate of
// `canon_state_resync_read_failures_total`, not the individual line.
// Matches reth's `maintain_transaction_pool` ("failed to load accounts").
debug!(
    target: "txpool",
    failed = failures.len(),
    "canonical account reads failed during pool resync; senders stay dirty for retry"
);
```

Capped backoff (100 ms -> 3.2 s over six consecutive all-failed iterations, reset on any success) would also drop the useless `spawn_blocking` dispatches, but that is more machinery than this needs.

grantkee (dismissed):
LGTM - 2 small nits for your consideration

Haxatron:
The fix for https://github.com/Telcoin-Association/telcoin-network/issues/1304 LGTM and has addressed the comment Grant raised by changing to from warn to debug.

grantkee (approved):
Thanks @MavenRain and @Haxatron !

PR #1317, opened 2026-08-30
fix(metrics): count inbound sync stream shedding by reason (#1307)

### LINKED ISSUE #1307: metric: Inbound sync stream shedding
## Problem

PR #1265 bounded the shed-task fan-out for denied inbound sync streams on the worker. Admission already ran before the spawn, but a denied stream still got its own task, so an over-cap substream burst fanned out one task per stream. The spawn is now gated by a `MAX_CONCURRENT_SHED_TASKS` budget of 8: past that budget the stream is dropped without spawning, and the requester never receives `Deny(AtCapacity)`.

The entire observability for that new drop path is a single `debug!` in `shed_inbound_sync_stream` (`crates/consensus/worker/src/network/mod.rs:446`). The default tracing filter is `info` (`bin/worker-gateway/src/cli.rs:209`), so in a default deployment the condition is invisible on the responder side.

The requester loses the signal too. When the shed task runs, the fetch loop logs `peer denied sync batch request: AtCapacity` (`crates/consensus/worker/src/network/handle.rs:428`). When the budget is exhausted and the stream is dropped instead, it logs `failed to read sync ack frame: unexpected end of file` (`handle.rs:417`), which is indistinguishable from a transport flap or a crashed peer. The `Deny` frame is what turns backpressure into a legible signal, and losing it converts that signal into generic I/O noise exactly when an operator most needs to see it. Instrumenting the responder is the only way to recover it.

No existing counter covers this. `WorkerMetrics` (`crates/consensus/worker/src/metrics.rs`) counts batches sealed, quorum waits and failures, report retries, validation failures, batches fetched, fetch duration, and dropped forwards. There is no admission, shed, or task-spawn counter anywhere in it. `ForwardDropReason` and `record_forward_dropped` look reusable, but the doc comment scopes them to observer transaction forwarding (#1133) and the only call sites are in `disburse_txns`. Reusing that enum here would be semantically wrong; the reusable part is the pattern, a labeled reason counter, not the enum. One level down, `crates/network-libp2p/src/metrics.rs` defines `SwarmMetrics` and `PeerManagerMetrics` with no stream metrics at all, and the inbound rate-limit drop at `stream/behavior.rs:329-337` is likewise counterless.

To be fair about scope: metric-less shedding is the established norm here, not the exception. The sibling `debug!` on the deny arm has no metric, the gossip-prefetch shed at `crates/consensus/worker/src/network/handler.rs:185-191` is a silent `return Ok(())` with no log and no metric, and both primary shed paths are the same. The gap is pre-existing policy rather than a regression, and the new drop is the least silent of the family. What makes it worth closing now is that shedding is no longer a rare edge case: with a global cap of 5 concurrent batch streams and a per-peer cap of 2, denial is the common case during catch-up, and the budget-exhausted arm is a condition an operator has no way to observe at all.

## Solution

Add a dedicated counter rather than overloading `ForwardDropReason`. In `crates/consensus/worker/src/metrics.rs`:

```rust
/// Why an inbound sync stream was refused without being served.
#[derive(Clone, Copy, Debug)]
pub(crate) enum SyncShedReason {
    /// Admission caps hit; the requester got `Deny(AtCapacity)`.
    Denied,
    /// Shed budget exhausted; the stream was closed with no reply.
    BudgetExhausted,
}

impl SyncShedReason {
    const fn label(self) -> &'static str {
        match self {
            Self::Denied => "denied",
            Self::BudgetExhausted => "budget_exhausted",
        }
    }
}

impl WorkerMetrics {
    /// Record an inbound sync stream refused without being served.
    pub(crate) fn record_sync_stream_shed(&self, reason: SyncShedReason) {
        metrics::counter!(
            "tn_worker.sync_streams_shed_total",
            "worker" => self.worker_id.to_string(),
            "reason" => reason.label(),
        )
        .increment(1);
    }
}
```

Plumbing is cheap. `WorkerNetwork::new` already receives `id: WorkerId`, and `metrics.rs` documents that components constructing `WorkerMetrics` independently for the same worker share the same underlying series because the registry identifies series by name and labels. So this is one new field on `WorkerNetwork` plus one constructor line, with no signature change: initialize with `WorkerMetrics::new_for_worker(id)` and call `record_sync_stream_shed` beside each of the two `debug!` lines in `shed_inbound_sync_stream`.

Instrument both arms, not just the new one. A single `denied` counter answers "are we shedding", but the pair answers "are we shedding fairly". The shed budget has no per-peer sub-cap even though the admission path it guards has one, so a nonzero `budget_exhausted` rate while `denied` stays low is the signature of a peer squatting on shed slots and starving other peers of their deny replies. Nothing else in the system would surface that.

The primary has the same two arms once the shed budget is mirrored there, so the same counter belongs in `PrimaryMetrics` at that point.

MavenRain:
Closes #1307.

## Problem

`WorkerNetwork::shed_inbound_sync_stream` refuses inbound sync streams in two ways, both on stream open before any request frame is read: when the admission caps hit it spawns a shed task that writes `Deny(AtCapacity)`, and when the shed budget (`MAX_CONCURRENT_SHED_TASKS`, #1265) is exhausted it drops the stream with no reply. Neither path reaches a counter, and the only trace is a `debug!` line the default `info` filter hides. On the requester side the budget-exhausted drop reads as a generic `failed to read sync ack frame` I/O error, so backpressure is indistinguishable from a transport fault.

## Fix

- [x] `crates/consensus/worker/src/metrics.rs`: `SyncShedReason { Denied, BudgetExhausted }` with a `const fn label`, and `WorkerMetrics::record_sync_stream_shed` recording `tn_worker.sync_streams_shed_total{worker, reason}` (same labeled-counter pattern as `ForwardDropReason` / `record_forward_dropped`).
- [x] `crates/consensus/worker/src/network/mod.rs`: `WorkerNetwork` gains a `metrics: WorkerMetrics` field built from the worker id in `new`. No signature change: the registry keys series by name and labels, so this instance and the `RequestHandler`'s record into the same per-worker series.
- [x] `admit_shed_and_record` wraps `try_admit_shed`: a reserved slot records `denied`, no slot records `budget_exhausted`. `shed_inbound_sync_stream` calls it at the decision point, before the spawn, so the count stays exact when the best-effort, `SYNC_REQUEST_READ_TIMEOUT`-bounded deny write later fails or the task never runs.
- [x] Why both arms: the shed budget has no per-peer sub-cap, while the admission path it guards has one. A nonzero `budget_exhausted` rate while `denied` stays low is the signature of shed slots pinned by peers that never read their deny reply (the squatting case from the issue). The pair, not the sum, is the signal. The series carry no peer label; attribution needs the per-peer `debug!` lines.

Threat model: the counter adds one labeled increment per refused stream, the same cost as the existing `quorum_failures_total` / `forwarded_txns_dropped_total` counters, and only on a path that already dropped the stream, so a flooding peer cannot make the responder do more work than before. No admission bound changes: `MAX_CONCURRENT_BATCH_STREAMS`, `MAX_PENDING_REQUESTS_PER_PEER`, and `MAX_CONCURRENT_SHED_TASKS` are untouched. The primary's `PrimaryMetrics` mirror is out of scope, per the issue.

## Testing

- `gateledger run -- cargo +nightly fmt -p tn-worker -- --check`: green.
- `metrics::tests::test_metrics_register_and_update` extended: two `Denied` and one `BudgetExhausted` refusal produce one series per reason under the worker label with counts 2 and 1. Confirmed by mutation (collapsing the `budget_exhausted` label into `denied` fails the test).
- New `network::tests::shed_admit_records_reason_per_arm`: fills the shed budget through `admit_shed_and_record` (every slot records `denied`), then one more call is refused and records `budget_exhausted`. Confirmed by mutation (swapping the two reasons in `admit_shed_and_record` fails the test).
- `gateledger run -- cargo +1.94 test -p tn-worker --lib` in an isolated target dir: green.
- CI runs on push; the pinned `cargo +1.94` attest runs on the second machine.


grantkee (dismissed):
LGTM!

grantkee (approved):

PR #1318, opened 2026-08-30
fix(primary): bound the shed-task fan-out for denied inbound sync streams (#1308)

### LINKED ISSUE #1308: Primary spawns a task per denied inbound sync stream
## Problem

`PrimaryNetwork::process_inbound_sync_stream` (`crates/consensus/primary/src/network/mod.rs:1431`) runs admission and then spawns unconditionally, checking the permit inside the spawned body:

```rust
// :1434
let permit = try_admit_sync(&self.epoch_stream_semaphore, &self.sync_stream_peers, peer);
// :1438  <-- unconditional spawn
self.task_spawner.spawn_task(task_name, async move {
    // :1446  <-- admission result checked inside the task
    let Some(_permit) = permit else {
        debug!(target: "primary::network", %peer, "denying inbound sync stream: at capacity");
```

This is the shape PR #1265 removed from the worker. A stream that admission already rejected still costs a full spawned task whose only job is to write `Deny(AtCapacity)` and close, so the cost of refusing work scales with the arrival rate of refused work rather than with the cap.

Denial is not a rare case here. The caps are `MAX_CONCURRENT_EPOCH_STREAMS = 5` globally (`:57`) and `MAX_PENDING_REQUESTS_PER_PEER = 2` (`:101`), so during catch-up, when many peers request epoch packs at once, denial is the common path. There is no `try_admit_shed` and no shed semaphore anywhere in the primary crate.

The fan-out is bounded, but loosely. The real limiter is the stream behavior's per-peer inbound rate limit, `MAX_INBOUND_PER_WINDOW = 256` per one-second window keyed by `PeerId` (`crates/network-libp2p/src/stream/behavior.rs:41-44`, enforced at `:329-337`). Each deny task is bounded at 5s by `tokio::time::timeout(SYNC_REQUEST_READ_TIMEOUT, ...)` (`:1450-1461`), so a non-reading peer cannot pin one indefinitely, but arrival rate times task lifetime still gives roughly 1,280 concurrent deny tasks from a single peer, and the limit is per peer so a coordinated set multiplies that. Against the worker's post-fix budget of 8 shed tasks, that is more than two orders of magnitude.

Two details make it worse. The rate-limit classification is deliberately not enforced: `StreamFailure::InboundRateLimited` is classified for scoring and then logged as "metrics-only, not enforced" (`crates/network-libp2p/src/consensus.rs:1783-1799`), so a peer flooding substreams is never penalized for it and the cost is a permanent steady state rather than a burst that ends in a ban. And the task manager's tracking channel is bounded at 1000 with `try_send` (`crates/types/src/task_manager.rs:298` and `:199-202`), so once a burst outruns the manager every untracked task emits its own error-level log line.

The primary already states the right principle on the neighboring request-response path. At `:1346-1357` it sheds in O(1) without spawning, with a comment noting that replying "at capacity" instead "would reintroduce a per-request task spawn, which is exactly what this bound removes." The sync stream path violates the rule stated a hundred lines above it.

## Solution

Mirror the worker fix from PR #1265 into `crates/consensus/primary/src/network/mod.rs`.

Add the budget constant beside the other caps near `:101`:

```rust
/// Maximum number of concurrent shed tasks for denied inbound sync streams.
pub(crate) const MAX_CONCURRENT_SHED_TASKS: usize = 8;
```

Add `fn try_admit_shed(semaphore: &Arc<Semaphore>) -> Option<OwnedSemaphorePermit>` beside `try_admit_sync` at `:253`, add a `shed_task_semaphore: Arc<Semaphore>` field to `PrimaryNetwork` (fields at `:1175-1190`, initialized near `:1214`), and rewrite `:1431-1463` to branch before the spawn:

```rust
let Some(permit) = try_admit_sync(&self.epoch_stream_semaphore, &self.sync_stream_peers, peer)
else {
    self.shed_inbound_sync_stream(peer, stream);
    return;
};
```

Move `let _permit = permit;` to the top of the spawned body and delete the in-task `let Some(_permit) = permit else { ... }` block. The new `shed_inbound_sync_stream` gates its spawn on `try_admit_shed` and keeps the existing 5s timeout wrapper around the `Deny(AtCapacity)` write and close.

One difference from the worker version: the primary's shed write uses the constant `MAX_SYNC_REQUEST_FRAME_SIZE` (`:146`) rather than `max_sync_frame_size(epoch)`, so the shed helper does not need to capture the epoch.

There is no shared code between the two paths to factor out. They have separate constants, separate `try_admit_sync` implementations, and different permit types (`PeerSlotPermit` versus `SyncStreamPermit`), so this is a deliberate parallel change to a mirrored file rather than a refactor.

Worth considering alongside this: enforcing the already-classified `InboundRateLimited` penalty (`crates/network-libp2p/src/consensus.rs:1783-1799`) would turn a sustained substream flood into a ban instead of a permanent steady-state cost, for both the primary and the worker.

MavenRain:
Closes #1308.

## Problem

`PrimaryNetwork::process_inbound_sync_stream` runs admission and then spawns unconditionally, checking the permit inside the spawned body.  A stream that admission already rejected still costs a full spawned task whose only job is to write `Deny(AtCapacity)` and close, so the cost of refusing work scales with the arrival rate of refused work rather than with the cap.  This is the shape #1265 removed from the worker.

Denial is the common path here, not a rare case: the caps are `MAX_CONCURRENT_EPOCH_STREAMS` (5) globally and `MAX_PENDING_REQUESTS_PER_PEER` (2), so during catch-up, when many peers request epoch packs at once, most arrivals are denied.  The fan-out is bounded only by the stream behavior's per-peer inbound rate limit (256 opens per one-second window, per `PeerId`), and each deny task lives up to the 5s `SYNC_REQUEST_READ_TIMEOUT`, so a single non-reading peer sustains roughly 1,280 concurrent deny tasks, and the limit is per peer so a coordinated set multiplies that.  No attacker sophistication is required: opening substreams is cheap for the requester, and every one of them costs the primary a spawned task before any admission decision applies to it.

The primary already states the right principle a hundred lines above, on the epoch-record request-response path: shed in O(1) without spawning, because replying "at capacity" "would reintroduce a per-request task spawn, which is exactly what this bound removes."

## Fix

Mirror the worker fix from #1265.  There is no shared code to factor out (separate constants, separate `try_admit_sync` implementations, different permit types), so this is a deliberate parallel change to the mirrored file.

- [x] `crates/consensus/primary/src/network/mod.rs`: add `MAX_CONCURRENT_SHED_TASKS` (8) and a `shed_task_semaphore` on `PrimaryNetwork`.  The shed write for a denied stream now runs on its own budgeted task (`shed_inbound_sync_stream`).  The budget check is a `try_acquire` ahead of the spawn, so it is cheap-first and attacker-independent.
- [x] Past the budget the stream is dropped without a spawn.  The requester sees a reset instead of a `Deny` frame and retries elsewhere, the same recovery it already has for a lost stream.
- [x] The admitted path now spawns only after `try_admit_sync` returns a permit, so no code path spawns ahead of an admission decision.  Total sync-task fan-out is bounded by `MAX_CONCURRENT_EPOCH_STREAMS` (5) admitted tasks plus `MAX_CONCURRENT_SHED_TASKS` (8) shed tasks.
- [x] The shed permit is held for the task lifetime and the deny write keeps its `SYNC_REQUEST_READ_TIMEOUT` bound, so a peer that never reads cannot pin a budget slot.  One difference from the worker version: the shed write uses the constant `MAX_SYNC_REQUEST_FRAME_SIZE` rather than `max_sync_frame_size(epoch)`, so the shed helper does not capture the epoch.

Kept for parity with the worker, and worth a follow-up in both files: the shed budget is global with no per-peer counter, so a peer that never reads can pin all 8 slots for the 5s write timeout each, and other denied peers then see only a reset.  That is the same trade the worker's merged version makes;  fixing it in one file but not its mirror would diverge the pattern this PR exists to keep aligned.

## Testing

- Nightly `fmt --check` on `tn-primary`
- `clippy -p tn-primary --no-deps --all-targets` with `-D warnings`
- `tn-primary` unit test `test_shed_admit_enforces_budget_and_frees_on_drop` (in `network_tests` next to the existing epoch-record admission tests;  confirmed by mutation: with `try_admit_shed` forced to `None` the test fails, restored green)
- CI (nightly fmt, workspace clippy, nextest) fires on push;  the pinned `+1.94` attest runs on the second machine.


grantkee (approved):
LGTM!

PR #1319, opened 2026-08-30
fix(prevrandao): pin the seed gate and document residual bias

### LINKED ISSUE #1310: PREVRANDAO seed fork is missing the rollout scaffolding its sibling forks carry
## Problem

#1249 (fix #1247) moved the `mix_hash`/`PREVRANDAO` derivation behind a fork gate: from `PREVRANDAO_FORK_EPOCH` onward the value is a domain-separated keccak over the epoch seed chain value, consensus block number, and batch index; before it, the legacy XOR of the consensus output digest and batch digest. The derivation itself verified consensus-safe: pre-fork replay is byte-identical, the gate reads the digest-pinned leader epoch, and it fails closed behind `seed_signature_active`. What did not land is the scaffolding the repo wraps around every epoch-gated fork, established twice over by the seed-signature and multi-workers forks. None of the gaps changes consensus today, since the gate is dormant on adiri and doubly guarded, but they ambush exactly the people the scaffolding exists for: operators diffing the fleet, whoever lands the arming PR, indexers reproducing `mix_hash`, and anyone running the engine suite against a testnet build.

Operationally, the startup fork-schedule log omits the new constant. That log exists so operators can diff compiled schedules across the fleet before an arming PR arrives, a compile-time constant that differs between binaries has no other in-protocol detection, and both siblings logged their dormant placeholder in the PR that introduced them. The non-adiri arm's message is now also incomplete: it enumerates the genesis-active forks, and seed-chain PREVRANDAO is one of them. Separately, the `TN_PREVRANDAO_FORK_EPOCH` test override exists in code but nothing sets it. The Makefile declares and exports only the two sibling variables, and the e2e harness pins only those two on spawned nodes, so the seeded derivation is unreachable in every default e2e lane. Worse, on non-adiri builds arming the seed fork alone silently arms PREVRANDAO too, because `prevrandao_seed_active` conjoins the two gates and the PREVRANDAO fork point is genesis-active off adiri. The harness documents its pins as arming independently; this fork breaks that property.

On coverage, no test pins the arm switch to the constant. Every `prev_randao` test builds its leader at epoch 0, so each build compiles and runs exactly one arm, selected by the feature set rather than by `PREVRANDAO_FORK_EPOCH`, and the fail-closed conjunction itself has zero direct coverage. One supported configuration is actually red: the engine's basefee-penalty test hardcodes the post-fork oracle and fails under `--features adiri`, where the fork is dormant and the fixture leader sits at epoch 0. CI cannot see the failure because the adiri test lane omits tn-engine, violating the lane's own stated rule that every crate forwarding `adiri` with gated tests belongs on it.

On documentation, `docs/src/evm-compatibility.md` and the tn-reth README still describe only the legacy XOR. Both tables were accurate days ago and name indexers and block explorers as their audience, exactly the readers who reproduce header fields from consensus data and would branch on the wrong derivation. And nothing at the EVM-visible surface records the residual the seed chain explicitly accepts: the committing leader can compute every PREVRANDAO its commit will produce before broadcasting, and can withhold the proposal if it dislikes them. This fork promotes that accepted last-actor bias into an opcode contracts can read, and the natural places to look say nothing about it.

## Solution

Bring the fork up to the sibling scaffolding. Log `PREVRANDAO_FORK_EPOCH` in the adiri arm of the startup fork schedule and name prevrandao in the non-adiri genesis-active message; carry the siblings' fleet-diff rationale on the constant's docs. Wire `TN_PREVRANDAO_FORK_EPOCH` as a third sibling: declare it in the Makefile defaulting to dormant, export it on the fork-pinned test lanes only (bare `--workspace` runs must stay variable-free so the override's inertness test keeps a clean process), and add it to the harness pin loop with the conjunction documented, so a seed-armed lane no longer arms PREVRANDAO as a side effect and a lane that wants the seeded derivation arms both knobs deliberately.

Close the coverage gap in both directions. The engine oracle should dispatch on the same predicate production dispatches on, recomposing by hand whichever derivation the build's fork schedule selects, and tn-engine should join the adiri CI lane so its pre-fork legacy arm actually runs somewhere. Add boundary keepers that derive both grid points from the constant, so an arming PR retargets them with no edit: under adiri, the legacy XOR must replay byte-identically at the epoch below the fork and the seeded derivation must fire at the fork epoch, with anti-vacuity tripwires against ambient overrides; without adiri, every epoch must take the seeded arm. A companion test should pin the conjunction itself: seed-dormant epochs stay on the legacy arm regardless of the fork point.

Update both public tables to the fork-gated derivation, with indexer guidance: branch on the block's epoch, recoverable from the upper 32 bits of `nonce`, and note that the seed chain value is the same one published as `extra_data` on the epoch's closing block. Record the accepted last-actor bias where readers will look (the derivation's docs, the fork constant, the payload attribute, the EVM block module), stating what the fork closes, grinding by re-cutting the payload, and what it does not: the leader keeps one propose-or-withhold choice per commit, so contracts needing unbiasable randomness must not use PREVRANDAO alone.

MavenRain:
Closes #1310. (Cantina #9)

## Problem

The PREVRANDAO rollout scaffolding is now largely present on `main`. The remaining gap is a direct regression test for the fail-closed seed-signature conjunction and documentation of the accepted leader withholding bias at the EVM-facing surfaces.

## Changes

- **Fork-gate regression test** (`crates/types/src/forks.rs`): run a child process with the seed-signature fork dormant and the PREVRANDAO fork point active from genesis. Assert that the public gate stays closed at every sampled seed-dormant epoch, with explicit override and anti-vacuity checks.
- **Derivation and indexer guidance** (`docs/src/evm-compatibility.md`, `crates/tn-reth/README.md`): document the epoch-gated legacy XOR and seeded keccak derivations, the fail-closed conjunction, integer encoding, and the closing block's seed checkpoint.
- **Accepted bias documentation** (`crates/types/src/forks.rs`, `crates/tn-reth/src/payload.rs`, `crates/tn-reth/src/evm/block.rs`): explain the committing leader's remaining propose-or-withhold choice.

## Threat model

The seeded derivation prevents payload grinding because its inputs are fixed by the committed order. The committing leader still sees the values before broadcasting and retains one propose-or-withhold choice per commit. Contracts requiring unbiasable randomness must not rely on PREVRANDAO alone. The regression test verifies that a misordered fork schedule cannot activate seeded PREVRANDAO while the seed-signature fork is dormant. This PR changes no production consensus behavior or activation epoch.

## Testing

- After merging `main` at `58c0a62d48a1c977870174ceb857491ab3102c2b`: `gateledger run -- cargo +nightly-2026-03-20 fmt --all -- --check` passed.
- `git diff --cached --check` passed. Automated merge-integrity checks confirmed disjoint file sets (five PR files and 29 incoming main files) and byte-identical preservation of both parent diffs, with no conflicts or manual edits.
- `diffclass --base HEAD --staged`, `apidelta --base HEAD --verb check`, and `buildplan --base HEAD --all-targets --emit json check` selected full-workspace validation for the incoming main changes. No compilation or tests were rerun for this merge update.
- The original PR validation recorded default/adiri PREVRANDAO tests and a mutation check that removed the seed conjunct. Those results predate this merge and are not validation of the new merge commit.
- The updated commit requires the second-machine test/attestation run. GitHub skips compile/test lanes for maintainer-authored PRs and checks the on-chain attestation; merge-queue lanes validate the eventual queue commit.


grantkee (changes requested):
Most of this was addressed in #1311 - the changes in docs/src/evm-compatibility.md and tn-reth README are great additions

Should be good to go after resolving merge conflicts

grantkee (approved):

PR #1322, opened 2026-08-31
refactor(worker): extract shed_sync_stream generic over the stream (#1314)

### LINKED ISSUE #1314: worker: extract shed_sync_stream generic over the stream so the shed path is unit-testable (follow-up to #1265)
## Summary

Follow-up to #1265 (bounded shed-task budget for denied inbound sync streams, #1254).  In review, @grantkee suggested a refactor: extract the shed path into a free function that is generic over the stream, in the style of `send_sync_batches_over_stream`.  That makes the full shed behavior testable with an in-memory writer and a paused clock.

Review comment: https://github.com/Telcoin-Association/telcoin-network/pull/1265#discussion_r3884604612

## Problem

The current unit test, `shed_admit_enforces_budget_and_frees_on_drop`, exercises only `try_admit_shed` against a bare semaphore.  From the review:

> This test is mostly pinning "non-blocking, fallible, owned-permit", but doesn't cover the deny -> shed routing, the binding to `self.shed_task_semaphore`, or the permit's hold-for-task-lifetime

Concretely, no test covers:

- The routing in `process_inbound_sync_stream`: a denied stream takes the shed path instead of the serve path.
- That the shed spawn is gated by `self.shed_task_semaphore` specifically.
- That a shed task holds its permit for the task's lifetime, so `MAX_CONCURRENT_SHED_TASKS` bounds live tasks, not spawn attempts.
- That the `SYNC_REQUEST_READ_TIMEOUT` bound on the deny write returns every slot, even when the peer never reads.

`shed_inbound_sync_stream` takes a libp2p `Stream` and reads `self.network_handle`, so a unit test cannot drive it directly today.

## Suggested fix (from the review)

Extract the shed body into a free function generic over the stream:

```rust
fn shed_sync_stream<S>(
    shed_sem: &Arc<Semaphore>,
    spawner: &TaskSpawner,
    epoch: Epoch,
    peer: BlsPublicKey,
    stream: S,
) where
    S: futures::AsyncWrite + Unpin + Send + 'static,
```

`shed_inbound_sync_stream` becomes a thin wrapper that passes `self.shed_task_semaphore`, the spawner, and the epoch.  Then test with a writer whose `poll_write` always returns `Poll::Pending`, under `#[tokio::test(start_paused = true)]`.  The review sketch (loop keyword replaced with the house combinator style):

```rust
(0..MAX_CONCURRENT_SHED_TASKS)
    .for_each(|_| shed_sync_stream(&sem, &spawner, 0, peer(), PendingWriter));
assert_eq!(sem.available_permits(), 0, "each shed task holds its slot while it runs");
// over budget: no spawn, no negative accounting
shed_sync_stream(&sem, &spawner, 0, peer(), PendingWriter);
assert_eq!(sem.available_permits(), 0);
// the bounded write trips SYNC_REQUEST_READ_TIMEOUT and every slot returns
tokio::time::advance(SYNC_REQUEST_READ_TIMEOUT + Duration::from_secs(1)).await;
assert_eq!(sem.available_permits(), MAX_CONCURRENT_SHED_TASKS);
```

This pins the budget as a bound on concurrent live tasks, the drop-without-spawn behavior past the budget, and the timeout-driven release of every slot.

## Scope

- `crates/consensus/worker/src/network/mod.rs` only.
- No behavior change: the refactor moves the existing shed body behind a generic signature.
- The existing `shed_admit_enforces_budget_and_frees_on_drop` test can stay as a cheap guard on the admit helper, or fold into the new test.

MavenRain:
Closes #1314.  Follow-up to #1265 (#1254): implements the refactor suggested in review (https://github.com/Telcoin-Association/telcoin-network/pull/1265#discussion_r3884604612).

## Summary

The shed path for denied inbound sync streams lived inside `WorkerNetwork::shed_inbound_sync_stream`, which takes a libp2p `Stream` and reads `self.network_handle`, so no unit test could drive it.  The existing test, `shed_admit_enforces_budget_and_frees_on_drop`, pins only the admit helper against a bare semaphore.  It does not cover the deny routing, the binding to `self.shed_task_semaphore`, the permit's hold-for-task-lifetime, or the timeout-driven release of stuck slots.

## Changes

- [x] `crates/consensus/worker/src/network/mod.rs`: extract the shed body into `shed_sync_stream<S>(shed_semaphore, spawner, epoch, peer, stream)` with `S: AsyncWrite + Unpin + Send + 'static`, in the style of `send_sync_batches_over_stream`.  No behavior change: the body moves as-is behind the generic signature.
- [x] `shed_inbound_sync_stream` becomes a thin wrapper that binds `self.shed_task_semaphore`, the handle's task spawner, and the current epoch.
- [x] New test `shed_sync_stream_bounds_live_tasks_and_times_out_stuck_writes` under `#[tokio::test(start_paused = true)]`: drives the full shed path with a `PendingWriter` whose polls never complete, modeling a peer that never reads its deny reply.
- [x] New test `shed_sync_stream_writes_deny_at_capacity_and_frees_slot`: a `CapturingWriter` records the shed task's bytes;  the test decodes them back through `read_frame` and asserts exactly the `Deny(AtCapacity)` frame, and that the slot returns  with no clock advance.
- [x] `shed_admit_enforces_budget_and_frees_on_drop` stays as the cheap guard on the admit helper.

Scope note: the wrapper now reads the epoch and clones the task spawner before admit, so the budget-exhausted drop path pays those two handle reads.  Accepted to keep the reviewed signature;  the drop path remains strictly cheaper than the pre-#1265 behavior, which spawned a task per denied stream.

## Threat model

Unchanged from #1265, now pinned by tests instead of asserted in prose.  `MAX_CONCURRENT_SHED_TASKS` bounds live shed tasks, not spawn attempts, so a burst of over-cap substreams cannot fan out to unbounded deny tasks.  The `SYNC_REQUEST_READ_TIMEOUT` bound on the best-effort deny write means a peer that opens streams and never reads (receive backpressure;  no attacker required for the slow-peer case) returns every budget slot at the timeout instead of pinning the shed budget.  A served deny is exactly one `Deny(AtCapacity)` frame, so the requester fails fast and retries elsewhere instead of seeing a bare reset.

## Testing

- `cargo +1.94 test -p tn-worker --lib -- network::tests` in an isolated target dir: all four network unit tests green.
- The new tests are mutation-confirmed three times from in-diff mutants: (a) dropping the shed permit at task start (the budget would then bound spawn attempts, not live tasks) fails the held-slot assertion;  (b) unbinding the deny-write timeout fails the slot-return assertion;  (c) writing `Err(Malformed)` instead of `Deny(AtCapacity)` fails the deny-frame decode assertion.  All mutants restored;  suite green after restore.
- `cargo +1.94 check -p tn-worker --all-targets` and `cargo +1.94 clippy -p tn-worker --no-deps --all-targets -- -D warnings` green;  nightly `cargo fmt -- --check` green.
- CI runs the nightly clippy and nextest workspace lanes on push;  the pinned `+1.94` attest runs on the second machine.


grantkee (approved):
LGTM!

PR #1324, opened 2026-09-01
fix(node): stop the restore-time RecentBlocks scan at the real-header floor

### LINKED ISSUE #1321: After a snapshot restore, the RecentBlocks backward scan can walk the dummy-header region to genesis, delaying startup by O(chain-height) database reads
## Summary

At startup, `try_restore_state` primes the `RecentBlocks` watch by calling `last_executed_output_blocks(block_capacity)` (`crates/node/src/manager/node.rs:1181-1185`).  The scan in `crates/node/src/engine/inner.rs:296-334` walks headers backward from the tip and counts a block each time the header nonce changes (each consensus output stamps one nonce on all of its blocks).  It stops only when it has collected `block_capacity` distinct nonces or reaches block 0.

A node restored from a state snapshot does not have real headers below the pack's window.  The restore scaffold writes the 257 real window headers (anchor plus `BLOCKHASH_ANCESTORS = 256`, `crates/storage/src/exec_state_pack.rs:83`), and fills every block below the window with a dummy header: `ExecHeader { number, ..Default::default() }` appended with hash `B256::ZERO` (`crates/tn-reth/src/snapshot.rs:649-650`).  Every dummy header has nonce 0.

If the 257 real headers contain fewer than `block_capacity` distinct nonces, the scan enters the dummy region.  The first dummy header counts once (its nonce 0 differs from the last real nonce).  After that the nonce never changes again, so the `blocks < number` condition never becomes false, and the loop reads every header down to block 0.  Each iteration is a synchronous `sealed_header_by_number` read, so a mature chain pays millions of pointless header reads inside the startup path, before `try_restore_state` returns and the node proceeds to networking and consensus.

## Preconditions

- The node was restored from a state snapshot (dummy headers exist below the 257-header window).
- The 257 real headers hold fewer than `block_capacity` distinct consensus-output nonces.  The default capacity is `gc_depth = MAX_GC_DEPTH = 50` (`crates/types/src/primary/mod.rs:49`), so this holds whenever the window averages more than 257/50 ≈ 5.1 execution blocks per consensus output.  Under real batch load this is a plausible steady state, and a larger configured `gc_depth` lowers the bar further.

No attacker is required;  the cost is self-inflicted on every restart of an affected node, and grows linearly with chain height forever.

## Secondary effect

When the scan crosses the boundary, it pushes one dummy header (nonce 0, stored hash `B256::ZERO`, default `parent_beacon_block_root`) into the result, so `try_restore_state` seeds `RecentBlocks` with a synthetic header and a `ConsensusNumHash` derived from default values (`crates/node/src/manager/node.rs:1185-1199`).

## Suggested fix

Bound the scan so it never leaves real history:

- Stop the walk when it reaches the snapshot scaffold region, for example break when a header comes back with the dummy signature (zero hash / zero nonce below the window start), or persist the restore floor and clamp `block_num` to it.
- Alternatively, cap total iterations at a small multiple of `block_capacity` times the window size;  the scan never legitimately needs more than the 257-header window plus the live tip region.
- Do not push the boundary dummy header into the result.

## Credit

Reported by Haxatron in the Cantina collaborative review (comment on `try_restore_state`, 2026-08-31).

Refs: origin/main `bbf08cd8`.

MavenRain:
Closes #1321.  (Cantina #30)

## Problem

At startup, `try_restore_state` primes the `RecentBlocks` watch through `last_executed_output_blocks(block_capacity)`.  The scan walks headers backward from the tip and counts one block per consensus-output nonce change, stopping only after `block_capacity` distinct nonces or at block 0.

A node bootstrapped from a state snapshot holds real headers only inside the restore window (the anchor plus the `BLOCKHASH_ANCESTORS = 256` lookback).  Every block below the window is a scaffold dummy: `ExecHeader::default()` with a zero hash and a zero nonce.  When the window holds fewer distinct nonces than `block_capacity` (the default capacity is `gc_depth = 50`, so any window averaging more than ~5.1 execution blocks per consensus output qualifies), the scan crosses into the dummy region.  The first dummy counts once (its nonce 0 differs from the last real nonce), and after that the nonce never changes again, so the walk reads every header down to genesis.  On a mature chain that is millions of synchronous `sealed_header_by_number` reads inside the startup path, before networking and consensus come up, and the cost grows linearly with chain height on every restart.  No attacker is required;  the cost is self-inflicted and permanent.

The boundary crossing also pushes one dummy header into the result, so `RecentBlocks` gets seeded with a synthetic header and a `ConsensusNumHash` derived from default values.

## Fix

Bound the scan so it never leaves real history, using state the restore already persists:

- [x] `crates/tn-reth/src/env/mod.rs`: new `RethEnv::real_header_floor()`, the lowest block number guaranteed to hold a real header.  It is `0` on a normally-synced node and `max(1, B - (BLOCKHASH_ANCESTORS - 1))` on a restored datadir, derived from the restored-state floor marker `B`.  The bound is not guessed: `import_chain_scaffold` admits a window only if it starts at or below that expression, so the derivation reuses the exact invariant the restore enforces.  The actual window may start one block lower, but only `B` is persisted, so the derived bound trims at most one legitimately readable header while never admitting a dummy read.
- [x] `crates/node/src/engine/inner.rs`: the backward walk is extracted into `collect_last_output_blocks(read_header, tip, number, scan_floor)` (combinator form, no imperative loop) and its termination changes from "block 0" to the floor.  With no restore the floor is 0 and the walk behaves exactly as before.  On a restored datadir the walk stops at the guaranteed-real floor: it never reads the dummy region and never returns a synthetic header, so the worst case is the restore window plus the live tip region instead of O(chain-height).

## Testing

- New unit tests for `collect_last_output_blocks` (`crates/node/src/engine/inner.rs`), driven by a synthetic reader that counts every header fetch:
  - Regression for #1321: a 10,000-block chain whose real window (9,745..=10,000) holds 32 distinct nonces against a request for 50.  Asserts the walk performs at most window-many reads, returns no header below the floor, and stays oldest-first.
  - Baseline without a floor: collects exactly the requested count, the newest head is the tip, and consecutive heads carry distinct nonces.
  - Genesis termination on a short, normally-synced chain (the pre-existing behavior, preserved by `scan_floor == 0`).
- `real_header_floor` coverage (`crates/tn-reth/src/snapshot.rs`): a new test asserts the absent-marker case returns 0 and a hand-written marker at 100000 yields 99745 (the non-clamped arm);  the existing scaffold test now also asserts the clamped arm (B = 5 yields floor 1) on an env built over a genuinely restored datadir.
- Each new load-bearing test was confirmed by mutation: removing the floor from the walk range, an off-by-one on the collection cap, and a floor derivation that returns `B` itself each make their covering test fail, and all pass again with the real code restored.
- Local static ladder under the pinned toolchain: `cargo +1.94 check -p tn-node -p tn-reth --all-targets`, `cargo +1.94 clippy --no-deps -p tn-node -p tn-reth --all-targets -- -D warnings`, nightly `fmt -- --check` on the touched crates.  CI owns the dynamic tier on push;  the pinned `+1.94` attest runs post-push on the second machine.

### DIFF
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -300,7 +300,7 @@ impl ExecutionNodeInner {
         let last_block_number = self.reth_env.last_block_number()?;
         debug!(target: "epoch-manager", ?last_block_number, "restoring last executed output blocks");
         collect_last_output_blocks(
-            |block_num| Ok(self.reth_env.sealed_header_by_number(block_num)?),
+            |block_num| self.reth_env.sealed_header_by_number(block_num).map_err(Into::into),
             last_block_number,
             number,
             self.reth_env.real_header_floor(),
--- a/crates/node/src/engine/inner.rs
+++ b/crates/node/src/engine/inner.rs
@@ -5,7 +5,7 @@
 use crate::error::ExecutionError;
 use eyre::{eyre, OptionExt};
 use jsonrpsee::http_client::HttpClient;
-use std::{net::SocketAddr, sync::Arc};
+use std::{net::SocketAddr, ops::ControlFlow, sync::Arc};
 use tn_batch_builder::BatchBuilder;
 use tn_batch_validator::BatchValidator;
 use tn_config::Config;
@@ -299,39 +299,12 @@ impl ExecutionNodeInner {
     ) -> eyre::Result<Vec<SealedHeader>> {
         let last_block_number = self.reth_env.last_block_number()?;
         debug!(target: "epoch-manager", ?last_block_number, "restoring last executed output blocks");
-        let mut result = Vec::with_capacity(number as usize);
-        if number > 0 {
-            let mut block_num = last_block_number;
-            let mut last_nonce;
-            if let Some(header) = self.reth_env.sealed_header_by_number(block_num)? {
-                last_nonce = header.nonce;
-                result.push(header);
-            } else {
-                return Err(eyre::Error::msg(format!("Unable to read block {block_num}")));
-            }
-            let mut blocks = 1;
-            while blocks < number {
-                if block_num == 0 {
-                    break;
-                }
-                block_num -= 1;
-                if let Some(header) = self.reth_env.sealed_header_by_number(block_num)? {
-                    // Do this check so we only track the "finalized" blocks not all the extra
-                    // batches.  Note that the nonce will be the same for all batches in the same
-                    // consensus output but will change for each output (composed of epoch and
-                    // round).
-                    if header.nonce != last_nonce {
-                        last_nonce = header.nonce;
-                        result.push(header);
-                        blocks += 1;
-                    }
-                } else {
-                    return Err(eyre::Error::msg(format!("Unable to read block {block_num}")));
-                }
-            }
-        }
-        result.reverse();
-        Ok(result)
+        collect_last_output_blocks(
+            |block_num| Ok(self.reth_env.sealed_header_by_number(block_num)?),
+            last_block_number,
+            number,
+            self.reth_env.real_header_floor(),
+        )
     }
 
     /// Return an database provider.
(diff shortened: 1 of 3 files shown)


Haxatron:
Fix verified: for snapshots produced by the standard exporter, startup reconstruction of RecentBlocks stops at the scaffold floor `max(1, B - 255)`, where B is the imported snapshot height. It therefore does not scan into the synthetic dummy-header region.

Edit: Latest main now enforces that any snapshot import reaches the scaffold floor B-255 here https://github.com/Telcoin-Association/telcoin-network/blob/66e0d14a52b042a383b666d4b4da0079ec5b4871/crates/tn-reth/src/snapshot.rs#L558-L574

grantkee (approved):

sstanfield (approved):

PR #1325, opened 2026-09-02
fix(tn-reth): floor snapshot-scaffold heights out of the canonical execution check

### LINKED ISSUE #1323: Snapshot-restored nodes attest zero-hash dummy headers as canonical execution blocks, letting a Byzantine proposer certify a poison latest_execution_block that halts full-history validators at commit time
**Severity:** High.  A single Byzantine committee member can craft a header whose `latest_execution_block` references a scaffold height with hash `B256::ZERO`;  if a voting quorum of the committee is snapshot-restored, the header certifies, and every validator that holds a *real* header at that height (full-history nodes, or nodes restored below a different height) self-halts (`NodeMode::CvvInactive` + `shutdown().notify()`) when the certified sub-dag reaches commit.  This is a targeted liveness attack that partitions the validator set by snapshot provenance.

**Attacker model:** One Byzantine committee member to author the poison header.  The commit-time halt then requires the header to certify, that is, a `quorum_threshold()` (2f+1 stake) of voters that accept it.  Snapshot-restored voters accept it;  the more of the committee that bootstrapped from a snapshot below the referenced height, the easier the certificate.  Honest operation never produces the poison value (honest proposers fill the field from their own real tip, `crates/consensus/primary/src/proposer.rs:307`), so this is Byzantine-triggered, not spontaneous.

**Credit:** surfaced by Haxatron in the current collaborative review.  This issue records the finding with the full read path and the commit-time consequence traced to repo evidence.

## Summary
Snapshot restore fabricates the pre-window header region with a placeholder hash.  In `crates/tn-reth/src/snapshot.rs:628-637`, for every block below the shipped window it writes a dummy header sealed with `B256::ZERO`:

```rust
None => {
    let dummy = ExecHeader { number, ..Default::default() };
    headers_writer.append_header(&dummy, &B256::ZERO)?;
}
```

The window is `[B-256, B]` (`BLOCKHASH_ANCESTORS = 256`, `crates/node/src/manager/export_exec.rs:37`);  heights `1..B-256` are zero-hash dummies.  The code comment at `crates/tn-reth/src/snapshot.rs:744` asserts "nothing a snapshot-bootstrapped node does ever reads them."  The consensus execution-result check falsifies that: it reads exactly these heights and treats `B256::ZERO` there as an authentic canonical hash.

An execution-result check gates whether a node will vote on a peer's header.  `wait_for_execution` (`crates/consensus/primary/src/consensus_bus.rs:717-746`) is called from the vote handler on the peer-supplied field:

```rust
// crates/consensus/primary/src/network/handler.rs:770
if self.consensus_bus.wait_for_execution(header.latest_execution_block()).await.is_err() {
    ...
    return Err(HeaderError::UnknownExecutionResult(header.latest_execution_block()).into());
}
```

`latest_execution_block` is a plain, author-set field on the signed header.  A Byzantine committee member sets it to `(dummy_height, B256::ZERO)` for any `dummy_height` in `1..B-256`.

## Why the check is asymmetric across honest nodes
`wait_for_execution` accepts on two paths, both of which a snapshot-restored node satisfies for `(dummy_height, B256::ZERO)` and a full-history node does not:

1. In-memory ring, number-agnostic (`crates/consensus/primary/src/recent_blocks.rs:82-89`):
   ```rust
   pub fn contains_execution_hash(&self, hash: BlockHash) -> bool {
       for block in &self.executed_blocks {
           if block.hash() == hash { return true; }
       }
       false
   }
   ```
   The comparison ignores the block number.  On restore, `try_restore_state` (`crates/node/src/manager/node.rs:1141-1170`) seeds the ring from `last_executed_output_blocks`, whose backward scan descends into the dummy region and pushes a dummy whose hash is `B256::ZERO` (this backward scan is the subject of #1321).  Any `contains_execution_hash(B256::ZERO)` then returns true.

2. Canonical-DB fallback (`crates/consensus/primary/src/consensus_bus.rs:338-344`, added by #1036):
   ```rust
   fn confirms(&self, block: &BlockNumHash) -> bool {
       let reader = self.0.lock().clone();
       reader
           .and_then(|reader| reader.canonical_execution_hash(block.number))
           .is_some_and(|hash| hash == block.hash)
   }
   ```
   This path is durable even after the ring evicts the dummy: a restored node's DB genuinely stores `B256::ZERO` at every dummy height, so `canonical_execution_hash(dummy_height) == B256::ZERO == block.hash` returns true.  A full-history node has the real hash there, so `confirms` returns false and it rejects.

Net at vote time: restored validators vote for the poison header;  full-history validators return `UnknownExecutionResult` and do not.

## Why certification is network-wide, not confined to restored nodes
The execution-result check gates **voting only**.  It is not re-applied when a full certificate is accepted from a peer.  Certificate acceptance (`crates/consensus/primary/src/state_sync/cert_validator.rs:100-128`) delegates to `Certificate::validate_and_verify` (`crates/types/src/primary/certificate.rs:225-249`), which checks epoch, quorum weight, and BLS signatures, then `Header::validate` (`crates/types/src/primary/header.rs:123-168`) checks epoch, parent/batch counts, digest, authority membership, and worker ids — there is no `latest_execution_block` / `wait_for_execution` check on this path.  A quorum-signed certificate is written into the DAG (`crates/consensus/primary/src/state_sync/cert_manager.rs:189-217`) on every node regardless of the execution field.

So once a restored-node quorum certifies the poison header, the certificate propagates to and is accepted by the full-history nodes too.

## The halt
The execution check re-surfaces at commit time, in the Bullshark output loop, for the committed sub-dag leader (`crates/consensus/primary/src/consensus/state.rs:501-514`):

```rust
let base_execution_block = committed_sub_dag.leader().latest_execution_block();
if self.consensus_bus.app().wait_for_execution(base_execution_block).await.is_err() {
    tracing::error!(target: "telcoin::consensus_state", "Got a bogus sub dag from bullshark, we are out of sync and probably can not recover!");
    self.consensus_bus.app().node_mode().send_modify(|v| *v = NodeMode::CvvInactive);
    self.consensus_config.shutdown().notify();
    ...
    return Ok(());
}
```

On a full-history node this check fails for `(dummy_height, B256::ZERO)`, so the node flips itself to `CvvInactive` and fires `shutdown().notify()` — a self-inflicted, likely unrecoverable liveness halt (per the comment).  It does not fork;  it stops rather than execute a block it cannot validate.  Restored nodes pass the same check and continue.  The validator set partitions by snapshot provenance.

## Impact
- A single Byzantine committee member, given a snapshot-restored voting quorum, halts every honest full-history validator (and any validator restored below the referenced height).  Because snapshot restore is a normal bootstrap/recovery path, a mixed committee is realistic.
- The defect chain is three parts, each independently worth closing: (1) dummies are persisted with a real-looking `B256::ZERO` canonical hash (`snapshot.rs:635`) and are readable via `canonical_execution_hash`;  (2) both accept paths bind the height to that fabricated hash rather than refusing the scaffold region (`recent_blocks.rs:82`, `consensus_bus.rs:338`);  (3) the execution-result check is vote-time only, so a poison header certifies and reaches commit network-wide (`cert_validator.rs`/`certificate.rs`/`header.rs` above).
- Not fixed by #1321.  Its scan-floor fix stops the dummy being seeded into the ring, but `B256::ZERO` stays in the DB, so the `confirms()` path still accepts `(dummy_height, B256::ZERO)` on a restored node.

## Scope note (worth confirming)
The commit-time check at `state.rs:502` validates the sub-dag **leader's** `latest_execution_block` only.  A poison value carried by a non-leader header inside a committed sub-dag is not gated by this specific call;  how non-leader headers' execution results are consumed downstream of the committed-sub-dag channel would need a separate trace to characterize.  The leader-header case above already traces cleanly to a halt, so this note only bounds how much worse the blast radius could be.

## Suggested directions
1. Refuse the scaffold region in the canonicity check.  A node must not attest to any execution block at or below its restored-state floor marker `B` (the marker already exists and is used by #1136/#1321): have `canonical_execution_hash` return `None` for heights below the floor, and have the ring seeding skip that region, so `confirms`/`contains_execution_hash` can never accept a dummy.  This closes both accept paths at the source.
2. Bind the hash to the height in `contains_execution_hash`: compare `(number, hash)` rather than hash alone, and reject a default/zero hash outright.  Defends the ring path even if a dummy is somehow present.
3. Optionally re-validate `latest_execution_block` at certificate-acceptance (or reject a header whose referenced height is below the local restored floor before it can be voted on), so a poison header cannot be certified network-wide even if one node's data layer is wrong.  Belt-and-suspenders over (1).

MavenRain:
## What (Cantina #28)

Fixes #1323.  A snapshot restore scaffolds every pre-window execution header
with a `B256::ZERO` placeholder hash so the reth static files stay contiguous
below the restore window.  The vote-time canonical-execution check read those
placeholders directly from the DB, so a snapshot-restored validator would
attest a `(scaffold_height, B256::ZERO)` pair as canonical execution and let a
peer clear the execution-result gate against a block that was never executed.

## Fix

`RethEnv` now carries the restore window start `B` as `restored_state_floor`.
`canonical_execution_hash` refuses any lookup below the floor, so a scaffold
height resolves to `None` instead of the zero placeholder and can never be
attested as canonical.  On a fresh or normally-synced node `restored_state_floor`
is `None`, so the guard is a strict no-op.

## Scope note

An earlier revision of this change also rejected the zero hash inside the
`RecentBlocks` ring as defence-in-depth.  That reject was **dropped** for two
reasons:

1. **It defends an unreachable path.**  The scaffold dummies are written only to
   the reth static files, never seeded into the `RecentBlocks` ring.  The ring is
   primed from `last_executed_output_blocks`, which carry real sealed hashes, so
   a scaffold height is never a ring member.  The canonical-DB floor guard is the
   only reachable accept path for a scaffold height, and it is fully closed here.
2. **It broke a legitimate sentinel.**  `{0, B256::ZERO}` is the default
   "genesis / no execution yet" execution pointer that healthy nodes query
   through `wait_for_execution` before their first execution.  A blanket zero-hash
   reject in the ring makes that query miss and hangs the gate;  two `tn-node`
   integration tests reproduced the hang.

The canonical-DB floor guard is the load-bearing fix and closes #1323 on its own.

## Testing

- `tn-reth` unit test asserting `canonical_execution_hash` returns `None` for a
  height below the restore floor and the real hash at/above it.
- Manual mutation confirm: neutering the floor guard makes the unit test fail
  (mutant killed).
- The two `tn-node` integration tests that caught the earlier ring-guard
  regression (`test_entry_reads_static_fee_at_boundary`,
  `test_sync_then_catchup_recovers_two_worker_accumulator`) pass single-threaded.
- `fmt --check`, `check --all-targets`, `clippy --all-targets --no-deps -D warnings`
  on `tn-reth`.

### DIFF
--- a/crates/consensus/primary/src/recent_blocks.rs
+++ b/crates/consensus/primary/src/recent_blocks.rs
@@ -79,13 +79,13 @@ impl RecentBlocks {
     }
 
     /// Is hash a recent block we have executed?
-    ///
-    /// A zero hash never matches: no real executed block seals to `B256::ZERO`, and a snapshot
-    /// restore fabricates its pre-window scaffold headers with exactly that hash. If such a dummy
-    /// is ever seeded into the ring, a peer could clear the execution-result check by referencing
-    /// it, so the zero hash is rejected outright before the ring is scanned.
     pub fn contains_execution_hash(&self, hash: BlockHash) -> bool {
-        hash != BlockHash::ZERO && self.executed_blocks.iter().any(|block| block.hash() == hash)
+        for block in &self.executed_blocks {
+            if block.hash() == hash {
+                return true;
+            }
+        }
+        false
     }
 
     /// Is hash (consensus output) in a recent block we have executed?
@@ -144,29 +144,5 @@ mod tests {
         assert_eq!(recent.latest_consensus_block_num_hash(), consensus_num_hash);
         assert_eq!(recent.latest_execution_block_num_hash(), executed.num_hash());
         assert!(recent.contains_execution_hash(executed.hash()));
-    }
-
-    #[test]
-    fn contains_execution_hash_rejects_the_zero_hash() {
-        use tn_types::BlockHash;
-
-        let mut recent = RecentBlocks::new(4);
-        let cnh = ConsensusNumHash::new(12, ConsensusHeaderDigest::from([3u8; 32]));
-
-        // a genuinely executed block resolves by its true (non-zero) sealed hash
-        let real = SealedHeader::seal_slow(ExecHeader { number: 9, ..Default::default() });
-        recent.push_latest(3, cnh, Some(real.clone()));
-        assert!(recent.contains_execution_hash(real.hash()));
-
-        // seed a zero-hash scaffold dummy the way a snapshot restore would, then confirm it can
-        // never clear the execution-result check: a Byzantine header referencing a scaffold
-        // height carries exactly this `B256::ZERO`
-        let dummy =
-            SealedHeader::new(ExecHeader { number: 3, ..Default::default() }, BlockHash::ZERO);
-        recent.push_latest(4, cnh, Some(dummy));
-        assert!(
-            !recent.contains_execution_hash(BlockHash::ZERO),
-            "a zero-hash dummy in the ring must not be reported as an executed block"
-        );
     }
 }
--- a/crates/consensus/primary/src/recent_blocks.rs
+++ b/crates/consensus/primary/src/recent_blocks.rs
@@ -79,13 +79,13 @@ impl RecentBlocks {
     }
 
     /// Is hash a recent block we have executed?
+    ///
+    /// A zero hash never matches: no real executed block seals to `B256::ZERO`, and a snapshot
+    /// restore fabricates its pre-window scaffold headers with exactly that hash. If such a dummy
+    /// is ever seeded into the ring, a peer could clear the execution-result check by referencing
+    /// it, so the zero hash is rejected outright before the ring is scanned.
     pub fn contains_execution_hash(&self, hash: BlockHash) -> bool {
-        for block in &self.executed_blocks {
-            if block.hash() == hash {
-                return true;
-            }
-        }
-        false
+        hash != BlockHash::ZERO && self.executed_blocks.iter().any(|block| block.hash() == hash)
     }
 
     /// Is hash (consensus output) in a recent block we have executed?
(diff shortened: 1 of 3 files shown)


Haxatron:
Fix verified: the canonical execution hash returns None for blocks below the snapshot restored-state floor B, which causes [`CanonicalReaderSlot::confirms`](https://github.com/Telcoin-Association/telcoin-network/blob/66e0d14a52b042a383b666d4b4da0079ec5b4871/crates/consensus/primary/src/consensus_bus.rs#L327-L345) to return false and subsequently the canonical execution check to reject attestations to dummy blocks.

grantkee (approved):
LGTM!

PR #1328, opened 2026-09-04
refactor: remove validator regions and the region-shuffle fork gate

### LINKED ISSUE #1327: Region-aware shuffle (#1294): a misreported region buys a guaranteed committee seat, and no attestation, monitoring, or slash path exists
Follow-up to Haxatron's comment on #1294 (2026-09-02):

> Note: The downside with a region aware shuffle design to consider is that the validators can submit the false region with lowest number of validators to increase their chances of getting into a committee, even if the validator does not live in the region
>
> Perhaps this should be monitored and then slashed via the smart contract

This issue records the concern against the code as merged, corrects one detail of the premise, quantifies the advantage, and lists the mitigation options.  Nothing here blocks #1294: the fork gate ships dormant on adiri (`REGION_SHUFFLE_FORK_EPOCH = u32::MAX`), and the region-aware sequence is output-identical to the flat shuffle while every region is 0.  The decision does need to land before the epoch-setting PR arms the gate, because any later change to the draw sequence is another fork.

## Premise, corrected

Validators cannot write their own region.  `setValidatorRegion(address, uint8)` is `onlyOwner` (tn-contracts `ConsensusRegistry.sol:350`), and `invariants.md` states "only governance (contract owner) can set or update a validator's region".  So the attack is not a direct on-chain self-report.  It is a false location claim made to governance during onboarding, which governance then writes on-chain.

The substance of the concern holds:

- The region is an off-chain claim.  Nothing on-chain or in the protocol can verify where a validator runs.  Governance relays whatever evidence it accepts.
- The shuffle turns that claim into a hard scheduling guarantee, not a soft preference (see below).
- No mechanism exists to detect, correct, or penalize a false claim after the fact beyond governance calling `setValidatorRegion` again.

## How large the advantage is

`region_aware_order` (`crates/tn-reth/src/evm/block.rs`, #1294) partitions the pool by nonzero region, Fisher-Yates each group, shuffles the region visit order, then round-robins one seat per region per round before the region-0 pool fills the rest.

Round one therefore seats exactly one validator from every assigned region.  With `R` assigned regions and committee size `C`:

- If `R <= C`, every assigned region gets a seat in round one with probability 1.
- A validator alone in its region is seated with probability 1.
- A validator in a region of `k` members is seated in round one with probability `1/k`.
- Under the flat shuffle every validator is seated with probability `C/N` for pool size `N`.

Worked example, pool of 12, committee of 4:

| Region layout | Singleton's seat probability | Others' seat probability | Flat shuffle |
|---|---|---|---|
| four regions of three | n/a | 1/3 | 1/3 |
| one singleton, eleven in one region | 1 | 3/11 | 1/3 |

The second row is the incentive: claiming an empty or sparse region converts a one-in-three chance into certainty, and pushes every honest validator below what the flat shuffle gave them.  The effect grows with the pool: for a pool of 40 and a committee of 8, a singleton still sits at 1 while the flat baseline is 1/5.

Consequences of a guaranteed seat go beyond rewards.  A validator that is always seated is always a proposer and always a voter, so any Byzantine budget analysis that assumes random committee membership is weakened for that validator.

## Why no existing mechanism covers it

- Detection: the protocol cannot observe geography.  The offense is only visible to governance through off-chain evidence (hosting provider, latency, network telemetry).
- Correction: `setValidatorRegion(addr, 0)` removes the advantage immediately.  That path exists today and needs no code.
- Penalty: `applySlashes` is `onlySystemCall` (`ConsensusRegistry.sol:190`) and the contract's own comment reads "when programmatic slashing ships".  There is no governance-callable slash, and the protocol's slashing pipeline has no offense it could raise for this, because it cannot detect it.
- Visibility: the client logs the post-shuffle order at debug level and nothing else.  Nobody is told that a region has one member or that one validator has been seated in every epoch since a region write.

## Proposed follow-ups

Ordered by cost.  Items 1 and 2 are telcoin-network; item 3 is tn-contracts; item 4 is a design choice that must precede arming the fork.

1. **Document the threat model where the guarantee is made.** Add the seat-probability analysis above to the `region_aware_order` doc comment and to the `REGION_SHUFFLE_FORK_EPOCH` constant's rollout notes, next to the existing "governance must not assign a nonzero region before the fork epoch" constraint.  State plainly that a region assignment is a governance attestation with a scheduling reward attached, and that sparse-region assignments deserve the most scrutiny.  Mirror the same paragraph in tn-contracts `design.md` under "Region-Aware Committee Shuffle" and `invariants.md` under "region".

2. **Make the advantage observable at epoch close.** When the region-aware arm runs, emit an info-level log (and a metric if the engine has a registry in scope) with the per-region pool counts and the per-region seat counts of the drawn committee.  Flag any region whose pool count is 1.  This is a few lines next to the existing `validators post-shuffle (region-aware)` debug line, costs no consensus change, and gives operators the signal Haxatron asked to monitor.

3. **Give governance an enforcement path (tn-contracts).** Today the only remedy is resetting the region.  If the team wants a penalty, the contract needs either a governance-callable slash entry point with an explicit reason, or the future programmatic slashing pipeline needs to accept governance-attested offenses.  Either way the `ValidatorRegionUpdated` event already gives auditors the write history.  Suggest filing a companion issue in tn-contracts that references this one.

4. **Optionally bound the guarantee before the fork is armed.** If a hard seat guarantee is unwanted, the cheapest design change is to treat regions below a minimum population `m` (for example 2) as unassigned, so a singleton earns nothing until a second validator shares the region.  This still rewards sparse regions, but removes the certainty and the empty-region trick.  It changes the documented algorithm and the draw sequence, so it is a consensus change: it must land while `REGION_SHUFFLE_FORK_EPOCH` is still `u32::MAX` on adiri, or it needs its own fork.  If the team accepts the guarantee as intended (regional diversity is the stated goal), close this item and keep 1 through 3.

## References

- PR #1294, Haxatron's comment: https://github.com/Telcoin-Association/telcoin-network/pull/1294#issuecomment-5516923392
- Issue #1279 (the shuffle implementation request)
- tn-contracts `src/consensus/ConsensusRegistry.sol` (`setValidatorRegion` `onlyOwner` at line 350, `applySlashes` `onlySystemCall` at line 190)
- tn-contracts `src/consensus/design.md` "Region-Aware Committee Shuffle", `invariants.md` "region"
- tn-contracts #149 (Cantina finding 8, staking clobbered a preconfigured region) for prior region-field history

MavenRain:
Replaces this PR's previous scope (#1327 items 1 and 2).  Closes #1327: the seat guarantee it documents no longer exists.

## Problem

Standup (2026-09-04): the network does not support validator regions yet.  A future spec will decide how regions are consumed.  Until then the client carries a region-aware committee draw behind a dormant fork gate (#1294), the per-region seat tally this branch had added (#1327), and a `setValidatorRegion` binding with no caller.  Dead machinery on a consensus-critical path costs review and audit time and gives nothing back.

## Changes

- [x] `crates/tn-reth/src/evm/block.rs`: `assemble_new_committee` is the flat Fisher-Yates draw again.  `region_aware_order`, `fisher_yates`, the `region_aware` parameter, the seat-tally wiring and the region tests are gone.  The file is byte-identical to main before #1294 (286888cb).
- [x] `crates/types/src/forks.rs`: `REGION_SHUFFLE_FORK_EPOCH`, `region_shuffle_active`, `region_shuffle_build_fork_active`, the `TN_REGION_SHUFFLE_FORK_EPOCH` test override and their two tests are gone.  Every other fork constant and gate is untouched.  Byte-identical to pre-#1294.
- [x] `crates/telcoin-network-cli/src/node.rs`: the startup fork-schedule log drops the region entry.  Byte-identical to pre-#1294.
- [x] `crates/tn-reth/src/metrics.rs` and `crates/tn-reth/README.md`: `RegionSeatReport`, `record_region_seats`, the three `tn_reth_epoch_close_region_*` gauges and their README section are gone.  Both files are byte-identical to main.
- [x] `crates/tn-reth/src/system_calls.rs`: the unused `setValidatorRegion` binding is gone.  `ValidatorInfo.region` stays, documented as a storage field the client never reads.

## Threat model

- No fork.  On adiri the gate was dormant (`u32::MAX`).  On every other build the region arm was active from genesis, but no validator carries a nonzero on-chain region (`setValidatorRegion` is the only writer, and the #1294 rollout rule forbade a write before the fork epoch), and with every region 0 that arm drew identically to the flat shuffle (the equivalence test in #1294 pinned it).  The committee draw sequence is unchanged on every network, and the restored code is the code that produced the chain's history.
- `ValidatorInfo.region` must stay in the `sol!` struct: the pinned ConsensusRegistry still returns it, and an ABI struct with a missing field fails to decode.  From the client's view the field is now write-only.
- Governance can still call `setValidatorRegion` on the contract.  The client ignores the value, so the call is a no-op for consensus, which is the state the chain was in before #1294.

## Out of scope

- The contract side: `ValidatorInfo.region`, `setValidatorRegion` and the `ValidatorRegionUpdated` event in tn-contracts, plus the "Region-Aware Committee Shuffle" section of `design.md` and `invariants.md` that says the client implements it.  Those are tn-contracts PRs, and the submodule pin is untouched here.
- The future region spec.

## Testing

- `cargo +nightly fmt -- --check` on tn-reth, tn-types and telcoin-network-cli.
- `cargo +1.94 check -p tn-reth -p tn-types -p telcoin-network-cli --all-targets`.
- `cargo +1.94 test -p tn-reth --lib -- assemble`: the four pre-existing committee-assembly tests, including the seed-pinned golden order.
- `cargo +1.94 test -p tn-types --lib -- forks::`: the remaining fork-gate tests.
- CI runs the nightly clippy and the workspace nextest lane on push;  the pinned `+1.94` attest runs on the second machine.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MavenRain:
Putting into draft in preparation for eradication of the region concept

MavenRain:
I've ripped as much of regions from the codebase in this PR as I could.  Removing the field entirely requires making the necessary changes to tn-contracts, and then consuming the updated submodule here.

grantkee (dismissed):

grantkee (approved):

PR #1341, opened 2026-09-08
fix(batch-builder): skip empty batch proposals

### LINKED ISSUE #1338: Batch builder seals and broadcasts an empty batch when the pending set drains mid-build
## Problem

The build gate and the transaction selection that follows it read the pool at two different times, in two different tasks. Nothing between them rejects the empty batch that results when the pending set drains in the gap.

The gate checks the pool in the `BatchBuilder` loop:

```rust
// crates/batch-builder/src/lib.rs:391
if self.pool.pending_transactions().is_empty() {
    self.max_delay_interval.reset();
} else {
    self.pending_task = Some(self.spawn_execution_task());
}
```

`spawn_execution_task` spawns a separate task (`lib.rs:207`), and `build_batch` calls `pool.best_transactions()` inside it. Anything that empties the pending sub-pool between the gate and that call produces a batch with no transactions: the `while let Some(pool_tx) = best_txs.next()` loop never runs and `build_batch` returns `Batch { transactions: vec![], .. }`.

From there the batch reaches the network unchecked:

1. `lib.rs:214-218` calls `batch.seal_slow()` and `to_worker.send(...)` unconditionally.
2. `Worker::seal` (`crates/consensus/worker/src/worker.rs:300-310`) goes straight to `quorum_waiter.verify_batch` on the committee-validator branch. The node never validates its own batch.
3. `quorum_waiter.rs` has no empty-batch guard, so vote requests go out to every peer.
4. Each peer rejects it: `validate_batch_size_bytes` calls `.reduce()` over an empty iterator, gets `None`, and returns `BatchValidationError::EmptyBatch` (`crates/batch-validator/src/validator.rs:133`).

Two things can drain the pending set inside that window, both on the canonical-update path:

- `process_canon_state_update` removes the last pending transaction when a `Commit` lands.
- `update_accounts` demotes the last pending transaction out of `pending` on the real post-execution balance. This is the underfunded-sender case from #1158: the sender spends on the transaction that was just mined, the optimistic balance kept the follow-up affordable, and the authoritative update parks it.

The second is the more likely trigger, and needs no attacker. I traced the code path but did not reproduce the race, so treat reachability as the unproven part of this report.

## Impact

Committee validators are exempt from peer penalties, so this does not ban the proposer. `Peer::apply_penalty` bypasses the score model for exempt peers (`crates/network-libp2p/src/peers/peer.rs:288-312`), and `TrustBasis` covers the previous, current, and next committees. The observer branch is already guarded: `disburse_txns` returns `Ok(())` on an empty batch (`worker.rs:256-258`).

What each occurrence costs:

- One batch round for that worker. The build, the vote requests to every peer, and the rejection all happen before `BuildOutcome::Failed` comes back. At the 1000 to 2500 ms header delay that is a full round of the worker's throughput.
- A false Byzantine signal on every peer. Each one logs `warn!` "skipping severe/fatal penalty for exempt peer" against an honest validator, and `tn_network.peer_penalties_total{severity="fatal"}` increments even when the ban is suppressed, because `manager.rs:508` records the metric before `all_peers.rs:264` applies the exemption. That counter is what an operator would watch to find a genuinely faulty committee member.

The penalty exemption is the only reason this is a throughput and observability problem rather than a liveness one. `EmptyBatch` is classified `Penalty::Fatal` (`crates/consensus/worker/src/network/error.rs:94`), and `Penalty::Fatal` maps to `config.min_score` (`crates/network-libp2p/src/peers/score.rs:93`), which disconnects and bans. If the exemption ever narrows, a validator that hits this race bans itself from the peers it asked for votes.

This is not a regression. The merge base for the current work (`87364b23`) has no guard either, the seal path is unchanged since #596 and #733, and the gate dates to #1153.

## Proposed solution

Short-circuit in the spawned task, before `seal_slow()` at `lib.rs:213-218`. An empty batch should skip the seal entirely, reset `max_delay_interval`, and take the deferred-build path that `mined_transactions.is_empty()` already uses at `lib.rs:454`. That existing check runs too late to help: the batch is on the wire by the time it is reached.

Add the same check to `Worker::seal` as well, so no caller can put an empty batch on the wire regardless of what the builder does. The observer branch already returns early on empty; the committee-validator branch should match.

A regression test needs to drain the pool between the gate and `best_transactions()`. That is also what would turn the reachability argument above into proof.

## How this was found

Reviewing the #1303 batched balance reads. The empty sender set first showed up as a wasted state-provider acquisition in `get_account_balances`; tracing whether that set could actually be empty led to the broadcast path, which costs far more than the read transaction.

MavenRain:
Closes #1338.

## Problem

A canonical account update can drain the pending pool after the batch builder's gate but before transaction selection. The resulting empty batch was sealed and sent to the worker, which requested quorum even though peers reject empty batches. This can happen to an honest validator with no attacker required, wasting a proposal and producing misleading peer-penalty signals.

## Changes

- [x] **Defer empty builds:** return an explicit empty outcome before sealing or sending to the worker. Reuse the existing timer reset and deferred-build path without updating the pool, recording a seal failure, or changing refusal backoff.
- [x] **Guard the worker boundary:** acknowledge empty batches as a successful no-op before forwarding, quorum, persistence, reporting, or broadcast.
- [x] **Add bounded regressions:** demote the pending transaction before the spawned build runs, verify no empty proposal, and verify that the restored transaction is proposed after the batch delay. Pause time only for the deterministic early-rebuild assertion, then resume before awaiting spawned work, with a 10-second watchdog. Cover empty validator and observer seals with no quorum, network, or storage effects.

## Testing

- Passed: `gateledger run -- cargo +nightly-2026-03-20 fmt --all --check` and `git diff --check`.
- Passed: independent diff review, including CI-weakening and timer/backoff checks.
- Passed in [hosted validation](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34290539093): `./etc/ci-lanes.sh clippy` (both passes), `./etc/ci-lanes.sh test-default --build-only`, and `./etc/ci-lanes.sh test-adiri --build-only`.
- Passed: `NEXTEST_PROFILE=issue1338-strict ./etc/ci-lanes.sh test-default`, 1,709 passed and 32 skipped in 141.337 seconds. The empty-build regression passed in 0.234 seconds within this full suite.
- Passed: `NEXTEST_PROFILE=issue1338-strict ./etc/ci-lanes.sh test-adiri`, 907 passed and 6 skipped in 59.449 seconds.
- The validation profile preserves existing test groups, disables retries, and fails any test reaching 60 seconds with no grace period. No `SLOW` events occurred in the run.
- Passed: all three production regressions before and after mutation checks. Disabling the builder guard fails with `empty batch reached worker`; disabling the worker guard fails the validator's immediate-success assertion. Both are actual test failures, not compilation failures or timeouts. The observer control still passes with the worker guard disabled, confirming its existing empty-disbursement behavior.
- Passed: deliberately replacing the initial build result with a permanently pending future makes the regression fail in 10.011 seconds with `empty-build regression exceeded its 10-second limit`. Restoring the source restores all three passing regressions.

Exact regression command:

```sh
cargo nextest run --locked --workspace --no-fail-fast --retries 0 -- \
  tests::empty_build_after_pending_pool_drain_is_deferred \
  batch_provider_tests::validator_empty_seal_is_noop \
  batch_provider_tests::observer_empty_seal_is_noop --exact
```

Validation commit `59dbb3c8` has identical Rust sources to PR head `a2ee92ae`. Its only additional change is the workflow code that runs the regressions, mutations, watchdog check, and strict full suites. That workflow change is outside this PR. Source and configuration mutations are restored before the job completes.

Pending: archive/RocksDB guards and the ignored all-feature e2e suite on the PR head, followed by maintainer on-chain attestation. Maintainer PR-head compile/test lanes are skipped by repository policy; `verify-on-chain` is currently red because this commit is unattested. This PR remains draft.

### DIFF
--- a/crates/batch-builder/src/lib.rs
+++ b/crates/batch-builder/src/lib.rs
@@ -575,11 +575,14 @@ mod tests {
         );
         assert!(txpool.pending_transactions().is_empty());
 
-        let outcome = timeout(Duration::from_secs(1), done)
-            .await
-            .map_err(std::io::Error::other)?
-            .map_err(std::io::Error::other)?
-            .map_err(std::io::Error::other)?;
+        let outcome = tokio::select! {
+            outcome = done => outcome
+                .map_err(std::io::Error::other)?
+                .map_err(std::io::Error::other)?,
+            _ = from_batch_builder.recv() => {
+                Err::<BuildOutcome, _>(std::io::Error::other("empty batch reached worker"))?
+            }
+        };
         assert_matches!(outcome, BuildOutcome::Empty);
         assert_matches!(
             from_batch_builder.try_recv(),
@@ -604,16 +607,25 @@ mod tests {
             .send(Ok(outcome))
             .map_err(|_| std::io::Error::other("build result receiver closed"))?;
         batch_builder.pending_task = Some(done);
-        let builder_task = tokio::spawn(batch_builder.run());
+        let mut builder_task = Box::pin(batch_builder.run());
 
-        // Virtual time makes the stale-tick and immediate-rebuild assertions deterministic.
-        assert!(timeout(delay - Duration::from_millis(1), from_batch_builder.recv())
-            .await
-            .is_err());
-        let (batch, ack) = timeout(delay, from_batch_builder.recv())
-            .await
-            .map_err(std::io::Error::other)?
-            .ok_or_else(|| std::io::Error::other("worker channel closed"))?;
+        // Register the reset deadline before advancing time. Reth's blocking validation
+        // tasks inhibit paused-clock auto-advance, so move the clock explicitly.
+        assert!(futures_util::poll!(&mut builder_task).is_pending());
+        tokio::time::advance(delay - Duration::from_millis(1)).await;
+        assert!(futures_util::poll!(&mut builder_task).is_pending());
+        assert_matches!(
+            from_batch_builder.try_recv(),
+            Err(tokio::sync::mpsc::error::TryRecvError::Empty)
+        );
+        tokio::time::advance(Duration::from_millis(1)).await;
+        let (batch, ack) = tokio::select! {
+            result = &mut builder_task => Err(std::io::Error::other(format!(
+                "builder exited before proposing a batch: {result:?}"
+            ))),
+            batch = from_batch_builder.recv() =>
+                batch.ok_or_else(|| std::io::Error::other("worker channel closed")),
+        }?;
         assert_eq!(batch.batch().transactions().len(), 1);
         let encoded = batch
             .batch()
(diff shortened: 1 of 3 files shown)


grantkee (approved):
LGTM!

PR #1342, opened 2026-09-08
fix(node): tolerate governance worker-count changes at epoch entry

### LINKED ISSUE #1336: node: strict local-vs-chain worker-count equality at epoch entry turns a setNumWorkers governance action into a fleet-wide crash loop
Found while reviewing #1315 (per-worker network swarms, #555 PR 3). The check added there is
correct at startup but fires again at every epoch entry, where halting is a self-inflicted
fleet-wide outage.

## Problem

`check_configured_worker_count` (`crates/node/src/manager/node.rs:87-103`) enforces **strict
bidirectional equality** between the node's local `node_info.p2p_info.workers` length and the
on-chain `WorkerConfigs` count. It is called from two places:

- startup, via `prepare_worker_networks` (`node.rs:121`, reached from `node.rs:845`)
- **every epoch entry**, via `check_committee_worker_count` (`start_epoch.rs:1143`), which runs it
  *before* the `num_workers == 1` early return and *before* the multi-workers fork gate

The equality is deliberate — the covering tests are titled *"Startup requires the local count to
match the raw chain-derived count, in both directions"* (`node.rs:1524-1541`) and
`epoch_entry_rejects_changed_worker_count` (`start_epoch.rs:1237-1256`).

`setNumWorkers(uint16) external onlyOwner` (`tn-contracts/src/consensus/WorkerConfigs.sol:121`) is
governance-mutable and takes effect at the next epoch boundary. On non-adiri builds
`multi_workers_build_fork_active` returns `true` unconditionally
(`crates/types/src/forks.rs:496-499`), so nothing gates this on mainnet.

## Consequence

An `Err` from the epoch-entry call site reaches `std::process::exit(1)` with no retry, no observer
fallback and no degraded mode:

| Hop | Location | Mechanism |
|---|---|---|
| 1 | `start_epoch.rs:365-369` | `check_committee_worker_count(...)?` in `create_committee_from_state` |
| 2 | `start_epoch.rs:1143` | `super::check_configured_worker_count(...)?` |
| 3 | `start_epoch.rs:180-181` | `self.create_committee_from_state(...).await?` |
| 4 | `run_epoch.rs:152` | `get_committee_with_epoch_start_info(engine).await?` |
| 5 | `node.rs:1231-1233` | `run_epochs`: `epoch_result.inspect_err(...)?` — no `RunEpochMode::ModeChange` conversion |
| 6 | `node.rs:1072` | select arm returns from `run()` |
| 7 | `crates/node/src/lib.rs:60-64` | logs `"Error running node"`, returns `Err` |
| 8 | `crates/telcoin-network-cli/src/cli.rs:194` | `node_join.await?` |
| 9 | `bin/telcoin-network/src/main.rs:18-23` | `eprintln!` -> `std::process::exit(1)` |

`retry_provider_faults` wraps only the state read (`start_epoch.rs:358`); the check runs after the
retry closure, so it is not retried.

Hop 4 sits near the top of `run_epoch`, which `run_epochs` calls once per loop iteration for every
mode (`Initial`, `NewEpoch`, `ModeChange`). Nothing in the path is CVV-gated, so **observers and
RPC nodes halt too**.

It is a **crash loop**, not a diagnosable halt: `exit(1)` under systemd/k8s restarts the process,
which then fails *earlier* at `node.rs:845` on the same check.

There is **no in-band recovery** — `spawn_node_networks` has exactly one call site (`node.rs:845`,
inside `run()`), `worker_event_streams` is sized once from config at `node.rs:690-692`, and
`self.builder.tn_config` is loaded once per process. Recovery requires editing `node-info.yaml`
and restarting on every host.

And because equality is strict in **both** directions, **no operator ordering avoids downtime**:
configure `N+1` before governance bumps and the node won't start; configure `N` after the bump and
it halts at the next boundary.

## Why this contradicts the execution layer

`crates/tn-reth/src/evm/block.rs:551-556`, verbatim:

> "The worker set comes from the CONTRACT's `numWorkers`, not the accumulator's. **Governance may
> have grown the worker set mid-epoch** (a `setNumWorkers` only takes effect at this boundary)... A
> worker with no accumulator slot yet prices from `(MIN_PROTOCOL_BASE_FEE, 0 gas)` — the fresh-slot
> rule."

And `test_boundary_added_worker_data_written_at_creation_close`
(`crates/node/tests/it/main.rs:2336-2355`) is a **pre-existing** integration test that drives
`setWorkerConfig(1, ..)` + `setNumWorkers(2)` in an epoch-0 block and asserts the boundary prices
the new worker correctly. Boundary-effective worker-count growth is a designed, tested,
protocol-supported governance operation — and the next epoch entry after it is now fatal fleet-wide.

## Why continuing is safe in this build

Each failure mode fail-stop would supposedly prevent was checked:

- **Header validation** (`crates/types/src/primary/header.rs:161-166`) rejects
  `worker_id >= committee.number_of_workers()`, reading the **committee**, never local config. A
  local shortfall cannot make this node accept or reject differently from its peers. No divergence.
- **Accumulator/committee agreement** (`run_epoch.rs:341-348`) already has its own independent
  `ensure!`, chain-derived on both sides.
- **Batch production**: `spawn_worker_node_components` hardcodes
  `let worker_id = DEFAULT_WORKER_ID;` (`start_epoch.rs:463`) and handles are read with graceful
  `.get(..).ok_or_else(..)` (`start_epoch.rs:479-481`, `:531-534`). No node in this build produces
  worker >= 1 batches, so there are none to miss.
- **Batch fetching**: only reachable once a peer commits a worker-1 batch, impossible until #557.

Running with `configured=1, on_chain=2` yields one extra fee-config row priced deterministically by
every node, one idle committee slot, and a warning. Degraded throughput at worst — the pre-#1315
`warn!` was substantively correct for this build.

## Fix

**A (recommended now) — make the epoch-entry check warn-only until #557 lands.** Keep the startup
equality check; it earns its keep on genesis and per-validator config drift.

```rust
// start_epoch.rs, replacing the unconditional `super::check_configured_worker_count(...)?`
if configured_workers != num_workers.get() {
    warn!(
        target: "epoch-manager",
        epoch, configured = configured_workers, on_chain = num_workers.get(),
        "local worker swarm count disagrees with chain state; continuing with worker 0 only. \
         Update `node_info.p2p_info.workers` and restart at your convenience"
    );
}
// Once #557 drives epoch components per worker, a local SHORTFALL becomes fatal (this node
// cannot fetch batches for ids it has no swarm for) while a surplus stays benign:
//   eyre::ensure!(configured_workers >= num_workers.get(), "...");
```

**B (recommended alongside A) — relax both call sites from `==` to `>=`,** so a zero-downtime
governance path exists at all:

```rust
eyre::ensure!(
    configured_workers >= on_chain_workers,
    "node config `node_info.p2p_info.workers` lists {configured_workers} workers but the \
     chain-derived count for epoch {epoch} is {on_chain_workers}: a node must configure at \
     least the worker count the committee carries"
);
```

That makes a bump survivable: roll the fleet to `N+1` in config (a surplus is tolerated), then
governance `setNumWorkers(N+1)`, and the boundary is a no-op for everyone. Surplus swarms are
provably inert today — higher-id swarms never call `start_listening`
(`start_epoch.rs:463-464`).

**C (larger, needs a decision)** — make the count reloadable in-band: refactor
`spawn_node_networks` to add swarms for new ids without touching live ones, and call it from the
epoch-entry path when the chain count rises. Keys are already derivable per id
(`KeyConfig::worker_network_keypair`, `crates/config/src/keys.rs:512-529`). Note a separate blocker
for going genuinely multi-worker: `self.bootstrap_servers` is frozen from the **genesis** committee
file at `node.rs:693-702`, and `spawn_worker_network_for_epoch` does `v.worker(*worker_id)` ->
`None` -> drops the peer (`start_epoch.rs:830-833`), so worker >= 1 swarms would have no bootstrap
peers.

## Regardless of the option chosen

1. Add the missing genesis cross-check between `NetworkGenesis::agreed_num_workers()`
   (`crates/config/src/genesis.rs:158-186`) and the `--worker-fee-config` count
   (`crates/telcoin-network-cli/src/genesis/mod.rs:169-226`). Nothing ties them together today, and
   this is the one mismatch class where failing loudly costs nothing.
2. Add an e2e test driving `setNumWorkers` across a live epoch boundary — that is what would have
   caught the contradiction with `test_boundary_added_worker_data_written_at_creation_close`.
3. Align check/fork-gate ordering between `node.rs:113-121` (fork gate first) and
   `start_epoch.rs:1143-1155` (equality first) so adiri produces the same diagnostic from both.

## Notes

Introduced by #1315; the pre-#1315 signature was `check_committee_worker_count(epoch, num_workers)`
with no `configured_workers` parameter at all
(`git show 66e0d14a:crates/node/src/manager/node/start_epoch.rs`), and post-fork it fell through to
a `warn!` and returned `Ok(())`.

Trigger requires a privileged `onlyOwner` transaction — deliberate, not attacker-reachable, not
reachable by accident in steady state. That is the only thing keeping this out of High. Should be
resolved before either `setNumWorkers` is used on a live network or #557 lands.

MavenRain:
Fixes #1336.

## Problem

A legitimate `WorkerConfigs.setNumWorkers` governance update takes effect at the next epoch boundary, while running nodes retain their original worker swarms. Requiring exact equality between local and on-chain counts stops validators, observers, and RPC nodes after a count change. No attacker is required. Strict startup equality also prevents operators from provisioning spare workers before an increase.

## Changes

- **Keep live epoch transitions running.** Warn on local/on-chain count mismatches and continue with this version's worker 0 epoch components. Reject an empty local worker configuration and retain the pre-fork layout guard. Full multi-worker participation and shortfall recovery remain part of #557.
- **Allow advance provisioning.** Startup requires at least the on-chain worker count, permits surplus swarms after the fork, and rejects zero counts before network creation.
- **Keep genesis counts consistent.** Reject mismatches between validator worker counts and `--worker-fee-config` entries before deploying contracts or writing ceremony outputs.
- **Cover governance increases and decreases.** Add startup, epoch-entry, fork-boundary, and genesis regressions, plus a process test that changes the on-chain count from 1 to 2 to 1 and verifies completed epochs from every original node without restarting it.

## Testing

- `git diff --check origin/main...HEAD` passed.
- `gateledger run --ledger .validation/gates -- cargo +nightly-2026-03-20 fmt --all -- --check` passed.
- `diffclass --base origin/main`, `apidelta --base origin/main --verb clippy`, and `buildplan --base origin/main --all-targets clippy` require build validation and select 14 packages. No local Clippy pass is claimed; the previous attempt reached the local free-space guard.
- [E2e run on `bc57acd1`](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34276558179) passed all 4 restart tests and all 5 selected epoch tests. `test_epoch_worker_count_changes_keep_nodes_running` passed in 22.405 seconds. The workflow also compiled the workspace test binaries under pinned Rust 1.94.
- After merging current `main`, fresh validation was dispatched on `01d84d8f`: [restart and epoch e2e suites](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34378338646) and [both Clippy feature sets plus default/adiri test compilation](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34378350989). These runs are pending; the earlier pass does not attest the merged commit.

The PR remains a draft pending the fresh runs, targeted unit-test execution in both feature sets, mutation confirmation, and full attestation. The [required attestation check](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34378311731) fails because `01d84d8f` is not attested; a maintainer must complete `make attest` on this commit and then request a review to rerun the check. The existing maintainer/draft rules skip normal PR compile/test lanes, so skipped jobs provide no test evidence. No CI configuration is changed by this PR.


sstanfield (approved):

grantkee (approved):
LGTM!

PR #1343, opened 2026-09-08
fix(node): validate per-worker network keys and addresses

### LINKED ISSUE #1337: node: prepare_worker_networks never validates per-worker network_key or address uniqueness, and its covering test assertion is tautological
Found while reviewing #1315 (per-worker network swarms, #555 PR 3).

## Problem

`prepare_worker_networks` (`crates/node/src/manager/node.rs:109-149`) is documented as *"Validate
every worker and pair its configuration with its persistent event stream"*, and it does validate:

- the multi-workers fork gate (`:116-120`)
- chain-vs-config worker count (`:121`)
- `WorkerId` range (`:122-125`)
- event-stream count (`:126-130`)
- each worker's `rpc` descriptor (`:137-141`)

It does **not** validate that `workers[i].network_key` matches the key actually derived for worker
`i`, nor that `network_address` values are distinct across workers. It does not even take a
`KeyConfig` parameter, so it cannot check the key.

Nothing else in the repo covers the gap. Exhaustive grep for `worker_network_public_key` yields 13
hits; the only production consumers are `start_epoch.rs:235` (stamps the **derived** key into RPC
`NodeInfo`, ignoring config), `crates/telcoin-network-cli/src/keytool/generate.rs:196` (**writes**
the derived key), and `crates/network-libp2p/src/consensus.rs:368` (`new_for_worker` derives the
swarm keypair from `key_config.worker_network_keypair(worker_id)`) — **the config `network_key`
field is never read to build the swarm.** Genesis validation
(`crates/config/src/genesis.rs:133-143`) checks only proof-of-possession and `agreed_num_workers()`.
`NodeP2pInfo`'s custom `Deserialize` (`crates/types/src/primary/info.rs:28-35`) only enforces
non-empty. `Config::update_worker_network_key` (`crates/config/src/node.rs:195-206`) is an
unchecked setter. No `HashSet`/`dedup` guard over `p2p_info.workers` exists in `crates/config`,
`crates/types` or `crates/node`.

## The covering test assertion is tautological

`prepare_worker_networks_preserves_worker_identity_and_event_streams` (`node.rs:1459-1492`) looks
like it covers the key check, but its own fixture builds the entry with that key:

```rust
// node.rs:1446-1456 -- fixture
network_key: keys.worker_network_public_key(worker_id),   // :1450
```
```rust
// node.rs:1475-1478 -- assertion
assert_eq!(worker.p2p.network_key, keys.worker_network_public_key(worker.worker_id));
```

Since `:1472` already asserts `worker.worker_id == expected_id` and `:1473` asserts
`worker.p2p == *configured`, the key assertion is logically implied by the two preceding assertions
plus the fixture. It would still pass if production stripped or scrambled the field. (The rest of
the test is worthwhile — it does pin index/event-stream alignment.)

## Consequence

Swarm identity is **correct regardless**, because `consensus.rs:368` derives the keypair from
`worker_id` — the bad config field never reaches the swarm. But `p2p.network_address` **is** used:
passed at `node.rs:1163` -> `swarm.add_external_address` (`consensus.rs:561`) and into the signed
`NodeRecord` (`consensus.rs:567-572`, `:605-611`). A duplicated entry therefore makes worker 1's
swarm sign a record advertising worker 0's address under the `Worker(1)` `RecordDomain` — a
self-inconsistent advertisement.

Today this is latent and there is no bind conflict, because only worker 0 gets per-epoch components
(`start_epoch.rs:463-464`), so higher-id swarms never call `start_listening`.

**Once #557 lands**, the duplicate address fails loudly — `libp2p-quic`'s `create_socket` binds with
no `SO_REUSEADDR`/`SO_REUSEPORT` (`libp2p-quic-0.13.0/src/transport.rs:183-196`) and `listen_on`
binds synchronously, so the second `swarm.listen_on` errors and `NetworkCommand::StartListening`
(`consensus.rs:812-814`) fails epoch start. The **key** mismatch would still be silent: peer-side,
`create_committee` (`crates/config/src/genesis.rs:189-208`) copies `worker_p2p_nodes()` verbatim, so
peers dial worker 0's address/PeerId on their worker-1 swarm; QUIC/TLS succeeds but every wire
protocol is worker-id-namespaced (`crates/network-libp2p/src/types.rs:87,96,109,124`) so all
negotiations fail. Net effect: worker 1 silently unreachable.

## Fix

Thread `&KeyConfig` in and add the two missing invariants:

```rust
fn prepare_worker_networks<Events: Clone>(
    workers: &[P2pNode],
    event_streams: &[Events],
    key_config: &KeyConfig,
    epoch: Epoch,
    on_chain_workers: usize,
) -> eyre::Result<Vec<PreparedWorkerNetwork<Events>>> {
    // ... existing count / range / event-stream checks unchanged ...

    // Distinct listen addresses: each swarm binds its own UDP socket, so a shared
    // address is a bind conflict (or, before per-worker listening lands, a silently
    // wrong advertisement).
    let mut seen = std::collections::HashSet::with_capacity(configured);
    for (worker_id, p2p) in workers.iter().enumerate() {
        eyre::ensure!(
            seen.insert(&p2p.network_address),
            "node config `node_info.p2p_info.workers[{worker_id}].network_address` ({}) \
             duplicates an earlier worker: every worker swarm binds its own socket",
            p2p.network_address
        );
    }

    workers
        .iter()
        .zip(event_streams)
        .zip(0..=WorkerId::MAX)
        .map(|((p2p, event_stream), worker_id)| {
            // The swarm always authenticates with the DERIVED keypair
            // (`ConsensusNetwork::new_for_worker`), so a config key that disagrees is an
            // advertisement peers can never reach. Reject it at startup rather than
            // publishing a signed NodeRecord that contradicts the swarm's own identity.
            let expected = key_config.worker_network_public_key(worker_id);
            eyre::ensure!(
                p2p.network_key == expected,
                "node config `node_info.p2p_info.workers[{worker_id}].network_key` does not \
                 match the key derived for worker {worker_id} (expected {expected:?}, found \
                 {:?}): re-run `keytool` or fix the entry",
                p2p.network_key
            );
            p2p.rpc.as_ref().map(tn_types::RpcInfo::validate).transpose().wrap_err_with(|| {
                format!(
                    "invalid `node_info.p2p_info.workers[{worker_id}].rpc` endpoint in node config"
                )
            })?;
            Ok(PreparedWorkerNetwork {
                worker_id,
                p2p: p2p.clone(),
                event_stream: event_stream.clone(),
            })
        })
        .collect()
}
```

Pass `&self.key_config` at the call site (`node.rs:1095-1100`). Replace the tautological assertion
at `:1475-1478` with a **negative** test — mutate `workers[1].network_key` to
`keys.worker_network_public_key(0)` and assert the error names `workers[1].network_key` — plus a
duplicate-address rejection test.

## Collateral effects

- **Test fixtures break on multi-worker paths.** `crates/test-utils-committee/src/builder.rs:167-175`
  assigns workers 1.. **fresh random** ed25519 keys (`NetworkKeypair::generate_ed25519()`) rather
  than derived ones, and `crates/test-utils-committee/src/authority.rs:153-171` copies those into
  `node_info.p2p_info.workers`. The builder needs to use
  `key_config.worker_network_public_key(worker_id)` for all ids — which also extends
  `fixture_tests.rs:25`'s existing invariant from worker 0 to every worker.
- **Operator-visible behaviour change**: a `node-info.yaml` whose worker keys were generated against
  a different BLS keystore than the one now loaded would be refused at startup. Intended semantics
  (such a node is unreachable anyway), but it is a new hard failure and belongs in release notes.
- No on-disk, wire-format or determinism impact — the check is startup-local.

## Notes

Low severity: operator self-harm on a path triple-gated today — `multi_workers_fork_active`
(`node.rs:116-120`), `keytool generate` hard-rejects `workers != 1` (`generate.rs:225-231`), and the
per-epoch loop is worker-0-only. No consensus-safety, determinism or cross-validator impact; damage
is confined to the misconfigured node's own reachability. The real cost is diagnosability, in a
function whose doc comment already claims to validate every worker before *"leaving partially
spawned networks"*.

Worth closing before #557 lands, since that is when the duplicate-address case turns into a failed
epoch start.

MavenRain:
Closes #1337.

## Problem

An operator can configure worker network keys that differ from the loaded BLS keystore or configure workers with colliding listen addresses. Startup previously accepted these configurations, allowing authentication failures or address conflicts after network startup. No attacker is required: this is a local configuration error.

## Changes

- [x] **Validate every worker before spawning swarms.** Each advertised key must match the identity derived for its worker ID. Duplicate-address checks remove trailing `/p2p/` identities before comparison and retain the configured address in errors.
- [x] **Print usable configuration values.** `NetworkPublicKey::Display` uses the same Base58 protobuf encoding as YAML. Key-mismatch errors print the expected and configured keys in that format, with a YAML round-trip regression.
- [x] **Keep fixtures consistent.** Advertised keys use the authority's per-worker derivation. Placeholder addresses use distinct loopback hosts derived from worker IDs; fixture regressions check keys and unique bootstrap and startup addresses across all authorities.
- [x] **Keep routine boundary coverage cheap.** The default overflow test derives one key. Full-range success remains an explicitly ignored test with its cost documented and avoids unused RPC parsing.
- [x] **Update operator documentation and merge current main.** The merge preserves main's spare-worker provisioning and pre-fork validation.

## Testing

- Pinned formatting passed: `gateledger run -- cargo +nightly-2026-03-20 fmt --all -- --check`. `git diff --check` and strict self-review passed.
- The refreshed source-extraction harness passed under Rust 1.94: `testledger -- cargo +1.94 test --offline --locked -j 2 -- --test-threads 2`, with an isolated Cargo cache and target directory and `RUSTC_WRAPPER=`. It uses the repository's serde_yaml 0.8.26. Default and `--features adiri` each passed 13 tests, with one ignored full-range case. Running `prepare_worker_networks_accepts_full_worker_id_range -- --ignored --test-threads 2` explicitly also passed.
- Four independent mutations compiled and failed their intended regressions: raw address comparison, Debug-formatted expected keys, Debug-formatted Display, and identical fixture placeholder addresses. All 15 extracted source spans match the current checkout byte for byte.
- The harness executes the actual startup logic and tests, cryptographic derivation, key formatting, and fixture address/mapping code. It substitutes the queue and complete fixture construction, so full node and committee integration coverage remains outstanding.
- The emitted 25-crate Clippy check completed under nightly-2026-03-20 with `--offline --locked -j 2 --all-targets --no-deps -- -D warnings` and the complete package list. It exited 101 on 25 lint errors in `crates/e2e-tests/tests/it/`; those files are unchanged from `main`, confirmed by `git diff --exit-code origin/main HEAD -- crates/e2e-tests`. This maintainer-authored PR skips GitHub's compile/test lanes. Full workspace tests and a fresh on-chain attestation remain outstanding; skipped checks are not counted as validation.
- [CI run 34683247866](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34683247866) ran for `8db124e600b577ce4688d60cb2a713215c5453dd`. Its `verify-on-chain` check failed because that commit has not been attested. A maintainer must run `make attest` on this commit and then request a review to rerun the check.


crates/node/src/manager/node.rs
```diff
@@ -129,11 +136,26 @@ fn prepare_worker_networks<Events: Clone>(
...
         .iter()
         .zip(event_streams)
         .zip(0..=WorkerId::MAX)
         .map(|((p2p, event_stream), worker_id)| {
+            eyre::ensure!(
+                addresses.insert(&p2p.network_address),
+                "node config `node_info.p2p_info.workers[{worker_id}].network_address` ({}) \
+                 duplicates an earlier worker: every worker swarm needs its own address",
+                p2p.network_address
+            );
```
grantkee, line 181: The duplicate-address check compares Multiaddrs byte for byte, so two worker entries that differ only in their trailing /p2p/ suffix pass it and then collide on the same UDP socket when the second one binds.

libp2p-quic's `multiaddr_to_socketaddr` parses and discards a trailing `Protocol::P2p` and binds only `(ip, udp port)`. The node-info address reaches `Swarm::listen_on` with the suffix intact: `consensus_config.worker_address(worker_id)` (`start_epoch.rs:849-851`) → `NetworkHandle::start_listening` (`network-libp2p/src/types.rs:574-579`) → `self.swarm.listen_on(multiaddr)` (`consensus.rs:812-815`). Production already depends on this, since every address keytool writes carries `/p2p/<peer_id>` (`keytool/generate.rs:202-215`), and nothing on the config path strips it — `NodeP2pInfo`'s deserializer only enforces a non-empty list, and none of the `validate()` functions in `tn-config` look at `network_address`.

So `/ip4/X/udp/9000/quic-v1/p2p/A` and `/ip4/X/udp/9000/quic-v1/p2p/B` are unequal here but the same socket. `create_socket` binds with no `SO_REUSEADDR`/`SO_REUSEPORT`, so the second bind returns `AddrInUse` → `NetworkError::Listen` → `start_epoch.rs:863 ?` → the node exits. That happens after `spawn_node_networks` has already constructed and spawned every swarm as a critical task, which is the partially-spawned state the doc comment on this function says the up-front check prevents. `/ip4/0.0.0.0/udp/9000` vs `/ip4/127.0.0.1/udp/9000` slips through the same way.

Two scope notes so this isn't overstated. keytool never writes a multi-worker file (`validate()` rejects `--workers != 1`), so a multi-worker `node-info.yaml` is hand-edited — but the natural way to hand-edit is to copy worker 0's entry, change the key (which this PR now forces) and the `/p2p/` suffix to match, and forget the port, which is exactly the shape this misses. And worker-vs-worker collision is unreachable until #557: `spawn_worker_node_components` hard-codes `DEFAULT_WORKER_ID` (`start_epoch.rs:463-464`), so workers 1..N are spawned but never bind today. The same gap between the primary and worker 0 on one port is reachable today, but that's outside #1337's remit.

The existing test `prepare_worker_networks_rejects_duplicate_network_addresses` copies a byte-identical address, and `worker_p2p` never appends `/p2p/`, so the suffix-only variant has no coverage.

Smallest fix I can see is to key the set on the address a QUIC listener actually binds. `tn_types` already re-exports `Protocol`. Keeping the raw configured address in the message means the existing `contains(&first_address.to_string())` assertion still holds.

```rust
use tn_types::{Multiaddr, Protocol};

/// The address a QUIC swarm actually binds: libp2p-quic parses and discards trailing
/// `/p2p/<id>` components on `listen_on`, so two entries that differ only there share a socket.
fn listen_address(addr: &Multiaddr) -> Multiaddr {
    let mut listen = addr.clone();
    while matches!(listen.iter().last(), Some(Protocol::P2p(_))) {
        listen.pop();
    }
    listen
}
```

```rust
let mut addresses: HashSet<Multiaddr> = HashSet::with_capacity(configured);
// ...
eyre::ensure!(
    addresses.insert(listen_address(&p2p.network_address)),
    "node config `node_info.p2p_info.workers[{worker_id}].network_address` ({}) \
     duplicates an earlier worker's listen address (a differing `/p2p/` suffix is still \
     the same socket): every worker swarm needs its own address",
    p2p.network_address
);
```

Plus one more case in the duplicate-address test where the duplicate is `first_address.clone().with(Protocol::P2p(keys.worker_network_public_key(duplicate_id).into()))`.

A fuller `(ip, port)` comparison that also treats `0.0.0.0` as colliding within its family and includes the primary address would close the rest, but it changes the signature and all nine call sites and widens the PR past #1337 — I'd take that as a follow-up alongside the primary-key check.

crates/node/src/manager/node.rs
```diff
@@ -129,11 +136,26 @@ fn prepare_worker_networks<Events: Clone>(
...
+            let expected = key_config.worker_network_public_key(worker_id);
+            eyre::ensure!(
+                p2p.network_key == expected,
+                "node config `node_info.p2p_info.workers[{worker_id}].network_key` does not \
+                 match the key derived for worker {worker_id} (expected {expected:?}, found \
+                 {:?}): update this entry to match the loaded BLS keystore",
+                p2p.network_key
+            );
```
grantkee, line 189: The key-mismatch error prints the expected key with Debug formatting, which is libp2p's unpadded-hex form rather than the base58 string node-info.yaml uses, so the operator cannot paste the value the message tells them to set.

`NetworkPublicKey` derives `Debug` (`crates/types/src/crypto/network.rs:11-12`), which delegates to `libp2p::identity::PublicKey`. The ed25519 impl writes `PublicKey(compressed): ` followed by each byte as `{byte:x}` — unpadded, so `0x0a` prints as `a` and the string isn't even reliably decodable back to 32 bytes by hand. What lands in the log is shaped like:

```text
NetworkPublicKey(PublicKey { publickey: Ed25519(PublicKey(compressed): 5f3a9c…e1) })
```

The value in `node-info.yaml` comes from the custom `Serialize` at `network.rs:46-57`, which emits `bs58::encode(self.encode_protobuf())` for human-readable serializers, and `Deserialize::visit_str` reverses exactly that. That's the only paste-able form. `PeerId` renders as a base58 multihash (`12D3KooW…`), which also isn't it.

There's no `Display` impl on the type, and this `ensure!` is the first operator-facing site in the repo that formats a `NetworkPublicKey`, so `{:?}` was the only formatter available rather than a convention being followed. The README paragraph this PR adds tells operators to "Restore worker key entries that match the node's BLS keystore", which presumes the error hands them a usable value. The `found {:?}` half echoes back the value they already have in the file.

Suggest adding `Display` next to the `Serialize` branch it mirrors — `bs58` is already a `tn-types` dependency — and switching this message to `{}`:

```rust
/// Renders the key exactly as `node-info.yaml` and genesis files carry it
/// (Base58 of the libp2p protobuf encoding), so error messages are paste-able.
impl fmt::Display for NetworkPublicKey {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.write_str(&bs58::encode(self.encode_protobuf()).into_string())
    }
}
```

```rust
let expected = key_config.worker_network_public_key(worker_id);
eyre::ensure!(
    p2p.network_key == expected,
    "node config `node_info.p2p_info.workers[{worker_id}].network_key` does not \
     match the key derived for worker {worker_id} (expected {expected}, found {}): \
     set this entry to the expected value to match the loaded BLS keystore",
    p2p.network_key
);
```

`prepare_worker_networks_rejects_mismatched_network_keys` could then also assert `error.to_string().contains(&keys.worker_network_public_key(invalid_id).to_string())` to pin the paste-able format. `Debug` output for `?`-formatted logs is unchanged.

crates/node/src/manager/node.rs
```diff
@@ -1561,29 +1644,44 @@ mod tests {
...
         let max_workers = usize::from(WorkerId::MAX) + 1;
-        let worker = P2pNode { rpc: None, ..worker_p2p(&worker_key_config(), 0)? };
-        let workers = vec![worker; max_workers + 1];
+        let workers = (0..=WorkerId::MAX)
+            .map(|worker_id| {
+                worker_p2p(&keys, worker_id).map(|worker| P2pNode { rpc: None, ..worker })
+            })
+            .collect::<eyre::Result<Vec<_>>>()?;
+        let overflow_workers: Vec<_> =
+            workers.iter().cloned().chain(std::iter::once(worker_p2p(&keys, 0)?)).collect();
```
grantkee: This test now performs 131,073 uncached BLS-backed key derivations and takes about 19 seconds under the debug test profile, where the version on main did one.

`worker_network_public_key` → `worker_network_keypair` (`crates/config/src/keys.rs:512-524`) → `generate_network_keypair` (`keys.rs:549-555`) does one BLS min_sig signature, one blake3 hash, and one ed25519 keygen per call, and nothing caches the result — `KeyConfigInner` has no derived-key field, and the test at `keys.rs:1060` explicitly guards against one being added. Under `cargo test`, `.cargo/config.toml` sets `[profile.test.package."*"] opt-level = 1`, and blst only bumps its C code to `-O2` when built without `debug_assertions`, so the signing runs at `-O1`.

Count for this test: 65,536 in `worker_p2p` for `0..=WorkerId::MAX` in the block anchored here, one more for the overflow tail on line 1663, and another 65,536 inside `prepare_worker_networks`'s success path when it re-derives every key at line 151 — zero on the overflow path. I measured 18.72 s for this test alone; the whole `prepare_worker_networks` filter (nine tests) runs in 18.71 s, so this is the entire cost of the module's unit tests. On `58c0a62d` the test built one `P2pNode` and did `vec![worker; max_workers + 1]`.

The two halves aren't equally expensive to keep. The overflow rejection fires at the count check (`node.rs:129-132`) before the `HashSet` and the per-worker loop, so it needs no distinct keys — the test already relies on that, since the overflow tail is a clone of worker 0 and would trip the duplicate-address check if the loop ran. The full-range success assertion (`prepared.last().worker_id == WorkerId::MAX`) genuinely needs 65,536 distinct, correctly keyed entries, and what it proves is that `.zip(0..=WorkerId::MAX)` is inclusive and agrees with the `<= MAX + 1` bound. That's a real off-by-one guard, but not one that needs to run in every default `cargo test`.

Suggest splitting it — cheap overflow rejection in the default run, full-range success behind `#[ignore]` with a reason, the way `crates/types/src/forks.rs:1339` already does for an expensive case. Building the `P2pNode` directly also skips the two URL parses per worker that `worker_p2p` does for an `RpcInfo` this test immediately discards.

```rust
/// One more worker than the `WorkerId` range holds is rejected before any per-worker
/// validation runs, so cloned entries are enough: the count check fires ahead of the loop.
#[test]
fn prepare_worker_networks_rejects_worker_id_overflow() -> eyre::Result<()> {
    let keys = worker_key_config();
    let max_workers = usize::from(WorkerId::MAX) + 1;
    let worker = P2pNode { rpc: None, ..worker_p2p(&keys, 0)? };
    let workers = vec![worker; max_workers + 1];
    let streams = vec![(); max_workers + 1];
    let error = prepare_worker_networks(&workers, &streams, &keys, Epoch::MAX, max_workers + 1)
        .err()
        .ok_or_else(|| eyre!("expected WorkerId overflow"))?;
    assert!(error.to_string().contains("WorkerId range"));
    Ok(())
}

/// Every id in the `WorkerId` range is accepted and the last prepared worker carries
/// `WorkerId::MAX`, so the `0..=WorkerId::MAX` zip cannot silently truncate the layout.
///
/// Each entry must carry its own derived key and a distinct address to pass the per-worker
/// checks, and `prepare_worker_networks` re-derives every key, so this test performs
/// 2 x 65,536 BLS-backed derivations (tens of seconds under the debug test profile).
#[test]
#[ignore = "derives 2 x 65,536 worker network keys; run with `cargo nextest run --run-ignored all`"]
fn prepare_worker_networks_accepts_full_worker_id_range() -> eyre::Result<()> {
    let keys = worker_key_config();
    let max_workers = usize::from(WorkerId::MAX) + 1;
    let workers = (0..=WorkerId::MAX)
        .map(|worker_id| {
            Ok(P2pNode {
                network_address: format!(
                    "/ip4/127.0.{}.1/udp/{}/quic-v1",
                    worker_id / 256,
                    9000 + worker_id % 256
                )
                .parse()?,
                network_key: keys.worker_network_public_key(worker_id),
                rpc: None,
            })
        })
        .collect::<eyre::Result<Vec<_>>>()?;
    let streams = vec![(); max_workers];
    let prepared = prepare_worker_networks(&workers, &streams, &keys, Epoch::MAX, max_workers)?;
    assert_eq!(prepared.len(), max_workers);
    assert_eq!(prepared.last().map(|worker| worker.worker_id), Some(WorkerId::MAX));
    Ok(())
}
```

The alternative is a production-side `eyre::ensure!(prepared.len() == configured, ...)` after the collect, which turns a silent zip truncation into a startup error covered by every small-N test already here and lets the full-range test go away entirely. That's a production change for a regression the bound at 129-132 already makes unlikely, so I'd leave it to your judgment.

crates/test-utils-committee/src/builder.rs
```diff
@@ -163,14 +162,11 @@ where
...
+            let worker_nodes: Vec<P2pNode> = (0..=WorkerId::MAX)
+                .take(self.number_of_workers.get())
                 .map(|worker_id| {
-                    let key = if worker_id == 0 {
-                        key_config.worker_network_public_key(DEFAULT_WORKER_ID)
-                    } else {
-                        NetworkKeypair::generate_ed25519().public().into()
-                    };
+                    let key = key_config.worker_network_public_key(worker_id);
                     (worker_address(), key).into()
```
grantkee: With the builder's default randomize_ports == false, every advertised worker here gets the identical placeholder address /ip4/127.0.0.1/udp/0/quic-v1, which the duplicate-address check this PR adds to prepare_worker_networks rejects for any fixture of three or more workers.

The code producing the address is the `worker_address` closure at `builder.rs:157-164`, which this PR doesn't change — anchoring here because this is the loop that calls it once per worker. The closure takes no worker id and yields port `0` unless randomizing, so every call returns the same string. Under `randomize_ports == true` the ports are distinct, because `get_available_udp_port` only returns a port after claiming it against the process-global `USED_PORTS`; the placeholder path has no such guard.

Those entries reach `node_info` verbatim. `AuthorityFixture::generate` (`authority.rs:145-165`) takes worker 0 from `Config::default_for_test_with_genesis` → `NodeP2pInfo::default()`, which allocates a real claimed port, then chains `bootstrap.workers.iter().skip(1).cloned()` for the rest. Default layout for N workers is `[127.0.0.1:<rand>, 127.0.0.1:0, 127.0.0.1:0, …]`, so two workers pass and three fail the new `HashSet` insert at `node.rs:145-150`.

This is latent rather than broken today: no test feeds a `CommitteeFixture` config through `prepare_worker_networks`. `.number_of_workers(` has exactly one caller (`fixture_tests.rs:15`, which only compares keys), and the e2e harness builds its `TnBuilder` from on-disk keytool output with one worker per validator. It will bite the first time #557 adds a multi-worker `EpochManager` test on a default fixture, and the failure will look like a bug in the new check rather than in the fixture.

I don't think the production check should bend for it — `/udp/0` is also the advertised external address, so two of them in a real config is a genuine misconfiguration. The fixture is the right place to fix this. Passing the worker id into the closure and varying the loopback host in placeholder mode keeps everything unbound (no fixed port) and distinct for any N. `127.{id / 256}.{id % 256}.1` maps every `WorkerId` to a valid octet pair; a `1 + id % 256` scheme would hit an invalid `.256` at `id % 256 == 255`.

```rust
            let randomize_ports = self.randomize_ports;
            // Placeholder addresses (`randomize_ports == false`) must still be distinct per
            // worker: `prepare_worker_networks` rejects duplicate worker addresses, and a
            // three-worker fixture would otherwise advertise `/udp/0` twice. Vary the loopback
            // host rather than the port so nothing binds a fixed port. Randomized addresses stay
            // on 127.0.0.1 with allocator-claimed (already distinct) ports.
            let worker_address = move |worker_id: WorkerId| -> Multiaddr {
                let (host, port) = if randomize_ports {
                    (host.to_string(), get_available_udp_port(host).unwrap_or(DEFAULT_WORKER_PORT))
                } else {
                    (format!("127.{}.{}.1", worker_id / 256, worker_id % 256), 0)
                };
                format!("/ip4/{host}/udp/{port}/quic-v1").parse().unwrap()
            };
            // Advertise the same derived identity each worker swarm uses to authenticate.
            let worker_nodes: Vec<P2pNode> = (0..=WorkerId::MAX)
                .take(self.number_of_workers.get())
                .map(|worker_id| {
                    let key = key_config.worker_network_public_key(worker_id);
                    (worker_address(worker_id), key).into()
                })
                .collect();
```

The fixture regression in `fixture_tests.rs` could pin it with a `HashSet` over `worker_p2p_nodes()` asserting `len() == number_of_workers.get()` per authority. One collateral to be aware of: `127.0.1.1` and above aren't bound by default on macOS, so a test that sets `randomize_ports(false)` and then actually listens on worker ≥ 1 would fail there — but the non-randomized mode has never promised bindable addresses, and worker 0 (the only worker any current test listens on) keeps `127.0.0.1`.

grantkee (changes requested):
Looks good - I have a few requests to improve the debugging ux and future proof the validation effort here

sstanfield (dismissed):
LGTM but fill need to re-approve after you address Grant's feedback.

crates/node/src/manager/node.rs
MavenRain: Adopted: duplicate checks now strip trailing `/p2p/` identities before comparing listen addresses. Added adjacent and nonadjacent suffix-only regression cases, and errors retain the full configured address.

crates/node/src/manager/node.rs
MavenRain: Added `Display` using the same Base58 protobuf encoding as YAML and switched the startup error to it. The mismatch regression checks the expected/found values, and a YAML round-trip test verifies the displayed key can be pasted into configuration.

crates/node/src/manager/node.rs
MavenRain: Split this into a default overflow rejection test using one derived key and an explicitly ignored full-range success test with the reason documented. The full-range case now constructs `P2pNode` directly to avoid unused RPC parsing.

crates/test-utils-committee/src/builder.rs
MavenRain: Placeholder workers now use distinct `127.high.low.1` hosts derived from the worker ID, with UDP port 0. Randomized mode keeps allocator-claimed ports, and the fixture regression checks distinct addresses in both bootstrap entries and node configuration for every authority.

grantkee (approved):
LGTM! Thanks @MavenRain

PR #1344, opened 2026-09-08
fix(txpool): offload batch builds and canonical updates

### LINKED ISSUE #1339: tn-reth: batch-builder balance reads and canonical pool updates run synchronous work on async runtime threads
## Problem

`WorkerTxPool` does synchronous MDBX and pool work at two hot-path call sites, both reached from async tasks on the runtime that also drives consensus and networking. A third site doing the same class of work is already offloaded to the blocking pool, so the treatment is inconsistent and the two remaining sites are the larger ones.

### 1. Sender balance reads in `build_batch`

`TxPool::get_account_balances` (`crates/tn-reth/src/txn_pool.rs:589-616`) acquires one state provider and then reads each sender's account against it. Its only caller is `build_batch` (`crates/batch-builder/src/batch.rs:191`), which runs inside the `next-batch` task spawned at `crates/batch-builder/src/lib.rs:207`. `TaskSpawner::spawn_task` is `tokio::spawn` (`crates/types/src/task_manager.rs:189`), not `spawn_blocking`.

So every account read happens on an async worker thread, with no yield point between reads, on the path that gates batch sealing. The sender count is bounded only by the 30M-gas/1MB batch limits, which #1303 put at roughly 1,400 distinct senders. #1303 reduced the per-read setup for this loop from one `ConsistentProvider` per sender to one for the whole set, but the reads themselves stayed where they were.

### 2. `on_canonical_state_change` in `update_canonical_state`

```rust
// crates/tn-reth/src/txn_pool.rs:452-455
// TODO: should this be a spawned blocking task?
//
// update pool to remove mined transactions
self.0.on_canonical_state_change(update);
```

This is the only call site of `on_canonical_state_change`, and two async contexts reach it:

- The `BatchBuilder` main loop, after quorum (`crates/batch-builder/src/lib.rs:473`).
- The pool's canonical critical task, via `process_canon_state_update` (`crates/tn-reth/src/txn_pool.rs:496`), which carries its own note: "sync fn so self will block until all pool updates are complete."

#1303 characterized this call as O(pool size) under the pool write lock, running for every `Commit`.

### 3. The site that is already offloaded

`reload_dirty_accounts` runs under `spawn_blocking` (`crates/tn-reth/src/txn_pool.rs:269-272`) with an explicit rationale:

> The reload is synchronous MDBX I/O. Run it on the blocking pool so it never occupies one of the async worker threads this runtime also uses for consensus and networking (reth offloads the same work: `maintain_transaction_pool` runs under `spawn_blocking_task`).

That path is capped at `MAX_RELOAD_ACCOUNTS` (100) addresses per iteration and is only reachable after a canonical-state broadcast lag. Per #1304 the offload was added during #1243's review round, which is how the asymmetry arose: the capped recovery path was fixed and the uncapped every-round path was not revisited.

## Impact

Runtime-thread occupancy on the consensus runtime, not a correctness problem. The pool has no cross-address invariant, so nothing here can diverge between validators.

This report is not backed by a profile. The sender bound is the analytic 30M-gas figure from #1303, not a measurement, and no latency attributable to these two sites has been observed. What is concrete is the inconsistency: the smallest and least frequent of the three sites is the one that got the offload, and its stated rationale applies verbatim to the other two.

The repository has fixed this class of bug before, in #1076 (unbounded `std::fs::copy` plus sync directory removals on runtime worker threads) and #976 (inline BLS G2 decompress on the shared consensus runtime).

## Stale comments that hide this

Two doc comments claim the batch build already runs on a blocking thread:

- `crates/batch-builder/src/lib.rs:92`: "Type alias for the blocking task that locks the tx pool and builds the next batch."
- `crates/batch-builder/src/lib.rs:103-104`: "Single active future that executes consensus output on a blocking thread and then returns the result through a oneshot channel."

Neither is accurate. The task is `tokio::spawn`ed, and the second also misdescribes the work: the batch builder builds batches, it does not execute consensus output. These are what would lead the next reader to believe the offload is already handled, so they should be corrected whether or not the offload lands.

## Proposed solution

Decide the policy once for all three sites rather than per call site, then apply it:

- For the balance reads, wrap the read in `spawn_blocking` inside `get_account_balances`, or hoist the offload so the whole synchronous body of `build_batch` runs on the blocking pool. The second is the larger change but covers the transaction encoding in the same loop, which is also synchronous CPU work.
- For `on_canonical_state_change`, resolve the existing TODO. The two callers are already async, so the offload is local to `update_canonical_state`. Note the ordering constraint: the canonical critical task applies notifications in sequence, so an offload must not let two updates interleave.
- Fix the two stale comments regardless.

Worth confirming with a measurement before restructuring `build_batch`, since the cheap fix (offload just the balance read) may be enough and the sender bound has never been observed in practice.

## How this was found

Review of #1303 (batched balance reads). The batching fix addresses the per-read provider setup that issue targets, but the reads stay on the async worker thread, which is the other half of the "runtime-thread occupancy" impact #1303 describes. Tracing why one path was offloaded and this one was not surfaced the TODO and the stale comments.

MavenRain:
Closes #1339.

## Problem

Batch construction reads sender balances and encodes transactions synchronously on an async runtime thread. Canonical pool updates also run synchronous pool maintenance on that runtime, even though dirty-account recovery already uses the blocking pool. Normal batch traffic is sufficient to occupy threads shared with consensus and networking; no attacker is required.

## Fix

- **Batch construction:** await a blocking task for the complete synchronous `build_batch` call before proposing the batch. The builder retains one build in flight.
- **Canonical pool maintenance:** make `update_canonical_state` asynchronous, move owned inputs into a blocking task, and await completion in both the canonical listener and the batch builder. The listener processes notifications sequentially, and blocking-task failures propagate through the existing task error paths.
- **Documentation and regressions:** correct the two stale task descriptions, check batch selection and balance reads execute off the runtime thread, and verify canonical updates wait for blocking work before returning. Existing epoch base-fee coverage now awaits the update.
- **Pool-state assertions:** four batch-builder tests wait for mined transactions to leave the pool before checking the result. A single runtime yield or observing batch storage does not establish that blocking pool maintenance has completed. Existing five-second limits and transaction, pending, queued, and blob assertions are preserved.

## Testing

- Merge commit `a90e4063` includes main `0d3192a4`. The merge preserves the original five-file PR patch exactly, with no manual resolution edits. Independent integration review found no required semantic adaptations.
- Passed: `gateledger run -- cargo +nightly-2026-03-20 fmt --all --check` and `git diff --cached --check`.
- Passed in [hosted validation](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34302333756): `./etc/ci-lanes.sh clippy` (both configurations), `./etc/ci-lanes.sh test-default --build-only`, and `./etc/ci-lanes.sh test-adiri --build-only`.
- Passed: `NEXTEST_PROFILE=issue1344-strict ./etc/ci-lanes.sh test-default`, 1,719 passed and 32 skipped in 92.783 seconds. `test_pool_updates_after_txs_mined` passed in 0.123 seconds; `test_batch_builder_produces_valid_batches` passed in 0.110 seconds.
- Passed: `NEXTEST_PROFILE=issue1344-strict ./etc/ci-lanes.sh test-adiri`, 908 passed and 6 skipped in 47.557 seconds.
- No `SLOW` events occurred in the validation run.
- Previously passed: the isolated scheduling and pool-wait harnesses under Rust 1.94 and Tokio 1.50.0. The scheduling harness's three in-diff mutants compiled and failed the intended thread-placement and awaited-completion assertions. These harnesses use fake dependencies; the hosted suites exercise the repository tests.

Validation commit `d9f94961` has identical Rust sources to PR head `a90e4063`. Its only additional change is a dispatch-only workflow step running the full default and Adiri CI lanes with profile `issue1344-strict`. That profile preserves test groups, disables retries, and fails a test at 60 seconds with zero grace. The workflow change is outside this PR and restores its temporary configuration after the run.

Remaining maintainer requirements: archive/RocksDB guards, the ignored all-feature e2e suite, and on-chain attestation of the actual PR head. The normal maintainer PR compile/test lanes are skipped by repository policy. This PR remains draft.


sstanfield (dismissed):

sstanfield (dismissed):
Good catch.

crates/batch-builder/src/lib.rs
```diff
@@ -213,7 +213,8 @@ impl BatchBuilder {
             let (ack, rx) = oneshot::channel();
 
             // this is safe to call without a semaphore bc it's held as a single `Option`
-            let BatchBuilderOutput { batch, mined_transactions, changed_accounts, peer_deferred } = build_batch(build_args, worker_id, base_fee);
+            let BatchBuilderOutput { batch, mined_transactions, changed_accounts, peer_deferred } =
+                batch::spawn_batch_build(build_args, worker_id, base_fee).await?;
```
grantkee: A JoinError from the blocking batch build never reaches the run loop, so a panic inside build_batch shuts the node down as AckChannelClosed with the real cause unlogged.

When `spawn_batch_build(...).await` returns `Err(JoinError)` (the blocking build panicked, or was cancelled at runtime shutdown), the `?` here converts it straight into `TaskError` — this async block returns `Result<(), TaskError>`, so the blanket `impl<E: Error> From<E> for TaskError` (`crates/types/src/task_manager.rs:58-62`) applies — and returns from the `next-batch` task without sending anything on `result`. `next-batch` is spawned with the non-critical `spawn_task`, and the `TaskManager` drops a non-critical task's `Err` with `continue` and no log line, so nothing records the join error. The run loop then sees `RecvError` at `res??` (`lib.rs:456`), which `From<oneshot::error::RecvError>` maps to `AckChannelClosed` (`error.rs:49-53`), and the critical `batch builder` task exits with "Fatal error: failed to receive ack reply that new block was built" (`crates/node/src/engine/inner.rs:127-129`). The new `BlockingTask` variant is never constructed on this path — it is only reached via the run loop's `update_canonical_state(...).await?` at `lib.rs:499-504`. The panic payload survives only in Rust's default panic-hook line on stderr.

Not a regression: before this PR a panic in the inline `build_batch` unwound through the same `tokio::spawn`ed task, dropped the same sender, and reached the same `AckChannelClosed` exit. The PR put the `JoinError` in hand and didn't use it. The node shuts down either way; this only changes what gets reported, which matters because the only realistic panic sources here are the overflow checks under the `e2e` profile (`batch.rs:101` `total_possible_gas + pool_tx.gas_limit()`, `batch.rs:230` `max_nonce + 1`) and anything inside reth's `best_transactions()` / `encoded_2718()` — exactly the cases where you'd want the message.

The sibling `to_worker.send` failure path at `lib.rs:242-247` already does the right thing (send the specific error through `result`, then return); this could mirror it:

```suggestion
            let BatchBuilderOutput { batch, mined_transactions, changed_accounts, peer_deferred } =
                match batch::spawn_batch_build(build_args, worker_id, base_fee).await {
                    Ok(output) => output,
                    Err(e) => {
                        error!(target: "worker::batch_builder", ?e, "blocking batch build failed");
                        // Surface the join error to the run loop. A dropped sender is misreported
                        // as `AckChannelClosed`, and the task manager discards a non-critical
                        // task's `Err` without logging it.
                        let _ = result.send(Err(BatchBuilderError::BlockingTask(e)));
                        return Ok(());
                    }
                };
```

Returning `Ok(())` is fine since the manager discards the non-critical `Err` anyway. With this the run loop's `debug!(?res, "pending task complete")` at `lib.rs:454` shows `Err(BlockingTask(..))`, `run()` returns it, `inner.rs:129` logs it, and the manager prints the panic message instead of the ack-channel one. `BatchBuilderError` needs no new derives — `JoinError` isn't `Clone`, but the error is built once and moved.

The `BlockingTask` doc at `error.rs:44-46` currently describes only the pool-update path; it would then cover both:

```rust
    /// A blocking batch build or transaction-pool update failed (panicked or was cancelled).
    #[error("Blocking task failed: {0}")]
    BlockingTask(#[from] tokio::task::JoinError),
```

No test covers a panicking build today — `batch_build_offloads_pool_work` exercises the `Ok` path only. A `TxPool` double whose `best_transactions` panics, driven through `spawn_execution_task`, asserting the run loop resolves to `Err(BatchBuilderError::BlockingTask(_))` rather than `AckChannelClosed`, would pin this.

crates/batch-builder/src/batch.rs
```diff
@@ -38,6 +38,17 @@ pub struct BatchBuilderOutput {
...
 
+/// Build a batch on the blocking pool, including transaction encoding and sender balance reads.
+///
+/// The caller awaits this task before proposing the batch and keeps at most one build in flight.
+pub(crate) fn spawn_batch_build<P: TxPool + Send + 'static>(
+    args: BatchBuilderArgs<P>,
+    worker_id: WorkerId,
+    base_fee: u64,
+) -> tokio::task::JoinHandle<BatchBuilderOutput> {
+    tokio::task::spawn_blocking(move || build_batch(args, worker_id, base_fee))
```
grantkee, line 49: spawn_batch_build drops the propose-batch tracing span, so every log line build_batch emits now lacks worker_id, epoch and base_fee.

The `next-batch` task is instrumented with the `propose-batch` span (`lib.rs:207`, `.instrument(span_clone)` at `lib.rs:307`) carrying `worker_id`, `base_fee` and `epoch`. Before this PR `build_batch` ran synchronously inside that task, so its `debug!` lines and the `warn!` in `WorkerTxPool::get_account_balances` inherited the span. `tokio::task::spawn_blocking` does not carry the caller's current `tracing` span onto the blocking thread — `Spawner::spawn_blocking` wraps the closure via `blocking_task(...)`, which is a no-op under `cfg_not_trace!` (i.e. without `tokio_unstable`, which this repo doesn't set: `RUSTFLAGS` in `.github/workflows/pr.yaml` / `etc/ci-lanes.sh` is `-D warnings -D unused_extern_crates`), and even with it tokio would enter its own `runtime.spawn{kind="blocking"}` span, not re-enter `propose-batch`. `.instrument` only enters the span around polls of the async task on the runtime thread; the blocking thread's span stack is empty.

Lines that now emit without the span: `batch.rs:115` (`deferring tx already packed by a validated peer batch`), `batch.rs:118` (`marking tx invalid due to gas constraint`), `batch.rs:134` (`marking eip4844 tx invalid`), `batch.rs:138` (`marking non-allowlisted tx type invalid`), `batch.rs:164` (`marking tx invalid due to bytes constraint`), and the one that matters operationally, `crates/tn-reth/src/txn_pool.rs:731-736` `warn!(target: "txpool", ?error, num_addresses, "failed to acquire state provider; reporting zero balance for all senders")` reached from `batch.rs:225`. Anything reth logs inside `best_transactions()` / `remove_transactions_and_descendants` loses it too. With reth's default `Full` fmt layer that's the `propose-batch{worker_id=.. base_fee=.. epoch=..}:` prefix going missing, and with `--tracing-url` the build events detach from the `propose-batch` OTel span. On a multi-worker node `worker_id` is the only thing that tells these lines apart. Everything after the `.await` in the task (`span.record("batch", ..)` at `lib.rs:239` onward) still carries the span, since the async task re-enters it on each poll.

Capturing the span on the runtime thread and re-entering it in the closure fixes this without touching `build_batch`:

```suggestion
    // `spawn_blocking` does not carry the caller's tracing span onto the blocking thread;
    // re-enter `propose-batch` so the build's log lines keep worker_id/base_fee/epoch.
    let span = tracing::Span::current();
    tokio::task::spawn_blocking(move || span.in_scope(|| build_batch(args, worker_id, base_fee)))
```

`Span::current()` is evaluated while the `next-batch` task is being polled inside `propose-batch`; `Span` is `Send + Sync + 'static`, and outside any span it returns a disabled span so `in_scope` is a no-op. `update_canonical_state` in `txn_pool.rs` doesn't need the same treatment — neither of its callers runs inside an application span (`BatchBuilder::run` is spawned uninstrumented at `inner.rs:127`, and the maintenance task is uninstrumented). The engine takes the other approach for `execute_consensus_output` (creates and enters its own span inside the blocking closure); either is fine, this one keeps the span identity so the build's events stay children of the same `propose-batch` span the rest of the task logs under.

crates/tn-reth/src/txn_pool.rs
```diff
@@ -347,14 +348,20 @@ impl WorkerTxPool {
...
+                                self.apply_canon_notification(notification).await?;
+                                Ok::<_, JoinError>(AddressSet::default())
+                            }
+                            .left_future()
+                        })
+                        .unwrap_or_else(|BroadcastStreamRecvError::Lagged(missed)| {
+                            futures::future::ready(Ok(self.mark_drifted(missed))).right_future()
+                        })
+                        .await?
+                }
```
grantkee: The Update arm builds an Either future to express what a nested match already says, and it's the only reason for the new FutureExt import.

`update.map(|n| async { ... }.left_future()).unwrap_or_else(|Lagged(m)| ready(Ok(..)).right_future()).await?` needs the `FutureExt as _` import at line 32 and a turbofished `Ok::<_, JoinError>` to line up the two `Either` halves. The enclosing `match event` is already a `match`, and the behaviour is the same either way: `apply_canon_notification` returns `Result<(), JoinError>`, `maintain_pool` returns `Result<(), TaskError>`, and the `?` converts through the same blanket `From` (`task_manager.rs:58-62`) whether it sits after `.await` on the `Either` or directly inside an arm. The `Ok` branch is awaited before the next event is pulled from the stream in both forms, and `mark_drifted` already runs eagerly inside `unwrap_or_else`. `BroadcastStreamRecvError` is a single-variant, non-`#[non_exhaustive]` enum, which the existing irrefutable closure pattern at line 360 already relies on.

```suggestion
                MaintenanceEvent::Update(update) => match update {
                    Ok(notification) => {
                        self.apply_canon_notification(notification).await?;
                        AddressSet::default()
                    }
                    Err(BroadcastStreamRecvError::Lagged(missed)) => self.mark_drifted(missed),
                },
```

`FutureExt` is then unused — `.left_future()` / `.right_future()` are its only uses in the file (every other `.map`/`.then`/`.take_while`/`.next` here is `Option`/`Result`/`Iterator`/`StreamExt`) — and would trip `unused_imports` under `-D warnings`, so line 32 becomes:

```rust
use futures::{stream, Stream, StreamExt as _};
```

grantkee (dismissed):
LGTM - I have a few nits to support tracing and error logs. One nit about readability for your consideration

grantkee (approved):

PR #1345, opened 2026-09-09
fix(tn-reth): honor transaction pool limits and lifetime

### LINKED ISSUE #1340: tn-reth: four --txpool.* flags parse and are silently ignored — TN's hand-built pool validator and replacement maintenance task never receive them
## Summary

TN builds its transaction-pool validator by hand rather than running reth's `EthereumPoolBuilder`, and replaces reth's `maintain_transaction_pool` with its own task. Four `--txpool.*` flags therefore parse, validate, land in `NodeConfig`, and are then silently discarded. This is the same defect class as #1156 (`--rpc.*` eth flags) and #1160 (`--rpc.txfeecap`), one layer down.

Three of the four are "silently ignored, never wrong value used" — TN's effective values happen to equal reth's builder defaults, so a default-configured node behaves identically to reth and the defect only manifests when an operator explicitly sets a flag. The fourth (`--txpool.lifetime`) is different: its consumer does not exist in TN at all, so queued transactions are never evicted.

Found while reviewing #1312 (not introduced by it — `git diff 87364b23…HEAD -- crates/tn-reth/src/txn_pool.rs` is empty). Filing separately as agreed on that PR.

## Root cause

`crates/tn-reth/src/txn_pool.rs:162-190` builds the validator with:

```
.no_eip4844()
.no_eip7702()
.kzg_settings(EnvKzgSettings::Default)
.set_tx_fee_cap(node_config.rpc.rpc_tx_fee_cap)
.with_local_transactions_config(pool_config.local_transactions_config.clone())
.with_additional_tasks(node_config.txpool.additional_validation_tasks)
.build_with_tasks(task_spawner.clone(), blob_store.clone());
```

reth's `EthereumPoolBuilder` (`crates/ethereum/node/src/node.rs:489-499`) makes three additional calls TN omits:

```
.with_max_tx_input_bytes(ctx.config().txpool.max_tx_input_bytes)
.with_max_tx_gas_limit(ctx.config().txpool.max_tx_gas_limit)
.with_minimum_priority_fee(ctx.config().txpool.minimum_priority_fee)
```

Passing `pool_config` to `Pool::eth_pool` does not deliver them: two are not `PoolConfig` fields at all, and the third is a `PoolConfig` field that reth itself never reads.

| Flag | Why it is dropped | Verified |
|---|---|---|
| `--txpool.max-tx-input-bytes` | Validator-only; not a `PoolConfig` field. Declared `validate/eth.rs:89`, enforced `:352-372` | reth `node.rs:493` calls it, TN does not |
| `--txpool.max-tx-gas` | Validator-only; not a `PoolConfig` field. Declared `validate/eth.rs:91`, enforced `:397-408` | reth `node.rs:496` calls it, TN does not |
| `--txpool.minimum-priority-fee` | Is a `PoolConfig` field (`config.rs:57`, defaulted `:124`) but **dead in reth** — no reader anywhere in `transaction-pool/src/pool/`. The only reader is the validator field at `validate/eth.rs:445`, set by `.with_minimum_priority_fee()` | reth `node.rs:497` calls it, TN does not |
| `--txpool.lifetime` | `PoolConfig::max_queued_lifetime` (`config.rs:70`, defaulted `:130`) has exactly one consumer in reth: `crates/node/builder/src/components/pool.rs:289` (`max_tx_lifetime: pool_config.max_queued_lifetime`), inside the `spawn_maintenance_task` TN replaces with its own (`txn_pool.rs:118+`), which has no lifetime eviction | grep for consumers returns only that site |

A repo-wide grep of `crates/` for `max_tx_input_bytes`, `max_tx_gas_limit`, `minimum_priority_fee`, and `max_queued_lifetime` returns **zero hits each**.

## Evidence this is accident, not policy

Every deliberate divergence in this same builder chain carries a comment with an issue number — `.no_eip4844()` / `.no_eip7702()` at `txn_pool.rs:166-176` (#1159), `.set_tx_fee_cap()` at `:180-183` (#1160). The three omissions carry nothing. And `crates/tn-reth/src/env/rpc.rs:45-49` documents this exact bug class one layer up: *"each `--rpc.*` eth flag parses and then silently keeps the reth default (#1156)"*.

## Impact

**No new DoS exposure from the three validator flags.** Neither omission lifts an existing cap:

- the 128 KiB input cap is still enforced via the validator's own default (`validate/eth.rs:352-372`); the operator merely cannot *tighten* it;
- per-transaction gas is still bounded by the block-gas-limit check (`validate/eth.rs:386-393`), kept current through `on_canonical_state_change` (`txn_pool.rs:446`), and batch gas is separately capped at 30M (`crates/types/src/worker/sealed_batch.rs:197-199`).

So for these three the defect is weak diagnostics: an operator sets a flag, gets no error, and the node ignores it.

**`--txpool.lifetime` is a genuine unbounded-growth concern.** Because nothing evicts by age, a transaction parked in the basefee or queued sub-pool stays there indefinitely. Two consequences worth noting:

- an underpriced transaction parks and creates a nonce gap that parks that sender's later transactions too, and nothing ever clears it;
- pool memory grows monotonically with abandoned transactions.

This is also the real mechanism behind "an underpriced transaction parks indefinitely", which came up on #1312 while discussing quote headroom. No amount of RPC-side fee headroom addresses it — the eviction is what is missing.

## Proposed fix

In `crates/tn-reth/src/txn_pool.rs`, extend the builder chain before `.with_additional_tasks(...)`:

```rust
// Apply the operator's remaining `--txpool.*` validator knobs.
//
// These three live on the validator, not on `PoolConfig`, so passing `pool_config` to
// `Pool::eth_pool` below does not deliver them:
// - `max_tx_input_bytes` and `max_tx_gas_limit` are not `PoolConfig` fields at all.
// - `PoolConfig::minimum_priority_fee` exists but is dead in reth v1.11.3 -- declared at
//   `transaction-pool/src/config.rs:57`, defaulted at `:124`, read nowhere in `pool/`. The
//   only reader is `EthTransactionValidator` (`transaction-pool/src/validate/eth.rs:445`).
//
// Without these calls each flag parses, lands in `NodeConfig`, and then silently keeps
// reth's validator default. Mirrors `EthereumPoolBuilder::build_pool` (reth v1.11.3,
// `crates/ethereum/node/src/node.rs:489-499`), which TN's launch path never runs -- the
// same gap `apply_eth_config` closes for `--rpc.*` (#1156).
.with_max_tx_input_bytes(node_config.txpool.max_tx_input_bytes)
.with_max_tx_gas_limit(node_config.txpool.max_tx_gas_limit)
.with_minimum_priority_fee(node_config.txpool.minimum_priority_fee)
```

Because `max_tx_input_bytes` becomes live, add a guard rejecting a per-transaction cap above `max_batch_size(epoch)` (`crates/types/src/worker/sealed_batch.rs:203-205`) — such a transaction could never be batched (`crates/batch-builder/src/batch.rs:124`) and would occupy pool memory forever.

For `--txpool.lifetime`, either implement age eviction in TN's maintenance task (`txn_pool.rs:118+`) using `pool_config.max_queued_lifetime`, or reject the flag at startup as unsupported. Silently accepting it is the worst of the three options.

Extend `crates/telcoin-network-cli/tests/txpool_defaults.rs` (currently covers only `--txpool.max-account-slots`) to assert each parsed value reaches the built validator.

## Dependency worth recording

If this fix plumbs `minimum_priority_fee` into the validator, then `eth_maxPriorityFeePerGas` must follow it. #1312 makes that method answer a constant zero, and the validator check is:

```
// reth crates/transaction-pool/src/validate/eth.rs:441-445
if !is_local && transaction.is_dynamic_fee() &&
    transaction.max_priority_fee_per_gas() < self.minimum_priority_fee
```

Both sides are `Option<u128>`, and Rust's derived `Ord for Option<T>` places `None` below every `Some(_)`, so `Some(tip) < None` is false for every tip — the check is currently **inert**, which is why a zero-tip quote is safe today. Wiring the flag makes it live, and a node with `--txpool.minimum-priority-fee N` would then quote a tip its own pool rejects. Either wire both, or reject a non-`None` `minimum_priority_fee` at startup as incoherent with a chain that has no fee market.

## Flags that DO reach live readers

Recorded so this is not over-read: `--txpool.gas-limit`, `--txpool.max-account-slots`, `--txpool.pricebump`, `--txpool.minimal-protocol-fee`, `--txpool.nolocals`, and the sub-pool size limits all reach live consumers via `PoolConfig`.

## Related

- #1156 — the same class for `--rpc.*` eth flags (closed)
- #1160 — the same class for `--rpc.txfeecap` (closed)
- #1050 — `--txpool.max-account-slots` silently promoted (closed; different mechanism, same surface)
- #1159 — TN not running `maintain_transaction_pool` also caused the blob-sidecar leak (closed); this is the queued-lifetime half of the same structural gap
- #1312 — the PR whose review surfaced this

Reth reference: v1.11.3 (`d6324d6`).

MavenRain:
Closes #1340.

## Problem

TN's custom pool construction accepts operator settings that never reach their consumers. Byte and gas admission limits keep the validator defaults, and queued or basefee transactions never expire by age. No attacker is required: ordinary abandoned transactions can remain parked indefinitely.

## Root cause

The byte and gas caps belong to reth's transaction validator rather than its pool configuration. TN also replaces reth's maintenance loop without consuming `max_queued_lifetime`.

## Changes

- **Admission limits:** pass `--txpool.max-tx-input-bytes` and `--txpool.max-tx-gas` to the validator. Reject a byte cap above the shared `min_batch_size()` floor before starting validation tasks. Admission and batching both measure full EIP-2718 encodings for supported transaction types.
- **Fork dependency:** document the once-per-process pool guard beside `max_batch_size`. A fork that lowers the limit must update the shared floor and its boundary coverage.
- **Queued expiry:** run a critical timer task independently of canonical notifications. Expire both parked sub-pools, preserve pending transactions, and honor reth's local/private-origin exemptions unless `--txpool.nolocals` is set. A zero lifetime expires on the next sweep, with a minimum timer period of 1 ms. Unexpected task exit or panic uses the same shutdown policy as canonical pool maintenance.
- **Expiry metrics:** register `tn_reth.txpool_expired_txs_total` at zero and count actual removals. A flat counter alone does not prove a stalled task because unexpired or exempt transactions can remain parked.
- **Priority fee policy:** reject an explicit `--txpool.minimum-priority-fee`, including zero, with an actionable startup error. This follows the issue's supported rejection option and TN's zero-tip fee policy.
- **Regression coverage:** exercise parsed settings against the production pool constructor, exact byte/gas boundaries, unsupported fee policy, age boundaries, local exemptions, and an idle-chain timer. Cover the shared batch floor and metric registration/accumulation. Merge current `main` while retaining its pool recovery and peer-batch changes.

## Testing

- Passed: `gateledger run --ledger /Users/oobi/Documents/gpt14/.gateledger -- cargo +nightly-2026-03-20 fmt --all -- --check` and `git diff --cached --check`.
- Passed three checks in an isolated harness built from the actual batch-limit functions, metrics derive, recording helper, and their tests, using Rust 1.94 and the locked metrics dependency versions. Increasing the batch floor above the limit and replacing cumulative counting with an absolute update each caused the corresponding assertion to fail. Restoring the source reproduced the passing test binary.
- Completed `diffclass --base origin/main --staged`, `apidelta --base origin/main --verb clippy`, and `buildplan --base origin/main clippy --emit json`. After committing the merge, the planner scopes the six-file PR diff to 25 of 26 packages and retains the dependency closure. These are planning results, not a successful workspace build.
- Static review found no weakened CI or test gates. The conflict resolution preserves main's canonical maintenance and peer-batch logic.
- Full workspace tests, clippy, production-pool mutation checks, and `make attest` for the updated commit remain pending. The isolated checks do not replace these gates. Maintainer-authored PRs skip GitHub's compile/test lanes, so the attestation machine must provide that coverage.


sstanfield (dismissed):

crates/tn-reth/src/txn_pool.rs
```diff
@@ -210,6 +242,22 @@ impl WorkerTxPool {
...
     ) -> eyre::Result<Self> {
+        // The batch byte limit is currently identical in every epoch. Non-blob reth
+        // validation measures the full EIP-2718 encoding, just like the batch protocol.
+        let maximum = max_batch_size(0);
+        (node_config.txpool.max_tx_input_bytes <= maximum).then_some(()).ok_or(
+            TxPoolConfigError::InputLimitExceedsBatch {
+                configured: node_config.txpool.max_tx_input_bytes,
+                maximum,
+            },
+        )?;
```
grantkee: The startup guard evaluates the batch byte limit at epoch 0, so a fork that lowers that limit later would leave this check accepting values the batch builder can never pack.

`max_batch_size(_epoch)` (`crates/types/src/worker/sealed_batch.rs:203-205`) takes an epoch and is documented "can change in the future at a fork", but the only thing recording that this guard depends on it is the comment here at the call site. The asymmetry that makes it bite: the pool and its validator are built **once per process** — `start_epoch.rs:490-504` calls `initialize_worker_components` only when `initial_epoch || !engine.are_workers_initialized()`, `engine/inner.rs:155` is the sole production caller, and `max_tx_input_bytes` is fixed at `build_with_tasks` with no runtime setter. The batch builder (`inner.rs:114-124`) and batch validator (`start_epoch.rs:536-538`) are rebuilt every epoch. So if a fork at epoch N lowers the batch limit, an operator value between the new and old limits keeps passing this guard for the life of the process.

What happens to the transaction it admits is the part worth noting, because it is not "gets rejected at batch time and goes away". It lands in the **pending** sub-pool, and `crates/batch-builder/src/batch.rs:124-132` tests `total_bytes_size + encoded.len() > max_size` — which conflates "the batch is full" with "this can never fit" — then calls `best_txs.max_batch_size(...)` → `mark_invalid`. In reth that only inserts the sender into the iterator-local `invalid: HashSet<SenderId>` (`pool/best.rs:112-118`), and the iterator is rebuilt on every build (`pending.rs:108-117`). Because it keys on *sender*, the stuck transaction also blocks every later-nonce transaction from that account, on every build, forever. Compare the blob and unsupported-type paths, which actually remove from the pool after the loop (`batch.rs:160-164` → `remove_transactions_and_descendants`). And the new expiry sweep cannot help: it iterates `queued_transactions()`, which is basefee + queued only (reth `pool/txpool.rs:487-489`).

I checked whether the repo's fork conventions would catch this and they would not. `crates/types/src/forks.rs` holds every fork epoch constant with `const _: () = assert!(...)` cross-constraints and there is a startup fork-schedule log (`telcoin-network-cli/src/node.rs:129-150`), but `grep -i batch forks.rs` is empty, and `git log -S max_batch_size -- sealed_batch.rs` shows a single introducing commit, so there is no precedent for annotating a change here either.

Nothing is wrong today — units match (reth measures `encoded_length()`, the batch validator sums `tx.len()`), and it needs both a fork that lowers the limit and an operator who raised `--txpool.max-tx-input-bytes` above reth's 128 KB default into the gap. It is the silence that seems worth fixing: a named floor plus a note at the fork site would make the dependency visible to whoever changes that constant.

```rust
// crates/types/src/worker/sealed_batch.rs
/// Max batch size in effect at `epoch`. Measured in bytes.
/// Currently always 1,000,000 but can change in the future at a fork.
///
/// FORK NOTE: `tn_reth::WorkerTxPool::build` checks the operator's
/// `--txpool.max-tx-input-bytes` against [`min_batch_size`] ONCE at node start (the pool
/// and its validator are built once per process, not per epoch). A fork that LOWERS this
/// value must lower [`min_batch_size`] in the same change.
pub fn max_batch_size(_epoch: Epoch) -> usize { 1_000_000 }

/// The smallest batch byte limit across every epoch this build can serve. For one-shot
/// startup guards that cannot be re-evaluated per epoch.
pub fn min_batch_size() -> usize { max_batch_size(0) }

#[cfg(test)]
#[test]
fn min_batch_size_bounds_every_epoch() {
    // extend with each fork epoch constant from forks.rs as they are added
    for epoch in [0, 1, Epoch::MAX] { assert!(min_batch_size() <= max_batch_size(epoch)); }
}
```

Then `let maximum = min_batch_size();` here, with the import at line 62 and `byte_limit_cannot_exceed_batch_limit` (`config_tests.rs:141`) updated to match. `sealed_batch.rs` is not in this diff, so this is a two-file change — happy for it to be a follow-up instead if you would rather keep this PR to the plumbing.

The deeper fix is splitting "can never fit" from "batch full" in `batch.rs` and removing the former from the pool alongside the blob path, which would close this for any cause rather than just this guard. I have opened that as #1383 since it is pre-existing and outside this diff.

crates/tn-reth/src/txn_pool.rs
```diff
@@ -270,6 +321,47 @@ impl WorkerTxPool {
...
+        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+        task_spawner.spawn_task("queued txn pool expiry", async move {
+            IntervalStream::new(interval)
+                .for_each(move |_| {
+                    pool.evict_stale_transactions(Instant::now());
+                    futures::future::ready(())
+                })
+                .await;
+            Ok(())
+        });
```
grantkee: The expiry task is spawned non-critical, and the task manager logs nothing at any level when a non-critical task exits, so if this one ever stops the pool silently reverts to having no age eviction.

Non-critical is a defensible choice for a hygiene task, and the body has no realistic exit path — the interval stream is infinite, so the only ways out are manager shutdown (via `create_task`'s `select!` at `task_manager.rs:189-198`, giving `Ok(())`) or a panic inside `evict_stale_transactions`. What I had assumed, and what turns out not to hold, is that the task manager would say something either way. It does not. In `until_exit_internal` (`crates/types/src/task_manager.rs:457-486`) the `Ok(info)` arm does `if !info.critical { continue; }` at 463-466 with no log, and the `Err((info, join_err))` arm — which covers both a returned `TaskError` and a panic `JoinError`, per `TaskHandle::poll` at 108-121 — does `if !info.critical { continue; }` at 474-477, also with no log. `spawn_task`/`create_task` (163-203) log nothing on spawn or on completion. Critical tasks by contrast get an `info!` or `error!` plus a shutdown (462-483). So a panic here is completely silent, and the regression is back to the pre-PR behaviour this PR exists to fix.

Two things make that hard to notice from the outside. Reth's own `TxPoolMetrics.removed_transactions` is not incremented by `remove_transactions` (only by the mined prune at `txpool.rs:671` and `discard_worst` at `:977`), so the sweep has no direct counter — it shows up only indirectly in the parked-size gauges. And the sibling task in this same file takes the opposite choice: `spawn_maintenance_task` at line 389 uses `spawn_critical_task` and returns `Err(TaskError::from_message(...))` on unexpected end, which is also where reth puts its stale sweep (inside the critical maintenance loop, `node/builder/src/components/pool.rs:281-283`).

I don't think there is an obviously right answer on critical vs non-critical, which is why I am asking rather than suggesting: critical means a panic in a hygiene task shuts the validator down, but it is consistent with the sibling task that drives the same reth structures under the same lock, and with reth itself. If you want it critical:

```rust
    task_spawner.spawn_critical_task("queued txn pool expiry", async move {
        IntervalStream::new(interval)
            .for_each(move |_| {
                pool.evict_stale_transactions(Instant::now());
                futures::future::ready(())
            })
            .await;
        // unreachable in practice: the interval stream is infinite
        Err(TaskError::from_message("queued txn pool expiry task ended: interval stream closed"))
    });
```

Either way a sweep counter would make the task's liveness observable, and it is nearly free because `remove_transactions` already returns what it removed. Following the existing `RETH_METRICS` convention in this file (`metrics.rs:80`, already used at `txn_pool.rs:502-503` and `:561-563`):

```rust
/// Parked (queued/basefee) transactions discarded by the age-based expiry sweep.
///
/// Reth's `removed_transactions` counter does not count these (it covers mined prunes
/// and size-based discards), so this is the only direct signal that the sweep is running
/// and what it removes. A flat line across several `--txpool.lifetime` periods on a node
/// with a non-empty queued gauge means the sweep task is gone.
pub(crate) txpool_expired_txs_total: Counter,
```

```rust
// in evict_stale_transactions, replacing `self.0.remove_transactions(stale);`
let removed = self.0.remove_transactions(stale);
RETH_METRICS.txpool_expired_txs_total.increment(u64::try_from(removed.len()).unwrap_or(u64::MAX));
```

grantkee (dismissed):

crates/tn-reth/src/txn_pool.rs
MavenRain: Added `tn_types::min_batch_size()` and changed the startup guard and its boundary test to use it. The fork note now lives beside `max_batch_size`, where a reduction must also update the shared floor and extend the test with the fork boundary and adjacent epochs. The current schedule is covered at epochs 0, 1, and `Epoch::MAX`.

This keeps the once-per-process admission limit valid across the epochs supported by the build. The separate handling of transactions that can never fit remains tracked in #1383.

crates/tn-reth/src/txn_pool.rs
MavenRain: I chose the critical-task behavior to match canonical pool maintenance. Losing the configured age limit while the validator continues running should be visible through the task manager's error and shutdown path. The expiry task now uses `spawn_critical_task` and returns `TaskError` if its interval stream unexpectedly ends; normal shutdown still follows the task manager's existing shutdown handling.

Added `tn_reth.txpool_expired_txs_total`, incremented by the number actually returned from `remove_transactions`, so a concurrent canonical removal is not double-counted. The counter is registered at zero with the existing metrics initialization. Its docs distinguish removals from liveness: an empty pool, unexpired transactions, or exempt local origins can also keep the counter flat.

Added coverage for zero registration and cumulative updates, including an empty sweep. The new batch-floor test and metrics checks passed in an isolated harness using the actual source and pinned dependencies, and both code mutations were caught. Full workspace validation and attestation of the updated commit are still required.

grantkee (approved):

PR #1347, opened 2026-09-09
fix(node): add opt-in epoch export retention (#1346)

### LINKED ISSUE #1346: `--enable-state-export` never prunes completed `epoch-{N}` bundles: every boundary writes a full state copy and nothing bounds their count, bytes, or age, so an export-enabled node eventually fills its data volume
**Severity:** Low.  Operational disk exhaustion on the opt-in state-export path (`--enable-state-export`, default off).  No attacker lever: the epoch clock sets the bundle count and chain state sets the bundle size, and a remote party can accelerate neither.  The consequence lands on the exporting node only, but it lands on the same volume as its databases.

**Credit:** surfaced by Haxatron as finding #35 of the current Cantina collaborative review.  This issue records the finding against origin/main `0d3192a4` with the read path traced to repo evidence and a concrete retention design.

## Summary

With `--enable-state-export` (`crates/telcoin-network-cli/src/node.rs:63-66`), `export_epoch_state` (`crates/node/src/manager/node/close_epoch.rs:529`) writes one bundle per closed epoch to `consensus-db/state_exports/epoch-{N}/` (`:539-540`).  The bundle is published by an atomic rename from `epoch-{N}.tmp` (`:704`) and is never touched again.  Nothing in the tree limits how many completed bundles accumulate, how many bytes they hold, or how old they are.

Each bundle is a complete copy of the execution state, not a delta.  `PinnedStateView::export_state_pack` (`crates/tn-reth/src/snapshot.rs:172`) opens cursors over the full `PlainAccountState` and `PlainStorageState` tables (`:182-185`), walks both from `first()` to the end (`:191-209`), and resolves every referenced bytecode (`:216`).  The bundle also carries the closed epoch's consensus pack and a self-contained `0..=N` records/certs set (`close_epoch.rs:509-511`).

## Mechanism

The only directory removals on the export path target temporary directories:

- `sweep_stale_tmp_exports` (`close_epoch.rs:68`, called once at startup from `crates/node/src/manager/node.rs:633-634`) removes entries that match exactly `epoch-{N}.tmp` (`is_stale_tmp_export_name`, `:52-56`).  Completed `epoch-{N}` directories never match, and the test at `:1075-1101` asserts that they survive.
- The pre-export clear of this epoch's own `epoch-{N}.tmp` (`:610-626`), and `remove_tmp_export` (`:103-116`) on the failure and skip paths (`:700`, `:715`, `:722`, `:727`, `:731`).

No code path calls `read_dir` or `remove_dir_all` on a completed `epoch-{N}` directory.  The code anticipates the gap: the previous bundle "can be legitimately absent (first epoch, a skipped or failed prior export, operator pruning)" (`:594`).  Manual pruning is therefore supported, but it is never performed and never documented.  `SYNC.md:99-131` tells operators how to enable export and which directory to copy, and says nothing about growth or cleanup.  The flag's help text (`node.rs:63-64`) says nothing either.  There is no metric or log line for bundle count or bytes.

Older bundles have no in-tree consumer.  `db load-state` (`crates/telcoin-network-cli/src/db.rs:59`, args at `:171`) takes one `epoch-{N}` directory (`:206-208`).  The exporter reads only the immediately previous bundle, and only as an optimization: the incremental records/certs path (`close_epoch.rs:592-600`) falls back to a full rebuild when `epoch-{N-1}` is missing.

## Scale

The genesis default epoch length is 8 hours (`crates/telcoin-network-cli/src/genesis/mod.rs:93`), so an export-enabled node publishes 3 bundles per day and about 1,100 per year.  Growth per bundle is the full state size plus the epoch's consensus pack.  Illustrative figures for the state term alone:

| State size | Growth per day | 1 TB volume full after |
|---|---|---|
| 2 GB | 6 GB | about 5.5 months |
| 10 GB | 30 GB | about 5 weeks |
| 20 GB | 60 GB | about 17 days |

The records/certs set is a second-order term: kilobytes per epoch, but each bundle carries all of `0..=N`, so the total across bundles is quadratic in N.

The exports live under `consensus-db/`, on the same volume as the consensus database and, unless an operator mounts datadir subdirectories separately, the execution database.  When the volume fills, the exporter itself fails cleanly (the temp directory is removed, `:727`), but MDBX writes for consensus and execution fail with the node, logging fails, and a restart can fail until an operator frees space by hand.

## Suggested fix

Add a count-based retention setting beside the export flag and apply it after each successful publish.

1. `--state-export-keep <N>` on the `node` command (`node.rs`, beside `enable_state_export`).  Unset keeps every bundle (current behavior), so no operator loses archived bundles on upgrade.  Reject `N = 0`.
2. In the completion task, after the `rename` at `close_epoch.rs:704` succeeds, list `export_root`, keep entries that are directories and whose name is a canonical `epoch-{K}` (the same strictness as `is_stale_tmp_export_name`, `:52-56`), sort by `K`, and remove every entry beyond the newest `N` with `remove_dir_all` on the blocking pool (`spawn_blocking`, per #1076).  Skip `.tmp` directories and anything that does not match, as the sweep does.  Order by epoch number, not by task completion time, so a late-finishing export cannot delete a newer bundle.
3. Run the same pass once at startup, after `sweep_stale_tmp_exports`, so a lowered `N` takes effect at the next restart rather than at the next boundary.
4. Log the pruned epochs at `info` on `tn::snapshot`, and expose a gauge for the completed bundle count.
5. Document the per-epoch full-state cost and the setting in the flag's help text and in the "The export bundle" section of `SYNC.md`.

A byte budget (`--state-export-max-bytes`) is a reasonable alternative or addition, but a count is simpler to reason about and matches how operators think about snapshots.  Document `N >= 2` as the setting that preserves the incremental records/certs path;  `N = 1` is valid and falls back to the full rebuild at every boundary (`:592-596`).

## Acceptance criteria

- With `--state-export-keep N`, at most `N` completed `epoch-{K}` directories exist after any successful publish, and they are the `N` highest epochs present.  A unit test in the style of `sweep_removes_only_temp_export_dirs` (`close_epoch.rs:1075`) covers: newest kept, `.tmp` and stray entries untouched, `N` larger than the count is a no-op.
- The bundle just published is never removed.  A failed or skipped export removes nothing.
- Pruning never runs on the epoch-close critical path (`run_epoch.rs:472` awaits `export_epoch_state`);  it runs inside the completion task on the blocking pool.
- Without the flag, behavior is unchanged: every bundle is kept.
- `SYNC.md` and the flag help state the growth rate and the retention setting.

## Related, not duplicate

- #973 added the exporter and #985 the importer;  neither discussed retention.
- #1076 (completion-task filesystem work on the blocking pool), #1084 and PR #1099 (the `.tmp` sweep predicate), #1078 (records/certs rebuild cost), and #1091 (exporter memory) all touch this path.  None bounds completed bundles.

MavenRain:
Closes #1346.

## Problem

With state export enabled, every successful epoch export adds a full execution-state copy to the same data volume as the node databases. Completed bundles currently accumulate without a retention limit. No attacker is required for the volume to fill.

## Changes

- **Opt-in retention:** add `--state-export-keep N`, rejecting zero and preserving unlimited retention when unset. The setting is inactive unless state export is enabled.
- **Safe cleanup:** retain the newest numeric epochs at startup and after successful publication. Scan and recursive removal run on the blocking pool. Temporary directories, noncanonical names, regular files, and symlinks are excluded.
- **Overlapping completions:** serialize publication and pruning. Skip an older completion before publication when newer bundles already fill the retention window. Failed or skipped exports do not prune completed bundles, and the just-published bundle is retained.
- **Operator visibility:** log removed epochs, report the completed-bundle count through `tn_state_export_completed_bundles`, and document full-state storage costs, retention, and copying bundles before pruning.

## Testing

- [x] `cargo +nightly-2026-03-20 fmt --all -- --check`.
- [x] `cargo +1.94 test --locked -p tn-node --lib -j 2 -- retention_`: 5 passed.
- [x] `cargo +1.94 test --locked -p telcoin-network-cli --lib -j 2 -- node::tests`: 9 passed.
- [x] Mutation confirmation: all five retention tests fail under incorrect retention boundaries/missing-root handling; the CLI regression fails with an incorrect finite default. Restored the fix and reran the suites with test replay disabled.
- [x] Independent strict review: no remaining correctness blockers or CI weakening. `ctxcat-review` has no runnable host in this environment.

Scoped Clippy was run with the pinned nightly and the complete 14-crate scope emitted by buildplan:

```sh
cargo +nightly-2026-03-20 clippy --locked --keep-going -p e2e-tests -p state-sync -p telcoin-network -p telcoin-network-cli -p tn-batch-builder -p tn-batch-validator -p tn-engine -p tn-executor -p tn-network-libp2p -p tn-node -p tn-primary -p tn-storage -p tn-test-utils -p tn-worker --all-targets --no-deps -j 2 -- -D warnings
```

Clippy exits 101 on 131 diagnostics also recorded in the saved `main` baseline at `0d3192a`. All 131 match by message, source location, and lint name; no new diagnostics remain. This is a documented baseline failure, not a passing lint result.

The [PR CI run](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34365023138) skipped compile/test lanes under maintainer scope and failed `verify-on-chain` because `9cd72c1` is not attested. A maintainer must run `make attest` on that commit, then request a review to rerun the check.


crates/node/src/manager/node/export_retention.rs
```diff
@@ -0,0 +1,369 @@
...
+
+/// The state-export root, retention policy, and publication lock shared by all export tasks.
+#[derive(Clone, Debug)]
+pub(super) struct StateExportRetention {
+    /// Parent directory of canonical epoch directories and unfinished temporary exports.
+    root: PathBuf,
+    /// Number of newest completed exports to retain, or no limit.
+    keep: Option<NonZeroUsize>,
+    /// Serializes startup cleanup with publication and cleanup by overlapping exports.
+    gate: Arc<Mutex<()>>,
```
sstanfield, line 25: You should prefer parking lot for sync locks.  They are faster and you don't have to worry about lock poisoning.  Unless you know something about std mutex I am missing they are strictly better/faster.  Std may in fact use them under the hood now but even if so it retains the poison API that parking lot does not need.

sstanfield (dismissed):
I think this is fine and an extra layer of safety I guess.  To be clear a node that is exporting bundles is supposed to be managing them outside of the node.  The design was to keep bundles as simple as possible (in the node) and allow operators as much flexibility as possible dealing with them.  This does increase complexity of the feature but I can accept it since it could help a careless operator (although they may not even set this).  If the goal is to help careless operators then defaulting this to some value might make more sense.  But honestly I am also good with leaving cleanup with operators exporting bundles.  Also suggest using parking lot over std locks but you can ignore if don't agree (parking lot is in use in the project already- I always reach for it for sync locks).

crates/node/src/manager/node/export_retention.rs
MavenRain: Thanks, switched to `parking_lot::Mutex` and removed the poison-error handling.

Haxatron (approved):
LGTM, I side with Steve that defaulting this to some value (instead of no pruning by default) can be added as an improvement

sstanfield (approved):

PR #1359, opened 2026-09-09
fix(node): keep bootstrap servers out of epoch committees

### LINKED ISSUE #1349: Stop baking bootstrap servers into epoch committees
## Problem

`EpochManager` loads bootstrap servers once from the genesis committee and holds them in a field
(`crates/node/src/manager/node.rs:223`). It then copies that map into every non-genesis `Committee`
it builds from chain state, and both per-epoch network setup paths read them straight back out of
the committee they were just stamped onto.

The round trip:

- copy in: `crates/node/src/manager/node/start_epoch.rs:381-387`
- primary reads back out: `start_epoch.rs:615-620`
- worker reads back out: `start_epoch.rs:826-834`

Bootstrap servers are dial hints for peer discovery. They say nothing about who sits in a committee,
and routing them through `Committee` puts purely local operator configuration inside a
consensus-facing type.

Registration is already once-per-process. `init_network_for_epoch` calls `add_bootstrap_peers` only
under `initial_epoch` (`start_epoch.rs:988-990`), so from epoch 1 onward the extract-and-pass path
is dead weight: the peers extracted at epoch 5 are the ones registered at epoch 0.

## Why this cannot fork the chain

Worth settling first, because "change what goes into a `Committee`" reads like a consensus change.
It is not. Nothing commits to bootstrap servers.

`CommitteeInner` has a hand-written `PartialEq` at `crates/types/src/committee.rs:381-389` comparing
epoch, both thresholds, worker count and authorities. `bootstrap_servers` is deliberately excluded,
so two committees differing only in bootstrap servers are already equal today.

Peer pack import checks the committee two ways and neither touches bootstrap servers.
`verify_epoch_meta` (`crates/storage/src/consensus_pack.rs:1510-1568`) compares
`epoch_meta.committee.epoch()` against the record's epoch, and `epoch_meta.committee.bls_keys()`
against the previous record's `next_committee`.

`EpochRecord` carries `Vec<BlsPublicKey>` for both `committee` and `next_committee`
(`crates/types/src/primary/epoch.rs:21-39`), not a `Committee`, so bootstrap servers never reach the
record that gets signed and certified.

There is no committee digest, no `IntentScope` that commits to one
(`crates/types/src/crypto/intent.rs:59-67`), and no committee table in the consensus DB. The genesis
committee file on disk is the only durable copy, and this change does not touch it.

## Proposed change

Delete the loop in `create_committee_from_state` at `start_epoch.rs:381-387`. Non-genesis committees
stop carrying bootstrap servers. The genesis branch keeps reading them from disk, since that file is
where they live.

Repoint both readers at `self.bootstrap_servers` (`node.rs:223`) instead of
`consensus_config.committee().bootstrap_servers()`. The primary reader wants the `primary` field of
each `BootstrapServer` (`crates/types/src/committee.rs:216-221`). The worker reader keeps its
`filter_map` over `v.worker(*worker_id)` unchanged, so a peer running fewer workers still drops out
for higher worker ids.

Then drop `CommitteeBuilder::add_bootstrap_server` (`crates/types/src/committee.rs:1082-1092`).
`start_epoch.rs:382` is its only non-test caller.

While in `new()`, collapse the duplicate genesis-committee read. The file is loaded at
`node.rs:666-675` to build the consensus chain (`node.rs:678`), then loaded again at
`node.rs:694-703` only to pull bootstrap servers off it. One read serves both.

## Risk

Low. The payload handed to `add_bootstrap_peers` stays byte-identical: it is the same
genesis-derived map either way, read from the manager instead of laundered through a committee.

The case to check is epoch 0. There the committee is loaded from disk with its bootstrap servers
intact, so both readers must produce the same map whether they read the committee or the manager
field. They do, because `node.rs:694-703` seeds that field from the same file.

## Acceptance criteria

- `create_committee_from_state` no longer writes bootstrap servers into non-genesis committees.
- Both per-epoch readers source from `EpochManager::bootstrap_servers`.
- `CommitteeBuilder::add_bootstrap_server` is gone, along with its test callers.
- `new()` reads the genesis committee once.
- A node with no configured bootstrap peers registers the same peer set it does today, at epoch 0
  and at a later epoch.

## Related

A follow-up issue makes the manager's bootstrap servers operator-configurable through
`NetworkConfig`. That change is only clean once the committee is out of this path, so it depends on
this one.

MavenRain:
Closes #1349.

## Problem

The epoch manager copies genesis bootstrap servers into every later committee, then reads them back during primary and worker network setup. These process-wide discovery hints already live in the manager.

## Changes

- [x] **Load genesis once in `EpochManager::new`** and retain its bootstrap servers before moving the committee into the consensus chain.
- [x] **Read primary and worker bootstrap peers from the manager**, preserving filtering for peers without the requested worker ID.
- [x] **Build later committees from on-chain validators** and remove the unused `CommitteeBuilder::add_bootstrap_server` method.

Peer authorization still comes from the previous/current/next committee sets. Bootstrap registration remains once per process, before the initial committee update, including when restarting after epoch 0. Genesis committee files and committee serialization retain their existing formats.

## Testing

- Passed: `cargo +nightly-2026-03-20 fmt --all -- --check` through `gateledger`.
- Passed: `git diff --check` and a strict diff review, including all remaining bootstrap readers and the removed API's callers.
- The `diffclass`, `apidelta`, and `buildplan` ladder retains the full 25-crate dependency closure for Clippy.
- Not run: the planned Clippy command exhausted its 300-second `tnquiet` wait behind another active build (exit 75, before Cargo started).
- Pending: workspace check, Clippy, nextest, and the maintainer's pinned Rust 1.94 attestation suite. GitHub skips the compile and test lanes for maintainer-authored PRs.
- Confirmed: [the PR workflow](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34409413557) skipped its compile/test lanes and failed `verify-on-chain` because commit `50e05dd4cbaba3615ded362080847fe8d66626a8` has no attestation. A maintainer must run `make attest` on that commit, then request a review to rerun the check.

### DIFF
--- a/crates/node/src/manager/node.rs
+++ b/crates/node/src/manager/node.rs
@@ -663,16 +663,13 @@ where
     ) -> eyre::Result<Self> {
         // Note this can only fail if the consensus DB is very broken (bad path for instance).
         // So we will panic for now, this will kill the node on startup for a critical error.
-        let committee_zero = if let Ok(committee_zero) =
+        let committee_zero =
             Config::load_from_path::<Committee>(tn_datadir.committee_path(), ConfigFmt::YAML)
-        {
-            committee_zero
-        } else {
-            error!(target: "epoch-manager", "Unable to load committee zero from the genesis committee!");
-            return Err(eyre::eyre!(
-                "unable to load committee zero (genesis committee), this is fatal"
-            ));
-        };
+                .map_err(|_| {
+                    error!(target: "epoch-manager", "Unable to load committee zero from the genesis committee!");
+                    eyre::eyre!("unable to load committee zero (genesis committee), this is fatal")
+                })?;
+        let bootstrap_servers = committee_zero.bootstrap_servers();
         let epochs_db_path = tn_datadir.epochs_db_path();
         let _ = std::fs::create_dir_all(&epochs_db_path);
         let consensus_chain = ConsensusChain::new(epochs_db_path, committee_zero)?;
@@ -691,17 +688,6 @@ where
         let worker_event_streams = (0..builder.tn_config.node_info.p2p_info.num_workers())
             .map(|_| QueChannel::new())
             .collect();
-        let bootstrap_servers = if let Ok(committee_zero) =
-            Config::load_from_path_or_default::<Committee>(
-                tn_datadir.committee_path(),
-                ConfigFmt::YAML,
-            ) {
-            committee_zero.bootstrap_servers()
-        } else {
-            error!(target: "epoch-manager", "Unable to load bootstrap servers from the genesis committee!");
-            BTreeMap::new()
-        };
-
         // Spawn the state exporter once, only when the feature is enabled.
         let exec_state_exporter =
             builder.enable_state_export.then(ExecStateExporter::spawn).transpose()?;
--- a/crates/node/src/manager/node/start_epoch.rs
+++ b/crates/node/src/manager/node/start_epoch.rs
@@ -339,7 +339,8 @@ where
     ///
     /// Epoch 0 has no on-chain history, so the genesis committee is loaded from the
     /// committee file on disk. Every later epoch is built with a [`CommitteeBuilder`] from the
-    /// statically configured bootstrap servers plus the on-chain validator set for that epoch.
+    /// on-chain validator set for that epoch. Bootstrap dial hints stay in the manager's
+    /// process-lifetime configuration instead of being copied into later committees.
     ///
     /// In both cases the committee's worker count comes from the on-chain `WorkerConfigs` state
     /// at the previous epoch's closing block (`epoch_first_block - 1`; genesis state for epoch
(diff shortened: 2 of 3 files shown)


sstanfield (approved):
Approving but you need to run down EpochMeta records in the consensus pack.  They contain a Committee object that now won't have bootstrap servers.  I am not seeing an obvious issue at first glance but this may introduce some validation issues to pack files.  I think the likely pain point would be ConsensusPack::open_append() but from how it is used it may be ok.  Probably worth a specific scan though to confirm there is no where where an old pack file meta with bootstrap servers can trip up a node after this.

MavenRain:
Thanks Steve. The source scan found no bootstrap-related compatibility issue. `open_append()` compares `EpochMeta` using `Committee` equality, which [already excludes `bootstrap_servers`](https://github.com/Telcoin-Association/telcoin-network/blob/50e05dd4cbaba3615ded362080847fe8d66626a8/crates/types/src/committee.rs#L382-L390). An old committee with bootstrap entries therefore matches a newly built committee without them when the consensus fields are the same.

Restart/static opens use the stored committee; import and `db validate` check epoch linkage and validator BLS keys. The bootstrap field's serialization is unchanged, so existing pack metadata remains readable. No migration or code change appears necessary for this concern.

This was a source review; runtime upgrade tests were not run.

PR #1360, opened 2026-09-09
refactor(node): hoist network bring-up before the epoch loop

### LINKED ISSUE #1351: Hoist one-time network bring-up out of the epoch loop
## Problem

Four network actions are one-time per process, but they live inside the epoch loop and are gated by
a boolean threaded through six function signatures.

The actions:

- bootstrap peer registration, `crates/node/src/manager/node/start_epoch.rs:988-990`
- primary listener bind, `start_epoch.rs:653-661`
- worker listener bind, `start_epoch.rs:848-864`
- worker component initialization, `start_epoch.rs:490-504`

The boolean, `initial_epoch`, appears on `start_epoch.rs:205`, `:409`, `:458`, `:599`, `:808` and
`:986`. It originates in `run_epoch.rs` as `network_first_init` (`run_epoch.rs:359`), computed from
`RunEpochMode::initial_epoch()` OR the `network_initialized` field (`node.rs:193`, set false at
`node.rs:726`).

The comment at `run_epoch.rs:350-358` explains why the flag cannot simply key off the initial epoch:
a restart that replays and closes an epoch boundary can return early before `create_consensus` runs,
so the first real setup happens on a following `NewEpoch` iteration instead. That comment exists to
describe a hazard the current structure creates.

The hazard disappears if the setup is unconditional and happens before the loop. There is no
iteration to be skipped if there is no iteration.

## Proposed change

Move bootstrap registration, both listener binds and the bootstrap dials into `run()`, next to
`spawn_node_networks` (`node.rs:845`). The worker path loops over the vector of worker handles
rather than a single one.

Then delete:

- the `initial_epoch` parameter from all six signatures listed above
- the `network_initialized` field (`node.rs:193`) and its initializer (`node.rs:726`)
- the `initial_epoch()` accessor on the epoch mode enum, if nothing else uses it
- the `if initial_epoch` branches, leaving `init_network_for_epoch` reduced to its
  `update_committees` call

Worker component initialization at `start_epoch.rs:490-504` needs its own decision. Its condition is
already `initial_epoch || !engine.are_workers_initialized().await`, so the second disjunct alone
preserves the behaviour once the first is gone. Keeping the `are_workers_initialized` guard is the
conservative move and matches what the guard was added for.

Committee slots stay per-epoch. They are set every epoch from authoritative state via
`update_committees`, and the doc comment at `node.rs:191-192` already says they are not gated on
this flag.

## Risk

Medium, and this is the largest of the startup issues. The restart-replay path described at
`run_epoch.rs:350-358` is the one to exercise: a node killed at an epoch boundary, restarted, and
allowed to replay-and-close before entering a fresh epoch. Under the current code that path relies
on the flag; under the change it should not care, because setup already happened.

Two ordering questions to settle during implementation. First, whether anything the moved code needs
from `ConsensusConfig` is available at `node.rs:845`, since the per-epoch config is built later.
Listener addresses come from node info, so they should be, but the primary listener currently reads
`consensus_config.primary_networkkey()` and `consensus_config.primary_address()`
(`start_epoch.rs:655-658`). Second, whether the peer wait (`start_epoch.rs:1037-1057`) moves with
them or stays per-epoch.

## Acceptance criteria

- Bootstrap registration, both listener binds and the bootstrap dials run once in `run()`.
- `initial_epoch` is gone from all six signatures.
- `network_initialized` is gone.
- The comment at `run_epoch.rs:350-358` is deleted, because the hazard it describes no longer
  exists.
- The restart-and-replay-boundary end-to-end tests pass unchanged.

## Related

A startup epoch-record sync gate depends on this. The gate belongs in the pre-loop startup sequence,
and putting it there is only coherent once network bring-up is there too.

MavenRain:
Closes #1351.

## Problem

Process-lifetime network setup lived inside the epoch loop behind `initial_epoch` and `network_initialized`. A restart that replays and closes an epoch can return before creating consensus, so network startup depended on carrying that state into the following iteration. No attacker is required to reach this path.

## Changes

- **Start networks before the epoch loop.** In `run()`, register bootstrap peers, bind the primary and every configured worker listener, and schedule bootstrap dials on the process task spawner. Addresses come directly from node info; the worker listener override still applies only to worker 0. Worker bootstrap registration and dialing include only peers advertising that worker id.
- **Keep committee state per epoch.** Bootstrap registration completes before committee resolution. Each epoch still updates authoritative previous/current/next committee slots, refreshes gossip publisher allowlists, dials committee peers, and waits for peer readiness. Boundary-topic authorization retains its committee window during rotation.
- **Remove startup flags.** Delete `network_initialized`, `RunEpochMode::initial_epoch()`, and all six `initial_epoch` parameters. Keep `are_workers_initialized()` as the worker component guard, including after replay closes an epoch before creating workers.

## Testing

- Passed: `cargo +nightly-2026-03-20 fmt --all -- --check`, run through `gateledger`.
- Passed: `git diff --check`; strict diff review, including startup task scheduling, committee ordering, listener address equivalence, worker-id selection, and CI weakening checks. The installed `ctxcat-review` workflow runner is unavailable in this session.
- Planning: `diffclass --base origin/main` requires a build; `apidelta --base origin/main --verb clippy` retains the dependency closure; `buildplan clippy --base origin/main --emit sh` supplies the package scope below.
- Pending: the scoped Clippy command below is queued behind an existing session's build. Workspace check/nextest and the unchanged restart/epoch-boundary end-to-end tests have not run. The existing GitHub workflow skips dynamic jobs for this maintainer account, so opening this draft does not provide those checks.

```sh
tnquiet -- gateledger run -- env CARGO_TARGET_DIR=/Users/oobi/Documents/.telcoin-shared-target cargo +nightly-2026-03-20 clippy --locked -p e2e-tests -p state-sync -p telcoin-network -p telcoin-network-cli -p tn-batch-builder -p tn-batch-validator -p tn-engine -p tn-executor -p tn-network-libp2p -p tn-node -p tn-primary -p tn-storage -p tn-test-utils -p tn-worker --all-targets --all-features --no-deps -- -D warnings
```

GitHub confirmed that the Rust jobs were skipped. `verify-on-chain` failed because commit `8e68107629e6aa8e8b73dfbc7761c86fd9173add` has no attestation. The pinned Rust 1.94 build, restart/boundary tests, and `make attest` for this commit are still required before this draft is ready for merge; request review after attestation to rerun that gate.


grantkee (approved):
LGTM - we may want to harden this further to actually wait for connections or dial preexisting kad peers, but we can assess this in a follow up once more of these refactors land

sstanfield (approved):
I think this is good.  Not sure how this code wound up in run epoch but can't see a reason it needs to stay there now.

PR #1361, opened 2026-09-09
fix(network): wait for established peers during epoch startup

### LINKED ISSUE #1352: Distinguish established peers from dialing peers in the peer-count API
## Problem

`NetworkHandle::connected_peer_count` (`crates/network-libp2p/src/types.rs:686`) sends
`NetworkCommand::ConnectedPeerIds`, which the swarm task answers from
`peer_manager.connected_or_dialing_peers()` (`crates/network-libp2p/src/consensus.rs:925`). A peer
that has only been dialed, with no connection established yet, counts.

Outbound "send to anybody" requests do not use that set. `NetworkCommand::SendRequestAny`
(`consensus.rs:986-999`) picks from `self.connected_peers`, a `VecDeque<PeerId>`
(`consensus.rs:294`) that is only ever pushed on `ConnectionEstablished` (`consensus.rs:1736`) and
retained-out on disconnect (`consensus.rs:1662`, `:1705`).

So the two views disagree during the exact window that matters. A readiness wait keyed on
`connected_peer_count` can return satisfied while `connected_peers` is still empty, and every
`SendRequestAny` in that window returns `NetworkError::NoPeers`.

Current callers of the count:

- `dial_peer_bls`'s give-up check, `crates/node/src/manager/node/start_epoch.rs:773`
- `wait_for_network_peers`, `start_epoch.rs:1037-1057`, polling at `:1041` and `:1054`

`wait_for_network_peers` exists to block epoch startup until the network can actually talk to
somebody. Right now it can unblock on a set of half-open dials.

## Proposed change

Separate the two concepts at the API boundary rather than silently changing what the existing name
means.

Add a command and handle method that report established peers only, answered from the same source
`SendRequestAny` selects from. Point `wait_for_network_peers` at it, since "can I send a request
yet" is the question it is actually asking.

Leave `dial_peer_bls`'s give-up check (`start_epoch.rs:773`) on the dialing-inclusive count, or
consider it separately. Its question is different: it is asking whether this node has any other
prospect at all before abandoning one unreachable peer, and counting in-flight dials there is
defensible.

Whatever the naming choice, the doc comment on each method should say which of the two sets it
reports, because the current name gives no hint.

## Risk

Low in isolation. The change makes an existing wait stricter, so a node that used to proceed on
half-open dials now waits for a real connection. That is the intended behaviour, but it does mean a
network that dials successfully and never completes a handshake now blocks where it previously
proceeded and failed later with `NoPeers`.

`wait_for_network_peers` already bounds itself at 240 attempts at 500ms (`start_epoch.rs:1043-1049`)
and returns an error rather than hanging, so the failure mode stays bounded.

## Acceptance criteria

- A handle method reports only peers present in `self.connected_peers`.
- `wait_for_network_peers` uses it.
- Both methods document which set they report.
- A node whose only peers are mid-dial does not pass the readiness wait.

## Related

A startup epoch-record sync gate depends on this. That gate short-circuits when no peers are
connected, and short-circuiting on a dialing-only count would make it exit immediately on every cold
boot.

MavenRain:
Closes #1352.

## Problem

Epoch startup can pass its peer-readiness check while every peer is still dialing. No attacker is required: the existing count includes pending dials, but `SendRequestAny` selects from the established-peer queue and returns `NoPeers` while that queue is empty.

## Changes

- [x] **Established-only API:** add `EstablishedPeerCount` and `NetworkHandle::established_peer_count()`, answered directly from the queue used by `SendRequestAny`.
- [x] **Startup readiness:** use the new count for the initial probe and every retry on both primary and worker networks. Preserve the existing retry limit and dialing-inclusive count used by `dial_peer_bls`.
- [x] **Document both counts:** explain pending-dial inclusion and the established count's snapshot semantics.
- [x] **Regression coverage:** add a deterministic dialing/connection/disconnect test through the handle API and a paused-clock test for the readiness loop's initial probe and retry.

## Testing

- Passed: `gateledger run -- cargo +nightly-2026-03-20 fmt --all -- --check` and `git diff --check`.
- Static scope analysis: `diffclass --base origin/main` requires compilation; `apidelta --base origin/main --verb clippy` retains the dependency closure; `buildplan clippy --base origin/main` selects 21 crates. Manual diff review found no CI weakening or confirmed correctness defect. `ctxcat-review` is unavailable on this machine.
- Pending: scoped Clippy was stopped during dependency compilation to avoid competing with existing Telcoin validation jobs. The focused command below was stopped while waiting for the machine-wide build lock, before compilation or test execution:

  ```sh
  tnquiet -- testledger -- cargo +1.94 test --offline --locked \
    -p tn-network-libp2p -p tn-node --lib established_peer -j 2
  ```

- [ ] Compile and execute both regressions, then confirm failure when the established count uses the dialing-inclusive set and when either readiness probe uses the old count.
- [ ] Complete workspace check, Clippy, nextest, and the required attestation on this commit.

This PR remains a draft because dynamic validation is incomplete. The current workflow skips build/test lanes for drafts and maintainer-authored PRs; a workflow trigger alone is not evidence that those checks passed.


sstanfield (approved):

grantkee (approved):
LGTM

PR #1362, opened 2026-09-09
fix(node): seed startup mode from epoch records

### LINKED ISSUE #1350: Seed the node-mode watch from the epoch record chain instead of defaulting to CvvActive
## Problem

The node-mode watch channel is seeded from `NodeMode::default()`
(`crates/consensus/primary/src/consensus_bus.rs:379`), and the derived `Default` on the enum picks
`CvvActive` (`consensus_bus.rs:179-194`). Every node therefore claims to be an active committee
validator from process start until the epoch loop first computes its real role in
`identify_node_mode` (`crates/node/src/manager/node/start_epoch.rs:213`, published at `:966`).

For a node that is genuinely an observer, that window is a lie about its own state. Today nothing
behavioural reads the watch inside it, because every behavioural reader (the proposer, consensus
state, the primary network handler) is constructed at or after `identify_node_mode`. That is an
ordering accident, not an enforced invariant.

Two surfaces do read the default in that window and report it outward:

- the metrics gauge mirror spawned at `crates/node/src/manager/node.rs:1005`, which primes gauges
  from the current value at `crates/consensus/primary/src/metrics.rs:100`
- the node-mode RPC accessor at `crates/node/src/lib.rs:134-136`

The metrics mirror self-heals on the next watch update. The RPC surface answers whatever it is asked
during the window.

## Proposed change

Seed the watch once at startup from the epoch record chain, before the epoch loop runs.

Read the latest stored epoch record via `EpochRecordDb::latest_record`
(`crates/storage/src/epoch_records.rs:802`) and look at its `next_committee` field
(`crates/types/src/primary/epoch.rs:29`). `next_committee` is the right field, not `committee`: a
record is written when its epoch closes, so the latest record names the committee serving now.

If the node's own BLS key is in that set, seed `CvvActive`. Otherwise seed `Observer`. With no
stored record, leave the watch alone and let the epoch loop decide as it does today.

Place the seed next to the startup re-vote hook at `node.rs:880`.

## Constraints

Never write `CvvInactive` from this path. `identify_node_mode` treats an existing `CvvInactive` as
sticky and returns it unchanged (`start_epoch.rs:936-939`), so seeding it would pin a node into the
catching-up state before anything has established that it is behind. `CvvInactive` stays reserved
for the epoch loop's demotion logic.

Run the seed exactly once, outside the loop. Re-running it per epoch would reset a node that the
loop had already demoted, undoing the demotion every boundary.

## Risk

Low, and strictly an improvement on the current default. The worst case is that a node seeds
`Observer` and `identify_node_mode` then promotes it to `CvvActive` a moment later, which is the
same transition the metrics mirror already handles.

The one behaviour to preserve: a node with no epoch records at all (cold genesis) must end up
exactly where it does today. Leaving the watch untouched in that case gives that for free.

## Acceptance criteria

- The watch is seeded from `latest_record().next_committee` before the epoch loop starts.
- The seed never writes `CvvInactive`.
- A node with no stored epoch record behaves as it does today.
- The node-mode RPC on a genuine observer reports `Observer` from startup, not `CvvActive`.

## Related

A follow-up issue deletes the construction-time `--observer` mode override in `node.rs:686-689`.
That removal opens exactly the window described above for a genuine observer, so it depends on this
issue landing first.

MavenRain:
Closes #1350.

## Problem

The node-mode watch initially reports `CvvActive`, even when the stored epoch record shows that the node is outside the current committee. An ordinary observer restart can therefore expose the wrong startup role through the watch before the epoch loop identifies it. No attacker is required.

## Changes

- [x] **Seed from durable membership:** read `latest_record()` once and check the node's BLS key against `next_committee`, since the record describes a closed epoch. Publish `CvvActive` for a member and `Observer` otherwise.
- [x] **Preserve epoch demotion:** run the seed beside the startup re-vote hook, before metrics and the epoch loop. This path never writes `CvvInactive` and cannot reset a demotion at a later epoch boundary.
- [x] **Preserve cold genesis:** leave both the value and notification state of the watch untouched when no stored record exists.
- [x] **Cover committee rotation:** regression tests exercise retiring and joining keys against the latest stored record, plus empty storage with default and observer modes.

## Testing

- Passed: `cargo +nightly-2026-03-20 fmt --all -- --check`.
- Passed: `git diff --check` and a strict diff review. `ctxcat-review` was unavailable as a callable workflow.
- Passed: `cargo +1.94 test --locked -p tn-node --lib startup_node_mode_ --target-dir target -j 1 -- --nocapture`, 2 tests passed.
- Mutation confirmed: both new tests failed at their intended assertions when the membership and empty-store branches were mutated. Both passed again after restoring the fix.
- Failed on existing test-code lints: the complete 14-package clippy scope emitted by `buildplan`, using default features and an isolated target directory. A second run with `--keep-going` reported failures in e2e, primary, libp2p, engine, executor, and storage test targets. All reported locations are outside the changed file and match base `58c0a62d`; there were no diagnostics in `manager/node.rs`.

The clippy command, with the isolated target path expressed relative to the checkout:

```sh
cargo +nightly-2026-03-20 clippy --locked --keep-going \
  -p e2e-tests -p state-sync -p telcoin-network -p telcoin-network-cli \
  -p tn-batch-builder -p tn-batch-validator -p tn-engine -p tn-executor \
  -p tn-network-libp2p -p tn-node -p tn-primary -p tn-storage \
  -p tn-test-utils -p tn-worker --all-targets -j 1 \
  --target-dir ../telcoin-1350-clippy-target --no-deps -- -D warnings
```

This PR remains a draft pending the required validation and maintainer attestation. CI run `34412273741` skipped the build and test jobs for this maintainer-authored PR. Its `verify-on-chain` check reports that commit `fb169e1d3783468a5a9ef46556f040852c399e65` needs `make attest`; the full attested suite is required before merge.


sstanfield (approved):
LGTM

grantkee (approved):
LGTM

We normally use the execution state as the source-of-truth, but I think it makes sense to read from the epoch records here in the event a node is recovering and in a future epoch. Epoch records sync more quickly so a node can determine the current/future committees from these records while execution state syncs

PR #1363, opened 2026-09-09
fix(node): start epoch pack workers before filling request queue

### LINKED ISSUE #1353: request_missing_packs fills a bounded queue before its consumers spawn
## Problem

`request_missing_packs` (`crates/state-sync/src/consensus.rs:726`) is called at
`crates/node/src/manager/node.rs:1013` during startup. It delegates to `request_epochs`
(`consensus.rs:203`), which walks forward from the last executed consensus epoch and, for every
epoch whose pack file is missing or incomplete, awaits `request_epoch_pack_file`
(`consensus.rs:230-232`).

`request_epoch_pack_file` (`crates/consensus/primary/src/consensus_bus.rs:834-840`) is one
`send().await` on `epoch_request_queue_tx`, a bounded `tokio::sync::mpsc::channel(1024)`
(`consensus_bus.rs:389`).

The three tasks that drain that queue are spawned immediately afterwards, at `node.rs:1016-1035`.

So the fill runs to completion against a queue with no consumers. A node missing more than 1024
epoch packs blocks on `send().await` at slot 1025 and never reaches the loop that would have spawned
the drainers. Startup stalls with no error and no log line saying why.

1024 epochs is a lot, and no current deployment is near it. But the ordering is load-bearing by
accident, and nothing in the code says so.

## Proposed change

Spawn the three consumer workers before calling `request_missing_packs`, so the queue drains while
it fills.

The workers go dormant when the queue is empty (the comment at `node.rs:1014-1015` already says so),
so starting them earlier costs nothing. Check that everything the worker closure captures is
available at the earlier point: it needs the shutdown notifier, the consensus bus, the primary
network handle and the consensus chain, all of which exist by `node.rs:1013`.

If the reorder turns out to be constrained by something not visible here, the alternative is a
bounded fill: cap the number of packs enqueued in one pass, or use `try_send` and stop on a full
queue, leaving the rest to the collector's periodic retry.

Either way, add a comment recording why the ordering matters, so a later reshuffle does not
reintroduce it.

## Risk

Low. Both tasks already run concurrently for the whole life of the process. The only change is that
they start overlapping a few statements earlier.

## Acceptance criteria

- The queue has live consumers before `request_missing_packs` starts filling it, or the fill is
  bounded so it cannot block on a full queue.
- The reason for the ordering is recorded in a comment at the call site.

MavenRain:
Closes #1353.

## Problem

A node missing more than 1,024 epoch packs can stall during startup with no attacker required. `request_missing_packs` fills a bounded queue before the three workers that drain it are spawned, so the 1,025th send waits indefinitely.

## Changes

- **Start consumers first:** move `request_missing_packs` immediately after spawning all three epoch-pack workers so the queue can drain while startup fills it.
- **Document the ordering:** explain the bounded-queue dependency at the call site and update the startup method documentation.

## Testing

- Passed `gateledger run -- cargo +nightly-2026-03-20 fmt --all -- --check` and `git diff --check`.
- Ran `diffclass --base origin/main`, `apidelta --base origin/main --verb clippy`, and `buildplan clippy --base origin/main`. The plan retains 14 affected crates because of feature-dependent code.
- Attempted the emitted Clippy scope with `CARGO_TARGET_DIR=/Users/oobi/Documents/.telcoin-shared-target gateledger run -- cargo +nightly-2026-03-20 clippy --offline --locked -p e2e-tests -p state-sync -p telcoin-network -p telcoin-network-cli -p tn-batch-builder -p tn-batch-validator -p tn-engine -p tn-executor -p tn-network-libp2p -p tn-node -p tn-primary -p tn-storage -p tn-test-utils -p tn-worker --all-targets -j 2 --no-deps -- -D warnings`. Stopped during dependency compilation while another Rust build was active on the machine; no Clippy result is claimed.
- Reviewed task startup through `TaskManager::create_task`: it calls `tokio::spawn` immediately, allowing the consumers to run while the producer awaits queue capacity.
- Workspace check, Clippy, nextest, and the Rust 1.94 attestation remain outstanding. Maintainer PRs skip the compile/test CI lanes; full validation and attestation must precede merging.


sstanfield (approved):
Good catch.

PR #1364, opened 2026-09-09
feat(network): configure bootstrap peers via YAML and CLI

### LINKED ISSUE #1354: Let operators configure bootstrap peers in NetworkConfig
Depends on #1349. That issue takes the committee out of the bootstrap-server path; this one makes
what is left operator-configurable.

## Problem

`EpochManager::bootstrap_servers` (`crates/node/src/manager/node.rs:223`) can only ever hold what
the genesis committee file says. It is read once at `node.rs:694-703` from
`tn_datadir.committee_path()` and never changed.

An operator who wants a node to bootstrap against different peers, a private relay, a regional
seed set, or a replacement for a genesis peer that no longer exists, has to edit the genesis
committee file. That file is the chain's trust root. Editing it to change a dial hint is the wrong
lever.

## Proposed change

Add a field to `NetworkConfig` (`crates/config/src/network.rs:11-27`):

```rust
bootstrap_peers: BTreeMap<BlsPublicKey, BootstrapServer>,
```

`BootstrapServer` is the type at `crates/types/src/committee.rs:216-221`. Note it holds
`workers: Vec<P2pNode>`, not a single worker, so the field and its tests must be written against
that shape. It has a hand-written `Deserialize` (`committee.rs:223-231`) and a legacy conversion
(`committee.rs:233-236`), both of which the map inherits.

`NetworkConfig` already carries `#[serde(default)]` (`network.rs:13`), so an existing config file
with no `bootstrap_peers` key keeps parsing and yields an empty map.

In `run()`, immediately after the network config is read at `node.rs:843`, override
`self.bootstrap_servers` with the configured set when it is non-empty. An empty set keeps today's
genesis fallback.

## Why this is inert for current deployments

No `network-config` file is checked in anywhere in the repo, and `NetworkConfig::read_config`
(`crates/config/src/network.rs:91`) falls back to the default when none exists. So the common case
today is `NetworkConfig::default()`, which under this change gives an empty `bootstrap_peers` and
therefore the genesis fallback, exactly as now.

## Open question: override or merge

The proposal above is full override: a non-empty configured set replaces the genesis set entirely.

Override is the more predictable of the two. An operator who writes a bootstrap list gets that list,
and can deliberately exclude a genesis peer that is gone. The cost is that an operator who wants to
add one peer has to restate all of them, and a partial list silently drops the rest.

Merge (union by BLS key, configured entry wins on collision) makes the additive case easy but gives
no way to remove a dead genesis peer, which is one of the motivating cases.

Pick one and document it on the field. Do not leave it implicit.

## Risk

Low. The failure mode is an operator writing a bootstrap list that cannot be reached, which is the
same failure mode as a stale genesis peer and produces the same symptom: the readiness wait at
`crates/node/src/manager/node/start_epoch.rs:1037-1057` times out with a clear error.

Worth stating in the field's doc comment that these are dial hints for peer discovery, not a trust
or authorization decision. Committee membership and gossip publisher allowlists are derived from
chain state and are unaffected by what goes here.

## Acceptance criteria

- `NetworkConfig` has a `bootstrap_peers` field that round-trips through YAML.
- An absent key, and an existing config file written before this change, both parse to an empty map.
- An empty map produces exactly the bootstrap payload the node registers today.
- A non-empty map produces the configured payload, under whichever of override or merge is chosen.
- The chosen semantics are documented on the field.

MavenRain:
Closes #1354.

Stacked on #1359 (`fix/1349-bootstrap-peer-ownership`). Merge that prerequisite first, then retarget this PR to `main`.

## Summary

Operators can choose bootstrap peers without editing the genesis committee. `NetworkConfig.bootstrap_peers` supplies a full replacement seed map, and `--bootstrap-peers <MAP>` accepts a YAML or JSON map as a process-only override.

Precedence is CLI, then network config, then genesis. A nonempty selected map replaces genesis completely; an absent config key or empty selected map preserves the genesis fallback. An explicitly empty CLI map also restores genesis when YAML specifies custom seeds.

These entries are discovery dial hints. Committee membership and gossip publisher authorization remain derived from chain state through the separate registration path introduced by #1359. Unreachable seeds can prevent network discovery, as with stale genesis seeds; no attacker is required.

## Changes

- **Network config:** private bootstrap map, accessor, documented replacement semantics, and shared precedence resolution.
- **Node startup:** resolve the selected map before constructing swarms; primary and worker registration consume that map.
- **CLI:** parse complete peer maps, including multiple workers and YAML document markers. Carry the override through `TnBuilder` without writing it to disk; update both builder construction sites to use its constructor.
- **Operator docs:** map shape, CLI usage, fallback behavior, and legacy single-worker compatibility.

## Testing

- Passed: `cargo +nightly-2026-03-20 fmt --all -- --check`, run through `gateledger`.
- Passed: `git diff --check` and strict static diff review, including the startup wiring and both bootstrap registration consumers. No CI configuration changes. `ctxcat-review` was unavailable.
- `diffclass` requires a build; `apidelta` retains the dependency closure; `buildplan clippy` selects 23 crates with `--all-targets`.
- Attempted the complete planned Clippy scope with `--no-deps -- -D warnings` under the pinned nightly and an isolated target directory. `tnquiet` timed out after 60 seconds waiting for another build (exit 75), before Cargo started.
- Added eight deterministic tests for YAML round trips, missing/legacy config, legacy worker decoding, replacement, CLI precedence, CLI YAML/JSON parsing, explicit empty overrides, and invalid entries. These tests and their mutation checks have **not run** because of build contention.
- Pending: workspace check, Clippy, nextest, and the maintainer's pinned Rust 1.94 attestation suite. The PR targets a prerequisite branch, so workflows restricted to `main` do not provide coverage here; the maintainer lanes also skip ordinary compile/test CI after retargeting.

Draft until the prerequisite and dynamic validation are complete.

### DIFF
--- a/crates/config/src/network.rs
+++ b/crates/config/src/network.rs
@@ -3,8 +3,8 @@
 use crate::{ConfigFmt, ConfigTrait, TelcoinDirs};
 use libp2p::kad::K_VALUE;
 use serde::{Deserialize, Serialize};
-use std::{num::NonZeroUsize, time::Duration};
-use tn_types::{Round, WorkerId};
+use std::{collections::BTreeMap, num::NonZeroUsize, time::Duration};
+use tn_types::{BlsPublicKey, BootstrapServer, Round, WorkerId};
 
 impl ConfigTrait for NetworkConfig {}
 
@@ -24,9 +24,38 @@ pub struct NetworkConfig {
     peer_config: PeerConfig,
     /// The hostname for the validator.
     hostname: String,
+    /// Bootstrap dial hints for peer discovery, keyed by BLS public key.
+    ///
+    /// A nonempty map replaces the genesis bootstrap set in full; entries are never merged.
+    /// An empty or absent map preserves the genesis fallback. A CLI override takes precedence
+    /// over this map, with an explicitly empty override selecting the genesis fallback.
+    /// Committee membership and gossip publisher authorization remain derived from chain state.
+    bootstrap_peers: BTreeMap<BlsPublicKey, BootstrapServer>,
 }
 
 impl NetworkConfig {
+    /// Return the configured bootstrap dial hints.
+    pub fn bootstrap_peers(&self) -> &BTreeMap<BlsPublicKey, BootstrapServer> {
+        &self.bootstrap_peers
+    }
+
+    /// Select bootstrap dial hints: CLI override, then network config, then genesis.
+    ///
+    /// The selected override or config map replaces genesis in full when nonempty. An empty
+    /// selected map falls back to genesis, including when the CLI explicitly supplies `{}`.
+    pub fn resolve_bootstrap_peers(
+        &self,
+        genesis: &BTreeMap<BlsPublicKey, BootstrapServer>,
+        cli_override: Option<&BTreeMap<BlsPublicKey, BootstrapServer>>,
+    ) -> BTreeMap<BlsPublicKey, BootstrapServer> {
+        let configured = cli_override.unwrap_or(&self.bootstrap_peers);
+        if configured.is_empty() {
+            genesis.clone()
+        } else {
+            configured.clone()
+        }
+    }
+
     /// Return a reference to the [SyncConfig].
     pub fn sync_config(&self) -> &SyncConfig {
         &self.sync_config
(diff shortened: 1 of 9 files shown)


sstanfield (dismissed):

grantkee (approved):
LGTM!

Issue #1001, opened 2026-07-23
network-libp2p: inbound Kademlia AddProvider records are stored with no ban or authenticity check, unlike the gated PutRecord path


MavenRain:
`network-libp2p`: inbound Kademlia `AddProvider` records are stored with no ban or authenticity check, unlike the gated `PutRecord` path

**Class:** input-validation / defense-in-depth asymmetry in the Kademlia inbound-request handler.  Confirmed mechanism on `origin/main` (`b6b3189e`).  The `kad::InboundRequest::AddProvider` arm persists an attacker-supplied provider record unconditionally, while the sibling `PutRecord` arm ban-checks the source and publisher and validates a BLS signature before it writes anything.  Severity is **Low**: the write is capped and TTL-evicted, the provider identity is not forgeable, and no TN application-level decision reads provider records.  It is a hardening gap rather than an exploitable break, but it is a genuine pre-auth write surface reachable by any non-banned peer (including a peer already banned at the app layer, whom the `PutRecord` path would reject and this path does not).

## Summary

Kademlia runs with `set_record_filtering(kad::StoreInserts::FilterBoth)` (`crates/network-libp2p/src/consensus.rs:390`), so libp2p does not auto-store inbound records or providers;  the application is handed each inbound request and is the sole write path into the persistent store.  The two write arms are asymmetric:

- `kad::InboundRequest::PutRecord` (`consensus.rs:1732-1736`) routes through `process_kad_put_request` (`consensus.rs:1887`), which ban-checks both publisher and source, validates the BLS signature via `peer_record_valid`, and enforces a newer-record check before storing.
- `kad::InboundRequest::AddProvider` (`consensus.rs:1721-1730`) calls `store_mut().add_provider(record)` unconditionally.  Its only guard is the `if let Some(record)` `None`-filter.  There is no ban check, no authenticity check, and no newer-record check.

The asymmetry is complete and appears to be an oversight rather than a deliberate design choice: the `PutRecord` reject path is carefully hardened (it even carries a multi-line comment explaining why it must not mutate the store on a sender-supplied key, from the own-record-deletion fix), while the `AddProvider` arm next to it was left ungated.

## The defect

**The `AddProvider` arm stores unconditionally** (`consensus.rs:1721-1730`):

```rust
kad::InboundRequest::AddProvider { record } => {
    if let Some(record) = record {
        self.swarm
            .behaviour_mut()
            .kademlia
            .store_mut()
            .add_provider(record)
[...]

PR #1368, opened 2026-09-10
fix(network-libp2p): derive KadStore::node_key from raw BLS key bytes (#1331)

### LINKED ISSUE #1331: KadStore::node_key BCS-encodes the BLS key while the DHT record uses raw bytes, so the node never republishes its own provider record
`KadStore::node_key` and `ConsensusNetwork::get_peer_record` derive the same DHT key two different ways, so the row `provided()` reads is never the row production writes.

## Problem

`KadStore::new` builds the node's own record key from the **BCS encoding** of the primary BLS public key, while `ConsensusNetwork::get_peer_record` builds it from the **raw** key bytes:

```rust
let node_key = RecordKey::new(&encode(&key_config.primary_public_key()));   // kad.rs:221  -> 97 bytes
let key      = kad::RecordKey::new(&self.key_config.primary_public_key());  // consensus.rs:618 -> 96 bytes
```

`encode` is BCS (`crates/types/src/codec.rs:69-71`). `BlsPublicKeyBytes` serializes through `serialize_bytes` on the non-human-readable path (`crates/types/src/crypto/bls_public_key.rs:230-240`), so BCS prepends a ULEB128 length byte — `0x60` — ahead of the 96 key bytes. `AsRef<[u8]>` returns `&self.bytes.0`, the bare 96 bytes (`bls_public_key.rs:177-181`).

Two different byte strings go into `key_to_hash`, so they land on two different rows.

## Consequence

`RecordStore::provided` reads `self.providers(&self.node_key)` (`kad.rs:604-620`) — the 97-byte slot. Production only ever writes the 96-byte slot. libp2p's `AddProviderJob` therefore enumerates a permanently empty slot and **the node never republishes its own provider record**. Once the initial `start_providing` entry ages out of remote peers' stores, nothing refreshes it.

## Why the test doesn't catch it

`test_provided_enumerates_only_self_authored_records` (`kad.rs:996-1034`) constructs its own `node_key` with the `encode`d form:

```rust
let node_key = RecordKey::new(&encode(&key_config.primary_public_key()));
```

Both sides of the test agree with `KadStore`, so the assertion passes vacuously against the slot production never touches.

## Fix

```rust
let node_key = RecordKey::new(&key_config.primary_public_key());
```

`kad.rs:996-1034` needs the same change on both sides, otherwise it keeps passing without exercising anything.

## Notes

Pre-existing, **not** introduced by #1315 — `git show origin/main:crates/network-libp2p/src/kad.rs` carries the identical line at `:221`, and `kad.rs` is unchanged on that branch. Surfaced independently by two agents while reviewing #1315 (per-worker network swarms, #555 PR 2) and split out here because it is unrelated to per-worker swarms.

Touches the issue #1001 provider-injection guard — the `provider == local_peer_id` filter in `provided()` — so the fix should keep that guard's test meaningful rather than just moving the key.

grantkee:
- helper function to consistently generate kad record keys
- consolidate record key generation (fix bug)

closes #1331


crates/network-libp2p/src/kad.rs
```diff
@@ -217,6 +217,19 @@ impl KadProviderRow {
 /// interval bounds that cost without losing any eviction a 48h TTL could yield.
 const PROVIDER_EVICT_INTERVAL: Duration = Duration::from_secs(60);
 
+/// The DHT key under which a node's [`NodeRecord`](crate::types::NodeRecord) lives: the
+/// raw bytes of its primary BLS public key.
+///
+/// This is deliberately *not* the BCS encoding of the key, which would prepend a ULEB128
+/// length prefix and land on a different row. Every producer and consumer of a node's
+/// record key must go through this function: `ConsensusNetwork::get_peer_record` publishes
```
MavenRain, line 225: **Doc/code drift (medium).** This doc says every producer and consumer of the node record key goes through `node_record_key`, and that "the lookups by BLS key query it". The lookup path does not. `find_authorities` builds the key inline at consensus.rs:1768 (`kad::RecordKey::new(&bls_key)`), as do the test-only `KadStoreGet` arm at consensus.rs:1088 and network_tests.rs:3621. The two forms agree today only because the helper body is `RecordKey::new`. If the helper ever changes (domain separation, a different encoding), publish and lookup split again while this doc still says they cannot. Either route those three sites through `node_record_key` (the commit message already names the first two), or narrow this sentence to the two callers that use it.

crates/network-libp2p/src/kad.rs
```diff
@@ -262,7 +275,7 @@ impl<DB: Database> KadStore<DB> {
         key_config: &KeyConfig,
         kad_type: NetworkType,
     ) -> Self {
-        let node_key = RecordKey::new(&encode(&key_config.primary_public_key()));
+        let node_key = node_record_key(&key_config.primary_public_key());
```
MavenRain, line 278: **Stale test comment (medium).** `test_kad_store` (kad.rs:1102-1105) still builds its provider key as `RecordKey::new(&encode(&key_config.primary_public_key()))` under the comment "Make manually to use our node key as key." After this line that key is the 97-byte BCS form and `node_key` is the 96 raw bytes, so the comment is false and those provider counts no longer cover the `provided()` slot. Switch that key to `node_record_key(&key_config.primary_public_key())`, or reword the comment to say the key is deliberately not the node key.

crates/network-libp2p/src/kad.rs
```diff
@@ -1250,6 +1266,46 @@ mod test {
...
+            kad_store.node_key.as_ref(),
+            encode(&primary_key).as_slice(),
+            "node_key must not be the BCS encoding of the key"
+        );
+
+        // A self-provide under the published key is what the republish job enumerates.
+        let ours = ProviderRecord {
+            key: published_key,
+            provider: local_peer_id,
+            expires: Instant::now().checked_add(Duration::from_secs(60 * 60 * 24)),
```
MavenRain, line 1299: **Test models the wrong expiry (low).** Production `start_providing` stores the local provider record with `expires: None` (libp2p `ProviderRecord::new`, record.rs:153), and `provided()` is the only reader of that row. This test stores it with `expires: Some(now + 24h)`, so the `None` round trip through `KadProviderRow` (kad.rs:109) and the eviction filters runs only in production. `expires: None` here makes the test match the row the republish job reads. The sibling test above has the same shape but tests the attacker filter, so it can stay as is.

crates/network-libp2p/src/kad.rs
```diff
@@ -1250,6 +1266,46 @@ mod test {
...
+            addresses: vec![],
+        };
+        kad_store.add_provider(ours).expect("add our provider record");
```
MavenRain, line 1302: **Missing `sync_persist` (low).** The sibling test calls `kad_store.db.sync_persist()` between `add_provider` and the read-back so the background writer catches up before `providers()` iterates. This test reads `provided()` straight after `add_provider`. The storage tests note that inserts show up immediately but an iteration during persist can observe a row twice, which would fail the `len() == 1` assert. Add the same call here.

crates/network-libp2p/src/kad.rs
```diff
@@ -217,6 +217,19 @@ impl KadProviderRow {
...
+/// The DHT key under which a node's [`NodeRecord`](crate::types::NodeRecord) lives: the
+/// raw bytes of its primary BLS public key.
+///
+/// This is deliberately *not* the BCS encoding of the key, which would prepend a ULEB128
+/// length prefix and land on a different row. Every producer and consumer of a node's
+/// record key must go through this function: `ConsensusNetwork::get_peer_record` publishes
+/// under it, the lookups by BLS key query it, and [`KadStore`] uses it as `node_key` so
+/// [`RecordStore::provided`] enumerates the row `start_providing` actually wrote (issue
+/// #1331).
+pub fn node_record_key(primary_public_key: &BlsPublicKey) -> RecordKey {
```
MavenRain, line 229: **Visibility (low).** All callers of `node_record_key` live in this crate (`consensus.rs` and the tests). `pub(crate)` carries the same guarantee without adding `tn_network_libp2p::kad::node_record_key` to the crate's public API.

MavenRain (dismissed):
**Verdict:** the fix is correct and minimal. `KadStore::new` now derives `node_key` with the same expression `get_peer_record` publishes under (the 96 raw key bytes, not the 97-byte BCS form), so `RecordStore::provided()` enumerates the row `start_providing` writes and libp2p's `AddProviderJob` republishes it. No migration is needed: the only production write keyed on the node's own key is `start_providing`, which always used the raw form, and no production path reads or writes the old encoded row. Worker stores get the same fix through the shared `get_peer_record`. The #1001 `provider == local_peer_id` guard in `provided()` becomes reachable and the sibling test covers it.

Five inline notes, none blocking. Two are doc and test drift this PR introduces or makes stale, three are test-hygiene nits.

1. kad.rs:225 (medium) The helper doc says every producer and consumer goes through `node_record_key`, but the lookup path still builds the key inline: `find_authorities` (consensus.rs:1768), the test-only `KadStoreGet` arm (consensus.rs:1088) and network_tests.rs:3621. Route them through the helper or narrow the doc.
2. kad.rs:278 (medium) `test_kad_store` (kad.rs:1102-1105) still keys its provider records on `encode(...)` under the comment "Make manually to use our node key as key." After this PR that key is not `node_key`, so the comment is false and the test no longer touches the `provided()` slot.
3. kad.rs:1299 (low) The new test stores the self-provide with `expires: Some(now + 24h)`. Production `start_providing` stores it with `expires: None`. Use `None` so the test covers the row the republish job reads.
4. kad.rs:1302 (low) The sibling test calls `kad_store.db.sync_persist()` between `add_provider` and the read-back. The new test does not. Add it for the same read-your-writes ordering.
5. kad.rs:229 (low) `node_record_key` can be `pub(crate)`. All callers are in this crate.

Also: the PR title ends in a literal "…" because GitHub truncated the 78-character commit subject. A squash merge would carry it into the merged subject. Retitle before merging.

Verification: static review at head 6c83addf. The PR-triggered fmt, clippy and test lanes are skipped for maintainer PRs by design. `verify-on-chain` passed, so this head SHA is attested in the adiri registry. I did not run a local build.

crates/network-libp2p/src/kad.rs
grantkee: All good points @MavenRain addressed your feedback and ready for rereview

grantkee:
Good feedback @MavenRain - I incorporated your suggestions and ready for re-review

MavenRain (approved):
Reviewed 34ad938c. All five earlier comments are addressed. Publishing, lookup, and provider enumeration now agree on the raw BLS key, with the self-authored provider filter and worker isolation preserved. The regression test covers the persisted `expires: None` case.

Validation: reviewed the complete diff and the pinned libp2p republishing implementation; `git diff --check` passed. The successful `verify-on-chain` job attests this exact head. I did not rerun the Rust suite locally.

Blockers: none. Approve.

PR #1370, opened 2026-09-11
fix(tn-reth): propagate RPC transaction pool batch size

### LINKED ISSUE #1365: Fix review: --txpool.max-batch-size is not propagated
In PR #1161, it omits applying `max_batch_size(config.txpool.max_batch_size());` to `config.rpc.eth_config()`. Thus, `--txpool.max-batch-size` config is ignored. Stock reth applies the setting in https://github.com/paradigmxyz/reth/blob/d6324d63e27ef6b7c49cdc9b1977c1b808234c7b/crates/node/builder/src/rpc.rs#L1002

--txpool.max-batch-size controls how many queued RPC transactions Reth inserts into its transaction pool together and is not to be confused with consensus batches.

https://github.com/Telcoin-Association/telcoin-network/issues/1340 and its fix does not cover this.

MavenRain:
Closes #1365.

## Problem

Setting `--txpool.max-batch-size` leaves RPC transaction pool insertion batching at reth's default. `get_rpc_server` constructs `EthConfig` from the RPC arguments alone, so the operator's transaction pool setting never reaches the existing `apply_eth_config` setter. This occurs during ordinary node startup, with no attacker required.

## Changes

- **Propagate the CLI setting:** apply `node_config().txpool.max_batch_size` to `EthConfig` before passing it to the eth API builder, matching reth v1.11.3's startup configuration. This controls insertion of queued RPC transactions into the pool.

## Testing

- Passed `cargo +nightly-2026-03-20 fmt --all -- --check` through `gateledger`.
- Passed `git diff --check` and static review of the configuration path against the pinned reth source.
- `diffclass`, `apidelta`, and `buildplan` selected a 21-crate Clippy scope. The local disk guard blocked that run with 19.3 GiB free, below its 30 GiB floor.
- Compile, runtime tests, and the full Rust 1.94 attestation remain pending. Maintainer-authored PRs skip the repository's compile and test lanes on `pull_request`.


Haxatron:
LGTM

sstanfield (approved):

grantkee (approved):

PR #1371, opened 2026-09-11
fix(network): refresh published node records

### LINKED ISSUE #1367: Self-published node records dropped as stale - remote copies lapse after the 48 h TTL
# A node's own kad record is signed once at boot and byte-identical republishes are dropped as stale, so remote copies lapse after the 48 h TTL

## Problem

`ConsensusNetwork::new` signs the node's record once (`create_node_record`, `crates/network-libp2p/src/consensus.rs:567-573` and `:603-613`) and stores it in `self.node_record`. `get_peer_record` (`:617-629`) re-encodes the same bytes for every publication, and `NodeRecord::build` (`crates/network-libp2p/src/types.rs:950-964`) is the only place the `timestamp` is set. Republication is libp2p's `PutRecordJob`, configured with `set_publication_interval(kad_publication_interval)` (`consensus.rs:455`; default 12 h, `crates/config/src/network.rs:202`), so every 12 hours each peer that stores the record receives the identical bytes again.

On the receiving side `process_kad_put_request` (`consensus.rs:2000-2083`) only writes the record when `is_newer_record` (`:2144-2162`) returns true, and that comparison is a strict `existing.timestamp < new.timestamp`. An identical republish is therefore logged as "ignoring stale but valid kad record" (`:2071-2074`) and never reaches `KadStore::put` (`crates/network-libp2p/src/kad.rs:501-530`), so the stored row's `expires` is never refreshed. `KadStore::get` filters expired rows (`kad.rs:485-498`) and `evict_expired_records` removes them, so a remote copy of any node's record lapses `kad_record_ttl` (48 h, `network.rs:201`) after it was first stored, regardless of how often the owner republishes.

Today this is masked: every first connection pushes the record directly (`publish_our_data_to_peer`, `:695-703`, gated by `mark_published_to_peer` at `:1729`) and `known_peers` retains the resolution independently of the store. For the sentry design it becomes load-bearing. A sentried validator's record only reaches the DHT through its sentries (PR 10 republishes it with `put_record_to`, publisher preserved), and if identical re-puts refresh nothing, the copies on non-sentry nodes lapse after 48 h and a new committee member cannot resolve the validator when it needs to (R11). The hourly kad replication job has the same problem: it re-puts the same bytes.

## Solution

Sender side. Keep the inputs of `create_node_record` on `ConsensusNetwork` (record domain, external addresses, `rpc`) and add a `refresh_own_record()` method that rebuilds `self.node_record` with a fresh `timestamp` through `NodeRecord::build`, writes it to the local store with `store_mut().put`, and calls `provide_our_data()` so the new bytes are what `PutRecordJob` snapshots next. Drive it from a `tokio::time::interval(kad_publication_interval)` arm in `run()`'s `select!`, ticking after the first interval (the boot-time `provide_our_data()` at `:721` covers the initial publication). Signing uses `key_config.request_signature_direct` as today. Read the interval from `LibP2pConfig` so tests can shorten it.

Receiver side. Replace the boolean `is_newer_record` with `fn record_freshness(&mut self, record) -> RecordFreshness { Newer, Identical, Older, Undecodable }`, comparing timestamps and, on equal timestamps, byte equality of `value`. In `process_kad_put_request` treat `Identical` as an expiry refresh: call `store_mut().put(record)` again so the row's `expires` is recomputed by `KadStore::put`, and keep the `add_self_advertised_peer` call (`:2064-2067`) so the identity confirmation is refreshed too. `Older` stays log-only with no penalty (`:2069-2074`), and `Undecodable` keeps the current behaviour.

The kad store's `expires` for the node's own record stays `None` (`:619-622`), so libp2p keeps recomputing `now + ttl` on every replication snapshot as today.

Compatibility. No wire or disk change. An old receiver treats the re-signed record as newer (its timestamp advanced) and stores it; an old sender's identical republishes are now refreshed by new receivers instead of dropped.

Out of scope: `KadStore::node_key` encoding the BLS key with bcs so the node never republishes its own provider record (#1331, separate prerequisite issue); sentry-side republish of third-party records (PR 10); the record's address list (PR 2).

## Tests

- `crates/network-libp2p/src/tests/network_tests.rs`: with `kad_publication_interval` set to one second, a peer's `kad_store_get` for the node's record shows a strictly increasing `timestamp` across two republish ticks.
- `crates/network-libp2p/src/tests/network_tests.rs`: pushing a byte-identical record a second time refreshes the stored row's `expires` (read it back through the `KadStore` after the second push and compare with the first).
- `crates/network-libp2p/src/tests/network_tests.rs`: a record with an older timestamp than the stored one is not written and the pusher's score is unchanged.
- `crates/network-libp2p/src/kad.rs` unit tests: `KadStore::put` of an existing key replaces the row and updates `expires`.

## References

- `crates/network-libp2p/src/consensus.rs` (`create_node_record`, `get_peer_record`, `provide_our_data`, `publish_our_data_to_peer`, `run`, `process_kad_put_request`, `is_newer_record`), `crates/network-libp2p/src/kad.rs` (`KadStore::get`, `KadStore::put`, `evict_expired_records`), `crates/network-libp2p/src/types.rs` (`NodeRecord::build`), `crates/config/src/network.rs` (`kad_record_ttl`, `kad_publication_interval`).
- Verified libp2p-kad 0.48 facts relied on: `PutRecordJob` re-puts every stored record on the publication interval and the replication job re-puts hourly, publisher preserved; `put_record_to` preserves the publisher while `put_record` overwrites it.
- Related issues: #1331 (adjacent bug in `KadStore::node_key`, not fixed here), #447 (publisher rule, unchanged).

MavenRain:
Closes #1367.

## Problem

A node signs its discovery record once at startup. Identical republications are treated as stale, so remote copies expire despite a healthy publisher. No attacker is required; a later committee member can lose the DHT record needed to resolve the node.

## Root cause

The publication job reuses the original signed bytes, while the receiver only stores records with a strictly greater timestamp.

## Changes

- **Periodic signing:** rebuild and publish the local record on `kad_publication_interval`, starting after the first full interval and skipping missed ticks. Preserve the configured address, network identity, RPC metadata, signing domain, and non-expiring local copy.
- **Compatible refresh:** distinguish newer, identical, older/conflicting, and undecodable records. Accept identical signed values after the existing ban, rate-limit, signature, and publisher checks. Extend their expiry while preserving a longer existing lifetime when a replica arrives with less remaining TTL. Retain identity confirmation and leave older or conflicting records unchanged without penalizing the sender.
- **Regression coverage:** exercise two publication ticks across connected peers, identical-record expiry refresh, stale/conflicting input and unchanged scores, and store replacement at capacity across primary and worker namespaces.

No wire or disk format changes. Upgraded publishers produce newer records for older receivers; upgraded receivers accept identical republishes from older publishers. Provider-key encoding in #1331 remains separate.

## Testing

- Passed: pinned formatting, whitespace validation, and strict diff review. No CI configuration changed.
- `diffclass` requires compilation. `apidelta` retains the dependency closure; `buildplan clippy` selects 21 packages with `--all-targets`. The complete scope ran under `nightly-2026-03-20`, with `--locked --offline --keep-going --no-deps -- -D warnings`. It failed on 131 lints in existing test code across six packages, with no diagnostics in the new or changed code. The networking test-target lint is `type_complexity` at the unchanged `src/tests/peer_manager.rs:550`. This is not a green full-scope check.
- Passed: 40 record-filter tests and the separate store-expiry test on Rust 1.94, including all four new regression tests.
- Five temporary production mutations were detected by test failures: omit re-signing, reject identical republishes, permit shortened expiry, accept equal-timestamp conflicting values, and accept older values. The final source was restored after the mutation checks.

```sh
cargo +1.94 test --locked --offline -p tn-network-libp2p --lib record
cargo +1.94 test --locked --offline -p tn-network-libp2p --lib test_kad_put_refreshes_expiry_at_capacity
git diff --check
```

Tests used an isolated target directory, `CARGO_PROFILE_TEST_DEBUG=0`, `CARGO_INCREMENTAL=0`, and `RUST_TEST_THREADS=2`.

This PR remains a draft because the full clippy gate failed and attestation is outstanding. The repository skips dynamic CI lanes for maintainer-authored PRs. The [final-commit PR workflow](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34628837175) failed `verify-on-chain`. Full `make attest` validation and its on-chain attestation remain required before merge.

### DIFF
--- a/crates/e2e-tests/tests/it/epochs.rs
+++ b/crates/e2e-tests/tests/it/epochs.rs
@@ -4,9 +4,13 @@ use crate::common::get_block;
 
 use super::common::{
     create_genesis_for_test, fetch_verified_epoch_record, generate_new_validator_txs, loop_epochs,
-    start_nodes, ProcessGuard, NEW_VALIDATOR,
+    start_nodes, wait_for_rpc, ProcessGuard, NEW_VALIDATOR,
 };
-use alloy::providers::{Provider, ProviderBuilder};
+use alloy::{
+    primitives::Bytes,
+    providers::{Provider, ProviderBuilder},
+    sol_types::SolCall,
+};
 use e2e_tests::NodeEndpoints;
 use rand::{rngs::StdRng, SeedableRng as _};
 use std::{
@@ -16,8 +20,9 @@ use std::{
     sync::Arc,
     time::Duration,
 };
+use tn_config::WORKER_CONFIGS_ADDRESS;
 use tn_reth::{
-    system_calls::{ConsensusRegistry, CONSENSUS_REGISTRY_ADDRESS},
+    system_calls::{ConsensusRegistry, WorkerConfigs, CONSENSUS_REGISTRY_ADDRESS},
     test_utils::TransactionFactory,
     RethChainSpec,
 };
@@ -28,7 +33,7 @@ use tn_types::{
         leader_seeded_ordering_fork_epoch_override, multi_workers_fork_active,
         seed_signature_active,
     },
-    keccak256, Address, Epoch, EpochCertificate, EpochRecord, Genesis, B256,
+    keccak256, Address, Epoch, EpochCertificate, EpochRecord, Genesis, B256, U256,
 };
 use tokio::time::timeout;
 use tracing::{debug, info};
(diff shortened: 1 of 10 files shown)


sstanfield (approved):

grantkee (approved):
Great work @MavenRain

PR #1372, opened 2026-09-11
fix(tn-reth): retry forwards after missing inclusion

### LINKED ISSUE #1366: Fix review: a byzantine validator reporting a false success can still censor forwarded transactions without having a 2nd validator
In [1175](https://github.com/Telcoin-Association/telcoin-network/pull/1175)

This PR reduces the censorship risk by forwarding to a second validator if the first fails, but does not completely eliminate it. A byzantine validator could also return a false acceptance and the prevent the second forwarding from occuring

MavenRain:
Closes #1366.

## Problem

An RPC success or `already known` reply does not prove inclusion. Sending a speculative second copy adds validator work on every healthy submission, and two validators can still acknowledge and discard it.

The observer's batch builder prunes an admitted forwarding batch, so monitoring inclusion requires retaining the signed payload separately from the ordinary pool.

## Changes

- **Inclusion-driven retries:** stop at the first acknowledgement. Retain the payload until canonical inclusion, then make another submission eligible only after three distinct newer consensus outputs omit it. Multiple execution blocks from the same output count once. The window can be changed through `with_retry_outputs`, with a nonzero value.
- **Validator rotation:** exclude every validator that previously acknowledged the transaction. A second false acknowledgement therefore starts another inclusion window. Existing transport fallback and rejection rules continue to apply before an acknowledgement.
- **Bounded retention:** use the pending pool's configured transaction-count and payload/routing-byte limits. The fixed 240-second lifetime is never refreshed by a duplicate or retry. Capacity refusal leaves the batch pooled; expiry is counted and logged.
- **Pool-owned recovery:** retain state across replacement of an epoch's forwarder. Check local canonical inclusion before reinsertion, and recover interrupted sends and insertions without overwriting a newer admission.
- **Regression coverage:** exercise real HTTP forwarding, actual canonical inclusion, pool reinsertion, output counting, duplicate suppression, validator history, capacity, expiry, and cancellation.

The healthy path costs one RPC send. This remains bounded best-effort recovery: inclusion depends on protocol progress and a reachable honest validator. The retained state is local memory and does not survive a node restart.

## Testing

- Pinned formatting passed: `cargo +nightly-2026-03-20 fmt --all -- --check` under `gateledger`.
- **72 selected production tests passed**, with zero failures or ignored tests, after restoring the mutation-tested source. This includes forwarding, retained-state, and transaction-pool regressions. The test command ran through `testledger --force` in an isolated target with incremental compilation disabled:

  ```sh
  cargo +1.94 test --offline --locked -j 2 --target-dir "$VALIDATION_TARGET" \
    -p tn-reth --lib -- --test-threads=2 forward txn_pool::tests
  ```

- All nine new retained-state tests failed under relevant mutations of changed production lines. Seven RPC/pool regressions also failed when first-acknowledgement termination, acknowledgement retention, inclusion cleanup, and retry scheduling were deliberately broken. Exact source bytes were restored before the final 72-test run.
- The full 21-package scope emitted by `apidelta`/`buildplan` was checked under `gateledger`, using the same isolated target:

  ```sh
  cargo +1.94 clippy --offline --locked --keep-going -j 2 \
    --target-dir "$VALIDATION_TARGET" \
    -p e2e-tests -p exex-indexer -p exex-lifecycle -p state-sync -p telcoin-network \
    -p telcoin-network-cli -p tn-batch-builder -p tn-batch-validator -p tn-engine \
    -p tn-executor -p tn-exex -p tn-network-libp2p -p tn-node -p tn-primary \
    -p tn-reth -p tn-rpc -p tn-storage -p tn-test-utils -p tn-test-utils-committee \
    -p tn-types -p tn-worker --all-targets --no-deps -- -D warnings
  ```

  Clippy remains blocked by existing lints. Every diagnostic source file (25 files) is byte-identical to the previous PR head, `99f2ea9b`; none of the four modified files has a diagnostic. No lints were suppressed. This stable pinned-toolchain check does not substitute for the nightly CI lanes or workspace attestation.

Full workspace attestation remains required. The current workflow skips compile/test lanes for maintainer-authored PRs, so a skipped lane is not validation evidence.

### DIFF
--- a/crates/tn-reth/src/forward.rs
+++ b/crates/tn-reth/src/forward.rs
@@ -5,7 +5,7 @@
 //! worker record (issue #804), so the submitter gets the same RPC experience they would get
 //! talking to a validator directly. Routing mirrors [`submit_txn_if_mine`]: a transaction is
 //! sent to the validator whose committee slot owns the sender, so all transactions from one
-//! account converge on a single validator and nonce ordering is preserved. A validator that has
+//! account reach their owning validator first and nonce ordering is preserved. A validator that has
 //! not advertised an endpoint (or is momentarily unreachable) is skipped in favor of one that
 //! has. The fallback order rotates per node and per forward (issue #1173), so redirected
 //! traffic spreads across the committee instead of concentrating on the lowest-keyed
@@ -27,10 +27,12 @@
 //!   ([`ForwarderMetrics::record_rejection_overridden`]). Endpoint-local failures (timeout,
 //!   transport error, full pool, internal error, a refusal tied to one validator's own pool
 //!   contents or admission config) fall through to the next advertised validator, and "already
-//!   known" counts as delivered. The symmetric fabrication - a validator that answers success and
-//!   drops the transaction - is still trusted at this boundary: a fabricated success is
-//!   indistinguishable from an honest one here, and catching it would cost redundant delivery or
-//!   inclusion tracking on every honest-path forward (issue #1167 records that trade).
+//!   known" counts as delivered. A first delivery also keeps the walk going until a second distinct
+//!   validator accepts (issue #1366), so a lone validator cannot stop forwarding by claiming
+//!   success and dropping the transaction. A delivery survives later rejections, chain exhaustion
+//!   and the budget expiring. This is bounded redundancy, not an inclusion or quorum guarantee: two
+//!   colluding validators, or no reachable honest fallback within the budget, can still prevent
+//!   inclusion.
 //! - The dial target is chosen by a committee member, not by this node. The endpoint arrives inside
 //!   a BLS-signed node record, so an arbitrary network peer cannot inject one, but a committee
 //!   member can still advertise any host it likes and every observer will dial it unattended.
@@ -85,8 +87,8 @@ const FORWARD_SEND_TIMEOUT: Duration = Duration::from_secs(5);
 
 /// Bounds the total time spent forwarding one transaction across all its fallback validators, so a
 /// whole unresponsive committee cannot make a single transaction cost `committee_size` back-to-back
-/// [`FORWARD_SEND_TIMEOUT`]s. When it elapses the transaction is left unforwarded and the next one
-/// proceeds.
+/// [`FORWARD_SEND_TIMEOUT`]s. When it elapses the walk preserves any received verdict and the next
+/// transaction proceeds.
 const FORWARD_TX_BUDGET: Duration = Duration::from_secs(15);
 
 /// Bounds the wall time one spawned forward spends working through its batch, so a task's
@@ -640,17 +642,17 @@ impl WorkerRpcForwarder {
                 // Bound the whole fallback chain for this transaction: even if every advertised
                 // validator accepts the connection but never answers, one transaction cannot cost
                 // more than `FORWARD_TX_BUDGET` before the next transaction proceeds.
-                // A first rejection the walk has banked escapes the budget through this slot:
+                // A first verdict the walk has banked escapes the budget through this slot:
                 // cancellation drops the walk and everything in it, and a verdict a validator
-                // already gave must not vanish with it (issue #1167).
-                let held_rejection = Mutex::new(None);
+                // already gave must not vanish with it (issues #1167, #1366).
+                let held_verdict = Mutex::new(None);
                 let chain = timeout(
                     txn_budget,
                     walk_fallback_chain(
                         tx.as_slice(),
                         ordered,
                         &providers,
-                        &held_rejection,
+                        &held_verdict,
                         &unreachable_endpoints,
                         &cache,
                     ),
@@ -661,7 +663,7 @@ impl WorkerRpcForwarder {
                 // tally below needs the two apart, so remember which happened before
                 // collapsing them into one outcome.
                 let budget_expired = chain.is_err();
-                let outcome = expired_walk_outcome(chain, &held_rejection);
+                let outcome = expired_walk_outcome(chain, &held_verdict);
 
                 match outcome {
                     ForwardOutcome::Delivered => delivered += 1,
@@ -933,8 +935,8 @@ fn owning_validator(
     committee_slots.get(slot).cloned()
 }
 
-/// Walk one transaction across the `ordered` validators until one delivers it or a second one
-/// confirms a considered rejection.
+/// Walk one transaction across the `ordered` validators until two accept it, or two reject it
+/// without an acceptance.
 ///
 /// The advisory-until-confirmed rejection is byzantine hardening (issue #1167). Sender routing
 /// hands every transaction from one account to the same owning validator first, so under a
@@ -952,13 +954,17 @@ fn owning_validator(
 /// censoring rejection needs from one validator to two; it is hardening against a single liar,
 /// not a quorum guarantee.
 ///
-/// The cost is one extra RPC round-trip per genuinely rejected transaction, which is rare on
-/// this path: every transaction here already passed this observer's own pool validation, so a
-/// considered rejection normally means the state moved (a nonce race), not spam. The symmetric
-/// fabrication - a validator that answers success and drops the transaction - is deliberately
-/// not defended here: a fabricated success is indistinguishable from an honest one at this
-/// boundary, and catching it would cost redundant delivery or inclusion tracking on every
-/// honest-path forward. Issue #1167 records that trade.
+/// A first acceptance also walks on (issue #1366): a validator can claim success and drop the
+/// transaction just as easily as it can fabricate a rejection. Two distinct accepting validators
+/// stop the walk, including "already known" replies. Once any validator accepts, later rejections
+/// cannot erase that delivery or stop attempts to reach a second accepting validator. The first
+/// acceptance replaces any held rejection in `held`, so cancellation preserves delivery too.
+///
+/// The healthy path costs two sends. Every validator is still tried at most once under the same
+/// send, transaction and batch budgets. Exhaustion or expiry after just one acceptance still
+/// reports delivery, including for a one-validator committee. This protects against one lying
+/// success when an honest fallback is reachable within the budget; it does not prove inclusion
+/// or protect against two colluding validators.
 ///
 /// The walk also carries the endpoint demotion of issue #1145: an endpoint already recorded in
 /// `unreachable` is skipped, a send that produced no JSON-RPC verdict demotes its endpoint -
@@ -968,7 +974,7 @@ async fn walk_fallback_chain(
     tx: &[u8],
     ordered: impl Iterator<Item = BlsPublicKey>,
     providers: &BTreeMap<BlsPublicKey, (String, RootProvider)>,
-    held: &Mutex<Option<(BlsPublicKey, String)>>,
+    held: &Mutex<Option<HeldVerdict>>,
     unreachable: &Mutex<BTreeSet<String>>,
     cache: &Mutex<EndpointCache>,
 ) -> ForwardOutcome {
@@ -985,13 +991,13 @@ async fn walk_fallback_chain(
         .filter(|(_, endpoint, _)| {
             !unreachable.lock().unwrap_or_else(|poisoned| poisoned.into_inner()).contains(*endpoint)
         });
-    // A short-circuiting fold: `Ok` carries the held first rejection (which validator gave it
-    // and why) between attempts, `Err` carries a terminal verdict out of the walk.
+    // A short-circuiting fold: `Ok` carries the first verdict between attempts, and `Err`
+    // carries a terminal verdict out of the walk. A delivery dominates any rejection.
     futures::stream::iter(targets)
         .map(Ok)
         .try_fold(
-            None::<(BlsPublicKey, String)>,
-            |pending_rejection, (key, endpoint, provider)| async move {
+            None::<HeldVerdict>,
+            |pending_verdict, (key, endpoint, provider)| async move {
                 // Bound each validator's round-trip: an endpoint that accepts the connection but
                 // never answers must not stall the whole fallback chain.
                 let disposition = timeout(FORWARD_SEND_TIMEOUT, provider.send_raw_transaction(tx))
@@ -1018,24 +1024,40 @@ async fn walk_fallback_chain(
                                 .unreachable
                                 .remove(endpoint);
                         }
-                        pending_rejection.iter().for_each(|(rejecting, reason)| {
-                        ForwarderMetrics::record_rejection_overridden();
-                        warn!(
-                            target: "worker::forward",
-                            rejected_by = %rejecting,
-                            delivered_by = %key,
-                            reason = %reason,
-                            "one validator rejected a transaction another validator accepted; \
-                             the rejecting validator answered from divergent state or is byzantine"
-                        );
-                    });
-                        Err(ForwardOutcome::Delivered)
+                        pending_verdict.iter().for_each(|verdict| match verdict {
+                            HeldVerdict::Rejected { validator, reason } => {
+                                ForwarderMetrics::record_rejection_overridden();
+                                warn!(
+                                    target: "worker::forward",
+                                    rejected_by = %validator,
+                                    delivered_by = %key,
+                                    reason = %reason,
+                                    "one validator rejected a transaction another validator accepted; \
+                                     the rejecting validator answered from divergent state or is byzantine"
+                                );
+                            }
+                            HeldVerdict::Delivered => {}
+                        });
+                        // Bank the first acceptance before another await, so a budget cut during
+                        // redundant forwarding cannot turn a delivery into a drop or requeue.
+                        *held.lock().unwrap_or_else(|poisoned| poisoned.into_inner()) =
+                            Some(HeldVerdict::Delivered);
+                        if pending_verdict == Some(HeldVerdict::Delivered) {
+                            Err(ForwardOutcome::Delivered)
+                        } else {
+                            Ok(Some(HeldVerdict::Delivered))
+                        }
+                    }
+                    // A prior acceptance wins over any number of subsequent rejections, and a
+                    // rejecting fallback cannot prevent the next redundant delivery attempt.
+                    Disposition::Rejected(_) if pending_verdict == Some(HeldVerdict::Delivered) => {
+                        Ok(pending_verdict)
                     }
                     Disposition::Rejected(reason) => {
                         // A second considered rejection confirms the held one. The `tried` set
                         // keeps the two verdicts from distinct validators, which is what a single
                         // byzantine validator cannot produce alone.
-                        if pending_rejection.is_some() {
+                        if pending_verdict.is_some() {
                             Err(ForwardOutcome::Rejected(reason))
                         } else {
                             debug!(
@@ -1047,9 +1069,11 @@ async fn walk_fallback_chain(
                             );
                             // Mirror the verdict into `held` so it survives the caller's budget
                             // cancelling this walk mid-chain (see [`expired_walk_outcome`]).
+                            let verdict =
+                                HeldVerdict::Rejected { validator: Box::new(key), reason };
                             *held.lock().unwrap_or_else(|poisoned| poisoned.into_inner()) =
-                                Some((key, reason.clone()));
-                            Ok(Some((key, reason)))
+                                Some(verdict.clone());
+                            Ok(Some(verdict))
                         }
                     }
                     // Endpoint-local, but the server answered (full pool, internal error, or a
@@ -1061,13 +1085,13 @@ async fn walk_fallback_chain(
                             reason = %reason,
                             "validator RPC did not accept the forwarded transaction; trying next"
                         );
-                        Ok(pending_rejection)
+                        Ok(pending_verdict)
                     }
                     // The endpoint gave no verdict: publish the demotion at the failure itself,
                     // not at task end, so batches sealed while this task is still walking already
                     // find the endpoint unusable at admission and the cooldown runs from the
                     // failure (issue #1145). No verdict is no confirmation either: the held
-                    // rejection, if any, walks on unchanged.
+                    // verdict, if any, walks on unchanged.
                     Disposition::Unreachable(reason) => {
                         if unreachable
                             .lock()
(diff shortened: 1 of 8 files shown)


grantkee (commented):
At some level, this is a limitation of the protocol. Instead of a follow-up speculative submission, would it make more sense to to monitor the transaction for inclusion for x amount of outputs before submitting again? (protocol expects 1-3s block times?)

I'm hesitant to endorse duplicate/redundant transaction submissions as an additional strain on validator throughput even with #1330 attempting to resolve some of this. This is a slippery slope since the second validator could also censor the transaction, in which case this didn't solve the root of the issue

MavenRain:
@grantkee Agreed. I changed this to monitor inclusion, as you proposed in [your review](https://github.com/Telcoin-Association/telcoin-network/pull/1372#pullrequestreview-5225569007). Updated in [d3a448e6](https://github.com/Telcoin-Association/telcoin-network/commit/d3a448e6b3d27e3869662c801ae7eb7dac4051b5).

The healthy path now stops after the first success or `already known` reply. The observer retains the signed transaction and retries only after three distinct newer consensus outputs without inclusion, excluding all validators that previously acknowledged it. If the second validator also acknowledges and drops it, the same inclusion window applies again.

Canonical inclusion clears the record. Retention has count/byte limits and a fixed 240-second lifetime that retries cannot extend. This avoids speculative copies on the healthy path while keeping the remaining protocol limitation explicit: inclusion still needs progress and an honest reachable validator.

The restored forwarding and transaction-pool selection passes 72 tests, with mutation checks for the new regressions. The PR description records the existing Clippy failures and outstanding workspace attestation.

sstanfield (approved):

crates/tn-reth/src/forward_pending.rs
```diff
@@ -0,0 +1,527 @@
...
+    /// acknowledged transaction requires output progress, so a stalled chain causes no retry storm.
+    pub(crate) fn ready(&mut self, now: Instant, limit: usize) -> (Vec<(Hash, Vec<u8>)>, usize) {
+        let expired: Vec<_> = self
+            .entries
+            .iter()
+            .filter(|(_, entry)| now >= entry.expires_at)
+            .map(|(hash, _)| hash.clone())
+            .collect();
+        let expired_count = expired.len();
+        expired.into_iter().for_each(|hash| self.remove(&hash));
```
grantkee: Expiry here drops the record but not the transaction, so bytes still in the pool or held by a forward task get re-admitted as new, and the fixed 240 s lifetime ends up bounding a record rather than a transaction.

The PR description and the retention docs say the lifetime "is never refreshed by a duplicate or retry", and that's true of the record: `admit` (:151) keys on hash presence and reuses the existing `expires_at`. But `ready()` reaps on `now ≥ expires_at` with no state check (the lease and `retry_not_before` only gate pickup at :264), `admit()` never looks at `expires_at` on an existing entry, and `remove()` (:203-208) drops only the record. On the task side, `requeue_one` (forward.rs:837-857) still calls `add_recovered_transaction_external` (forward.rs:842-848) when the hash is absent — `reinserting`/`reinserted` are no-ops on a missing entry but the insert isn't — and the abandoned tail (forward.rs:811-814) calls it unconditionally. Whether `forward_txns` returns true (builder prunes) or false (`NotValidator`, builder keeps the txs pooled and re-seals, worker.rs:288-296), any untracked pooled tx is re-packed and hits `admit` as new: `accepted_by = {}`, `expires_at = now + 240s`.

Two ways to get there, both reproduced against this file as-is with a small harness:

- Mid-`Sending`. With T0 = first admission and E = T0 + 240 s, a retry re-admitted at `T_r > T0 + 120` gets `Sending(T_r + 120)`, which outlives E. `ready(E)` reaps it mid-walk. If the walk ends with no verdict and `had_acceptance == false` (a fresh record reaped before `accepted_by` is read at forward.rs:657-660, or every endpoint demoted), or the tx lands in the abandoned tail, `requeue_one` inserts it → next seal → fresh 240 s.
- `Queued`, and deterministic. A record in `Queued` at E is reaped while its bytes sit in the pool; the next seal (≤1 s) re-admits it. Under a committee outage `forward_txns` refuses (forward.rs:885-899) without calling `admit`, so entries sit in `Queued` through the whole refusal/backoff window and most expiries refresh rather than expire.

A quiet chain doesn't add a third path: the 100 ms `RetryTick` (txn_pool.rs:134, :452-465, :506) keeps `ready()` running, so the residual window there is ≤100 ms and the fix below covers it.

Memory stays bounded — `RetentionLimits` and pool capacity still hold — but the invariant stated at the top of this file, on `PendingTransaction` (:67, :76), on `FORWARD_PENDING_LIFETIME` (forward.rs:112-113), and in the test name at :462 bounds a record, not a transaction. Concretely: `accepted_by` is reset so the retry re-sends to validators that already hold it ("already known" counts as Delivered, which softens it), `forwarded_txns_expired_total` over-counts, and an un-includable tx is never actually shed. The existing tests don't see it because `retries_and_duplicates_do_not_refresh_expiry` (:462-475) and `interrupted_sends_recover_but_stalled_inclusion_does_not_retry` (:479-492) never re-`admit` the reaped hash, and `missing_inclusion_requeues_the_retained_payload` (txn_pool.rs:1152) stops before expiry.

Suggestion: expiry should also evict the pooled copy. `self.0.remove_transactions` is already used at txn_pool.rs:396.

```rust
// forward_pending.rs
pub(crate) fn ready(&mut self, now: Instant, limit: usize) -> (Vec<(Hash, Vec<u8>)>, Vec<Hash>) {
    let expired: Vec<_> = self.entries.iter().filter(|(_, e)| now >= e.expires_at).map(|(h, _)| h.clone()).collect();
    expired.iter().for_each(|hash| self.remove(hash));
    let ready = /* unchanged */;
    (ready, expired)
}
```

```rust
// txn_pool.rs::retry_forwarded
let (ready, expired) = self.pending_forwards().ready(now, limit);
if !expired.is_empty() {
    ForwarderMetrics::record_txns_expired(u64::try_from(expired.len()).unwrap_or(u64::MAX)); // see the counter comment in txn_pool.rs
    warn!(target: "worker::forward", expired = expired.len(), "forward inclusion tracking expired");
    self.0.remove_transactions(expired);
}
```

With the lease guard from the `reinserting` comment below, `requeue_one` already returns 0 when it no longer owns the entry, which closes the task-side path — the two are one change. Optionally `admit` could pre-scan for `now ≥ expires_at` entries and remove-and-skip them. An `Expired` tombstone state would also work, but it needs no pool API and adds state surface (and counts against limits); redefining expiry as "stop tracking" and fixing the docs is the other option, but I don't think that's what anyone wants.

Tests worth adding: admit → defer → 3 outputs → `retry_forwarded(now + GRACE)` → `retry_forwarded(now + LIFETIME)` ⇒ `pool.get(&hash).is_none()`; `requeue_one` with no record returns 0 and leaves the pool unchanged; `reinserting` returns false after a reap.

crates/tn-reth/src/forward_pending.rs
```diff
@@ -0,0 +1,527 @@
...
+            entry.state = PendingState::RetryDue;
+        });
+    }
+
+    /// Begin an asynchronous pool insertion without losing ownership on cancellation.
+    pub(crate) fn reinserting(&mut self, hash: &Hash, now: Instant) {
+        self.entries.get_mut(hash).into_iter().for_each(|entry| {
+            entry.state = PendingState::Reinserting(now + entry.insertion_budget);
+        });
+    }
```
grantkee: reinserting and retry_later set the entry state unconditionally, so a stale forward task's tail requeue can overwrite a lease that ready() or a newer task now owns and put an already-acknowledged transaction back in the pool.

`reinserting` (:291-295) and `retry_later` (:284-288) write `entry.state` with no guard, while the other side of the same handoff is guarded: `reinserted` (:298-313) only acts on `Reinserting(_)`, and `admit` (:163-169) only transitions from `Queued | Reinserting(_)`. The doc on `reinserted` and the comment at txn_pool.rs:566 ("A new batch may already own the inserted transaction by this point") show the concurrency was anticipated on the exit side of the insert but not the entry side. The PR description says interrupted insertions recover "without overwriting a newer admission"; that holds for `reinserted` but not for the call that precedes it.

The window isn't theoretical, because the lease always expires no later than the task's tail requeue starts:

- `forward_txns` (forward.rs:924-936) calls `admit_forwards` before `spawn_forward`. `admit_forwards` (txn_pool.rs:520-530) passes `now = Instant::now()` and `send_budget = FORWARD_BATCH_BUDGET` (120 s), so each admitted entry becomes `Sending(T_admit + 120s)` with `retry_not_before = T_admit + 1s` and `insertion_budget = 120s`.
- The task computes its own `deadline = Instant::now() + 120s` at forward.rs:614, i.e. `T_admit + δ + 120s` with δ ≥ 0.
- `map_while` (forward.rs:641-643, via `next_txn_budget` at forward.rs:958-961) ends the loop at the first pull where `now ≥ deadline`. The tail requeue (forward.rs:811-814) is a sequential `.then(requeue_one)` chain starting at `T_end ≥ deadline`; each `requeue_one` awaits `add_recovered_transaction_external` (forward.rs:842-848), so the k-th entry's `reinserting` runs at `T_end + Σ(k−1 insert latencies)`. `sleep_until(requeue_ready)` (forward.rs:838) is long past by then.
- Meanwhile `ready()` fires within ~100 ms of lease expiry: `maintain_pool` (txn_pool.rs:445-509) calls `retry_forwarded(Instant::now(), 100)` at txn_pool.rs:506 after every event, and there's a `RetryTick` every `RELOAD_RETRY_INTERVAL = 100 ms` (txn_pool.rs:134, :452-457). `ready()` (:263-273) selects `Sending(d)` when `now ≥ d && now ≥ retry_not_before`, moves up to 100 entries to `Reinserting(now + 120s)` (:276), and `retry_forwarded` (txn_pool.rs:541-571) re-inserts each into the pool then calls `reinserted(hash, true)` → `Queued`.

Only the still-`Sending` tail entries are eligible at that moment (pulled entries are `AwaitingInclusion`, removed, or already `Reinserting`/`Queued` via the in-loop `requeue_one`), so `ready()` and the stale tail requeue walk the same set in different orders — BTreeMap hash order vs batch order. Timeline for an observer whose batch overruns 120 s with a tail of N entries:

1. `T0`: admit → `Sending(T0+120s)`.
2. `T0+δ`: task A starts, `deadline = T0+δ+120s`.
3. `T0+120s ≤ t_r ≤ T0+120.1s`: tick → `ready()` moves ≤100 tail entries to `Reinserting` → pool insert → `Queued`.
4. `t_b` (tens of ms if the builder is cycling, ≤ `max_batch_delay` 1 s if idle; the builder also wakes on pending-transaction notifications, batch-builder/src/lib.rs:516-530): builder seals the `Queued` txs → `disburse_txns` (worker.rs:254-296) → `forward_txns` → `admit` → `Sending(t_b+120s)`, task B spawned, builder prunes them.
5. `t_b + j·RTT`: task B acks the j-th tx → `accepted` → `AwaitingInclusion(3)` (forward.rs:691-696).
6. `T_end + k·l` (l ≈ 1-5 ms): stale task A calls `reinserting(h_k)` and overwrites B's `Sending`/`AwaitingInclusion` with `Reinserting(now+120s)` for every `h_k` it reaches after step 4/5.

For a tail of hundreds of entries walked at comparable rates in different orders, a substantial fraction lands in step 6 after step 5. Three outcomes from there:

- After the builder's prune, the stale insert succeeds. reth's `TxPool::add_transaction` refuses only on `contains(hash)`, and the nonce it checks is `state_nonce` from the canonical DB via the validator, not the builder's optimistic `changed_accounts` nonce (batch.rs:229-240); `add_transaction` overwrites `sender_info` with the DB nonce. So `added = true` → `reinserted(h, true)` sees the stale `Reinserting` → `Queued`. The tx is back in the pool while B has already had it acknowledged; the builder repacks → `admit` (`Queued` → `Sending`) → task C forwards it, and `remaining_targets` (forward.rs:983-989) excludes the acking validator, so the duplicate goes to a different validator inside the 3-output window. `admit` (:165) and `accepted` (:188) reset `submitted_at` and the window. That's the speculative retransmission #1366 was meant to rule out.
- With a transient insert failure (sender capacity, replacement conflict, pool refusal): `added = false`, `pool.get` is `None` because of the prune → `reinserted(h, false)` → `remove` (:310). Inclusion tracking for an acknowledged transaction is dropped.
- Stale insert after B's admit but before the prune → `AlreadyImported` → `present = true` → `Queued`; the prune then removes the tx, leaving an entry in `Queued` with the tx in no pool. B's terminal action normally repairs it; if B is cancelled at an epoch boundary the entry sits in `Queued` (ignored by `ready()`, :270) until `expires_at` and the tx is silently lost.

`retry_later` (called at txn_pool.rs:553) has the same missing guard but a much smaller window — the `spawn_blocking` duration, and it needs a provider error — and the worst case is `Sending` → `RetryDue`, which self-heals on the next tick. Worth guarding for consistency. `accepted` (:181) and `defer` (:196) are likewise unguarded with a δ-sized window and a benign outcome.

No test covers the race: `interrupted_insertions_recover_without_overwriting_a_new_send` (:497-511) only exercises the `reinserted` guard, and `forward_txns_abandons_the_rest_of_a_batch_at_the_batch_budget` (forward.rs:2030-2069) runs with `requeue_pool = None`.

A `Sending(_)`-only guard on `reinserting` isn't enough on its own, since a stale task could still clobber a newer `Sending` lease. What I'd suggest is a lease generation, so a task can only take over the lease it was granted:

```rust
// forward_pending.rs
enum PendingState { Sending(Instant, u64), /* rest unchanged */ }
struct PendingTransaction<..> { /* ... */ generation: u64 }

// admit(): -> Option<(u64, Vec<Vec<u8>>)>; when transitioning to Sending:
entry.generation = entry.generation.wrapping_add(1);
entry.state = PendingState::Sending(now + send_budget, entry.generation);

/// Take over insertion only from the lease this task was granted; false if gone or owned elsewhere.
pub(crate) fn reinserting(&mut self, hash: &Hash, lease: u64, now: Instant) -> bool {
    self.entries.get_mut(hash).is_some_and(|entry| {
        matches!(entry.state, PendingState::Sending(_, g) if g == lease)
            .then(|| entry.state = PendingState::Reinserting(now + entry.insertion_budget))
            .is_some()
    })
}
/// Only the in-flight reinsertion may defer itself.
pub(crate) fn retry_later(&mut self, hash: &Hash) {
    self.entries.get_mut(hash).into_iter()
        .filter(|entry| matches!(entry.state, PendingState::Reinserting(_)))
        .for_each(|entry| entry.state = PendingState::RetryDue);
}
```

```rust
// forward.rs — thread the lease admit_forwards -> spawn_forward -> requeue_one
async fn requeue_one(pool: Option<&WorkerTxPool>, ready_at: Instant, lease: u64, tx: Vec<u8>) -> usize {
    tokio::time::sleep_until(ready_at).await;
    let hash = keccak256(&tx);
    if !pool.is_some_and(|pool| pool.pending_forwards().reinserting(&hash, lease, std::time::Instant::now())) {
        return 0; // ready() or a newer task owns this entry now
    }
    /* existing insert + reinserted(&hash, present) unchanged */
}
```

A single global counter on `PendingForwards` would be a fine simplification over a per-entry one. A cheaper alternative is to reuse the `Sending` deadline `Instant` as the lease identity — no new field, and unique in practice since re-admissions of one hash are ≥100 ms apart. Dropping the task-side tail requeue and relying on `ready()` alone doesn't close it, because the in-loop requeues at forward.rs:726 and :736 can still run up to δ after lease expiry, and it changes what `requeued` means. Applying the generation check to `accepted`/`defer` as well would be consistent but isn't required.

Test shape: admit → past SEND → `ready()` → `reinserted(true)` → `admit` (new lease) → `accepted` → stale `reinserting(old_lease)` returns false and leaves `AwaitingInclusion`; `retry_later` on `AwaitingInclusion` is a no-op.

This composes with the expiry comment on `ready()` above: once `requeue_one` bails when it doesn't own the entry, it also stops re-inserting bytes for a record that expiry already reaped, so the two are one change across `reinserting`, `ready()`, and `requeue_one`.

crates/tn-reth/src/forward.rs
```diff
@@ -661,11 +685,28 @@ impl WorkerRpcForwarder {
...
+                    }
+                    ForwardOutcome::Rejected(_) | ForwardOutcome::NoEndpointReached
+                        if had_acceptance =>
+                    {
+                        // Earlier acknowledgements remain pending until inclusion or expiry.
+                        // Exhausting fresh endpoints never resets the route to a known censor.
+                        requeue_pool.as_ref().into_iter().for_each(|pool| {
+                            pool.pending_forwards().defer(&hash);
+                        });
+                    }
```
grantkee: A deferred retry gets tallied and warned as abandoned, because this arm increments none of the counters the abandoned computation at line 769 subtracts from queued.

The code I'm asking about is `let abandoned = queued.saturating_sub(delivered + rejected + unreached);` at forward.rs:769, which this PR doesn't change — commenting here because this is the new arm that broke its accounting. On main every arm tallied one of `delivered`, `rejected`, `unreached` (except the clamped-budget arm at :723-731, which is meant to fall into abandoned); this arm touches none of `delivered` (:692), `rejected` (:710), `unreached` (:733), `no_verdict`, or `requeued`, and :769 wasn't updated to know about it.

How it's reached: `had_acceptance = !accepted_by.is_empty()` (:657-661), and `accepted_by` is only ever populated by `accepted` (forward_pending.rs:181-193), so this arm fires on a retry walk — `accepted` → `AwaitingInclusion` → `committed` → `RetryDue` (forward_pending.rs:228-235) → `ready()` → `Reinserting` → `reinserted(true)` → `Queued` → builder repacks → `admit` → `Sending`. Match-arm ordering means the guard captures every `Rejected`/`NoEndpointReached` under `had_acceptance`, including the zero-dial case where `remaining_targets` (:983-989) is empty because every committee member already acknowledged. In that case the "batch forward budget elapsed" warn at :771-779 fires after zero elapsed time, and it recurs every `retry_outputs` outputs until `FORWARD_PENDING_LIFETIME`. Each one also bumps `forwarded_txns_abandoned_total` via `record_txns_abandoned` (:780-782), whose doc (metrics.rs:450-454, "left unforwarded because their batch hit `FORWARD_BATCH_BUDGET`") is false for a deferred retry.

The requeue side is consistent — `pulled` includes the index — so this is metrics and logging only. No test drives this arm through the spawned task or reads `forwarded_txns_abandoned_total`.

Suggestion:

```rust
let mut deferred = 0_usize;                                   // ~:618, beside the other tallies
ForwardOutcome::Rejected(_) | ForwardOutcome::NoEndpointReached if had_acceptance => {
    deferred += 1;
    requeue_pool.as_ref().into_iter().for_each(|pool| pool.pending_forwards().defer(&hash));
}
let abandoned = queued.saturating_sub(delivered + rejected + unreached + deferred); // :769; add `deferred` to the warn! fields
```

It'd be worth recording `deferred` too (a `ForwardDropReason::Deferred` label, or a `tn_reth.forwarded_txns_deferred_total`) so the `queued − dropped − abandoned = delivered` invariant documented at metrics.rs:332-334 stays exact. I'd avoid folding it into `unreached`, which is documented as "requeued" and isn't true here. A spawned-task retry test asserting `abandoned_total == 0` would pin it.

This and the two other metrics comments (on the admission path in `forward_txns` below, and on `retry_forwarded` in txn_pool.rs) are really one change across `forward.rs` and `metrics.rs`.

crates/tn-reth/src/forward.rs
```diff
@@ -873,14 +921,28 @@ impl TxnForwarder for WorkerRpcForwarder {
...
+                    let retained = if self.requeue_pool.is_some() {
+                        self.requeue_pool.as_ref().and_then(|pool| pool.admit_forwards(transactions))
+                    } else {
+                        Some(transactions)
+                    };
+                    if retained.is_none() {
+                        ForwarderMetrics::record_batch_shed();
+                        ForwarderMetrics::record_txns_dropped(ForwardDropReason::BatchShed, queued_total);
+                        warn!(target: "worker::forward", "cannot retain forwarding batch; leaving it pooled");
+                    }
```
grantkee: forwarded_txns_queued_total counts transactions that admit then suppresses as duplicates, and a retention refusal records the same BatchShed reason as a permit shed, so neither shows up as what it is on a dashboard.

Two related accounting gaps on this admission path.

The first is on the `admit_forwards` call at :925. `record_txns_queued(queued_total)` runs at :869-873 before any admission decision — that's by design, it's the denominator (metrics.rs:477-483) — but `admit` (forward_pending.rs:163-169) then drops `Sending`/`AwaitingInclusion`/`RetryDue`/`ObserveOnly` entries from the returned vec with no metric. The task's `queued` (:617) is the post-filter length, `forward_txns` never compares `retained.len()` to `queued`, and so a suppressed transaction lands in no terminal series. That breaks the `queued − dropped − abandoned = delivered` relation documented at metrics.rs:332-334. Every pre-admission refusal pairs `queued_total` with a drop; suppression is the only admission outcome that doesn't. It triggers on a client resubmission while the first copy is outstanding (the ordinary requeue path goes through `Queued` and isn't suppressed). `accepted_forward_is_retained_and_duplicate_admission_is_suppressed` (:2353) asserts only hits and `accepted_by`, not the counters.

The second is the refusal branch at :929-933. It emits `record_batch_shed` + `ForwardDropReason::BatchShed`, identical to the permit-exhaustion branch at :910-921, but the docs on both name permit exhaustion as the sole cause — `BatchShed` (metrics.rs:352-355) says "because every forward permit was in flight", and `record_batch_shed` (metrics.rs:440-448) says the same. `ForwardDropReason::ALL` (metrics.rs:377-383) has no label that fits a retention refusal. The operator action differs: a permit shed says raise RPC concurrency, a retention refusal says there's an acked-but-unincluded backlog or `RetentionLimits` is too small. `forward_txns_sheds_a_batch_when_every_permit_is_in_flight` (:1893) covers the permit path only.

Naming is your call, but something like:

- `ForwardDropReason::AlreadyPending` ("already_pending"), recording `queued_total − admitted.len()` on the admitted path, so suppressed duplicates subtract from the denominator like every other non-delivery.
- `ForwardDropReason::RetentionFull` ("retention_full") replacing `BatchShed` at :931.
- Add both to `ALL`, and update the `BatchShed` / `record_batch_shed` docs to say the batch-level series aggregates both refusals (or leave `forwarded_batches_shed_total` as-is and only split the transaction-level reason).

Alternatives I'd rank lower: recording `queued` post-admission (loses the denominator meaning), a standalone suppressed-duplicates series, or doc-only for the second half. Extending the :2353 duplicate-suppression test with a counter assertion and adding a `RetentionLimits::new(0, 0)` refusal test with metric assertions would cover both.

This is one metrics change together with the deferred-tally comment above and the expired-counter comment in txn_pool.rs.

crates/tn-reth/src/txn_pool.rs
```diff
@@ -483,10 +503,74 @@ impl WorkerTxPool {
...
+        )
+    }
+
+    /// Return eligible forwards to the ordinary pool after checking canonical inclusion locally.
+    /// The existing maintenance batch limit also bounds provider reads and validations per event.
+    async fn retry_forwarded(&self, now: Instant, limit: usize) {
+        let (ready, expired) = self.pending_forwards().ready(now, limit);
+        if expired > 0 {
+            ::metrics::counter!("tn_reth.forwarded_txns_expired_total")
+                .increment(u64::try_from(expired).unwrap_or(u64::MAX));
```
grantkee: forwarded_txns_expired_total is the only forwarder counter not pre-registered at zero in ForwarderMetrics::init, and the only one emitted through an inline string counter! instead of a constant and a record_ helper.

This is the sole occurrence of `tn_reth.forwarded_txns_expired_total` in the crate. Every other forwarder series — `FORWARDED_BATCHES_SHED`, `FORWARDED_TXNS_ABANDONED`, `FORWARD_ENDPOINTS_DEMOTED`, `FORWARDED_TXNS_REQUEUED`, `FORWARDED_TXNS_QUEUED`, the `FORWARDED_TXNS_DROPPED` reasons — is a `const` in metrics.rs:320-326, registered at zero by `init()` (metrics.rs:428-438), emitted through a `record_*` helper, and asserted by `test_forwarder_metrics_register_at_zero_and_count_separately` (metrics.rs:738). The `init()` doc explains why that matters: a series that only appears on first increment can't be told apart from a broken exporter, and a `rate()` over a series whose first sample is already nonzero renders no step, so a one-shot expiry burst on an otherwise quiet node stays invisible.

Suggestion: a `FORWARDED_TXNS_EXPIRED` const, zero-registration in `init()`, a `record_txns_expired(count: u64)` helper, and a line in the register-at-zero test. With that in place this call becomes `ForwarderMetrics::record_txns_expired(...)`, which is also what the expiry-eviction sketch in the `ready()` comment on forward_pending.rs assumes.

One metrics change together with the deferred-tally and admission-path comments in forward.rs.

PR #1382, opened 2026-09-15
perf(engine): compact cumulative trie overlays geometrically

### LINKED ISSUE #1381: perf(engine): avoid quadratic cumulative trie-overlay merges
## Summary

Each block in a multi-block consensus output merges its sorted state and trie deltas into an accumulated `OutputTrieOverlay`. For distinct keys whose sorted ranges interleave, these merges repeatedly move the accumulated entries into new vectors. With `N` blocks adding `M` entries each to a merged vector, this retains `O(N^2 * M)` total merge work.

Verified against Telcoin Network [`08a507835011675a32348ac60d0488bbba891552`](https://github.com/Telcoin-Association/telcoin-network/commit/08a507835011675a32348ac60d0488bbba891552), current `main` on September 15, 2026, and its [locked Reth v1.11.3 revision](https://github.com/Telcoin-Association/telcoin-network/blob/08a507835011675a32348ac60d0488bbba891552/Cargo.lock#L7481), `d6324d63e27ef6b7c49cdc9b1977c1b808234c7b`.

Related: #1301 and #1276. The current block-root path uses `OverlayRootStateProvider` to bypass the earlier `MemoryOverlayStateProvider::trie_input()` rebuild. This issue covers the remaining cost of updating the accumulated sorted overlay.

## Current execution path

1. The engine creates [one `OutputTrieOverlay` per consensus output](https://github.com/Telcoin-Association/telcoin-network/blob/08a507835011675a32348ac60d0488bbba891552/crates/engine/src/payload_builder.rs#L100-L112). [Persistence happens after the block-building loop](https://github.com/Telcoin-Association/telcoin-network/blob/08a507835011675a32348ac60d0488bbba891552/crates/engine/src/payload_builder.rs#L288-L300).
2. Each block finishes through the custom root provider, then calls [`overlay.extend_from_block(&computed_trie_data)`](https://github.com/Telcoin-Association/telcoin-network/blob/08a507835011675a32348ac60d0488bbba891552/crates/tn-reth/src/env/execution.rs#L232-L256).
3. [`extend_from_block`](https://github.com/Telcoin-Association/telcoin-network/blob/08a507835011675a32348ac60d0488bbba891552/crates/tn-reth/src/env/output_overlay.rs#L64-L75) calls `extend_ref_and_sort` on both accumulated structures. Unique ownership avoids an `Arc::make_mut` deep copy, but the sorted merge still processes existing entries.
4. Reth routes [account-state merges](https://github.com/paradigmxyz/reth/blob/d6324d63e27ef6b7c49cdc9b1977c1b808234c7b/crates/trie/common/src/hashed_state.rs#L622-L637) and [account-trie-node merges](https://github.com/paradigmxyz/reth/blob/d6324d63e27ef6b7c49cdc9b1977c1b808234c7b/crates/trie/common/src/updates.rs#L603-L622) through [`extend_sorted_vec`](https://github.com/paradigmxyz/reth/blob/d6324d63e27ef6b7c49cdc9b1977c1b808234c7b/crates/trie/common/src/utils.rs#L31-L79). On its general path, this helper takes the old vector, allocates a new vector with capacity `left.len() + other.len()`, and moves or clones the merged entries into it.

## Why the cost remains quadratic

The helper has an append fast path when all incoming keys sort after the existing maximum. Distinct keys alone do not guarantee that ordering. Interleaving ranges take the general merge path.

For one merged vector, assume each block contributes `M` distinct keys and each subsequent batch's range overlaps the accumulated range. Block `i` writes `i * M` entries into the resulting vector. Across `N` blocks, the total is:

```text
M + 2M + ... + NM = M * N * (N + 1) / 2
```

As a small ordering example, incoming batches `[1, 4]`, `[2, 5]`, and `[3, 6]` produce vectors of lengths 2, 4, and 6. The second and third batches both require the general merge. This illustrates the merge algorithm, rather than an end-to-end transaction reproducer.

The existing [execution documentation explicitly acknowledges the remaining quadratic bound](https://github.com/Telcoin-Association/telcoin-network/blob/08a507835011675a32348ac60d0488bbba891552/crates/tn-reth/src/env/execution.rs#L123-L130). Its statement that the merge has "no allocation churn" should also be corrected: the pinned helper allocates a replacement vector on each general merge.

## Impact

Large consensus outputs with growing sets of distinct keys can consume quadratic merge work and cumulative allocation volume, delaying execution and durable finalization. For the vector described above, peak live storage remains linear in accumulated entries; the quadratic quantity is work and allocation volume over the output.

This is a source-based complexity finding. No fresh runtime benchmark or production denial-of-service impact is claimed.

## Suggested direction and validation

Consider geometrically compacted sorted runs or another representation that incorporates new deltas without merging the full accumulated vector after every block. Preserve latest-write precedence, account-deletion tombstones, storage wipes, and the existing output-level persistence and rollback behavior.

Validation should cover:

- Fixed per-block delta size with increasing block counts, using distinct interleaving keys, repeated keys, and append-only ranges as separate workloads.
- Entries moved and allocation volume alongside elapsed time, to expose the remaining full-vector merge even when constants improve.
- State-root and trie-update equivalence against the existing [layered-overlay and unpersisted-chain oracle tests](https://github.com/Telcoin-Association/telcoin-network/blob/08a507835011675a32348ac60d0488bbba891552/crates/tn-reth/tests/it/trie_overlay.rs#L116-L164), including deletion and storage-wipe cases.

MavenRain:
Closes #1381.

## Problem

Each block can merge every accumulated account and trie-node entry into replacement vectors. Honest outputs with growing, interleaving hashed keys therefore incur quadratic cumulative merge work; no attacker is required.

## Changes

- **Geometric compaction:** retain chronological sorted runs and merge adjacent runs only when they represent the same number of blocks. Cumulative compaction work is `O(N M log N)` for `N` blocks with at most `M` delta entries.
- **Layered cursors:** read at most `O(log N)` runs directly, using statically dispatched recursive cursor wrappers around Reth's existing deletion, wipe, reset, and address-switch logic. Newer writes take precedence and block-owned `Arc`s remain immutable.
- **Regression coverage:** add three unit tests and extend the existing database-backed root/update oracle to 65 blocks. Output-level persistence and rollback behavior retain their existing paths.
- **Measurements:** add `cargo +1.94 bench --locked -p tn-reth --bench output_overlay` for interleaving, repeated, and append-only keys. Correct the execution documentation's allocation and complexity claims.

## Measurements

Local Rust 1.94, unoptimized build, 32 account entries and 32 trie-node tombstones per block. The standalone harness compiles the benchmark and run implementation directly from this checkout. Input preparation is excluded.

| Workload | Blocks | Allocated MB, before → after | Elapsed ms, before → after |
| --- | ---: | ---: | ---: |
| Interleaving | 64 | 13.31 → 2.67 | 284.7 → 7.6 |
| Interleaving | 256 | 210.53 → 13.95 | 522.5 → 28.8 |
| Interleaving | 1,024 | 3,358.72 → 68.89 | 16,237.4 → 132.7 |
| Repeated | 1,024 | 13.10 → 16.45 | 14.23 → 14.21 |
| Append-only | 1,024 | 13.10 → 68.89 | 3.40 → 15.82 |

Allocation totals count requested allocation/reallocation bytes and reproduced across two runs. Timings are single-run observations on a busy host. These measurements cover overlay accumulation.

For the 1,024-block interleaving fixture, the two sorted vectors emit 33,587,200 merge-output entries before and 655,360 after, derived from the fixture sizes and compaction schedule. Allocation totals additionally include copy-on-write clones and run metadata.

The tradeoff is extra allocation for repeated and append-only keys, plus duplicate keys retained across separate runs. Root-walk and end-to-end execution performance still need measurement.

## Testing

- `cargo +nightly-2026-03-20 fmt --all -- --check`: passed.
- `cargo +1.94 test --offline --locked --manifest-path /Users/oobi/Documents/gpt13/tn1381-harness/Cargo.toml --lib -j 2`: three tests passed, using the production modules and repository dependency lock.
- `cargo +1.94 clippy --offline --locked --manifest-path /Users/oobi/Documents/gpt13/tn1381-harness/Cargo.toml --tests --bins --no-deps -j 2 -- -D warnings`: passed for the focused harness.
- Each new test failed under an in-diff mutation: merging all runs every block, reversing compaction precedence, and replacing address propagation with reset. Restored tests passed.

Draft pending the full attested run. The current PR workflow skips build/test lanes for maintainer authors, including MavenRain.

The [PR workflow](https://github.com/Telcoin-Association/telcoin-network/actions/runs/34976641605) ran on `de5e5da18f7438d1dc38d8a8a5d10f965e8ab253`. Build/test lanes were skipped under the maintainer policy, and `verify-on-chain` failed because this commit has no attestation. The job requires `make attest` on this commit, followed by a review request to rerun verification.


sstanfield (approved):

crates/tn-reth/src/env/output_overlay.rs
```diff
@@ -93,12 +96,9 @@ impl OutputTrieOverlay {
...
-            InMemoryTrieCursorFactory::new(DatabaseTrieCursorFactory::new(tx), self.nodes.as_ref()),
+            RunCursorFactory::new(DatabaseTrieCursorFactory::new(tx), &self.runs),
             HashedPostStateCursorFactory::new(
-                HashedPostStateCursorFactory::new(
-                    DatabaseHashedCursorFactory::new(tx),
-                    self.state.as_ref(),
-                ),
+                RunCursorFactory::new(DatabaseHashedCursorFactory::new(tx), &self.runs),
                 &current_sorted,
             ),
```
grantkee: The layered run stack answers is_storage_empty as false where the flat newest-wins merge answers true, so when an older run holds a non-zero slot that a newer run zeroes, the emitted TrieUpdates diverge from reth's even though the state root matches.

Anchoring here because this is where the whole hashed-cursor stack is composed and, as far as I can tell, the only place a fix can go. The delegation in `cursors.rs:140-146` is faithful to its inner cursors and isn't where the problem is.

The mechanism is in reth's own `HashedPostStateCursor::is_storage_empty` (v1.11.3, `crates/trie/trie/src/hashed_cursor/post_state.rs:351-359`): it does `if self.post_state_cursor.has_any(|(_, v)| !v.is_zero()) { return Ok(false) }` and only otherwise delegates inward. `has_any` scans that layer's entire slot slice and is blind to shadowing by outer layers. So an older run holding `(S, 5)` still answers "not empty" even though a newer run zeroed `S`. Pre-PR, there was one merged layer so this couldn't happen; now each run is its own layer and the question gets asked of every one of them.

Direct cursor probe, empty (Noop) database, runs `{A: [(S,5)]}` / empty filler / `{A: [(S,0)]}`:

```
[a] flat merged slots for A: Some((false, [(0x…01, 0)]))
[a] LAYERED is_storage_empty = false
[a] FLAT    is_storage_empty = true
[a] LAYERED seek(ZERO) = None      [a] FLAT seek(ZERO) = None
[a] DIVERGES = true
```

`seek` agrees; only `is_storage_empty` diverges. Downstream, `StorageRoot::calculate` (reth `trie.rs:618-629`) short-circuits on `true` and returns `StorageTrieUpdates::deleted()`, while `false` performs the full walk. At the root/updates level:

```
--- block 1 --- root equal = true   updates equal = false
  layered root = 0x764f1e7b…5644     oracle root = 0x764f1e7b…5644
  layered storage_tries[contract] = None
  oracle  storage_tries[contract] = Some(StorageTrieUpdates { is_deleted: true, … })
```

With multiple slots the layered side emits `removed_nodes` explicitly where the oracle emits the `is_deleted` marker. The integration guard test in this PR fails when a fixture produces the pattern — adding a "zero every slot of `c1`" block to `seeded_deltas` yields:

```
assertion `left == right` failed: trie updates diverged at block 8
```

with the state-root assertion at `trie_overlay.rs:151` still passing. The whole diff is one extra `StorageTrieUpdates { is_deleted: true, storage_nodes: {}, removed_nodes: {} }` entry on the oracle side.

All four of these have to hold for one account at one block:

1. The outermost hashed layer (current block's delta) has only zero-valued slots for the account, no wipe.
2. Some inner run holds a non-zero value for a slot a newer layer zeroed, with no wipe above it.
3. The flattened newest-wins view has no non-zero slot.
4. The DB's hashed storage for the account is empty, or some layer wipes it. (Any DB row at all makes reth's flat path return `false` too, so no divergence — this is what contains it.)

Plus a compaction interaction: the non-zero write and the zeroing write must land in different runs. Two adjacent level-0 blocks compact immediately in `SortedTrieRuns::extend`, which flattens the hazard away.

What it does and doesn't affect, traced rather than assumed:

- State root: identical in every case tested, including the 65-block integration sequence. The header commitment is unaffected.
- Persisted database: identical. `write_storage_trie_updates_sorted` (reth `trie/db/src/trie_cursor.rs:124-131`) turns `is_deleted` into `delete_current_duplicates()`; the layered path instead emits explicit `removed_nodes` covering exactly the nodes it created, and those removals cover everything the wholesale delete would have. Checked with a simulated writer in both the empty-DB and populated-DB-with-wipe shapes.
- Consensus: not a fork vector. All validators run identical code and the root matches the pre-PR/reth value regardless.
- What it actually is: a violation of the exact-equality invariant the module doc advertises ("must produce EXACTLY the root and trie updates reth itself computes"), plus latent test fragility.

On the test side: `src/env/output_overlay/tests.rs:80` already contains `(key(1), U256::ZERO)`. The test passes only because `base` (line 107) seeds the same contract with `(key(9), U256::from(99))`, which is never zeroed, so `has_any` is true in the flat merge and the oracle never short-circuits. Remove that masking seed and `run_roots_match_flattened_oracle_across_compaction_boundaries` fails at block 1. The integration fixture can't produce a zero today only because values come from `U256::from(rng.next_u64())` with fixed seed `0x1301` (`trie_overlay.rs:63`) — probability 2^-64.

On fixing it: the obvious fix doesn't work, and I tried it. Implementing `is_storage_empty` as a `seek(B256::ZERO)?.is_none()` probe inside `RunCursorFactory` still fails, because in `layered_root_with_updates` the outermost hashed layer is reth's own `HashedPostStateCursorFactory` wrapping the current block's delta — it sits above `RunCursorFactory` and starts the faulty descent where a fix inside `RunCursorFactory` can't see it. A seek-based probe is also not oracle-equivalent (reth returns `false` whenever the DB has any row, even one the overlay zeroes).

The fix has to live here, where the stack is composed, and replicate reth's exact predicate:

```text
is_storage_empty(A) =
    if  merged_newest_wins_overlay(A) has any non-zero slot   -> false
    else if any layer (current or run) wipes A                -> true
    else                                                      -> raw_db_cursor.is_storage_empty()
```

Concretely: wrap the composed hashed-cursor factory in an adapter holding (a) the full layered factory, (b) `&SortedTrieRuns`, (c) `&current_sorted`, and (d) the raw `DatabaseHashedCursorFactory`. Delegate `seek`/`next`/`reset`/`set_hashed_address` to the layered cursor unchanged; implement `is_storage_empty` as a newest-to-oldest scan over `once(current).chain(runs.iter().rev())`, tracking seen slot keys so a newer `ZERO` shadows an older non-zero, stopping at the first wiped layer, and falling through to a raw DB storage cursor when the overlay contributes nothing non-zero. Cost is O(overlay slots for the account), once per account per block — the same order as reth's own `has_any`.

A cheaper but weaker alternative: fold the current block's `HashedPostStateSorted` into the run stack as the newest layer so the entire hashed stack is `RunCursorFactory`-owned, then implement the predicate once inside `RunCursorFactory`.

Either way, I'd want a regression case that actually bites — either drop the masking `(key(9), U256::from(99))` seed from the unit test's `base`, or add a block to `seeded_deltas` that zeroes every slot of `c1`.

Whether to fix now or document the deviation and land the regression test is your call. Nothing reachable today is broken, but the invariant is load-bearing for future work — ExEx consumers, witness/proof generation, or any merge path that relies on `is_deleted` to clear accumulated `storage_nodes` via `StorageTrieUpdates::extend_common`.

crates/tn-reth/benches/output_overlay.rs
```diff
@@ -0,0 +1,148 @@
+//! Compare cumulative allocations of the former merge path and production sorted runs.
```
grantkee, line 1: The bench target emits 60 unused_crate_dependencies warnings, which are hard errors under the RUSTFLAGS=-D warnings that CI exports and permanent noise in make check.

Exact count is 60, all attributed to the bench; the lib and both test targets are clean:

```
$ cargo check -p tn-reth --all-targets --offline --message-format short
warning: extern crate `serde` is unused in crate `output_overlay`
...
warning: `tn-reth` (bench "output_overlay") generated 60 warnings
$ ... | grep -c "^warning: extern crate"   → 60
```

The lint comes from `Cargo.toml:60-66` (`[workspace.lints.rust] unused_crate_dependencies = "warn"`), inherited at `crates/tn-reth/Cargo.toml:120-121`. `-D warnings` is set in four places with the identical string `RUSTFLAGS="-D warnings -D unused_extern_crates"`: `etc/ci-lanes.sh:49`, `etc/test-and-attest.sh:64`, `.github/workflows/pr.yaml:70`, `.github/workflows/cache-deps.yaml:68`. That it's a hard error and not just noise:

```
$ cargo rustc -p tn-reth --bench output_overlay --offline -- -D warnings
error: could not compile `tn-reth` (bench "output_overlay") due to 60 previous errors
```

No CI lane builds benches today, so this doesn't fail the merge queue — but `make check` (`Makefile:109-110`) uses `--all-targets` without exporting `RUSTFLAGS`, so the documented local check now warns 60 lines and exits 0 every time.

There are two house precedents for this: `crates/tn-reth/tests/it/main.rs:3` and `crates/tn-reth/tests/reth_defaults.rs:9-10`, the second of which carries the rationale ("An integration test binary links every dependency of its crate but uses only a few."). An inner `#![allow(unused_crate_dependencies)]` at the bench crate root is sufficient — the lint is the sole source of all 60, and the inner attribute also covers the `#[path]`-included `sorted_runs` module. (The `-D unused_extern_crates` half of `RUSTFLAGS` is a different lint that fires on explicit `extern crate` items; it isn't what produces these and the allow doesn't disturb it.)

```suggestion
//! Compare cumulative allocations of the former merge path and production sorted runs.

// A bench binary links every dependency and dev-dependency of its crate but uses only a few.
#![allow(unused_crate_dependencies)]
```

crates/tn-reth/benches/output_overlay.rs
```diff
@@ -0,0 +1,148 @@
...
+        atomic::{AtomicBool, AtomicUsize, Ordering},
+        Arc,
+    },
+    time::Instant,
+};
+use tn_types::{B256, U256};
+
+#[path = "../src/env/output_overlay/sorted_runs.rs"]
+mod sorted_runs;
+use sorted_runs::SortedTrieRuns;
```
grantkee: The #[path] include compiles sorted_runs.rs a second time at the bench crate root, which only works while the file references nothing through crate:: or super:: and nothing in CI would catch it breaking.

This declares `#[path = "../src/env/output_overlay/sorted_runs.rs"] mod sorted_runs;` while `src/env/output_overlay.rs:41` declares the same file as `mod sorted_runs;`. The file is compiled twice: once at `crate::env::output_overlay::sorted_runs`, once at the bench crate root.

It compiles today because of a property nothing enforces. The complete path inventory of `sorted_runs.rs` (81 lines) is `reth_trie::{updates::TrieUpdatesSorted, HashedPostStateSorted}` (line 3), `std::sync::Arc` (line 4), `Arc::make_mut` (lines 34-35), and file-local `RunLevel`/`TrieRun`/`SortedTrieRuns`/`Self`. Zero `crate::`, zero `super::` in path position. The six `pub(super)` items (lines 12, 23, 28, 47, 54, 63) degrade to crate-visible because the module sits at the bench crate root, where `super` is the crate root. Any of these breaks the bench and no CI lane builds benches: adding a `use crate::` or `use super::`; tightening any `pub(super)` to private; moving the file deeper and relying on `pub(super)` reaching a grandparent (there is none at the bench root); adding a `#[cfg(test)] mod tests` with crate-internal imports. The property is already under pressure — `cursors.rs:7` does `use super::sorted_runs::SortedTrieRuns;`, so the sibling module already reaches across in the way this one can't.

On the baseline at lines 123-129: the reimplementation is faithful — it matches `extend_from_block` at the merge base (`git show 08a50783:crates/tn-reth/src/env/output_overlay.rs`, lines 72-75) call-for-call, same two `Arc::make_mut(...).extend_ref_and_sort(...)` in the same order. And calling the shipped code wasn't an option, since this same commit deletes it; a before/after micro-benchmark on the branch that replaces the "before" can't call it short of a git-history dependency. So I read that as a design decision with a documentation gap rather than a defect. What's missing is that nothing pins the baseline: if `extend_ref_and_sort` semantics shift under a reth bump, the reported speedup silently becomes fiction.

Two independent things I'd suggest. For the include, the least invasive option is to make the coupling explicit at the source:

```diff
--- a/crates/tn-reth/src/env/output_overlay/sorted_runs.rs
+++ b/crates/tn-reth/src/env/output_overlay/sorted_runs.rs
@@ -1,4 +1,9 @@
 //! Chronological sorted runs compacted like carries in a binary counter.
+//!
+//! `benches/output_overlay.rs` includes this file a second time via `#[path]`, as a direct
+//! child of that bench crate's root. That compiles only while this file references nothing
+//! through `crate::` or `super::` and its `pub(super)` items stay `pub(super)` (at the bench
+//! root, `super` IS the crate root). No CI lane builds benches, so breaking either of those
+//! is caught by `make check`, not by the merge queue.
```

If the bench is meant to outlive the PR, the structurally sounder option is to drop `#[path]`, add a `bench-internals = []` feature to `crates/tn-reth/Cargo.toml` gating `pub mod sorted_runs`, and have the bench reach the type through `tn_reth::` — which also retires the `tn_reth is unused` entry among the 60 warnings in the other thread.

For the baseline, one comment so the next reader knows it's a snapshot:

```diff
+                // Snapshot of the pre-#1381 `OutputTrieOverlay::extend_from_block`
+                // (two `Arc::make_mut` + `extend_ref_and_sort` calls per block). The
+                // original was deleted by this commit, so it cannot be called; see
+                // `git show 08a50783:crates/tn-reth/src/env/output_overlay.rs`.
                 let old = measure(|| {
```

grantkee (dismissed):
A few nits, but nothing blocking. LGTM @MavenRain !

crates/tn-reth/src/env/output_overlay.rs
MavenRain: Fixed in 9790112. The adapter wraps the complete cursor stack, including the current block, and preserves newer zeros, wipes, and the raw database fallback. Four focused tests pass, including the unmasked root/update oracle and the new cursor matrix; mutations of zero shadowing and the database fallback both fail.

crates/tn-reth/benches/output_overlay.rs
MavenRain: Added the documented `#![allow(unused_crate_dependencies)]` at the benchmark crate root in 9790112, following the existing integration-test precedent.

crates/tn-reth/benches/output_overlay.rs
MavenRain: Adopted the feature-gated approach in 9790112: the benchmark now uses `tn_reth::OutputTrieOverlay::extend_sorted`, with `bench-internals` required. This removes the `#[path]` include and keeps the run type private. The baseline comment now names `08a50783` and Reth v1.11.3.

grantkee (approved):
LGTM!

PR #1385, opened 2026-09-15
fix(node): derive node role from committee membership

### LINKED ISSUE #1350: Seed the node-mode watch from the epoch record chain instead of defaulting to CvvActive

### LINKED ISSUE #1355: Derive node role from committee membership; deprecate --observer
Depends on #1350, and should not merge before it. That issue seeds the node-mode watch from the
epoch record chain; without it, deleting the override here leaves a genuine observer reporting
itself as an active validator until the epoch loop runs.

## Problem

`--observer` is read in two places to force a node out of the committee role:

- a construction-time override that writes `NodeMode::Observer` straight into the watch,
  `crates/node/src/manager/node.rs:686-689`
- a disjunct in `identify_node_mode`, `crates/node/src/manager/node/start_epoch.rs:957`

Both exist so an operator can run a node that follows consensus without participating. For a node
whose key is not in the committee, the flag changes nothing: `identify_node_mode` already resolves
`Observer` from `!in_committee` alone.

For a node whose key *is* in the committee, the flag appears to offer a warm standby: a validator
that is seated on chain but deliberately not proposing. That capability does not actually exist.

## The flag is already half-broken on a seated node

`identify_node_mode` returns `Observer`, so the primary does not propose. But the worker's
`QuorumWaiter` is built purely from committee membership, not node mode. `new_worker_internal`
constructs it from `consensus_config.authority()` (`crates/consensus/worker/src/worker.rs:92`), and
that authority comes from `committee.authority_by_key(&primary_public_key)`
(`crates/config/src/consensus.rs:203`). A seated node gets a quorum waiter regardless of the flag,
so the worker keeps building and sealing batches.

Meanwhile the same node unsubscribes from the topic those batches are published on.
`should_subscribe_batch_topic` returns false for `Observer` (`start_epoch.rs:1116-1121`), and
`spawn_worker_network_for_epoch` acts on it at `start_epoch.rs:897-898`.

So a seated node run with `--observer` today builds batches and is unsubscribed from the gossip
topic that carries them. That is not a warm standby. It is a validator with one half of its worker
path disconnected from the other.

## Proposed change

Delete the construction-time override at `node.rs:686-689` and the
`|| self.builder.tn_config.observer` disjunct at `start_epoch.rs:957`. Node role then follows
committee membership alone, which is what `identify_node_mode` already computes and documents
(`start_epoch.rs:925-930`).

Keep the CLI argument (`crates/telcoin-network-cli/src/node.rs:58-61`) as a hidden no-op that logs a
deprecation warning naming this issue. Documented observer command lines keep working on the live
testnet, which is past epoch 407 on chain 2017 (`crates/types/src/forks.rs:180`). Deleting the
argument outright would make every one of those command lines fail to parse.

Update the flag's help text and the docs that advertise it (`README.md`, `SYNC.md`,
`crates/telcoin-network-cli/README.md`, `etc/local-testnet.sh`) to say that role is derived from
committee membership.

## What operators lose

An operator running a seated validator with `--observer` as a non-signing brake loses a capability
they believed they had. Per the section above, they did not have it: the worker half kept running.

The supported way to take a validator out of consensus is to exit it on chain. Say so in the
deprecation warning, so an operator who hits it knows what to do instead.

## Risk

Medium, and concentrated on the live testnet. The check to run before merge: a node whose key is
absent from the committee must still resolve to `Observer` with the flag gone, both at genesis and
at a later epoch. That is the case every current observer deployment is in.

## Acceptance criteria

- `node.rs:686-689` and the disjunct at `start_epoch.rs:957` are gone.
- A non-committee node resolves to `Observer` without the flag.
- `--observer` still parses, logs a deprecation warning naming this issue, and changes no behaviour.
- Docs and the local testnet script no longer present the flag as a way to control node role.
- The observer end-to-end tests pass unchanged.

## Related

A follow-up issue deletes the flag and its plumbing entirely, one release later.

MavenRain:
Closes #1355.

## Problem

No attacker is required: starting a seated validator with `--observer` forces its primary into observer mode while its worker still follows committee membership. The worker can keep producing batches while unsubscribed from their gossip topic, so the flag does not provide a working non-signing standby.

The startup-mode prerequisite is already merged in #1362, which closes #1350.

## Changes

- **Derive role from membership:** remove both observer overrides. Preserve startup seeding from the latest epoch record and the epoch loop's sticky `CvvInactive` catch-up state.
- **Preserve command compatibility:** keep `--observer` as a hidden, ignored argument. Warn with #1355 and tell operators to exit a validator on chain to take it out of consensus. Remove the flag's tracing-name override.
- **Update operator guidance:** remove the flag from launch examples and the local testnet script, and explain membership-based roles in both READMEs and `SYNC.md`.
- **Cover CLI compatibility:** add a regression test for legacy parsing and omission from both help views. Existing observer end-to-end tests are unchanged.

## Testing

- Passed: `cargo +nightly-2026-03-20 fmt --all -- --check` through `gateledger`, `bash -n etc/local-testnet.sh`, and `git diff --check`.
- Ran `diffclass --base origin/main`, `apidelta --base origin/main --verb clippy`, and `buildplan clippy --base origin/main --explain`. They require full workspace validation.
- The exact observer argument and new regression-test body passed in an isolated Clap 4.6.0 probe under Rust 1.94. Changing the added `hide = true` to `hide = false` failed the help assertion; restoring it passed again. This probe does not compile the full Telcoin CLI.
- The full-crate attempt, `cargo +1.94 test --locked -p telcoin-network-cli --lib deprecated_observer_flag_parses_but_is_hidden`, could not proceed offline because `aead 0.5.2` was not cached.
- **Outstanding:** the workflow deliberately skips build and test lanes for maintainer PRs. [Run 35023912255](https://github.com/Telcoin-Association/telcoin-network/actions/runs/35023912255) failed `verify-on-chain` because commit `86922e774f7433d0259bc3cfe7d8fed147d66a74` is not attested. Full workspace clippy, both nextest lanes, and observer end-to-end coverage remain unverified until `make attest` succeeds on that commit, followed by rerunning the attestation check.


etc/local-testnet.sh
```diff
@@ -180,8 +180,8 @@ if [ "$START" = true ]; then
     DATADIR="${ROOTDIR}/observer"
     CONSENSUS_METRICS="127.0.0.1:9104"
     echo "Starting Observer in background, rpc endpoint http://localhost:8541"
```
sstanfield, line 182: Thanks for not letting this bit rot.

sstanfield (dismissed):
LGTM

sstanfield (approved):

grantkee (approved):
LGTM too!

PR #1386, opened 2026-09-15
fix(txpool): move canonical update preparation off async threads

### LINKED ISSUE #1378: tn-reth: update_canonical_state copies the full SealedBlock body and process_canon_state_update still collects accounts/hashes on the runtime thread
## Problem

`WorkerTxPool::update_canonical_state` (after #1344) moves the pool update onto the blocking pool, but it gets the tip there by cloning the whole `SealedBlock` on the calling async thread first, and the caller on the maintenance path still does its O(accounts + txs) collection before the hop. So the runtime-thread work #1339 targets is only partly moved.

Line numbers are against #1344 at `1396873`.

### 1. The body copy is wasted

```rust
// crates/tn-reth/src/txn_pool.rs:561-563
let pool = self.clone();
let new_tip = new_tip.clone();
tokio::task::spawn_blocking(move || {
```

`tn_types::SealedBlock` is `reth_primitives_traits::SealedBlock<Block<TransactionSigned>>` = `{ header: SealedHeader, body: BlockBody { transactions: Vec<TransactionSigned>, ommers, withdrawals } }`, `#[derive(Clone)]`. The clone deep-copies every `Signed<Tx>` envelope (~300 B each plus a refcount bump on `input: Bytes`, plus a deep copy of any access list) and the ~700 B header.

Reth reads only header fields from `CanonicalStateUpdate::new_tip` (v1.11.3): the `traits.rs` accessors are `number()`, `hash()`, `timestamp()` and `gas_limit()` via `block_info()`; `Pool::on_canonical_state_change` (`pool/mod.rs:508-532`) calls `update.block_info()` and `validator.on_new_head_block(new_tip)`, which narrows to `new_tip_block.header()` (`validate/eth.rs:876-878`); `TxPool::on_canonical_state_change` receives only `BlockInfo`, hashes and senders. Only the lazily-evaluated `trace!(?update)` would touch the body. Reth's own maintenance loop passes `new_tip.sealed_block()` borrowed from the `Arc<Chain>` it holds, because that whole loop runs under `spawn_blocking_task`.

Before #1344 the function took `new_tip` by reference straight into `CanonicalStateUpdate`; the clone is new.

### 2. The collection before the hop is still on the runtime thread

```rust
// crates/tn-reth/src/txn_pool.rs:598-611
let changed_accounts: Vec<ChangedAccount> = state
    .accounts_iter()
    ...
    .collect();
...
let mined_transactions: Vec<TxHash> = blocks.transaction_hashes().collect();
```

`process_canon_state_update` (`txn_pool.rs:590-621`) runs `state.accounts_iter()` (O(accounts)) and `blocks.transaction_hashes()` (O(txs)) before the `.await` at `:620`, so on the `canonical txn pool` critical task's runtime thread, on every `Commit`. This is pre-existing, but it now sits directly in front of a `spawn_blocking` that could absorb it.

### 3. Both call sites pay, and the batch builder pays twice

- The `canonical txn pool` critical task, on every `Commit` (`txn_pool.rs:614`).
- `BatchBuilder::run`, on every mined batch (`crates/batch-builder/src/lib.rs:499-504`). The builder already does a full-block clone per canonical notification at `lib.rs:443` (`self.last_canonical_update = latest.tip().sealed_block().clone()`), so a mined batch currently costs two body copies on the runtime thread.

## Impact

Runtime-thread occupancy, not correctness. At the ~1,400-tx batch bound the clone is roughly 400-500 KB of memcpy, one `Vec` allocation and ~1,400 `Bytes` refcount increments, repeated on drop: tens to low-hundreds of µs, measurable in a profile but small next to the pool update itself. Same order as the neighbouring inline collection. This is not backed by a profile; the sender bound is the analytic figure from #1303.

## Proposed solution

Two steps, the first small enough to land alone.

**Clone only the header.** No signature or call-site change; removes the body copy on both paths.

```rust
// crates/tn-reth/src/txn_pool.rs:561-574
let pool = self.clone();
// The pool reads only header fields (number/hash/timestamp/gas_limit) from the tip;
// skip the body copy. The header-only block below is synthetic (empty body) and must
// not escape this closure.
let header = new_tip.clone_sealed_header();
tokio::task::spawn_blocking(move || {
    let new_tip = SealedBlock::from_sealed_parts(header, BlockBody::default());
    let update = CanonicalStateUpdate { new_tip: &new_tip, /* as today */ };
    pool.0.on_canonical_state_change(update);
})
.await
```

`SealedBlock::clone_sealed_header`, `SealedBlock::from_sealed_parts` and `BlockBody: Default` exist; `BlockBody` is re-exported from `tn_types` but not yet imported in `txn_pool.rs`. The existing `test_canonical_update_cannot_clobber_epoch_base_fee` and `test_canonical_update_awaits_blocking_pool_work` exercise `block_info()` on this path.

**Move the collection onto the blocking pool and drop the builder's clone.** Change `update_canonical_state` to take `&SealedHeader`, factor the sync body into an `apply_canonical_update(&self, &SealedBlock, ..)` helper, and have `process_canon_state_update` move the `Arc<Chain>` into the closure and collect `changed_accounts` / `mined_transactions` there, borrowing `tip.sealed_block()` as reth does. On the builder side, `last_canonical_update` becomes `SealedHeader`, `latest_canon_block` (`lib.rs:313-320`) returns one via `reth_env.sealed_header_by_number(num)?` / `chainspec().sealed_genesis_header()`, and `lib.rs:443` becomes `latest.tip().clone_sealed_header()`. The call at `lib.rs:499-504` is textually unchanged. Tests at `txn_pool.rs` and `batch-builder/src/lib.rs:589, 621` need a `SealedHeader` (`genesis_block.clone_sealed_header()` works; the `base_fee_per_gas` assertion still compiles since `SealedHeader: Deref<Target = Header>`). `Arc<Chain>`, `SealedHeader` and `WorkerTxPool` are all `Send + Sync + 'static`.

## How this was found

Review of #1344. Two independent tn-review scans both traced every reader of `new_tip` inside reth's pool and found none that touches the body.

MavenRain:
Fixes #1378.

## Problem

Canonical pool updates still copy the full tip block and collect changed accounts and mined transaction hashes on async runtime threads. The batch builder also retains a full block for a path that only needs its header. Ordinary block and batch processing pays these costs, with no attacker required.

## Changes

- **Pool updates:** accept a `SealedHeader` and construct a temporary empty-body block inside the blocking task. The pinned Reth pool and validator only read header fields from this value.
- **Canonical maintenance:** move the shared `Arc<Chain>` into the blocking task, collect account changes and mined hashes there, and borrow the original tip block.
- **Batch builder:** retain and refresh only the canonical header, and load a header at startup with the existing genesis fallback.
- **Update semantics:** share the synchronous application path, preserve awaited completion and error propagation, and read the epoch base fee when applying the update. Adapt the existing epoch-fee and blocking-completion tests to the header input.

## Testing

- Passed: `cargo +nightly-2026-03-20 fmt --all -- --check` and `git diff --check`.
- `diffclass` requires a build; `apidelta` retains the dependency closure. `buildplan clippy --base origin/main --all-targets --explain` selects 21 affected packages.
- Passed: Clippy over all 21 affected packages with CI's default target selection, using the exact command below. Both Clippy runs used `tnquiet` and `gateledger`.
- Adding `--all-targets` checked both modified crates, then failed with 25 lint errors in the E2E integration test target. Every reported file under `crates/e2e-tests/tests/it/` is unchanged from base `f4a89ef8`; examples include `needless_question_mark` in `common.rs:396` and `ptr_arg` in `epochs.rs:336`. That expanded check is not green.

```sh
cargo +nightly-2026-03-20 clippy --locked -j 2 \
  -p telcoin-network -p telcoin-network-cli -p tn-batch-builder \
  -p tn-batch-validator -p tn-engine -p tn-executor -p tn-exex \
  -p tn-network-libp2p -p tn-node -p tn-primary -p tn-reth -p tn-rpc \
  -p tn-storage -p tn-test-utils -p tn-test-utils-committee \
  -p tn-types -p tn-worker --no-deps -- -D warnings
```

Test execution and the Rust 1.94 attestation remain pending. The PR workflow skipped its compiler and test lanes for this maintainer PR. The [attestation check](https://github.com/Telcoin-Association/telcoin-network/actions/runs/35025381026/job/104571028199) confirms that commit `dd225ebce60d48d2ef624bc6cb6d52c81491acdd` is not yet attested. A maintainer must complete `make attest` on that commit and request a review to rerun the check.


sstanfield (dismissed):

crates/tn-reth/src/txn_pool.rs
```diff
@@ -642,29 +642,52 @@ impl WorkerTxPool {
...
     ) -> Result<(), JoinError> {
         let pool = self.clone();
         let new_tip = new_tip.clone();
         tokio::task::spawn_blocking(move || {
-            let update = CanonicalStateUpdate {
-                new_tip: &new_tip,
-                pending_block_base_fee: pool.2.base_fee(),
+            // Reth's pool and validator only read header fields from the tip. This synthetic
+            // block satisfies that API without copying a body and stays inside this closure.
+            let new_tip = SealedBlock::from_sealed_parts(new_tip, BlockBody::default());
```
grantkee, line 756: This is not what I would have expected, but confirmed accurate for this version of reth. This might change in a future upgrade, so might be worth a regression test so maintainers don't forget about this observer optimization

grantkee (dismissed):
LGTM - one suggestion about future proofing reth assumptions

crates/tn-reth/src/txn_pool.rs
MavenRain: I've filed an immediate follow-up for this: https://github.com/Telcoin-Association/telcoin-network/issues/1420

MavenRain: This got kicked from the merge queue, so I'll likely satisfy #1420 on this branch.

grantkee (approved):
LGTM!

Issue #1358, opened 2026-09-09
End-to-end coverage for cold-genesis startup


grantkee:
Depends on #1356. This is the coverage that exercises that issue's gate.

## Problem

Nothing in the end-to-end suite starts a node alone, with no peers, and requires it to survive.
That is the shape that drives the startup gate's zero-peer short-circuit and the network readiness
wait at `crates/node/src/manager/node/start_epoch.rs:1037-1057`. Both are timeouts, and a timeout
with no test is a timeout nobody notices until it fires in production.

Two scenarios that an earlier version of this work proposed are already covered on main and should
not be rebuilt:

- a fresh joiner catching up is `test_observer_late_join_catchup`
  (`crates/e2e-tests/tests/it/restarts.rs:613`)
- a full-cohort simultaneous restart is phase 2 of `do_restarts` (`restarts.rs:436-452`), which
  kills all four nodes and restarts them together

## Proposed test

One new scenario: cold genesis, alone.

Start a single validator from genesis with no other node running. It must:

1. stay alive through the network readiness wait, which is bounded at 240 attempts of 500ms
   (`start_epoch.rs:1043-1049`)
2. stay alive through the startup gate's own deadline
3. serve RPC while alone
4. become an active validator once the rest of the cohort starts and consensus forms

Step 3 is the part that distinguishes a passing gate from a hung one. A node that blocks on either
timeout without serving RPC has failed even if it eventually recovers.

Use `wait_until_blocking` (`crates/test-utils/src/wait.rs:72`) for the deadline loops, matching the
other tests in that file. `start_validator` in `crates/e2e-tests/tests/it/common.rs` is the spawn
helper.

## Name it into the scheduled lane

The test name must contain `test_epoch`. The `test-epochs` make target filters on that substring
(`Makefile:200-201`), and the scheduled Durable end-to-end lane runs that target
(`.github/workflows/durable-e2e.yaml:91-94`). A name outside the filter means the scheduled lane
never runs it, which for a test about startup timeouts under the slower Durable MDBX regime is
exactly the run that matters.

## Risk

Low. It adds a test, changes no production code. The thing to watch is flakiness: the scenario is
built on two timeouts, so the assertions should be about liveness and eventual promotion, not about
how long either phase takes.

## Acceptance criteria

- One new end-to-end test covering cold genesis with no peers.
- Its name matches the `test_epoch` filter.
- It asserts the node serves RPC while alone, and becomes an active validator after the cohort
[...]

PR #1387, opened 2026-09-15
fix(node): sync epoch records before the first epoch

### LINKED ISSUE #1356: Sync the epoch record chain to the network tip before the epoch loop starts
Depends on #1351 (network bring-up must already be in the pre-loop startup sequence) and #1352 (the
zero-peer short-circuit needs a count of established peers, not dialing ones).

## Problem

A node enters its first epoch without knowing how far behind the epoch record chain it is. The
record collector that closes that gap, `collect_epoch_records`
(`crates/state-sync/src/epoch.rs:18`), runs as a background task spawned at
`crates/node/src/manager/node.rs:858` and is armed by a watch channel. Nothing waits for it.

So epoch entry races the sync. A node restarted after being down for several epochs picks a
committee, a worker count and a node role from whatever it happens to know at that instant, and the
collector catches up underneath it.

## Proposed change

Add a public function to `crates/state-sync/src/epoch.rs` that drives the existing private collector
to quiescence before the epoch loop starts, and returns the epoch it reached.

Exit conditions, all four required:

1. Two consecutive passes return the same epoch.
2. A hard deadline elapses.
3. A per-pass cap on work is hit without database progress.
4. Shutdown is signalled.

It returns immediately when no peers are connected. It never fails the node: a gate that cannot
reach the tip returns what it has and lets startup continue.

### Quiescence is equality, not ordering

Test consecutive passes for equality, not for "did not advance". `collect_epoch_records` returns
`epoch.saturating_sub(1)` on three paths: a failed save (`epoch.rs:71`), an invalid downloaded
record (`epoch.rs:101`), and a missing parent anchor (`epoch.rs:107`).

A pass can therefore move backwards while it repairs. Treating a backwards move as quiescence ends
the gate early, precisely on the node that most needs it.

### The per-pass cap is not optional

An earlier review of this design found the gate burning its full deadline on every simultaneous
cohort boot: peers are connected, so the zero-peer short-circuit does not fire, but no peer is
serving epoch certs yet, so every pass returns the same epoch and the gate waits out its deadline
before proceeding. That was rated high severity. Ship the cap from the start.

## The ordering constraint, which is the hard part

The gate needs the dummy epoch-0 record as an anchor. `validate_downloaded_record` returns
`NoAnchor` when the previous record is missing (`epoch.rs:104-108`), and the collector then walks
backwards instead of forwards. The dummy is currently seeded inside the epoch loop, at
`crates/node/src/manager/node/run_epoch.rs:237`, which is after the gate would run.

Moving that seed earlier is not free. The startup re-vote hook added by #1198 states in its own
contract (`crates/node/src/manager/epoch_votes.rs:517-520`) that it may treat any stored epoch-0
record as a genuine closed-epoch record *because* `run_epochs` seeds the dummy after the hook runs
and `save_dummy_epoch0` never persists it. Seed the dummy before that hook and the hook re-arms a
vote round for a record that can never be certified.

The resolution is an ordering, not a code change:

1. `spawn_epoch_vote_collector` (currently `node.rs:867`)
2. `revote_uncertified_epoch_record_on_startup` (currently `node.rs:880`)
3. seed the dummy epoch-0 record
4. run the gate
5. seed the node mode (see #1350)
6. `spawn_epoch_record_collector` (currently `node.rs:858`) last, so it does not race the gate's
   first collection

If that ordering turns out not to hold, the fallback is to teach the re-vote hook to skip the dummy.
A predicate already exists: `EpochRecordDb::contains_dummy_epoch0`
(`crates/storage/src/epoch_records.rs:782`). It is store-wide rather than record-scoped, so it would
need care.

## Two other interactions

The record validation path was rewritten by #922 and #937 and now anchors downloaded records to the
locally trusted committee. The gate's safety argument has to be derived against that code, not
carried over from an earlier design.

Two guards deliberately refuse to mistake the dummy record for a real one, at
`crates/node/src/manager/node/close_epoch.rs:448` and `run_epoch.rs:1014`. Moving when the dummy is
seeded must not disturb either.

## Risk

High, the highest in this group. It changes when a node is allowed to enter its first epoch, and it
moves a seed that another component's documented contract depends on.

Two shapes to exercise: a cold cohort booting simultaneously, which is what produced the earlier
high-severity finding, and a single node booting alone with no peers, which drives the zero-peer
short-circuit.

## Acceptance criteria

- A public gate in `crates/state-sync/src/epoch.rs`, exported from the crate root.
- Quiescence tested by equality of consecutive passes.
- A per-pass cap, a hard deadline, a zero-peer short-circuit and a shutdown arm, all four present.
- The gate never returns an error that fails startup.
- The startup sequence follows the ordering above, with a comment recording why.
- The #1198 contract at `epoch_votes.rs:517-520` is still true after the move, or is updated to
  match what is actually true.
- The guards at `close_epoch.rs:448` and `run_epoch.rs:1014` still behave as before.
- A cold cohort boot does not burn the full deadline.
- Workspace `cargo check`, `clippy` and `nextest` pass, plus the restart and epoch end-to-end
  suites.

MavenRain:
Closes #1356.

## Problem

A restarted node can enter its first epoch while the background epoch record collector is still catching up. Its initial role can therefore use stale committee information. No attacker is required.

## Changes

- **Bound startup collection:** export `sync_epoch_records_to_tip`, using the existing collector and locally anchored record validation. Stop after two equal consecutive pass results, shutdown, a 30-second deadline, or a 5-second pass cap without certified-prefix progress. Skip collection when no peers have established connections.
- **Preserve recovery ordering:** subscribe the vote collector, re-vote durable records, seed dummy epoch 0, run the gate, seed node mode, then start background record collection. The startup re-vote hook still cannot see the uncertifiable dummy.
- **Cover stalled and recovering nodes:** add deterministic scheduling tests and public-gate tests for syncing from the dummy anchor and booting without established peers. Backward pass results do not establish quiescence, and certificate backfill counts as progress even when the latest record is unchanged.

## Testing

- Passed: `cargo +nightly-2026-03-20 fmt --all -- --check`.
- Passed: all 12 epoch tests on Rust 1.94, including the 10 new regressions and both existing record-validation tests, using this isolated-target command:

```sh
CARGO_TARGET_DIR=.validation/target \
CARGO_PROFILE_DEV_DEBUG=0 CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 \
testledger --force -- cargo +1.94 test --locked -p state-sync --lib epoch:: -j 2
```

- **Mutation confirmation:** six mutations of added scheduling lines were rejected in a harness using the unchanged production driver and repository tests. An inverted peer-readiness condition was then rejected by both public-gate tests in the full crate. Every new regression test was observed to fail under a mutation; the restored implementation passed all 12 epoch tests.
- Static review confirms the existing downloaded-record validation and dummy-record guards are unchanged.
- `diffclass`, `apidelta`, and `buildplan` require the full workspace validation scope. Workspace check, clippy, nextest, and the restart/epoch end-to-end suites remain pending the separate full-suite attestation. Current CI deliberately skips compile/test lanes for maintainer PRs, and `verify-on-chain` confirms this commit is not yet attested.


sstanfield (dismissed):
Lot of changes in some finicky code but I don't see any obvious problems and having the epoch record DB pre-loaded seems mostly upside.

crates/state-sync/src/epoch.rs
```diff
@@ -13,6 +14,108 @@ use tracing::{debug, error, info};
...
+    peer_count: impl Future<Output = usize>,
+    collect: impl FnMut(Epoch) -> Collect,
+    progress: impl FnMut() -> Progress,
+    shutdown: impl Future<Output = ()>,
+) where
+    Collect: Future<Output = Epoch>,
+    Progress: Future<Output = Epoch>,
+{
+    let sync = async {
+        if peer_count.await > 0 {
```
grantkee: I think it's better to poll here. If a node restarts and hasn't connected with any peers yet, this will skip which seems very likely in WAN setting. The e2e test replaced a 2s sleep with 45s too, which makes me think this has a high chance of being 0, especially in production

crates/node/src/manager/node.rs
```diff
@@ -1006,6 +990,50 @@ where
         )
         .await;
 
+        // Startup ordering is load-bearing: subscribe the vote collector, re-vote durable records,
```
grantkee, line 1028: This block gets the startup ordering right, but nothing downstream reads the records it just synced when the node's role is chosen, so identify_node_mode re-decides that role from this node's own execution state a moment later and overwrites the seed.

First, credit where it's due: the gate already buys something neither issue mentions — syncing records to tip is what makes the epoch N-1 record present for `resolve_prior_epoch_record`, which hard-errors on its absence at `crates/node/src/manager/node/run_epoch.rs:984-990` (called from `:529`). What I'd like to talk through is the follow-on work. The flow I have in mind:

1. The node checks canonical EVM state. If the tip is block 0, assume genesis. Otherwise, check whether this node is in the committee.
2. If it is in the committee, contact other validators and concurrently fetch epoch records.
3. Once epoch records are synced: if the node is not in the committee named at the epoch-record tip, continue as an observer.
4. Otherwise, assume validator status based on the initial canonical-tip check.

Today the role is decided by `identify_node_mode` (`crates/node/src/manager/node/start_epoch.rs:860-898`), which computes membership from `consensus_config.in_committee()` at `:870-873` (`crates/config/src/consensus.rs:366`) — derived from this node's OWN execution state — and writes it unconditionally with `send_modify` at `:895`. Called as the first statement of `create_consensus` (`:208`) from `run_epoch` (`run_epoch.rs:321`), it has no notion of the epoch-record committee at all: `EpochRecord`, `latest_record` and `get_committee_keys` appear nowhere in it, and `self.consensus_chain` appears only as an argument forwarded into `state_sync::prime_consensus` at `:879`. So the precedence today is the inverse of step 3 — the on-chain check inside the epoch loop overwrites whatever the epoch-record tip said at startup, and the only value that survives is the sticky `CvvInactive` early return at `:865-868`, which `seed_node_mode_on_startup` can never write (`crates/node/src/manager/node.rs:1605-1613`).

Worth being explicit that the seed reads the right thing: `latest_record()` (`node.rs:1604`) tested against `next_committee` (`:1607`), and since a stored record for epoch N closes epoch N, `next_committee` is the committee of the epoch now running — the same convention `get_committee_keys` uses for its fallback (`crates/storage/src/epoch_records.rs:838-841`). The input is correct; the precedence is what doesn't match. Concretely: a node down for several epochs restarts with its execution tip at epoch E and syncs records to epoch T greater than E. If it was seated at E but rotated out by T, the on-chain check says in-committee and the epoch-record tip says not — and today the on-chain answer wins, where step 3 wants the record tip to win.

One hazard before step 1 gets built: a membership check on the canonical-tip path has to route through `epoch_state_at_epoch_start_from_tip`, not `epoch_state_from_canonical_tip`. The doc at `crates/tn-reth/src/env/epoch.rs:208-212` says the committee arrays a tip read returns mutate mid-epoch — a governance `burn` swap-and-pops the ejected validator out of the CURRENT epoch's stored committees immediately — so tip reads before and after a burn disagree, and `epoch_state_from_canonical_tip` is flagged "not committee-deterministic" at `:216-217`. Two honest nodes checking either side of a burn would classify themselves differently.

The cheapest way to reach step 1 is already on the safe side of that. `get_committee_with_epoch_start_info` uses the pinned read (`start_epoch.rs:157`) and is already called at `node.rs:998`, just scoped inside the `if !contains_epoch(0)` block, with `self.key_config.primary_public_key()` in scope three lines later at `:1024`:

```rust
// hoist out of the `if !contains_epoch(0)` block
let (committee, ..) = self.get_committee_with_epoch_start_info(&engine).await?;
let seated_on_chain =
    committee.authority_by_key(&self.key_config.primary_public_key()).is_some();
```

`authority_by_key` is at `crates/types/src/committee.rs:857`, so no new dependencies. The one cost to weigh: that function hard-halts on both `StateReadError` classes by design (policy at `start_epoch.rs:134-142`, halt at `:160-165`), so making the call unconditional makes every startup inherit that halt.

Two smaller notes. There's no `is_genesis` helper anywhere in the tree — `canonical_tip().number == 0` (`crates/tn-reth/src/env/helpers.rs:289`) is the only existing expression of it, nearest precedent being `check_restore_consistency` at `node.rs:1576`, whose doc at `:1567` reads "Genesis / fresh nodes (`tip.number == 0`) are exempt." And at epoch 0 the committee doesn't come from chain state at all: `create_committee_from_state` loads it from `committee.yaml` at `start_epoch.rs:366-372`, only `epoch > 0` building from on-chain validators — relevant since step 1 treats block 0 specially. Step 2's "contact other validators" already exists as `wait_for_network_peers` (`start_epoch.rs:960-980`), but per-epoch only (`:673`, `:812`) and hard-erroring after ~2 minutes (`:968`), so a startup-scoped variant needs its own failure policy.

None of this has to land here — the ordering this PR establishes is the prerequisite for all of it, and reworking the precedence is more than #1356 asks for. Would you rather I open a follow-up issue for steps 1 through 4 and leave this PR as the sequencing fix it is?

crates/node/src/manager/node.rs
```diff
@@ -1006,6 +990,50 @@ where
...
+            );
+            let committee: Vec<BlsPublicKey> = committee.bls_keys().iter().copied().collect();
+            let next_committee = committee.clone();
+            let epoch_rec =
+                EpochRecord { epoch: 0, committee, next_committee, ..Default::default() };
+            // This in-memory anchor is neither persisted nor signed and is replaced at epoch close.
+            self.consensus_chain.epochs().save_dummy_epoch0(epoch_rec).await?;
+        }
+
+        let synced_epoch = state_sync::sync_epoch_records_to_tip(
```
grantkee: The startup epoch record sync is awaited inline here, so every healthcheck and metrics listeners can't bind until after this point

This PR's own head commit concedes the symptom. `git log -1 --format=%B 7abb593b` reads: "Startup epoch record synchronization can keep the observer from listening longer than that, so the first transaction hit connection refused." The patch went to the test — `crates/e2e-tests/tests/it/restarts.rs:485-489` now polls the observer's block-number endpoint for up to 45 seconds before submitting — not to the ordering that left the observer unreachable.

The budget is `STARTUP_SYNC_TIMEOUT`, 30 seconds (`crates/state-sync/src/epoch.rs:18`), enforced by the outer `select!` at `epoch.rs:109-116`. Three operator-facing things sit below it in `run()`:

- `HealthcheckServer::spawn` at `crates/node/src/manager/node.rs:1136`, behind `--healthcheck`.
- `tn_metrics::start_metrics_server` at `node.rs:1147`, behind `--metrics`.
- The worker JSON-RPC listener, reached from `run_epochs` at `node.rs:1225`:

```
run_epoch → create_consensus              run_epoch.rs:321
  spawn_worker_node_components            start_epoch.rs:249
    engine.initialize_worker_components   start_epoch.rs:471-479
      reth_env.start_rpc                  engine/inner.rs:182
        server_config.start(server).await tn-reth/src/env/rpc.rs:350
```

#1358 states the requirement this lands on: "A node that blocks on either timeout without serving RPC has failed even if it eventually recovers." A readiness or liveness probe pointed at the healthcheck port gets connection-refused for that whole window, and a probe whose failure threshold is under 30 seconds restarts the node straight back into the same gate. Load balancers and external RPC clients see the same refusal.

I think we should move the healthcheck and metrics binds above line 1012, so operator-facing endpoints are never behind the gate

crates/node/src/manager/node.rs
```diff
@@ -1006,6 +990,50 @@ where
...
+        info!(target: "epoch-manager", synced_epoch, "startup epoch record sync finished");
+
+        seed_node_mode_on_startup(
+            self.consensus_chain.epochs(),
+            self.consensus_bus.node_mode(),
+            &self.key_config.primary_public_key(),
+        )
+        .await;
+
+        state_sync::spawn_epoch_record_collector(
```
grantkee, line 1088: nit: the rustdoc on EpochManager::run still describes the pre-PR startup ordering, so the function now documents a sequence it no longer performs.

The real location is node.rs:778-789, which this PR doesn't touch and so has no commentable diff line; I'm anchoring at this spawn because it's the step the doc now places wrongly.

The clause reads "...schedule process-lifetime bootstrap dials, spawn the epoch-record and vote collectors, restore execution state (`try_restore_state`), and spawn the engine-update task." Two parts of that are now false. The collectors are no longer spawned together: the vote collector stays at `:974` while the record collector moved to `:1028`, behind both the sync gate and the node-mode seed. And the two new steps in between, the dummy epoch-0 anchor at `:997-1010` and the blocking `sync_epoch_records_to_tip` at `:1012`, aren't mentioned at all.

Housekeeping rather than a defect: issue #1356 asked for one thing here, "the startup sequence follows the ordering above, with a comment recording why", and the new inline comment at `:993` is exactly that. The rustdoc directly above it still states the old ordering, so `run` now carries two comments describing two different sequences.

Replacement for the stale clause:

```
/// listeners, schedule process-lifetime bootstrap dials, spawn the vote collector, re-vote
/// durable epoch records, seed the in-memory epoch-0 anchor, run the bounded startup
/// epoch-record sync (up to `STARTUP_SYNC_TIMEOUT`, skipped when no peers are established),
/// seed the node-mode watch, then spawn the background epoch-record collector and the
/// engine-update task.
```

I walked the body against that and the order holds. `STARTUP_SYNC_TIMEOUT` is private to `tn-state-sync` (epoch.rs:18, 30s), so keep it plain code text, not an intra-doc link.

Pre-existing and optional: `try_restore_state` runs at `:875`, before `spawn_node_networks`, not after the collectors where the doc puts it. Already imprecise before this PR, so fix it only if you are in the paragraph anyway.

crates/e2e-tests/tests/it/restarts.rs
```diff
@@ -483,7 +483,10 @@ fn wait_for_block(node: &str, target_block: u64) -> eyre::Result<()> {
 /// Run some test to make sure an observer is participating in the network.
 fn run_observer_tests(client_urls: &[String; 4], obs_url: &str) -> eyre::Result<()> {
     network_advancing(client_urls)?;
-    std::thread::sleep(Duration::from_secs(2)); // Advancing, so pause so that upcoming checks will fail if a node is lagging.
+    // The observer may still be syncing startup epoch records after the validators are ready.
+    wait_until_blocking(Duration::from_secs(45), "observer RPC ready", || {
```
grantkee, line 532: The 45 second timeout here is an attempt count rather than a wall-clock deadline, so if the observer never binds its RPC this wait takes about three hours to fail instead of 45 seconds.

`wait_until_blocking` (`crates/test-utils/src/wait.rs:72-92`) folds over `attempts_for(timeout)` (`wait.rs:23-25`) — `timeout.as_millis() / POLL_INTERVAL.as_millis()`, with a 10ms `POLL_INTERVAL` (`wait.rs:20`), so 4500 attempts here. `grep -n Instant crates/test-utils/src/wait.rs` returns nothing; there is no clock in the file. The duration is only honest if each attempt is free.

This one isn't. `get_block_number` (`common.rs:526-529`) reaches `get_block` (`common.rs:306`), which calls `call_rpc(node, "eth_getBlockByNumber", params, 10, ...)` on line 320, and `call_rpc_inner` (`common.rs:346-362`) sends the request then retries ten more times, sleeping 250ms before each on a fresh client with a 10s request timeout. A refused connection returns immediately, so each attempt costs the backoff: 10 × 250ms = 2.5s. 4500 × 2.5s = 11,250s, about 3h 8m, plus 45s of poll intervals. It terminates rather than hangs — just far past the number in the call. Behind that, `common.rs:322-326` retries 30 times at 1s each on a null result, ~52.5s more per attempt — not reached when the connection is refused, since `?` on line 320 fires first, but a syncing observer answering null hits it.

`Ok(get_block_number(obs_url).is_ok())` turns an error into "not ready yet" rather than an early exit — right for a readiness probe, but nothing short-circuits the fold.

This is the helper's behaviour, not something the PR introduced: `wait.rs` and `common.rs` are untouched by the diff and `network_advancing` (`common.rs:400-410`) has the same shape. It bites here because an observer that never brings up its RPC is exactly what these startup changes could cause, and CI would find out hours late.

Cheapening the probe realigns the count with the deadline; `restarts.rs:442-446` already passes `retries = 0` "so a known-down node fails immediately instead of burning ~5s of retries per check", and a no-retry block-number helper here would make 4500 attempts really cost 45 seconds. The fuller fix, worth its own issue since #1358 prescribes this helper for these loops, is a real deadline beside the attempt fold:

```rust
let deadline = Instant::now() + timeout;
(0..attempts_for(timeout))
    .fold(None, |resolved: Option<eyre::Result<()>>, attempt| {
        if resolved.is_some() || Instant::now() >= deadline {
            resolved
        } else {
            if attempt > 0 {
                std::thread::sleep(POLL_INTERVAL);
            }
            condition().map(|met| met.then_some(())).transpose()
        }
    })
    .unwrap_or_else(|| Err(eyre::eyre!("timed out waiting for condition: {description}")))
```

For the async twin, `tokio::time::Instant` keeps `start_paused = true` working.

crates/state-sync/src/epoch.rs
```diff
@@ -13,6 +14,108 @@ use tracing::{debug, error, info};
...
+                    let stalled = collected.is_none() && after == before;
+                    let next_epoch = collected.unwrap_or_else(|| after.saturating_sub(1));
+                    // Keep the outer deadline and shutdown responsive for immediately ready passes.
+                    tokio::task::yield_now().await;
+                    (!quiescent && !stalled)
+                        .then_some(((), (next_epoch, collected, collect, progress)))
+                },
+            )
+            .count()
+            .await;
```
grantkee: nit: the scheduling loop is spelled as a stream unfold over a four-element tuple whose count result is computed and discarded, and the Epoch it carries out through a parking_lot Mutex is only ever logged.

That Epoch travels out of the closures through the Mutex at epoch.rs:37, written at epoch.rs:42 and epoch.rs:56 and unwrapped at epoch.rs:66. I grepped for the return value: the only production consumer is the info! at crates/node/src/manager/node.rs:1019. Carrying it costs an extra latest_record() round trip inside progress() at epoch.rs:54-55, and progress() runs twice per pass (epoch.rs:88 and epoch.rs:96), so each pass makes four actor round trips of which two exist only to feed that log line. Those are oneshot hops to an in-process actor, so this is a readability point, not a latency one.

Two pieces here are deliberate and should stay. The higher-order generic split of drive_epoch_record_sync is intentional — the doc at epoch.rs:69-73 says "Separate scheduling from transport to exercise repair, stalled peers, and shutdown deterministically" — and the eight scheduling tests in crates/state-sync/src/epoch/startup_tests.rs inject collect, progress and shutdown through exactly that seam, which the PR body's mutation evidence rests on. The return value is asserted by startup_sync_downloads_from_dummy_anchor and startup_sync_zero_peers_preserves_dummy in crates/state-sync/src/epoch.rs, so dropping it is only worth it if those two assertions move to reading the store directly.

Narrow version: same scheduling, plain loop, no Mutex.

```rust
// drive_epoch_record_sync(.., mut collect: .., mut progress: .., ..)
let (mut last_epoch, mut previous_pass) = (0, None);
loop {
    let before = progress().await;
    let collected = tokio::select! {
        biased;
        _ = tokio::time::sleep(STARTUP_SYNC_PASS_TIMEOUT) => None,
        epoch = collect(last_epoch) => Some(epoch),
    };
    // A backward step repairs an anchor; only equality establishes quiescence.
    let quiescent = collected.is_some_and(|epoch| previous_pass == Some(epoch));
    let after = progress().await;
    let stalled = collected.is_none() && after == before;
    last_epoch = collected.unwrap_or_else(|| after.saturating_sub(1));
    previous_pass = collected;
    // Keep the outer deadline and shutdown responsive for immediately ready passes.
    if quiescent || stalled {
        break;
    }
}
```

The ordering there matches the current pass line for line: quiescent computed before the second progress() call, the yield ahead of the exit check. That yield is load-bearing — with a ready collect the biased select resolves without ever returning Pending, so without it the sync future never yields back and the deadline and shutdown arms at epoch.rs:109-116 could not fire. Two knock-ons: collect and progress need mut in the parameter list, and the futures::StreamExt import at epoch.rs:6 goes with the last stream use.

grantkee (changes requested):
This is going in the right direction, but I think the peer count is likely always 0 still and would like to see a poll with a timeout for startup peers

crates/state-sync/src/epoch.rs
MavenRain: Updated in [2eca7238](https://github.com/Telcoin-Association/telcoin-network/commit/2eca7238fcb7c03c1cdf8eeaf164dfdba9349ece): startup sync now polls for peers every 500 ms within the shared 30-second deadline and shutdown guard. An initially empty peer set no longer skips synchronization.

crates/node/src/manager/node.rs
MavenRain: Updated in [2eca7238](https://github.com/Telcoin-Association/telcoin-network/commit/2eca7238fcb7c03c1cdf8eeaf164dfdba9349ece): startup now reads the pinned epoch-start committee (the configured committee at genesis) and starts committee dials before syncing. Both the initial role seed and `identify_node_mode` apply the latest applicable record's `next_committee` veto, so stale local execution state cannot re-promote a removed validator.

crates/node/src/manager/node.rs
MavenRain: Updated in [2eca7238](https://github.com/Telcoin-Association/telcoin-network/commit/2eca7238fcb7c03c1cdf8eeaf164dfdba9349ece): health and metrics now start immediately after the execution engine starts, before the startup sync gate. The existing readiness predicate is preserved.

crates/node/src/manager/node.rs
MavenRain: Updated in [2eca7238](https://github.com/Telcoin-Association/telcoin-network/commit/2eca7238fcb7c03c1cdf8eeaf164dfdba9349ece): rewrote `run()`'s startup-order documentation to match the implementation, including restore, vote collection/re-voting, the epoch-zero anchor, foreground sync, role seeding, and the background collector.

crates/e2e-tests/tests/it/restarts.rs
MavenRain: Updated in [2eca7238](https://github.com/Telcoin-Association/telcoin-network/commit/2eca7238fcb7c03c1cdf8eeaf164dfdba9349ece): observer readiness now uses a single RPC request with no retries, and both wait helpers enforce elapsed-time deadlines. The async wait cancels at its deadline; the blocking wait can overrun only by its current request (bounded to 10 seconds here).

crates/state-sync/src/epoch.rs
MavenRain: Updated in [2eca7238](https://github.com/Telcoin-Association/telcoin-network/commit/2eca7238fcb7c03c1cdf8eeaf164dfdba9349ece): replaced the four-element tuple and discarded pass counter with a named state, and removed the mutex and tip reads used only for logging. The stream-based driver preserves deadlines, shutdown, backward repair, quiescence, and cooperative yielding.

sstanfield (approved):

grantkee (approved):
Merge conflicts now, but all comments were resolved. Thanks @MavenRain !

sstanfield (approved):

PR #1389, opened 2026-09-16
fix(network-libp2p): late-joining nodes never learn committee JSON-RPC endpoints

### LINKED ISSUE #1388: Late-joining nodes never learn committee JSON-RPC endpoints, so observer tx forwarding is dead for the life of the process
## Summary

A node that joins an already-running network with a cold datadir never caches the committee's rpc-bearing worker records, so every transaction it accepts sits in its batch builder forever instead of being forwarded to a validator. This hits every observer or NVV that is not started at genesis; the node keeps following the chain and serving reads, so nothing looks wrong until someone submits a transaction through it. No CI lane catches it: the only e2e test that both starts an observer after an epoch boundary and submits a transaction through it is `test_observer_pack_imports_after_epoch_close`, which is `#[ignore]`d and matches neither the `test_restarts` nor the `test_epoch` filter that `durable-e2e.yaml` runs, and the other observer test (`crates/e2e-tests/tests/it/state_export_import.rs:429-433`) deliberately submits to a validator instead because, in its own words, that "needs no observer -> committee tx forwarding".

Verified against `origin/main` @ `f4a89ef8`. Not reported and not fixed: no open PR touches `crates/network-libp2p/src/peers/manager.rs`, and neither #1361 (wait for established peers) nor #1364 (bootstrap peers via YAML/CLI) changes the mechanism below.

## Reproduction

`test_observer_pack_imports_after_epoch_close` in `crates/e2e-tests/tests/it/sync.rs:33` starts four validators, waits for an epoch to close, then starts an observer on a cold datadir. The pack import it was written to guard works: the observer imports the epoch-0 pack and tracks the validators to the same height (epoch 6, consensus number 181, identical to node0 and node1). The final liveness check (`sync.rs:163-168`), a transaction submitted to the observer and confirmed on a validator, is what fails.

Build the node binary first and pass it in, so each run uses a fresh binary rather than a stale one:

```sh
make build-e2e-bin
TN_BIN_PATH="$(pwd)/target/e2e/telcoin-network" \
  cargo nextest run -p e2e-tests --test it --run-ignored all \
  test_observer_pack_imports_after_epoch_close
```

The observer's log shows what happened to the transaction:

```sh
grep -c "advertised a JSON-RPC endpoint" crates/e2e-tests/test_logs/observer_pack_import/node4-run0.log
# 36
```

36 warnings, zero successful forwards, across all six epochs the observer was up. `num_batches=` appears 1323 times across the four validator logs and every one of them is `num_batches=0`.

## Root cause

The worker swarm's `PeerManager::known_peers` holds two different things under one key: config-derived dial hints (a bootstrap stub with `rpc: None`, inserted at startup and pinned) and network-learned advertisement records (the peer's signed `NodeRecord`, which is where `rpc` comes from). Every path that decides whether to fetch or refresh a record asks "is there an entry?" when it means "have we learned this peer's advertisement?". A late joiner ends up with the stub, and the stub blocks the real record forever.

1. At process start, `add_bootstrap_peers` inserts a stub `NetworkInfo { .., rpc: None }` for every committee validator into each worker swarm and pins it. Call site: `crates/node/src/manager/node.rs:939`; handler: `crates/network-libp2p/src/consensus.rs:855-872` (the stub is built with a hard-coded `rpc: None` at `:867`); insertion: `manager.rs:1009` (`add_bootstrap_peer`). The primary's call site at `node.rs:898` states the ordering on purpose: "Register bootstrap peers before per-epoch committee updates resolve known peers."
2. Each validator pushes its real, rpc-carrying worker `NodeRecord` to the observer on first connect. The receiving path, `add_self_advertised_peer` (`manager.rs:1109`), only caches it when `self.peers.is_committee_member(&bls_key)` is already true (`manager.rs:1116`). Otherwise it falls into the `else if source == advertised` branch, which confirms the transport identity and discards the `NetworkInfo`, `rpc` included. The worker swarm's committee slots are written later, by `init_network_for_epoch` inside `spawn_worker_network_for_epoch` (`crates/node/src/manager/node/start_epoch.rs:787`), and `create_consensus` (`start_epoch.rs:197`) only gets there after the primary has finished its own `wait_for_network_peers` (`start_epoch.rs:673`). In the failing run the gap between the push and the committee write was 100 ms; structurally it spans the whole primary bring-up, so it is wider under load.
3. The push is one-shot. `published_to_peers` (`consensus.rs:328`) is a per-process LRU that records which peers we have already pushed to, and the next DHT republish is `kad_publication_interval`, 12 hours by default (`crates/config/src/network.rs:275`).
4. Both re-discovery triggers filter on presence in `known_peers`: `trigger_missing_authorities` at `manager.rs:960` and `find_authorities` at `manager.rs:1214`. The pinned stub is present, so `kad::get_record` is never issued for those keys again. `prune_known_peers` (`manager.rs:933`) keeps pinned entries across every committee rotation, so nothing ever evicts the stub either.
5. `current_committee_rpcs` (`manager.rs:1240`) returns an empty vector for the rest of the process. `disburse_txns` (`crates/consensus/worker/src/worker.rs:280`) warns `no committee validator has advertised a JSON-RPC endpoint`, and the batch builder keeps the transactions pending on an exponential backoff.

The doc comment on `current_committee_rpcs` encodes the wrong invariant: "Members whose record is known but carries no RPC info did not advertise one and are skipped without a re-fetch." A bootstrap stub is indistinguishable from "advertised nothing".

The hazard was already understood in one place. `kad_record_is_stale` (`manager.rs:1168`) exempts pinned entries specifically because an operator stub "would otherwise block the peer's real record (fresh multiaddrs, advertised rpc) forever". That guard sits downstream of the `is_committee_member` gate at `:1116`, which returns first, so it never gets the chance to do its job for a late joiner.

Cold vs warm datadir: `add_restored_peer` (`manager.rs:995`) runs inside `ConsensusNetwork::new`, before `add_bootstrap_peers`, and does carry `rpc`. A node with a warm kad store may not exhibit the bug because `add_bootstrap_peer` never overwrites an existing entry. It reproduces on a cold datadir, which is every fresh join.

## Evidence

Logs in `crates/e2e-tests/test_logs/observer_pack_import/`.

| Time (UTC) | Event |
|---|---|
| 20:12:20.13-.18 | Validators publish their worker records, each carrying `http://127.0.0.1:1652x/` |
| 20:12:30.026 | Observer starts, 10 s and one epoch boundary later |
| 20:12:30.4746-.4751 | All four validators push their rpc-bearing worker records to the observer |
| 20:12:30.575254 | Observer's worker swarm sets its committee slots, 100 ms too late |
| 20:12:31.0999 | First `no committee validator has advertised a JSON-RPC endpoint` |
| 20:13:23.889 | Last one; 36 warnings, 0 successful forwards, across all 6 epochs |

## Why tests didn't catch it

The unit tests cover the pinned-stub refresh, but only in the ordering that works. `test_pinned_peer_kad_record_bypasses_staleness_guard` (`crates/network-libp2p/src/tests/peer_manager.rs:1243`) calls `update_committees` first, then inserts the stub, then delivers the real record. The production ordering is stub, then record, then committee.

The swarm-level test says so outright. `test_advertise_rpc_via_kad` (`crates/network-libp2p/src/tests/network_tests.rs:3154`) seeds the committee before connecting, with this comment at `:3224-3227`:

> seed the nvv's committee (target + the rest) BEFORE it connects so target's rpc-bearing record is retained the moment it arrives via kad; the committee gate added in #827 drops records for non-committee keys, and a late seed misses the record's first delivery. an nvv following the chain learns the committee from epoch state.

The test works around the production ordering instead of reproducing it. "A late seed misses the record's first delivery" is this bug.

## Proposed fix

Three changes, all in the network layer, on branch `fix/observer-forward-rpc-discovery`. The first closes the window, the second makes the state self-healing, the third closes an adjacent gap that #1364 made reachable from operator config.

Cache self-advertised records from pinned peers (`manager.rs:1116`). Widen the cache condition in `add_self_advertised_peer` from `is_committee_member(&bls_key)` to also accept `pinned_peers.contains(&bls_key)`. A pinned key is operator-provisioned, so the set stays config-bounded and the #827 injection guard holds; the record's BLS signature is already verified in `peer_record_valid`, and the self-advertised branch already requires `source == advertised`. With the gate widened, the existing pinned exemption in `kad_record_is_stale` lets the real record replace the stub, which is what it was written for.

Make re-discovery ask "have we learned this record?" instead of "is there an entry?" (`manager.rs:960` and `:1214`). Add `stub_records: HashSet<BlsPublicKey>` to `PeerManager`, inserted by the three operator-provisioned writers (`add_trusted_peer_and_dial`, `add_known_peer`, `add_bootstrap_peer`) and removed by the three network-learned writers (`add_discovered_peer`, `add_self_advertised_peer`, `add_restored_peer`). A private `record_unlearned(&self, k)` returns true when the key is absent or still a stub, and both `trigger_missing_authorities` and `find_authorities` filter on it; prune it alongside `known_peers`. This deliberately does not chase a learned record that carries no rpc, so a validator that genuinely advertises nothing is still skipped and there is no kad churn; re-emission is already deduped in flight by #1140. The `current_committee_rpcs` doc comment gets corrected at the same time. One existing test has to change: `test_current_committee_rpcs_ignores_members_without_advertised_rpc` (`peer_manager.rs:1339`) seeds via `add_known_peer`, which is an operator stub, and asserts no re-discovery; it switches to `add_discovered_peer` so it keeps testing "a peer that genuinely advertised nothing".

Stop discarding the operator's advertised rpc (`consensus.rs:867`). The `AddBootstrapPeers` handler receives `BTreeMap<BlsPublicKey, P2pNode>`, and `P2pNode` carries `rpc: Option<RpcInfo>` (`crates/types/src/committee.rs:99`), but the handler hard-codes `rpc: None`. Pass `info.rpc` through. #1364 added YAML/CLI bootstrap config that deserializes the full `P2pNode` shape, so an operator following `docs/src/getting-started/bootstrap-peers.md` can set `rpc:` under a worker entry today and have it silently dropped. This is not sufficient on its own (the genesis ceremony leaves `worker.rpc` unset), so it does not replace the first two changes.

New unit tests cover the production ordering (stub, then push while the committee is empty, then `update_committees`; asserts `current_committee_rpcs()` returns the rpc, fails on current main), self-healing (stub plus committee, no push; asserts `PeerEvent::MissingAuthorities` for the stubbed key), and no churn (a learned record with `rpc: None` emits nothing). A late-joining variant of `test_advertise_rpc_via_kad` seeds the committee after connecting. No new e2e test: `test_observer_pack_imports_after_epoch_close` is already the end-to-end regression proof, and the assertion that matters is zero `advertised a JSON-RPC endpoint` warnings in `node4-run0.log` after the fix.

## Related

- #1367 (open): its body says the 48 h record lapse is "masked" today because "every first connection pushes the record directly ... and `known_peers` retains the resolution". For a late joiner that push is discarded, so the masking does not exist. PR #1371 (open) re-signs and republishes on `kad_publication_interval`, so at best it would mitigate this at the 12 h horizon.
- #1135 (closed) / #1140 (merged): same `!known_peers.contains_key` predicate, opposite symptom (duplicate kad queries rather than none).
- #905 (closed) / #906 (merged): introduced the pinned, never-overwrite bootstrap entries that now shadow the real record.
- #827 (closed) / #857 (merged): introduced the `is_committee_member` gate in `add_self_advertised_peer` that discards the first delivery.
- #904 (merged): introduced `current_committee_rpcs` and the doc comment that states the wrong invariant.
- #729 (closed): the rpc-advertisement feature this defeats for late joiners.
- #1132 (closed) / #1145 (closed) / #1133 (closed): the observer-forwarding lineage. #1132 attributes the "no advertised endpoint" case to `worker.rpc` being unset by the genesis ceremony, and #1145 describes the refusal persisting "for a whole epoch"; neither considers a node that received the advertisement and threw it away, which is why this looked like operator misconfiguration rather than a node bug.
- #714 (open): the `AllPeers` single-source-of-truth refactor; the stub/record split above would fold into it.

## Follow-ups not addressed here

The e2e catch-up gate is vacuous. `sync.rs:124-134` waits until the observer's `tn_latestConsensusHeader` number reaches the validator's, but that RPC returns `consensus_bus.last_consensus_header()` (`crates/node/src/lib.rs:114`), the "most recently seen" header fed by state-sync's fetch path (`crates/state-sync/src/consensus.rs:193`), not applied state. The wait guards nothing about what the observer has actually executed.

`is_epoch_complete` logs at ERROR on its normal path. `crates/storage/src/consensus.rs:1042` emits `DB error checking epoch completeness` whenever `consensus_header_by_number` returns `Err`, and for the current pack a header that simply is not there yet comes back as `PackError::ConsensusNumberTooHigh` (`consensus_pack.rs:1268-1271`). "Not complete yet" is the expected answer for a pack still being filled, so this line fires as an error during ordinary epoch-completeness checks.

grantkee:
A node that joins a running network with a cold datadir accepts transactions it can never forward. `AddBootstrapPeers` pins an rpc-less stub for every committee validator, each validator pushes its real record once on first connect, and that push lands before the worker swarm's committee slots are seeded, so the committee gate in `add_self_advertised_peer` discards it. From then on the stub satisfies every "is the record known?" check, no kad lookup is ever issued for the key, and the next republish is 12h out. `disburse_txns` logs `no committee validator has advertised a JSON-RPC endpoint` for the life of the process. `test_observer_pack_imports_after_epoch_close` fails on its final liveness check for exactly this reason.

- `add_self_advertised_peer` caches a pushed record for a pinned key even when no committee slot is seeded yet. The pinned set is operator config, so the #827 bound is unchanged.
- A new `stub_records` set on `PeerManager` separates config dial hints from network-learned records. `trigger_missing_authorities` and `find_authorities` chase a stub the same way they chase an unknown member. A learned record that carries no rpc is still left alone, so there is no new kad churn, and re-emission stays capped by the per-key in-flight dedup from #1140.
- The `AddBootstrapPeers` handler passes the operator's configured `rpc` through instead of hard-coding `None`. #1364 made that field reachable from YAML/CLI and it was being dropped silently.
- The kad staleness exemption is keyed on `stub_records` rather than `pinned_peers`. Pins are never cleared, so a pinned validator's learned record could be regressed by any connected peer relaying an older validly-signed copy. Pre-existing on `main`; surfaced in review because the first change widened it.

Tests: a unit test for the production ordering (stub, push, seed) that fails on `main`; self-healing and no-churn tests for `current_committee_rpcs` and `find_authorities`; a late-seed variant of `test_advertise_rpc_via_kad` with an assertion before the seed so it pins the admission change on its own; a bootstrap-config rpc test; replay regression tests on both the push and query paths. `test_current_committee_rpcs_ignores_members_without_advertised_rpc` was seeded through an operator stub, which is now chased by design, so it seeds through `add_discovered_peer` and keeps testing a learned rpc-less record.

Verification: `test_observer_pack_imports_after_epoch_close` against a binary built from `main` fails after 69s with 35 warnings in the observer log; against this branch it passes in 15s with 0. `make test-restarts` 4/4, `make test-epochs` 5/5, workspace 1828/1828.

Not in this PR: `add_discovered_peer` still admits committee keys only, so a pinned bootstrap peer seated only in epoch N+2 is queried by the prefetch and its result dropped (one wasted lookup per epoch per key); and an operator-configured bootstrap `rpc` is replaced by the peer's first learned record even when that record advertises none. Both are policy choices worth a separate discussion.

Related: #1367 assumes this mechanism is healthy; #1371 would only mitigate at the 12h republish horizon.

closes #1388


sstanfield (approved):

MavenRain (approved):
LGTM

PR #1390, opened 2026-09-17
feat(node): start all committee workers each epoch

### LINKED ISSUE #557: feat: Multi-Worker Support — PR 4: `EpochManager` Multi-Worker Loop

MavenRain:
Closes #557.

## Summary

Start every worker in the committee's on-chain worker range at epoch entry. Each worker gets its own pool, RPC listener, batch validator, batch builder, and network interface while sharing the primary and execution engine. One-worker committees keep the existing behavior.

## Changes

- **Epoch startup:** initialize workers sequentially by id, refresh each worker's epoch handle, and prefetch future committees on every active swarm. Each RPC server advertises its own worker identity and address.
- **Epoch rollover:** reuse existing pools and RPC listeners, refresh only the matching worker's peer-count task, and start a batch builder for every active worker.
- **Recovery:** drain the orphan cache once before new builders start and forward each orphan through its originating worker. Preserve the existing per-worker, chain-pinned accumulator and base-fee recovery.
- **Worker-count changes:** require enough configured swarms before entering consensus, allow a provisioned surplus, and refresh persistent pool fee handles when a removed worker is reactivated. Provision two local workers and their bootstrap addresses in the live count-change e2e fixture while retaining a one-worker genesis and the existing epoch-progress assertions.

The worker count remains a consensus input read from pinned chain state. A local configuration shortfall fails before participation, so validators cannot silently start different worker subsets. No peer-supplied value determines the number of workers started. The existing bootstrap path resizes the accumulator from pinned on-chain state before execution, and epoch startup reuses its per-worker fee handles.

## Testing

Validation used an isolated Cargo target directory. Formatting and the following Clippy checks passed under the pinned nightly:

```sh
cargo +nightly-2026-03-20 clippy --locked --keep-going -j 2 \
  -p telcoin-network -p telcoin-network-cli -p tn-batch-builder \
  -p tn-batch-validator -p tn-engine -p tn-executor -p tn-exex \
  -p tn-network-libp2p -p tn-node -p tn-primary -p tn-reth -p tn-rpc \
  -p tn-storage -p tn-test-utils -p tn-test-utils-committee -p tn-types \
  -p tn-worker --no-deps -- -D warnings
cargo +nightly-2026-03-20 clippy --locked -j 2 \
  -p tn-node -p tn-reth --all-targets --no-deps -- -D warnings
```

The default-feature worker regression suite passed: 30 passed, with one pre-existing ignored test.

```sh
cargo +1.94 test --locked -p tn-node -j 2 --no-fail-fast worker -- --test-threads 2
```

Confirmed all three new tests fail under in-diff regressions: a global initialization check, acceptance of too few local swarms, and removal of the retained pool fee update. Restored both source files byte-for-byte and forced the worker-filtered suite to execute again: 30 passed, with one pre-existing ignored test.

The live worker-count fixture correction passed the all-target e2e compile check and its focused test (1 passed in 32.3 seconds). The test used a newly built Rust 1.94 node executable with `tn-storage/test-utils`, selected through `TN_BIN_PATH`.

```sh
cargo +1.94 check --locked -p e2e-tests --all-targets -j 2
cargo +1.94 test --locked -p e2e-tests --test it -j 2 \
  epochs::test_epoch_worker_count_changes_keep_nodes_running -- --ignored --exact
```

The 21-package Clippy command was also attempted with `--all-targets`. That broader check reports 131 pre-existing test lints in `e2e-tests`, `tn-engine`, `tn-executor`, `tn-primary`, `tn-storage`, and `tn-network-libp2p`. No lint suppressions or CI changes are included.

Full workspace attestation remains outstanding. The initial maintainer run reported 23 passing e2e tests and a failure in the count-change fixture, which had provisioned only one worker before activating two. The fixture now supplies the capacity required by the epoch-entry check. A maintainer must run `make attest` on the updated PR head, then request a review to rerun verification.


sstanfield (approved):

grantkee (approved):
LGTM - I filed a few follow-up issues we should address before fulling supporting multi-worker lanes. Good work @MavenRain

sstanfield (approved):

grantkee (approved):

PR #1392, opened 2026-09-17
fix(node): validate primary key and correct worker listener identity

### LINKED ISSUE #1374: node: primary network_key in node info is never validated against KeyConfig; worker-0 listener override stamps the primary key
Found while reviewing #1343 (per-worker network key validation, closes #1337). Line numbers are against `a3f0d001`.

## Problem

#1343 makes `prepare_worker_networks` reject a `node_info.p2p_info.workers[i].network_key` that does not match `key_config.worker_network_public_key(i)`. The same gap is still open for the primary.

`node_info.p2p_info.primary.network_key` is read through `ConsensusConfig::primary_networkkey()` (`crates/config/src/consensus.rs:357-359`) and used in exactly two places, both to stamp a `/p2p/` component onto a listener env override in `parse_listener_address_for_swarm`:

- `crates/node/src/manager/node/start_epoch.rs:656` — `PRIMARY_LISTENER_MULTIADDR`
- `crates/node/src/manager/node/start_epoch.rs:857` — `WORKER_LISTENER_MULTIADDR` (worker 0)

Nothing compares it with `key_config.primary_network_public_key()` (`crates/config/src/keys.rs:501`). The swarm's actual identity comes from the key config (`ConsensusNetwork::new_for_primary` takes `key_config`; `start_epoch.rs:233` stamps the derived key into the RPC `NodeInfo`), so a mismatched config field is silently ignored by the node itself while still being what the operator hands to genesis and to peers. That is the same failure shape #1337 described for workers: swarm identity is correct regardless, the advertised identity is wrong, and nothing says so at startup.

A related wrinkle on the second call site: `start_epoch.rs:857` passes the *primary* key when building the worker-0 listener address. libp2p-quic parses and discards a trailing `Protocol::P2p` on `listen_on`, so it does not affect binding, but it is inconsistent with `keytool generate`, which writes the worker key there (`crates/telcoin-network-cli/src/keytool/generate.rs:202-215`), and with the `/p2p/` the worker's `NodeRecord` advertises.

## Fix

Add the same check for the primary in `spawn_node_networks` before the primary swarm is constructed (`crates/node/src/manager/node.rs`, next to the `prepare_worker_networks` call):

```rust
let expected = self.key_config.primary_network_public_key();
eyre::ensure!(
    self.builder.tn_config.node_info.p2p_info.primary.network_key == expected,
    "node config `node_info.p2p_info.primary.network_key` does not match the key derived \
     from the loaded BLS keystore (expected {expected}, found {}): update this entry",
    self.builder.tn_config.node_info.p2p_info.primary.network_key
);
```

`{expected}` assumes the `Display` impl for `NetworkPublicKey` proposed on #1343 (Base58 of the protobuf encoding, matching the YAML form); with `{:?}` the operator gets libp2p's unpadded hex, which cannot be pasted back into `node-info.yaml`.

And at `start_epoch.rs:857`, pass `consensus_config.key_config().worker_network_public_key(*worker_id)` instead of `primary_networkkey()`.

A `check_primary_network_key_*` unit test alongside the `prepare_worker_networks_*` tests would cover the first change; the second is exercised by any e2e run that sets `WORKER_LISTENER_MULTIADDR`.

MavenRain:
Closes #1374.

## Problem

A stale `node_info.p2p_info.primary.network_key` lets startup proceed with an advertised identity that differs from the loaded BLS keystore. No attacker is required: the swarm derives the correct key while peers and genesis configuration can retain the wrong one.

`WORKER_LISTENER_MULTIADDR` also uses the primary key when appending or checking its `/p2p/` suffix, so an override containing the correct worker peer ID is rejected.

The delayed restart tests also depend on observing the transient `CvvInactive` mode. Catch-up can finish before an RPC poll sees it. The lagged test then expects the old balance after catch-up, even though the missed transfer may already be applied.

## Changes

- **Validate the primary identity before network construction.** Compare the configured key with `KeyConfig::primary_network_public_key()` before worker preparation or either swarm is created. On mismatch, identify the config field and display expected and configured keys in the same Base58 representation used by configuration.
- **Use worker 0's derived identity for its listener override.** Preserve the existing worker-0-only override and the configured addresses for other workers.
- **Add deterministic primary-key regression tests.** Cover matching keys, worker keys, primary keys from another keystore, and the operator-facing error representation.
- **Require durable catch-up evidence in delayed restart tests.** Enable metrics on the restarted process and wait for applied state-sync headers plus `CvvActive`. The fresh process's counter cannot contain sync activity from before the restart. Reuse the existing metrics scraper.
- **Check the lagged node's old balance before shutdown.** Retain the post-restart assertion that the missed transfer reaches the expected balance. Add regressions for catch-up finishing before the first poll, missing sync evidence, and a node that has not returned to active consensus.

## Testing

- Passed `cargo +nightly-2026-03-20 fmt --all -- --check` through `gateledger`, using the repository's `rust-nightly` pin. `git diff --check` also passed.
- The `diffclass -> apidelta -> buildplan clippy` ladder identifies the change as executable Rust and retains 14 packages for scoped clippy.
- Passed `cargo +1.94 test --offline --locked -j 2 --lib` through `testledger` in an isolated source-extraction harness: 3 tests passed. The harness executes both new repository tests plus the actual worker listener expression and parser with real BLS/libp2p keys. It covers an unset override, appending the worker identity, accepting the worker suffix, rejecting a primary suffix, and preserving other workers' configured addresses. Its surrounding `KeyConfig` and listener context are reduced; this does not compile the full `tn-node` crate.
- Confirmed all three mutations fail their intended test: remove the primary guard, invert its comparison, and restore the primary key in the worker listener call. Each mutant compiled successfully before its regression assertion failed.
- Passed the same pinned `cargo +1.94 test --offline --locked -j 2 --lib` command through `testledger` in a separate harness containing the exact restart predicate, `NodeMode`, and three new repository tests. All three tests pass. Mutating the required mode back to inactive, allowing zero applied headers, or removing the active-mode requirement each fails its intended test. This validates the extracted helpers, not the complete E2E binary.
- Started the complete 14-package clippy plan with `--all-targets --no-deps`; stopped it during the cold dependency build. Clippy and the full node/workspace/e2e suites are not claimed as passed.
- The reported `make attest` run passed 23 E2E tests and failed `test_restarts_lagged_delayed`, with four skipped. The revised restart checks still require a full E2E rerun.
- Full repository validation and exact-head attestation remain required. Maintainer-authored PRs skip the Rust CI lanes under the current workflow. This PR remains in draft until that validation is available.


sstanfield (approved):

grantkee (approved):
LGTM!

PR #1410, opened 2026-09-19
Migrate adiri's Safe stack to the canonical v1.4.1 suite at a fork boundary

### LINKED ISSUE #1409: Safe on adiri lands at non-canonical addresses, breaking cross-chain Safe parity
## Problem

Adiri's Safe stack was deployed by compiling `safe-contracts` with this repo's toolchain (solc 0.8.26) instead of using the canonical v1.4.1 release bytes (solc 0.7.6). The result is semantically equivalent but byte-different code at the canonical addresses, and that difference is load-bearing: `SafeProxyFactory.createProxyWithNonce` derives the proxy address via CREATE2 over the factory's **embedded proxy creation code**. Different factory bytes mean every Safe created on adiri lands at a different address than the identical creation call on Ethereum, Sepolia, Base, or any other chain Safe supports. A user's counterfactual Safe address does not carry onto Telcoin Network.

The gap is wider than the factory. Adiri carries only three of the Safe deployments — the `Safe` singleton, `SafeProxyFactory`, and the governance proxy. The other ten contracts in the official safe-deployments v1.4.1 registry are absent: `SafeL2`, `CompatibilityFallbackHandler`, `SafeToL2Setup`, `MultiSend`, `MultiSendCallOnly`, `SignMessageLib`, `CreateCall`, `SimulateTxAccessor`, both migration helpers, and the Safe Singleton Factory that deployed all of them on live chains. Safe ecosystem tooling — the protocol-kit SDK, Safe{Wallet}, Safe Transaction Service — assumes those singletons and handlers exist at their canonical addresses with canonical behavior. Several of them bake `address(this)` into their runtime bytes as an immutable and guard on it, so they are only valid at the canonical address, which is exactly where adiri has nothing.

The governance Safe itself sits on the L1 `Safe` singleton with no fallback handler set. Mainnet genesis puts governance on `SafeL2`, because `SafeToL2Setup` routes every counterfactual Safe on a non-mainnet chain onto SafeL2 and SafeL2's transaction events are what Safe Transaction Service indexes. Adiri therefore diverges from mainnet on the one account governance actually runs through.

None of this is fixable by a transaction the chain can send. The sanctioned in-Safe path, `SafeToL2Migration.migrateToL2`, carries an `onlyNonceZero` guard that runs after `execTransaction` has already incremented the nonce — it must be the Safe's first transaction ever. Adiri governance is at nonce 2, so no transaction its owners can ever sign performs the migration. `SafeMigration.migrateL2Singleton` has no nonce guard, but it leaves the missing suite and the recompiled factory in place, so counterfactual addresses stay wrong. Leaving adiri as-is means the testnet cannot be used to rehearse anything involving Safe addresses, and the divergence from mainnet genesis is permanent.

## Solution

Introduce an epoch-gated in-protocol fork, `GOVERNANCE_SAFE_FORK_EPOCH`, that brings live adiri's Safe stack to parity with mainnet genesis in a single deterministic commit. It should fire from the epoch-closing block that concludes `GOVERNANCE_SAFE_FORK_EPOCH - 1` — a one-shot `==` trigger, the same shape as the existing `CONSENSUS_REGISTRY_FORK_EPOCH` — and should ship dormant at `u32::MAX`, armed later by a dedicated epoch-setting PR once the gate-capable build is fleet-wide.

The boundary should do three things. **Etch** the ten canonical contracts adiri lacks plus SafeL2, using the byte-exact runtime bytes captured from live mainnet deployments and vendored in `tn-contracts`, seeding SafeL2's `threshold` slot to 1 so nobody can `setup()`-hijack the fresh singleton, and clamping each account's nonce up to the EIP-161 contract value that mainnet genesis carries. **Swap** the two recompiled deployments — the `Safe` singleton and `SafeProxyFactory` — to canonical bytes, code-only so balance, nonce, and all storage survive; the v1.4.1 storage layout is identical between the recompile and the canonical build, so the preserved slots stay valid. **Migrate** the governance proxy onto SafeL2 by writing exactly two slots, the singleton slot and the fallback-handler slot, leaving owners, threshold, the Safe nonce, and the TEL balance untouched by omission.

Every write must be a pure function of committed state so the whole fleet passes or aborts in lockstep and re-derives a byte-identical state root. The two swap targets and the governance proxy should be gated fail-closed on pinned pre-fork code hashes, since those paths preserve storage whose layout only the pin vouches for; each gate should additionally accept its own post-fork end state so re-executing the boundary is a no-op rather than a halt. The etch targets and the fallback-handler slot are addresses and slots the fork itself defines, so they should force the canonical value and record any displaced occupant at `warn!` rather than stalling the fleet on an epoch-closing block. All canonical hashes belong in a pinned table in `tn-types::forks` with a test tying each row to its vendored file, so that a `tn-contracts` bump changing a vendored byte fails CI instead of silently changing historical state roots after the fork has run.

The whole mechanism should be `adiri`-gated — mainnet genesis already carries the full canonical suite with governance on SafeL2, so non-adiri builds have nothing to migrate and should compile it out entirely. That implies a dedicated e2e lane building an `adiri` node binary, since the default e2e binary cannot execute the fork at all; arming the fork's environment override on any other lane should be a named test failure rather than a silent no-op.

grantkee:
- Adds `GOVERNANCE_SAFE_FORK_EPOCH` (dormant at `u32::MAX`) and `apply_governance_safe_fork`, firing once from the epoch-closing block that concludes `FORK_EPOCH - 1`. It etches the eleven canonical contracts adiri lacks, swaps the recompiled `Safe` singleton and `SafeProxyFactory` to canonical bytes code-only, migrates the governance proxy's singleton and fallback-handler slots onto SafeL2, and marks the singleton factory's deployer EOA at nonce 1 to close the last mainnet-genesis parity leaf. The factory swap is what restores cross-chain counterfactual Safe addresses — `createProxyWithNonce` CREATE2s over the factory's embedded proxy creation code, so the recompile is why adiri's addresses diverge.
- Gates that preserve storage (`Safe`, `SafeProxyFactory`, the governance proxy's slot 0) fail closed on pinned pre-fork hashes and additionally accept their own post-fork end state, so re-execution is a no-op rather than a fleet-wide abort. Gates over addresses and slots the fork defines — the etch targets, the handler slot — force the canonical value and `warn!` on a displaced occupant instead of halting the epoch-closing block.
- `GOVERNANCE_SAFE_FORK_CANONICAL_SUITE` in `tn-types::forks` pins all 13 addresses and code hashes, with a default-feature CI test hashing each vendored `tn-contracts` file against its row. Callers resolve rows by name (`governance_safe_fork_canonical_address`), since row order is coupled to `tn-reth`'s vendored bytecode list.
- Bumps `tn-contracts` to the canonical Safe suite and regenerates mainnet `genesis.yaml` / `committee.yaml`: governance moves to the canonical `SafeProxy` bytecode on the SafeL2 singleton, the full 12-contract v1.4.1 registry plus the Safe Singleton Factory are predeployed, and every contract account carries the EIP-161 nonce-1 value it has on Ethereum. Precompile account count in `test_precompile_genesis_accounts` goes 9 → 25.
- The whole mechanism is `#[cfg(feature = "adiri")]`, including its `fork_epoch_overrides` row — a non-adiri build stays silent about a pin it cannot honor rather than advertising a schedule change that never happens. `make test-e2e-governance-safe` builds a separate `adiri` e2e binary into its own target root and runs the fork across a live boundary on spawned nodes; it is the only lane where `TN_GOVERNANCE_SAFE_FORK_EPOCH` is not inert, and arming that variable elsewhere now fails by name instead of quietly doing nothing.
- Why a protocol fork rather than a governance transaction: `SafeToL2Migration.migrateToL2` guards `onlyNonceZero` after `execTransaction` increments, so it must be the Safe's first transaction ever. Adiri governance is at nonce 2 — no signable transaction can perform the migration, and `SafeMigration.migrateL2Singleton` would leave the missing suite and the recompiled factory in place.

closes #1409


MavenRain (approved):
All solid

